PR #19, opened 2015-05-28
Ignore directories created by editors

sethladd:


sethladd:
cc @mkustermann

mkustermann:
Generally speaking, I agree with kevin on this - everybody uses different editors/tools, so the list will never be complete, the repo specific gitignore is for repo-specific files (e.g. build files). But I don't mind your commit. All is good :)

PR #42, opened 2015-10-13
fix #41: meta description tag set with package description

### LINKED ISSUE #41: meta description tag in html package page should be the package description
When sharing a package page such as https://pub.dartlang.org/packages/test (on slack, facebook), the title is correct (i.e. it is the package title), but the description is always "Pub is a package manager for the Dart programming language."

````html
<title>test 0.12.5 | Pub Package Manager</title>
<meta name="description"
  content="Pub is a package manager for the Dart programming language.">
````

The description should be updated to be the package description such as

````html
<title>test 0.12.5 | Pub Package Manager</title>
<meta name="description"
  content="test provides a standard way of writing and running tests in Dart.">
````

Sharing (on slack and others) would be definitely better!

### PR DESCRIPTION
**Pull request by a14n**

### CODE REVIEW

**Comment by a14n** [comment]
WARNING: not tested!

**Comment by sethladd** [comment]
cc @whesse for review.

Thanks!

File: app/lib/templates.dart [hunk 1]
```diff
@@ -211,8 +211,9 @@ class TemplateService {
         'changelog_filename': changelogFilename,
         'version_count' : '$totalNumberOfVersions',
     };
-    return _renderPage(
-        'pkg/show', values, title: '${package.name} ${latestVersion.id} | Pub Package Manager');
+    return _renderPage('pkg/show', values,
+        title: '${package.name} ${latestVersion.id} | Pub Package Manager',
```
mkustermann: Instead of using `latestVersion.id` you could use `latestVersion.semanticVersion` here.

mkustermann:
Thanks Alexandre for the contribution :)

Sorry for the late reply. I've seen you've signed the CLA, so I'll merge it.

PR #51, opened 2016-06-03
Add link to Crossdart on the pkg/show page

astashov:
**Note**: I didn't test it, it's pretty hard to run it locally, it depends on Google Cloud account heavily.
But it type checks! :)

Crossdart contains the hyperlinked source code of all the packages on
pub. Somewhat like dartdocs.org, but in case you want to read the source
code of a package instead of its documentation :)

It gonna look like this:

image

/cc @filiph

### DIFF
--- a/app/lib/models.dart
+++ b/app/lib/models.dart
@@ -155,6 +155,13 @@ class PackageVersion extends db.ExpandoModel {
 
   String get documentationNice => niceUrl(documentation);
 
+  String get crossdart {
+    var name = Uri.encodeComponent(packageKey.id);
+    var version = Uri.encodeComponent(id);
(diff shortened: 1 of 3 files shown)


filiph:
We might want to call this "Source code (hyperlinked)" on the page, not "Crossdart".

astashov:
Makes sense, will change.

astashov:
Fixed

filiph:
Since this is not tested, I'd like someone familiar with the codebase to lgtm this.
// cc @kevmoo

mkustermann:
LGTM

mkustermann:
The change is live. The hyperlinked source is really nice :)

astashov:
Awesomeness! Thank you very much!

PR #76, opened 2017-03-15
compress pngs

kevmoo:
Also remove unused pngs


mkustermann:
Rietveld CLs are preferred!

PR #88, opened 2017-03-31
Add latest dev version to the search result page.

### LINKED ISSUE #47: Ability to see both stable and dev release versions
Related to issue #46, if you visit https://pub.dartlang.org/packages/polymer you now see only the stable version, not the release candidate version. These are very different, and have totally different readme files etc.

I think maybe the best solution would be to actually have two different pages? Maybe stable is still the default but it has a link to a page for the dev version?

isoos:
Fixes #47.

Disclaimer: first time touching gcloud database code. The proposed solution is suboptimal, it would be faster if we would store latest_dev_version value at the package object.

### DIFF
--- a/app/lib/search_service.dart
+++ b/app/lib/search_service.dart
@@ -81,14 +81,27 @@ class SearchService {
         var versionKeys =
             packages.where(exists).map((p) => p.latestVersionKey).toList();
         if (versionKeys.isNotEmpty) {
+          // select latest development versions for each package
+          var packageKeys = versionKeys.map((vk) => vk.parent).toList();
+          List<PackageVersion> devVersions = await Future.wait(
+              packageKeys.map((p) async {
+                List<PackageVersion> all = await db
+                    .query(PackageVersion, ancestorKey: p)
+                    .run()
+                    .toList();
+                all.sort((a, b) =>
+                    b.semanticVersion.compareTo(a.semanticVersion));
+                return all.first;
+              }));
           var versions = await db.lookup(versionKeys);
           int count = min(int.parse(search.searchInformation.totalResults),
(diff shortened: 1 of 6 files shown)


mkustermann:
It's not super pretty, but fine for now.

If you like you could add an optional `latestDevVersion` property to the `Package` model class, make a small batch program (see examples in `app/bin/tools*dart`) for converting the existing `Package` objects. Then you could simplify this code (as well as the one in `handlers.dart`, which is similar).

But I'll pull this change anyways now.

app/lib/search_service.dart
```diff
@@ -81,14 +81,27 @@ class SearchService {
         var versionKeys =
             packages.where(exists).map((p) => p.latestVersionKey).toList();
         if (versionKeys.isNotEmpty) {
+          // select latest development versions for each package
+          var packageKeys = versionKeys.map((vk) => vk.parent).toList();
+          List<PackageVersion> devVersions = await Future.wait(
```
mkustermann: Strictly speaking, this adds more latency then necessary, because it first does `await db.lookup(keys)`, waits for the result and then does `await Future.wait(...)`.

The second call will download all versions of all `[packages]` and contains therefore also the information from the first rpc.

isoos:
thnx for merging, I'll do a 1-2 follow-up PRs for the suggested new field

mkustermann:
Maybe an advice: Usually I have done changes like this in two stages
a) make a CL which makes the app forwards compatible (adding an optional `latestDevVersionKey` field to the `Package` model)
b) make code which uses this new field

Then one can deploy a), run the batch jobs to update the packages and deploy b) afterwards. If something goes wrong in b) one can always switch to the a) version (but we cannot switch to a version before a)).

PR #89, opened 2017-04-03
Simple validation of the package name in the uploaded pubspec.

### LINKED ISSUE #50: Validate package name in the backend
Currently the pub client validates package names (see https://github.com/dart-lang/pub/blob/master/lib/src/validator/name.dart#L51). However the backend does not, so if packages are uploaded by some other means than the pub client (or using a modified pub client) illegal package names will be accepted.

isoos:
Fixes #50.

### DIFF
--- a/app/lib/utils.dart
+++ b/app/lib/utils.dart
@@ -52,9 +52,9 @@ String canonicalizeVersion(String version) {
   return canonicalVersion.toString();
 }
 
-RegExp _identifierExpr = new RegExp(r'^[a-zA-Z0-9_]*$');
-RegExp _startsWithLetterOrUnderscore = new RegExp(r'^[a-zA-Z_]');
-const _reservedWords = const [
+final RegExp _identifierExpr = new RegExp(r'^[a-zA-Z0-9_]+$');
+final RegExp _startsWithLetterOrUnderscore = new RegExp(r'^[a-zA-Z_]');
+const List<String> _reservedWords = const <String>[
   'assert',
   'break',
   'case',
--- a/app/lib/utils.dart
+++ b/app/lib/utils.dart
@@ -51,3 +51,56 @@ String canonicalizeVersion(String version) {
   }
   return canonicalVersion.toString();
 }
+
+RegExp _identifierExpr = new RegExp(r'^[a-zA-Z0-9_]*$');
+RegExp _startsWithLetterOrUnderscore = new RegExp(r'^[a-zA-Z_]');
+const _reservedWords = const [
+  'assert',
+  'break',
+  'case',
+  'catch',
+  'class',
+  'const',
+  'continue',
+  'default',
+  'do',
+  'else',
+  'extends',
+  'false',
+  'final',
+  'finally',
+  'for',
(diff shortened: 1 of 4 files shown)


app/lib/utils.dart
```diff
@@ -51,3 +51,56 @@ String canonicalizeVersion(String version) {
   }
   return canonicalVersion.toString();
 }
+
+RegExp _identifierExpr = new RegExp(r'^[a-zA-Z0-9_]*$');
+RegExp _startsWithLetterOrUnderscore = new RegExp(r'^[a-zA-Z_]');
```
mkustermann: Consider always using `final` if you don't modify variables (also local variables)

app/lib/utils.dart
```diff
@@ -51,3 +51,56 @@ String canonicalizeVersion(String version) {
...
+const _reservedWords = const [
```
mkustermann: Consider also typing it `const <String> [ ... ]`. Now that we are moving towards the strong mode type system, it's probably better.

app/test/backend_test.dart
```diff
@@ -731,6 +731,37 @@ main() {
...
 
+        scopedTest('bad package names are rejected', () async {
+          var tarballStorage = new TarballStorageMock();
+          var transactionMock = new TransactionMock();
+          var db = new DatastoreDBMock(transactionMock: transactionMock);
+          var repo = new GCloudPackageRepository(db, tarballStorage);
+          registerLoggedInUser('hans@juergen.com');
+
+          // Returns the error message as String or null if it succeeded.
+          Function fn = (String name) async {
```
mkustermann: The
```dart
Function fn = (String name) async {
```
can be simplified to 
```dart
fn(String name) async {
```

app/lib/utils.dart
```diff
@@ -51,3 +51,56 @@ String canonicalizeVersion(String version) {
   }
   return canonicalVersion.toString();
 }
+
+RegExp _identifierExpr = new RegExp(r'^[a-zA-Z0-9_]*$');
```
mkustermann: Replace `*` by `+`, we already know it's not the empty string (which is handled before).

mkustermann:
Looks fine with me. Added a few comments. 

Would be **really nice** to share the entire logic of `pubspec.yaml` validation with the client. (or have at least a library we can copy entirely)

isoos:
Done w/ the requested changes. Re: `pubspec.yaml` validation: we could either depend on the `pub` package, or create a new one (under `/dart-lang/`, I think a Googler should own it). Either way, I can do the PRs to share the same codebase.

If you don't mind, I'll open a new PR with more rigorous dart analyzer settings (e.g. finals enforced).

app/test/backend_test.dart
```diff
@@ -731,6 +731,37 @@ main() {
           });
         });
 
+        scopedTest('bad package names are rejected', () async {
+          var tarballStorage = new TarballStorageMock();
```
mkustermann: nit: These variables could also be final. Just for future CLs.

PR #90, opened 2017-04-04
Enable strong mode, a few linter rules and run dartfmt on most files

isoos:
Apologies for the long PR, the changes are mostly non-functional, with the exception of calling `.toString()` in `app/bin/tools_remove_package.dart` line 95, and using `dart:math`'s `min` in `app/lib/search_service.dart`.

They started out as separate commits, but got squashed along the way, and it would have been too painful to get separated again.

### DIFF
--- a/app/lib/atom_feed.dart
+++ b/app/lib/atom_feed.dart
@@ -23,17 +23,17 @@ class FeedEntry {
   final String alternateTitle;
 
   FeedEntry(this.id, this.title, this.updated, this.authors, this.content,
-            this.alternateUrl, this.alternateTitle);
+      this.alternateUrl, this.alternateTitle);
 
   void writeToXmlBuffer(StringBuffer buffer) {
-    var escape = _HtmlEscaper.convert;
+    final escape = _HtmlEscaper.convert;
 
     var authorTags = '';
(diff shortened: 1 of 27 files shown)


app/lib/atom_feed.dart
```diff
@@ -23,25 +23,25 @@ class FeedEntry {
   final String alternateTitle;
 
   FeedEntry(this.id, this.title, this.updated, this.authors, this.content,
-            this.alternateUrl, this.alternateTitle);
+      this.alternateUrl, this.alternateTitle);
```
mkustermann: There is another place further up, where you replaced a `Configuration.prod() ...;` with `Configuration.prod() ... {}`. But here you don't add an empty clause.

I guess it means if there are field initializes **only then** do we also need the empty block?

isoos: I believe I have only removed `{}` (and replaced them with `;`).

mkustermann: Oh, sorry. Got probably irritated by this github diff showing tool. We use normally rietveld which is much nicer for reviewing.

mkustermann:
You could update `README.md` how to run the analyzer on the package if you like.

PR #93, opened 2017-04-05
Adding the latestDevVersionKey field.

isoos:


app/lib/backend.dart
```diff
@@ -289,11 +289,7 @@ class GCloudPackageRepository extends PackageRepository {
       package.updated = newVersion.created;
 
       // Keep the latest version in the package object up-to-date.
-      if (package.latestSemanticVersion < newVersion.semanticVersion &&
-          (package.latestSemanticVersion.isPreRelease ||
-              !newVersion.semanticVersion.isPreRelease)) {
-        package.latestVersionKey = newVersion.key;
-      }
+      package.updateVersion(newVersion.key);
```
mkustermann: Since you have the `PackageVersion` available on new uploads, I think it would be better to pass it instead of just a `Key` object (see comment below)

app/lib/models.dart
```diff
@@ -63,6 +66,23 @@ class Package extends db.ExpandoModel {
         .where((email) => email != lowerEmail)
         .toList();
   }
+
+  void updateVersion(db.Key newVersionKey) {
+    final Version newVersion = new Version.parse(newVersionKey.id);
```
mkustermann: It is normally not encouraged to directly use `.id` because that is making assumptions. Instead if you passed in a `PackageVersion pv` object you could just use `pv.latestVersion` and `pv.latestSemanticVersion` (see the getters above). I would also add two new getters for the dev version.

app/lib/models.dart
```diff
@@ -188,3 +208,55 @@ String niceUrl(String url) {
...
+}
+
+/// Compares two versions according to the semantic versioning specification.
+///
+/// If [pubSorting] is `true` then pub's priorization ordering is used, which
+/// will rank pre-release versions lower than stable versions (e.g. it will
+/// order "0.9.0-dev.1 < 0.8.0").  Otherwise it will use semantic version
+/// sorting (e.g. it will order "0.8.0 < 0.9.0-dev.1").
+int _compareSemanticVersionsDesc(Version a, Version b,
+    {bool decreasing: true, bool pubSorting: true}) {
```
mkustermann: This is a private method and only called from two places, I would make the named parameters, required positional ones (both call sites know the information).

app/bin/tools_backfill_packages.dart
```diff
@@ -0,0 +1,50 @@
...
+import 'server_common.dart';
+import 'tools_common.dart';
+
+Future main(List<String> args) async {
+  int updatedPackages = 0;
+  await withProdServices(() async {
+    await fork(() async {
+      if (Platform.isMacOS) {
+        await initializeApiaryDatastore();
+      }
```
mkustermann: I think it would be more readable to encapsulate the `Platform.isMacOS` specific code (as already mentioned somewhere).

Something like this, should do it:

```dart
Future withProdServices(Future fun()) {
  if (Platform.isMacOS) {
    return _withProdServices(() {
      return fork(() async {
        await initializeApiaryDatastore();
        return fun();
      });
    });
  } else {
    return _withProdServices(fun);
  }
}

Future _withProdServices(Future fun()) async {
  if (!Platform.environment.containsKey('GCLOUD_PROJECT') ||
      !Platform.environment.containsKey('GCLOUD_KEY')) {
    throw 'Missing GCLOUD_* environments for package:appengine';
  }
  return withAppEngineServices(() {
    registerUploadSigner(
        new ServiceAccountBasedUploadSigner(activeConfiguration.credentials));
    initBackend();
    return fun();
  });
```

Then everyone using `withProdServices()` doesn't need to care about it anymore

isoos:
Updated most request, but I haven't merged the server.dart's `Platform.isMacOS` yet, I'd defer it to a subsequent change.

PR #102, opened 2017-04-10
Adding some templates test.

isoos:


mkustermann:
Here's a suggestion: Instead of just rendering a template and grepping for random pieces in it, we:
* have a tool to write out golden files
* have a unit test which compares against the golden files

Then we would also see immediately what changes in the output if we change the templates.

isoos:
Changed the test to golden files. I haven't added a separate tool yet, for these simple tests, I've used the following snippet in each test case to write it: `new File('path.html').writeAsStringSync(html)`

PR #106, opened 2017-04-18
Use the new latest_dev_version field in search.

isoos:


app/lib/search_service.dart
```diff
@@ -78,19 +78,14 @@ class SearchService {
...
-          final packageKeys = versionKeys.map((vk) => vk.parent).toList();
-          final List<PackageVersion> devVersions =
-              await Future.wait(packageKeys.map((p) async {
-            final List<PackageVersion> all =
-                await db.query(PackageVersion, ancestorKey: p).run().toList();
-            all.sort((a, b) => b.semanticVersion.compareTo(a.semanticVersion));
-            return all.first;
-          }));
           final versions = await db.lookup(versionKeys);
+          final devVersions = await db.lookup(devVersionKeys);
```
mkustermann: This is serializing 2 RPCs, you can reduce the latency by doing only one RPC or running the two in parallel.

```dart
await db.lookup([]..addAll(versionKeys)..addAll(devVersionKeys))
```

or

```dart
await Future.wait([db.lookkup(versionKeys), db.lookup(devVersionKeys)]);
```

isoos: Fixed. I went with the first version, because most dev versions will be the same as the stable version, and I hope that the DataStore will figure that out and won't do duplicate reads.

PR #109, opened 2017-04-21
Introduce new field: detectedTypes.

isoos:


mkustermann:
Added a few comments, looks fine to me. Though the name `scopes` is not that descriptive -- I'd prefer `searchTerms` or something similar.

app/bin/tools_backfill_scopes.dart
```diff
@@ -0,0 +1,54 @@
...
+  final int lengthA = a?.length ?? 0;
+  final int lengthB = b?.length ?? 0;
+  if (lengthA != lengthB) return false;
+  if (lengthA == 0) return true;
+  return (a.toSet()..removeAll(b)).isEmpty;
+}
+
+Future _updatePackageVersion(Key pvKey) async {
+  await dbService.withTransaction((Transaction t) async {
+    final list = (await t.lookup([pvKey.parent, pvKey]));
```
isoos: I think simplicity of the code is more beneficial here than the one-off gain in performance. If/when we introduce more search features, we may need to modify this backfill job and make it more complex. I'd defer this optimization until then.

In addition to that: we could choose an indexing/search mechanism that doesn't require the `scope` field on the Package object, and we could remove the lookup here.

app/lib/models.dart
```diff
@@ -86,6 +93,19 @@ class Package extends db.ExpandoModel {
       latestDevVersionKey = pv.key;
     }
   }
+
+  /// Updates the search-related fields (eg. [scope]) to match the latest stable
+  /// [PackageVersion].
+  ///
+  /// Returns whether any field has been updated.
+  bool updateSearchFields(PackageVersion pv) {
```
isoos: Discussed parts of this offline. Let's wait until the search/indexing solution is decided, because that would make the question of naming irrelevant here.

mkustermann:
@isoos 
Is this CL still waiting for something or should it be merged}


About the original TBDs:

> the field's name (scope)

Would prefer a more descriptive name (tags, searchTerms, ...) - though since I don't have a strong opinion, you can stick with scope.

> the values (e.g. flutter vs flutter_plugin)

It's just a marker name, right? So we can use whatever we want. I'd go for `flutter_plugin` if it is a plugin.

> whether the Package instance should also have this field (e.g. whether we will need it for querying the Datastore)

It would be preferred to not have duplication if it is not necessary: If all places where we query the package (e.g. for rendering the package page template) we also query the latest version, then it should be enough to have it on the latest version I think. It's always possible to add it later, if need be.

isoos:
This is still pending, I'm waiting on @kevmoo's feedback on the search choice.

e.g. if we go ahead with using the current custom search API with additional attributes, than "listing the flutter plugins" page will need to have the duplicated values on the `Package` object, otherwise we may not need it there.

isoos:
Updated the value (`flutter` -> `flutter_plugin`).

isoos:
@mkustermann please do another round of review, this version should be ready to merge.

I've renamed `scope` to `detectedTypes`, and tried to be keep the other parts consistent with it.(I'm open for naming suggestions thought.)

Talked with Kevin, and we concluded that we start with the custom search API, and for the "list of flutter plugins" page we'll need that field on the `Package` object too.

kevmoo:
@mkustermann cool w/ this change?

app/lib/backend.dart
```diff
@@ -640,11 +644,22 @@ Future<models.PackageVersion> parseAndValidateUpload(
     ..changelogContent = changelogContent
     ..libraries = libraries
     ..downloads = 0
```
mkustermann: unrelated comment: at some point we should do some cleanup and get rid of the 'downloads' field.

mkustermann:
Looks fine with me.

Issue #108, opened 2017-04-20
Change the title: "Dart Packages"


kevmoo:
"Pub Package Manager" seems a weird title. We should have Dart right there – this is **the** site for Dart Packages.

Right?

kevmoo:
...I'd like to discuss this a bit before running with it. Just brainstorming.

isoos:
Related observations:
- Searching for "dart packages" gets https://pub.dartlang.org/packages as the first result, right before the domain root.
- Searching for "dart libraries" have many related results, but none of them is pub.dartlang.org on the first page (not sure if this is a concern, but I expected at least one first-page result).
- Search for "angular dart package" gets https://pub.dartlang.org/packages/angular as the first result, and presents angular2, Angular2_components, angular_ui... as they were sub-sections of it.
- The prior search has the page title: "angular 1.1.2+2 | Pub Package Manager - Dart Pub" - The "Pub" name is duplicated, and I'm not sure if it is beneficial to include the version there, e.g. "angular library | Pub - Dart Package Manager" could be better.

mkustermann:
Should this issue closed, since https://github.com/isoos/pub-dartlang-dart/commit/1adda949cf75c0652d51e595b7b8468a881f41d7 has landed?

PR #110, opened 2017-04-24
Changing the page title in reference to #108

isoos:


mkustermann:
There might have been a template test which fails now (the golden files). Maybe you can update them in another CL?

isoos:
Ah, I did forget to run tests, and also failed to turn it on travis for my branch. Fixing it.

PR #112, opened 2017-04-24
consolidating the Platform.isMacOS calls

isoos:
@mkustermann: this is a slightly modified version of your patch (https://codereview.chromium.org/2840533002/), thanks for the suggestions. The modification: I'm storing the apiary database in a local field, that way we don't need to pass it around.

### DIFF
--- a/app/bin/server_common.dart
+++ b/app/bin/server_common.dart
@@ -89,7 +89,36 @@ Future<String> obtainServiceAccountEmail() async {
   return response.body.trim();
 }
 
-Future<DatastoreDB> initializeApiaryDatastore() async {
+/// Wrapper function used to work around issue with running gRPC Datastore on
+/// MacOS.
+///
+/// On
+///    * Linux this function simply returns `fun()`
+///
+///    * MacOS this function makes a new service scope, optionally builds an
+///      apiary Datastore and registers an apiary Datastore in the new service
+///      scope
+///
+Future withCorrectDatastore(Future fun()) {
+  if (Platform.isMacOS) {
+    return fork(() async {
+      DatastoreDB apiaryDatastore = lookup(_apiaryDatastoreSymbol);
+      if (apiaryDatastore == null) {
+        apiaryDatastore = await _initializeApiaryDatastore();
+        register(_apiaryDatastoreSymbol, apiaryDatastore);
+      }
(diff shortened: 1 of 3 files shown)


app/bin/server_common.dart
```diff
@@ -89,7 +89,32 @@ Future<String> obtainServiceAccountEmail() async {
   return response.body.trim();
 }
 
-Future<DatastoreDB> initializeApiaryDatastore() async {
+/// Wrapper function used to work around issue with running gRPC Datastore on
+/// MacOS.
+///
+/// On
+///    * Linux this function simply returns `fun(dbService)`
```
mkustermann: simply returns `fun()`

app/bin/server_common.dart
```diff
@@ -89,7 +89,32 @@ Future<String> obtainServiceAccountEmail() async {
...
+///
+///    * MacOS this function makes a new service scope, optionally builds an
+///      apiary Datastore and registers an apiary Datastore in the new service
+///      scope
+///
+Future withCorrectDatastore(Future fun()) {
+  if (Platform.isMacOS) {
+    return fork(() async {
+      _apiaryDatastore ??= await _initializeApiaryDatastore();
```
mkustermann: Consider resetting `_apiaryDatastore` after `await fun()`. So it's nesting, but the outermost one, resets it to `null`.

app/bin/tools_common.dart
```diff
@@ -6,34 +6,22 @@ import 'dart:io';
...
-      return fork(() async {
-        await initializeApiaryDatastore();
-        return fun();
-      });
+  return withCorrectDatastore(() {
+    if (!Platform.environment.containsKey('GCLOUD_PROJECT') ||
+        !Platform.environment.containsKey('GCLOUD_KEY')) {
+      throw 'Missing GCLOUD_* environments for package:appengine';
+    }
+    return withAppEngineServices(() {
```
mkustermann: The `withCorrectDatastore` and `withAppEngineServices` need to be switched (maybe I got that wrong in my example).

app/bin/server_common.dart
isoos: I think resetting it to null may cause unintended race conditions, I've switched it to store it session scope, that way the nested will use the instance created in the outer one.

PR #117, opened 2017-05-02
Handler for /flutter/plugins.

isoos:
Additional handlers:
- `/flutter` redirects to `/flutter/plugins`
- `/flutter/plugins` uses the same pipeline as `/packages`, except it will add an additional filter for `detectedTypes = flutter_plugin`.

Prior releasing please update datastore indexes via `gcloud datastore create-indexes index.yaml`.


app/lib/handlers.dart
```diff
@@ -166,19 +168,22 @@ Future<shelf.Response> staticsHandler(shelf.Request request) async {
 
 /// Handles requests for /packages - JSON
```
mkustermann: Since you pass in "basePath" now, please update the comment what URLs are coming here.

app/lib/handlers.dart
```diff
@@ -201,17 +206,19 @@ Future<shelf.Response> packagesHandlerJson(
 }
 
 /// Handles requests for /packages - HTML
```
mkustermann: ditto: Please update comment.

app/views/pkg/index.mustache
```diff
@@ -2,7 +2,7 @@
     for details. All rights reserved. Use of this source code is governed by a
     BSD-style license that can be found in the LICENSE file. }}
 
-<h1>Packages</h1>
+<h1>{{title}}</h1>
```
mkustermann: Would you mind adding a test for `/flutter/...` and update the golden files?

app/lib/handlers.dart
```diff
@@ -450,6 +457,28 @@ Future<shelf.Response> apiPackagesHandler(shelf.Request request) async {
...
+  if (path == '/flutter' || path == '/flutter/') {
+    return _redirectResponse(basePath);
+  }
+  final int page = _pageFromUrl(request.url);
+  if (path.endsWith('.json')) {
+    return packagesHandlerJson(request, page, true,
+        basePath: basePath, detectedType: BuiltinTypes.flutterPlugin);
+  } else if (request.url.queryParameters['format'] == 'json') {
+    return packagesHandlerJson(request, page, false,
+        basePath: basePath, detectedType: BuiltinTypes.flutterPlugin);
```
mkustermann: Do we need to expose these two json api endpoints? Who is going to use them?

(These `.json` and `?format=json` apis are not the real apis used by the pub client. )

isoos: AFAIK we don't have any users, I've just kept the same pattern we had for /packages. Happy to remove it, especially because the URL spec may change in the near future, and the simpler the better for now.

app/views/pkg/index.mustache
isoos: Done. (Squashed commits, because of removal of the json methods.)

app/test/handlers_test.dart
```diff
@@ -120,6 +124,43 @@ void main() {
...
+          expect(packages.first, testPackage);
+          return [testPackageVersion];
+        });
+        registerBackend(backend);
+        expectHtmlResponse(await issueGet('/flutter/plugins'));
+      });
+
+      tScopedTest('/flutter/plugins&page=2', () async {
+        final backend =
+        new BackendMock(latestPackagesFun: ({offset, limit, detectedType}) {
```
mkustermann: Is this dartfmt formatted code? Looks kind of weired :-/

app/lib/handlers.dart
```diff
@@ -448,6 +453,20 @@ Future<shelf.Response> apiPackagesHandler(shelf.Request request) async {
   return _jsonResponse(json);
 }
 
+/// Handles requests for /flutter
+Future<shelf.Response> flutterHandler(shelf.Request request) async {
+  final String path = request.requestedUri.path;
+  final String basePath = '/flutter/plugins';
+  if (path == '/flutter' || path == '/flutter/') {
```
mkustermann: If this function gets called for `/flutter/foo/bar?page=2` we will treat it as `/flutter/plugins?page=2`. We should rather return a 404 in such a case.

The functionality of this function should be, if I see this correctly, only handling URLs of the form:
```
/flutter/plugins
/flutter/plugins?page=<number>
```

See comment above.

app/lib/handlers.dart
```diff
@@ -62,6 +62,8 @@ Future<shelf.Response> appHandler(
     return docHandler(request);
   } else if (path.startsWith('/static')) {
     return staticsHandler(request);
+  } else if (path.startsWith('/flutter')) {
+    return flutterHandler(request);
```
mkustermann: I think you should move the handling of this to be precisely like we do with '/packages', namely add it to the `handlers` map above, then it gets called only if the url path is correct (see comment below why)

You can add an extra redirection handler here for `/flutter` -> `/flutter/plugins`

mit-mit:
@mkustermann can we go ahead and merge this?

PR #120, opened 2017-05-03
Fixing PageLink.rightmostPage, added tests. 

isoos:
I'm planning to do a follow-up change which removes the `<a>` element from the disabled entries.

### DIFF
--- a/app/test/golden/pagination_first.html
+++ b/app/test/golden/pagination_first.html
@@ -1,0 +1,45 @@
+
+<div class="pagination">
+  <ul>
+      <li class="disabled">
+        <a href="/packages?page=0">
+         &laquo
+        </a>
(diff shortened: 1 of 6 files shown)


app/test/golden/pagination_first.html
```diff
@@ -0,0 +1,45 @@
+
+<div class="pagination">
+  <ul>
+      <li class="disabled">
+        <a href="/packages?page=0">
+         &laquo
```
kevmoo: this seems weird. Shouldn't it end w/ a `;`?

isoos: It should, but for some reason it is working, nevertheless. Fixing it in the subsequent change, where I'm modifying the related template.

PR #131, opened 2017-05-08
Use favicons. Fixes #129 for now.

### LINKED ISSUE #129: Feature request: Use favicons for pub.dartlang.org pages
Could be used to distinguish between the different kind of packages for the package specific pages and the filtered views.

isoos:
I believe the icon-selection logic will be more complex over time, but for now this is good enough.

### DIFF
--- a/app/lib/templates.dart
+++ b/app/lib/templates.dart
@@ -88,7 +88,7 @@ class TemplateService {
   /// Renders the `views/pkg/index.mustache` template.
   String renderPkgIndexPage(
       List<Package> packages, List<PackageVersion> versions, PageLinks links,
-      {String title}) {
+      {String title, String faviconUrl}) {
     final packagesJson = [];
     for (int i = 0; i < packages.length; i++) {
       final package = packages[i];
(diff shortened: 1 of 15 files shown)


app/lib/templates.dart
```diff
@@ -528,3 +538,8 @@ class PackageLinks extends PageLinks {
   @override
   String formatHref(int page) => '${_basePath ?? '/packages'}?page=$page';
 }
+
+abstract class LogoUrls {
+  static const String smallDartFavicon = '/static/favicon.ico';
+  static const String flutterLogo32x32 = '/static/img/flutter-logo-32x32.png';
```
mkustermann: It is a bit confusing that we:
* have the two images in different extensions `.png` and `.ico`
* have the two images possibly in different resolutions

but do use them at the same places (see e.g. the `<link rel="short cut icon"` ...)

isoos: Agreed. I believe `favicon.ico` is the only one that will be `.ico`, the rest will be `.png` going forward. I'm not even sure that we need to stick with `.ico` anymore, need to do some research before ditching it entirely.

PR #132, opened 2017-05-09
Refactor search query and result page

isoos:
... in preparation for additional filtering logic. The idea there is that list of filtered types goes to the `SearchQuery` object as well, and we need to use that info in `PageLinks`, and also in some formatting of the search result page too.

### DIFF
--- a/app/lib/search_service.dart
+++ b/app/lib/search_service.dart
@@ -56,15 +56,14 @@ class SearchService {
 
   SearchService(this.httpClient, this.csearch);
 
-  /// Search for packes using [query], starting at offset [offset] returning
+  /// Search for packes using [queryText], starting at offset [offset] returning
   /// max [numResults].
-  Future<SearchResultPage> search(
-      String query, int offset, int numResults) async {
+  Future<SearchResultPage> search(SearchQuery query) async {
     bool exists(x) => x != null;
     final db = dbService;
 
-    final search = await csearch.cse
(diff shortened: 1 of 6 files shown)


app/lib/search_service.dart
```diff
@@ -91,22 +90,31 @@ class SearchService {
...
-  final String query;
+  String text;
+
+  /// The offset used for the search.
+  int offset;
 
-  /// The offset used for the search/
-  final int offset;
+  /// The maximum number of items queried when search.
+  int limit;
```
mkustermann: Could these fields be final?

app/lib/templates.dart
```diff
@@ -361,14 +362,15 @@ class TemplateService {
...
     final values = {
-      'query': query,
+      'query': queryText,
       'results': results,
       'pagination': renderPagination(pageLinks),
       'hasResults': results.length > 0,
     };
-    return _renderPage('search', values, title: 'Search results for $query.');
+    return _renderPage('search', values,
+        title: 'Search results for $queryText.');
```
mkustermann: Do we need to escape `queryText` here?

isoos: It will get html-escaped in `renderLayoutPage`. Or do you want to put that inside quotes?

mkustermann: Couldn't remember whether _renderPage will do escaping by default or not. It's fine!

PR #140, opened 2017-05-15
Strict HTML check for templates in test.

isoos:


isoos:
This would have prevented the `<a><a>` issue that was left in a template last week.

PR #141, opened 2017-05-15
Move environment variable access to a single place.

isoos:


app/bin/configuration.dart
```diff
@@ -50,3 +50,25 @@ class Configuration {
       : projectId = 'dartlang-pub-dev',
         packageBucketName = 'dartlang-pub-dev--pub-packages';
 }
+
+/// Configuration from the environment variables.
+class EnvConfig {
+  final String gcloudKey;
+  final String gcloudProject;
+
+  EnvConfig._({this.gcloudKey, this.gcloudProject});
```
mkustermann: Why are these named optional parameters?

app/bin/configuration.dart
```diff
@@ -50,3 +50,25 @@ class Configuration {
...
+
+  factory EnvConfig._detect() {
+    return new EnvConfig._(
+      gcloudKey: Platform.environment['GCLOUD_KEY'],
+      gcloudProject: Platform.environment['GCLOUD_PROJECT'],
+    );
+  }
+
+  bool get hasGcloudKey => gcloudKey != null;
```
mkustermann: This getter is unnecessary, since we always require `GCLOUD_PROJECT` to be there.

app/bin/configuration.dart
```diff
@@ -50,3 +50,25 @@ class Configuration {
...
+/// Configuration from the environment variables.
+class EnvConfig {
+  final String gcloudKey;
+  final String gcloudProject;
+
+  EnvConfig._({this.gcloudKey, this.gcloudProject});
+
+  factory EnvConfig._detect() {
+    return new EnvConfig._(
+      gcloudKey: Platform.environment['GCLOUD_KEY'],
```
mkustermann: I would throw an exception if the `GCLOUD_PROJECT` is not set, because that is a requirement (for local and prod), the key is optional iirc.

app/bin/configuration.dart
isoos: Kept it, because tools_common.dart still references it, after we make `GCLOUD_PROJECT` mandatory.

PR #151, opened 2017-05-24
Update hash on tab changes. Better tab IDs. Fixes #150.

### LINKED ISSUE #150: On clicking on the installing tab, add the #-pkg-tab-installing to the url.
Original at: https://github.com/dart-lang/pub/issues/1621

cc @branflake2267

On clicking on the installing tab, it would be nice to add the hash and fragment to the url. This would allow for easy copy and paste. 

This would make much easier to share the installing link in other guides. 

For example: 
https://pub.dartlang.org/packages/google_sign_in#-pkg-tab-installing

screen shot 2017-05-23 at 3 56 18 pm

isoos:


static/js/script.js
```diff
@@ -12,6 +12,20 @@ $(function() {
     $('a[href="' + window.location.hash + '"][data-toggle="tab"]').tab('show');
   }
 
+  // Update location hash after a tab is selected.
```
mkustermann: Strictly speaking this JavaScript code doesn't need to run on every page, just on the one where we have the package view. But it's fine with me.

PR #153, opened 2017-05-24
Preserve search query. Fixes #152.

### LINKED ISSUE #152: Search field text should be persisted when searching
Currently, if you fill in a search string in the search field -- say foo'' and then hit Enter, you get search results for 'foo', however 'foo' is not preserved in the search field, so if you want to refine your search you have to enter the whole search term again.

Would be good to preserve the search string, just like on google.com and other search pages.

isoos:


kevmoo:
@mkustermann would you take a look and merge?

PR #155, opened 2017-05-30
async analyzer skeleton with PubSub

isoos:
@mkustermann: This part of gcloud is new to me, any advice on how to initialize the PubSub object? It needs an authorizing http Client, and I'm unsure from where to get it.

### DIFF
--- a/app/bin/analyzer_service.dart
+++ b/app/bin/analyzer_service.dart
@@ -1,0 +1,64 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:appengine/appengine.dart';
+
+import 'package:pub_dartlang_org/analyzer_topic.dart';
+
+import 'server_common.dart';
+
+void main() {
+  useLoggingPackageAdaptor();
+
+  withAppEngineServices(() async {
+    return withCorrectDatastore(() async {
+      // Trigger idle check every minute.
+      new Timer.periodic(new Duration(minutes: 1), _onIdle);
+
+      // TODO: call await setupServices(activeConfiguration);
+      // TODO: make sure AnalyzeTopic is initialized in setupServices
+
+      for (;;) {
+        _lastNonIdle = new DateTime.now();
+        final AnalyzeRequest rq = await analyzerTopic.pullRequest();
+        logger.info('Got AnalyzeRequest: ${rq.package} ${rq.version}');
+        _processing = true;
+        try {
+          final Stopwatch sw = new Stopwatch()..start();
+          await _analyze(rq.package, rq.version);
+          sw.stop;
+          logger.info('Analysis of ${rq.package} ${rq.version} completed '
+              'in ${sw.elapsed.inSeconds} seconds.');
+        } catch (e, st) {
+          logger.severe(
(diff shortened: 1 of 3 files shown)


mkustermann:
@isoos @kevmoo 

Maybe we should first talk about this before the implementation starts?

The current architecture is super nice: One can simply run the entire server locally as a standalone process. Once we introduce pub/sub we might loose that ability (and/or have very weird behavior) - quite obviously this makes me a bit worried. For developing the app a **fast and simple local dev cycle** is critical.

Also it's not quite clear where this work is going, so it might make sense to talk about it first.

isoos:
Feasibility on keeping it simpler:

As an alternative to a separate service/process, we could do the analysis in a different isolate. I wouldn't put it in the same web-serving isolate due to the potentially larger response latency, however in a separate isolate it should work fine.

We could choose polling the Datastore instead of Pub/Sub, the only question there is how to prevent processing the same package in parallel on each of the instances. In its simplest form we could fetch analysis candidates, and pick one randomly. Not ideal, but should work for now.

However, this would introduce scaling issues:
- we cannot scale it separately from the web-serving process
- we will need to have a way to temporarily increase the number of instances, e.g. when we want to trigger a batch (re-)analysis.

All-in-all: yes, we can certainly keep it simpler, given the above constraints.

kevmoo:
I like the idea of using a separate service.

We can have a separate `indexer.yaml` or similar. We can ever put it in a directory parallel to `app` if needed.

I think it's obvious we want to keep the datastore code, etc in the same repo. But that doesn't mean we should bolt the analysis logic on top of the current user-facing app.

...right?

isoos:
I've explored how to create a separate service/module alongside the main `app`. According to the [gcloud custom runtime setup](https://cloud.google.com/appengine/docs/flexible/custom-runtimes/build), we can have an `analyzer.yaml` for the analyzer service, but it will use the same `Dockerfile`, and AFAICT there is no configuration or command line option that selects another one. Which means that we need to find another way for setting up the analyzer service instance, for example the following ideas came to me:

- Tweak the `Dockerfile`, to be smart about the environment: conditionally copy either the `app` or the soon-to-be `analyzer` application into the `/project/app/` directory.
- Tweak the [Dart docker files](https://github.com/dart-lang/dart_docker), to be aware of gcloud services.
- Modify our server.dart to detect the environment and start either the web serving service, or the analyzer service (when certain environment variables are set). In this case we won't have a clear separation of the services.
- Refactor the directories:
  - move everything we have now into a new root e.g. `app/` becomes `public-facing-frontend-server/app/` (directory name intentionally long to indicate I would like to have a shorter version). The current `Dockerfile` from the root is also moved.
  - introduce a new top-level directory `analyzer/` that has its own `analyzer.yaml`, `Dockerfile` and separate project.

In most of the cases the `analyzer` project would depend on `app` with `path: ../../public-facing-frontend-server/app/` links.

Any thoughts?

isoos:
Update: probably a better option:
- Refactor only the `*.yaml` and `Dockerfiles`, and provide scripts to deploy either of them. The `Dockerfile` cannot reference a directory outside of its own, and won't follow symlinks, therefore the scripts will be using either carbon-copy of the files or a hard link (probably the later).

kevmoo:
@isoos what's the plan here?

isoos:
we have offline discussion with @mkustermann

isoos:
Closing this in favor of https://github.com/dart-lang/pub-dartlang-dart/pull/164.

PR #156, opened 2017-06-01
Experimental search result ordering with CSE bias and mock scores.

isoos:


app/lib/search_service.dart
```diff
@@ -28,6 +28,14 @@ SearchService get searchService => ss.lookup(#_search);
 /// Register a new [SearchService] in the current service scope.
 void registerSearchService(SearchService s) => ss.register(#_search, s);
 
+abstract class CseTokens {
+  static const pageMapDocument = 'document';
+  static const detectedTypePrefix = 'dt_';
+  static const experimentalScore = 'exp_score';
+
+  static String detectedType(String type) => 'dt_$type';
```
mkustermann: Consider using detectedTypePrefix, since you added it, instead of repeating `dt_` ?

(Also maybe add `String` type to the constants)

app/lib/search_service.dart
```diff
@@ -98,6 +109,16 @@ class SearchService {
   }
 }
 
+// https://developers.google.com/custom-search/docs/structured_search#bias-by-attribute
+enum SearchBias { strong, weak }
+
+SearchBias parseExperimentalBias(String value) {
+  if (value == null) return null;
+  if (value == 'strong') return SearchBias.strong;
+  if (value == 'weak') return SearchBias.weak;
```
mkustermann: Consider adding `'hard'` here as well, which seems to be the default, i.e. what we are using atm.

mkustermann:
Looks good, a few minor comments.

Maybe we should increase the test timeout to make travis less flaky? I think `package:test` allows one to just put a anotation on the `_test.dart` file on the top (i.e. library metadata)  -- e.g. `@Timeout(const Duration(seconds: 50))`.

isoos:
test timeout in a separate PR, otherwise fixed.

PR #157, opened 2017-06-01
Add survey code

kevmoo:


app/lib/templates.dart
```diff
@@ -300,14 +302,13 @@ class TemplateService {
...
-    String faviconUrl,
-    String searchQuery,
-  }) {
+  String renderLayoutPage(String title, String contentString,
+      {PackageVersion packageVersion,
+      Map<String, String> pageMapAttributes,
+      String faviconUrl,
+      String searchQuery,
+      bool includeSurvey: true}) {
+    includeSurvey ??= true;
```
mkustermann: It makes little sense to have `bool includeSurvey: true` if you check again for `null` and set it to `true`.

Either the caller is responsible for never passing `includeSuvey: null` or you remove the `: true`.

app/lib/templates.dart
```diff
@@ -384,24 +386,21 @@ class TemplateService {
...
-    String searchQuery,
-  }) {
+  String _renderPage(String template, values,
+      {String title: 'pub.dartlang.org',
+      PackageVersion packageVersion,
+      Map<String, String> pageMapAttributes,
+      String faviconUrl,
+      String searchQuery,
+      bool includeSurvey: true}) {
+    includeSurvey ??= true;
```
mkustermann: ditto

app/views/layout.mustache
```diff
@@ -45,6 +45,9 @@
     </PageMap>
     -->
     {{/has_pagemap}}
+    {{#include_survey}}
+    <script async="" defer="" src="//www.google.com/insights/consumersurveys/async_survey?site=5wdvu4vc5lhop3xbatxof6lzcm"></script>
```
mkustermann: Why would we want to link to the **http** version of `www.google.com` if the pub server is running on `http`?

Even if the pub server is running on **non-https** the link should still be **https://www.google.com/...** I think.

So consider changing `//` with `https://`.

mkustermann:
Looks fine, two minor comments.

app/views/layout.mustache
mkustermann: Or is this not allowed for some reason?

kevmoo: This is copy-pasted the recommended usage. Given that we host on https, I'm going to leave as-is.

app/lib/templates.dart
kevmoo: ack

PR #163, opened 2017-06-07
Fix UTF8 parse exception when Readme or Changelog contains invalid UTF-8 characters.

isoos:


kevmoo:
Is there a bug on this?

isoos:
no bug entry, but try to extract the README.md from https://storage.googleapis.com/pub-packages/packages/puremvc-2.0.6.tar.gz, it will fail without this change.

app/lib/utils.dart
```diff
@@ -36,10 +37,14 @@ Future<List<String>> listTarball(String path) async {
 }
 
 Future<String> readTarballFile(String path, String name) async {
-  final result = await Process.run('tar', ['-O', '-xzf', path, name]);
+  final result = await Process.run(
+    'tar',
+    ['-O', '-xzf', path, name],
+    stdoutEncoding: null,
```
kevmoo: Couldn't you just pass in `const UtfCodec.allowMalformed: true` to the `stdoutEncoding` param?

isoos: fixed.

kevmoo: Now to be really annoying: should this be an optional argument? (allowing malformed)

I could see us being "tolerant" for the readme/changelog. It makes me more nervous for the pubspec.

Thoughts, @mkustermann ?

isoos: In `pubspec` the most probable cause for weird characters will be the author's name, the description or a comment block. I see very little reason to throw on these.

kevmoo:
K – just checkin...

PR #164, opened 2017-06-08
Analyzer service skeleton inside app/, using env variable.

isoos:
- `analyzer.yaml` sets the `PUB_SERVICE=analyzer` env variable
- `bin/server.dart` checks the environment variable and dispatches application to either `frontend.dart` or to `analyzer.dart`.
- refactored `lib/`

git didn't mark `server.dart` as moved, rather updated, if you would like to, I could split up this PR and do it in two, so that history is better preserved.


analyzer.yaml
```diff
@@ -0,0 +1,24 @@
+# Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+# for details. All rights reserved. Use of this source code is governed by a
+# BSD-style license that can be found in the LICENSE file.
+
+runtime: custom
+env: flex
+service: analyzer
+env_variables:
```
mkustermann: Instead of relying on such a special variable, it might be better to use the `GAE_SERVICE` environment variable.

app/bin/service/analyzer.dart
```diff
@@ -0,0 +1,22 @@
...
+
+void main() {
+  useLoggingPackageAdaptor();
+
+  withAppEngineServices(() async {
+    await runAppEngine((HttpRequest ioRequest) async {
+      // TODO: implement API handlers
+      final HttpResponse response = ioRequest.response;
+      response.statusCode = 200;
+      response.add([]);
```
mkustermann: Adding 0 bytes to the response seems to me as a NOP, maybe just get rid of this?

app/bin/server.dart
```diff
@@ -1,121 +1,16 @@
...
-
-shelf.Request sanitizeRequestedUri(shelf.Request request) {
-  final uri = request.requestedUri;
-  final resource = uri.path;
-  final normalizedResource = path.normalize(resource);
-
-  if (resource == normalizedResource) {
-    return request;
+  if (Platform.environment['PUB_SERVICE'] == 'analyzer') {
+    analyzer.main();
```
mkustermann: Please use the 'GAE_SERVICE' environment variable instead, then we can also have 3 cases: the analyzer service, the default service and an error if this script is invoked with any other service.

mkustermann:
Maybe also update the `README.md` file with the instructions on how to start the two services now.

analyzer.yaml
isoos: That is a well-hidden env variable. I was looking for similar, but was without luck, and I have found little reference even after I knew what to search for... fix in progress

isoos:
added notes in README.

analyzer.yaml
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+# for details. All rights reserved. Use of this source code is governed by a
+# BSD-style license that can be found in the LICENSE file.
+
+runtime: custom
+env: flex
+service: analyzer
+
+resources:
+  cpu: 4
```
kevmoo: These are kinda expensive to be running all the time.

Since we don't care about instant through-put, I'm wondering if using one CPU and setting the max instances higher would be a bitter idea.

@mkustermann ?

analyzer.yaml
```diff
@@ -0,0 +1,22 @@
...
+  memory_gb: 8
+
+#manual_scaling:
+#  instances: 1
+
+automatic_scaling:
+  min_num_instances: 1
+  max_num_instances: 4
```
kevmoo: max instances = 1

isoos: at this point I switched back to the simpler setting:
```
manual_scaling:
  instances: 1
```

kevmoo: SGTM

isoos:
/ping @mkustermann: it would be great if you could merge this in the next half day or so, it will unblock a few dev path that could go in parallel

PR #165, opened 2017-06-12
Flutter SDK setup in Dockerfile

isoos:


Dockerfile
```diff
@@ -4,9 +4,17 @@ FROM google/dart-runtime-base:1.23.0
 # docker image diff small.
 RUN apt-get update && \
     apt-get upgrade -y && \
-    apt-get install -y memcached && \
+    apt-get install -y git memcached unzip && \
     rm -rf /var/lib/apt/lists/*
 
+ENV PATH="/flutter/bin:${PATH}"
+# To ensure actions from this container are not seen as "typical" users.
+ENV PUB_ENVIRONMENT="bot.pub_dartlang_org.docker"
```
isoos: /cc @kevmoo

kevmoo: 👍

Dockerfile
```diff
@@ -4,9 +4,17 @@ FROM google/dart-runtime-base:1.23.0
...
+
+RUN git clone -b alpha https://github.com/flutter/flutter.git
```
mkustermann: Using a branch might make this a non-reproducible docker build. Maybe use the tag '0.0.7' instead (to which the 'alpha' branch is currently set)

Dockerfile
```diff
@@ -4,9 +4,17 @@ FROM google/dart-runtime-base:1.23.0
 # docker image diff small.
 RUN apt-get update && \
     apt-get upgrade -y && \
-    apt-get install -y memcached && \
+    apt-get install -y git memcached unzip && \
     rm -rf /var/lib/apt/lists/*
 
+ENV PATH="/flutter/bin:${PATH}"
+# To ensure actions from this container are not seen as "typical" users.
```
mkustermann: Maybe rephrase this a bit:

`Let the pub server know that this is not a "typical" pub client but rather a bot.`

Dockerfile
```diff
@@ -4,9 +4,17 @@ FROM google/dart-runtime-base:1.23.0
...
+ENV PUB_ENVIRONMENT="bot.pub_dartlang_org.docker"
+
+RUN git clone -b alpha https://github.com/flutter/flutter.git
+# Downloads the Dart SDK and disables analytics tracking – which we always want
+RUN flutter config --no-analytics
```
mkustermann: Maybe add a warning comment here that this will add 400 MB to our custom docker layer

Dockerfile
```diff
@@ -4,9 +4,17 @@ FROM google/dart-runtime-base:1.23.0
 # docker image diff small.
```
mkustermann: Maybe change the `FROM ...` to upgrade to the latest stable version, namely 1.24.0.

Dockerfile
isoos: Please check this comment: https://github.com/dart-lang/pana/issues/17#issuecomment-306454385

I don't mind fixing it to a specific version.

Dockerfile
isoos: Alternatively, we could do most of these from the analyzer service, on its initialization. That way this extra 400MB will be on that instance only, and not on `frontend` or on soon-to-be `search`.

kevmoo:
+1

We could have a script that manages this on boot. One concern: I'm not sure
if we have to worry about reboot...

isoos:
Closing this in favor of a boot-time setup.

PR #168, opened 2017-06-13
Setup Flutter only for analyzer.

isoos:


Dockerfile
```diff
@@ -15,6 +18,9 @@ RUN pub get
 ADD app /project/app
 RUN pub get --offline
 
+# Flutter setup runs only if GAE_SERVICE=="analyzer"
+RUN /project/app/script/setup-flutter.sh
```
mkustermann: It's a nice idea, but I'm not sure if `GAE_SERVICE` is available during docker build time. Did you try this out?

I think the environment variables mentioned [here](https://cloud.google.com/appengine/docs/flexible/python/upgrading) are only guaranteed to be available during execution of the container. Even if it happens to work, it's quite fishy and I would not rely on it.

As mentioned on the original PR, we can just sacrifice the 400 MB for now to get going.

app/script/setup-flutter.sh
```diff
@@ -0,0 +1,26 @@
...
+
+if [[ -z "$FLUTTER_SDK" ]];
+then
+  echo "Flutter setup aborted: FLUTTER_SDK must point to a target directory."
+  exit 0
+fi
+
+if [ -d "$FLUTTER_SDK" ];
+then
+  echo "Flutter setup aborted: $FLUTTER_SDK already exists, assuming proper setup."
```
mkustermann: Since docker builds happen always from scratch, how can we already have this directory?

isoos: It can exists if you run the script from local dev environment (which I do - without running docker).

Dockerfile
isoos: Good catch! I wouldn't put that extra 400MB in the image, rather I did a facelift, PTAL: I've remove `setup-flutter.sh` from `Dockerfile`, and calling it from `analyzer.dart` instead.

Env variables should be initialized at that point, and there is minimal overhead compared to the setup in Docker.

app/bin/service/analyzer.dart
```diff
@@ -5,10 +5,26 @@
...
+import 'package:logging/logging.dart';
+
+final Logger logger = new Logger('pub.analyzer');
 
 void main() {
   useLoggingPackageAdaptor();
 
+  final String flutterSdkDir = Platform.environment['FLUTTER_SDK'];
+  if (flutterSdkDir == null) {
+    logger.warning('FLUTTER_SDK is not set, assuming flutter is in PATH.');
```
mkustermann: These logging messages are not going where they are supposed to go without having setup the appengine services first.

isoos:
PTAL

app/script/setup-flutter.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+if [ "$GAE_SERVICE" != "analyzer" ];
+then
+  echo "Flutter setup aborted: GAE_SERVICE must be \"analyzer\"."
+  exit 0
```
mkustermann: This could also be an `exit 1` since it should never happen.

app/script/setup-flutter.sh
```diff
@@ -0,0 +1,27 @@
...
+fi
+
+if [ -d "$FLUTTER_SDK" ];
+then
+  echo "Flutter setup aborted: $FLUTTER_SDK already exists, assuming proper setup."
+  # This shouldn't happen in docker, making sure this fails during an image build.
+  exit 1
+fi
+
+git clone -b alpha https://github.com/flutter/flutter.git $FLUTTER_SDK
```
mkustermann: Could we use `set -e` in the top of this file, to ensure a `git clone` failure (or any other failure in this shell script) is fatal?

app/bin/service/analyzer.dart
```diff
@@ -2,20 +2,44 @@
...
+  if (envConfig.flutterSdkDir == null) {
+    logger.warning('FLUTTER_SDK is not set, assuming flutter is in PATH.');
+  } else {
+    // If the script exists, it is very likely that we are inside the appengine.
+    // In local development environment the setup should happen only once, and
+    // running the setup script multiple times should be safe (no-op if
+    // FLUTTER_SDK directory exists).
+    if (FileSystemEntity.isFileSync('/project/app/script/setup-flutter.sh')) {
+      logger.warning('Setting up flutter checkout. This may take some time.');
+      await Process.run('/project/app/script/setup-flutter.sh', []);
```
mkustermann: Could we do a check for exit code here?
```dart
final ProcessResult result = await Process.run(...);
if (result.exitCode != 0) {
  logger.error('Failed to setup flutter checkout (exited with ${result.exitCode})\nstdout: ${result.stdout}\nstderr: ${result.stderr}');
}
```

PR #171, opened 2017-06-14
Use /flutter/packages instead of plugins. Fixes #143.

### LINKED ISSUE #143: Users can't find packages to use with Flutter
/flutter/plugins only shows packages which include iOS and Android code.

However users are interested in finding what packages they can use with Flutter, which is a much larger set.

My understanding is @kevmoo is working on a plan to find all of these, but in the short-term I think we can change /flutter/plugins to include all packages which have a direct dependency on sdk: flutter.

isoos:
The work is based on instructions from: https://github.com/dart-lang/pub-dartlang-dart/issues/143#issuecomment-305508387

@mkustermann: `bin/tools/backfill_search_fields.dart` must run in prod before promoting this to live.

### DIFF
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -658,11 +658,15 @@ Future<models.PackageVersion> parseAndValidateUpload(
 /// Automatically detects the built-in types.
 /// The returned list should be alphabetically sorted.
 List<String> detectTypes(Pubspec pubspec) {
-  final List<String> detectedTypes = [];
+  final Set<String> detectedTypes = new Set();
   if (pubspec.hasFlutterPlugin) {
+    detectedTypes.add(models.BuiltinTypes.flutterPackage);
     detectedTypes.add(models.BuiltinTypes.flutterPlugin);
   }
-  return detectedTypes;
+  if (pubspec.dependsOnFlutterSdk) {
+    detectedTypes.add(models.BuiltinTypes.flutterPackage);
+  }
+  return detectedTypes.toList()..sort();
 }
 
 /// Helper utility class for interfacing with Cloud Storage for storing
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -41,9 +41,10 @@ Future<shelf.Response> appHandler(
     '/search': searchHandler,
     '/packages': packagesHandler,
     '/packages.json': packagesHandler,
-    '/flutter': flutterHandler,
-    '/flutter/': flutterHandler,
-    '/flutter/plugins': flutterPluginsHandler,
+    '/flutter': redirectToFlutterPackages,
+    '/flutter/': redirectToFlutterPackages,
+    '/flutter/packages': flutterPackagesHandler,
+    '/flutter/plugins': redirectToFlutterPackages,
   }[path];
 
   if (handler != null) {
@@ -199,7 +200,7 @@ Future<shelf.Response> packagesHandlerJson(
   return jsonResponse(json);
 }
 
-/// Handles requests for `/packages` (default behavior) or `/flutter/plugins`
+/// Handles requests for `/packages` (default behavior) or `/flutter/packages`
 /// (when [basePath] is specified) - HTML
 Future<shelf.Response> packagesHandlerHtml(
   shelf.Request request,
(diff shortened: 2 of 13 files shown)


app/lib/frontend/templates.dart
```diff
@@ -578,6 +577,5 @@ abstract class LogoUrls {
   static const String flutterLogo32x32 = '/static/img/flutter-logo-32x32.png';
 }
 
-const String flutterPluginsDescriptionHtml =
-    '<p><a href="https://flutter.io/platform-plugins/">'
-    'Learn more about Flutter plugins.</a></p>';
+const String flutterPackagesDescriptionHtml =
+    '<p><a href="https://flutter.io/">Learn more about Flutter.</a></p>';
```
kevmoo: @isoos I think this is fine for now

@mit-mit @eseidelGoogle 

At the moment we have a link on the site (see here https://pub.dartlang.org/flutter/plugins) that points folks to flutter/platform-plugins

Any desire to do something different now that the index is covering all flutter packages?

eseidelGoogle:
@kevmoo we probably don't have a thing to point at yet (other than https://stackoverflow.com/questions/43101373/how-do-i-create-a-pub-package-that-works-with-flutter/43101394#43101394 ) see https://github.com/flutter/flutter/issues/10070

app/lib/frontend/templates.dart
kevmoo: @isoos – Put a comment in here referencing https://github.com/flutter/flutter/issues/10070

app/lib/frontend/handlers.dart
```diff
@@ -467,24 +468,28 @@ Future<shelf.Response> apiPackagesHandler(shelf.Request request) async {
   return _jsonResponse(json);
 }
 
-/// Handles requests for /flutter (redirects to /flutter/plugins).
+/// Handles requests for /flutter (redirects to /flutter/packages).
 shelf.Response flutterHandler(shelf.Request request) =>
```
isoos: rename to what?

isoos: nvm. `redirectToFlutterPackages` because `/flutter/plugins` can use the same.

isoos:
> For the index update – I assume running it won't break existing code before the new site is updated – right?

Code won't break, but `/flutter/packages` will be empty, until the backfill happens or a new version of them is being uploaded.

kevmoo:
Right.

From mobile. Sorry for brevity and/or spelling.

app/lib/frontend/backend.dart
```diff
@@ -658,11 +659,15 @@ Future<models.PackageVersion> parseAndValidateUpload(
 /// Automatically detects the built-in types.
 /// The returned list should be alphabetically sorted.
 List<String> detectTypes(Pubspec pubspec) {
-  final List<String> detectedTypes = [];
+  final Set<String> detectedTypes = new SplayTreeSet();
```
mkustermann: Usually we don't use the splay data structures much -- they also might go away soon. I'd just use the normal `Set()`.

app/test/frontend/handlers_test.dart
```diff
@@ -125,40 +125,41 @@ void main() {
       });
 
       tScopedTest('/flutter', () async {
-        expectRedirectResponse(await issueGet('/flutter'), '/flutter/plugins');
-        expectRedirectResponse(await issueGet('/flutter/'), '/flutter/plugins');
+        expectRedirectResponse(await issueGet('/flutter'), '/flutter/packages');
+        expectRedirectResponse(
+            await issueGet('/flutter/'), '/flutter/packages');
```
mkustermann: You could also ensure `/flutter/plugins` goes to `/flutter/packages` now.

isoos:
Updated with latest rebase.

app/lib/frontend/handlers.dart
kevmoo: 😄

app/test/frontend/handlers_test.dart
kevmoo: @isoos add this one test and I think we're good...

isoos: I did (see the line below).

kevmoo:
Sweet!

PR #173, opened 2017-06-15
Move items to shared. 

isoos:
Most of these will be used either by the `analyzer` or the `search` service.


app/lib/shared/handlers.dart
```diff
@@ -0,0 +1,56 @@
...
+      : JSON.encode(json);
+  return new shelf.Response(
+    status,
+    body: body,
+    headers: {'content-type': 'application/json; charset="utf-8"'},
+  );
+}
+
+shelf.Response htmlResponse(String content, {int status: 200}) =>
+    new shelf.Response(
```
mkustermann: For future CLs: If it doesn't fit on one line, I think it's more readable with a block instead of using `=>`.

mkustermann:
Seems like the travis build timed out again.

As I already suggested in the past: Maybe it's better to just use a global `@TestTimeout` annotation instead of tuning it very sensitively on a test basis.

PR #175, opened 2017-06-19
Experimental search service.

isoos:
Try it out:
- [experimental on staging](https://dartlang-pub-dev.appspot.com/search?experimental-service=true&q=json+xml)
- [original on staging](https://dartlang-pub-dev.appspot.com/search?q=json+xml)

The disk-based ducene is commented out, and for the time being we'll use an in-memory index. I've reached out to the project author, who implemented it promptly, but haven't published a new version yet. I'll need to run a few experiments with the proper indexing batch sizing before switching to disk index.

The use of search service is behind the `experimental-service=true` query parameter, and in case it fails, it will switch back to custom search.

### DIFF
--- a/app/bin/service/search.dart
+++ b/app/bin/service/search.dart
@@ -1,0 +1,35 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:appengine/appengine.dart';
+import 'package:shelf/shelf_io.dart' as shelf_io;
+
+import 'package:pub_dartlang_org/shared/configuration.dart';
+import 'package:pub_dartlang_org/shared/service_utils.dart';
+
+import 'package:pub_dartlang_org/search/handlers.dart';
+import 'package:pub_dartlang_org/search/index_ducene.dart';
+import 'package:pub_dartlang_org/search/updater.dart';
+
+void main() {
+  useLoggingPackageAdaptor();
+
+  withAppEngineServices(() async {
+    return withCorrectDatastore(() async {
+      registerPackageIndex(new DucenePackageIndex());
+      if (envConfig.duceneDir == null) {
+        // Don't push too much info into the in-memory index.
+        new PackageIndexUpdater(packageIndex).update(limit: 300);
+      } else {
+        // TODO: move the indexing into a different isolate.
+        new PackageIndexUpdater(packageIndex).startPolling();
+      }
+
+      await runAppEngine((HttpRequest request) =>
+          shelf_io.handleRequest(request, searchServiceHandler));
+    });
+  });
+}
--- a/app/lib/frontend/search_service.dart
+++ b/app/lib/frontend/search_service.dart
@@ -5,16 +5,23 @@
 library pub_dartlang_org.search_service;
 
 import 'dart:async';
+import 'dart:convert';
 import 'dart:math' show min;
 
 import 'package:gcloud/db.dart';
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:googleapis/customsearch/v1.dart' as customsearch;
 import 'package:googleapis_auth/auth_io.dart' as auth;
+import 'package:logging/logging.dart';
 import 'package:http/http.dart' as http;
+
+import '../shared/configuration.dart';
+import '../shared/search_service.dart' as search_service;
 
 import 'keys.dart';
 import 'models.dart';
+
+Logger _logger = new Logger('pub.frontend.search');
 
 /// The Custom Search ID used for making calls to the Custom Search API.
 const String _CUSTOM_SEARCH_ID = "009011925481577436976:h931xn2j7o0";
@@ -45,7 +52,8 @@ Future<SearchService> searchServiceViaApiKeyFromDb() async {
   final String keyString = await customSearchKeyFromDB();
   final httpClient = auth.clientViaApiKey(keyString);
   final csearch = new customsearch.CustomsearchApi(httpClient);
-  return new SearchService(httpClient, csearch);
+  final searchServiceClient = new http.Client();
+  return new SearchService(httpClient, csearch, searchServiceClient);
 }
 
 /// A wrapper around the Custom Search API, used for searching for pub packages.
@@ -62,11 +70,72 @@ class SearchService {
   /// The Custom Search client API stub.
   final customsearch.CustomsearchApi csearch;
 
-  SearchService(this.httpClient, this.csearch);
+  /// The HTTP client used for making calls to our search service.
+  final http.Client searchServiceClient;
+
+  SearchService(this.httpClient, this.csearch, this.searchServiceClient);
 
   /// Search for packes using [queryText], starting at offset [offset] returning
   /// max [numResults].
-  Future<SearchResultPage> search(SearchQuery query) async {
+  Future<SearchResultPage> search(SearchQuery query, bool useService) async {
+    if (useService) {
+      try {
+        final SearchResultPage page = await _searchService(query);
+        if (page != null) return page;
+        _logger.warning('Search service was not ready.');
+      } catch (e, st) {
+        _logger.severe('Unable to call search service.', e, st);
+      }
+    }
+    return _searchCSE(query);
+  }
+
+  Future<SearchResultPage> _searchService(SearchQuery query) async {
+    final search_service.PackageQuery packageQuery =
+        new search_service.PackageQuery(query.text,
+            type: query.type, offset: query.offset, limit: query.limit);
+    final Uri serviceUrl = new Uri(
+      scheme: 'https',
+      host: activeConfiguration.searchServiceHost,
+      path: '/search',
+      queryParameters: packageQuery.toServiceQueryParameters(),
+    );
+    final http.Response response = await searchServiceClient.get(serviceUrl);
+    if (response.statusCode == search_service.searchIndexNotReadyCode) {
+      // Search request before the service initialization completed.
+      return null;
+    }
+    if (response.statusCode != 200) {
+      // There has been an issue with the service
+      throw new Exception(
+          'Service returned status code ${response.statusCode}');
+    }
+    final search_service.PackageSearchResult result =
+        new search_service.PackageSearchResult.fromJson(
+            JSON.decode(response.body));
+    if (!result.isLegit) {
+      // Search request before the service initialization completed.
+      return null;
+    }
+
+    Key versionKey(search_service.PackageScore ps) => dbService.emptyKey
+        .append(Package, id: ps.package)
+        .append(PackageVersion, id: ps.version);
+    Key devVersionKey(search_service.PackageScore ps) => dbService.emptyKey
+        .append(Package, id: ps.package)
+        .append(PackageVersion, id: ps.devVersion);
+    final List<Key> allKeys = []
+      ..addAll(result.packages.map(versionKey))
+      ..addAll(result.packages.map(devVersionKey));
+    final List<PackageVersion> allVersions = await dbService.lookup(allKeys);
+    final versions = allVersions.sublist(0, result.packages.length);
+    final devVersions = allVersions.sublist(result.packages.length);
+
+    return new SearchResultPage(
+        query, result.totalCount, versions, devVersions);
+  }
+
+  Future<SearchResultPage> _searchCSE(SearchQuery query) async {
     bool exists(x) => x != null;
     final db = dbService;
 
--- a/app/lib/search/handlers.dart
+++ b/app/lib/search/handlers.dart
@@ -1,0 +1,51 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:shelf/shelf.dart' as shelf;
+
+import '../shared/handlers.dart';
+import '../shared/search_service.dart';
+
+import 'index_ducene.dart';
+
+/// Handlers for the search service.
+Future<shelf.Response> searchServiceHandler(shelf.Request request) async {
+  final path = request.requestedUri.path;
+  final handler = {
+    '/debug': debugHandler,
+    '/search': searchHandler,
+    // TODO: add handler for update notification
+  }[path];
+
+  if (handler != null) {
+    return handler(request);
+  } else {
+    return notFoundHandler(request);
+  }
+}
+
+/// Handler /debug requests
+Future<shelf.Response> debugHandler(shelf.Request request) async {
+  return jsonResponse({
+    'ready': packageIndex.isReady,
+    'indexSize': await packageIndex.indexSize(),
+    'currentRss': ProcessInfo.currentRss,
+    'maxRss': ProcessInfo.maxRss,
+  }, indent: true);
+}
+
+/// Handles /search requests.
+Future<shelf.Response> searchHandler(shelf.Request request) async {
+  if (!packageIndex.isReady) {
+    return htmlResponse(searchIndexNotReadyText,
+        status: searchIndexNotReadyCode);
+  }
+  final bool indent = request.url.queryParameters['indent'] == 'true';
+  final PackageSearchResult result = await packageIndex.search(
+      new PackageQuery.fromServiceQueryParameters(request.url.queryParameters));
+  return jsonResponse(result.toJson(), indent: indent);
+}
--- a/app/lib/search/index_ducene.dart
+++ b/app/lib/search/index_ducene.dart
@@ -1,0 +1,243 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+import 'dart:math';
+
+import 'package:ducene/analysis.dart';
+import 'package:ducene/document.dart';
+import 'package:ducene/search.dart';
+import 'package:ducene/util.dart';
+import 'package:gcloud/service_scope.dart' as ss;
+
+import '../shared/configuration.dart';
+import '../shared/search_service.dart';
+
+/// The [PackageIndex] registered in the current service scope.
+PackageIndex get packageIndex => ss.lookup(#packageIndexService);
+
+/// Register a new [PackageIndex] in the current service scope.
+void registerPackageIndex(PackageIndex index) =>
+    ss.register(#packageIndexService, index);
+
+class DucenePackageIndex implements PackageIndex {
+  final _MultiNgramAnalyzer _analyzer = new _MultiNgramAnalyzer(3);
+  Directory _fsDir;
+  IndexHolder _index;
+  DateTime _lastUpdated;
+  Future _initFuture;
+
+  Future _init() async {
+    if (_initFuture != null) return _initFuture;
+    assert(_index == null);
+    _initFuture = new Future(() async {
+      IndexHolderDirectory indexDir;
+      if (envConfig.duceneDir != null) {
+        // Temporarily everything is in memory, because ducene is not using async
+        // IO, and that blocks appengine's health checks, killing the machine
+        // before it got any chance to setup.
+        // TODO: fix ducene to handle async FS properly
+        // _fsDir = new Directory(envConfig.duceneDir);
+        // indexDir = new FSIndexHolderDirectory(_fsDir);
+        indexDir = new RAMIndexHolderDirectory();
+      } else {
+        indexDir = new RAMIndexHolderDirectory();
+      }
+      _index = await DirectoryHolder.open(indexDir);
+    });
+    return _initFuture;
+  }
+
+  @override
+  bool get isReady => _lastUpdated != null;
+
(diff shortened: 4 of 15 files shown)


isoos:
I've also structured the code in a way that requires very little modification on the current search API, because I wanted to limit the surface area in case of a rollback is needed. If the search service is to our satisfaction, the refactor and merge of the interfaces will be easy to do.

Btw. the important part of the example query above is the quality: with the current query, searching for `xml json` yields the `xml2json` package, but other than that, the results on the first page are not really relevant. I think the new search service seems to provide better results.

search.yaml
```diff
@@ -0,0 +1,25 @@
...
+# BSD-style license that can be found in the LICENSE file.
+
+runtime: custom
+env: flex
+service: search
+env_variables:
+  # Needs to be in sync with lib/shared/configuration.dart
+  DUCENE_DIR: '/ducene'
+
+resources:
```
kevmoo: https://cloud.google.com/compute/docs/machine-types#standard_machine_types

FYI – I think you get a machine type that is at least the requirements you put here.

...so I believe this would bump us up to a 2-cpu box.

So drop ram to 3.5?

app/lib/shared/search_service.dart
```diff
@@ -0,0 +1,152 @@
...
+
+  factory PackageSearchResult.fromJson(Map json) => new PackageSearchResult(
+        lastUpdated: json['lastUpdated'],
+        totalCount: json['totalCount'],
+        packages: (json['packages'] ?? [])
+            .map((Map m) => new PackageScore.fromJson(m))
+            .toList(),
+      );
+
+  bool get isLegit => lastUpdated != null;
```
kevmoo: For most things, I can figure out what's going on.

`isLegit` – could probably use a doc comment. 😄

app/lib/shared/search_service.dart
```diff
@@ -0,0 +1,152 @@
...
+  Future<PackageSearchResult> search(PackageQuery query);
+}
+
+/// A summary information about a package that goes into the search index.
+///
+/// It is also part of the data structure returned by a search query, except for
+/// the [readme] and [popularity] fields, which are excluded when returning the
+/// results.
+class PackageDocument {
+  String url;
```
kevmoo: could be finals?

app/lib/shared/search_service.dart
isoos: Good catch! I was planning to use it but forgot, fixed.

The error case it tries to detect is the following: a new search service instance is up, and by accident we migrate traffic to it before it has been able to update its index. In such cases the results are not to be trusted, and we'll just fall back to custom search.

app/lib/search/handlers.dart
```diff
@@ -0,0 +1,47 @@
...
+    'ready': packageIndex.isReady,
+    'indexSize': await packageIndex.indexSize(),
+    'currentRss': ProcessInfo.currentRss,
+    'maxRss': ProcessInfo.maxRss,
+  }, indent: true);
+}
+
+/// Handles /search requests.
+Future<shelf.Response> searchHandler(shelf.Request request) async {
+  final PackageSearchResult result = await packageIndex.search(
```
mkustermann: Shouldn't we check for `packageIndex.isReady` here and return some kind of "not ready" status code back to the requester?

app/lib/frontend/search_service.dart
```diff
@@ -66,7 +73,55 @@ class SearchService {
...
+        new search_service.PackageQuery(query.text,
+            type: query.type, offset: query.offset, limit: query.limit);
+    final Uri serviceUrl = new Uri(
+      scheme: 'https',
+      host: activeConfiguration.searchServiceHost,
+      path: '/search',
+      queryParameters: packageQuery.toServiceQueryParameters(),
+    );
+    final http.Response response = await httpClient.get(serviceUrl);
+    final search_service.PackageSearchResult result =
```
mkustermann: We should probably check `response.statusCode` here.

app/lib/search/updater.dart
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:logging/logging.dart';
+import 'package:gcloud/db.dart';
```
mkustermann: nit: sorting imports

app/lib/search/updater.dart
```diff
@@ -0,0 +1,83 @@
...
+
+    int count = 0;
+    await for (Package p in dbService.query(Package).run()) {
+      versionKeys.add(p.latestVersionKey);
+      devVersions[p.name] = p.latestDevVersion;
+      count++;
+      if (limit != null && limit <= count) break;
+    }
+
+    for (List<Key> slices in sliceList(versionKeys, 100)) {
```
mkustermann: Maybe lower the list from 100 to more like 20 here.

The issue is that `PackageVersion` contains README/CHANGELOG/pubspec.yaml so they can make the protobuf we get back quite big. There is a limit on how big a response to a single lookup request can be, so to avoid flakiness it might make sense to lower this from the current 100.

app/lib/search/updater.dart
```diff
@@ -0,0 +1,83 @@
...
+      final List<PackageVersion> versions = await dbService.lookup(slices);
+      final List<PackageDocument> documents = versions
+          .map((pv) => new PackageDocument(
+                url: 'https://pub.dartlang.org/packages/${pv.package}',
+                package: pv.package,
+                version: pv.version,
+                devVersion: devVersions[pv.package] ?? pv.version,
```
mkustermann: When will we ever execute `?? pv.version` ?

app/lib/search/updater.dart
```diff
@@ -0,0 +1,83 @@
...
+      _updating = false;
+    }
+  }
+
+  Future _doUpdate(int limit) async {
+    final List<Key> versionKeys = [];
+    final Map<String, String> devVersions = {};
+
+    int count = 0;
+    await for (Package p in dbService.query(Package).run()) {
```
mkustermann: Consider adding a bit more logging information, so one can better follow what the search service is doing atm (but not too much output) to have some context if an exception occurs / ....

search.yaml
mkustermann: Not sure what kevin means, but the cpu/memory_gb just get rounded up sometimes. It's fine to specify our minimum requirements, to make it explicit how much cpu/ram we're guaranteed to get.

It might be better to double the cpu/ram here.

app/pubspec.yaml
```diff
@@ -6,6 +6,7 @@ sdk: ">=1.19.0 <2.0.0"
 dependencies:
   appengine: '^0.4.0'
   args: '^0.13.7'
+  ducene: '^0.4.0'
```
mkustermann: This is a 3rd-party package and every upgrade can potentially cause us to pull in malicious code. The only other package is `uuid` I think.

Maybe put them into their own section and add a comment that these are 3rd-party packages we are relying on.

We could also think about pin-pointing the version exactly instead of having a loose constraint.

app/bin/service/search.dart
```diff
@@ -0,0 +1,34 @@
...
+  useLoggingPackageAdaptor();
+
+  withAppEngineServices(() async {
+    return withCorrectDatastore(() async {
+      registerPackageIndex(new DucenePackageIndex());
+      if (envConfig.duceneDir == null) {
+        // Don't push too much info into the in-memory index.
+        new PackageIndexUpdater(packageIndex).update(limit: 300);
+      } else {
+        new PackageIndexUpdater(packageIndex).startPolling();
```
mkustermann: Please add a TODO here to move the indexing into a different isolate.

We should ensure that the isolate running `runAppEngine` (which is serving http requests) is always responsive with low latency. The indexing can happen in the background.

The other isolate can use the same setup procedure as here, just without `runAppEngine()`.

app/lib/search/index_ducene.dart
```diff
@@ -0,0 +1,226 @@
...
+import 'package:gcloud/service_scope.dart' as ss;
+
+import '../shared/configuration.dart';
+import '../shared/search_service.dart';
+
+/// The [PackageIndex] registered in the current service scope.
+PackageIndex get packageIndex => ss.lookup(#_ducene_package_index);
+
+/// Register a new [PackageIndex] in the current service scope.
+void registerPackageIndex(PackageIndex index) =>
```
mkustermann: For the other things we use the service scope for we postfix the name with service. Maybe we should also do that here: `packageIndexService`.

app/lib/search/index_ducene.dart
```diff
@@ -0,0 +1,226 @@
...
+  DateTime _lastUpdated;
+  final _MultiNgramAnalyzer _analyzer = new _MultiNgramAnalyzer(3);
+
+  Future _init() async {
+    if (_index != null) return;
+    IndexHolderDirectory indexDir;
+    if (envConfig.duceneDir != null) {
+      // Temporarily everything is in memory, because ducene is not using async
+      // IO, and that blocks appengine's health checks, killing the machine
+      // before it got any chance to setup.
```
mkustermann: As mentioned above, we should move the actual indexing into a different isolate, so the main isolate can provide low-latency responses to search queries.

AFAIK even when code runs synchronous i/o it should not block other isolates (fingers crossed that this is true)

app/lib/search/index_ducene.dart
```diff
@@ -0,0 +1,226 @@
...
+
+/// Register a new [PackageIndex] in the current service scope.
+void registerPackageIndex(PackageIndex index) =>
+    ss.register(#_ducene_package_index, index);
+
+class DucenePackageIndex implements PackageIndex {
+  Directory _fsDir;
+  IndexHolder _index;
+  DateTime _lastUpdated;
+  final _MultiNgramAnalyzer _analyzer = new _MultiNgramAnalyzer(3);
```
mkustermann: nit: maybe move the final above the non-finals.

app/lib/search/index_ducene.dart
```diff
@@ -0,0 +1,226 @@
...
+      // before it got any chance to setup.
+      // TODO: fix ducene to handle async FS properly
+      // _fsDir = new Directory(envConfig.duceneDir);
+      // indexDir = new FSIndexHolderDirectory(_fsDir);
+      indexDir = new RAMIndexHolderDirectory();
+    } else {
+      indexDir = new RAMIndexHolderDirectory();
+    }
+    _index = await DirectoryHolder.open(indexDir);
+  }
```
mkustermann: There is a race, when two calls to `_init()` are in the middle of `DirectoryHolder.open()` only the last one will win and set the `_index`.

Consider just doing something like this:

```dart
class ... {

  Future readFuture;

  Future _init()  {
    if (readyFuture != null) return readyFuture;
    final Completer completer = new Completer();
    readyFuture = completer.future;

    new Future(() async {
      ...
      await ...;
      completer.complete();
    });
  }
}
```

app/lib/search/index_ducene.dart
```diff
@@ -0,0 +1,226 @@
...
+
+  @override
+  Iterable<String> tokenize(String text) {
+    final Set<String> ngrams = new Set();
+    for (int ngramLength = 1; ngramLength <= _n; ngramLength++) {
+      if (text.length <= ngramLength) {
+        ngrams.add(text);
+      } else {
+        for (int i = 0; i <= text.length - ngramLength; i++) {
+          ngrams.add(text.substring(i, i + ngramLength));
```
mkustermann: Is this doing character-n-grams ? That seems like a lot of data. I thought we're talking about word-n-grams.

app/lib/search/index_ducene.dart
```diff
@@ -0,0 +1,226 @@
...
+    }
+    return t;
+  }
+
+  String _normalizeText(String text, int maxLength) {
+    if (text == null) return '';
+    String t = text
+        .toLowerCase()
+        .replaceAll(new RegExp('\\W'), ' ')
+        .replaceAll(new RegExp('\\s+'), ' ')
```
mkustermann: Make these global final RegExp's so the VM can optimize them properly when invoking many times.

app/lib/search/index_ducene.dart
```diff
@@ -0,0 +1,226 @@
...
+    return new PackageSearchResult(
+      lastUpdated: _lastUpdated.toIso8601String(),
+      totalCount: totalCount,
+      packages: packages,
+    );
+  }
+
+  String _normalizeWhitespaces(String text, int maxLength) {
+    if (text == null) return '';
+    String t = text.replaceAll(new RegExp('\\s+'), ' ').trim();
```
mkustermann: See comment further down about RegExps

app/lib/search/index_ducene.dart
```diff
@@ -0,0 +1,226 @@
...
+          .addFilter(new TermQuery(new Term('detectedTypes', query.type)));
+    }
+
+    final int offset = query.offset ?? 0;
+    final int limit = max(minSearchLimit, query.limit ?? 100);
+
+    final TopDocs topPackages = await searcher.search(
+      duceneQuery,
+      maxSearchResults + minSearchLimit, // additional results for extended bias
+      scoreSort: new TFIDFScoreSort(),
```
mkustermann: tf-idf : Nice :+1:

app/lib/search/index_ducene.dart
```diff
@@ -0,0 +1,226 @@
...
+        doc: new PackageDocument(
+          url: doc.get('id'),
+          package: doc.get('package'),
+          version: doc.get('version'),
+          devVersion: doc.get('devVersion'),
+          description: doc.get('originalDescription'),
+          lastUpdated: doc.get('lastUpdated'),
+          detectedTypes: doc.getValues('detectedTypes'),
+        ),
+        score: sd.score + bias,
```
mkustermann: Maybe add a comment here how these two doubles get combined. What is the range of them? Do they have 50/50 weight?

app/lib/search/index_ducene.dart
isoos: Yeah, it is the 1-gram, 2-gram, and 3-gram split of the string.
We may switch to words, but as long as we fit in memory easily, I don't mind having these around.

app/lib/search/index_ducene.dart
isoos: Added a TODO. This part is underdeveloped right now.

isoos:
I've also added a check in order to skip packages that haven't changed since last indexing.

app/lib/search/index_ducene.dart
mkustermann: We can iterate on it, I just thought that using up to 3-character n-grams gives worse result than using word n-grams with some preprocessing/normalization of the words. Actually I'm quite positively surprised that the character-n-grams work so well.

app/lib/frontend/search_service.dart
```diff
@@ -62,11 +70,65 @@ class SearchService {
...
+  SearchService(this.httpClient, this.csearch, this.searchServiceClient);
 
   /// Search for packes using [queryText], starting at offset [offset] returning
   /// max [numResults].
-  Future<SearchResultPage> search(SearchQuery query) async {
+  Future<SearchResultPage> search(SearchQuery query, bool useService) async {
+    if (useService) {
+      try {
+        final SearchResultPage page = await _searchService(query);
+        if (page != null) return page;
```
mkustermann: Maybe add a logging call after the if, so we know when we fall back to CSE in the case where no exception is raised.

app/lib/frontend/search_service.dart
```diff
@@ -62,11 +70,65 @@ class SearchService {
...
+            type: query.type, offset: query.offset, limit: query.limit);
+    final Uri serviceUrl = new Uri(
+      scheme: 'https',
+      host: activeConfiguration.searchServiceHost,
+      path: '/search',
+      queryParameters: packageQuery.toServiceQueryParameters(),
+    );
+    final http.Response response = await searchServiceClient.get(serviceUrl);
+    if (response.statusCode == search_service.searchIndexNotReadyCode) {
+      // Search request while service instance is updating index.
```
mkustermann: The comment is a bit misleading, since in the future, even if the index is being updated (in one isolate) we still expect the search service to serve using the old index. So the only downtime is during the very first time the index is being built (even new deployments will be able to use the saved old index, causing therefore no downtime).

mkustermann: We might be better off checking for three exit codes: 
  * the positive one if the response was correctly satisfied
  * the negative one where the service is up & running but is building the very first index
  * everything else, e.g. 500s, internal errors, ... (should log these as errors)

app/lib/shared/search_service.dart
```diff
@@ -0,0 +1,156 @@
...
+
+  Map toJson() => {
+        'lastUpdated': lastUpdated,
+        'totalCount': totalCount,
+        'packages': packages,
+      };
+}
+
+class PackageScore {
+  final PackageDocument doc;
```
mkustermann: I think actually there is no need to return the entire `PackageDocument` which was used for indexing: The user of the search service only needs to know:
  * the package/version
  * the score
  * maybe a text snippet where the search terms occured (though even that is not used by our user of the search service atm)

Just imagine what would happen if we extend the `PackageDocument` to not only include the readme/changelog but also dart source code of the package to be indexed. It's not really useful to return all of that data I think.

app/lib/shared/search_service.dart
```diff
@@ -0,0 +1,156 @@
...
+  final double score;
+
+  PackageScore({this.doc, this.score});
```
mkustermann: Shouldn't we make these two mandatory positional arguments?

app/lib/shared/search_service.dart
```diff
@@ -0,0 +1,156 @@
...
+  Map<String, String> toServiceQueryParameters() => {
+        'q': text,
+        'type': type,
+        'offset': offset?.toString(),
+        'limit': limit?.toString(),
+      };
+}
+
+class PackageSearchResult {
+  final String lastUpdated;
```
mkustermann: Maybe add a comment that this `lastUpdated` means actually the last run of the indexing system (as opposed to the `PackageDocument.lastUpdated`, which should just be the last time the package/version was indexed, which can be much longer ago, if there were no new "pub publish"es since then)

app/lib/shared/search_service.dart
```diff
@@ -0,0 +1,156 @@
...
+  bool get isLegit => lastUpdated != null;
+
+  Map toJson() => {
+        'lastUpdated': lastUpdated,
+        'totalCount': totalCount,
+        'packages': packages,
+      };
+}
+
+class PackageScore {
```
mkustermann: Consider renaming these classes a bit:
  * `PackageSearchResult` -> `(Package)?SearchResults`
  * `PackageScore` -> `(Package)?SearchResult`

Not sure if the `Package` prefix is necessary.

app/lib/search/index_ducene.dart
```diff
@@ -0,0 +1,245 @@
...
+          devVersion: doc.get('devVersion'),
+          description: doc.get('originalDescription'),
+          lastUpdated: doc.get('lastUpdated'),
+          detectedTypes: doc.getValues('detectedTypes'),
+        ),
+        // TODO: normalize values and calibrate bias weight
+        score: sd.score + bias,
+      ));
+    }
+    packages.sort((a, b) => -Comparable.compare(a.score, b.score));
```
mkustermann: I think you can just do `packages.sort((a, b) => return -a.score.compareTo(b.score))` since they are both doubles.

app/lib/search/index_ducene.dart
```diff
@@ -0,0 +1,245 @@
...
+  Future removeUrl(String url) async {
+    await _init();
+    await _index.deleteDocuments(new BoolQuery()..append('id', url));
+  }
+
+  @override
+  Future merge() async {
+    await _init();
+    await _index.forceMerge();
+    _lastUpdated = new DateTime.now();
```
mkustermann: When a service is storing datetimes or returning them to other services, we should always use UTC time.

Maybe it would also make sense to add an `assert(lastUpdated.isUtc)` in the serialization logic.

mkustermann:
A few last comments, mostly nits. Then we're ready for merging.

app/lib/shared/search_service.dart
isoos: I was contemplating to introduce a new type for it, but the overlap was too big. I will do it now.

We do not returning the readme content (nor the popularity score), but I do think there is some value returning the extra fields, e.g. we could take them on their face value and skip the datastore for additional sanity check. Nevertheless, I'll strip the results down a bit for now, we can extend them later if needed.

app/lib/shared/search_service.dart
isoos: A little nit: if/when we were to introduce a package-name-typeahead feature, and serve it through the search API, what would be the name of that?
Similar question for code search, but I strongly suspect that it can be part of the `PackageSearchResult`.

mkustermann: Would think we could have something like this:
```dart

class FuzzySearchApi {
  ...
  Future<FuzzyMatchResults> fuzzyMatch(String searchTerm);
  ...
}

class FuzzyMatchResults {
  final String searchTerm;

  final int totalResultCount; // <- maybe unnecessary, we might only want to support top-N
  final int offsetInResults;    // <- maybe unnecessary, we might only want to support top-N
  final List<FuzzyMatchResult> matches;
}

class FuzzyMatchResult {
  final String package;
  final double score;
  FuzzyMatchResult(this.package, this.score);
}
```

Then keep the list of all package names in memory and do fuzzy matching on them (e.g. dwt or something else).

PR #176, opened 2017-06-19
Proposed URLs for analyzer service

isoos:
I think the URLs should do the following:
- GET /packages/[package]: returns the latest analysis of the latest stable version of the package
- GET /packages/[package]/[version]: returns the latest analysis of the given package + version (if exists)
- GET /packages/[package]/[version]/[analysis]: returns the specific analysis of the given package + version (if exists)
- POST /packages/[package]: trigger an asynchronous analysis of the package (latest version will be analyzed)
- POST /packages/[package]/[version]: trigger an asynchronous analysis of the package + version

### DIFF
--- a/app/lib/analyzer/handlers.dart
+++ b/app/lib/analyzer/handlers.dart
@@ -1,0 +1,75 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:shelf/shelf.dart' as shelf;
+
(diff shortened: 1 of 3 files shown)


app/lib/analyzer/handlers.dart
```diff
@@ -0,0 +1,76 @@
...
+import 'dart:async';
+import 'dart:io';
+
+import 'package:shelf/shelf.dart' as shelf;
+
+import '../shared/handlers.dart';
+
+/// Handlers for the analyzer service.
+Future<shelf.Response> analyzerServiceHandler(shelf.Request request) async {
+  logPubHeaders(request);
```
mkustermann: As in the PR about search service, this is never invoked by the `pub` client itself and the other users of the API don't supply the `X-Pub-*` headers, so I think this can be safely removed.

PR #177, opened 2017-06-20
Analyzer model + backend

isoos:
Datastore model + a bit of backend operations (to store and get the analysis model object).

### DIFF
--- a/app/lib/analyzer/backend.dart
+++ b/app/lib/analyzer/backend.dart
@@ -1,0 +1,88 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:math';
+
+import 'package:gcloud/db.dart';
+
+import 'models.dart';
+
+/// Datastore-related access methods for the analyzer service
+class AnalysisBackend {
+  final DatastoreDB db;
+
+  AnalysisBackend(this.db);
+
+  /// Gets the [Analysis] object from the Datastore. [version] and [analysis] is
+  /// optional, when they are missing the analysis from the latest rollup will
+  /// be returned.
+  Future<Analysis> getAnalysis(String package,
+      [String version, int analysis]) async {
+    final Key packageKey = db.emptyKey.append(PackageAnalysis, id: package);
+
+    if (version == null) {
+      final list = await db.lookup([packageKey]);
+      final PackageAnalysis pa = list[0];
+      if (pa == null) return null;
+      version = pa.latestVersion;
+    }
+
+    // version was set
+    final Key versionKey =
+        packageKey.append(PackageVersionAnalysis, id: version);
+
+    if (analysis == null) {
+      final list = await db.lookup([versionKey]);
+      final PackageVersionAnalysis pva = list[0];
+      if (pva == null) return null;
+      analysis = pva.latestAnalysis;
+    }
+
+    // analysis was set
+    final Key analysisKey = versionKey.append(Analysis, id: analysis);
+    final List list = await db.lookup([analysisKey]);
+    return list[0];
+  }
+
+  /// Stores the analysis, and either creates or updates its parent
+  /// [PackageAnalysis] and [PackageVersionAnalysis] records.
+  Future storeAnalysis(Analysis analysis) async {
+    assert(analysis.id == null);
+    await db.commit(inserts: [analysis]);
+    assert(analysis.id != null);
+
+    // update package and version too
+    await db.withTransaction((Transaction tx) async {
+      final Key packageKey =
+          db.emptyKey.append(PackageAnalysis, id: analysis.packageName);
+      final Key packageVersionKey = packageKey.append(PackageVersionAnalysis,
+          id: analysis.packageVersion);
+      final List parents = await tx.lookup([packageKey, packageVersionKey]);
+      final PackageAnalysis package = parents[0];
+      final PackageVersionAnalysis version = parents[1];
+
+      final List<Model> inserts = [];
+      if (package == null) {
+        inserts.add(new PackageAnalysis.fromAnalysis(analysis));
+      } else if (package.updateWithLatest(analysis)) {
+        inserts.add(package);
+      }
+
+      if (version == null) {
+        inserts.add(new PackageVersionAnalysis.fromAnalysis(analysis));
+      } else if (version.updateWithLatest(analysis)) {
+        inserts.add(version);
+      }
+
+      if (inserts.isEmpty) {
+        await tx.rollback();
+        return;
+      }
+
+      tx.queueMutations(inserts: inserts);
+      await tx.commit();
+    });
+  }
+}
--- a/app/lib/analyzer/model_properties.dart
+++ b/app/lib/analyzer/model_properties.dart
@@ -1,0 +1,51 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'package:gcloud/db.dart';
+
+/// These status codes mark the status of the analysis, not the result/report.
+///
+/// WARNING: Do not rename these enum values, since they are stored in the DB.
+enum AnalysisStatus {
+  /// Analysis was aborted without a report.
+  /// Probably an issue with pana or an unhandled edge case.
+  aborted,
+
+  /// Analysis was completed but there are missing parts.
+  /// One or more tools failed to produce the expected output.
+  failure,
+
+  /// Analysis was completed without issues.
+  success,
+}
+
+class AnalysisStatusProperty extends StringProperty {
+  const AnalysisStatusProperty({String propertyName, bool required: false})
+      : super(propertyName: propertyName, required: required, indexed: true);
+
+  @override
+  bool validate(ModelDB db, Object value) =>
+      (!required || value != null) &&
+      (value == null || value is AnalysisStatus);
+
+  @override
+  String encodeValue(ModelDB db, Object value, {bool forComparison: false}) {
+    if (value is AnalysisStatus) {
+      return value.toString().split('.').last;
+    }
+    return null;
+  }
+
+  @override
+  Object decodePrimitiveValue(ModelDB db, Object value) {
+    if (value is String) {
+      return AnalysisStatus.values
+          .firstWhere((as) => as.toString().endsWith('.$value'));
+    }
+    if (value == null) return null;
+    throw new Exception('Unknown analysis status: $value');
+  }
+}
--- a/app/lib/analyzer/models.dart
+++ b/app/lib/analyzer/models.dart
@@ -1,0 +1,134 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:gcloud/db.dart' as db;
+import 'package:pub_semver/pub_semver.dart';
+
+import 'model_properties.dart';
+
+/// Analyzed package report.
+@db.Kind(name: 'PackageAnalysis', idType: db.IdType.String)
+class PackageAnalysis extends db.ExpandoModel {
+  String get packageName => id;
+
+  @db.ModelKeyProperty(required: true)
+  db.Key latestVersionKey;
+
+  String get latestVersion => latestVersionKey.id;
+  Version get latestSemanticVersion => new Version.parse(latestVersion);
+
+  PackageAnalysis();
+  PackageAnalysis.fromAnalysis(Analysis analysis) {
+    parentKey = db.dbService.emptyKey;
+    id = analysis.packageName;
+    latestVersionKey = analysis.packageVersionKey;
+  }
+
+  /// Returns true if there was a change that warrants an update in Datastore.
+  bool updateWithLatest(Analysis analysis) {
+    if (isNewer(analysis.semanticPackageVersion, latestSemanticVersion)) {
+      // Current package version is already the latest stable version.
+      return false;
+    }
+    package.latestVersionKey = analysis.packageVersionKey;
+    return true;
+  }
+}
+
+/// Analyzed version report.
+@db.Kind(name: 'PackageVersionAnalysis', idType: db.IdType.String)
+class PackageVersionAnalysis extends db.ExpandoModel {
+  db.Key get packageKey => parentKey;
+  String get packageName => packageKey.id;
+  String get packageVersion => id;
+
+  Version get semanticPackageVersion => new Version.parse(packageVersion);
+
+  @db.ModelKeyProperty(required: true)
+  db.Key latestAnalysisKey;
+
+  int get latestAnalysis => latestAnalysisKey.id;
+
+  @db.DateTimeProperty()
+  DateTime analysisTimestamp;
+
+  @db.StringProperty()
+  String analysisVersion;
+
+  Version get semanticAnalysisVersion => new Version.parse(analysisVersion);
+
+  @AnalysisStatusProperty()
+  AnalysisStatus analysisStatus;
+
+  PackageVersionAnalysis();
+  PackageVersionAnalysis.fromAnalysis(Analysis analysis) {
+    parentKey =
+        db.dbService.emptyKey.append(PackageAnalysis, id: analysis.packageName);
+    id = analysis.packageVersion;
+    latestAnalysisKey = analysis.key;
+    analysisTimestamp = analysis.timestamp;
+    analysisVersion = analysis.analysisVersion;
+    analysisStatus = analysis.analysisStatus;
+  }
+
+  /// Returns true if there was a change that warrants an update in Datastore.
+  bool updateWithLatest(Analysis analysis) {
+    if (isNewer(analysis.semanticAnalysisVersion, semanticAnalysisVersion)) {
+      // the new analysis' pana version is older, result probably obsolete
+      return false;
+    }
+    version.latestAnalysisKey = analysis.key;
+    version.analysisTimestamp = analysis.timestamp;
+    version.analysisVersion = analysis.analysisVersion;
+    version.analysisStatus = analysis.analysisStatus;
+    return true;
+  }
+}
+
+/// Immutable record of an analysis.
+@db.Kind(name: 'Analysis', idType: db.IdType.Integer)
+class Analysis extends db.ExpandoModel {
+  @db.StringProperty(required: true)
+  String packageName;
+
+  @db.StringProperty(required: true)
+  String packageVersion;
+
+  int get analysis => id;
+  db.Key get packageVersionKey => parentKey;
+
+  Version get semanticPackageVersion => new Version.parse(packageVersion);
+
+  @db.DateTimeProperty()
+  DateTime timestamp;
+
+  // Analysis fields contain information about the pana version being used,
+  // how its execution went, and the raw output.
+
+  @db.StringProperty()
+  String analysisVersion;
+
+  Version get semanticAnalysisVersion => new Version.parse(analysisVersion);
+
+  @AnalysisStatusProperty()
+  AnalysisStatus analysisStatus;
+
+  @db.StringProperty(indexed: false)
+  String analysisJsonContent;
+
+  // Report fields contain our interpretation and relevant extracts of the pana
+  // analysis output.
+  // TODO: add report fields
+
+  /// Empty constructor only for appengine.
+  Analysis();
+
+  /// Proper constructor that initializes the required fields.
+  Analysis.init(this.packageName, this.packageVersion) {
+    timestamp = new DateTime.now().toUtc();
+    parentKey = db.dbService.emptyKey
+        .append(PackageAnalysis, id: package)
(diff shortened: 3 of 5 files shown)


app/lib/shared/utils.dart
```diff
@@ -67,6 +67,38 @@ String canonicalizeVersion(String version) {
...
 }
 
+/// Compares two versions according to the semantic versioning specification.
+///
+/// If [pubSorting] is `true` then pub's priorization ordering is used, which
+/// will rank pre-release versions lower than stable versions (e.g. it will
+/// order "0.9.0-dev.1 < 0.8.0").  Otherwise it will use semantic version
+/// sorting (e.g. it will order "0.8.0 < 0.9.0-dev.1").
+int compareSemanticVersionsDesc(
+    semver.Version a, semver.Version b, bool decreasing, pubSorting) {
```
kevmoo: `bool pubSorted`

app/lib/analyzer/models.dart
```diff
@@ -0,0 +1,99 @@
...
+class PackageVersionAnalysis extends db.ExpandoModel with AnalysisRollup {
+  @db.ModelKeyProperty(required: true, propertyName: 'package')
+  db.Key packageKey;
+
+  String get package => packageKey.id;
+
+  @db.StringProperty(required: true)
+  String version; // Same as id
+
+  /// See also [AnalysisRollup] fields.
```
kevmoo: Weird spot for a doc comment – should be on class?

app/lib/analyzer/backend.dart
```diff
@@ -0,0 +1,153 @@
...
+      final List list = await db.lookup([packageKey]);
+      return loadRollup(list[0]);
+    }
+
+    // version was set
+    final Key versionKey =
+        packageKey.append(PackageVersionAnalysis, id: version);
+
+    if (analysis == null) {
+      final List list = await db.lookup([versionKey]);
```
kevmoo: `var list` is fine, too.

https://www.dartlang.org/guides/language/effective-dart/usage#consider-omitting-the-types-for-local-variables

😄

app/lib/analyzer/models.dart
isoos: If you look at the class, if is clear that it inherits fields from another class. However, if you just glance at the list of fields, it may not be obvious. I thought a reminder will be nice there.

app/lib/analyzer/backend.dart
isoos: I will leave it this way, because:
- with `var`, the linter complains about being non-final
- without `List` the analyzer complains that there is a type missmatch

app/lib/analyzer/models.dart
kevmoo: Then (perhaps) just use `//` instead of `///`.

The latter implies doc comments.

app/lib/analyzer/backend.dart
kevmoo: Ah! Use `as` then. Then we'll get runtime checks that the value returned is actually a list.

I should look at the lints...that seems overly strict.

app/lib/analyzer/models.dart
```diff
@@ -0,0 +1,99 @@
...
+  Version get semanticVersion => new Version.parse(version);
+
+  @db.DateTimeProperty()
+  DateTime timestamp;
+
+  @db.StringProperty()
+  String analysisVersion;
+
+  @db.StringProperty()
+  String analysisStatus;
```
mkustermann: If you like, you can make your own subclass of `StringProperty` which is encoding/decoding from an enum, so we can use a real dart class instead of the string:

```dart
@db.AnalysisStatusProperty()
AnalysisStatus analysisStatus;

....

enum AnalysisStatus {
  Pending,
  Failure,
  Success,
  Aborted,
}
```

There should be an example of how to do this somewhere, iirc `model_properties.dart`, it's pretty simple.

(Maybe `AnalysisStatus` -> `AnalysisResult`)

Could do the same for the `reportJson`

app/lib/analyzer/models.dart
```diff
@@ -0,0 +1,99 @@
...
+  String get package => packageKey.id;
+
+  @db.StringProperty(required: true)
+  String version; // Same as id
+
+  // See also [AnalysisRollup] fields.
+}
+
+/// Immutable record of an analysis.
+@db.Kind(name: 'Analysis', idType: db.IdType.String)
```
mkustermann: If it's an immutable record, we could just use `idType: db.IdType.Integer` and let datastore generate the id for us. Otherwise there's always the question of whether the new `Analysis` object we want to store is already present in the datastore or not.

app/lib/analyzer/models.dart
```diff
@@ -0,0 +1,99 @@
...
+  String analysisVersion;
+
+  @db.StringProperty()
+  String analysisStatus;
+
+  @db.StringProperty()
+  String reportVersion;
+
+  @db.StringProperty(indexed: false)
+  String reportJson;
```
mkustermann: Don't we have an implicit link between analysisVersion and reportVersion? We are determining which version of kevin's analysis package we use, so it's almost pinpointed. So whenever we roll a new version of that, we can bump analysis version, avoiding the distinction between `analysisVersion` and `reportVersion`, since the first implies the second.

app/lib/analyzer/models.dart
```diff
@@ -0,0 +1,99 @@
...
+
+  /// Analysis was completed without issues.
+  static const success = 'success';
+}
+
+/// Fields that contain the properties of the latest analysis.
+///
+/// For [PackageAnalysis] the rollup is collected only for the latest stable
+/// version.
+class AnalysisRollup {
```
mkustermann: Please make it an abstract class.

mkustermann: Personally I'm not a big fan of the name `rollup` at all :-/

Would highly prefer, something like `LatestAnalysisMixin`, `LatestAnalysisPropertiesMixin`.

app/lib/analyzer/models.dart
```diff
@@ -0,0 +1,99 @@
...
+
+  db.Key get analyzedVersionKey => analysisKey.parent;
+  Version get semanticAnalyzedVersion =>
+      new Version.parse(analyzedVersionKey.id);
+
+  @db.StringProperty()
+  String analysisVersion;
+
+  @db.StringProperty()
+  String analysisStatus;
```
mkustermann: See comment further down about this.

app/lib/analyzer/models.dart
```diff
@@ -0,0 +1,99 @@
...
+class AnalysisRollup {
+  @db.DateTimeProperty()
+  DateTime analysisTimestamp;
+
+  @db.ModelKeyProperty()
+  db.Key analysisKey;
+
+  db.Key get analyzedVersionKey => analysisKey.parent;
+  Version get semanticAnalyzedVersion =>
+      new Version.parse(analyzedVersionKey.id);
```
mkustermann: Could we get rid of `analyzedVersionKey` here? This is a mixin applied to different model classes. It should be model-class specific how the key space looks like.

For example:
```dart
PackageVersionAnalysis analysis;
analysis.analyzedVersionKey; // <-- this getter is completely unnecessary
                             // because it's the same as `analysis.key`.
```

app/lib/analyzer/backend.dart
```diff
@@ -0,0 +1,160 @@
...
+  AnalysisBackend(this.db);
+
+  /// Gets the [Analysis] object from the Datastore. [version] and [analysis] is
+  /// optional, when they are missing the analysis from the latest rollup will
+  /// be returned.
+  Future<Analysis> getAnalysis(String package,
+      [String version, String analysis]) async {
+    final Key packageKey = db.emptyKey.append(PackageAnalysis, id: package);
+
+    Future<Analysis> loadRollup(AnalysisRollup rollup) async {
```
mkustermann: It's not loading any "rollup", it's loading the latest analysis result (feeling very uncomfortable with "rollup" :-/ )

app/lib/analyzer/models.dart
```diff
@@ -0,0 +1,99 @@
...
+/// Fields that contain the properties of the latest analysis.
+///
+/// For [PackageAnalysis] the rollup is collected only for the latest stable
+/// version.
+class AnalysisRollup {
+  @db.DateTimeProperty()
+  DateTime analysisTimestamp;
+
+  @db.ModelKeyProperty()
+  db.Key analysisKey;
```
mkustermann: `analysisKey` => `latestAnalysisKey`

app/lib/analyzer/models.dart
```diff
@@ -0,0 +1,99 @@
...
+  String analysisStatus;
+}
+
+/// Analyzed package report.
+@db.Kind(name: 'PackageAnalysis', idType: db.IdType.String)
+class PackageAnalysis extends db.ExpandoModel with AnalysisRollup {
+  @db.StringProperty(required: true)
+  String package; // Same as id
+
+  // See also [AnalysisRollup] fields.
```
mkustermann: We could consider letting `PackageAnalysis` point to the most recent `PackageVersionAnalysis` instead of having the same fields here duplicated several times.

The additional latency due to multiple lookups is not that huge and can (as always) be mitigated by memcache in the future.

app/lib/analyzer/backend.dart
```diff
@@ -0,0 +1,160 @@
...
+    // analysis was set
+    final Key analysisKey = versionKey.append(Analysis, id: analysis);
+    final List list = await db.lookup([analysisKey]);
+    return list[0];
+  }
+
+  /// Create a new [Analysis] object with the current timestamp.
+  Analysis prepareAnalysis(String package, String version) {
+    final DateTime now = new DateTime.now().toUtc();
+    final String id = _generateAnalysisId(now);
```
mkustermann: We don't utilize that the id contains a mangled datetimestamp, so we could let datastore auto-generate an id for us.

app/lib/analyzer/backend.dart
```diff
@@ -0,0 +1,160 @@
...
+    final Key packageKey = db.emptyKey.append(PackageAnalysis, id: package);
+    final Key versionKey =
+        packageKey.append(PackageVersionAnalysis, id: version);
+    return new Analysis()
+      ..parentKey = versionKey
+      ..id = id
+      ..packageKey = packageKey
+      ..versionKey = versionKey
```
mkustermann: These two "back pointing" keys are not necessary at all, since the `Analysis.key` contains all the necessary information.

We might still want to store `Analysis.packageName` and `Analysis.versionName` as Strings though, for ease of querying `Analysis` objects and because it makes semantically.

app/lib/analyzer/backend.dart
```diff
@@ -0,0 +1,160 @@
...
+    final version = new PackageVersionAnalysis()
+      ..parentKey = analysis.packageKey
+      ..id = analysis.version
+      ..packageKey = analysis.packageKey
+      ..version = analysis.version;
+    _updateRollup(version, analysis);
+    return version;
+  }
+
+  bool _updateRollup(AnalysisRollup rollup, Analysis analysis) {
```
mkustermann: -> `updateLatestAnalysis`

isoos:
PTAL, almost everything changed :)

Important note: I've switched to int id, but I couldn't figure out how to get the appengine-assigned id, I did a fallback to use millisecondsSinceEpoch instead. I think collision here is minimal, and even if it happens, we won't end up in a bad state.

mkustermann:
@isoos It is super easy:

```dart
final model = new Analysis();
assert(mode.id == null);
await dbService.commit(inserts: [model]);
assert(mode.id is int);
```

Will take another look at the PR later on.

app/lib/analyzer/model_properties.dart
```diff
@@ -0,0 +1,49 @@
...
+
+  @override
+  bool validate(ModelDB db, Object value) =>
+      (!required || value != null) &&
+      (value == null || value is AnalysisStatus);
+
+  @override
+  String encodeValue(ModelDB db, Object value, {bool forComparison: false}) {
+    if (value is AnalysisStatus) {
+      return value.toString().split('.').last;
```
mkustermann: If you do it automatically, please add a warning comment above at `AnalysisStatus`, something like:

"Warning: Do not rename these enum values, since they are stored in the DB".

app/lib/analyzer/model_properties.dart
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @override
+  Object decodePrimitiveValue(ModelDB db, Object value) {
+    if (value is String) {
+      return AnalysisStatus.values.firstWhere(
+          (as) => as.toString().endsWith('.$value'),
+          orElse: () => null);
+    }
+    return null;
```
mkustermann: It's fine for this method to throw if you get anything which is neither `null` nor `String`.

app/lib/analyzer/model_properties.dart
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @override
+  Object decodePrimitiveValue(ModelDB db, Object value) {
+    if (value is String) {
+      return AnalysisStatus.values.firstWhere(
+          (as) => as.toString().endsWith('.$value'),
+          orElse: () => null);
```
mkustermann: Could issue an error here (or at least `logger.error`) because it means the DB contains values we don't understand. After a read-from-db, write-to-db the old value is simply gone. (i.e. instead of keeping the old value, we'll write a `null` instead).

app/lib/analyzer/models.dart
```diff
@@ -0,0 +1,78 @@
...
+
+import 'model_properties.dart';
+
+/// Analyzed package report.
+@db.Kind(name: 'PackageAnalysis', idType: db.IdType.String)
+class PackageAnalysis extends db.ExpandoModel {
+  String get packageName => id;
+
+  @db.ModelKeyProperty(required: true)
+  db.Key latestAnalysisKey;
```
mkustermann: Why not make this `latestVersionAnalysisKey` ?

Alternatively we have to also remember what `analysisVersion` (aka package:pana version) this one has (similar to `PackageVersionAnalysis`), because when writing a new `Analysis` we need to know whether to update `PackageAnalysis` and/or `PackageVersionAnalysis` or not.

app/lib/analyzer/backend.dart
```diff
@@ -0,0 +1,152 @@
...
+    _updatePackageLatest(package, analysis);
+    return package;
+  }
+
+  bool _updatePackageLatest(PackageAnalysis package, Analysis analysis) {
+    if (isNewer(analysis.semanticVersion, package.latestSemanticVersion)) {
+      // analysis version is older than latest known stable version
+      return false;
+    }
+    if (package.latestAnalysis > analysis.analysis) {
```
mkustermann: That's actually not something to rely on. The numbers are not always guaranteed to be sequentially increasing.

But we also don't need that I think:

If we just created a new `Analysis` for a given `package` / `version` we simply update it's `PackageVersionAnalysis`. The only case when we don't want that is if the analysis available on the `PackageVersionAnalysis` had a newer `Analysis.analysisVersion` (e.g. was generated with a newer `package:pana` version, because we prefer newer over older reports).

app/lib/analyzer/backend.dart
```diff
@@ -0,0 +1,158 @@
...
+        .append(PackageVersionAnalysis, id: version);
+    return new Analysis()
+      ..parentKey = versionKey
+      ..packageName = package
+      ..packageVersion = version
+      ..timestamp = now;
+  }
+
+  /// Stores the analysis, and either creates or updates its parent
+  /// [PackageAnalysis] and [PackageVersionAnalysis] rollups records.
```
mkustermann: "rollups" :-/

app/lib/analyzer/backend.dart
```diff
@@ -0,0 +1,158 @@
...
+          id: analysis.packageVersion);
+      final List parents = await tx.lookup([packageKey, packageVersionKey]);
+      final PackageAnalysis package = parents[0];
+      final PackageVersionAnalysis version = parents[1];
+
+      final List<Model> inserts = [];
+      if (version == null) {
+        inserts.add(_createVersion(analysis));
+      } else {
+        if (_updateVersionLatest(version, analysis)) {
```
mkustermann: Maybe in another CL: In the normal pub code you added methods to the models for this IIRC, maybe we should do that here as well.

app/lib/analyzer/backend.dart
```diff
@@ -0,0 +1,158 @@
...
+      ..parentKey = db.emptyKey
+      ..id = analysis.packageName
+      ..latestVersionAnalysisKey = analysis.key;
+    return package;
+  }
+
+  bool _updatePackageLatest(PackageAnalysis package, Analysis analysis) {
+    if (isNewer(
+        analysis.semanticPackageVersion, package.latestSemanticVersion)) {
+      // current package version is older than latest known stable version
```
mkustermann: This comment is not correct, since this is really comparing semantic versions (published package versions) not the `analysisVersion` (aka pana version).

app/lib/analyzer/backend.dart
```diff
@@ -0,0 +1,158 @@
...
+      // [PackageVersionAnalysis] to always store the latest.
+      if (inserts.isEmpty) {
+        await tx.rollback();
+        return;
+      }
+
+      if (package == null) {
+        inserts.add(_createPackage(analysis));
+      } else {
+        if (_updatePackageLatest(package, analysis)) {
```
mkustermann: This is updating `PackageAnalysis` purely based on the new `Analysis` irrespective of the semver of the pub version (e.g. `PackageAnalysis.latestAnalysis` is for 1.0.0 with pana 0 and analysis is 0.1.0 with pana 1, so here we don't want to update `latestAnalysis` (at least that is what I thought)).

We can avoid this issue by letting `PackageAnalysis` always point to the latest stable semver `PackageVersionAnalysis` and let this one point to the latest pana version of `Analysis`.

This was I mentioned in the last 2 reviews, that the `Key` in `PackageAnalysis` points to `PackageVersionAnalysis` (and not to an `Analysis` directly).

app/lib/analyzer/models.dart
```diff
@@ -0,0 +1,82 @@
...
+
+import 'model_properties.dart';
+
+/// Analyzed package report.
+@db.Kind(name: 'PackageAnalysis', idType: db.IdType.String)
+class PackageAnalysis extends db.ExpandoModel {
+  String get packageName => id;
+
+  @db.ModelKeyProperty(required: true)
+  db.Key latestVersionAnalysisKey;
```
mkustermann: Similar concern as in the last 2 reviews (see comment above)

app/lib/analyzer/backend.dart
isoos: This is true if you look only at the lines here. However, we update the version in a few lines earlier, and if there is no update (== pana version is older), we abort the update, including the package update.

app/lib/analyzer/backend.dart
isoos: the comment is correct, because it is about the packageVersion comparison, and right, we don't compare pana versions here.

the packageVersion comparison is need here, because we might be analyzing `packageVersion=1.1.0-dev` with newer pana, but we want to store only the analysis of the latest stable version at the package level.

I can see that this and the above is causing confusion, I'll rewrite it, probably storing more info on the package level.

isoos:
PTAL. Now package points only to version, and version to analysis. Also moved the object creation and updates into model.

app/lib/analyzer/backend.dart
mkustermann: That is a fair point.

>  However, we update the version in a few lines earlier, and if there is no update (== pana version is older),
>  we abort the update, including the package update.

Though the pana version can be older and we still come here iff the the analysis is for a new version (but for an already existent package).

mkustermann:
> Now package points only to version,

I don't think it does, the current PR is still setting it to `analysis.key`:

```dart
   PackageAnalysis _createPackage(Analysis analysis) {
     final package = new PackageAnalysis()
       ..parentKey = db.emptyKey
       ..id = analysis.packageName
       ..latestVersionAnalysisKey = analysis.key;
     return package;
   }
```

The github reviewing makes this quite difficult, since I cannot see the changes you uploaded since the last review. (there would be an option to use rietveld)

Have you uploaded your latest changes or is there something missing?

isoos:
> Have you uploaded your latest changes or is there something missing?

My bad, typed in `git push ...`, hit enter, but didn't check if it was running all right. Turns out it sometimes fails, fixed it.

app/lib/analyzer/backend.dart
```diff
@@ -0,0 +1,93 @@
...
+  Future<Analysis> getAnalysis(String package,
+      [String version, int analysis]) async {
+    final Key packageKey = db.emptyKey.append(PackageAnalysis, id: package);
+
+    if (version == null) {
+      final list = await db.lookup([packageKey]);
+      final PackageAnalysis pa = list[0];
+      if (pa == null) return null;
+      version = pa.latestVersion;
+      analysis = pa.latestAnalysis;
```
mkustermann: There's no `PackageAnalysis.latestAnalysis` now I think. Just remove this line.

(Since the travis bot didn't catch it and the variable is typed, I assume we're not running the analyzer on travis. We should keep that in mind and enable it at some point.)

isoos: Good catch! I think `dartanalyzer` didn't raise an alarm because the model's superclass `ExpandoModel` implements `noSuchMethod`.

mkustermann: That makes sense, yes. Though there is a good reason to inherit from `ExpandoModel`: it helps with forward/backwards compatibility of schema changes.

PR #179, opened 2017-06-22
Analyzer isolate, scheduler and trigger.

isoos:
- polls datastore every 10 minutes (if otherwise idle)
- the current batching makes sure that a carefully timed trigger-attack won't starve the datastore polling

### DIFF
--- a/app/lib/shared/task_scheduler.dart
+++ b/app/lib/shared/task_scheduler.dart
@@ -1,0 +1,116 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:async/async.dart';
+import 'package:logging/logging.dart';
+
+final Logger _logger = new Logger('pub.scheduler');
+
+/// Interface for task execution.
+typedef Future TaskRunner(Task task);
+
+// ignore: one_member_abstracts
+abstract class TaskSource {
+  /// Returns a stream of currently available tasks at the time of the call.
+  Stream<Task> currentTasks();
+}
+
+/// Tasks coming from through the isolate's receivePort, originating from a
+/// HTTP handler that received a ping after a new upload.
+class ManualTriggerTaskSource implements TaskSource {
+  final Stream taskReceivePort;
+  final int capacity;
+  StreamSubscription _subscription;
+  final Set<Task> _triggered = new Set();
+
+  ManualTriggerTaskSource(this.taskReceivePort, {this.capacity: 100}) {
+    _subscription = taskReceivePort.listen((Task task) {
+      // protect against spamming the manual trigger
+      if (_triggered.length < capacity) {
+        _triggered.add(task);
+      }
+    });
+  }
+
+  @override
+  Stream<Task> currentTasks() {
+    final List<Task> tasks = _triggered.toList();
+    _triggered.clear();
+    return new Stream.fromIterable(tasks);
+  }
+
+  void close() {
+    // It is unlikely that we will close this, but cancelling the subscription
+    // makes dartanalyzer happy.
+    _subscription.cancel();
+  }
+}
+
+/// Task source that has a limit on the polling frequency.
+abstract class PollingTaskSource implements TaskSource {
+  final Duration minimumPeriod;
+  DateTime _lastPollTime;
+
+  PollingTaskSource(this.minimumPeriod);
+
+  @override
+  Stream<Task> currentTasks() {
+    final DateTime now = new DateTime.now();
+    if (_lastPollTime == null ||
+        now.difference(_lastPollTime) > minimumPeriod) {
+      _lastPollTime = now;
+      return pollTasks();
+    }
+    return new Stream.fromIterable([]);
+  }
+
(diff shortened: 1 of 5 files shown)


app/lib/analyzer/scheduler.dart
```diff
@@ -0,0 +1,101 @@
...
+    for (;;) {
+      if (_tasks.isEmpty) {
+        await _populateTasks();
+        // short nap, just in case
+        await new Future.delayed(new Duration(seconds: 2));
+      } else {
+        final ScheduledTask task = _tasks.removeLast();
+        await _runAnalyze(task);
+        // long nap, to give GC ample time
+        await new Future.delayed(new Duration(seconds: 15));
```
mkustermann: Actually I'm not sure if our GC is collecting during idle times. Usually the collections are triggered by new/old-space allocation failures (i.e. if current capacity is exhausted). But I don't mind if we are waiting for a few seconds :)

app/lib/analyzer/scheduler.dart
```diff
@@ -0,0 +1,101 @@
...
+      newTasks.addAll(await _pollDatastoreForTasks());
+      _lastDatastorePoll = new DateTime.now();
+    }
+
+    _tasks.addAll(newTasks);
+    _tasks.shuffle();
+  }
+
+  Future<bool> _sanityCheck(ScheduledTask task) async {
+    // TODO: check if the package and version exists
```
mkustermann: Even if we return `true` here, if we manually delete a package, it can still be the case that it doesn't exist after calling this method. Maybe it would be better handling the non-exist case when we actually try to analyze it.

app/lib/analyzer/scheduler.dart
```diff
@@ -0,0 +1,101 @@
...
+
+  Future<List<ScheduledTask>> _pollDatastoreForTasks() async {
+    // TODO: get packages from datastore to analyze
+    return [];
+  }
+
+  Future _runAnalyze(ScheduledTask task) async {
+    // TODO: implement analysis
+  }
+}
```
mkustermann: Since the search indexing might do the same thing: monitor new packages and index them (as well as going back in history), you could generalize this a bit.

For example something like this:

```dart

Future analysisIsolate() {
  await for (final Task task in getTasks()) {
    try {
       final analysis = await analyzePackage(task);
       await writeAnalysisResult(analysis);
     } catch (e, s) { ... }
  }
}

Future searchIndexingIsolate() {
  await for (final Task task in getTasks()) {
    try {
       final result = await indexPackage(task);
       index.update(result);
       await index.saveTo...();
     } catch (e, s) { ... }
  }
}

Stream<Task> getTasks() {
  // there is probably a `combineStreams` somewhere which does this.
  return combineStreams([
    new ManualTriggerSource().streamTasks(),
    new DatastoreHistorySource().streamTasks(),
    new DatastoreHeadPollerSource().streamTasks(),
  ]);
}

class TaskSource {
  Stream<Task> streamTasks();
}

class ManualTriggerSource extends TaskSource {
  ReceivePort port;

  Stream<Task> streamTasks() async* {
    await for (final message in port) {
      yield new Task(....);
    }
  }
}

class DatastoreHistorySource  extends TaskSource {
  Stream<Task> streamTasks() async* {
    await for (final candidate in dbService.query(PackageVersion, -date)) {
       yield new Task(candidate.package, candidate.version);
    }
  }
}

class DatastoreHeadPollerSource extends TaskSource {
  Stream<Task> streamTasks() async* {
    final DateTime begin = (await dbService.query(PackageVersion, -date, limit: 1)).first;
    while (true) {
      await new Future.delayed(const Duration(minutes: 1)); // polll interval.
      await for (final candidate in dbService.query(PackageVersion, -date)) {
        if (candidate.timeStamp > begin) yield candidate;
        else break;
      }
      yield new Task();
    }
  }
}

class Task { package, version, receivedTimestamp }
```

isoos:
Thanks for the suggestions, I've applied them with a few changes.

app/lib/shared/task_scheduler.dart
```diff
@@ -0,0 +1,116 @@
...
+  final int capacity;
+  StreamSubscription _subscription;
+  final Set<Task> _triggered = new Set();
+
+  ManualTriggerTaskSource(this.taskReceivePort, {this.capacity: 100}) {
+    _subscription = taskReceivePort.listen((Task task) {
+      // protect against spamming the manual trigger
+      if (_triggered.length < capacity) {
+        _triggered.add(task);
+      }
```
mkustermann: Consider logging if we ever hit more than 100 manual tasks, since we never expect it.

You can do it in an upcoming CL.

mkustermann:
I think the travis build had an exception. Do you know why?

Otherwise: LGTM

isoos:
> I think the travis build had an exception. Do you know why?

````
 Expected: 'Exception: Package name must begin with a letter or underscore.'
    Actual: 'Failed to make tarball of test package.'
````

Looks like the `tar` command in `withTestPackage` (`backend_test_utils.dart`) failed, I don't see any immediate reason for that.

mkustermann:
> Would you like me to rewrite it that way?

No, it's fine. Was just curious. We can change it if need be.

mkustermann:
> I don't see any immediate reason for that.

Then it's unrelated to this CL.

PR #180, opened 2017-06-22
Serve analysis data through HTTP.

isoos:


isoos:
rebased

kevmoo:
No tests?

isoos:
Not yet, several pieces missing for it. Trivial test could have been done
but not worth it yet.

isoos:
Added handler tests, PTAL.

app/test/analyzer/handlers_test.dart
```diff
@@ -0,0 +1,178 @@
...
+import '../shared/handlers_test_utils.dart';
+import '../shared/utils.dart';
+
+import 'handlers_test_utils.dart';
+
+void main() {
+  group('handlers', () {
+    group('not found', () {
+      scopedTest('/xxx', () async {
+        expectNotFoundResponse(await issueGet('/xxx'));
```
mkustermann: I think you need to either return the future from `expectNotFoundResponse` or `await` it, otherwise `package:test` will think the test was done without errors and mark it as `PASS` even if it fails later.

(also in the other tests)

app/test/analyzer/handlers_test.dart
```diff
@@ -0,0 +1,178 @@
...
+    group('trigger analysis', () {
+      scopedTest('/packages/pkg', () async {
+        final ReceivePort rp = new ReceivePort();
+        final Future<Task> taskFuture = rp.first;
+        registerTaskSendPort(rp.sendPort);
+        expectJsonResponse(await issuePost('/packages/pkg'),
+            body: {'status': 'OK'});
+        final Task task = await taskFuture;
+        expect(task.package, 'pkg');
+        expect(task.version, isNull);
```
mkustermann: Is the receive port closed automatically by `rp.first`? If not, you might need to close it (also below)

isoos: It closes it (ReceivePort is a non-broadcast stream), but better to be explicit about it.

app/test/analyzer/handlers_test.dart
isoos: Good point, calling `await` on it. I'm not sure why this was working, maybe because it was a scoped test and it is running in a zone? Btw. the frontend's handler tests are similar.

PR #191, opened 2017-06-26
Run pana analyzer.

isoos:
One behavioral change is that trigger needs to have both a package and a version. We know these at upload time, there is no need to handle the case where we need to lookup the latest stable version for the package.


app/lib/analyzer/backend.dart
```diff
@@ -92,4 +99,47 @@ class AnalysisBackend {
...
+          .append(Package, id: packageName)
+          .append(PackageVersion, id: packageVersion)
+    ]);
+    // package or version does not exists
+    if (versions.isEmpty || versions.first == null) return false;
+
+    final List<PackageAnalysis> list =
+        await db.lookup([db.emptyKey.append(PackageAnalysis, id: packageName)]);
+    // no analysis for the package
+    if (list.isEmpty || list.first == null) return true;
```
mkustermann: This is guaranteed to always return a list of length 1.

app/lib/analyzer/backend.dart
```diff
@@ -92,4 +99,47 @@ class AnalysisBackend {
...
+
+    final PackageAnalysis packageAnalysis = list.first;
+    final PackageVersionAnalysis versionAnalysis =
+        (await db.lookup([packageAnalysis.latestVersionKey])).first;
+
+    if (versionAnalysis.analysisVersion == analysisVersion) {
+      final DateTime now = new DateTime.now();
```
mkustermann: To make the reader not care about whether `DateTime.difference` automatically converts one argument to utc if the other is utc already, you could make `now` be utc.

mkustermann:
LGTM with minor comments

PR #193, opened 2017-06-27
Analyzer task sources: head and outdated.

isoos:
I am undecided whether to move the datastore-accessing parts to `analyzer/backend.dart`, but we can do that in a subsequent change too.

### DIFF
--- a/app/lib/analyzer/task_sources.dart
+++ b/app/lib/analyzer/task_sources.dart
@@ -4,24 +4,118 @@
 
 import 'dart:async';
 
+import 'package:async/async.dart';
+import 'package:gcloud/db.dart';
+
+import '../frontend/models.dart';
 import '../shared/task_scheduler.dart';
 
+import 'models.dart';
+
+/// Creates a task when a version uploaded in the past hour has no analysis yet.
 class DatastoreHeadTaskSource extends PollingTaskSource {
-  DatastoreHeadTaskSource() : super(new Duration(minutes: 10));
+  final DatastoreDB _db;
+  DateTime _lastTs;
+  DatastoreHeadTaskSource(this._db) : super(const Duration(minutes: 10));
 
   @override
   Stream<Task> pollTasks() {
-    // TODO: implement pollTasks
-    return new Stream.empty();
+    final DateTime now = new DateTime.now().toUtc();
+    final DateTime oneHourAgo = now.subtract(new Duration(hours: 1));
+    DateTime minCreated;
+    if (_lastTs == null) {
+      minCreated = oneHourAgo;
+    } else if (_lastTs.isBefore(oneHourAgo)) {
+      // more than an hour passed since the last poll
+      minCreated = _lastTs;
+    } else {
+      minCreated = oneHourAgo;
+    }
+    _lastTs = oneHourAgo;
+    final Query q = _db.query(PackageVersion)..filter('created >=', minCreated);
+    return q.run().asyncMap(_mapPackageVersion).where((v) => v != null);
+  }
+
+  Future<Task> _mapPackageVersion(Model m) async {
+    final PackageVersion pv = m;
+    final List<PackageVersionAnalysis> items = await _db.lookup([
+      _db.emptyKey
+          .append(PackageAnalysis, id: pv.package)
+          .append(PackageVersionAnalysis, id: pv.version)
+    ]);
+    if (items.first == null) {
+      return null;
+    } else {
+      return new Task(pv.package, pv.version);
+    }
   }
 }
 
+/// Creates a task when the most recent analysis is older than [afterDays] days.
+///
+/// When [analysisVersion] is set, it also checks whether the current one is
+/// newer and creates a task if needed.
 class DatastoreHistoryTaskSource extends PollingTaskSource {
-  DatastoreHistoryTaskSource() : super(new Duration(hours: 1));
+  final DatastoreDB _db;
+  final int afterDays;
+  final String analysisVersion;
+  final int _batchSize = 10;
+  int _offset = 0;
+
+  DatastoreHistoryTaskSource(this._db,
+      {this.afterDays: 30, this.analysisVersion})
+      : super(const Duration(minutes: 10));
 
   @override
   Stream<Task> pollTasks() {
-    // TODO: implement pollTasks
-    return new Stream.empty();
+    final StreamCompleter<Task> resultCompleter = new StreamCompleter();
+    final Query q = _db.query(PackageVersion)
+      ..order('-created')
+      ..offset(_offset)
+      ..limit(_batchSize);
+    _offset += _batchSize;
+
+    q.run().asyncMap(_mapPackage).toList().then(
+      (List<Task> list) {
+        if (list.isEmpty) {
+          // Query results were empty, we have no further versions,
+          // restart history scanning.
+          _offset = 0;
+          resultCompleter.setEmpty();
+        } else {
+          resultCompleter.setSourceStream(
+              new Stream.fromIterable(list.where((t) => t != null)));
+        }
+      },
+      onError: (e, st) {
+        resultCompleter.setError(e, st);
+      },
+    );
+
+    return resultCompleter.stream;
+  }
+
+  Future<Task> _mapPackage(Model m) async {
+    final PackageVersion pv = m;
+    final List<PackageVersionAnalysis> list = await _db.lookup([
+      _db.emptyKey
+          .append(PackageAnalysis, id: pv.package)
+          .append(PackageVersionAnalysis, id: pv.version)
+    ]);
+    if (list.first == null) {
+      return new Task(pv.package, pv.version);
+    }
+
+    final PackageVersionAnalysis version = list.first;
+    final Duration diff =
+        new DateTime.now().toUtc().difference(version.analysisTimestamp);
+    final bool versionDiffers =
+        analysisVersion != null && version.analysisVersion != analysisVersion;
+
+    if (versionDiffers || diff.inDays >= afterDays) {
+      return new Task(version.packageName, version.packageVersion);
+    } else {
+      return null;
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


app/lib/analyzer/task_sources.dart
```diff
@@ -4,24 +4,79 @@
...
+import '../frontend/models.dart';
 import '../shared/task_scheduler.dart';
 
+import 'models.dart';
+
+/// Creates a task when a package has no analysis for its latest stable version.
 class DatastoreHeadTaskSource extends PollingTaskSource {
-  DatastoreHeadTaskSource() : super(new Duration(minutes: 10));
+  final DatastoreDB _db;
+  DatastoreHeadTaskSource(this._db) : super(new Duration(minutes: 10));
```
mkustermann: Could be `const Duration(...)`

app/lib/analyzer/task_sources.dart
```diff
@@ -4,24 +4,79 @@
...
 
   @override
   Stream<Task> pollTasks() {
-    // TODO: implement pollTasks
-    return new Stream.empty();
+    return _db
+        .query(Package)
```
mkustermann: A few comments about `DatastoreHeadTaskSource` and `DatastoreHistoryTaskSource`

  * it would be fine if we ensured to analyze all package versions, not just the latest package.
  * it would be nice to not do repeated queries for things we know we have already processed.

So the idea with the split between `DatastoreHeadTaskSource` and `DatastoreHistoryTaskSource` was:

  * the head poller is just querying the latest `PackageVersion` uploads since the last poll (not going back in  history infinitely)
  * the history poller is just going back in history *once* and then it's done (i.e. can return an empty stream)

This should be easily doable via a query on `PackageVersion` with `-date` sorting and a field remembering, which was the last item we processed already, thereby making both `getTasks()` calls a NOP (after initial history traversal).

app/lib/analyzer/task_sources.dart
```diff
@@ -4,24 +4,79 @@
...
+        .run()
+        .asyncMap(_mapPackage)
+        .where((v) => v != null);
+  }
+
+  Future<Task> _mapPackage(Model m) async {
+    final Package package = m;
+    final List<PackageAnalysis> items = await _db
+        .lookup([_db.emptyKey.append(PackageAnalysis, id: package.name)]);
+    if (items.isEmpty ||
```
mkustermann: This list is guaranteed to have length 1.

app/lib/analyzer/task_sources.dart
isoos: > the history poller is just going back in history once and then it's done

There is a conflicting requirement here: we want to re-analyze packages with their latest resolved dependencies, and that will require polling history again and again (at least until we have a better way of detecting the need of re-analysis).

app/lib/analyzer/task_sources.dart
```diff
@@ -4,24 +4,78 @@
...
+
+  Future<Task> _mapPackage(Model m) async {
+    final PackageAnalysis package = m;
+    final List<PackageVersionAnalysis> list =
+        await _db.lookup([package.latestVersionKey]);
+    if (list.first == null) return null;
+    final PackageVersionAnalysis version = list.first;
+
+    final Duration diff =
+        new DateTime.now().difference(version.analysisTimestamp);
```
mkustermann: Consider using utc instead.

app/lib/analyzer/task_sources.dart
mkustermann: Why not use something like this :
```dart

Key latestHandledKey;

Stream<Task> pollTasks() async* {
  final query = db.query(PackageAnalysis)..order('-date');
  await for (final PackageVersion pv in query.run()) {
    if (pv.key == latestHandledKey) break;
    if (!isUpToDate(pv)) {
      yield new Task(pv.package, pv.packageVersion);
    }
    latestHandledKey = pv.key;
  }
}
```
?

isoos:
PTAL:
- head is now polling for new versions in the last hour (+ some extra time window in case the polling frequency gets larger than an hour).
- history is now querying on versions, descending created time, with offset and limit controlled, but re-starting when it completes.

Both of the changes will now target every version, in contrast to the previous latest-stable-only.

isoos:
> Consider using utc instead.

Done. The rest of the comments you have submitted around the same time I've updated the sources became obsolete I think.

app/lib/analyzer/task_sources.dart
```diff
@@ -4,24 +4,118 @@
...
+    final DateTime oneHourAgo = now.subtract(new Duration(hours: 1));
+    DateTime minCreated;
+    if (_lastTs == null) {
+      minCreated = oneHourAgo;
+    } else if (_lastTs.isBefore(oneHourAgo)) {
+      // more than an hour passed since the last poll
+      minCreated = _lastTs;
+    } else {
+      minCreated = oneHourAgo;
+    }
```
mkustermann: May I ask why you want to have this 1 hour logic?

app/lib/analyzer/task_sources.dart
```diff
@@ -4,24 +4,118 @@
...
+  final DatastoreDB _db;
+  DateTime _lastTs;
+  DatastoreHeadTaskSource(this._db) : super(const Duration(minutes: 10));
 
   @override
   Stream<Task> pollTasks() {
-    // TODO: implement pollTasks
-    return new Stream.empty();
+    final DateTime now = new DateTime.now().toUtc();
+    final DateTime oneHourAgo = now.subtract(new Duration(hours: 1));
```
mkustermann: Another opportunity for `const Duration()` :)

app/lib/analyzer/task_sources.dart
```diff
@@ -4,24 +4,118 @@
...
+    final PackageVersion pv = m;
+    final List<PackageVersionAnalysis> items = await _db.lookup([
+      _db.emptyKey
+          .append(PackageAnalysis, id: pv.package)
+          .append(PackageVersionAnalysis, id: pv.version)
+    ]);
+    if (items.first == null) {
+      return null;
+    } else {
+      return new Task(pv.package, pv.version);
```
mkustermann: Shouldn't this be reversed? i.e. if we couldn't find an analysis, we should make a `Task` and otherwise skip it.

app/lib/analyzer/task_sources.dart
```diff
@@ -4,24 +4,118 @@
...
   @override
   Stream<Task> pollTasks() {
-    // TODO: implement pollTasks
-    return new Stream.empty();
+    final StreamCompleter<Task> resultCompleter = new StreamCompleter();
+    final Query q = _db.query(PackageVersion)
+      ..order('-created')
+      ..offset(_offset)
+      ..limit(_batchSize);
+    _offset += _batchSize;
```
mkustermann: Since there can be new `PackageVersions`s the offset/limit mechanism is not really good (also because it is very costly).

Unfortunately I haven't had the time to implement `package:gcloud` support for cursors yet.

Why do we need to limit this? Because it takes too long to call generate the list of tasks?

That's precisely why I would've prefered having infinite streams and a merging function, like
```dart
  final sources = [new HeadPollerSource(), new HistoryPollerSource(), new ManualSource()];
  await for (final Task task in combineStreams(sources)) {
    ...
  }
```

This way the `combineStreams` would listen on all incoming streams and interleave the tasks, giving each task source equal priority.

mkustermann: Alternatively I would do it based on the `created` timestamp, similarly to the head poller. So you can continue where you left off (effectively using the timestamp as the db query cursor).

app/lib/analyzer/task_sources.dart
isoos: T0: issue query
T0+1ms sec: new version is getting uploaded and eventually it reaches datastore
T0+100ms: query completes, but misses the newly uploaded version
T0+10 minutes: the new poll will not catch the newly uploaded version, unless we do address the potential latency above.

Now, 1 hour may be a bit too much, but I don't see too much harm with it, as we have 6 of such queries per hour.

app/lib/analyzer/task_sources.dart
isoos: good catch!

app/lib/analyzer/task_sources.dart
isoos: Offset missmatch is not really an issue, as the new version will cause us repeated check, but we won't miss a version with that. Bad performance however is an issue indeed.

I'll rewrite it with the `created` timestamp shortly, however that also will need the same sorting.

I think we want to have some kind of throttling. Even if we have a single stream that is provided only once, you don't want to have the history source to flood the analyzer with tasks (e.g. re-analyzing after 30 days), and potentially cause days of backlog, while newly uploaded versions didn't get any attention. AFAICT in that regard the single stream wouldn't give much benefit for throttling, without the same limits I did here.

isoos:
I've implemented https://github.com/dart-lang/pub-dartlang-dart/pull/198 as an alternative take on this. Please review that before doing this.

app/lib/analyzer/task_sources.dart
mkustermann: The throttling could be nicely implemented with a customized version of `StreamGroup.merge` which accepts priorities and handles only one item at a time. Basically whenever a new `Task` is available from the manual trigger, we pick that one, and only if none is there, we use head poller, and if none is there, we'll use the history.

So the complexity can be nicely abstracted with a `combineStreams` function.

app/lib/analyzer/task_sources.dart
mkustermann: Not quite sure I understand your description.

Every single newly uploaded `PackageVersion` will have a `created` timestamp. This gives us a total utc timestamp order. The probability of two newly-uploaded packages having the very same timestamp is very unlikely. So we just remember the very last `PackageVersion.created` we analyzed and continue the next query from there.

The only two possible issues are: two `PackageVersion`s have the exact same `created` timestamp or there is a clock difference where one VM's clock is ahead of another one and it basically creates a `PackageVersion` with an wrong/too-old `created` timestamp.

Anyways, I'm fine with the 1 hour approach, if you prefer it.

mkustermann:
I've pulled now #198  and left a few comments which could be addressed in a future CL.

Closing this PR.

PR #196, opened 2017-06-27
Search service to use TaskScheduler, plus backend and handler tests.

isoos:


app/lib/search/handlers.dart
```diff
@@ -49,3 +52,17 @@ Future<shelf.Response> searchHandler(shelf.Request request) async {
       new PackageQuery.fromServiceQueryParameters(request.url.queryParameters));
   return jsonResponse(result.toJson(), indent: indent);
 }
+
+/// Handles requests for:
+///   - /packages/<package>
+Future<shelf.Response> packageHandler(shelf.Request request) async {
+  final String path = request.requestedUri.path.substring('/packages/'.length);
+  final String packageName = path;
+  final String requestMethod = request.method?.toUpperCase();
```
mkustermann: There should always be a `request.method`, it's not possible to receive an HTTP request without a method.

app/lib/search/updater.dart
```diff
@@ -4,92 +4,65 @@
...
-  Timer _timer;
-  bool _updating = false;
+Logger _logger = new Logger('pub.search.updater');
 
-  PackageIndexUpdater(this.index);
+class IndexUpdateTaskSource extends PollingTaskSource {
+  final bool isLocalDevMode;
+  bool _isFirst = true;
+  IndexUpdateTaskSource({this.isLocalDevMode: false})
+      : super(new Duration(hours: 1));
```
mkustermann: Could be `const Duration`

app/lib/search/updater.dart
```diff
@@ -4,92 +4,65 @@
...
-      _updating = true;
-      await _doUpdate(limit);
-    } catch (e, st) {
-      _logger.severe('Error while updating search index.', e, st);
-    } finally {
-      _updating = false;
+  @override
+  Stream<Task> pollTasks() {
+    final DateTime updatedAfter =
+        _isFirst ? null : new DateTime.now().subtract(new Duration(days: 1));
```
mkustermann: use utc since `searchBackend.listPackages` might compare against db timestamps

app/lib/search/updater.dart
```diff
@@ -4,92 +4,65 @@
...
+}
 
-  void startPolling() {
-    if (_timer == null) {
-      _timer = new Timer.periodic(new Duration(hours: 4), (_) {
-        update();
-      });
-      update();
-    }
+Future mergeIndex(int taskCount) async {
```
mkustermann: Consider adding a comment here: What is being merged? What does `taskCount` mean? Why is it not a boolean?

app/lib/shared/task_scheduler.dart
```diff
@@ -73,22 +76,42 @@ abstract class PollingTaskSource implements TaskSource {
...
+        if (afterTaskNapDuration != null) {
+          await new Future.delayed(afterTaskNapDuration);
+        }
+      }
+      if (loopEndCallback != null) {
+        await loopEndCallback(count);
+      }
+      // taking a nap, GC may kick in?
+      if (afterLoopNapDuration != null) {
+        await new Future.delayed(afterLoopNapDuration);
```
mkustermann: Strictly speaking I think sleeping because of GC is overkill. Not even sure if our GC uses idle time for GCing atm. Normally GCs are driven by (new/old space) allocation failures, storebuffer overflows, ... which only happen when mutator executes.

All isolates are independently GCed, so if the analysis/search tasks run in different isolates, there will be no harm to the main isolate.

The only impact is CPU: If the total CPU usage of the analysis/search task (including subprocesses, ...) is taking up all the cores, then the main http serving latency will suffer. But we should ensure our services get enough CPUs to prevent this.

Maybe just get rid of these naps entirely.

app/lib/search/backend.dart
```diff
@@ -0,0 +1,71 @@
...
+      ));
+    }
+    return documents;
+  }
+
+  Stream<String> listPackages({DateTime updatedAfter}) {
+    return _db
+        .query(Package)
+        .run()
+        .map((Model m) => m as Package)
```
mkustermann: Does the analyzer complain, or why did you add this map? Don't think it should be necessary.

app/lib/search/backend.dart
```diff
@@ -0,0 +1,71 @@
...
+        readme: pv.readmeContent,
+        popularity: mockScores[pv.package] ?? 0.0,
+      ));
+    }
+    return documents;
+  }
+
+  Stream<String> listPackages({DateTime updatedAfter}) {
+    return _db
+        .query(Package)
```
mkustermann: You should be able to use filters:
```dart
Stream<String> listPackages({DateTime updatedAfter}) {
  final query = db.query(Package);
  if (updatedAfter != null) query.filter('>= updated', updatedAfter);
  for (final Package package in query.run()) {
    yield package.name;
  }
}
```

So the datastore can do the filtering instead of querying everything and doing filtering in Dart.

app/lib/search/updater.dart
```diff
@@ -4,92 +4,65 @@
...
+        _isFirst ? null : new DateTime.now().subtract(new Duration(days: 1));
+    _isFirst = false;
+    Stream<Task> stream = searchBackend
+        .listPackages(updatedAfter: updatedAfter)
+        .map((String package) => new Task(package, null));
+    if (isLocalDevMode) {
+      // limit memory use and faster warm-up time
+      stream = stream.take(300);
     }
+    return stream;
```
mkustermann: Why should we re-index all packages updated in the last day every hour (so on average we're doing 24 re-indexing of the same tarball redundantly) ?

Similar to the analysis, we should remember, e.g. based on timestamp, which ones we already  handled, so we only re-index new packages.

All of this complication is due to the fact that `pollTasks()` is called many times, instead of a single `startStreaming()` :-/

isoos:
PTAL. I've updated this PR with the latest TaskScheduler. I've got a few tries on how to make the first listing of packages be complete and then trigger the first merge, eventually settled with this counter + fake Task approach. Not proud of it, but the other versions were even more crappier.

app/lib/search/updater.dart
```diff
@@ -4,92 +4,107 @@
...
+  Stream<Task> startStreaming() async* {
+    for (;;) {
+      final DateTime now = new DateTime.now().toUtc();
+      int counter = 0;
+      if (isLocalDevMode && counter > 300) break;
+      await for (String package
+          in searchBackend.listPackages(updatedAfter: _lastTs)) {
+        counter++;
+        yield new Task(package, null);
+        if (isLocalDevMode && counter > 300) break;
```
mkustermann: Not sure why this special `isLocalDevMode` logic is here, it's quite disturbing.

The user of all the task sources can limit it.

You could even make a nice wrapper, which either stops after a limit, like this:

```dart
class LocalDevModeTaskSource implements TaskSource {
  final TaskSource inner;

  LocalDevModeTaskSource(this.inner);

    int emitted = 0;
    await for (final Task task in inner.startStreaming()) {
     yield task;
     if (emitted++ > 300) break;
   }
  }
}
```

or throttling by doing sleeps in-between.

app/lib/search/updater.dart
```diff
@@ -4,92 +4,107 @@
...
+      }
+      if (_firstPackageListCount == null) {
+        _firstPackageListCount = counter;
+        // A bogus task that will be cleared easily and do no harm, and it makes
+        // sure that the first index merge will get called eventually.
+        yield new Task('___placeholder_for_index_trigger___', '');
```
mkustermann: Why is this needed?

Somehow this comment doesn't quite make sense to me. I think the decision when to merge should not be done by this class.

app/test/search/handlers_test_utils.dart
```diff
@@ -0,0 +1,21 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:shelf/shelf.dart' as shelf;
+
+import 'package:pub_dartlang_org/search/handlers.dart';
+
+Future<shelf.Response> issueGet(String path) async {
+  final uri = 'https://pub.dartlang.org$path';
```
mkustermann: Shouldn't the URL be `search-dot-dartlang.org` (not that it matters for the test, but a reader might get confused)

app/lib/search/updater.dart
```diff
@@ -4,92 +4,107 @@
...
+    try {
+      final List<Task> tasks = new List.from(_batch);
+      _batch.clear();
+      _logger.info(
+          'Updating index with ${tasks.length} packages [example: ${tasks.first
+              .package}]');
+      final List<PackageDocument> docs = await searchBackend
+          .loadDocuments(tasks.map((t) => t.package).toList());
+      await packageIndex.addAll(docs);
+      final bool doMerge = _firstPackageListCount != null &&
```
mkustermann: If you need shared state, instead of using a global variable, you could consider just letting one class know the other:
```dart
class BatchIndexUpdater {
  final IndexUpdateTaskSource source;

  Future _updateBatch() async {
    ...
    if (source.firstScanComplete) {
      ...
    }
  }
}

```

app/lib/search/updater.dart
```diff
@@ -4,92 +4,107 @@
...
+    }
   }
 
-  void startPolling() {
-    if (_timer == null) {
-      _timer = new Timer.periodic(new Duration(hours: 4), (_) {
-        update();
-      });
-      update();
+  Future _currentBatchRunning() async {
```
mkustermann: You could simplify the code by getting rid of this function (see comment above).

app/lib/search/updater.dart
```diff
@@ -4,92 +4,107 @@
...
-    }
+  Future _updateBatch() async {
+    _batchUpdateTimer?.cancel();
+    _batchUpdateTimer = null;
 
-    _logger.info('Calling index.merge() ...');
-    await index.merge();
-    _logger.info('index.merge() completed.');
+    await _currentBatchRunning();
+    if (_batch.isEmpty) return;
```
mkustermann: There is almost a race here (when the timer and `updateTask` will call this function, and both of them get yielded via the `await`). It's quite difficult to see whether it can happen or not, so I think we should simplify it, so it becomes crystal clear there is no race.

It's easier to read if you bundle **checking** and **setting** of `_ongoingBatchUpdate`  (and `_batch`)  together.

Like in
```dart

  while (_ongoingBatchUpdateFuture != null) {
    await _ongoingBatchUpdateFuture;
  }
  final Completer completer = new Completer();
  _ongoingBatchUpdateFuture = completer.future;


  final List<Task> batch = _batch.toList(growable: false);
  _batch.clear();

```

app/lib/search/updater.dart
isoos: I was using this limit to make the fine-tuning of the indexing and search snappier. Re-indexing the entire staging each time I made a small change was taking long, and this was a reasonable limit. There are other ways for that, so I'll just remove this part of the logic.

app/lib/search/updater.dart
isoos: The complexity of this synchronization comes from the latencies between (1) task gets created -> batched tasks get added to the index (2) first scan completes -> marking the search index ready.

Merge was a convenient way of doing the 'ready' mark with the previous approach, even with the round-based scans, but clearly needs a different approach here. I'll try to simplify the entire process.

isoos:
PTAL. Couple of major changes:
- fixed the `loadDocuments()` index issue
- removed the `isLocalDevMode` flag
- removed the global state, but counting seems to be the simplest synchronization mechanism for marking the end of the first scan, so using it for that purpose. removed the extra dummy task, that was unnecessary.

There are slight differences with the task source here and in the analyzer, once we have both working I'll merge their code, but I'd rather keep them separate for the time being.

app/lib/search/updater.dart
```diff
@@ -4,92 +4,100 @@
...
+  Stream<Task> startStreaming() async* {
+    for (;;) {
+      final DateTime now = new DateTime.now().toUtc();
+      int count = 0;
+      await for (String package
+          in searchBackend.listPackages(updatedAfter: _lastTs)) {
+        count++;
+        yield new Task(package, null);
+      }
+      _batchIndexUpdater.reportScanCount(count);
```
mkustermann: Still I find this quite a hack in many ways. It could be removed and just let the indexer re-build the index whenever either one is true:
   * got X new tasks
   * got 1 new task and Y minutes passed since then

But I'll merge it now anyways.

isoos: I agree, that would be simpler, but then: how does the search service know that it completed the indexing of all the packages at least once (not counting the last whatever minutes of changes if they may happen)?

In other words: when spinning up a new search service instance, how do we know when we can migrate traffic to it?

mkustermann: Ideally we would not build up a new index on each deployment of a new service. We would always have an already built index on cloud storage, which a newly created VM can just load. It can add more newly uploaded packages to that index.

It will periodically write out the newest index to cloud storage, so when the VM gets killed or a new instance gets deployed it can just load it from there.

Also the 'search' service is not making the decision when to migrate traffic. We manually select when it's time to change the default version of a service (for any service, including "search")

mkustermann: So basically all the search service has to do is: 
  * (Optionally) Load a base index from cloud storage when starting
  * Ensure all packages are in the index, if not, add them (going back in history)
  * Ensure newly uploaded packages will be included eventually

isoos:
Agreed most of it, but there is a constraint there at the moment: the ducene on-disk-index is sync only (with the currently deployed package version), and unusable for our purposes. Until a new version is created, we are going to build the index each time we create a new service.

mkustermann:
Actually not just each time we create a new service (or rather deploy a new version to the "search" service). AppEngine also sometimes just kills VMs and restarts them. So basically on every VM start it will re-index all packages.

It's fine for now if it rebuilds it every time. But once we switch over from custom search api, then we should have the index backed up and ready for serving on VM restarts.

PR #197, opened 2017-06-28
Notify analyzer on new package upload.

isoos:


app/lib/shared/utils.dart
```diff
@@ -10,9 +10,13 @@ import 'dart:io';
 import 'dart:math';
 
 import 'package:logging/logging.dart';
+import 'package:http/http.dart';
```
mkustermann: Please prefix usages of `package:http` with `http`, it makes it so much easier to read, because `Client` could be anything, whereas `http.Client` read much more clearly.

app/lib/frontend/backend.dart
```diff
@@ -329,6 +330,14 @@ class GCloudPackageRepository extends PackageRepository {
...
       }
     });
+
+    storeFuture.then((PackageVersion pv) {
+      /// Notify analyzer services about a new version, and *DO NOT* do the
+      /// same with search service. The later will get notified after analyzer
+      /// ran the first analysis on the new version.
+      notifyAnalyzer(pv.packageName, pv.versionString);
+    }, onError: (_) => null);
+    return await storeFuture;
```
mkustermann: No need to await the future, can just return it. But why doing this in the first place:

```dart
final PackageVersion pv = await db.withTransaction(....);
/// Notify analyzer services about a new version, and *DO NOT* do the
/// same with search service. The later will get notified after analyzer
/// ran the first analysis on the new version.
notifyAnalyzer(pv.package, pv.versionString);
return pv;

PR #198, opened 2017-06-29
Analyzer task sources: head and history.

isoos:
This is an alternative to implementation, most notable differences:
- at task scheduler we have only a single stream, and do not repeatedly call currentTasks()
- manual trigger has a buffer and a timer to protect against flooding
- head task source is polling every minute for the last 10-minutes window
- history task source is a simple query with an asyncMap that delays each item's processing with 30 seconds.

also removed the nap.

### DIFF
--- a/app/lib/analyzer/task_sources.dart
+++ b/app/lib/analyzer/task_sources.dart
@@ -4,24 +4,93 @@
 
 import 'dart:async';
 
+import 'package:gcloud/db.dart';
+
+import '../frontend/models.dart';
 import '../shared/task_scheduler.dart';
 
+import 'models.dart';
+
+/// Creates a task when a version uploaded in the past 10 minutes has no
+/// analysis yet.
 class DatastoreHeadTaskSource extends PollingTaskSource {
-  DatastoreHeadTaskSource() : super(new Duration(minutes: 10));
+  final DatastoreDB _db;
+  DateTime _lastTs;
+  DatastoreHeadTaskSource(this._db) : super(const Duration(minutes: 1));
 
   @override
-  Stream<Task> pollTasks() {
-    // TODO: implement pollTasks
-    return new Stream.empty();
+  Future poll() async {
+    final DateTime now = new DateTime.now().toUtc();
+    final DateTime tenMinutesAgo = now.subtract(const Duration(minutes: 10));
+    DateTime minCreated;
+    if (_lastTs == null) {
+      minCreated = tenMinutesAgo;
+    } else if (_lastTs.isBefore(tenMinutesAgo)) {
+      // more than ten minutes passed since the last poll
+      minCreated = _lastTs;
+    } else {
+      minCreated = tenMinutesAgo;
+    }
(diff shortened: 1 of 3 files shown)


app/lib/analyzer/task_sources.dart
```diff
@@ -4,24 +4,93 @@
...
+    DateTime minCreated;
+    if (_lastTs == null) {
+      minCreated = tenMinutesAgo;
+    } else if (_lastTs.isBefore(tenMinutesAgo)) {
+      // more than ten minutes passed since the last poll
+      minCreated = _lastTs;
+    } else {
+      minCreated = tenMinutesAgo;
+    }
+    _lastTs = tenMinutesAgo;
```
mkustermann: Since you unconditionally set `_lastTs = tenMinutesAgo`, the `else if (_lastTs.isBefore(tenMinutesAgo)` will always be true (except in the very first call), so I don't think we ever go into the `else` branch.

Also as mentioned in another comment, `_lastTs` can be simply set to the newest `PackageVersion.created` that the query returned.

app/lib/analyzer/task_sources.dart
```diff
@@ -4,24 +4,93 @@
...
+          analysisVersion != null && version.analysisVersion != analysisVersion;
+
+      if (versionDiffers || diff.inDays >= afterDays) {
+        addTask(new Task(version.packageName, version.packageVersion));
+      }
+    }
   }
+
+  Future<PackageVersion> _delay(Model m) =>
+      new Future.delayed(period, () => m as PackageVersion);
```
mkustermann: The `asyncMap` is not really necessary, you can simply have `await new Future.delayed()` inside the `await for (...)` loop above, it's easier to read anyways.

app/lib/shared/task_scheduler.dart
```diff
@@ -14,60 +14,70 @@ typedef Future TaskRunner(Task task);
...
       }
+      if (_timer == null) {
+        _timer = new Timer(const Duration(minutes: 1), () {
+          _timer = null;
+          for (Task task in _triggered) {
+            _controller.add(task);
+          }
+          _triggered.clear();
+        });
+      }
```
mkustermann: This would also look **much** nicer if we just used an `async*` generator with a sleep in the for loop body :-/

app/lib/shared/task_scheduler.dart
```diff
@@ -14,60 +14,70 @@ typedef Future TaskRunner(Task task);
...
-abstract class PollingTaskSource implements TaskSource {
-  final Duration minimumPeriod;
-  DateTime _lastPollTime;
-
-  PollingTaskSource(this.minimumPeriod);
+/// Task source that has a limit on how frequently it will poll its data source.
+abstract class PollingTaskSource extends TaskSource {
+  final Duration _period;
+  Timer _timer;
+  PollingTaskSource(Duration period):_period = period;
```
mkustermann: Consider running the formatter on the code.

PR #199, opened 2017-06-30
Follow-up on task scheduler and analyzer task sources.

isoos:


app/lib/analyzer/task_sources.dart
```diff
@@ -5,44 +5,52 @@
...
-    await for (PackageVersion pv in q.run()) {
-      final List<PackageVersionAnalysis> items = await _db.lookup([
-        _db.emptyKey
-            .append(PackageAnalysis, id: pv.package)
-            .append(PackageVersionAnalysis, id: pv.version)
-      ]);
-      if (items.first == null) {
-        addTask(new Task(pv.package, pv.version));
+        final Query q = _db.query(PackageVersion)
+          ..filter('created >=', minCreated);
```
mkustermann: You could add `..order('created')` here, to ensure if an exception occurs during querying/analysis, the update of `_lastTs` is strictly increasing (i.e. we don't start from newest to oldest, but from oldest to newest).

app/lib/shared/task_scheduler.dart
```diff
@@ -49,37 +44,16 @@ class ManualTriggerTaskSource extends TaskSource {
     });
   }
 
+  @override
+  Stream<Task> startStreaming() => _controller.stream;
```
mkustermann: To simplify the code in the constructor, you can actually just listen on the port here:
```dart
  await for (final Task task in _taskReceivePort) {
    // Could sleep here to throttle if wanted.
    yield task;
  }
}
```

mkustermann:
Looks really good, two minor comments

app/lib/shared/task_scheduler.dart
isoos: I believe that this is still insufficient for a certain kind of flooding attack, but overall it is a very low risk and low impact if it happens, keeping it simple is just fine.

PR #202, opened 2017-07-03
Trigger search reindexing after analysis is done.

isoos:


isoos:
I've found two issues with this, will re-do in another way.

PR #206, opened 2017-07-04
Fix Analysis store: its parents objects must exist.

isoos:
This is also switching from Datastore-generated ID to manual, using the current timestamp's millis since epoch as ID.

Because both `PackageAnalysis` and `PackageVersionAnalysis` stores the id of the `Analysis`, we either need to create their instance first with null id (with all of the additional error handling that comes with it), or generate the id ourselves.

Using millis since epoch is low risk here, as `Analysis` composite id is prefixed with both package and version.

### DIFF
--- a/app/lib/analyzer/backend.dart
+++ b/app/lib/analyzer/backend.dart
@@ -63,12 +63,10 @@ class AnalysisBackend {
   /// Stores the analysis, and either creates or updates its parent
   /// [PackageAnalysis] and [PackageVersionAnalysis] records.
   Future storeAnalysis(Analysis analysis) async {
-    assert(analysis.id == null);
-    await db.commit(inserts: [analysis]);
-    assert(analysis.id != null);
-
     // update package and version too
     await db.withTransaction((Transaction tx) async {
+      analysis.id =
+          await (tx.db.datastore.allocateIds([analysis.key])).first.id;
       final Key packageKey =
           db.emptyKey.append(PackageAnalysis, id: analysis.packageName);
       final Key packageVersionKey = packageKey.append(PackageVersionAnalysis,
@@ -90,11 +88,7 @@ class AnalysisBackend {
         inserts.add(version);
       }
 
-      if (inserts.isEmpty) {
-        await tx.rollback();
-        return;
-      }
-
+      inserts.add(analysis);
       tx.queueMutations(inserts: inserts);
       await tx.commit();
     });


app/lib/analyzer/backend.dart
```diff
@@ -63,10 +63,7 @@ class AnalysisBackend {
   /// Stores the analysis, and either creates or updates its parent
   /// [PackageAnalysis] and [PackageVersionAnalysis] records.
   Future storeAnalysis(Analysis analysis) async {
-    assert(analysis.id == null);
-    await db.commit(inserts: [analysis]);
-    assert(analysis.id != null);
-
+    analysis.id ??= new DateTime.now().millisecondsSinceEpoch;
```
mkustermann: Use utc for anything that goes into the database.

mkustermann:
It would be nice if you could write in the CL description **why** you are changing from the auto-assigned ids.

app/lib/analyzer/backend.dart
isoos: The API doc says: "This value is independent of the time zone."

isoos:
> It would be nice if you could write in the CL description why you are changing from the auto-assigned ids.

Updated.

app/lib/analyzer/backend.dart
mkustermann: Not sure which API doc you are referring to. Of course the int can be anything. But if we use timestamps, we should make them utc.

There is also a way to ask the datastore to generate uniqe ids using [dbService.allocateIds()](https://www.dartdocs.org/documentation/gcloud/0.4.0/gcloud.datastore/Datastore/allocateIds.html). This way we would not even risk having low probability of clash (as long as all `Analysis` objects use `allocateIds` then it's guaranteed to be unique).

If you think it would be cleaner, then let's do it like this, otherwise let me know and I'll merge it as it is now.

isoos: Something like this [the current new code]? I'm not sure if it is supposed to be called in our outside of the transaction.

Btw. with the `DateTime`-based solution, I was referring to this API:
https://api.dartlang.org/stable/1.24.2/dart-core/DateTime/millisecondsSinceEpoch.html

with that, `new DateTime.now().millisecondsSinceEpoch == new DateTime.now().toUtc().millisecondsSinceEpoch`

app/lib/analyzer/backend.dart
```diff
@@ -63,12 +63,10 @@ class AnalysisBackend {
...
   /// [PackageAnalysis] and [PackageVersionAnalysis] records.
   Future storeAnalysis(Analysis analysis) async {
-    assert(analysis.id == null);
-    await db.commit(inserts: [analysis]);
-    assert(analysis.id != null);
-
     // update package and version too
     await db.withTransaction((Transaction tx) async {
+      analysis.id =
+          await (tx.db.datastore.allocateIds([analysis.key])).first.id;
```
mkustermann: I think setting the key is also fine:
```dart
analysis.key = (await dbService.allocateIds([analysis.key]).first;
```

app/lib/analyzer/backend.dart
mkustermann: Thanks for the explanation, that makes perfect sense.

It doesn't matter if you call it outside a transaction or not, once you get new id's back from the datastore, it is guaranteed they are not returned again. So in worst case there are just a few ids which got returned but are unused(e.g. when rolling back the inserting transaction).

PR #207, opened 2017-07-04
Notify search service after analysis is stored.

isoos:
This depends on the handler change in https://github.com/dart-lang/pub-dartlang-dart/pull/196

### DIFF
--- a/app/lib/shared/utils.dart
+++ b/app/lib/shared/utils.dart
@@ -178,3 +178,19 @@ void notifyAnalyzer(String package, String version) {
     _logger.info('Environment was not initialized: $e');
   }
 }
+
+
+void notifySearch(String package) {
+  try {
+    final String host = activeConfiguration.searchServiceHost;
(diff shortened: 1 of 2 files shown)


app/lib/shared/utils.dart
```diff
@@ -178,3 +178,19 @@ void notifyAnalyzer(String package, String version) {
...
+
+
+void notifySearch(String package) {
+  try {
+    final String host = activeConfiguration.searchServiceHost;
+    final String uri = 'https://$host/packages/$package';
+    // Don't block on the notification request, and don't fail even if there was
+    // an error.
+    _client.post(uri).then((_) => null, onError: (e) {
+      _logger.info('Notification request on $uri failed: $e');
```
mkustermann: Please make it an error, so we can see it in the cloud console logs under errors. We have so many "info" log entries, this is just going under there.

(also below)

PR #209, opened 2017-07-05
Show example content on package page + model attributes + backfill.

isoos:
The backfill part provides a good testing ground for using the new version of the `archive` package instead of `tar`, this way we'll be sure that everything in the pub repo is extractable by it.

### DIFF
--- a/app/lib/frontend/models.dart
+++ b/app/lib/frontend/models.dart
@@ -150,8 +150,9 @@ class PackageVersion extends db.ExpandoModel {
   String readmeContent;
 
   FileObject get readme {
-    if (readmeFilename != null)
+    if (readmeFilename != null) {
       return new FileObject(readmeFilename, readmeContent);
+    }
     return null;
   }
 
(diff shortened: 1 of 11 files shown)


app/pubspec.yaml
```diff
@@ -4,6 +4,7 @@ author: Dart Team
 description: The pub.dartlang.org website.
 sdk: ">=1.19.0 <2.0.0"
 dependencies:
+  archive: '^1.0.31'
```
mkustermann: Would you mind moving it to the third-party section and pinning the version?

app/lib/frontend/models.dart
```diff
@@ -167,6 +167,18 @@ class PackageVersion extends db.ExpandoModel {
...
   }
 
+  @db.StringProperty(indexed: false)
+  String exampleFilename;
+
+  @db.StringProperty(indexed: false)
+  String exampleContent;
+
+  FileObject get example {
+    if (exampleFilename != null)
```
mkustermann: Normally we prefer using `{ ... }` if the statement doesn't fit on one line.

isoos: Fixed the other two as well.

PR #216, opened 2017-07-07
Improve analysis target check: flat structure, handle more cases.

isoos:


app/lib/analyzer/backend.dart
```diff
@@ -110,33 +111,52 @@ class AnalysisBackend {
...
+      _logger.warning(
+          'Newer analysis version detected, current instance is obsolete.');
+      return false;
+    }
+
+    assert(versionAnalysis.analysisVersion == analysisVersion);
+
+    // Was latest analysis aborted?
+    if (versionAnalysis.analysisStatus == AnalysisStatus.aborted) {
+      return true;
```
mkustermann: I don't exactly remember when we get to `AnalysisVersion.aborted` but we have to be careful to avoid re-analyzing again and again if analysis failed (e.g. due to broken pana or whatever).

app/lib/analyzer/backend.dart
```diff
@@ -110,33 +111,52 @@ class AnalysisBackend {
...
+    }
+
+    // Is latest analysis older than 4 hours?
+    final DateTime now = new DateTime.now().toUtc();
+    if (now.difference(versionAnalysis.analysisTimestamp).inHours < 4) {
```
mkustermann: I would consider significantly increasing this, since we could potentially re-analyze all uploaded packages every 4 hours (depending on how often this function is called).

Didn't we have another `Duration` somewhere else in the code which represented the same conceptual timeout (i.e. "allow re-analysis of same package/version again after X days")?

app/lib/analyzer/backend.dart
isoos: Good point. In the theory the abort could happen for reasons outside of `pana`, e.g. a transient network error, but in either case, it is better to not do it again immediately. I'll remove this part and the following block with the 4-hour window could act as a safeguard.

app/lib/analyzer/backend.dart
isoos: A malicious source could, through the URL-based task trigger, add the same task in every minute or so. We need to check this either here, or do an additional DB lookup when we are adding the task. This one looks more efficient, as we already have the data, and this edge-case is a low-risk attack vector, which I wanted to address, but not overdo.

kevmoo:
@isoos did you address the comments from @mkustermann ?

isoos:
> @isoos did you address the comments from @mkustermann ?

I've removed the `aborted` part, but kept the other one (commented on both for the reasons).

app/lib/analyzer/backend.dart
mkustermann: iirc we already discussed that it would be an option to make the "frontend -> analyzer/search/... service" communication restricted, thereby just not allowing 3rd parties to call these services.

afaik this can be done via the special x-* headers which will be stripped from 3rd-party requests.

app/lib/analyzer/backend.dart
```diff
@@ -110,33 +111,47 @@ class AnalysisBackend {
...
+      _logger.warning(
+          'Newer analysis version detected, current instance is obsolete.');
+      return false;
+    }
+
+    assert(versionAnalysis.analysisVersion == analysisVersion);
+
+    // Is latest analysis older than 4 hours?
+    final DateTime now = new DateTime.now().toUtc();
+    if (now.difference(versionAnalysis.analysisTimestamp).inHours < 4) {
```
mkustermann: So I've looked up the constant we should share, namely the one defined [here](https://github.com/dart-lang/pub-dartlang-dart/blob/fb5f9479db5233fdea5ae4fa6da48308e5609b20/app/lib/analyzer/task_sources.dart#L111):

```dart
  ....
  await new Future.delayed(const Duration(days: 1));
```

This duration has precisely the meaning to not re-analyze historical data more often than once a day. The duration in this CL right here should have the same meaning, so we should put it into a shared place and use the constant in both places.

app/lib/analyzer/backend.dart
isoos: That is true, we can restrict that. But even then, we could have duplicates e.g. when
- (a) first the trigger comes in via HTTP
- (b) the analyzer starts working on it, but there is no sign of it in the DB yet
- (c) the head poller source creates a task too, because it notices the new version
- (d) the analyzer completes the task, and schedules the next one, which happens to be the same package and version.

mkustermann:
I'll merge it anyways now, but I think we should share the constant as described in my comment above.

app/lib/analyzer/backend.dart
isoos: > This duration has precisely the meaning to not re-analyze historical data more often than once a day. The duration in this CL right here should have the same meaning, so we should put it into a shared place and use the constant in both places.

I'll do the follow-up with the consts soon.

PR #217, opened 2017-07-07
Analyzer task history in reverse upload order, throttle only the hits.

isoos:


app/lib/analyzer/task_sources.dart
```diff
@@ -77,14 +77,15 @@ class DatastoreHistoryTaskSource implements TaskSource {
...
-          await new Future.delayed(const Duration(seconds: 30));
+        final Query query = _db.query(PackageVersion)..order('-created');
+        await for (PackageVersion pv in query.run()) {
           final List<PackageVersionAnalysis> list = await _db.lookup([
             _db.emptyKey
                 .append(PackageAnalysis, id: pv.package)
                 .append(PackageVersionAnalysis, id: pv.version)
           ]);
           if (list.first == null) {
+            await new Future.delayed(const Duration(seconds: 30));
```
kevmoo: Please document why we're waiting 30s

app/lib/analyzer/task_sources.dart
```diff
@@ -96,6 +97,7 @@ class DatastoreHistoryTaskSource implements TaskSource {
               version.analysisVersion != analysisVersion;
 
           if (versionDiffers || diff.inDays >= afterDays) {
+            await new Future.delayed(const Duration(seconds: 30));
```
kevmoo: ditto

app/lib/analyzer/task_sources.dart
```diff
@@ -77,14 +77,17 @@ class DatastoreHistoryTaskSource implements TaskSource {
...
+        await for (PackageVersion pv in query.run()) {
           final List<PackageVersionAnalysis> list = await _db.lookup([
             _db.emptyKey
                 .append(PackageAnalysis, id: pv.package)
                 .append(PackageVersionAnalysis, id: pv.version)
           ]);
           if (list.first == null) {
+            // Waiting to throttle the historical tasks. If there is a newly
+            // uploaded package, give it a chance before continuing with these.
+            await new Future.delayed(const Duration(seconds: 30));
```
mkustermann: Since the sleep is orthongonal to other functionality, you could separate the concerns and make a wrapper:

```dart
class ThrottlingTaskSource implements TaskSource {
  final TaskSource innerSource;
  final Duration sleepDuration;

  ThrottlingTaskSource(this.innerTaskSource, this.sleepDuration);

  Stream<Task> startStreaming() async {
    await for (final Task task in innerSource.startStreaming()) {
      yield task;
      await new Future.delayed(sleepDuration);
    }
  }
}
```

isoos: I'm not sure that an additional class and redirection will make it easier to follow what is happening. I'd rather focus on making the Stream merger part to handle this (either by priorities or doing the throttling there). `package:async`'s `StreamQueue` could help there a lot, but it lacks the synchronous check whether there is an item waiting in the queue.

mkustermann: The fact that the comment is explaining things about "newly uploaded package" is a clear sign that it's code which doesn't fit in there (the only responsibility of the class should be scanning historical packages, it shouldn't care about how many task sources the user of the class has and how the tasks are prioritized).

But if you want to move that functionality in the stream merging in the future, it's fine with me.

PR #222, opened 2017-07-10
Point README's relative links to the package homepage. Fixes #221.

### LINKED ISSUE #221: Relative links in GitHub READMEs are broken on pub.dartlang.org
On this page:

https://pub.dartlang.org/packages/w_flux

There is this text:

> Simple examples of `w_flux` usage can be found in the example directory. The example `README` includes instructions for building / running them.

The text `README` is a relative link work renders fine on GitHub:

`The example [README](example/README.md)
includes instructions for building / running them.`

However on the pub page it's still relative and therefore [a 404](https://pub.dartlang.org/packages/example/README.md).

I don't know whether you'll consider this as a bug, but I think there are only two fixes:

1. The pub site rewrites relative links based on the original location
2. Relative links are not supported and the user must make them absolute

The problem with 2 is that it's not portable - the links will then always go to the same place (the master branch at GitHub). If you push the code somewhere else and it gets rendered, or if you're viewing it locally, it won't link to the file within the same folder as you'd expect.

isoos:
Refactored markdown processing, overriding the inline syntax parser for links.

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -9,25 +9,13 @@ import 'dart:io';
 import 'dart:math';
 
 import 'package:gcloud/service_scope.dart' as ss;
-import 'package:markdown/markdown.dart';
 import 'package:mustache/mustache.dart' as mustache;
 
+import '../shared/markdown.dart';
 import '../shared/mock_scores.dart';
 
 import 'models.dart';
 import 'search_service.dart' show CseTokens, SearchQuery, SearchResultPage;
-
-final List<BlockSyntax> _blockSyntaxes =
-    ExtensionSet.gitHub.blockSyntaxes.toList();
-
-final List<InlineSyntax> _inlineSyntaxes = ExtensionSet.gitHub.inlineSyntaxes
-    .where((s) => s is! InlineHtmlSyntax)
-    .toList();
-
-String _markdownToHtml(String text) => markdownToHtml(text,
-    extensionSet: ExtensionSet.none,
-    blockSyntaxes: _blockSyntaxes,
-    inlineSyntaxes: _inlineSyntaxes);
 
 String _escapeAngleBrackets(String msg) =>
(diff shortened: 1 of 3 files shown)


app/lib/frontend/templates.dart
```diff
@@ -9,26 +9,14 @@ import 'dart:io';
...
 import '../shared/mock_scores.dart';
 
 import 'models.dart';
 import 'search_service.dart' show CseTokens, SearchQuery, SearchResultPage;
 
-final List<BlockSyntax> _blockSyntaxes =
-    ExtensionSet.gitHub.blockSyntaxes.toList();
-
-final List<InlineSyntax> _inlineSyntaxes = ExtensionSet.gitHub.inlineSyntaxes
-    .where((s) => s is! InlineHtmlSyntax)
```
mkustermann: This was here in order to prevent 3-rd party html on pub.dartlang.org. Why did you remove it?

isoos: Haven't removed it, only moved to `lib/shared/markdown.dart`

app/lib/shared/markdown.dart
```diff
@@ -0,0 +1,76 @@
...
+class _RelativeLinkSyntax extends m.LinkSyntax {
+  final Uri baseUri;
+  _RelativeLinkSyntax(this.baseUri);
+
+  static _RelativeLinkSyntax parse(String url) {
+    try {
+      final Uri uri = Uri.parse(url);
+      if (uri.scheme != 'http' && uri.scheme != 'https') return null;
+      if (uri.host == null || uri.host.isEmpty || !uri.host.contains('.'))
+        return null;
```
mkustermann: For multiline if's we use {}s. Maybe there's an analyzer linter flag we can turn on?

app/lib/frontend/templates.dart
mkustermann: Sorry my bad.

PR #223, opened 2017-07-11
Display platform info from the analyzer service on the package page.

isoos:
I'm not sure that this is the best way to do this wrt. the `AnalysisClient` and the `PlatformReport`, but I wanted to flush out the idea and see how it sticks.

The working version can be seen here:
https://20170711t074110-dot-dartlang-pub-dev.appspot.com/packages/async?experimental-service=true
https://20170711t074110-dot-dartlang-pub-dev.appspot.com/packages/angular2?experimental-service=true

Not: the current report will say that angular2 works on flutter, because there is at least one `.dart` file in the angular package that works with it. A per-file report would be more factually correct, but would take much more screen space, probably a separate tab.


mkustermann:
This is something for @kevmoo to decide.

A few comments:

* The html icon is not quite obvious to me, not sure if there's a better one. The icons could have hover-over text, which describes it as text "web", "flutter mobile", "server / console".

* Why is the platform detection done on the client of analysis service and not on the analysis service side? It could even be done by pana and stored in the DB, instead of re-computing it over and over again.

* We should only use public libraries (e.g. under 'lib/' but not '/lib/src'). Is pana doing this or where is that logic?

* Not sure if it's best to have all 3 icons for platform-independent packages. As a user I have then the impression that the package provides specific flutter support / is optimized for flutter, even though it's platform-neutral. We could split the concepts: "has platform independent public library" and "has flutter-specific public library", "has web specific public library", ...

isoos:
> We should only use public libraries (e.g. under 'lib/' but not '/lib/src').

This is already how pana works: it *reports* only about these, but of course analyzes the transitive dependencies, including the ones inside `lib/srv`.

> Why is the platform detection done on the client of analysis service and not on the analysis service side? It could even be done by pana and stored in the DB, instead of re-computing it over and over again.

`pana` should be doing more strict schema evolution, and even when it is breaking, it should support to read back an older version of the analysis. That means everything we introduce there, it will live for a long time, if nowhere else, in the deserialization code.

I am undecided about the usefulness of the current `PlatformReport`. Baking these into `pana` may be good in the long term, but I'd do that only after we are happy about the data structures we are actually using in pub. Keeping these as thin and calculable views will achieve this goal: once we are happy, it could be moved into `pana` without much hassle, but if we change our mind (e.g. the UI requires a different view), we can just drop it and don't tax `pana` with it.

> icons, UX:

I haven't put much effort into those parts, just wanted to see *something* that is coming from the analyzer service. I'm still waiting for the new UI mocks. Btw. the icons have both `title` and `alt` attributes, they should have a small hint of text when you hover over them for enough time.

app/lib/shared/analyzer_client.dart
```diff
@@ -0,0 +1,125 @@
...
+    }
+    final Summary summary = new Summary.fromJson(_analysisData.analysisContent);
+
+    final Set<String> uses = new Set();
+    final Set<String> worksOn = new Set();
+    final Set<String> conflictsWith = new Set();
+
+    bool worksEverywhere = true;
+    bool worksAnywhere = false;
+    for (DartFileSummary dfs in summary.dartFiles.values) {
```
mkustermann: Is this `summary.DartFiles` all dart libraries or just the public libraries of the package?

We should exclude any libraries under `lib/src`.

mkustermann:
Just a few comments

app/lib/frontend/handlers.dart
```diff
@@ -356,20 +361,28 @@ Future<shelf.Response> packageVersionHandlerHtml(
...
+    final Future<AnalysisReport> analysisReportFuture = useAnalyzerService
+        ? analyzerClient.getAnalysisData(packageName, selectedVersion.version)
+        : new Future.value();
+
     final versionDownloadUrls =
         await Future.wait(first10Versions.map((PackageVersion version) {
       return backend.downloadUrl(packageName, version.version);
     }).toList());
 
+    final AnalysisReport report = await analysisReportFuture;
```
isoos: To run it in concurrently to the async construction of `versionDownloadUrls` on the next few lines. It reduces a bit of latency, and doesn't seem to make things complicated.

app/lib/shared/analyzer_client.dart
```diff
@@ -0,0 +1,125 @@
...
+/// analysis data. This keeps the interface narrow over the raw analysis data.
+class AnalyzerClient {
+  final http.Client _client = new http.Client();
+  final String _analyzerServiceHost;
+  AnalyzerClient(this._analyzerServiceHost);
+
+  /// Gets the analysis data from the analyzer service via HTTP.
+  Future<AnalysisReport> getAnalysisData(String package, String version) async {
+    final String uri =
+        'https://$_analyzerServiceHost/packages/$package/$version';
```
isoos: Only the host is coming from there. I can do a follow-up to change all of those to include the schema too.

app/lib/shared/analyzer_client.dart
isoos: The `dartFiles` contains all of the files, but files in `lib/src` do not contain `DartFileSummary.platform` field (that is the reason to check for it on the line below).

I'll update pana to make this distinction more obvious. I'm going to add a comment and TODO for now.

isoos:
@kevmoo: would splitting up this change helpful? I'd like to use something like the AnalyzerClient in search too, extracting that would also make the rest of this easier...

isoos:
I've started a few unrelated PRs that will introduce too many merge conflict here, will close this and re-introduce in a smaller batches, hopefully clearing all of the pending question.

PR #225, opened 2017-07-11
Poor man's priority stream merger, removing the need for manual stream throttle.

isoos:
I've been looking into `async` and similar packages, but couldn't find anything that does it similarly to our needs. `StreamQueue` came close, but it lack a synchronous `hasPending` or similar indication whether the queue has anything immediate, and without it the code was more complex than the current PR.

At this point I think we could remove the `StreamSource` interface altogether, and use streams only, but keeping that to a follow-up, once we are settled with this one.

### DIFF
--- a/app/lib/shared/task_scheduler.dart
+++ b/app/lib/shared/task_scheduler.dart
@@ -3,8 +3,8 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:collection';
 
-import 'package:async/async.dart';
 import 'package:logging/logging.dart';
 
 final Logger _logger = new Logger('pub.scheduler');
@@ -25,12 +25,7 @@ class ManualTriggerTaskSource implements TaskSource {
   ManualTriggerTaskSource(this._taskReceivePort);
 
   @override
-  Stream<Task> startStreaming() async* {
-    await for (final Task task in _taskReceivePort) {
-      yield task;
-      await new Future.delayed(const Duration(seconds: 10));
-    }
-  }
+  Stream<Task> startStreaming() => _taskReceivePort;
 }
 
 /// Schedules and executes package analysis.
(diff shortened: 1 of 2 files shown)


isoos:
It is early to tell, but this change increased average CPU utilization from 60% to well above 85%. Hopefully it is not spinning in the scheduler, rather doing more task without much waiting in-between.

app/lib/shared/task_scheduler.dart
```diff
@@ -41,16 +36,34 @@ class TaskScheduler {
...
+      source.startStreaming().listen((Task task) {
+        queue.add(task);
+        newTaskTrigger.add(null);
+      });
+    }
 
-    await for (Task task in stream) {
+    // On new task, pick the first queue that has a task and execute that.
+    await for (var _ in newTaskTrigger.stream) {
+      final Queue<Task> queue = priorityQueues.firstWhere((q) => q.isNotEmpty);
```
kevmoo: Do we have to worry about `priorityQueues` ever being empty?

isoos: there is 1:1 mapping between adding one item to any queue + triggering the newTaskTrigger.stream. Also, on each item of that stream we remove exactly one task. At this point one of the queues should be always non-empty.

kevmoo: Might be worth a note inline

app/lib/shared/task_scheduler.dart
```diff
@@ -41,16 +36,37 @@ class TaskScheduler {
...
+
+    // Listen on the streams and put tasks into their own queues.
+    for (int i = 0; i < sources.length; i++) {
+      final TaskSource source = sources[i];
+      final Queue<Task> queue = priorityQueues[i];
+      source.startStreaming().listen((Task task) {
+        queue.add(task);
+        newTaskTrigger.add(null);
+      });
+    }
```
mkustermann: Consider encapsulating this functionality into it's own function/class, just like `StreamGroup.merge`, maybe `class PrioritizedStreamGroup<T> {}`.

isoos:
> Just one comment: it would be nicer if the functionality was in a separate class, since it's cross-cutting functionality.

I have settled with a pull-based async iterator, I think that suited this use case the best.

kevmoo:
What's the status w/ this, @isoos ?

isoos:
> What's the status w/ this?

I think it is ready to merge, barring the comment in the other PR to unit test the async iterator. If you don't mind, I'd do that in a follow-up.

PR #226, opened 2017-07-17
Snapshots to speed up search service initialization

isoos:


app/lib/search/backend.dart
```diff
@@ -76,3 +91,102 @@ class SearchBackend {
...
+        await bucket.delete(name);
+      } catch (e, st) {
+        _logger.warning('Snapshot delete failed: $name', e, st);
+      }
+    }
+  }
+}
+
+class SearchSnapshot {
+  DateTime updated;
```
kevmoo: can be final?

isoos: it could, but it would make tracking the last updated timestamp much more complex and awkward. so, no, can't. :)

kevmoo: ack

app/lib/search/backend.dart
```diff
@@ -76,3 +91,102 @@ class SearchBackend {
...
+    if (names.isEmpty) return null;
+    // Select the first entry, because there is chance that later entries were
+    // aborted for some reason, and the most probable correct entry is the first
+    // one.
+    names.sort();
+    final String selected = names.first;
+    // reading
+    final Stream<List<int>> storageStream = bucket.read(selected);
+    final Stream<List<int>> unzippedStream = _gzip.decoder.bind(storageStream);
+    final String json = await UTF8.decoder.bind(unzippedStream).join();
```
mkustermann: Small nit: This can be also chained like `bucket.read(selected).transform(_gzip.decoder).transform(UTF8.decoder).single` I think.

app/lib/search/backend.dart
```diff
@@ -76,3 +91,102 @@ class SearchBackend {
...
+    final List<String> names = list
+        .where((entry) => entry.isObject)
+        .map((entry) => entry.name)
+        .toList();
+    if (names.isEmpty) return null;
+    // Select the first entry, because there is chance that later entries were
+    // aborted for some reason, and the most probable correct entry is the first
+    // one.
+    names.sort();
+    final String selected = names.first;
```
mkustermann: Since the names are sorted by timestamp, this will effectively pick the oldest one. Why not picking `names.last`?

Also: The cloud storage uploads are atomic: Only if all data got written will the object be created.

app/lib/search/backend.dart
```diff
@@ -76,3 +91,102 @@ class SearchBackend {
...
+    final List<int> buffer =
+        _gzip.encode(UTF8.encode(JSON.encode(snapshot.toJson())));
+
+    // generate name from current timestamp
+    final String ts =
+        new DateTime.now().toUtc().toIso8601String().replaceAll(':', '-');
+    final currentName = 'snapshot-$ts.json.gz';
+
+    // upload
+    final StreamSink<List<int>> sink = bucket.write(currentName);
```
mkustermann: Since you have all data in `buffer`, why not use `bucket.writeBytes`?

PR #229, opened 2017-07-18
Replace ducene with a simple inverted index (due to perf reasons).

isoos:


app/lib/search/index_simple.dart
```diff
@@ -0,0 +1,208 @@
...
+      for (int i = 0; i <= text.length - ngramLength; i++) {
+        ngrams.add(text.substring(i, i + ngramLength));
+      }
+    }
+  }
+  return ngrams;
+}
+
+class _PackageResult {
+  String url;
```
isoos: made `url` final, but `score` is more complicated, mostly because of the way I update it with popularity score. To rewrite that part (and align it with the other weights) requires a bit more iterations, I'll happy to do that in a subsequent change too.

PR #235, opened 2017-07-19
Prioritize most recent versions (per-package) over historical ones during re-analysis

isoos:
Update: removed comment about base branch, this was now rebased on master.

### DIFF
--- a/app/lib/analyzer/task_sources.dart
+++ b/app/lib/analyzer/task_sources.dart
@@ -75,26 +75,25 @@ class DatastoreHistoryTaskSource implements TaskSource {
   Stream<Task> startStreaming() async* {
     for (;;) {
       try {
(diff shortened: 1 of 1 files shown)


app/lib/shared/task_scheduler.dart
```diff
@@ -74,3 +70,111 @@ class Task {
   @override
   int get hashCode => version.hashCode;
 }
+
+/// A pull-based interface for accessing events from multiple streams, in the
+/// priority order of the streams provided.
+///
+/// Inspired by Iterator and package:async's StreamQueue.
+class PrioritizedAsyncIterator<T> {
```
kevmoo: Would it be worth having some sanity/unit tests for this class? Seems non-trivial

PR #237, opened 2017-07-19
Fix async iterator's bug when it is closing.

isoos:


kevmoo:
@isoos - this can be removed, right?

isoos:
Yeah, this was included in https://github.com/dart-lang/pub-dartlang-dart/pull/247

PR #238, opened 2017-07-20
Better search relevancy: symmetrical to both query and target text.

isoos:
Started as a follow-up to yesterdays review, but found an easy way to solve many search quirks at once (line 186 was the key, the others are only fine-tuning). Also added tests for the edge cases that document the underlying scores.

### DIFF
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -19,10 +19,10 @@ void registerPackageIndex(PackageIndex index) =>
     ss.register(#packageIndexService, index);
 
 class SimplePackageIndex implements PackageIndex {
-  final Map<String, PackageDocument> _documents = {};
-  final _TokenIndex _nameIndex = new _TokenIndex();
-  final _TokenIndex _descrIndex = new _TokenIndex();
-  final _TokenIndex _readmeIndex = new _TokenIndex();
+  final Map<String, PackageDocument> _documents = <String, PackageDocument>{};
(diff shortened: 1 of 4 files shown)


app/lib/search/index_simple.dart
```diff
@@ -62,64 +62,70 @@ class SimplePackageIndex implements PackageIndex {
 
   @override
   Future<PackageSearchResult> search(PackageQuery query) async {
-    final Map<String, _PackageResult> resultMap = {};
+    final Map<String, double> total = {};
```
kevmoo: For future – put types on the right for literals.

app/lib/search/index_simple.dart
```diff
@@ -62,64 +62,70 @@ class SimplePackageIndex implements PackageIndex {
...
-            !doc.detectedTypes.contains(query.type);
-      });
-    }
+    final Map<String, double> popularityScores = new Map.fromIterable(
+      total.keys,
+      value: (String url) => _documents[url].popularity * 100,
+    );
+    addAll(popularityScores, 0.05);
+
+    List<PackageScore> results = [];
```
kevmoo: ditto – var results = <PackageScore>[];

isoos:
> For future – put types on the right for literals.

Fixed the ones in this file. Shall we turn on `prefer_collection_literals`?

kevmoo:
> Fixed the ones in this file. Shall we turn on prefer_collection_literals?

Sure!

PR #239, opened 2017-07-21
Randomize analyzer tasks to reduce duplicates in case multiple instances are running

isoos:
`stream_transform`'s `buffer` was published ~9 days ago, and it is a perfect fit for us.
There will be collisions (when two instances start to re-analyze everything for a new pana version), but not too many, and if with increasing the slice sizes or the buffer time, it can be reduced further.

### DIFF
--- a/app/lib/shared/utils.dart
+++ b/app/lib/shared/utils.dart
@@ -12,6 +12,7 @@ import 'dart:math';
 import 'package:logging/logging.dart';
 import 'package:http/http.dart' as http;
 import 'package:pub_semver/pub_semver.dart' as semver;
+import 'package:stream_transform/stream_transform.dart';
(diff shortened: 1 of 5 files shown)


app/lib/shared/utils.dart
```diff
@@ -208,3 +209,27 @@ List<String> exampleFileCandidates(String package) => [
       'example/$package.dart',
       'example/example.dart',
     ];
+
+/// Buffers for [duration] and then randomizes the order of the items in the
+/// stream. For every single item, their final position would be in the range of
+/// [maxPositionDiff] of its original position.
+Stream<T> randomizeStream<T>(
```
kevmoo: dartfmt?

PR #247, opened 2017-07-24
Make PrioritizedAsyncIterator more like AsyncIterator

isoos:
This is branched off of https://github.com/dart-lang/pub-dartlang-dart/pull/237

On the weekend I was playing with some async concepts, and figured that nobody created an `AsyncIterator` interface like the following:

```
abstract class AsyncIterator<T> {
  Future<bool> moveNext();
  T get current;
}
```

This is very close to the `Iterator` interface, except `moveNext()` is async. I think it is useful to have a lazy async API like this, and I'm planning to publish it soon.

I've rewritten the pub site's code to conform the same API.

### DIFF
--- a/app/lib/shared/task_scheduler.dart
+++ b/app/lib/shared/task_scheduler.dart
@@ -39,8 +39,8 @@ class TaskScheduler {
     final PrioritizedAsyncIterator<Task> taskIterator =
         new PrioritizedAsyncIterator(
             sources.map((TaskSource ts) => ts.startStreaming()).toList());
-    while (await taskIterator.hasNext) {
-      final Task task = await taskIterator.next;
+    while (await taskIterator.moveNext()) {
+      final Task task = taskIterator.current;
       try {
         await taskRunner(task);
       } catch (e, st) {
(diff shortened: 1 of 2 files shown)


kevmoo:
Isn't this called Stream?

isoos:
> Isn't this called Stream?

While Stream could use StreamController with pause/resume listeners to have a lazy evaluation, neither the provider, nor the consumer implementation is easy or can be short. With this AsyncIterator, the consumer side is as simple as `while (await iter.moveNext()) { print(iter.current); }`. The producer side is more complex in pub's case, but in my experiments I used a very simple `Future<T> next()` callback that feeds the producer side (until a `null` value is returned).

AFAIK `await for` iteration of a Stream doesn't pause/resume it, nor does it suspend methods with `async*`. I see an unfilled niche here.

kevmoo:
Have you looked here first? https://pub.dartlang.org/packages/async

isoos:
> Have you looked here first? https://pub.dartlang.org/packages/async

Yes, but I've found nothing comparable. `StreamQueue` was close but that solves a different problem and is not really applicable for such case. Do you know anything specific that I shall look into?

mkustermann:
> On the weekend I was playing with some async concepts, and figured that nobody created an 
> AsyncIterator interface like the following:

> abstract class AsyncIterator<T> {
>     Future<bool> moveNext();
>     T get current;
> }

There is such an interface in `dart:async` and it's called [StreamIterator](https://api.dartlang.org/stable/1.24.2/dart-async/StreamIterator-class.html).

isoos:
Ah, good catch! I'll rename it accordingly.

On Thu, Jul 27, 2017 at 5:16 PM, Martin Kustermann <notifications@github.com

PR #250, opened 2017-07-24
URL param and query text parsing to restrict search for package prefixes

isoos:
The current Google Custom Search doesn't have any option to enable prefix searches, this only works with the new service. In a subsequent change we can enable that with PageMap tokens within GCE.

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -24,6 +24,9 @@ import 'templates.dart';
 final String StaticsLocation =
     Platform.script.resolve('../../static').toFilePath();
 
+RegExp _packageRegexp =
+    new RegExp('package:([_a-z0-9]+)\\*?', caseSensitive: false);
+
 /// Handler for the whole URL space of pub.dartlang.org
 ///
 /// The passed in [shelfPubApi] handler will be used for handling requests to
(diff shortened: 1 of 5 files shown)


app/lib/frontend/handlers.dart
```diff
@@ -109,7 +112,19 @@ shelf.Response docHandler(shelf.Request request) {
...
 /// Handles requests for /search
 Future<shelf.Response> searchHandler(shelf.Request request) async {
-  final queryText = request.url.queryParameters['q'];
+  String queryText = request.url.queryParameters['q'] ?? '';
+  String packagePrefix = request.url.queryParameters['pkg-prefix'];
+
+  final Match pkgMatch = _packageRegexp.firstMatch(queryText);
+  if (pkgMatch != null) {
+    // only if query attribute was not specified
+    if (packagePrefix == null) {
```
kevmoo: so if not explicit, see if the search values is a valid (prefix) of a package name and use that as the param?

app/lib/frontend/search_service.dart
```diff
@@ -210,12 +218,16 @@ class SearchQuery {
...
     this.type,
+    this.packagePrefix,
     this.bias,
   });
 
   /// Returns the query text to use in CSE.
   String buildCseQueryText() {
     String queryText = text;
+    if (packagePrefix != null) {
+      queryText += ' $packagePrefix';
```
kevmoo: So if the prefix is not explicit, wouldn't this mean the search term is duplicated here?

app/lib/frontend/handlers.dart
kevmoo: So if the prefix is not explicitly given, add it – but only if the search string looks like a package name?

app/lib/frontend/search_service.dart
isoos: we remove the package:[whatever] pattern from the original query string, and in case of CSE, adding it to the query may actually yield some good (although far for correct) results.

app/lib/frontend/handlers.dart
isoos: Yes. I've added the query parameter in case flutter decides to do search from their page with a user-provided query string, they won't need to override that string, rather just specify the url param.

PR #259, opened 2017-07-26
Expose some non-revealing stats in the /debug handlers.

isoos:
Based on https://github.com/dart-lang/pub-dartlang-dart/pull/258
These are very simple, non-revealing metrics, so that I can monitor some crucial internals.

### DIFF
--- a/app/lib/shared/utils.dart
+++ b/app/lib/shared/utils.dart
@@ -5,6 +5,7 @@
 library pub_dartlang_org.utils;
 
 import 'dart:async';
+import 'dart:collection';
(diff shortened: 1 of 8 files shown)


app/lib/shared/utils.dart
```diff
@@ -233,3 +234,29 @@ Stream<T> randomizeStream<T>(
     },
   ));
 }
+
+class LastNTracker<T> {
```
kevmoo: `<T extends Comparable<T>>` – just to keep you honest about sort

PR #260, opened 2017-07-27
Enable multiple isolates for analyzer and search.

isoos:


mkustermann:
I've just published a new `package:appengine` version, which includes the new `shared: ` parameter.

isoos:
> I've just published a new package:appengine version, which includes the new shared: parameter.

Thanks! I've updated the PR with multi-isolate support for search too.

mkustermann:
Will do this the search indexing in multiple isolates (redundantly) ?

app/bin/service/analyzer.dart
```diff
@@ -28,16 +28,17 @@ Future main() async {
...
+        final ReceivePort mainReceivePort = new ReceivePort();
         await Isolate.spawn(
           _runScheduler,
           [mainReceivePort.sendPort],
           onError: errorReceivePort.sendPort,
           onExit: errorReceivePort.sendPort,
+          errorsAreFatal: true,
         );
         final List<SendPort> sendPorts = await mainReceivePort.take(1).toList();
         registerTaskSendPort(sendPorts[0]);
```
isoos: That's true and a good catch! In practice, I don't think this will be a big issue, but I'll do a follow-up on that.

isoos:
> Will do this the search indexing in multiple isolates (redundantly) ?

Yes it will, but that is the only way to make the CPU-bound search run in parallel.

app/bin/service/analyzer.dart
mkustermann: Since we know ahead-of-time the number of isolates, it might make sense to shard the stream of `Task`s.

i.e. letting isolate `isolateId` only analyze packages `"$name/$version".hashCode % #num-isolates == isolateId`. Otherwise the isolates might be traversing the very same stream of `Task`s and analyze them, just to figure out that another isolate was faster.

app/bin/service/search.dart
```diff
@@ -23,7 +23,23 @@ import 'package:pub_dartlang_org/search/handlers.dart';
...
+      _main,
+      i,
+      onError: errorPort.sendPort,
+      onExit: errorPort.sendPort,
+      errorsAreFatal: true,
+    );
+  }
+}
+
+void _main(int isolateId) {
```
mkustermann: If we have several isolate running the normal search indexing & serving code, they will also race when writing checkpoints of the index and deleting old checkpoints. But I think it should be guaranteed that there's always one checkpoint left.

isoos:
> There can be one isolate doing new indexing, checkpoints the index to cloud storage and other isolates (or even VMs) could use it for serving.

Indexing is *fast*. It takes only a couple of seconds on my macbook to index the entire snapshot, and there is no need to separately maintain a serialized format for the index. Separating the indexing, its storage and sharding is just not worth it.

>  they will also race when writing checkpoints of the index and deleting old checkpoints

Good point. I believe this can be mitigated if we restrict snapshot deletion to the names that are earlier than the current one. Worst case there will be 2-4 recent snapshots around, but there won't be a case where two isolates delete each others' updates. (Updated the code with it.)

> Since we know ahead-of-time the number of isolates, it might make sense to shard the stream of Tasks.

analyzer is better off if we have only one isolate working on a triggered.

search is better off with a broadcast notification, but as the isolates deal with the trigger http handler separately, it would require much more effort to ensure every search index is up-to-date with low latency. I'd rather keep it with the current state and in 10 minutes they will be up-to-date anyway.

mkustermann:
Looks fine.

> > Since we know ahead-of-time the number of isolates, it might make sense to shard the stream of Tasks.

> analyzer is better off if we have only one isolate working on a triggered.

Agreed, only one should handle a trigger.

Though, as mentioned, I still think that analyzer isolates will repeatedly analyze the same package (when going back in history). By sharding based on name/version we could prevent that.

PR #263, opened 2017-07-28
Store search pages in memcache.

isoos:
Disclaimer: I was not able to test this in staging, not sure why, but memcache is not working for me there.

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -19,6 +19,7 @@ import 'atom_feed.dart';
 import 'backend.dart';
 import 'handlers_redirects.dart';
 import 'models.dart';
+import 'search_memcache.dart';
 import 'search_service.dart';
 import 'templates.dart';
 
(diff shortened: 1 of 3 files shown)


app/lib/frontend/handlers.dart
```diff
@@ -169,11 +175,15 @@ Future<shelf.Response> searchHandler(shelf.Request request) async {
...
   final links = new SearchLinks(query, resultPage.totalCount);
+  final String content = templateService.renderSearchPage(resultPage, links);
+
   _searchBackendTracker.add(resultPage.backend);
   _searchBackendLatencyTracker.add(resultPage.latency);
   _searchOverallLatencyTracker.add(sw.elapsed);
 
-  return htmlResponse(templateService.renderSearchPage(resultPage, links));
+  // do not await on memcache update
+  searchMemcache?.setUiSearchPage(cacheUrl, content);
```
mkustermann: I think it's better to await for the memcache update for two reasons: Firstly the additional latency is negligible in comparison to the rest of what the search takes (maybe around 1-2 ms), secondly if memcache isn't working or throws an error we will only see it in the logs if the request wasn't done. (If the request handler is done, the error won't show up on the request log entry.)

isoos: Updated with await. Is there any timeout in the `memcache` client that will make sure we don't hang with an otherwise prepared response page?

PR #267, opened 2017-07-31
Use search service by default.

isoos:
Deployment instructions:
- Deploy `search` service in production. Wait until the the `/debug` url reports back that it is ready.
- Deploy `app` in production.

Demo: https://dartlang-pub-dev.appspot.com/search?q=http

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,12 +105,6 @@ To view your application in the web browser run:
   $ gcloud app browse
 ```
 
-Similarly, to deploy a new version of the `analyzer` service:
-
-```
-pub-dartlang-dart $ gcloud app deploy --no-promote analyzer.yaml
-```
-
 This will do a remote docker build in the cloud, push the generated docker image layers to a
 GCS bucket and deploy a new version to AppEngine.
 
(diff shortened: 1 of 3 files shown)


mkustermann:
Are we ready to deploy analysis/search service to prod version? If so, please update the `README.md` file for deployment.

isoos:
> Are we ready to deploy analysis/search service to prod version? If so, please update the README.md file for deployment.

analysis: no, search: yes.

I'll update it shortly, but the "wait for the search service to become available" only applies for the very first deployment. Soon (in half a day) there will be a snapshot, and the service will be ready within a few seconds in every case (either by us creating a new version or by appengine's reboot).

kevmoo:
See https://dartlang-pub-dev.appspot.com/search?q=yaml

Weird to see `haml` here. Is this the search bits matching substrings?

isoos:
> Weird to see haml here. Is this the search bits matching substrings?

Yes, we do match 1-grams, 2-grams, 3-grams and 4-grams (in any place in the phrase) and prefixes, with increasing weight with length.

`yaml` and `haml` have 75% clear overlap, so not surprising that it is listed.

kevmoo:
> yaml and haml have 75% clear overlap, so not surprising that it is listed.

Sure. I guess if we added metrics that weighted age, that'd get better...right?

isoos:
> Sure. I guess if we added metrics that weighted age, that'd get better...right?

We could, but then when `haml` gets updated, it will get more weight... I'd prefer more weight on popularity/actual use, dependent packages etc.

PR #276, opened 2017-08-03
Delete obsolete (older than 180 days) Analysis entries.

isoos:


app/lib/analyzer/backend.dart
```diff
@@ -166,4 +166,27 @@ class AnalysisBackend {
 
     return true;
   }
+
+  Future deleteObsoleteAnalysisEntries() async {
+    final DateTime threshold =
+        new DateTime.now().toUtc().subtract(const Duration(days: 180));
```
kevmoo: Good thing you used `toUtc` here

app/bin/service/analyzer.dart
```diff
@@ -111,3 +113,18 @@ Future _initFlutterSdk() async {
 void _registerServices() {
   registerAnalysisBackend(new AnalysisBackend(db.dbService));
 }
+
+final Random _random = new Random.secure();
```
kevmoo: Good thing this is secure

app/lib/analyzer/backend.dart
isoos: @mkustermann has a strict policy on it :)

app/bin/service/analyzer.dart
isoos: I once worked with an audit guy for a short project. He loved to flag the non-secure random things as 'high-risk' security issues, regardless of the context, and it doesn't harm too much. So why not? :)

kevmoo: Yup. Just being silly...

app/lib/analyzer/backend.dart
kevmoo: 😄

PR #278, opened 2017-08-03
Latency tracking for the analyzer service use too

isoos:
FYI, the latency metrics on staging:
```
  "package": {
    "analysis_latency": {
      "median": 50,
      "p90": 97,
      "p99": 169
    },
    "overall_latency": {
      "median": 125,
      "p90": 213,
      "p99": 386
    }
  },
```

I think we are good to go ahead with displaying some of its info without the experimental flag. (well, when we get to deploy it in prod anyway).

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -31,6 +31,8 @@ RegExp _packageRegexp =
     new RegExp('package:([_a-z0-9]+)\\*?', caseSensitive: false);
 
 // Non-revealing metrics to monitor the search service behavior from outside.
+final _packageAnalysisLatencyTracker = new LastNTracker<Duration>();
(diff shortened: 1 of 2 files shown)


kevmoo:
> I think we are good to go ahead with displaying some of its info without the experimental flag. (well, when we get to deploy it in prod anyway).

Like what info?

isoos:
> Like what info?

I was thinking to display the license from there on the right side block.

Btw. we have so far this behind flag:
https://dartlang-pub-dev.appspot.com/packages/http?experimental-service=true#pub-pkg-tab-analysis

But as it will be behind flag for the upcoming 1-2 weeks anyway (I don't think we'll launch analyzer in prod next week, and I'll be out the week after), it can be just on that tab for now.

PR #285, opened 2017-08-04
Deploy script.

isoos:
There are a couple of quirks that needed special handling, that's why I've ended up with the current version:
- not deleting the old version after a successful migration may push the resource use above the limit (e.g. on staging it is 5, so 3 non-promoted + 3 old version are not able to run together)
- traffic migration fails if either the new or the old version is not on autoscale

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,12 @@ pub-dartlang-dart/app $ pub run test
 
 ## Deploying a new version to production
 
+### Shortcut
+
+Run `dart deploy.dart all --delete-old`
+
+### Longer version
+
 Before being able to deploy, please ensure you have an up-to-date version of the
 [Google Cloud SDK](https://cloud.google.com/sdk/) installed.
 
@@ -120,8 +126,8 @@ pub-dartlang-dart $ git push origin <version>
 ### Services
 
 The pub site uses two services:
-- `search` (production ready)
-- `analyzer` (experimental, staging only)
+- `search`
+- `analyzer`
 
 To deploy each, use the following:
 
--- a/deploy.dart
+++ b/deploy.dart
@@ -1,0 +1,178 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+HttpClient httpClient = new HttpClient();
+
+Future main(List<String> args) async {
+  List<String> services;
+  if (args.isNotEmpty) {
+    switch (args[0]) {
+      case 'all':
+        services = ['analyzer', 'search', 'default'];
+        break;
+      case 'analyzer':
+      case 'analyzer.yaml':
+        services = ['analyzer'];
+        break;
+      case 'search':
+      case 'search.yaml':
+        services = ['search'];
+        break;
+      case 'app':
+      case 'app.yaml':
+      case 'default':
+        services = ['default'];
+        break;
+    }
+  }
+
+  if (services == null) {
+    print('Specify at least one argument:');
+    print('deploy.dart ( app | analyzer | search | all ) [ --delete-old ]');
+    exit(1);
+  }
+
+  String newVersion = new DateTime.now()
+      .toIso8601String()
+      .replaceAll('-', '')
+      .replaceAll(':', '')
+      .replaceAll('T', 't')
+      .split('.')
+      .first;
+  print('New version: $newVersion');
+
+  final bool deleteOld = args.contains('--delete-old');
+  for (String service in services) {
+    print('\nDeploying $service...\n');
+    await new _ServiceDeployer(service, newVersion, deleteOld).deploy();
+  }
+
+  httpClient.close(force: true);
+}
+
+class _ServiceDeployer {
+  final String project;
+  final String service;
+  final String newVersion;
+  final bool migrateTraffic;
+  final bool deleteOld;
+  String _oldVersion;
+
+  _ServiceDeployer(this.service, this.newVersion, this.deleteOld)
+      : project = Platform.environment['GCLOUD_PROJECT'],
+        migrateTraffic =
+            Platform.environment['GCLOUD_PROJECT'] != 'dartlang-pub-dev' {
+    if (project == null) {
+      throw new StateError('GCLOUD_PROJECT must be set!');
+    }
+  }
+
+  Future deploy() async {
+    await _detectOldVersion();
+    await _gcloudDeploy();
+    await _checkHealth();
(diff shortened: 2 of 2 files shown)


kevmoo:
Dude! You could have written this in Dart!

isoos:
> You could have written this in Dart!

Tbh, it started out as a simple script, but gained complexity along the way... It would be worth to rewrite it in Dart if we want to have more test & validation of the service before migrating the traffic, but otherwise this is not too bad for a shell script...

deploy.sh
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env bash
+
+set -e
+
+VERSION=`date "+%Y%m%dt%H%M%S"`
```
mkustermann: Please prefer using`$(...)` over using backticks.

deploy.sh
```diff
@@ -0,0 +1,105 @@
...
+LOGFILE="$VERSION.log"
+
+MIGRATE="--migrate"
+if [ "$GCLOUD_PROJECT" = "dartlang-pub-dev" ]; then
+  MIGRATE=""
+fi
+
+ANALYZER_HOST="https://""$VERSION""-dot-analyzer-dot-""$GCLOUD_PROJECT"".appspot.com"
```
mkustermann: Why don't you use normal interpolation: `https://${VERSION}-dot-....` ?

The ending and beginning `"` are quite hard to read.

deploy.sh
```diff
@@ -0,0 +1,105 @@
...
+SEARCH_HOST="https://""$VERSION""-dot-search-dot-""$GCLOUD_PROJECT"".appspot.com"
+APP_HOST="https://""$VERSION""-dot-""$GCLOUD_PROJECT"".appspot.com"
+
+if [ "$1" = "all" ] || [ "$1" = "analyzer" ]; then
```
mkustermann: One expression is actually enough: ``[ "$1" = "all" -o  "$1" = "analyzer" ]`

see "man test"

deploy.sh
```diff
@@ -0,0 +1,105 @@
...
+  OLD_ANALYZER=`gcloud app versions list --service analyzer --format='value(id)'`
+
+  echo "Deploying analyzer ($VERSION)..."
+  echo "Deploying analyzer ($VERSION)..." >>$LOGFILE
+  gcloud app deploy analyzer.yaml --no-promote -v $VERSION -q >>$LOGFILE 2>&1
+
+  echo "Checking analyzer status on $ANALYZER_HOST ..."
+  echo "Checking analyzer status on $ANALYZER_HOST ..." >>$LOGFILE
+  curl "$ANALYZER_HOST""/debug" -s -f  >/dev/null
```
mkustermann: Doesn't it usually take some time until the service is actually ready? iirc it usually takes a bit more than just waiting for `gcloud app deploy` to complete.

deploy.sh
```diff
@@ -0,0 +1,105 @@
...
+
+set -e
+
+VERSION=`date "+%Y%m%dt%H%M%S"`
+LOGFILE="$VERSION.log"
+
+MIGRATE="--migrate"
+if [ "$GCLOUD_PROJECT" = "dartlang-pub-dev" ]; then
+  MIGRATE=""
+fi
```
mkustermann: Why do we handle dev differently from prod?

deploy.sh
```diff
@@ -0,0 +1,105 @@
...
+  echo "Deploying analyzer ($VERSION)..." >>$LOGFILE
+  gcloud app deploy analyzer.yaml --no-promote -v $VERSION -q >>$LOGFILE 2>&1
+
+  echo "Checking analyzer status on $ANALYZER_HOST ..."
+  echo "Checking analyzer status on $ANALYZER_HOST ..." >>$LOGFILE
+  curl "$ANALYZER_HOST""/debug" -s -f  >/dev/null
+
+  echo "Migrating traffic on analyzer..."
+  echo "Migrating traffic on analyzer..." >>$LOGFILE
+  gcloud app services set-traffic analyzer --splits $VERSION=1 $MIGRATE -q >>$LOGFILE 2>&1
```
mkustermann: Is there a reason why you want to split traffic?

So far I've always manually checked that the deployed version works and then migrated all of the traffic to the new version.

If one splits the traffic than it might take longer for us to see if a version has bugs, since only a subset of the IPs are routed to the newly deployed version. Also reproducing user reported bugs is more difficult due to IP based traffic splitting.

deploy.sh
```diff
@@ -0,0 +1,105 @@
...
+ANALYZER_HOST="https://""$VERSION""-dot-analyzer-dot-""$GCLOUD_PROJECT"".appspot.com"
+SEARCH_HOST="https://""$VERSION""-dot-search-dot-""$GCLOUD_PROJECT"".appspot.com"
+APP_HOST="https://""$VERSION""-dot-""$GCLOUD_PROJECT"".appspot.com"
+
+if [ "$1" = "all" ] || [ "$1" = "analyzer" ]; then
+  OLD_ANALYZER=`gcloud app versions list --service analyzer --format='value(id)'`
+
+  echo "Deploying analyzer ($VERSION)..."
+  echo "Deploying analyzer ($VERSION)..." >>$LOGFILE
+  gcloud app deploy analyzer.yaml --no-promote -v $VERSION -q >>$LOGFILE 2>&1
```
mkustermann: If the deployment was successful we should make a git tag with the version number (and for the prod version we should push the tag to github.).

deploy.sh
```diff
@@ -0,0 +1,105 @@
...
+      echo "No existing analyzer version, exiting."
+      exit 1;
+    fi
+    echo "Delete old analyzer ($OLD_ANALYZER)..."
+    echo "Delete old analyzer ($OLD_ANALYZER)..." >>$LOGFILE
+    gcloud app versions delete --service analyzer $OLD_ANALYZER -q >>$LOGFILE 2>&1
+  fi
+fi
+
+if [ "$1" = "all" ] || [ "$1" = "search" ]; then
```
mkustermann: There's a lot of duplicated code, it's hard to see what the difference is. You could use bash functions with arguments to remove the duplication.

isoos:
> Agree with kevin, it would be nice having less duplication and maybe have it in dart.

I've done a rewrite in dart. Code size increased a lot, and I'm not sure this makes it easier to understand. I've still got the bash version, if you'd rather revert back to it...

> Generally speaking I think I would prefer if we separate deployment from cleaning up of old versions (at least into 2 different scripts). It's good to have old versions around, in case there are issues, one can fall back to an older one in the developers console.

As far as I see it, this script is useful for automating the version change in staging, do your testing with the new version, and if everything looks fine, automate the version change in prod too. Any more complex workflow could be done manually without much effort, and isn't really worth the time to automate. (I've already spent too much time on this small script.)

PR #290, opened 2017-08-09
Expose analyzer stats through /debug handler.

isoos:
Example output:

```
  "scheduler": {
    "pending": 23791,
    "status": {
      "skip": 325,
      "normal": 166,
      "race": 9
    },
    "taskPerHour": 279.7442418608048,
    "remaining": "1 days 1 hours 6 mins"
  }
```

There are a couple of caveats:
- Each isolate sends its stats to the main isolate, if there are more than one, only one of them will be reported.
- Metrics and estimates are far from exact, and if you hit the main host's `/debug` URL, you will see different numbers.

### DIFF
--- a/app/lib/analyzer/backend.dart
+++ b/app/lib/analyzer/backend.dart
@@ -65,9 +65,11 @@ class AnalysisBackend {
 
   /// Stores the analysis, and either creates or updates its parent
   /// [PackageAnalysis] and [PackageVersionAnalysis] records.
-  Future storeAnalysis(Analysis analysis) async {
+  ///
+  /// Returns whether an analysis race was detected.
+  Future<bool> storeAnalysis(Analysis analysis) {
     // update package and version too
-    await db.withTransaction((Transaction tx) async {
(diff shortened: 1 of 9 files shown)


app/lib/analyzer/backend.dart
```diff
@@ -88,12 +90,15 @@ class AnalysisBackend {
...
 
+      final DateTime prevTimestamp = version?.analysisTimestamp;
       if (version == null) {
         version = new PackageVersionAnalysis.fromAnalysis(analysis);
         inserts.add(version);
       } else if (version.updateWithLatest(analysis)) {
         inserts.add(version);
       }
+      final bool wasRace = prevTimestamp != null &&
+          version.analysisTimestamp.difference(prevTimestamp) < freshThreshold;
```
mkustermann: If it was a race, why would we insert this one? We already know there is an up-to-date one stored, right?

isoos: Good point, updated.

PR #295, opened 2017-08-17
Use JsonSerializable, update score logic

kevmoo:


kevmoo:
@isoos – I wanted to pull the scores out into a JSON file so it's easy to update them from a script. While I was at it, I converted the rest of the JSON logic to JsonSerializable.

app/lib/shared/search_service.g.dart
```diff
@@ -0,0 +1,87 @@
...
+        url: json['url'] as String,
+        package: json['package'] as String,
+        version: json['version'] as String,
+        devVersion: json['devVersion'] as String,
+        description: json['description'] as String,
+        lastUpdated: json['lastUpdated'] as String,
+        readme: json['readme'] as String,
+        detectedTypes:
+            (json['detectedTypes'] as List)?.map((e) => e as String)?.toList(),
+        popularity: json['popularity'] as double);
```
mkustermann: Side note: I'm not quite sure if the generator is doing this correctly: Often the wireformat json contains numbers as either as doubles or integers (if there is no fractional part). A `0 as double` will throw an exception!

So in order to work around this, it's better if the generator would generate: `(json['popularity'] as num).toDouble()` here. This ensures that the resulting value is a double, even if it was encoded as a whole integer.

app/lib/shared/mock_scores.dart
```diff
@@ -1,863 +1,9 @@
+library pub_dartlang_org.shared.mock_scores;
```
mkustermann: Theoretically there should be a copy right notice here!

app/lib/shared/mock_scores.json
```diff
@@ -0,0 +1,1182 @@
...
+ "bbs": 0,
+ "preprocessor": 0,
+ "highcharts_options": 0,
+ "cupid": 0,
+ "jaguar_facebook_client": 0,
+ "multi_reg_exp": 0,
+ "newton": 0,
+ "sync_socket": 0,
+ "hub": 0,
+ "tentacle_response_formatter": 0
```
mkustermann: Here, e.g., we have a json file with either integers or doubles!

app/lib/shared/search_service.g.dart
kevmoo: Great catch!

app/lib/shared/mock_scores.json
kevmoo: On second thought, I think dart2js handles (`0 as double`) quite fine.

kevmoo: ...in fact, it must – because I"ve used this code a great deal on web apps.

kevmoo:
@isoos – thoughts?

isoos:
lgtm pending the instructions in readme.md

app/lib/shared/mock_scores.json
mkustermann: We're not running on dart2js but rather on the VM, which has the proper number semantics.

```
 % cat test.dart
main() => print(1 as double);

 % dart test.dart
Unhandled exception:
type 'int' is not a subtype of type 'double' in type cast where
  int is from dart:core
  double is from dart:core

#0      Object._as (dart:core-patch/object_patch.dart:73)
#1      main (file:///usr/local/google/home/kustermann/repositories/vipunen-sdk/sdk/test.dart:1:19)
#2      _startIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:265)
#3      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:151)
```

kevmoo: I just realized this and verified. Sorry....digging...

app/lib/shared/search_service.g.dart
kevmoo: Fixed

kevmoo:
README work coming...

app/lib/shared/mock_scores.dart
kevmoo: ack

PR #302, opened 2017-08-22
Use analyzed data in search ranking

isoos:
-  Package health is 5% of the total score.
- Using bezier scoring for both popularity and health.

This is the first step of the chicken-and-egg problem of using bezier (relative) scores for package health (where you need to have some synchronization and get to know all the scores). The search index needs all the raw numbers to calculate the relative score, so it will fetch it from analyzer, and stores it in the snapshot.

Later on either the frontend or the analyzer will need to know about the score of a particular package, so most likely they will turn to the search index for the calculation.

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -423,13 +423,8 @@ Future<shelf.Response> packageVersionHandlerHtml(
       final Stopwatch serviceSw = new Stopwatch()..start();
       final AnalysisData analysisData = await analyzerClient.getAnalysisData(
           selectedVersion.package, selectedVersion.version);
-      analysisView = (analysisData != null &&
-              analysisData.analysisStatus != AnalysisStatus.aborted)
-          ? new AnalysisView(analysisData)
-          : null;
-      analysisTabContent = analysisView == null
-          ? null
-          : templateService.renderAnalysisTab(analysisView);
+      analysisView = new AnalysisView(analysisData);
+      analysisTabContent = templateService.renderAnalysisTab(analysisView);
       _packageAnalysisLatencyTracker.add(serviceSw.elapsed);
     }
 
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -6,6 +6,8 @@ import 'dart:async';
 import 'dart:math';
 
 import 'package:gcloud/service_scope.dart' as ss;
+// TODO: move scoring to a separate package or outside of pana/src
+import 'package:pana/src/scoring.dart' as scoring;
 
 import '../shared/search_service.dart';
 
@@ -23,6 +25,7 @@ class SimplePackageIndex implements PackageIndex {
   final TokenIndex _nameIndex = new TokenIndex();
   final TokenIndex _descrIndex = new TokenIndex();
   final TokenIndex _readmeIndex = new TokenIndex();
+  Map<String, double> _healthScores;
   DateTime _lastUpdated;
   bool _isReady = false;
 
@@ -36,6 +39,7 @@ class SimplePackageIndex implements PackageIndex {
     _nameIndex.add(doc.url, doc.package);
     _descrIndex.add(doc.url, compactDescription(doc.description));
     _readmeIndex.add(doc.url, compactReadme(doc.readme));
+    _clearScores();
   }
 
   @override
@@ -52,6 +56,7 @@ class SimplePackageIndex implements PackageIndex {
     _nameIndex.removeUrl(url);
     _descrIndex.removeUrl(url);
     _readmeIndex.removeUrl(url);
+    _clearScores();
   }
 
   @override
@@ -59,12 +64,14 @@ class SimplePackageIndex implements PackageIndex {
     final Map<String, double> total = <String, double>{};
     void addAll(Map<String, double> scores, double weight) {
       scores?.forEach((String url, double score) {
-        final double prev = total[url] ?? 0.0;
-        total[url] = prev + score * weight;
+        if (score != null) {
+          final double prev = total[url] ?? 0.0;
+          total[url] = prev + score * weight;
+        }
       });
     }
 
-    addAll(_nameIndex.search(query.text), 0.80);
+    addAll(_nameIndex.search(query.text), 0.75);
     addAll(_descrIndex.search(query.text), 0.10);
     addAll(_readmeIndex.search(query.text), 0.05);
 
@@ -73,11 +80,8 @@ class SimplePackageIndex implements PackageIndex {
       addAll(_nameIndex.search(query.packagePrefix), 0.8);
     }
 
-    final Map<String, double> popularityScores = new Map.fromIterable(
-      total.keys,
-      value: (String url) => _documents[url].popularity * 100,
-    );
-    addAll(popularityScores, 0.05);
+    addAll(getHealthScore(total.keys), 0.05);
+    addAll(getPopularityScore(total.keys), 0.05);
 
     List<PackageScore> results = <PackageScore>[];
     for (String url in total.keys) {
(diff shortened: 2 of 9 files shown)


app/lib/frontend/handlers.dart
```diff
@@ -422,13 +422,10 @@ Future<shelf.Response> packageVersionHandlerHtml(
       final Stopwatch serviceSw = new Stopwatch()..start();
       final AnalysisData analysisData = await analyzerClient.getAnalysisData(
           selectedVersion.package, selectedVersion.version);
-      final AnalysisView analysisView = (analysisData != null &&
-              analysisData.analysisStatus != AnalysisStatus.aborted)
-          ? new AnalysisView(analysisData)
+      final AnalysisView analysisView = new AnalysisView(analysisData);
+      analysisTabContent = analysisView.hasAnalysisData
+          ? templateService.renderAnalysisTab(analysisView)
```
mkustermann: Since `renderAnalysisTab` returns `null` now, you can remove the condition here.

app/lib/shared/analyzer_client.dart
```diff
@@ -65,4 +86,43 @@ class AnalysisView {
...
+  double get health {
+    if (_data == null) return 0.0; // missing analysis
+    if (!hasAnalysisData) return -1.0; // aborted analysis
+
+    // Starting out with max.
+    double score = 1.0;
+
+    // Penalty for each major issue.
+    if (_summary.issues != null && _summary.issues.isNotEmpty) {
+      score -= _summary.issues.length * 0.2;
```
mkustermann: What is a "major issue"?

Can an `analyzer_options.yaml` inside the package disable hints/warnings?

app/lib/shared/analyzer_client.dart
```diff
@@ -65,4 +86,43 @@ class AnalysisView {
...
+
+    if (_summary.dartFiles != null) {
+      for (DartFileSummary dfs in _summary.dartFiles.values) {
+        // penalties only for files inside the library
+        if (!dfs.isInLib) continue;
+
+        // penalty for formatting
+        if (!dfs.isFormatted != true) {
+          score -= 0.001;
+        }
```
mkustermann: Since all of the penalities are O(number-of-files in lib/) it seems larger packages will get penalized more, even if they have less issues/kloc ?

app/lib/shared/analyzer_client.dart
```diff
@@ -65,4 +86,43 @@ class AnalysisView {
...
+
+        // penalty for code structure
+        if (dfs.analyzerItems != null) {
```
mkustermann: What is an "analyzer item"? What does that have to do with "code structure"?

app/lib/shared/analyzer_client.dart
```diff
@@ -65,4 +86,43 @@ class AnalysisView {
     list.sort();
     return list;
   }
+
+  /// Returns the raw health score between -1.0 and 1.0
+  /// These need to be normalized for search and frontend purposes.
+  double get health {
```
mkustermann: Since this is a computation on the immutable analysis result, why are we not caching it somewhere? Seems like we should avoid re-computing this.

app/lib/shared/analyzer_client.dart
```diff
@@ -53,8 +54,28 @@ class AnalysisView {
...
+          data.analysisContent != null) {
+        summary = new Summary.fromJson(data.analysisContent);
+      }
+    } catch (e, st) {
+      // don't block on faulty serialization
+      _logger.warning(
+          'Unable to read analysis content for package: '
+          '${data.packageName} ${data.packageVersion}',
+          e,
+          st);
```
mkustermann: Since you put this try/catch in here, I assume you hit a case where we were crashing here? Why was that the case?

app/lib/search/index_simple.dart
```diff
@@ -141,6 +146,43 @@ class SimplePackageIndex implements PackageIndex {
...
+    _healthScores = <String, double>{};
+    _popularityScores = <String, double>{};
+
+    final healthScorer =
+        new scoring.Summary(_documents.values.map((pd) => pd.health ?? 0.0));
+    final popularityScorer = new scoring.Summary(
+        _documents.values.map((pd) => pd.popularity ?? 0.0));
+    for (PackageDocument doc in _documents.values) {
+      if (doc.health != null) {
+        _healthScores[doc.url] = healthScorer.bezierScore(doc.health) * 100.0;
```
mkustermann: What is a `bezierScore` and why can't we use **the already normalized** scores, which lie between `[-1.0, 1.0]`?

Either a package is healthy or not healthy, but it's health shouldn't depend on some complicated statistics of all other packages.

app/lib/search/index_simple.dart
```diff
@@ -141,6 +146,43 @@ class SimplePackageIndex implements PackageIndex {
...
+      }
+      if (doc.popularity != null) {
+        _popularityScores[doc.url] =
+            popularityScorer.bezierScore(doc.popularity) * 100.0;
```
mkustermann: Here I would assume that our bigquery batch job gives us already normalized popularity metric, something like `#of-package-X-downloads / #of-total-downloads` (or something more complicated), scaled to an appropriate interval, like `[0, 1]`.

app/lib/search/index_simple.dart
```diff
@@ -73,11 +81,8 @@ class SimplePackageIndex implements PackageIndex {
       addAll(_nameIndex.search(query.packagePrefix), 0.8);
     }
 
-    final Map<String, double> popularityScores = new Map.fromIterable(
-      total.keys,
-      value: (String url) => _documents[url].popularity * 100,
-    );
-    addAll(popularityScores, 0.05);
+    addAll(getHealthScore(total.keys), 0.05);
+    addAll(getPopularityScore(total.keys), 0.05);
```
mkustermann: I would assume that the popularity has a higher influence than 5%

app/lib/shared/analyzer_client.dart
isoos: We will migrate every non-trivial calculation from the `AnalysisView` to `pana` once we are happy with its use. However as this is entirely experimental and we don't know yet how we are going to use it, doing these in `pana` first would introduce a much bigger development roundtrip.

app/lib/shared/analyzer_client.dart
isoos: There is a chance that a new `analyzer` instance with a new `pana` version will produce a format that we don't parse for whatever reason. I'm planning to introduce a `panaVersion` parameter in the `frontend`->`analyzer` request so that if the `frontend` is on older, it still gets the older version.

app/lib/shared/analyzer_client.dart
isoos: major issue: `dartfmt`, `dartanalyzer` or `pub` finishing with non-zero exit code or some unexpected error while running their output.

`analyzer_options.yaml` could disable hints/warnings, that is true, but that goes to the minor issues down below. I'm not sure whether we want to override `analyzer_options.yaml`. Moved discussion to https://github.com/dart-lang/pana/issues/63

app/lib/shared/analyzer_client.dart
isoos: That is true, and I have no idea yet how we want to address that. Both filesize-based normalization and ignoring it altogether has its pros and cons. Luckily this is easy to change later if we make up our mind to do a different thing.

app/lib/shared/analyzer_client.dart
isoos: This is the stored output of `dartanalyzer` warnings or errors, each item separately.

app/lib/search/index_simple.dart
isoos: How much would you attribute to it, and which part of the score would you decrease?

app/lib/search/index_simple.dart
isoos: @kevmoo has started the score summary to do similar thing to npmjs (at least at that time npmjs was doing it, I'm not able to point to it right now). Kevin, could you please expand on it?

app/lib/search/index_simple.dart
isoos: Yeah, popularity may already use it, I'm going to remove this part.

app/lib/shared/analyzer_client.dart
mkustermann: Consider changing the comment to explain what `_summary.issues` is and refer to the pana github issue.

app/lib/shared/analyzer_client.dart
mkustermann: Please file a pana issue then and refer to the bug in this code.

app/lib/shared/analyzer_client.dart
mkustermann: Consider adding a comment then, so one can easily see that this is a place which might need a change.

app/lib/shared/analyzer_client.dart
mkustermann: Maybe rephrase the comment to say exactly that `Penalty for analyzer errors/warnings.`

app/lib/search/index_simple.dart
isoos: I agree, but my quick experiments showed that it is easy to screw up the balance, e.g. the test case for "chrome socket" will put the `chrome_net` package as last if I increase the popularity arbitrarily. Let's go with the current low value for now, and I'll experiment with the fine-tuning in a separate PR (with probably more test cases).

mkustermann:
@isoos Another timing out test. Maybe we need another increase in timeout for this upload-too-big test.

mkustermann:
Pulling it now. Filed https://github.com/dart-lang/pub-dartlang-dart/issues/305 for discussing this artificial bias in the health metric.

Issue #275, opened 2017-08-02
Check service calls, whether they are coming from the frontend.


isoos:
p2, not urgent, just to keep track of it. At the moment the open API helps to debug issues.

isoos:
It looks like that with the current setup (accessing services through their public URL) we don't have any appengine header that we could use to automatically restrict access from outside.

Headers from HTTP access through frontend:
```
x-cloud-trace-context = 7ed48e9c4e984bfba5b18f883d45838b/16292583403674841989
x-appengine-city = ?
content-length = 0
host = search-dot-dartlang-pub-dev.appspot.com
via = 1.1 google
x-forwarded-proto = https
x-appengine-citylatlong = 0.000000,0.000000
x-forwarded-for = 35.188.121.74, 209.85.147.153
x-appengine-country = US
x-appengine-region = ?
```

Headers from HTTP access through public URL:

```
x-cloud-trace-context = 1368b3a4357604e6ecb2293d5a2c5309/17388231850812814525
user-agent = Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.90 Safari/537.36
x-appengine-city = budapest
accept-encoding = gzip, deflate
x-forwarded-proto = http
x-appengine-country = HU
accept = text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
accept-language = en-US,en;q=0.8,hu;q=0.6
host = search-dot-dartlang-pub-dev.appspot.com
via = 1.1 google
x-appengine-citylatlong = 47.497912,19.040235
x-forwarded-for = 176.63.27.42, 172.217.18.84
upgrade-insecure-requests = 1
x-appengine-region = bu
```

@mkustermann: are you aware of any header that we could set and won't be stripped during the inside-appengine calls?

Otherwise would need a more elaborate token passing, and I'd do that only if this becomes a real issue. (Btw. only the notifications are affected).

mkustermann:
@isoos The question is not which headers are present by default, but what happens to extra `x-` or `x-appengine` headers we pass. The idea is that GAE will strip/mangle certain external header names whereas we would get unstripped ones from `frontend -> search/analysis`.

The reason that this is possible is that all services are in the same cloud project / compute engine virtual network. This means the instances should be able to talk to each other directly, bypassing the appengine header manging/stripping.

[...]

PR #304, opened 2017-08-22
Command line notification tool to prepare for shared secret in #275.

isoos:


app/bin/tools/notify_service.dart
```diff
@@ -0,0 +1,32 @@
...
+
+  final String service = args[0];
+  if (service == 'analyzer' && args.length == 3) {
+    notifyAnalyzer(args[1], args[2]);
+  } else if (service == 'search' && args.length == 2) {
+    notifySearch(args[1]);
+  } else {
+    _printHelp();
+  }
+  closeNotifyClient();
```
mkustermann: Whatever the http client is doing, it's probably not synchronous, so maybe we should change `notifyAnalyzer`  / `notifySearch` to return a `Future` and `await` here before we close the http client.

isoos: Yeah, probably better to do proper async await here, although my experiments showed to be working.

PR #313, opened 2017-08-25
Display license information without experimental flag

isoos:
I'm hoping that we'll get reassuring latency numbers after this change gets live, as we'll do a lot of analyzer lookup, and also collect their latencies.

- The 'analysis tab' remains behind experimental flag.
- The memcache behavior is the same: the analysis tab is not cached, otherwise the page gets cached.

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -386,10 +386,10 @@ Future<shelf.Response> packageVersionHandlerHtml(
         await backend.uiPackageCache.getUIPackagePage(packageName, versionName);
   }
 
-  final bool useService =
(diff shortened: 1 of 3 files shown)


mkustermann:
Just to clarify: The default mode of operation will be whenever we render a new package page, we'll ask the analyzer service for it's results for the package. This will increase the latency (also because the analysis service is not using memcache but always hitting the datastore).

The reason why it's not an issue is, that we cache it and serve the cached version for a while, right?

The caching is only disabled when, in addition to the above work, we also render the analysis tab, and this is guarded by the flag.

isoos:
Yeah, that's about right.

PR #314, opened 2017-08-25
Update Angular link after 4.0 is published.

isoos:
Note: Merge and publish after 4.0 is published.


kevmoo:
I will handle the merge once we land pkg/angular v4

PR #319, opened 2017-08-28
Refactor notification handler and put it in the service scope.

isoos:
This may also improve the issues with the upload test's timeouts, because it no longer initializes the http client.

### DIFF
--- a/app/lib/shared/notification.dart
+++ b/app/lib/shared/notification.dart
@@ -1,0 +1,65 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:gcloud/service_scope.dart' as ss;
+import 'package:http/http.dart' as http;
+import 'package:logging/logging.dart';
+
+import 'configuration.dart';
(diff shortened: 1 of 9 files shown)


app/lib/shared/notification.dart
```diff
@@ -0,0 +1,65 @@
...
+      _logger.info('Environment was not initialized: $e');
+    }
+  }
+
+  Future _doNotify(String uri) async {
+    // Don't block on the notification request, and don't fail even if there was
+    // an error.
+    try {
+      final response =
+          await _client.post(uri, headers: await prepareNotificationHeaders());
```
kevmoo: Doesn't this block?

app/test/frontend/backend_test_utils.dart
```diff
@@ -366,3 +367,20 @@ Future withTestPackage(Future func(List<int> tarball),
     return func(bytes);
   });
 }
+
+class NotificationClientMock implements NotificationClient {
+  List<String> notificationLog = [];
```
kevmoo: final?

app/lib/shared/notification.dart
isoos: It does. The comment a few lines above is outdated, and removing it. (We actually do no block on it, because we don't await the `notifyAnalyzer` call).

PR #320, opened 2017-08-29
Refactor and merge datastore head task pollers.

isoos:
The benefit is slim right now, but later on the `dartdoc` service should be able to use the shared one right out of the box.

### DIFF
--- a/app/lib/analyzer/task_sources.dart
+++ b/app/lib/analyzer/task_sources.dart
@@ -9,6 +9,7 @@ import 'package:logging/logging.dart';
 
 import '../frontend/models.dart';
 import '../shared/task_scheduler.dart';
+import '../shared/task_sources.dart';
 import '../shared/utils.dart';
 
 import 'models.dart';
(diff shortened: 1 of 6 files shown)


app/lib/shared/task_sources.dart
```diff
@@ -0,0 +1,101 @@
...
+    }
+    int count = 0;
+    await for (Package p in q.run()) {
+      final task = new Task(p.name, p.latestVersion ?? p.latestDevVersion);
+      if (await shouldYieldTask(task)) {
+        count++;
+        yield task;
+      }
+    }
+    await reportTurnDone(count);
```
mkustermann: Move this statement to the call site.

app/lib/analyzer/task_sources.dart
```diff
@@ -18,43 +19,20 @@ final Logger _logger = new Logger('pub.analyzer.source');
...
-          if (items.first == null) {
-            yield new Task(pv.package, pv.version);
-          }
-        }
-      } catch (e, st) {
-        _logger.severe('Error polling head.', e, st);
-      }
-      await new Future.delayed(const Duration(minutes: 1));
-    }
+  Future<bool> shouldYieldTask(Task task) async {
```
mkustermann: Normally I would prefer separating these concerns via composition rather than inheritance, e.g. `new FilteredTaskSource(new DatastoreHeadTaskSource())`.

app/lib/search/updater.dart
```diff
@@ -4,35 +4,26 @@
...
-      await for (String package
-          in searchBackend.listPackages(updatedAfter: _lastTs)) {
-        count++;
-        yield new Task(package, null);
-      }
-      _batchIndexUpdater.reportScanCount(count);
-      _lastTs = now.subtract(const Duration(minutes: 10));
-      await new Future.delayed(new Duration(minutes: 30));
-    }
+  Future reportTurnDone(int count) async {
```
mkustermann: The name does not seem to be very descriptive, maybe `dbScanComplete` or something like that.

isoos:
> Normally I would prefer separating these concerns via composition rather than inheritance, e.g. `new FilteredTaskSource(new DatastoreHeadTaskSource())`.

What would you think about `shouldYieldTask` and `dbScanComplete` to become `Function`s passed in the `DatastoreVersionsHeadTaskSource` class?

app/lib/shared/task_sources.dart
isoos: Do you mean that counting the Stream's length at one level up? To make things less complicated, I think we could operate without the count.

mkustermann:
> > Normally I would prefer separating these concerns via composition rather than inheritance, e.g. new 
> > FilteredTaskSource(new DatastoreHeadTaskSource()).

> What would you think about shouldYieldTask and dbScanComplete to become Functions passed in the > DatastoreVersionsHeadTaskSource class?

Let's just stick with this version for the sake of making progress. It was just a note/recommendation, since composed code is much easier to read/understand than using subclasses.

PR #321, opened 2017-08-29
Prepare dartdoc notification after new version is uploaded.

isoos:


app/lib/shared/notification.dart
```diff
@@ -32,6 +32,17 @@ class NotificationClient {
     }
   }
 
+  Future notifyDartdoc(String package, String version) async {
```
kevmoo: Now that we have three, might be an opportunity to DRY up the notifyX methods.

But no rush...

PR #322, opened 2017-08-29
Run dartdoc for the given task.

isoos:


app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -19,7 +28,68 @@ class DartdocRunner implements TaskRunner {
...
+      final pkgLocation =
+          await pubEnv.getLocation(task.package, version: task.version);
+      final pkgPath = pkgLocation.location;
+      await _runDartdoc(task, pkgPath);
+      await _writeMetadata(task, pkgPath);
+      // TODO: upload doc/api to the appropriate bucket
+    } finally {
+      await tempDir.delete(recursive: true);
+    }
+    return false; // no race detection
```
kevmoo: explain the comment?

isoos: The method should return true if there were a race of processing detected (e.g. in `analyzer`), false otherwise. Not sure how we want to comment on it in such case.

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -19,7 +28,68 @@ class DartdocRunner implements TaskRunner {
...
-    return false; // no race detected
+    final tempDir =
+        await Directory.systemTemp.createTemp('pub-dartlang-dartdoc');
+    final pubCacheDir = await tempDir.resolveSymbolicLinks();
+
+    final pubEnv = new PubEnvironment(pubCacheDir: pubCacheDir);
+
+    try {
+      final pkgLocation =
+          await pubEnv.getLocation(task.package, version: task.version);
```
mkustermann: Does this end up downloading from  the proper `https://pub.dartlang.org/packages/<name>/versions/<version>.tar.gz` url?

(so we don't have the memcache issue)

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -19,7 +28,68 @@ class DartdocRunner implements TaskRunner {
...
+      final pkgPath = pkgLocation.location;
+      await _runDartdoc(task, pkgPath);
+      await _writeMetadata(task, pkgPath);
+      // TODO: upload doc/api to the appropriate bucket
```
mkustermann: Ideally the bucket organization would use a scheme like `gs://<bucket-name>/<dartdoc-version>/<package>/<version>/{index.html,...}` here.

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -19,7 +28,68 @@ class DartdocRunner implements TaskRunner {
...
+    final pubCacheDir = await tempDir.resolveSymbolicLinks();
+
+    final pubEnv = new PubEnvironment(pubCacheDir: pubCacheDir);
+
+    try {
+      final pkgLocation =
+          await pubEnv.getLocation(task.package, version: task.version);
+      final pkgPath = pkgLocation.location;
+      await _runDartdoc(task, pkgPath);
+      await _writeMetadata(task, pkgPath);
```
mkustermann: If `_runDartdoc` failed, this call will fail, since the `doc` folder is not there.

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -19,7 +28,68 @@ class DartdocRunner implements TaskRunner {
 
   @override
   Future<bool> runTask(Task task) async {
-    // TODO: implement doc generation and upload
-    return false; // no race detected
+    final tempDir =
+        await Directory.systemTemp.createTemp('pub-dartlang-dartdoc');
+    final pubCacheDir = await tempDir.resolveSymbolicLinks();
```
mkustermann: Ideally we would reuse a pub cache, instead of just re-downloading all transitive dependencies for every single package we generate documentation for.

What is this `PubEnvironment` stuff doing?

Is `pubCacheDir` the real `PUB_CACHE` or is it a directory where this particular (package, version) tarball will be downloaded to?

app/lib/dartdoc/dartdoc_runner.dart
isoos: Unfortunately `pub` does a sanity check before doing that:
https://github.com/dart-lang/pub/blob/master/lib/src/command/cache_add.dart#L57
https://github.com/dart-lang/pub/blob/master/lib/src/source/hosted.dart#L130

Btw. this affects `pana` the very same way. /cc @kevmoo

app/lib/dartdoc/dartdoc_runner.dart
isoos: We have had an issue with the `analyzer` service early on, when I've left out the temp dir creation and cleanup: the packages fill up the disk quickly.

In this case `pubCacheDir` is a clean environment, initialized and clear for every `dartdoc` run. However, `pub cache` *does not* download the transitive dependencies, only the specified package, so it is fine to use it a a clean env.

`PubEnvironment` encapsulates the various executables, commands and environment variables for `pana`. It felt better to reuse it here, than to reinvent everything.

app/lib/dartdoc/dartdoc_runner.dart
isoos: `TaskScheduler` has a proper error handling, so we don't really need to worry about handling each of these errors separately. In fact, not having a `doc` directory helps us to get an exception without explicitly checking for it (and then throwing ourselves).

mkustermann: It's generally not advisable running commands / steps if they have assumptions about previous commands without ensuring that these previous commands completed successfully.

That's precisely why shell scripts often use `set -e` to abort after the first failed command.

Just assume that the package author already had a `doc` folder, then we might accidentally upload that, incorrect doc folder, to cloud storage and claim everything is fine, even though we didn't generate proper docs.

I'll add a comment further down about it.

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -19,7 +28,68 @@ class DartdocRunner implements TaskRunner {
...
+  }
+
+  Future _runDartdoc(Task task, String pkgPath) async {
+    final pr = await Process.run('dartdoc', [], workingDirectory: pkgPath);
+    if (pr.exitCode != 0) {
+      _logger.severe('Error while running dartdoc for $task.\n'
+          'exitCode: ${pr.exitCode}\n'
+          'stdout: ${pr.stdout}\n'
+          'stderr: ${pr.stderr}\n');
+    }
```
mkustermann: Please throw an exception or give a boolean to the caller to indicate whether it was successful. If we fail to generate dartdocs, we should not try running gcs uploading code!

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -19,7 +28,68 @@ class DartdocRunner implements TaskRunner {
...
+
+  Future<String> _getDartdocVersion() async {
+    if (_cachedDartdocVersion != null) return _cachedDartdocVersion;
+    final pr = await Process.run('dartdoc', ['--version']);
+    if (pr.exitCode != 0) {
+      _logger.severe('Unable to detect dartdoc version\n'
+          'exitCode: ${pr.exitCode}\n'
+          'stdout: ${pr.stdout}\n'
+          'stderr: ${pr.stderr}\n');
+      return null;
```
mkustermann: Ditto, if we can't determine the version, we should not continue.

isoos:
PTAL, I think I've addressed most of the concerns (dependencies are fetched, separate directory for the dartdoc content, throwing on dartdoc failures).

Postponed:
- direct download (some `pana` support is required for `analyzer` too, want to do that the same time)
- upload (btw. I'm not sure that the bucket file path needs to have the dartdoc version, a direct mirror of the dartdocs.org dir structure could be more beneficial)

mkustermann:
> upload (btw. I'm not sure that the bucket file path needs to have the dartdoc version, a direct mirror of the 
> dartdocs.org dir structure could be more beneficial)

The normal dartdocs URL space could just redirect to the active dartdoc-version GCS URLs.

There are multiple advantages of doing this: If we roll a new dartdoc version
  * it could change the entire look & feel:  So we would have a time period where some of the generated docs have the old style and some new, it's not the best user experience
  * we avoid needing to delete old files & create new files (and have short time where there are no docs)
  * the new dartdoc version might have a bug and fail on some percentage of packages, then we can easily go back to the old available docs

mkustermann:
> direct download (some pana support is required for analyzer too, want to do that the same time)

I've filed https://github.com/dart-lang/pana/issues/78, so we don't forget about it.

PR #328, opened 2017-09-01
Reduce backend load of re-indexing fresh data

isoos:
Since search is using `analyzer`, it is no longer super-cheap to re-index packages, better to restrict too frequent updates.

### DIFF
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -33,6 +33,21 @@ class SimplePackageIndex implements PackageIndex {
   bool get isReady => _isReady;
 
   @override
+  Future<bool> containsPackage(String package,
+      {String version, Duration maxAge}) async {
+    final String url = pubUrlOfPackage(package);
+    final PackageDocument doc = _documents[url];
+    if (doc == null) return false;
+    if (version != null && doc.version != version) return false;
+    if (maxAge != null &&
+        (doc.timestamp == null ||
+            new DateTime.now().toUtc().difference(doc.timestamp) > maxAge)) {
+      return false;
+    }
+    return true;
+  }
+
+  @override
   Future add(PackageDocument doc) async {
     await removeUrl(doc.url);
     _documents[doc.url] = doc;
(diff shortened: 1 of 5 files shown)


app/lib/shared/search_service.g.dart
```diff
@@ -18,7 +18,10 @@ PackageDocument _$PackageDocumentFromJson(Map<String, dynamic> json) =>
         detectedTypes:
             (json['detectedTypes'] as List)?.map((e) => e as String)?.toList(),
         health: (json['health'] as num)?.toDouble(),
-        popularity: (json['popularity'] as num)?.toDouble());
+        popularity: (json['popularity'] as num)?.toDouble(),
+        timestamp: json['timestamp'] == null
```
kevmoo: isn't json_seriailazable nice? 😄

app/lib/search/index_simple.dart
```diff
@@ -33,6 +33,21 @@ class SimplePackageIndex implements PackageIndex {
...
   @override
+  Future<bool> containsPackage(String package,
+      {String version, Duration maxAge}) async {
+    final String url = pubUrlOfPackage(package);
+    final PackageDocument doc = _documents[url];
+    if (doc == null) return false;
+    if (version != null && doc.version != version) return false;
+    if (maxAge != null &&
+        (doc.timestamp == null ||
+            new DateTime.now().difference(doc.timestamp) > maxAge)) {
```
mkustermann: Is `doc.timestamp` utc or not? We should use the same time zone.

app/lib/search/index_simple.dart
```diff
@@ -33,6 +33,21 @@ class SimplePackageIndex implements PackageIndex {
   bool get isReady => _isReady;
 
   @override
+  Future<bool> containsPackage(String package,
+      {String version, Duration maxAge}) async {
+    final String url = pubUrlOfPackage(package);
+    final PackageDocument doc = _documents[url];
+    if (doc == null) return false;
+    if (version != null && doc.version != version) return false;
```
mkustermann: Since `_documents[url]` seems to return only a single `PackageDocument` I assume if this method is not called with the latest version, it will therefore always return false. Is that the case? Does the caller guarantee it's always calling with the latest `{version: ...}`?

isoos: I wouldn't call it guarantee, but yeah, the notification client only passes the package name and no version, while the task source will get only the latest versions.

app/lib/search/index_simple.dart
isoos: good catch, fixing

PR #330, opened 2017-09-05
Index package dependencies and trigger re-analysis of packages

isoos:
The name `dependee` and the REST URL for it is a bit weird for me, but I couldn't come up with a better name - suggestions welcome :)

In an early version I've cached the dependees, but it is cheap to re-compute and we don't need to maintain another cache invalidation there.

### DIFF
--- a/app/lib/analyzer/backend.dart
+++ b/app/lib/analyzer/backend.dart
@@ -221,6 +221,12 @@ class AnalysisBackend {
       }
     }
   }
+
+  Future<String> getLatestStableVersion(String package) async {
+    final Package p =
+        (await db.lookup([db.emptyKey.append(Package, id: package)])).single;
+    return p?.latestVersion ?? p?.latestDevVersion;
+  }
 }
 
 class BackendAnalysisStatus {
--- a/app/lib/analyzer/pana_runner.dart
+++ b/app/lib/analyzer/pana_runner.dart
@@ -11,6 +11,7 @@ import 'package:pana/pana.dart';
 import '../shared/analyzer_service.dart';
 import '../shared/configuration.dart';
 import '../shared/notification.dart';
+import '../shared/search_client.dart';
 import '../shared/task_scheduler.dart' show Task, TaskRunner;
 
 import 'backend.dart';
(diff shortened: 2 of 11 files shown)


mkustermann:
We already have a `search` -> `analyzer` dependency. This adds a `analyzer` -> `search` dependency. We should avoid circular dependencies in our services as much as possible!

isoos:
If we were to separate out the search index (update+query) from the search index updating service (which queries datastore and analysis data), it would remove the circular dependency, but I see a lot of work with that, with very little benefit for now. Any suggestion how to do it?

mkustermann:
> If we were to separate out the search index (update+query) from the search index updating service 

Search functionality has nothing to do with the information you need, namely package dependency graphs. Search should be just about indexing text, ranking and serving queries.

The analyzer service needs to know more about package dependency graphs, since it fetches dependencies and runs the analyzer on the transitive graph.

> Any suggestion how to do it?

What is the concrete problem you are trying to solve? Is this in order to avoid polling?

isoos:
> Search should be just about indexing text, ranking and serving queries.

Well, it needs to know about and filter on platform, rank based on quality, popularity, and further down the line it should also suggest similar packages, or list packages from the same author. I doubt that it will remain simple.

> What is the concrete problem you are trying to solve? Is this in order to avoid polling?

As you can see, search already has the relevant data from the datastore, and has a place for caching (and cache invalidation). 

If analyzer would want to have the same information, it would need to duplicate the same thing: (1) reading through every package analysis and check whether the package in question is a dependency. (2) Build an efficient cache around it and make sure it is invalidated properly, as the new analysis or a new version appears.

Another alternative is to store more of these calculated values in the Datastore, along or parallel to the Package object (hiding the circularity in the database access).

isoos:
Gave it some thought and also discussed some of it with Kevin. I'm going to drop this PR, and think on how we can change all of our services from pull-based to push-based (maybe with the help of the new memcache or some lightweight Datastore synchronization). With that, `search` should become a "dumb" store of the index, without connecting to `analyzer`, while `analyzer` could (a) push the dependency information to search (b) query the same information if needed (c) trigger re-analysis when needed.

kevmoo:
@isoos – should close?

PR #335, opened 2017-09-06
Handle null window in DatastoreVersionsHeadTaskSource ctor

kevmoo:


kevmoo:
Hit this deploying dev, @isoos – will TBR it

app/lib/shared/task_sources.dart
```diff
@@ -42,8 +42,9 @@ class DatastoreVersionsHeadTaskSource implements TaskSource {
       : _window = window ?? _defaultWindow,
         _sleep = sleep ?? _defaultSleep,
         _onlyLatest = onlyLatest,
-        _lastTs =
-            skipHistory ? new DateTime.now().toUtc().subtract(window) : null;
```
isoos: ah, yeah, this should have been _window instead of window

PR #339, opened 2017-09-06
Fix some strong-mode hints

kevmoo:


kevmoo:
Thoughts, @mkustermann ?

app/lib/frontend/handlers.dart
```diff
@@ -81,6 +69,20 @@ Future<shelf.Response> appHandler(
   }
 }
 
+final _handlers = new Map<String, shelf.Handler>.unmodifiable({
```
mkustermann: Why `new Map.unmodifiable()` instead of `const <String, shelf.Handler> {...}`? Static tearoffs can be used in constant expression afaik.

app/lib/shared/task_scheduler.dart
```diff
@@ -132,7 +132,7 @@ class PrioritizedStreamIterator<T> implements StreamIterator<T> {
 
     // Listen on the streams and put items into their own queues.
     for (int i = 0; i < sources.length; i++) {
-      final Stream source = sources[i];
+      final source = sources[i];
```
mkustermann: Personally I don't think removing types helps readability of the code. I understand that there are lots of cases where a type can be inferred, but that doesn't mean that the code is easier to read/understand.

So why not change this to `Stream<T>` instead?

kevmoo: > So why not change this to Stream<T> instead?

Just a matter of style. I don't mind inference. Less typing. Can go either way.

app/lib/frontend/handlers.dart
kevmoo: Hrm...will try...

kevmoo:
PTAL

PR #340, opened 2017-09-08
PackageMemcache: Enable namespacing for index

kevmoo:
And other cleanup

Helps with https://github.com/dart-lang/pub-dartlang-dart/issues/337


mkustermann:
The CL title says **enabling namespacing for index**? What does that mean? To me it seems like this is just renaming a few fields.

kevmoo:
The only behavior change here is prefixing the index key the same way the json and html package pages are namespaced.

mkustermann:
Ah, I see. It's almost impossible to notice due to the (unnecessary) renames/...

PR #343, opened 2017-09-12
Upgrade to pana 0.4

isoos:


app/lib/shared/analyzer_client.dart
```diff
@@ -91,53 +91,24 @@ class AnalysisView {
   }
 
   List<String> getDependencies() {
```
mkustermann: Just a nit: If these are transitive dependencies, a `getTransitiveDependencies` would make it explicit.

app/test/frontend/golden/analysis_0.4.0_http.json
```diff
@@ -0,0 +1,448 @@
+{
```
mkustermann: To see changes in these files (i.e. diff's) it would be better if the filename was not including the version. Then we could easily see the small difference in the json file with changing pana versions.

PR #344, opened 2017-09-12
Index and search the analysis's platform instead of the pubspec detectedType

isoos:
Both the frontend and the search service supports the notation for "package must run on X and must not run on Y":
- `?platform=X&platform=!Y` (for form-submissions)
- `?platforms=X,!Y` (for simplified URL where multiple values are not supported)

This will allow us to distinguish between the following cases:
- `?platforms=flutter`: show me packages that I can use on flutter
- `?platforms=flutter,web`: show me packages that I can use both on flutter and on the web
- `?platforms=flutter,!server,!web`: show me the packages that are flutter-only
- `?platforms=!flutter,!server,!web`: show me the packages that have platform conflicts

### DIFF
--- a/app/lib/shared/analyzer_client.dart
+++ b/app/lib/shared/analyzer_client.dart
@@ -82,6 +82,8 @@ class AnalysisView {
   DateTime get timestamp => _data.timestamp;
   AnalysisStatus get analysisStatus => _data.analysisStatus;
 
+  DartPlatform get platform => _summary?.platform;
+
   String get licenseText {
     final String text = _summary?.license?.toString();
     if (text == LicenseNames.unknown || text == LicenseNames.missing) {
--- a/app/lib/shared/platform.dart
+++ b/app/lib/shared/platform.dart
@@ -1,0 +1,89 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:pana/pana.dart';
+
+abstract class KnownPlatforms {
+  static const String flutter = PlatformNames.flutter;
+  static const String server = PlatformNames.server;
+  static const String web = PlatformNames.web;
+  static const List<String> all = const [flutter, server, web];
+
+  static bool isKnownPlatform(String platform) => all.contains(platform);
+}
+
+List<String> indexDartPlatform(DartPlatform platform) {
+  if (platform == null) {
+    return null;
+  }
+  if (platform.worksEverywhere) {
+    return KnownPlatforms.all;
+  }
+  if (platform.restrictedTo == null || platform.restrictedTo.isEmpty) {
+    return null;
+  }
+  return new List.from(platform.restrictedTo);
+}
+
+class PlatformPredicate {
+  final List<String> required;
+  final List<String> prohibited;
+
+  PlatformPredicate._(this.required, this.prohibited);
+
+  factory PlatformPredicate({List<String> required, List<String> prohibited}) {
+    List<String> reduce(List<String> platforms) {
+      final List<String> list = platforms
+          ?.where((p) => p != null && p.isNotEmpty)
+          ?.where(KnownPlatforms.isKnownPlatform)
+          ?.toList();
+      return (list != null && list.isNotEmpty) ? list : null;
+    }
+
+    return new PlatformPredicate._(reduce(required), reduce(prohibited));
+  }
+
+  factory PlatformPredicate.fromUri(Uri uri) {
+    final String pluralStr = uri.queryParameters['platforms'];
+    List<String> platforms;
+    if (pluralStr != null && pluralStr.isNotEmpty) {
+      platforms = pluralStr.split(',');
+    } else {
+      platforms = uri.queryParametersAll['platform'];
+    }
+    final List<String> required = <String>[];
+    final List<String> prohibited = <String>[];
+    platforms?.forEach((String p) {
+      if (p.startsWith('-') || p.startsWith('!')) {
+        prohibited.add(p.substring(1).trim());
+      } else {
+        required.add(p.trim());
+      }
+    });
+    return new PlatformPredicate(required: required, prohibited: prohibited);
+  }
+
+  bool get isNotEmpty => required != null || prohibited != null;
+
+  bool matches(List<String> platforms) {
+    if (required != null) {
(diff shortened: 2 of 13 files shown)


kevmoo:
Not sure the practical benefit of this. ... other than being able to see platform conflicts, I guess...

isoos:
> Not sure the practical benefit of this. ... other than being able to see platform conflicts, I guess...

Reasons to have the negative predicates:
- Flutter devs may be interested only in Flutter plugins
- web devs may be interested in web-only widgets

I can easily remove that part, but most of the PR is about changing from `detectedType` to `DartPlatform`, so we won't get this CL reduced in half...

kevmoo:
No worries. We could/should ponder positive searches for these cases.

Packages that depend on Flutter
Packages that depend on Angular

...etc etc

app/lib/shared/search_service.dart
```diff
@@ -73,46 +75,51 @@ class PackageDocument extends Object with _$PackageDocumentSerializerMixin {
 
 class PackageQuery {
   final String text;
-  final String type;
+  final PlatformPredicate platform;
```
mkustermann: To avoid confusion, we should name variables of type `PlatformPredicate` also `platformPredicate` (also at other places)

app/lib/shared/platform.dart
```diff
@@ -0,0 +1,89 @@
...
+
+import 'package:pana/pana.dart';
+
+abstract class KnownPlatforms {
+  static const String flutter = PlatformNames.flutter;
+  static const String server = PlatformNames.server;
+  static const String web = PlatformNames.web;
+  static const List<String> everywhere = const [flutter, server, web];
+
+  static bool isPredicate(String platform) => everywhere.contains(platform);
```
mkustermann: This doesn't test whether `platform` is predicate, but rather whether `platform` is a **known** platform.

Maybe rename this to `isKnownPlatform`.

app/lib/shared/analyzer_client.dart
```diff
@@ -82,6 +82,8 @@ class AnalysisView {
   DateTime get timestamp => _data.timestamp;
   AnalysisStatus get analysisStatus => _data.analysisStatus;
 
+  DartPlatform get platform => _summary?.platform;
```
mkustermann: Where is that `DartPlatform` coming from? I cannot see any class definition. Is this from the pana?

app/lib/shared/platform.dart
```diff
@@ -0,0 +1,89 @@
...
+  }
+  if (platform.restrictedTo == null || platform.restrictedTo.isEmpty) {
+    return null;
+  }
+  return new List.from(platform.restrictedTo);
+}
+
+class PlatformPredicate {
+  final List<String> required;
+  final List<String> prohibited;
```
mkustermann: Why do we have a `DartPlatform` class if we then represent a `platform` mostly as plain `String`s?

app/lib/shared/platform.dart
```diff
@@ -0,0 +1,89 @@
...
+  if (platform == null) {
+    return null;
+  }
+  if (platform.worksEverywhere) {
+    return KnownPlatforms.everywhere;
+  }
+  if (platform.restrictedTo == null || platform.restrictedTo.isEmpty) {
+    return null;
+  }
+  return new List.from(platform.restrictedTo);
```
mkustermann: Side note: `DartPlatform` seems also like not the best name, since it's not really an identifier but behaves more like a set/list.

app/lib/shared/platform.dart
```diff
@@ -0,0 +1,89 @@
...
+import 'package:pana/pana.dart';
+
+abstract class KnownPlatforms {
+  static const String flutter = PlatformNames.flutter;
+  static const String server = PlatformNames.server;
+  static const String web = PlatformNames.web;
+  static const List<String> everywhere = const [flutter, server, web];
+
+  static bool isPredicate(String platform) => everywhere.contains(platform);
+}
```
mkustermann: We have now at least 5 different types we use for two concepts (platform identifier, platform predicate):
```
PlatformNames, KnownPlatforms, PlatformPredicate, DartPlatform, String
```
e.g. Is it obvious what the difference between `PlatformNames.everywhere` and `KnownPlatforms.everywhere` is? To me not at all (of course I can find out, but it's not obvious)!

Also there a difference between a **platform name** and a **platform** ? Both seem to be just plain `String`s.

app/lib/shared/platform.dart
```diff
@@ -0,0 +1,89 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:pana/pana.dart';
+
+abstract class KnownPlatforms {
+  static const String flutter = PlatformNames.flutter;
+  static const String server = PlatformNames.server;
+  static const String web = PlatformNames.web;
+  static const List<String> everywhere = const [flutter, server, web];
```
mkustermann: Maybe use `all` instead of `everywhere` (I would only use `everywhere` together with a verb, e.g. `runsEverywhere`).

app/lib/shared/analyzer_client.dart
isoos: It is from `pana`: https://github.com/dart-lang/pana/blob/master/lib/src/platform.dart#L30

app/lib/shared/platform.dart
isoos: The platform name is simple, and it is a `String`. But their relation is complex, e.g. "runs of Flutter" can mean several things:
- requires Flutter
- runs only on Flutter
- runs on Flutter but also on other platforms.

`DartPlatform` (from pana) gives a DRY format of the above, e.g. a package either runs everywhere, or it is restricted to only a few platforms or a single one. However, the search use cases are more complex, hence the `PlatformPredicate`.

`KnownPlatforms` is a helper, so that the rest of the pub site doesn't need to depend on `pana` internal `PlatformNames` (which also has additional entries for the debug-centric `DartPlatform.description` field).

app/lib/shared/platform.dart
isoos: Yeah, I agree it is a crappy name, and I'll be happy to refactor it if there is a better one. Platform classification has way too many edge cases, so it either gets too simple (leaving out legitimate concerns), or somewhat confusing (like the current state). I've erred on the later, because I think the "which platform will this be running on" will be an important info (icons, search filters), and I'd rather have a bit more complexity in how we handle it, but get it correct from the user's point of view.

PR #345, opened 2017-09-12
Small review follow-ups

isoos:
- `getDependencies()` -> `getTransitiveDependencies()`
- removed pana version from test file to better track changes across future versions
- fixing string interpolation


app/lib/frontend/templates.dart
```diff
@@ -132,7 +132,7 @@ class TemplateService {
     }
 
     final List<Map> dependencies = analysis
-        .getDependencies()
+        .getTransientDependencies()
```
mkustermann: Wouldn't `transitive` be better than `transient`?

isoos: ah, my bad, good catch

PR #346, opened 2017-09-13
Model and backend for storing PackageDocuments in Datastore

isoos:


isoos:
I was a bit undecided whether this should be part of the `AnalysisBackend` or not, but decided against it. With the current separation, the `analyzer` service will depend on the `SearchBackend` class, but the `PackageSearch` will be masked from it.

isoos:
Waiting for @mkustermann to comment on the linked issue first.

PR #353, opened 2017-09-14
Render platform icons from analysis data

isoos:
Examples on staging:
- https://dartlang-pub-dev.appspot.com/packages/async
- https://dartlang-pub-dev.appspot.com/packages/build


kevmoo:
This will get rejiggered w/ the new design...but I'm okay having a place-holder.

PR #354, opened 2017-09-14
Re-add installing packages to Dockerfile

isoos:


kevmoo:
Why? Just curious...

isoos:
> Why? Just curious...

Flutter depends on unzip. Not sure about git, but it was there previously, so let's keep it.

kevmoo:
squash and merge as you like...

PR #357, opened 2017-09-15
Memcache for analysis data

isoos:
Improves a lot on the page build latency, and will be a must for accessing analysis data for search results.

### DIFF
--- a/app/lib/shared/analyzer_memcache.dart
+++ b/app/lib/shared/analyzer_memcache.dart
@@ -1,0 +1,62 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:gcloud/service_scope.dart' as ss;
+import 'package:logging/logging.dart';
+import 'package:memcache/memcache.dart';
+
+final Logger _logger = new Logger('pub.analyzer_memcache');
+
+/// Sets the analyzer memcache.
+void registerAnalyzerMemcache(AnalyzerMemcache value) =>
+    ss.register(#_analyzerMemcache, value);
+
+/// The active analyzer memcache.
+AnalyzerMemcache get analyzerMemcache => ss.lookup(#_analyzerMemcache);
+
+class AnalyzerMemcache {
+  final Memcache _memcache;
+  final String _prefix = 'dart_analyzer_api_';
+  final Duration _expiration = new Duration(minutes: 60);
+
+  AnalyzerMemcache(this._memcache);
+
+  Future<String> getContent(
+      String package, String version, String panaVersion) async {
+    try {
+      return await _memcache.get(_key(package, version, panaVersion));
+    } catch (e, st) {
+      _logger.severe('ERROR', e, st);
+      // ignore errors
+    }
+    _logger.fine('Couldn\'t find memcache entry for $package $version');
+    return null;
+  }
+
+  Future setContent(String package, String version, String panaVersion,
+      String content) async {
+    try {
+      await _memcache.set(_key(package, version, panaVersion), content,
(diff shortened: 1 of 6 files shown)


app/lib/shared/analyzer_memcache.dart
```diff
@@ -0,0 +1,62 @@
...
+void registerAnalyzerMemcache(AnalyzerMemcache value) =>
+    ss.register(#_analyzerMemcache, value);
+
+/// The active analyzer memcache.
+AnalyzerMemcache get analyzerMemcache => ss.lookup(#_analyzerMemcache);
+
+class AnalyzerMemcache {
+  final Memcache _memcache;
+  final String _prefix = 'dart_analyzer_api_';
+  final Duration _expiration = new Duration(minutes: 60);
```
mkustermann: Maybe we should keep this in sync with `AppEnginePackageMemcache`. I have reduced the timeout back then from 60 to 10 minutes after using flexible runtimes with a per-instance memcache.

Now that we have a shared memcache (again), we should increase the other timeout also to 60 minutes again. 

If everything works, then the memcache invalidation will take care of taking the value out, once a new upload happened.

app/lib/shared/analyzer_memcache.dart
```diff
@@ -0,0 +1,62 @@
...
+    try {
+      await _memcache.remove(_key(package, version, panaVersion));
+    } catch (e, st) {
+      _logger.warning(
+          'Couldn\'t remove memcache entry for $package $version', e, st);
+    }
+  }
+
+  String _key(String package, String version, String panaVersion) =>
+      '$_prefix/$package/$version/$panaVersion';
```
mkustermann: Not quite sure, but we might also want to include the flutter version in the key. wdyt?

app/lib/shared/analyzer_memcache.dart
isoos: Will do in a follow-up. I think more than half of the memcache-related code could be shared, I just wanted to put this out first.

app/lib/shared/analyzer_memcache.dart
isoos: Here only the panaVersion is important, because we want to have cheap protection against breaking changes in pana version upgrade. When flutter version changes, it does affect the analysis and will trigger a new one, but that is not breaking in the serialization format.

kevmoo:
So basically we're trying to "communicate" over memcache first, then use RPC?

I guess that should be faster, right?

isoos:
> So basically we're trying to "communicate" over memcache first, then use RPC?
> I guess that should be faster, right?

Yeah, working as expected.

mkustermann:
> I guess that should be faster, right?

Well, memcache latency might be similar to GCE -> GCE latency. But since we want to cache anyway **to avoid datastore latency**, we might as well do the caching on the analyzer client side (instead of analyzer server side).

isoos:
> Well, memcache latency might be similar to GCE -> GCE latency. But since we want to cache anyway to avoid datastore latency, we might as well do the caching on the analyzer client side (instead of analyzer server side).

As long as we limit the memory use of the cache, I believe a local cache will be OK. How much free memory do we have on the prod instances?

mkustermann:
> As long as we limit the memory use of the cache, I believe a local cache will be OK.

We could do that, but I think it's unnecessary. Memcache should be <= 1ms roundtrip.

> (kevmoo) So basically we're trying to "communicate" over memcache first, then use RPC?

What I ment is that we hit the network anyways, whether we talk to memcache or to the analysis service.

But

frontend -> memcache service  (which istvan implemented, in `AnalyzerClient`)

is faster than

frontend -> analyzer service -> memcache service  (which would be "ok" as well)
frontend -> analyzer service -> datastore service (slow)

kevmoo:
SGTM

On Tue, Sep 19, 2017 at 1:51 AM, Martin Kustermann <notifications@github.com

PR #359, opened 2017-09-18
Fetch icons data from analyzer for search results

isoos:
Demo: https://dartlang-pub-dev.appspot.com/search?q=game
The UI part is considered placeholder for now.

### DIFF
--- a/app/lib/frontend/search_service.dart
+++ b/app/lib/frontend/search_service.dart
@@ -14,6 +14,7 @@ import 'package:googleapis_auth/auth_io.dart' as auth;
 import 'package:logging/logging.dart';
 import 'package:http/http.dart' as http;
 
+import '../shared/analyzer_client.dart';
 import '../shared/search_client.dart';
 import '../shared/search_service.dart';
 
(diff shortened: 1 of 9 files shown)


app/lib/frontend/search_service.dart
```diff
@@ -85,14 +86,27 @@ Future<SearchResultPage> _loadResultForPackages(SearchQuery query,
...
+            (p) => analyzerClient.getAnalysisData(p.name, p.latestVersion)));
+
     final allVersions =
         await dbService.lookup([]..addAll(versionKeys)..addAll(devVersionKeys));
     final versions = allVersions.sublist(0, versionKeys.length);
     final devVersions = allVersions.sublist(versionKeys.length);
-    return new SearchResultPage(
-        query, totalCount, versions, devVersions, backend);
+
+    final List<AnalysisData> analysisDataList = await analysisBatchFuture;
```
mkustermann: We have to be quite careful here:
```dart
Future a = ...;
Future b = ...;
await b;
await a;
```
If during `await b` `a` completes with an error, there will be no handler registered for `a`. This will cause the error to be propagated to the current zone.

So usually it's better to do:

```dart
Future a =  ...;
// ... only synchronous code here
Future b = ...; 
// ... only synchronous code here
await Future.wait([a, b]);
```
Then one avoids propagating an error to the zone, since it's guaranteed that there is a registered callback on both futures, which will make `_loadResultForPackages` return with that error.

app/test/frontend/golden/flutter_packages_index_page2.html
```diff
@@ -99,7 +99,9 @@
   </thead>
   <tbody>
       <tr>
-        <td><img src="/static/img/flutter-logo-32x32.png" class="tablecol-icons-img" alt="Flutter package" title="Flutter package"/>
+        <td>  <div style="min-width: px">
```
mkustermann: As a TODO, once we have a proper CSS story, it might make sense to move that into the `tablecol-icons-img` css class.

also: Isn't there a number missing before `px`? Maybe avoid emitting it alltogether if there is no pixel with given.

app/lib/frontend/search_service.dart
isoos: good point, fix ongoing

app/test/frontend/golden/flutter_packages_index_page2.html
isoos: missing px: good catch, that is for the tables that are still using detectTypes. Fixed their template for now.

style vs css: Agree and noted, but without a TODO. I hope the new UI templates should already handle it, and will revisit on migration.

PR #361, opened 2017-09-18
Extracting memcache constants to a shared file.

isoos:


isoos:
No change in the functionality yet, let me know if you want me to do the expiration increase in the same PR.

PR #365, opened 2017-09-18
Simplify package memcache with shared impl

isoos:


isoos:
Removed a lot of logging, let me know if they were important and I'll add them back.

app/lib/shared/package_memcache.dart
```diff
@@ -33,105 +32,63 @@ abstract class UIPackageCache {
...
 
   @override
-  Future setUIPackagePage(String package, String version, String data) async {
-    _logger.info('memcache["$package"] setting to new rendered UI data');
-    return _ignoreErrors(_memcache.set(
-        _packageUIKey(package, version), UTF8.encode(data),
-        expiration: packageUiPageExpiration));
-  }
+  Future setUIPackagePage(String package, String version, String data) =>
+      _uiPage.setText(new _PVKey(package, version), data);
```
mkustermann: I would prefer if we would not introduce extra classes like `_PVKey`. They provide very little value and introduce extra overhead.

The `SimpleMemcache` doesn't need to be generic. The value it provides atm is auto-prefixing and auto-expiration.

app/lib/shared/memcache.dart
```diff
@@ -47,6 +47,24 @@ class SimpleMemcache<T> {
...
+    try {
+      await _memcache.set(_key(key), content, expiration: _expiration);
+    } catch (e, st) {
+      _logger.warning('Couldn\'t set memcache entry for $key', e, st);
+    }
+  }
+
   Future invalidate(T key) async {
     try {
       await _memcache.remove(_key(key));
```
mkustermann: I think you can move the invalidation logging here. It is beneficial to have the cache invalidation log messages and they happen almost never (only on new package uploads).

app/lib/shared/package_memcache.dart
isoos: It also masks the way how keys are generated, and through the type (if it is not `String`) it makes sure that the key generation is used consistently between `get*`, `set*` and `invalidate`.

If you want to remove the `_PVKey` we need to put the key creation somewhere else, potentially outside of the `SimpleMemcache`, because it may be 0-3 parameters that could go into the key.

Alternatively, I could make the API like the following, allowing the caller to decide how many of these params it will use
```
Future<String> getText({String package, String version, String context});
```
(`context` being the `panaVersion` for `SearchMemcache`).

I'm fine with either approach, which would you like to have?

mkustermann: Just instead of relying on `new _PVKey(a, b, c).toString()` `_pvKey(a, b, c)` is enough.

The extra "safety" by making it generic is dminishing small, because there are max 3 calls to it with the same combined key, namely get/set/invalidate.

Also others just use `String` as the generic argument. You could argue they should then suddenly use `SimpleMemcache<Uri>`. Then we might need to do URI parsing and toString just for that.

Just thinking this is overkill.

isoos: Yeah, good point with the uri. Changing it to String...

PR #366, opened 2017-09-19
Search order implementation with additional tests.

isoos:
There is one side-effect for the text-search: there is a clipping where we remove low score values from the results, and this happens earlier (before adding the quality and health scores). The result is fewer results, which have roughly the same score as before.

Some examples:
- ['game' order by health](https://20170919t153718-dot-search-dot-dartlang-pub-dev.appspot.com/search?q=game&order=health&indent=true)
- ['game' order by updated](https://20170919t153718-dot-search-dot-dartlang-pub-dev.appspot.com/search?q=game&order=updated&indent=true)
- ['game' order by text](https://20170919t153718-dot-search-dot-dartlang-pub-dev.appspot.com/search?q=game&order=text&indent=true)

### DIFF
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -71,66 +71,59 @@ class SimplePackageIndex implements PackageIndex {
 
   @override
   Future<PackageSearchResult> search(SearchQuery query) async {
-    final Map<String, double> total = <String, double>{};
-    void addAll(Map<String, double> scores, double weight) {
-      scores?.forEach((String url, double score) {
-        if (score != null) {
-          final double prev = total[url] ?? 0.0;
-          total[url] = prev + score * weight;
-        }
-      });
+    // do text matching
+    final Score textScore = _searchText(query.text, query.packagePrefix);
+
+    // The set of urls to filter on.
+    final Set<String> urls =
+        textScore?.getKeys()?.toSet() ?? _documents.keys.toSet();
+
+    // filter on package prefix
+    if (query.packagePrefix != null) {
+      urls.removeWhere(
+        (url) => !_documents[url]
+            .package
+            .toLowerCase()
+            .startsWith(query.packagePrefix.toLowerCase()),
+      );
     }
 
-    addAll(_nameIndex.search(query.text), 0.70);
-    addAll(_descrIndex.search(query.text), 0.10);
-    addAll(_readmeIndex.search(query.text), 0.05);
-
-    if ((query.text == null || query.text.isEmpty) &&
-        query.packagePrefix != null) {
-      addAll(_nameIndex.search(query.packagePrefix), 0.8);
+    // filter on platform
+    if (query.platformPredicate != null) {
+      urls.removeWhere(
+          (url) => !query.platformPredicate.matches(_documents[url].platforms));
     }
 
-    addAll(getHealthScore(total.keys), 0.05);
-    addAll(getPopularityScore(total.keys), 0.10);
+    // reduce text results if filter did remove an url
+    textScore?.removeWhere((key) => !urls.contains(key));
 
-    List<PackageScore> results = <PackageScore>[];
-    for (String url in total.keys) {
-      final PackageDocument doc = _documents[url];
-
-      // filter on platform
-      if (query.platformPredicate != null &&
-          !query.platformPredicate.matches(doc.platforms)) {
-        continue;
-      }
-
-      // filter on package prefix
-      if (query.packagePrefix != null &&
-          !doc.package
-              .toLowerCase()
-              .startsWith(query.packagePrefix.toLowerCase())) {
-        continue;
-      }
-
-      results.add(new PackageScore(
-        url: doc.url,
-        package: doc.package,
-        score: total[url],
-      ));
-    }
-
-    results.sort((a, b) => -a.score.compareTo(b.score));
-
-    // filter out the noise (maybe a single matching ngram)
-    if (results.isNotEmpty) {
-      final double bestScore = results.first.score;
-      final double scoreTreshold = bestScore / 25;
-      results.removeWhere((pr) => pr.score < scoreTreshold);
+    List<PackageScore> results;
+    switch (query.order ?? SearchOrder.overall) {
+      case SearchOrder.overall:
+        final Score overallScore = new Score()
+          ..addValues(textScore?.values, 0.85)
+          ..addValues(getPopularityScore(urls), 0.10)
+          ..addValues(getHealthScore(urls), 0.05);
+        results = _rankWithValues(overallScore.values);
+        break;
+      case SearchOrder.text:
+        results = _rankWithValues(textScore.values);
+        break;
+      case SearchOrder.updated:
+        results = _rankWithComparator(urls, _compareUpdated);
+        break;
+      case SearchOrder.popularity:
+        results = _rankWithValues(getPopularityScore(urls));
+        break;
+      case SearchOrder.health:
+        results = _rankWithValues(getHealthScore(urls));
+        break;
     }
 
     // bound by offset and limit
-    final int totalCount = min(maxSearchResults, results.length);
+    final int totalCount = results.length;
     if (query.offset != null && query.offset > 0) {
-      if (query.offset > totalCount) {
+      if (query.offset >= results.length) {
         results = <PackageScore>[];
       } else {
         results = results.sublist(query.offset);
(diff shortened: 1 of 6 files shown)


isoos:
I'm also planning to simplify the index in a subsequent PR: we could remove `url` as an id and use only `package`. When we introduce the Dart SDK API index, it is likely that it will get into a different structure than `PackageDocument`, so there is no reason to have it around now, and it will simplify some of the lookups in this current change.

app/lib/search/index_simple.dart
```diff
@@ -168,6 +161,80 @@ class SimplePackageIndex implements PackageIndex {
...
+    });
+  }
+
+  void removeWhere(bool keyCondition(String key)) {
+    final Set<String> keysToRemove = values.keys.where(keyCondition).toSet();
+    keysToRemove.forEach(values.remove);
+  }
+
+  void removeLowScores(double fraction) {
+    final double maxValue = values.values.fold(0.0, max);
```
mkustermann: I think this is a big weird: Basically if I search for "foobar" and none of the packages are relevant, though our character n-grams returns some low-score results, we find the max here, and someone passes like `fraction = 0.9` (meaning should be in top 10%) then we get results, even though none are relevant.

I think a better way of doing this is to always ensure that if a package is perfect match, then it's score needs to be `1.0` then you can just say

```dart
removeWhere((key) => values[key] < 0.9);
```

Then it's even possible to have optimizations which pushes this filter down to the things calling `addValues` so you can avoid work if you know already it can't pass the filter.

app/lib/shared/search_service.g.dart
```diff
@@ -83,6 +83,19 @@ abstract class _$PackageScoreSerializerMixin {
...
+    var val = <String, dynamic>{
+      'url': url,
+      'package': package,
+    };
+
+    void writeNotNull(String key, dynamic value) {
+      if (value != null) {
+        val[key] = value;
+      }
+    }
```
mkustermann: side note: not the most efficient code our "great json serializer source gen" is doing :-/

app/lib/search/index_simple.dart
```diff
@@ -168,6 +161,80 @@ class SimplePackageIndex implements PackageIndex {
...
+  Score _searchText(String text, String packagePrefix) {
+    Score textScore;
+    if (text != null && text.isNotEmpty) {
+      textScore = new Score()
+        ..addValues(_nameIndex.search(text), 0.82)
+        ..addValues(_descrIndex.search(text), 0.12)
+        ..addValues(_readmeIndex.search(text), 0.06);
+      textScore.removeLowScores(0.05);
+    } else if (packagePrefix != null) {
+      textScore = new Score()..addValues(_nameIndex.search(packagePrefix), 0.8);
```
mkustermann: Is this `_nameIndex.search(packagePrefix)` really searching for a **prefix** or does it match anywhere in the name?

If it doesn't search for a prefix, then we shouldn't use the name "prefix".

app/lib/search/index_simple.dart
```diff
@@ -168,6 +161,80 @@ class SimplePackageIndex implements PackageIndex {
...
+      return _compareUpdated(_documents[a.url], _documents[b.url]);
+    });
+    return list;
+  }
+
+  List<PackageScore> _flattenFromUrls(
+      Set<String> urls, int compare(PackageDocument a, PackageDocument b)) {
+    final List<PackageScore> list = urls
+        .map((url) =>
+            new PackageScore(url: url, package: _documents[url].package))
```
mkustermann: It's not clear why you are not adding a `, score: values[url]` here!

app/lib/search/index_simple.dart
```diff
@@ -71,66 +71,59 @@ class SimplePackageIndex implements PackageIndex {
...
-      addAll(_nameIndex.search(query.packagePrefix), 0.8);
+    // do text matching
+    final Score textScore = _searchText(query.text, query.packagePrefix);
+
+    // The set of urls to filter on.
+    final Set<String> urls =
+        textScore?.getKeys()?.toSet() ?? _documents.keys.toSet();
+
+    // filter on package prefix
+    if (query.packagePrefix != null) {
```
mkustermann: For future optimizations: There might be possibilities where one can avoid constructing the big maps in `Score` if we know only a small subset of keys will survive.

i.e. Instead of building up big datastructures and removing from them later, we could try filtering early on and construct smaller datastructures.

app/lib/search/index_simple.dart
```diff
@@ -71,66 +71,59 @@ class SimplePackageIndex implements PackageIndex {
...
+    textScore?.removeWhere((key) => !urls.contains(key));
+
+    List<PackageScore> results;
+    switch (query.order ?? SearchOrder.overall) {
+      case SearchOrder.overall:
+        final Score overallScore = new Score()
+          ..addValues(textScore?.values, 0.85)
+          ..addValues(getPopularityScore(urls), 0.10)
+          ..addValues(getHealthScore(urls), 0.05);
+        results = _flattenFromValues(overallScore.values);
```
mkustermann: Instead of 'flatten' we could use 'rank', because what it really is doing is ordering/ranking the results.

app/lib/search/index_simple.dart
```diff
@@ -71,66 +71,59 @@ class SimplePackageIndex implements PackageIndex {
...
-      final double scoreTreshold = bestScore / 25;
-      results.removeWhere((pr) => pr.score < scoreTreshold);
+    // reduce text results if filter did remove an url
+    textScore?.removeWhere((key) => !urls.contains(key));
+
+    List<PackageScore> results;
+    switch (query.order ?? SearchOrder.overall) {
+      case SearchOrder.overall:
+        final Score overallScore = new Score()
+          ..addValues(textScore?.values, 0.85)
```
mkustermann: You could avoid creating a copy of `textScore` and modify the copy, by just modify the `textScore` itself.

app/lib/search/index_simple.dart
```diff
@@ -71,66 +71,59 @@ class SimplePackageIndex implements PackageIndex {
...
+          ..addValues(getPopularityScore(urls), 0.10)
+          ..addValues(getHealthScore(urls), 0.05);
+        results = _flattenFromValues(overallScore.values);
+        break;
+      case SearchOrder.text:
+        results = _flattenFromValues(textScore.values);
+        break;
+      case SearchOrder.updated:
+        results = _flattenFromUrls(urls, _compareUpdated);
```
mkustermann: The name of the function is misleading, it has notting to do with urls, it's ordering/ranking based on updated timestamp.

isoos:
>> Dart SDK API index

> What is this?

https://github.com/dart-lang/pub-dartlang-dart/issues/190 (we shall eventually do code search, including Dart SDK APIs).

app/lib/search/index_simple.dart
isoos: This is the use case for e.g. sort by `updated`, where we don't have scores (no similarity, no health/popularity).

app/lib/search/index_simple.dart
isoos: Yeah, this was tricky to write (it is there in the older version too), and now it may not be needed at all. The idea was that if you don't specify any text query, yet specify a package name prefix, we still need to order the results by something, and at that point we only had this kind of "something".

I'll remove this, and disallow queries that have no text but filter on package prefix and order on text match score.

app/lib/search/index_simple.dart
isoos: In this case I think it is cleaner to keep it separate, as it is explicit how the text match score relates to the rest.

app/lib/search/index_simple.dart
isoos: Good point, will address it in a follow-up.

app/lib/search/index_simple.dart
isoos: This may happen automatically, because we weight longer N-grams and prefixes more, and if there is a 6+ character match, we will most likely remove the low-quality matches anyway.
I'll experiment with this, and do a follow-up in a subsequent PR.

mkustermann: Maybe slightly said in another way: If a search query has **only results with poor scores**, we should **display none of them** - not even the highest scored one.

isoos: Yeah, that makes sense. I've added a low-score filter, but it is very conservative now. I would still experiment with it more, because it is not always the package name that people are searching for, and the rules around it may become complex.

isoos:
Merging this now, further comments will be addressed in a subsequent PR.

app/lib/search/index_simple.dart
```diff
@@ -168,6 +161,81 @@ class SimplePackageIndex implements PackageIndex {
...
     );
   }
+
+  Score _searchText(String text, String packagePrefix) {
+    if (text != null && text.isNotEmpty) {
+      final Score textScore = new Score()
+        ..addValues(_nameIndex.search(text), 0.82)
+        ..addValues(_descrIndex.search(text), 0.12)
+        ..addValues(_readmeIndex.search(text), 0.06);
+      // removes scores that are less than 5% of the best
```
mkustermann: This comment is confusing, since `less than 5% of the best` could be interpreted as we drop the lowest `95%`. Maybe just consider removing these comments if the code itself is self-explaining.

PR #369, opened 2017-09-20
WIP: Running pana more frequently (don't merge yet)

isoos:
The goal here is to re-run analysis on the latest versions every day, and on the rest less frequently: as they age and reach one-year we re-run it only once a month, after that even less frequently.
Of course we re-run analysis if pana/flutter version changes, regardless of age.

I'm still observing how this behaves on staging (don't merge it yet), but I thing the code is ready for a review.

### DIFF
--- a/app/lib/analyzer/backend.dart
+++ b/app/lib/analyzer/backend.dart
@@ -24,7 +24,7 @@ AnalysisBackend get analysisBackend => ss.lookup(#_analysisBackend);
 
 final Logger _logger = new Logger('pub.analyzer.backend');
 
-const Duration freshThreshold = const Duration(hours: 12);
+const Duration freshThreshold = const Duration(hours: 22);
 const Duration obsoleteThreshold = const Duration(days: 180);
 
 /// Datastore-related access methods for the analyzer service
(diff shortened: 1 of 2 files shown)


mkustermann:
Please don't spend time on this, as I promised on the issue about component dependencies, I'll make a simple implementation of fixpoint-based dependency tracking, then we can avoid eagerly running everything every day (remember that running analysis on a package is e.g. using a fresh pub-cache, which implies we would tens of thousands of times download tarballs redundantly).

I don't think this is very salable and ideally we would not have any GC of analysis datastore entities.

isoos:
>  I'll make a simple implementation of fixpoint-based dependency tracking

65% of this change will be useful regardless of the frequency. I'll start to submit the smaller changes separately, and we can decide whether we want to have the increased frequency or the trigger on dependency tracking.

> ideally we would not have any GC of analysis datastore entities

`test`, `async` and other popular packages will trigger a re-index on most of the packages anyway, so I think it is worth to have the GC there. The currently running experiment shows that it is cheap to do.

isoos:
Current status: there seems to be an over-scheduling bug somewhere: versions keep getting selected for analysis, and then skipped by every running instance. I believe when I fix that bug, we have a good chance that the 1-day period can work.

isoos:
Folding this in favor of https://github.com/dart-lang/pub-dartlang-dart/pull/388, will revisit the repeated scheduling bug regardless (soon).

PR #370, opened 2017-09-20
Add types to public functions

kevmoo:


isoos:
can you add the linter rule too?

kevmoo:
The linter rules ends up spamming us in tests and top-level final fields – things that should be inferred

...so I left it off

PR #374, opened 2017-09-21
Remove duplicates in waiting tasks

isoos:
When the sources keep pushing the tasks in a higher frequency, the waiting queues may fill up with a lot of redundant tasks, and it is better to deduplicate them.

### DIFF
--- a/app/lib/shared/task_scheduler.dart
+++ b/app/lib/shared/task_scheduler.dart
@@ -52,7 +52,9 @@ class TaskScheduler {
   Future run() async {
     final PrioritizedStreamIterator<Task> taskIterator =
         new PrioritizedStreamIterator(
-            sources.map((TaskSource ts) => ts.startStreaming()).toList());
+      sources.map((TaskSource ts) => ts.startStreaming()).toList(),
+      deduplicateWaiting: true,
+    );
     while (await taskIterator.moveNext()) {
       final Task task = taskIterator.current;
       _pendingCount = taskIterator.pendingCount;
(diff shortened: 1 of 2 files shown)


app/lib/shared/task_scheduler.dart
```diff
@@ -136,8 +140,10 @@ class PrioritizedStreamIterator<T> implements StreamIterator<T> {
       final Queue<T> queue = _priorityQueues[i];
       _subscriptions[i] = source.listen(
         (T item) {
-          queue.add(item);
-          _triggerComplete();
+          if (!deduplicateWaiting || !queue.contains(item)) {
+            queue.add(item);
+            _triggerComplete();
+          }
```
mkustermann: I'm not sure how efficient `Queue.contains()` is, I suppose it does a linear scan.

Our [LinkedHashMap](https://api.dartlang.org/stable/1.24.2/dart-collection/LinkedHashMap-class.html) implementation preserves insertion order. So effectively you might be able to use it as a Queue with deduplication support (i.e. always insert new tasks into the map and only when the caller/user is able to get a new task you just take `map.keys.first` and remove that task).

It would also allow when delivering a new task to the caller/user, remove that task from all the `sources`s queues.

isoos: Good point! Is it the same with `LinkedHashSet`? It feels a bit strange to use the same key and value with a map.

mkustermann: The [documentation for LinkedHashSet](https://api.dartlang.org/stable/1.24.2/dart-collection/LinkedHashSet-class.html) claims iteration order equals insertion order, so I guess yes :)

Issue #348, opened 2017-09-13
Remove detectedTypes and use search in frontend for more features


isoos:
Now that we have proper platform classification from `pana`, using `detectedTypes` is no longer preferred (in `Package`, `PackageVersion` and throughout in `frontend`). The change requires a few things:
- [x] `search` to:
  - [x] support list of packages in reverse update order
  - [x] accept a query as valid without text or package prefix
  - [x] ~return more info (incl. platform information as part of the result for displaying the icon)~
  - [x] strip of extra info from search results
- [x] `fronend` to:
  - [x] use `search` to query `/packages` and `/flutter/packages`, because that is the quickest way to support all kinds of queries (e.g. later filtering on popularity, health, maintenance...)
  - [x] display icons in `search` from the `analyzer` service, using memcache
  - [x] `AnalysisView` to render flutter icon on package show page

- [ ] cleanup `detectedTypes` in Datastore

isoos:
@kevmoo @mkustermann: I don't see the `detectedType` property in staging Datastore anymore. Did you remove it from there and from prod too?

isoos:
Ah, I've realized that the datastore console did hide the column from me, maybe because the field hasn't been accessed in a while?

kevmoo:
I didn't touch it. 🤷‍♂️

PR #377, opened 2017-09-21
Use search service for /flutter/packages/

isoos:
Part of #348.

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -15,7 +15,8 @@ import 'package:shelf/shelf.dart' as shelf;
 import '../shared/analyzer_client.dart';
 import '../shared/handlers.dart';
 import '../shared/platform.dart';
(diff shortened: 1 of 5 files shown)


app/lib/frontend/handlers.dart
```diff
@@ -266,16 +268,29 @@ Future<shelf.Response> packagesHandlerHtml(
...
+      null,
+      platformPredicate: platformPredicate,
+      order: SearchOrder.updated,
+      offset: offset,
+      limit: limit,
+    ));
+    packages =
+        await backend.lookupPackages(results.packages.map((ps) => ps.package));
+  } else {
+    packages = await backend.latestPackages(offset: offset, limit: limit);
```
isoos: This leaves the listing of `/packages/` coming from Datastore. We could spare a few lines if this was coming from `search` too, but I left it here, because it has one remote call instead of two.

mkustermann:
lgtm, no comments. seems like you need to rebase first

PR #380, opened 2017-09-22
No-surpise Analysis GC

isoos:
Instead of doing a long cleanup, which may be never triggered if Appengine keeps restarting the instances every few days, we can do a quick, focused cleanup just after we store a new `Analysis`. My staging experiment showed that it is cheap to do.

### DIFF
--- a/app/lib/analyzer/backend.dart
+++ b/app/lib/analyzer/backend.dart
@@ -203,26 +203,50 @@ class AnalysisBackend {
     return true;
   }
 
-  Future deleteObsoleteAnalysisEntries() async {
+  /// Deletes the obsolete [Analysis] instances from Datastore. An instance is
+  /// obsolete, if:
+  /// - it is **not** the latest for the pana version, or
+  /// - it is older than 6 months (except if it is the latest one).
+  Future deleteObsoleteAnalysis(String package, String version) async {
     final DateTime threshold =
         new DateTime.now().toUtc().subtract(obsoleteThreshold);
-    final Query query = db.query(Analysis)
-      ..order('timestamp')
-      ..filter('timestamp <', threshold);
-    await for (Analysis analysis in query.run()) {
-      final PackageVersionAnalysis pva =
-          (await db.lookup([analysis.parentKey])).single;
-      if (pva.latestAnalysis != analysis.analysis) {
-        _logger.info('Deleting obsolete Analysis: ${analysis.packageName} '
-            '${analysis.packageVersion} ${analysis.analysis}');
-        await db.withTransaction((Transaction tx) async {
-          final List list = await tx.lookup([analysis.key]);
-          if (list[0] != null) {
-            tx.queueMutations(deletes: [list[0].key]);
-          }
-          await tx.commit();
-        });
+    final Query scanQuery = db.query(
+      Analysis,
+      ancestorKey: db.emptyKey
+          .append(PackageAnalysis, id: package)
+          .append(PackageVersionAnalysis, id: version),
+    );
+    final List<Key> obsoleteKeys = <Key>[];
+
+    final List<Analysis> existingAnalysis = await scanQuery.run().toList();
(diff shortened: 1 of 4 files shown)


mkustermann:
Ideally we would generally avoid GCs. If we would calculate a hash of the contents of the data in the entity which we care about, and do a lookup with the hash before we insert it, we can at least avoid inserting duplicated entries.

With a dependency-based triggering of analysis, we can avoid doing the analyze everything every X days.

Having these two things, should allow us to just keep the old data around, or do we expect that we will have **lots of distinct analysis results** in the DB?

isoos:
There are good arguments for keeping the data around, but I also find it refreshing that we explicitly don't care about old results. This allows us to introduce breaking changes in `pana`, without sweating about compatibility or conversions.

If we were to keep the data around and updated, what is the benefit of knowing the analysis of a version one year back? It is unlikely that the health score will improve. There may be newer releases. The attention is most likely on the latest stable, and its usability at that time, not in the past.

Re: hash and not storing the same result. We shall certainly do it, but the timestamp needs to be updated, because it has a great value to know how fresh the analysis is (also it can control when we do a re-analysis). Is there a way to only update only the timestamp in Datastore (and not the data payload)? Or shall we just update the full entry?

Wrt: GC. If you prefer, I will refactor this code into a script inside `bin/tools/`, and we can use it later if we want to.

app/lib/analyzer/backend.dart
```diff
@@ -203,27 +203,50 @@ class AnalysisBackend {
...
+
+    final List<_AnalysisRef> refs = <_AnalysisRef>[];
+    await for (Analysis analysis in scanQuery.run()) {
+      refs.add(new _AnalysisRef(
+          analysis.key, analysis.panaVersion, analysis.timestamp));
+    }
+    refs.sort((a, b) => a.timestamp.compareTo(b.timestamp));
+
+    // (length - 1) makes sure that we keep at least the latest version
+    for (int i = 0; i < refs.length - 1; i++) {
```
mkustermann: Maybe add a comment here: `NOTE: This is an O(n^2) algorithm`.

Or much better: Use a `Map<String, Analysis> panaVersion2LatestAnalysis = {}`. Then you can just walk through the `refs` / `existingAnalysis` **once** update the `panaVersion2LatestAnalysis` and every time you replace an entry you put it into `List<Key> toBeDeleted`.

app/lib/analyzer/backend.dart
```diff
@@ -203,27 +203,50 @@ class AnalysisBackend {
...
+      ancestorKey: db.emptyKey
+          .append(PackageAnalysis, id: package)
+          .append(PackageVersionAnalysis, id: version),
+    );
+    final List<Key> obsoleteKeys = <Key>[];
+
+    final List<_AnalysisRef> refs = <_AnalysisRef>[];
+    await for (Analysis analysis in scanQuery.run()) {
+      refs.add(new _AnalysisRef(
+          analysis.key, analysis.panaVersion, analysis.timestamp));
```
mkustermann: No need for the `_AnalysisRef` class, all the information is available in `Analysis`, just use:

```dart
final List<Analysis> existingAnalysis = scanQuery.run().toList();
existingAnalysis.sort(...);
```

app/lib/analyzer/backend.dart
```diff
@@ -203,26 +203,50 @@ class AnalysisBackend {
...
+      obsoleteKeys.removeLast();
+    }
+
+    if (obsoleteKeys.isNotEmpty) {
+      _logger.info('Deleting Analysis entries for $package $version: '
+          '${obsoleteKeys.map((k)=> k.id).join(',')}');
+      await db.withTransaction((tx) async {
+        tx.queueMutations(deletes: obsoleteKeys);
+        await tx.commit();
+      });
```
mkustermann: Whenever a transaction is used, you are supposed to be reading before writing. So I think here it would be ok to just remove them without a transaction.

PR #388, opened 2017-09-25
Use global (reverse) package dependency graph to determine which packages need re-analysis

mkustermann:
When startin up, the analysis service builds up a dependency graph
consisting of all packages in the database.  We track two different uses
of a package (as a direct or a dev dependency) via a special "dev/"
prefix.

The building of the initial dependency graph takes around 90 seconds
and consumes maybe around 20 MB.  Afterwards the dependency graph is
changed incrementally when new [PackageVersion]s are uploaded and is
used to answer the question of which packages need re-analysis.

Currently the implementation is a conservative estimate: We
say "A" needs re-analysis if a new [PackageVersion] of "B" is uploaded iff

  * any version of "A"'s pubspec.yaml file has "B" in the
    dependencies/dev_dependencies section

  * any version of "A"'s pubspec.yaml file has "X" in the dependencies
    section and "X" needs re-analysis if a new version of "B" is
    uploaded

This conservative estimate gives a superset of the actual set of
packages affected.  The current algorithm can be improved by taking
constraints into account.  Namely prevent the fixpoint algorithm to
including packages which have constraint violations.

It would also be possible to add the flutter sdk as another node in the
graph.  This would then also allow re-analysis of packages on new
flutter versions.  (Though currently we pinpoint the flutter version, so
it would make no difference at all).

Together with
[Create new Analysis only iff something changed](https://github.com/dart-lang/pub-dartlang-dart/issues/381)
it should allow us to get rid of re-analyzing all packages every X days
and the GC of analysis datastore entities (though we might still keep it
for other reasons).

### DIFF
--- a/app/lib/shared/deps_graph.dart
+++ b/app/lib/shared/deps_graph.dart
@@ -1,0 +1,218 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:gcloud/db.dart';
+import 'package:logging/logging.dart';
+
+import 'package:pub_dartlang_org/frontend/models.dart';
+import 'package:pub_dartlang_org/shared/task_scheduler.dart';
+
+final Logger _logger = new Logger('pub.package_graph');
+
+// To avoid having the same package names many times in memory we canonicalize them.
+class StringCanonicalization {
+  final Map<String, String> _map = <String, String>{};
+
+  String canonicalize(String value) {
+    // Fast case is when [value] is already in the map.
+    final canonicalized = _map[value];
+    if (canonicalized != null) return canonicalized;
+
+    return _map[value] = value;
+  }
+}
+
+// TODO(kustermann): We could incorporate the pubspec constraints to make the
+// sets smaller.
+class TransitiveDependencyGraph {
+  final Map<String, Set<String>> _deps = <String, Set<String>>{};
+
+  void addAll(String node, Set<String> deps) {
+    final nodes = _deps.putIfAbsent(node, _newSet);
+    final int before = nodes.length;
+    nodes.addAll(deps);
+    if (before != nodes.length) {
+      _fixpointNode(node);
+    }
+  }
+
+  void add(String node, String dep) {
+    if (_deps.putIfAbsent(node, _newSet).add(dep)) {
+      _fixpointNode(node);
+    }
+  }
+
+  Set<String> transitiveNodes(String node) => _deps.putIfAbsent(node, _newSet);
+
+  void _fixpointNode(String node) {
+    final Set<String> directs = _deps[node];
+    final temp = <Set<String>>[];
+    for (;;) {
+      final int before = directs.length;
+      temp.clear();
+      for (final String direct in directs) {
+        if (node != direct) {
+          final Set<String> transitives = _deps[direct];
+          if (transitives?.isNotEmpty == true) {
+            temp.add(transitives);
+          }
+        }
+      }
+      for (final Set<String> transitives in temp) {
+        directs.addAll(transitives);
+      }
+      if (before == directs.length) break;
+    }
+  }
+
+  static Set<String> _newSet() => new Set<String>();
+}
+
+class PackageDependencyBuilder {
+  static const Duration pollingInterval = const Duration(seconds: 10);
+  static const String devPrefix = 'dev/';
+
+  final DatastoreDB db;
+
+  // The forward (transitive) dependencies could be easily calculated, but
+  // right now we have no need for them.
+  //final TransitiveDependencyGraph normalDeps = new TransitiveDependencyGraph();
+
+  final TransitiveDependencyGraph reverseDeps = new TransitiveDependencyGraph();
+  final String Function(String) canonicalize = new StringCanonicalization().canonicalize;
+
+  DateTime _lastTs;
+
+  PackageDependencyBuilder(this.db);
+
+  Future scanExistingPackageGraph() async {
+    final sw = new Stopwatch()..start();
+    for (;;) {
+      _logger.info('Scanning existing package graph');
+      try {
+        // We scan from oldest to newest and therefore keep [_lastTs] always
+        // increasing.
+        final query = dbService.query(PackageVersion)..order('created');
+        await for (PackageVersion pv in query.run()) {
+          _addPackageVersion(pv);
+          _lastTs = pv.created;
+        }
+      } catch (e, s) {
+        _logger.severe(e, s);
+        continue;
+      }
+      _logger.info('Scanned package graph in ${sw.elapsed}');
+      return;
+    }
+  }
+
+  Stream<Set<String>> run() async* {
+    _logger.info('Monitoring new package uploads & trigger analysis');
+    for (;;) {
+      try {
+        final query = dbService.query(PackageVersion)
+          ..filter('created >', _lastTs)
+          ..order('created');
+        await for (PackageVersion pv in query.run()) {
+          // Emit a set of packages depending on this one.
+          _addPackageVersion(pv);
+          yield _affectedPackages(pv.package);
+
+          _lastTs = pv.created;
+        }
+      } catch (e, s) {
+        _logger.severe(e, s);
+      }
+      await new Future.delayed(pollingInterval);
+    }
+  }
+
+  void _addPackageVersion(PackageVersion pv) {
+    final String package = pv.package;
+    final Map pubspec = pv.pubspec.asJson;
+
+    final Set<String> depsSet = new Set<String>();
+    final normalDeps = pubspec['dependencies'];
+    if (normalDeps is Map) {
+      depsSet.addAll(normalDeps.keys);
+    }
+
+    final Set<String> devDepsSet = new Set<String>();
+    final devDeps = pubspec['dev_dependencies'];
+    if (devDeps is Map) {
+      devDepsSet.addAll(devDeps.keys);
+    }
+
+    // First we add the [package] together with the dependencies /
+    // dev_dependencies to the graph.  This will update the graph transitively,
+    // thereby calculating new users of the [package].
+    _add(package, depsSet, devDepsSet);
+  }
+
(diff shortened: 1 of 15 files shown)


mkustermann:
@isoos Haven't had enough time to test it out, but that would be a simple in-memory solution of triggering transitively dependent packages.

If we have more than 1 VM running analysis service we might want to find a way to avoid doing the re-analysis on all of them. One solution I was thinking of, is basically doing this computation (i.e. determine the set of packages which need re-analysis) on the VM which handles the upload of the package. Then it can use the "manual-trigger" mechanism to RPC call one analysis instance to handle it.

Let me know what you think about it.

mkustermann:
Issue https://github.com/dart-lang/pub-dartlang-dart/issues/317

isoos:
On a conceptual level this looks OK (I'll look into the code in a few hours), I'm just curious: was there any specific reason not using `pana`'s output for this?

mkustermann:
There's no specific reason, it's just much simpler to implement and doesn't need the pana results, so we could do that computation on package uploads (which trigger all transitively dependent packages). If we used pana results, then that might not be an option.

app/lib/shared/deps_graph.dart
```diff
@@ -0,0 +1,218 @@
...
+
+    //normalDeps.addAll(canonicalizedDevPackage, canonicalizedDevDeps);
+    for (final String dep in canonicalizedDevDeps) {
+      reverseDeps.add(dep, canonicalizedDevPackage);
+    }
+  }
+
+  // We use a synthetic package to handle weak dependencies / dev
+  // dependencies (no package is depending on this one).
+  String _devPackage(String package) => '$devPrefix$package';
```
isoos: I'm not sure that it is worth to differentiate between dev and direct dependencies in our case. I've had written a package where I put a dependency as dev, because the idea was to have a common interface for various databases, and put db-specific implementations alongside. Not a great design, but it happens, I've seen in another project too. Better to be more pessimistic and treat every dependency the same way.

app/lib/shared/deps_graph.dart
```diff
@@ -0,0 +1,218 @@
...
+import 'package:gcloud/db.dart';
+import 'package:logging/logging.dart';
+
+import 'package:pub_dartlang_org/frontend/models.dart';
+import 'package:pub_dartlang_org/shared/task_scheduler.dart';
+
+final Logger _logger = new Logger('pub.package_graph');
+
+// To avoid having the same package names many times in memory we canonicalize them.
+class StringCanonicalization {
```
isoos: The name gave me a pause, as I was expecting some kind of text transformation or an URI something that is an encoded form of the package name.

I'm trying to figure out a better name for it (around the expressions of singleton, dictionary related phrases), but haven't found a good one yet.

app/lib/shared/deps_graph.dart
```diff
@@ -0,0 +1,218 @@
...
+
+class PackageDependencyBuilder {
+  static const Duration pollingInterval = const Duration(seconds: 10);
+  static const String devPrefix = 'dev/';
+
+  final DatastoreDB db;
+
+  // The forward (transitive) dependencies could be easily calculated, but
+  // right now we have no need for them.
+  //final TransitiveDependencyGraph normalDeps = new TransitiveDependencyGraph();
```
isoos: Not sure if we need to keep these here in commented out code blocks. Remove?

app/lib/shared/deps_graph.dart
```diff
@@ -0,0 +1,218 @@
...
+      await new Future.delayed(pollingInterval);
+    }
+  }
+
+  void _addPackageVersion(PackageVersion pv) {
+    final String package = pv.package;
+    final Map pubspec = pv.pubspec.asJson;
+
+    final Set<String> depsSet = new Set<String>();
+    final normalDeps = pubspec['dependencies'];
```
isoos: I'd put these in the `Pubspec` class (which hopefully will get merged with `pana`'s `Pubspec` class), let's remove the duplicate functions here and there...

app/lib/shared/deps_graph.dart
```diff
@@ -0,0 +1,218 @@
...
+  String canonicalize(String value) {
+    // Fast case is when [value] is already in the map.
+    final canonicalized = _map[value];
+    if (canonicalized != null) return canonicalized;
+
+    return _map[value] = value;
+  }
+}
+
+// TODO(kustermann): We could incorporate the pubspec constraints to make the
```
isoos: I think this would be a lot of complexity here, for very little gain for now. I'd rather keep this class without version constraints, as long as the `analyzer` keeps up with the tasks.

app/lib/shared/deps_graph.dart
```diff
@@ -0,0 +1,218 @@
...
+    }
+  }
+
+  void add(String node, String dep) {
+    if (_deps.putIfAbsent(node, _newSet).add(dep)) {
+      _fixpointNode(node);
+    }
+  }
+
+  Set<String> transitiveNodes(String node) => _deps.putIfAbsent(node, _newSet);
```
isoos: The naming of the methods here doesn't make it clear in which directions it is working and what this returns (dependees vs. dependencies).

app/lib/shared/deps_graph.dart
```diff
@@ -0,0 +1,218 @@
...
+  final PackageDependencyBuilder builder;
+
+  DepsGraphBasedAnalysisTaskSource(this.db, this.builder);
+
+  @override
+  Stream<Task> startStreaming() async* {
+    await for (final Set<String> packages in builder.run()) {
+      for (final String packageName in packages) {
+        final Key packageKey = db.emptyKey.append(Package, id: packageName);
+        final Package package = (await db.lookup([packageKey]))[0];
```
isoos: this should be batched by the 10 or something (using e.g. `sliceList` from utils.dart)

isoos:
@kevmoo: I've updated the PR with a commit to fix the new Task() creation. Could you please approve/merge this? (I'll do the follow-up on my review comments.)

app/lib/shared/deps_graph.dart
isoos: Found the correct term: https://en.wikipedia.org/wiki/String_interning

mkustermann: That is a nice find, Istvan. Never heard of "string interning" before. But it's better to use a term that almost everyone is familiar with and yes, [canonicalization](https://en.wikipedia.org/wiki/Canonicalization) is a very good name for it.

You can try searching for "canonicalization" and "interning" in the dart sdk and see which is more common.

mkustermann:
A bit busy right now, though I'll take another stab at this soon and try to move it to the uploading backend. So we notify the analysis service with the expanded list (this has the benefit of avoiding N different analysis VMs to do the same thing, since the uploading VM is uniq).

mkustermann:
Please take another look, Istvan. This is as much as I would implement.

app/lib/shared/deps_graph.dart
isoos: With the recent discussion on [dependencies](https://github.com/dart-lang/pub-dartlang-dart/issues/441) I'm wondering whether this is needed at all.

app/lib/shared/notification.dart
```diff
@@ -21,23 +21,33 @@ NotificationClient get notificationClient => ss.lookup(#_notificationClient);
 class NotificationClient {
   final http.Client _client = new http.Client();
 
-  Future notifyAnalyzer(String package, String version) =>
-      _doNotify(activeConfiguration.analyzerServicePrefix, package, version);
+  Future notifyAnalyzer(
```
isoos: Do we need this new batching here? Notification is cheap, infrequent and can be non-blocking on the main caller. We could issue them in serial, and it would be still fine, while it would preserve the round-robin distribution of the workload.

app/pubspec.lock
```diff
@@ -12,7 +12,7 @@ packages:
       name: analyzer
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "0.30.0+2"
+    version: "0.30.0+4"
```
isoos: Can these version upgrades go in a separate change?

mkustermann: They can go before but not after, because my change alone causes the following:

```
$ pub run test 
... Skipped all tests
```

Only an upgrade of `package:test` seems to make it run tests again.

app/lib/shared/notification.dart
mkustermann: > Notification is cheap

Well, this can end up causing 1000 packages to be re-analyzed, so a 1000 http requests are not that cheap (and take some time as well).

Though I agree that from load balancing perspective it's not good. It did came to my mind, but then I weighted on the batch version.

Don't really have strong opinions on it.

app/lib/shared/deps_graph.dart
mkustermann: I would prefer to submit it with the support, then it's at least in the git history in case we change our minds later on.

We can remove it, yes. It depends what `package:pana` does. If it only looks into the `lib` folder, then I would be fine with it. If it also looks into other folders, I guess we should keep it.  (Actually I have no idea what the policy about the `bin` folder is: It contains entry points which are exposed like the `lib` folder. So I assume if one uses a package in `bin` then it has to go into the normal `dependencies` section???)

isoos: sgtm, but you need to fix the tests too (mocks require the new signatures)
`lib` and `bin` is treated the same way wrt. dependencies, because you could run the script from the `bin/` directory with `pub run`.

app/lib/shared/notification.dart
isoos: No strong opinions here either. Let's go with yours then.

mkustermann:
(Not sure what's wrong with the cla checker, all my accounts look good on https://cla.developers.google.com/)

app/lib/shared/notification.dart
mkustermann: I've made them individual requests now, but no longer await for the notifications to be out in the upload finish callback.

(The reason it's nice to wait for things inside the request handler is that then we'll get the logging messages associated with the request. This way if the analyzer notification causes an exception we would see it. But now I trigger it in background again as before, since I call the notification many times.)

mkustermann:
Thanks for reviewing, @isoos !

PR #392, opened 2017-09-28
Keep track of package creation time and order by it on the search service

isoos:


kevmoo:
This is just an *option*, right? It doesn't feed into the default results...

isoos:
> This is just an option, right? It doesn't feed into the default results...

Yeah, I'm working on a separate PR for that.

isoos:
(deferring merge a bit, as it is possible to do the same query with Datastore only, and if there isn't any other requirement for it, it is better leave it out)

isoos:
Closing without merging for now, because home page has roughly the same latency with the Datastore-only version, and otherwise there is no need for this.

isoos:
Re-opening and merging this, because the new homepage concept we have discussed requires the list of newest packages by platform.

PR #394, opened 2017-09-29
 Cleanup search results preparation and rendering

isoos:
- There is no need to query the latest dev versions, as we don't display any data off of them. If we worry about data consistency, I'd rather put a periodic job up there to check the Datastore for such cases, but there is no need to do it every time we render search results.

- With that, I've refactored `PackageVersionView` in a way that should be reusable on the home page and other index pages too: the version is presented together with the analysis, and stuff like the latest dev versions is optional.

- I've also did a little cleanup in the test dir and moved the shared services' mocks into a single place for easier referencing.

### DIFF
--- a/app/lib/frontend/search_service.dart
+++ b/app/lib/frontend/search_service.dart
@@ -48,28 +48,28 @@ Future<SearchResultPage> _loadResultForPackages(
 
   final List<Key> versionKeys =
       packageEntries.map((p) => p.latestVersionKey).toList();
-  final List<Key> devVersionKeys =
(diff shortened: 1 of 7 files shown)


isoos:
@mkustermann: Found a different (and simple) way to prepare the home page, therefore the `PackageVersionView` is not strictly needed. Please let me know if you agree with the removal of the latest dev version lookups, and I will re-introduce the changes here in separate, smaller PRs.

isoos:
Closing PR, I'm going to re-submit the unrelated parts in separate pieces.

PR #396, opened 2017-10-03
The new design's assets (built on 20171003)

isoos:


isoos:
> Do you think it would be good to use SASS (the Dart version) instead of raw CSS? Despite having done not much web development I heard that it is more composable/maintainable. Since there's a dart transformer it should be very easy to integrate.

I think the designer said in the early emails that there is some gulp-based process to build these. I'd also prefer the dart-based SASS, but I'd focus on that after the bulk of the migration is done.

PR #397, opened 2017-10-03
URL and cookie-based method to decide which design to serve

isoos:


mkustermann:
Suggestion: Instead of using cookies, could we simply serve it under a new URL space? i.e. 

```
old:
pub.dartlang.org/search
pub.dartlang.org/packages
pub.dartlang.org/packages/<name>/versions/<version>

new:
pub.dartlang.org/experimental/search
pub.dartlang.org/experimental/packages
pub.dartlang.org/experimental/packages/<name>/versions/<version>
....
```

This way I have less fear that someone has the cookie, loads a page, which will add it to memcache and suddenly all users will get that new page from the memcache.

Since we're using `shelf.Request` it should be quite easy to do this.

wdyt?

isoos:
that is a valid option, with a few different caveats: (1) we need to generate the `/experimental/` URLs everywhere now, and (2) after we are switching and removing the old design, we'll need to make sure that we don't render URLs for /experimental/. The later is more or less covered with a few simple text searches, so I'm going to close this for now and go ahead with the `/experimental/`

PR #398, opened 2017-10-03
New design: layout template rendering.

isoos:


app/lib/frontend/templates.dart
```diff
@@ -376,6 +376,39 @@ class TemplateService {
     return _renderPage('index', values, title: 'Pub: Dart Package Manager');
   }
 
+  /// Renders the `views/layout_new.mustache` template.
```
mkustermann: Could we move these into a new directory, like `views/v2/*.mustache`?

app/lib/frontend/templates.dart
```diff
@@ -636,6 +669,8 @@ class PackageLinks extends PageLinks {
 }
 
 abstract class LogoUrls {
+  static const String newDesignAssetsDir = '/static/20171003';
```
mkustermann: Could we make this `/static/v2/...`?

mkustermann:
lgtm, only comment about moving it to a `views/v2` directory.

Haven't looked much at the html.

app/views/v2/layout.mustache
```diff
@@ -0,0 +1,137 @@
+{{! Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
mkustermann: When adding more v2 layout templates, I think we should consider slowly also adding v2 golden files.

PR #400, opened 2017-10-03
Simple platform switch on the SearchQuery.

isoos:


app/lib/shared/search_service.dart
```diff
@@ -161,6 +161,16 @@ class SearchQuery {
     );
   }
 
+  SearchQuery scopeOnPlatform(PlatformPredicate platformPredicate) =>
```
mkustermann: Some packages have adopted a style where immutable classes (e.g. Uri/shelf.Request/...) provide a method (either `replace` or `change`) which returns a copy of the immutable object with some properties changed.

e.g. from [here](https://www.dartdocs.org/documentation/shelf/0.6.7+2/shelf/Request/change.html)

```dart
class Request {
  ... 
  Request change({Map<String, String> headers,
                  Map<String, Object> context,
                  String path,
                  body}) => ...;

  ...
}
```

Would be nice adopting this style here as well, so maybe rename this method to `change({PlatformPredicate platformPredicate})`?

The call sites would then look like

```dart
final SearchQuery newQuery = searchQuery.change(platformPredicate: newPlatformPredate);
```

isoos: It would work, with the note that we will need to create an empty `PlatformPredicate` or an empty query string to make something the 'default'. I'm fine with that, so renaming it to `change`.

PR #402, opened 2017-10-03
Don't fetch dev versions to render search result page.

isoos:


mkustermann:
lgtm - maybe we should have a golden file for the rendered search results page as well. wdyt?

isoos:
> maybe we should have a golden file for the rendered search results page as well. wdyt?

We do have that, but it isn't effected by this change:

https://github.com/dart-lang/pub-dartlang-dart/blob/master/app/test/frontend/golden/search_page.html

PR #403, opened 2017-10-03
New design: pagination.

isoos:


app/lib/frontend/templates.dart
```diff
@@ -567,26 +575,24 @@ abstract class PageLinks {
 
     final bool hasPrevious = currentPage > 1;
     results.add({
-      'state': {'state': hasPrevious ? null : 'disabled'},
```
kevmoo: Why the churn here?

app/test/frontend/golden/flutter_packages_index_page2.html
```diff
@@ -114,20 +114,22 @@
 
 <div class="pagination">
   <ul>
-      <li class="">
-        <a href="/packages?page=1">
+      <li class=" ">
+        <a href="&#47;packages?page=1">
```
kevmoo: This seems broken...

app/test/frontend/golden/flutter_packages_index_page2.html
```diff
@@ -114,20 +114,22 @@
 
 <div class="pagination">
   <ul>
-      <li class="">
-        <a href="/packages?page=1">
+      <li class=" ">
```
kevmoo: Why the extra whitespace?

app/test/frontend/golden/flutter_packages_index_page2.html
isoos: We have `&` in these URLs, and that needs to be escaped. Unfortunately the HtmlEscaper is a bit too eager to do other escaping, but it is valid XML attribute, and the browser treats it OK.

app/lib/frontend/templates.dart
isoos: I'm switching to simpler template preparation and mustache structure (esp. for the new pagination).

app/test/frontend/golden/flutter_packages_index_page2.html
isoos: Simpler rendering has its downsizes. Do you prefer the tradeoff for the better HTML output?

app/test/frontend/golden/flutter_packages_index_page2.html
isoos: Switched to an attribute-specific HTML escaper, this is now better.

app/lib/frontend/templates.dart
```diff
@@ -26,6 +26,7 @@ String _escapeAngleBrackets(String msg) =>
     const HtmlEscape(HtmlEscapeMode.ELEMENT).convert(msg);
 
 const HtmlEscape _HtmlEscaper = const HtmlEscape();
+const HtmlEscape _attrEscaper = const HtmlEscape(HtmlEscapeMode.ATTRIBUTE);
```
mkustermann: Could be consistent with `_Html` vs. `_attr`. I know our new style guide is very confusing regarding constants, previously we had nice `CONST_VALUE` names.

app/lib/frontend/templates.dart
```diff
@@ -404,6 +405,14 @@ class TemplateService {
     return _renderTemplate('layout', values, escapeValues: false);
   }
 
+  /// Renders the `views/pagination_new.mustache` template.
+  String renderNewPagination(PageLinks pageLinks) {
+    final values = {
+      'page_links': pageLinks.hrefPatterns(),
+    };
+    return _renderTemplate('pagination_new', values, escapeValues: false);
```
mkustermann: Maybe have a `v2/pagination` here.

app/lib/frontend/templates.dart
isoos: I'll do a follow-up PR to rename `_HtmlEscaper` -> `_htmlEscaper`.

PR #405, opened 2017-10-04
Tuning the search index weights

isoos:
The change addresses the too-low relevancy score of description and readme content (see https://github.com/dart-lang/pub-dartlang-dart/issues/404 for user-reported search):
- increased the weight of both `description` and `readme`
- decreased the TF-IDF weight we are using for scoring
- limited the `description` and `readme` to use minimum 3-grams, while the `name` index will use minimum `2-grams`; one-letter queries will no longer yield results;
- increased the term-weights of N-grams with N>=5, and it will also "back-scale" the length with the minimum length of the index

I've added the test items for the full `haversine` context, and I've also hand-checked it with the `game` search.

### DIFF
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -20,9 +20,9 @@ void registerPackageIndex(PackageIndex index) =>
 
 class SimplePackageIndex implements PackageIndex {
   final Map<String, PackageDocument> _packages = <String, PackageDocument>{};
-  final TokenIndex _nameIndex = new TokenIndex();
-  final TokenIndex _descrIndex = new TokenIndex();
-  final TokenIndex _readmeIndex = new TokenIndex();
+  final TokenIndex _nameIndex = new TokenIndex(minLength: 2);
+  final TokenIndex _descrIndex = new TokenIndex(minLength: 3);
+  final TokenIndex _readmeIndex = new TokenIndex(minLength: 3);
   DateTime _lastUpdated;
   bool _isReady = false;
 
@@ -167,9 +167,9 @@ class SimplePackageIndex implements PackageIndex {
   Score _searchText(String text, String packagePrefix) {
     if (text != null && text.isNotEmpty) {
       final Score textScore = new Score()
-        ..addValues(_nameIndex.search(text), 0.82)
-        ..addValues(_descrIndex.search(text), 0.12)
-        ..addValues(_readmeIndex.search(text), 0.06);
+        ..addValues(_nameIndex.search(text, coverageWeight: 0.90), 0.60)
+        ..addValues(_descrIndex.search(text, coverageWeight: 0.75), 0.20)
+        ..addValues(_readmeIndex.search(text, coverageWeight: 0.50), 0.20);
       textScore.removeLowScores(0.05);
       textScore.removeWhere((id) => textScore.values[id] < 1.0);
       return textScore;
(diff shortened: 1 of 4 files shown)


app/lib/search/index_simple.dart
```diff
@@ -282,7 +285,13 @@ class TokenIndex {
     }
     for (String id in counts.keys.toList()) {
       final double current = counts[id];
-      counts[id] = 100.0 * (current / _weights[id]) * (current / sumWeight);
+      double score = 100.0 * (current / sumWeight);
+      if (idfWeight > 0.0) {
+        // adjust score with inverse document frequency
+        final double idf = current / _weights[id];
```
mkustermann: The inverse document frequency `idf` should tell us how special the term is among our corpus (for very unique terms it will be high, for common terms it will be very low).

For example in english text, the term `the` will be common among many documents so it will get a very low `idf` score. Whereas the term `geospatial` will have a very high `idf` because only very few documents contain that term. (Normally the logarithm of the ratio is taken, see [wikipedia](https://en.wikipedia.org/wiki/Tf%E2%80%93idf) )

This computation can be done when building the index not when querying it! After building the idf's for all terms when indexing, we can normalize them to be within `[0.0, 1.0]`.

mkustermann: The term-frequency `tf(term, document)` for each term and document can also be pre-computed when building the index.

isoos:
> It seems like this is working around the fact that we don't do word-level based indexing and does so with all kinds of heuristics.

There is a bit more going on, because we index not only the words, but also:
- N-grams,
- prefix runs (e.g. longbow -> lon, long, longb, longbo, longbow),
- and parts of CamelCasedWords -> camel, cased, words

If we would stick with word-only indexing, we could have a very simple search logic, but we would lose:
- typo-tolerance,
- word variants and derived phrases (like word compositions).

app/lib/search/index_simple.dart
```diff
@@ -348,6 +366,9 @@ Set<String> _tokenize(String originalText) {
       prevLower = lower;
     }
   }
+  if (minLength > 0) {
+    tokens.removeWhere((t) => t.length < minLength);
```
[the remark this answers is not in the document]
isoos: we do have that for n-grams (`max(minNgram, minLength)`), and prefix is also starting at maxNgram+1, but we add the word, and also the parts of the CamelCasedWords, and instead of putting the check everywhere, it seemed to be easier to just remove them if they happen to be shorter.

mkustermann:
> but we would lose: ....

That is not necessarily the case, as I already mentioned in another discussion. (e.g. It's possible to just map the search term to the closest N words in corpus and use them for normal word-based searching/ranking).

Anyways, it was just a side note pointing out that the character n-grams causes us to easily miss obvious cases like the ranking `package:latlong` very high since when searching for the very unique `haversine` word (as reported in https://github.com/dart-lang/pub-dartlang-dart/issues/404).

As I mentioned above, it's fine if we land it as it is now -- with an additional `assert` and removal of the `if (idfWeight > 0.0)` condition.

app/lib/search/index_simple.dart
isoos: So, here we are a bit sloppy, and this is not really true TF-IDF. Making it a true TF-IDF was on my roadmap, but it never was a priority, and worked out okay-ish so far, and the two additional steps for it doesn't seemed to be justified (neither in complexity, nor time involved). I'm making it more obvious in the documentation, PTAL, and let me know if you'd prefer true TF-IDF (I'd guess it shall be after merging and deploying this).

app/lib/search/index_simple.dart
```diff
@@ -282,7 +285,13 @@ class TokenIndex {
     }
     for (String id in counts.keys.toList()) {
       final double current = counts[id];
-      counts[id] = 100.0 * (current / _weights[id]) * (current / sumWeight);
+      double score = 100.0 * (current / sumWeight);
+      if (idfWeight > 0.0) {
+        // adjust score with inverse document frequency
+        final double idf = current / _weights[id];
+        score *= 1 - (idfWeight * (1 - idf));
```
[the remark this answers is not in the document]
isoos: The expanded formula doesn't tell me why it is calculated that way, while at the same time I could reason about the closed form. I'm refactoring this part a bit, PTAL whether it makes more sense now?

isoos:
> It's possible to just map the search term to the closest N words in corpus and use them for normal word-based searching/ranking.

It is possible, but only with the assumption of (a) you know all of the words being used or (b) you trust new words. Which puts rare words that are important variation of "known" words in a hard spot: we either classify them as the 'common' word, or accept them as a trusted word. But with the later, what is the difference between that, and a typo?

I know, this may be an edge case, but it also affects brand names, creative project names and domain-specific abbreviations. So while that approach has merits, it is not clear that it is better than the n-grams.

PR #409, opened 2017-10-09
Calculate release frequency.

isoos:
This is not used directly for ordering, we will need to decide (1) whether this is going to be normalized across the packages (2) what is the "ideal" release schedule we value the most (e.g. not too frequent, but not too much time between releases).


mkustermann:
Maybe we should first decide whether it should be used for scoring at all (I mentioned my concerns about this before).

What I could understand if making a metric out of how much time passes between major version changes (i.e. breaking changes) which does in fact indicate if a package is under heavy development.

isoos:
discussed offline, postponing the idea

PR #412, opened 2017-10-09
New design: package show page + versions page

isoos:
I've tried to stick to the following refactor logic:
- Reuse as much as possible, don't duplicate code. I've extracted value calculation, old method goes to the front, new method afterward (that way the diff is reasonable).
- I've commented a few lines with //experimental_design: obsolete or new. These are to make it easier to clean up after the merge.

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -61,7 +61,7 @@ Future<shelf.Response> appHandler(
   } else if (path.startsWith('/packages/')) {
     return packageHandler(request);
   } else if (path.startsWith('/experimental/packages/')) {
-    return experimentalPackageHandler(request);
+    return packageHandlerV2(request);
   } else if (path.startsWith('/doc')) {
     return docHandler(request);
   } else if (path.startsWith('/static')) {
@@ -370,7 +370,7 @@ FutureOr<shelf.Response> packageHandler(shelf.Request request) {
 /// Handles the following URLs:
 ///   - /packages/<package>
 ///   - /packages/<package>/versions
-FutureOr<shelf.Response> experimentalPackageHandler(shelf.Request request) {
+FutureOr<shelf.Response> packageHandlerV2(shelf.Request request) {
   var path =
       request.requestedUri.path.substring('/experimental/packages/'.length);
   if (path.length == 0) {
@@ -387,7 +387,7 @@ FutureOr<shelf.Response> experimentalPackageHandler(shelf.Request request) {
     if (responseAsJson) {
       return packageShowHandlerJson(request, Uri.decodeComponent(path));
     } else {
-      return experimentalPackageVersionHandlerHtml(
+      return packageVersionHandlerHtmlV2(
           request, Uri.decodeComponent(path), null);
     }
   }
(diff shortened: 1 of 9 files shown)


isoos:
Shall I put this into multiple pieces? E.g. tag rendering can be easily extracted and it is being shared between all of the new pages.

app/test/frontend/handlers_test_utils.dart
```diff
@@ -202,12 +202,31 @@ class TemplateMock implements TemplateService {
   }
 
   @override
+  String renderNewPkgShowPage(
```
mkustermann: Maybe we should be consistent with what we use `new` or `experimental` or `v2`. I would just go for `v2` and  only use `experimental` in the URL, so users know that this is not stable yet.

app/views/v2/pkg/show.mustache
```diff
@@ -0,0 +1,147 @@
+{{! Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
mkustermann: We should consider adding golden files for the v2 templates as well.

app/lib/frontend/templates.dart
```diff
@@ -325,12 +379,43 @@ class TemplateService {
...
       'version_count': '$totalNumberOfVersions',
+      'icons': LogoUrls.versionsTableIcons, // experimental_design: new
     };
-    return _renderPage(
-      'pkg/show',
-      values,
+    return values;
+  }
+
+  /// Renders the `views/private_keys/show.mustache` template.
```
mkustermann: This comment seems wrong.

app/lib/frontend/handlers.dart
```diff
@@ -465,6 +547,13 @@ Future<shelf.Response> packageVersionHandlerHtml(
   return htmlResponse(cachedPage);
 }
 
+/// Handles requests for /experimental/packages/<package>/versions/<version>
+Future<shelf.Response> experimentalPackageVersionHandlerHtml(
+    shelf.Request request, String packageName, String versionName) {
+  return _packageVersionHandlerHtml(request, packageName, versionName,
+      templateService.renderNewPkgShowPage, false);
```
mkustermann: I would prefer to have a `backed.uiPackageCacheV2` which can be used for caching in the new v2 handlers and invalidation is invalidating `backed.uiPackageCache` and `backend.uiPackageCacheV2`.

Then we don't need all these booleans.

app/views/v2/pkg/show.mustache
isoos: added

app/test/frontend/handlers_test_utils.dart
isoos: Renamed them with a V2 postfix.

app/lib/frontend/templates.dart
isoos: Yeah, the original of the copy-paste was wrong too, fixed both.

app/lib/frontend/handlers.dart
isoos: The flag in https://github.com/dart-lang/pub-dartlang-dart/pull/415 will substitute the bool here.

PR #413, opened 2017-10-09
Remove detectedTypes from models.

isoos:
I might be wrong with this, but AFAICT we need to remove the property from the model before deleting from the Datastore. I have no idea how to do the later using the Dart APIs though...

### DIFF
--- a/app/bin/tools/remove_detected_types.dart
+++ b/app/bin/tools/remove_detected_types.dart
@@ -25,7 +25,7 @@ Future main(List<String> args) async {
 Future _updatePackageVersion(Key pvKey) async {
   await dbService.withTransaction((Transaction t) async {
     final PackageVersion pv = (await t.lookup([pvKey])).single;
-    pv.detectedTypes = null;
+    pv.additionalProperties.remove('detectedTypes');
     t.queueMutations(inserts: [pv]);
     await t.commit();
   });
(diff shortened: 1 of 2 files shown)


mkustermann:
The important part is not whether to remove the property first or not, but to ensure there is no longer any deployed versions/services which use the given properties.

Once this is guaranteed, one can simply make the list empty for all properties in a batch script and then remove the property alltogether.

Alternatively:
All our models inherit from `db.ExpandoModel`, which is a "catch all" for additional properties (used for forward/backwards compatible reader/writers). So once you remove the field from the model it will go into a generic map in `db.ExpandoModel:additionalProperties`. It can simply be removed from that map (just as if you clear the list as explained above).

mkustermann:
Why do you remove the batch script? Has someone already run it?

isoos:
> Why do you remove the batch script? Has someone already run it?

Once I had removed the property, the remove script didn't make much sense. But with your description of the `additionalProperties`, it does now, will change it.

isoos:
PTAL

> The important part is not whether to remove the property first or not, but to ensure there is no longer any deployed versions/services which use the given properties.

Staging is good to go with it, running the script now. For a quick check, everything that is deployed after October is not using the detectedTypes.

PR #415, opened 2017-10-10
Package memcache to use flag for pages with the new design.

isoos:
I've opted for the non-optional bool parameter, in order to make sure we get all of the uses both this time and when we are merging them back again.


mkustermann:
Couldn't we just have a `backend.uiPackageCacheV2` which has the `final bool version2;` property instead of every caller passing in a boolean?

PR #416, opened 2017-10-10
Very simple maintenance score: mostly penalties for trivial items and age.

isoos:


isoos:
@mkustermann: any comment on this?

app/lib/search/backend.dart
```diff
@@ -90,11 +95,50 @@ class SearchBackend {
...
+  double _calculateMaintenance(Package p, PackageVersion pv) {
+    final DateTime now = new DateTime.now();
+    final Duration age = now.difference(pv.created);
+    if (age > _twoYears) {
+      return 0.0;
+    }
+
+    double score = 1.0;
+
+    // over a year old?
```
mkustermann: Personally I think comments which just say the same thing as the source are not really necessary. Here the line below says `age > _year` which is the same as `over a year old`.

isoos:
> I have some concerns about scoring/ranking packages based on publish frequency, size of changelog...

I agree, and with this change, I think we are not doing anything that is too assuming or opinionated. Information in changelog and readme may not be the best proxy of maintenance, but their lack of is a bit of telling. Anyway, let's go with this one for now, and see how it goes.

PR #418, opened 2017-10-11
Use json_annotation, include copyright in generated files

kevmoo:
And also ignore prefer_final_locals

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -13,7 +13,7 @@ import 'package:logging/logging.dart';
 import 'package:gcloud/db.dart';
 import 'package:gcloud/storage.dart';
 import 'package:gcloud/service_scope.dart' as ss;
-import 'package:json_serializable/annotations.dart';
+import 'package:json_annotation/json_annotation.dart';
 
 import '../frontend/models.dart';
 import '../shared/analyzer_client.dart';
--- a/app/lib/search/backend.g.dart
+++ b/app/lib/search/backend.g.dart
@@ -1,4 +1,10 @@
+// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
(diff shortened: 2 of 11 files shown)


app/lib/shared/configuration.dart
```diff
@@ -91,12 +91,13 @@ class EnvConfig {
...
   EnvConfig._(
-    this.gaeService,
+    String gaeService,
     this.gcloudProject,
     this.gcloudKey,
     this.flutterSdkDir,
     this.isolateCount,
-  ) {
+  )
+      : this.gaeService = gaeService ?? 'default' {
```
isoos: I think @mkustermann asked me not to have a default value here and we keep being explicit about what service we want to start.

mkustermann: @kevmoo This is an unrelated change (again), this change should do only one thing, not several different orthogonal things :-/

app/test/frontend/handlers_test_utils.dart
```diff
@@ -177,12 +177,8 @@ class TemplateMock implements TemplateService {
...
 
   @override
-  String renderPkgIndexPage(
-      List<PackageView> packages,
-      PageLinks links,
-      {String title,
-      String faviconUrl,
-      String descriptionHtml}) {
+  String renderPkgIndexPage(List<PackageView> packages, PageLinks links,
+      {String title, String faviconUrl, String descriptionHtml}) {
```
mkustermann: ditto

app/test/frontend/handlers_test.dart
```diff
@@ -252,8 +252,7 @@ void main() {
           return [testPackageVersion];
         });
         registerBackend(backend);
-        await expectAtomXmlResponse(await issueGet('/feed.atom'),
-            regexp: '''
+        await expectAtomXmlResponse(await issueGet('/feed.atom'), regexp: '''
```
mkustermann: ditto.

kevmoo:
PTAL – sticking w/ one change 😄

PR #419, opened 2017-10-11
New design: index page

isoos:


isoos:
It also fixes issues in tags rendering and navigation on the main layout.

PR #421, opened 2017-10-11
Add missing explicit dependencies to pubspec, also fixed SDK constraint

kevmoo:


isoos:
Hm, this got missing too. github issues?

PR #423, opened 2017-10-11
Check analyzer and formatter on travis

kevmoo:
Using the mono_repo tools

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,26 @@
+# Created with https://github.com/dart-lang/mono_repo
 language: dart
-sudo: false
+
 dart:
-  - stable
   - dev
-script:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,7 +1,26 @@
...
+
 dart:
-  - stable
   - dev
-script:
-  - cd app && pub get && pub run test
+  - stable
+
+env:
+  - PKG=app TASK=dartanalyzer
```
mkustermann: Normally I'd rather call this "SUBFOLDER" or something like this, because it's only use is for pushd, though it is generated ...

kevmoo: there tends to be a one-to-en w/ the dir name and the pkg name. I hear ya.

PR #430, opened 2017-10-12
New design: package list page.

isoos:
This has a lot of overlap with the `/search` page, with the notable differences being:
- search has no platform-specific title and description
- package list page has no count on the number of packages
- package list includes the author (in the old design), while the search page didn't (in both) - TBD later

At first I've tried to merge this with the search page in one go, but ended up not doing that, rather doing it in separate steps (too many knobs involved, which may be better to do in isolation). In the end I expect that we'll have a unified `/packages` page with both the search functions and the platform switcher.

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -124,6 +124,66 @@ class TemplateService {
         title: pageTitle, faviconUrl: faviconUrl);
   }
 
+  /// Renders the `views/v2/pkg/index.mustache` template.
+  String renderPkgIndexPageV2(
+      List<PackageView> packages, PageLinks links, String currentPlatform) {
+    final packagesJson = [];
+    for (int i = 0; i < packages.length; i++) {
+      final view = packages[i];
+      packagesJson.add({
+        'url': '/packages/${view.name}',
+        'name': view.name,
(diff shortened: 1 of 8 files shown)


mkustermann:
If you update this PR, I'll review and we can land it & do a prod deployment with `/experimental` :)

app/lib/frontend/templates.dart
```diff
@@ -124,6 +124,66 @@ class TemplateService {
...
+
+    final platformTabs = [
+      platformTabData('Flutter', KnownPlatforms.flutter),
+      platformTabData('Web', KnownPlatforms.web),
+      platformTabData('Server', KnownPlatforms.server),
+      platformTabData('All', null),
+    ];
+
+    final values = {
+      'platform_tabs': platformTabs, // experimental_design: new
```
mkustermann: I don't even understand what `// experimental_design: new` means. Maybe just get rid of these comments. If the map entries are used in old/new templates, maybe rephrase these to `// used in v2 only`.

isoos: yeah, I'll do a followup to change everything. it was meant to be a unique string that is easy to search for when we'll do the merging and cleanup, a shorter version works just as well. (initially I was aiming for naming everything as "experimental", but that is no longer the case)

PR #433, opened 2017-10-12
update to latest popularity scores

kevmoo:


app/lib/shared/mock_scores.g.dart
```diff
@@ -14,1183 +14,1110 @@ part of pub_dartlang_org.shared.mock_scores;
 
 const _$mockScoresJsonLiteral = const {
```
mkustermann: Just as a side note: The const maps implementation in the VM is not very efficient, since it just puts (key, value) tuples into a list and does a linear search on lookup, so I would consider using a non-const map literal :D

For the interested reader, see [the source of _ImmutableMap](https://github.com/dart-lang/sdk/blob/master/runtime/lib/immutable_map.dart#L8) !

kevmoo: ack

PR #434, opened 2017-10-13
URI-based package memcache.

isoos:
@mkustermann: I was going back-and-forth with this before introducing the v2-flag for the memcache. This one is a simpler solution, and I believe we are not exposing ourselves to cache-filling attacks (e.g. one the expected path values will be cached, and you cannot introduce new cached keys by altering the request URI). I'm not sure if it is a valid concern though...

wdyt?

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -108,8 +108,9 @@ Future<shelf.Response> debugHandler(shelf.Request request) async {
 }
 
 /// Handles requests for /
-Future<shelf.Response> indexHandler(_) async {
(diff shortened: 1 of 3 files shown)


mkustermann:
I think it's fine the way we have it now. Once v2 is the only option we just remove the old v1 caching lines and the booleans.

The reason I think URIs are not as good as explicit package/version input is that it's hard to control what the caller provides. Maybe they provide different URIs and the invalidation logic then misses to invalidate them, causing strange effects.

By having the interface to the cache object get explicit package/version we avoid that problem.

PR #440, opened 2017-10-13
remove reference to crossdart in v2

devoncarew:
- remove reference to crossdart in v2
- fix https://github.com/dart-lang/pub-dartlang-dart/issues/439

(not 100% sure how to run the tests locally, but I assume travis will take care of it)

### DIFF
--- a/app/views/v2/pkg/show.mustache
+++ b/app/views/v2/pkg/show.mustache
@@ -124,11 +124,6 @@ import 'package:{{package}}/{{library}}';
       <a class="link" href="{{package.documentation}}">{{package.nice_documentation}}</a>
     {{/package.documentation}}
 
-    {{#package.crossdart}}
-      <h3 class="title">Source code (hyperlinked)</h3>
-      <a class="link" href="{{package.crossdart}}">{{package.nice_crossdart}}</a>
-    {{/package.crossdart}}
-
     <h3 class="title">{{package.uploaders_title}}</h3>
     <p>{{& package.uploaders_html}}</p>
 
--- a/app/lib/frontend/models.dart
+++ b/app/lib/frontend/models.dart
@@ -204,14 +204,6 @@ class PackageVersion extends db.ExpandoModel {
 
   String get documentationNice => niceUrl(documentation);
 
-  String get crossdart {
-    final name = Uri.encodeComponent(packageKey.id);
-    final version = Uri.encodeComponent(id);
-    return 'https://www.crossdart.info/p/$name/$version/';
-  }
-
-  String get crossdartNice => niceUrl(crossdart);
-
   String get homepage {
     return pubspec.homepage;
   }
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -418,8 +418,6 @@ class TemplateService {
         'nice_homepage': selectedVersion.homepageNice,
         'documentation': selectedVersion.documentation,
         'nice_documentation': selectedVersion.documentationNice,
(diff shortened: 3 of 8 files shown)


app/views/v2/pkg/show.mustache
```diff
@@ -124,11 +124,6 @@ import 'package:{{package}}/{{library}}';
       <a class="link" href="{{package.documentation}}">{{package.nice_documentation}}</a>
     {{/package.documentation}}
 
-    {{#package.crossdart}}
-      <h3 class="title">Source code (hyperlinked)</h3>
-      <a class="link" href="{{package.crossdart}}">{{package.nice_crossdart}}</a>
```
kevmoo: Should cleanup the `package.*crossdart` values, too!

devoncarew: The model references? I assume those are stilled used by the v1 code?

kevmoo: We should kill the feature there, too – while you're at it 😄

devoncarew: Done!

mkustermann:
May I ask why we remove these references?

devoncarew:
Yup:

- crossdart doesn't have a maintainer, is not a dart team resource, and
- linking to it from here implies an SLA that it doesn't have. If it goes down, who fixes it? We've had many issues linking to dartdocs.org; something with minimal SLA and frequent issues
- crossdart - while very nice - doesn't serve a critical use case, unlike something like dartdocs.org

mkustermann:
> crossdart doesn't have a maintainer, is not a dart team resource, and

It has the same status as the dartdocs project and the dart team is an owner of both, crossdart and dartdocs.

> linking to it from here implies an SLA that it doesn't have ...

Same arguments are as well true for dartdocs (does pub.dartlang.org have a SLA?, I don't think so)

> crossdart - while very nice - doesn't serve a critical use case, unlike something like dartdocs.org

Did we ever look at usage stats to see how much it's used? Or do we just conclude nobody clicks on these links without knowing?


Well, don't misunderstand me - I have not that strong opinions on this, but sometimes I use it if I want to browse a package's sources. So it seems to me as we remove functionality that quite a few users might have gotten accustomed to. The reason for removal is just "nobody is working on it atm", which has been true for a long time, and yet it still works.

devoncarew:
> It has the same status as the dartdocs project and the dart team is an owner of both, crossdart and dartdocs.

From my understanding, they were both built by external contributors, hosted on non-dart team urls, for a long time were lightly maintained, and have more recently become unmaintained.

> Same arguments are as well true for dartdocs (does pub.dartlang.org have a SLA?, I don't think so)

pub does! if it goes down, the dart team is on the hook for fixing it.

dartdocs.org serves an important enough use case that we're looking at hosting it ourselves as part of this pub site refresh

> Did we ever look at usage stats to see how much it's used? Or do we just conclude nobody clicks on these links without knowing?

These services can still exist, this is more a call about what we link out to. This is more about taking a careful look at how many projects we're supporting.

> Well, don't misunderstand me - I have not that strong opinions on this, but sometimes I use it if I want to browse a package's sources. So it seems to me as we remove functionality that quite a few users might have gotten accustomed to. The reason for removal is just "nobody is working on it atm", which has been true for a long time

Thanks for the feedback, and I apologize that there was no communication here before the PR.

> and yet it still works

The dartdoc team does gets pinged every time there's an issue with dartdocs.org (historically, ~monthly). One is maintained by the dart team, the other an external contributor. People conflate the two because of the name and because we link from the pub package page.

PR #445, opened 2017-10-18
Tweak to the menu styles

kevmoo:


static/v2/css/style.css
```diff
@@ -128,12 +128,11 @@ body{
 
 }
 
-.site-nav > .link{
```
isoos: This is a built CSS, not the "master" file. First we'll need to port the design files + its build process into this repository. If you submit this now, it will be hard to backport the change to the base....

kevmoo: Oh! So what should I do?

isoos: @liyangguang: could you please open a PR that ports the new design's files and build scripts into this repo?

liyangguang: Do you mean gulp etc.? I thought we'll just throw gulp away and just use this one css file, since gulp is not really doing much, and my gulp setup is unnecessarily complex.

What do we need here? I can make changes to the css and PR. (e.g. separate into different parts, etc.)

isoos: @liyangguang: Ah, I see. In other words, this CSS is all we've got, and we are good to do changes on it, right? If you can confirm that, @kevmoo is good to go with the change.

liyangguang: Yes!

PR #447, opened 2017-10-19
Platform info for search preserved (on search and index pages).

isoos:


app/test/frontend/golden/v2/search_page.html
```diff
@@ -91,6 +91,7 @@ <h1 class="_visuallyhidden">Dart pub</h1>
     <main class="package-banner">
       <form class="search-bar" action="/experimental/search">
         <input class="input" name="q" placeholder="Search packages" autocomplete="on" autofocus value="foobar">
+
```
kevmoo: shouldn't the hidden input show up here? Or just not yet?

isoos: It is a conditional part of the template, because there is no reason to put the `platform` parameter in the URL if there is no platform set.

kevmoo: Ah...k

kevmoo: So we just don't have a `golden` that causes the input to be rendered?

isoos: yeah, I'll add the golden files with the landing blurb change (mostly because I wanted to spare a few merge conflicts)

mkustermann:
Why do we use hidden inputs? 

It seems much better if the URL contains all necessary information to render a page instead of relying on URL + some hidden input fields. For example can the same URL be rendered with different values for the input fields / the same request to a URL have different results depending on hidden input values? This makes memcache'ing much more fragile.

isoos:
> Why do we use hidden inputs?

Hidden inputs are required for adding the extra platform parameter in the URL:
- No platform URL: `/search?q=text`
- Platform URL: `/search?q=text&platform=flutter`

Kevin's question was whether to include the hidden field in the form unconditionally, which will result in the following non-platform URL: `/search?q=text&platform=`

I believe the URL without the `&platform=` is better, and unless you specifically ask for it, I'd rather land this PR.

isoos:
> It seems much better if the URL contains all necessary information to render a page instead of relying on URL + some hidden input fields.

I agree, and there is nothing like that around, especially because it messes with caching. There is an unrelated plan to redirect the `/` landing page to a platform-specific landing based on a cookie, but it has nothing to do with the current change.

isoos:
PTAL: I've updated the same thing with the platform-specific landing pages, and in the golden files you can see that the `<form>` now contains the `platform` field (line 65 in each).

kevmoo:
Needs a dartfmt, @isoos

mkustermann:
> Hidden inputs are required for adding the extra platform parameter in the URL:

Sorry about that, as long as they end up in the URL, I'm fine with it. (somehow my bad html skilled brain thought these values might get POSTed).

> to a platform-specific landing based on a cookie

We should think carefully if we ever want cookies **ever**.

PR #453, opened 2017-10-20
New search score combination: text match is based on union, overall match is done with multiplication

isoos:
Improves `haversine` and `angular`, doesn't change `travis`. I think it also helps for `game`, but that and `travis` would really need the TF-IDF scoring for more improvements.

### DIFF
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -3,7 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:math';
+import 'dart:math' as math;
 
 import 'package:gcloud/service_scope.dart' as ss;
 
@@ -72,10 +72,10 @@ class SimplePackageIndex implements PackageIndex {
   Future<PackageSearchResult> search(SearchQuery query) async {
     // do text matching
     final Score textScore = _searchText(query.text, query.packagePrefix);
+    final Score base = textScore ?? _allPackages();
 
     // The set of packages to filter on.
-    final Set<String> packages =
-        textScore?.getKeys()?.toSet() ?? _packages.keys.toSet();
+    final Set<String> packages = base.getKeys();
 
     // filter on package prefix
     if (query.packagePrefix != null) {
@@ -93,20 +93,22 @@ class SimplePackageIndex implements PackageIndex {
           !query.platformPredicate.matches(_packages[package].platforms));
     }
 
-    // reduce text results if filter did remove a package
-    textScore?.removeWhere((key) => !packages.contains(key));
+    // reduce base results if filter did remove a package
+    final Score filtered = base.project(packages);
 
     List<PackageScore> results;
     switch (query.order ?? SearchOrder.overall) {
       case SearchOrder.overall:
-        final Score overallScore = new Score()
-          ..addValues(textScore?.values, 0.85)
-          ..addValues(getPopularityScore(packages), 0.10)
-          ..addValues(getHealthScore(packages), 0.05);
-        results = _rankWithValues(overallScore.values);
+        final Score overallScore = Score.multiply([
+          filtered,
+          new Score(getPopularityScore(packages)).map((v) => 0.5 + 0.5 * v),
+          new Score(getHealthScore(packages)).map((v) => 0.75 + 0.25 * v),
+          new Score(getMaintenanceScore(packages)).map((v) => 0.9 + 0.1 * v),
+        ]).removeLowValues(fraction: 0.01);
+        results = _rankWithValues(overallScore.getValues());
         break;
       case SearchOrder.text:
-        results = _rankWithValues(textScore.values);
+        results = _rankWithValues(textScore.getValues());
         break;
       case SearchOrder.created:
         results = _rankWithComparator(packages, _compareCreated);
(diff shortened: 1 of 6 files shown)


app/lib/search/index_simple.dart
```diff
@@ -155,35 +156,38 @@ class SimplePackageIndex implements PackageIndex {
...
   Map<String, double> getMaintenanceScore(Iterable<String> packages) {
     return new Map.fromIterable(
       packages,
-      value: (String package) => (_packages[package].maintenance ?? 0.0) * 100,
+      value: (package) => (_packages[package].maintenance ?? 0.0),
     );
   }
 
+  Score _allPackages() =>
+      new Score(new Map.fromIterable(_packages.keys, value: (_) => 1.0));
```
mkustermann: This would cause us to create a map of the size of `O(number-of-packages)` with values all being `1.0` - for every search. This is quite inefficient, given that the caller can simply use `1.0` if there is no score for it.

As mentioned on previous CLs, it is not very advisable to build up big data structures and then prune them later. Only build what you need.

app/lib/search/index_simple.dart
```diff
@@ -227,29 +231,70 @@ class SimplePackageIndex implements PackageIndex {
...
+  double getMaxValue() => _values.values.fold(0.0, max);
+  Map<String, double> getValues() => _values;
+
+  double operator [](String key) => _values[key] ?? 0.0;
+
+  /// Calculates the intersection of the two Scores, by multiplying the values.
+  ///
+  /// [weight] can be used to dampen the removal-effect of multiplication with
+  /// zero or very low values, and to keep results from the original set.
+  Score multiply(Score other, {double weight: 0.0}) {
```
mkustermann: Maybe it would be better making this a static method and rename it to `interpolate`:
```dart
/// Parameter t in [0, 1] interpolates between [scoreA] and [scoreB].
/// The resulting [score]'s keys are the common keys between [scoreA] and [scoreB]
Score.interpolate(scoreA, scoreB, 0.1);
````

File: app/lib/search/index_simple.dart [hunk 3]
```diff
@@ -227,29 +231,70 @@ class SimplePackageIndex implements PackageIndex {
...
+    final Set<String> keys = getKeys().intersection(other.getKeys());
```
mkustermann: Maybe add an assertion that the interpolation variable traces the line segment, i.e. is between [0,1]

app/lib/search/index_simple.dart
```diff
@@ -227,29 +231,70 @@ class SimplePackageIndex implements PackageIndex {
...
+        final double otherValue = other[key];
+        final double m = otherValue * weight + (1.0 - weight);
+        return this[key] * m;
+      },
+    ));
+  }
+
+  /// Calculates the union of the two Scores, by using the maximum values from
+  /// the sets.
+  Score union(Score other) {
```
mkustermann: Would prefer having this as

```dart
Score.max(scoreA, scoreB);
```

app/lib/search/index_simple.dart
```diff
@@ -227,29 +231,70 @@ class SimplePackageIndex implements PackageIndex {
...
+      result[key] = value;
+    }
+    return new Score(result);
   }
 
-  void removeLowScores(double fraction) {
-    final double maxValue = values.values.fold(0.0, max);
-    final double cutoff = maxValue * fraction;
-    removeWhere((key) => values[key] < cutoff);
+  Score withKeys(Iterable<String> keys) {
```
mkustermann: Maybe `restrictToKeys` or `project` (from [relational database terminology](https://en.wikipedia.org/wiki/Projection_(relational_algebra))) would be better.

app/lib/search/index_simple.dart
isoos: For now I'm not worried about this. Most queries I've measured are about 1-3 ms, and the maximum memory use of a search instance is between 380 and 420MB. I'd prefer to keep the code simpler in expense of memory use for now, and when we are happy with it, we can take a look in optimizations.

app/lib/search/index_simple.dart
isoos: I've added interpolate, but kept the multiple (rewrote it a bit similar to `max`). The interpolation as direct part of the scoring is not great because it can not only decrease good scores, but also it can increase bad scores, which is not what we want here.

app/lib/search/index_simple.dart
```diff
@@ -227,29 +230,82 @@ class SimplePackageIndex implements PackageIndex {
...
+    final result = <String, double>{};
+    for (String key in keys) {
+      final double value = _values[key];
+      if (value == null) continue;
+      result[key] = value;
+    }
+    return new Score(result);
+  }
+
+  Score interpolate(double t) {
```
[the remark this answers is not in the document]
isoos: How would you call a function that maps the region of [0.0 ... 1.0] values to [0.9 .. 1.0]? Because the role of interpolate is only to do that to prepare multiplication but to prevent to much penalty if e.g. a popularity value is zero.

The problem with the interpolation in your suggestion is that it does seem to depend on the application order of the popularity/health/maintenance values, and therefore not symmetric, not associative. (I haven't actually checked it, but it felt that way). Multiplication on the other hand is associative and doesn't depend on the order, it felt more "right".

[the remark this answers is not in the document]
mkustermann: I have never proposed to use use linear interpolation, I'm just stating the fact that this function is not doing linear interpolation (and thought the previous version of the code did, that's why I suggested the name).

How about `map`:

```dart
final Score result = scoreA.map((double value) => 0.8 + 0.2 * value);
```

?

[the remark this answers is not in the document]
isoos: `map` it is then.

Yeah, a previous version was a bit confusing in that regard, because while it appeared from the parameter passing that it does interpolation, it was the same logic as now. It was more memory efficient though, because it spared the creation of the intermediate Map. Nevertheless, I like the current version better, it is easier to follow.

PR #454, opened 2017-10-20
Fetch and use popularity data from bucket.

isoos:


isoos:
This is using Kevin's suggestion of calculating the total votes with the bezier normalization.

app/bin/service/search.dart
```diff
@@ -51,9 +51,16 @@ void _main(int isolateId) {
...
         storageService, activeConfiguration.searchSnapshotBucketName);
-    registerSnapshotStorage(new SnapshotStorage(storageService, bucket));
+    registerSnapshotStorage(
+        new SnapshotStorage(storageService, snapshotBucket));
+
+    final Bucket popularityBucket = await _createOrGetBucket(
+        storageService, activeConfiguration.popularityDumpBucketName);
+    registerPopularityStorage(
+        new PopularityStorage(storageService, popularityBucket));
+    await popularityStorage.init();
```
mkustermann: Be a bit careful when to `await` during initialization, if we wait too long until we call `runAppEngine` then GAE will think the service is unhealthy and kill the VM. Guess it should not take too long to fetch the latest popularity data?

app/lib/search/backend.dart
```diff
@@ -138,6 +147,71 @@ class SearchBackend {
...
 
+class PopularityStorage {
+  final Storage storage;
+  final Bucket bucket;
+
+  final Map<String, double> _values = {};
+
+  PopularityStorage(this.storage, this.bucket);
+
+  double operator [](String package) => _values[package];
```
mkustermann: I would prefer not overloading operators, to avoid confusion on call sites if this is a map or something else. Maybe rename this to `lookup`?

app/lib/search/backend.dart
```diff
@@ -138,6 +147,71 @@ class SearchBackend {
...
+  Future init() async {
+    await fetch();
+    new Timer.periodic(const Duration(days: 1), (_) {
+      fetch();
+    });
+  }
+
+  Future fetch() async {
+    _logger.info('Updating popularity storage...');
+    final List<BucketEntry> list = await bucket.list().toList();
```
mkustermann: Please wrap this in try-catch and re-try. If the first call has an error, e.g. a timeout, then the VM will never get the scores.

I cannot recall if `package:gcloud` has specific exception classes, so we can avoid re-trying if e.g. the bucket doesn't exist or we don't have permission. If so, we could special case on the exception type.

mkustermann: After time this will become a quite slow operation. Instead of listing the entire bucket contents, which can not only do several RPCs (under the hood), we should only try to fetch a `popularity-latest.json`.

The uploader should be able to upload as `popularity-YYYY-MM-DD.json` (so we have the historic values) and then do a compare-and-swap way to copy it over `popularity-latest.json`. There will be no races and everything is atomic: Either the new file is there entirely or the old file is there, this is guaranteed by GCS.


(Same is true for snapshots of search index: If we cache it on GCS I have probably added the same comment there)

app/lib/search/backend.dart
```diff
@@ -138,6 +147,71 @@ class SearchBackend {
...
+      final int devVotes = int.parse(item['votes_dev']);
+      final int directVotes = int.parse(item['votes_direct']);
+      final int totalVotes = int.parse(item['votes_total']);
+      final int finalVotes = directVotes * 25 + devVotes * 5 + totalVotes;
+      rawTotals[package] = finalVotes;
+    }
+    final summary = new Summary(rawTotals.values);
+    for (String package in rawTotals.keys) {
+      final int raw = rawTotals[package];
+      _values[package] = summary.bezierScore(raw);
```
mkustermann: There is still no document explaining all the scoring logic and furthermore no clear motivation what benefit (or drawbacks) this non-linear scaling has, except for: "npmjs does it, so let's hope they had a good reason, the same reason applies to us and therefore let's just copy it".

app/lib/search/backend.dart
```diff
@@ -138,6 +147,71 @@ class SearchBackend {
...
+      _logger.warning('No popularity data found!');
+      return;
+    }
+    // Try to load the available snapshots in reverse order (latest first).
+    names.sort();
+    for (String selected in names.reversed) {
+      _logger.info('Loading popularity data: $selected');
+      try {
+        final String json =
+            await bucket.read(selected).transform(UTF8.decoder).join();
```
mkustermann: Also we should consider zipping the file, usually it gives 10x size reductions and the cost of unzipping is low.

app/bin/service/search.dart
isoos: I think the fetch and processing shouldn't take longer than a couple of seconds. I think it is safe for now to assume that the storage is available and ready, and later we could add error handling if there is an issue with that. (e.g. the UI could just hide the related data)
However such "deep" error handling is not something I'd like to focus on at the moment, because I think there is very little risk for it, and the current "mock_scores" fallback is good enough for now.

app/lib/search/backend.dart
isoos: Fair point. I'd like to keep the current logic until @kevmoo confirms that the upload process will do the swap. In a separate PR I'll update the snapshot code to use the same thing.

Btw. I think it may be worth to keep the full list fetch as a fallback if the file is not there for some reason. WDYT?

app/lib/search/backend.dart
isoos: Yeah, I'll make it a priority to start a doc on it before the end of this week.

For the score value in this code, we need a normalization, and bezier is just as good as anything else for now. (I believe that once we have the building blocks in place, changing the normalization or the weights, or in general figuring out what we really want with the metrics will much easier with live data at hand.)

app/lib/search/backend.dart
mkustermann: To ensure this gets addressed I've filed https://github.com/dart-lang/pub-dartlang-dart/issues/468

app/lib/search/backend.dart
```diff
@@ -138,10 +148,83 @@ class SearchBackend {
...
+    names.sort();
+    for (String selected in names.reversed) {
+      _logger.info('Loading popularity data: $selected');
+      try {
+        Stream<List<int>> stream = bucket.read(selected);
+        if (selected.endsWith('.gz')) {
+          stream = stream.transform(_gzip.decoder);
+        }
+        final String json = await stream.transform(UTF8.decoder).join();
+        final Map map = JSON.decode(json);
```
mkustermann: There is no need for `.join()`:

```dart
final Map map = await stream.transform(UTF8.decoder).transform(JSON.decoder).single;
```

app/lib/search/backend.dart
```diff
@@ -138,10 +148,83 @@ class SearchBackend {
...
+
+  void _updateLatest(Map raw) {
+    final Map<String, int> rawTotals = {};
+    final List items = raw['items'];
+    for (Map item in items) {
+      final String package = item['pkg'];
+      final int devVotes = int.parse(item['votes_dev']);
+      final int directVotes = int.parse(item['votes_direct']);
+      final int totalVotes = int.parse(item['votes_total']);
+      final int finalVotes = directVotes * 25 + devVotes * 5 + totalVotes;
```
mkustermann: I don't know what these votes concretely are, since there hasn't been any code reviews for the generation of this data. But the `+ totalVotes` could be much higher than `directvotes * 25 + devVotes * 5` if the `votes_total` are transitive votes.

PR #457, opened 2017-10-24
No need to include local type vars

kevmoo:
See https://www.dartlang.org/guides/language/effective-dart/usage#consider-omitting-the-types-for-local-variables

### DIFF
--- a/app/analysis_options.yaml
+++ b/app/analysis_options.yaml
@@ -32,6 +32,7 @@ linter:
     - list_remove_unrelated_type
     - literal_only_boolean_expressions
 #    - non_constant_identifier_names
+    - omit_local_variable_types
     - one_member_abstracts
 #    - only_throw_errors
     - overridden_fields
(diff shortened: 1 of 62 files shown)


isoos:
I'm 50/50 on this, for short blocks I am fine removing them, for complex ones it may be better for readability.

mkustermann:
Agree with @isoos, instead of inferring always if we can, we should only do it where we save space and not reduce readability. Mainly if the right hand side makes it clear for the reader what type the expression has, then it's ok to let the inference do it.

For example, your change includes
```dart
final apiHandler = await setupServices(activeConfiguration)
```
Is it clear what `apiHandler` is, what kind of arguments it takes? It's probably a function type, but could be anything. a `shelf.Handler apiHandler = ...` will make this clear.

kevmoo:
Closing this out. Clear there is no way to lean here that a linter can help w/

PR #462, opened 2017-10-24
Changing from Popular to Top on the v2 landing pages.

isoos:


app/test/frontend/templates_test.dart
```diff
@@ -33,7 +33,6 @@ void main() {
       } else {
         htmlParser.parse();
       }
-      new File('$goldenDir/$fileName').writeAsStringSync(content);
```
isoos: Sorry, this got sneaked in on one of the previous PRs.

PR #467, opened 2017-10-25
Extract platform tab rendering.

isoos:
The new method is a bit bigger, but (a) it is needed for the landing pages too (b) with the new URL structure proposed in the document it will get shorter.

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -148,24 +148,8 @@ class TemplateService {
       descriptionHtml = flutterPackagesDescriptionHtml;
     }
 
-    Map platformTabData(String text, String platform) {
-      final String param = platform == null
-          ? null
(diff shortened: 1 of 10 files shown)


kevmoo:
Check travis...

app/lib/frontend/templates.dart
```diff
@@ -768,6 +735,45 @@ class TemplateService {
...
+      }
+      String url;
+      if (searchQuery != null) {
+        final newQuery = searchQuery.change(
+            platformPredicate: tabPlatform == null
+                ? new PlatformPredicate()
+                : new PlatformPredicate(required: [tabPlatform]));
+        url = newQuery.toSearchLink();
+      } else {
+        url = new Uri(path: '/packages', queryParameters: params).toString();
```
mkustermann: Move the computation of `params` into this `} else {` branch, since it's only used here.

Issue #468, opened 2017-10-25
Data used from GCS should be accessed in O(1) instead of O(n) - i.e. don't rely on bucket.list()


mkustermann:
We should never use linear scanning for finding objects in GCS buckets via `bucket.list()` but use a fixed name which contains whatever we need and have the historic values named by timestamp.

isoos:
Is there a Dart API for renaming a file in a bucket? I can't find it.

The handling of search snapshots [deletes](https://github.com/dart-lang/pub-dartlang-dart/blob/master/app/lib/search/backend.dart#L278) the older entries. If we were to do the same for the popularity data, the issue wouldn't be that critical, would it?
(For historical storage we could upload the same data in another bucket too.)

mkustermann:
There's at least [Storage.copy](https://www.dartdocs.org/documentation/gcloud/0.4.0+1/gcloud.storage/Storage/copyObject.html) which is atomic and should override the destination if it already exists. If multiple writer do the same thing  concurrently we might get any of them, but I don't see this as a problem. (The GCS api exposes also a compare-and-write, but I don't think our dart api exposes this).

It is **generally a bad idea to list the contents of buckets** and we have **no reason to do so**.

kevmoo:
Considered critical for https://github.com/dart-lang/pub-dartlang-dart/milestone/1 ?

kevmoo:
Blocked by https://github.com/dart-lang/pub-dartlang-dart/issues/473

kevmoo:
I closed https://github.com/dart-lang/pub-dartlang-dart/issues/473

See https://github.com/dart-lang/pub-dartlang-dart/issues/473#issuecomment-339864510

Since we can enable versioning, I think it's fine to delete the code that looks for json files – just to make things simple.

SGTY @isoos ?

isoos:
sgtm. I will modify the search snapshot too (to upload / lookup a single file), could you please enable the same for that bucket too?

kevmoo:
> could you please enable the same for that bucket too?

Versions? Sure...

kevmoo:
Done for gs://dartlang-pub--search-snapshot/

kevmoo:
@isoos @mkustermann – can we close this?

isoos:
I would keep it open until the last related code is removed, which should be done in #499.

mkustermann:
Strictly speaking I would have preferred having the old objects be there with a date in the name instead of using bucket versioning, but well ...

PR #472, opened 2017-10-25
Search snapshot to try loading the latests first. #468

isoos:
I'd keep the listing for the transitioning period, but also could keep it for fallback.


mkustermann:
Thanks!

PR #477, opened 2017-10-26
Fetching popularity data with the latest file 

isoos:


app/lib/search/backend.dart
```diff
@@ -166,7 +167,29 @@ class PopularityStorage {
...
   Future fetch() async {
-    _logger.info('Updating popularity storage...');
+    Future<Map> load(String path) async {
+      _logger.info('Loading popularity data: $path');
+      try {
+        Stream<List<int>> stream = bucket.read(path);
+        if (path.endsWith('.gz')) {
+          stream = stream.transform(_gzip.decoder);
+        }
+        final String json = await stream.transform(UTF8.decoder).join();
```
mkustermann: Same comment as with the other code, the `.join()` is an unnecessary operation since the json decoder can act on chuncked input:

```dart
return await stream.transform(UTF8.decoder).transform(JSON.decoder).single;
```

PR #478, opened 2017-10-26
Code cleanup: JSON.decoder.

isoos:


mkustermann:
Maybe do the same for the code which downloads popularity stats from GCS?

isoos:
> Maybe do the same for the code which downloads popularity stats from GCS?

That's line 177-180.

PR #480, opened 2017-10-27
mockScores are `num` - could be int. Handle conversion

kevmoo:
To avoid crashes in checked mode

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -150,7 +150,7 @@ class SearchBackend {
 }
 
 class PopularityStorage {
-  final String _latestPath = 'popularity-latest.json.gz';
+  static final String _latestPath = 'popularity-latest.json.gz';
   final Storage storage;
   final Bucket bucket;
 
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -90,8 +90,9 @@ class SearchBackend {
       if (pv == null) continue;
 
       final analysisView = analysisViews[i];
-      final double popularity =
-          popularityStorage.lookup(pv.package) ?? mockScores[pv.package] ?? 0.0;
+      final double popularity = popularityStorage.lookup(pv.package) ??
+          mockScores[pv.package]?.toDouble() ??
+          0.0;
       final double maintenance = _calculateMaintenance(p, pv);
 
       results[i] = new PackageDocument(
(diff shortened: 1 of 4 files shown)


app/lib/search/index_simple.dart
```diff
@@ -30,6 +30,21 @@ class SimplePackageIndex implements PackageIndex {
...
   @override
+  Map<String, dynamic> get debugInfo {
+    final data = {
+      'packageCount': _packages.length,
+      'lastUpdated': _lastUpdated?.toIso8601String(),
+    };
+
+    if (_lastUpdated != null) {
+      data['lastUpdateDelta'] =
+          new DateTime.now().difference(_lastUpdated).toString();
```
mkustermann: Since `_lastUpdated` is utc, you should also use utc for `now`.

app/lib/search/backend.dart
```diff
@@ -233,7 +234,7 @@ class PopularityStorage {
     final summary = new Summary(rawTotals.values);
     for (String package in rawTotals.keys) {
       final int raw = rawTotals[package];
-      _values[package] = summary.bezierScore(raw);
+      _values[package] = summary.bezierScore(raw).toDouble();
```
mkustermann: I would assume a non-linear interpolation based on bezier curves returns **always** a double

app/lib/search/backend.dart
```diff
@@ -149,11 +150,11 @@ class SearchBackend {
 }
 
 class PopularityStorage {
-  final String _latestPath = 'popularity-latest.json.gz';
+  static final String _latestPath = 'popularity-latest.json.gz';
```
mkustermann: How about const?

app/lib/search/index_simple.dart
kevmoo: Doesn't affect delta, since timezones are included.

When you hit it right after an update it shows as ~seconds – not hours.

app/lib/search/backend.dart
kevmoo: There's a non-zero cost in the VM to try to canonicalize const values IIRC, so I generally avoid 'em unless there's a big win.

Not like this change is huge, either. Not many instances of `PopularityStorage` during a run...

app/lib/search/backend.dart
kevmoo: See https://github.com/dart-lang/pana/pull/114

PR #484, opened 2017-10-27
Retry analysis (one-time only) if there has been a tool error.

isoos:


app/lib/analyzer/pana_runner.dart
```diff
@@ -31,27 +31,36 @@ class PanaRunner implements TaskRunner {
...
+        if (tempDir != null) {
+          await tempDir.delete(recursive: true);
+        }
       }
+      return null;
+    }
+
+    Summary summary = await analyze();
+    if (summary == null || (summary.toolProblems?.isNotEmpty ?? false)) {
+      await new Future.delayed(new Duration(seconds: 15));
```
kevmoo: Put a log in here before we start...saying that we're waiting?

app/lib/analyzer/pana_runner.dart
```diff
@@ -31,27 +31,36 @@ class PanaRunner implements TaskRunner {
...
+        if (tempDir != null) {
+          await tempDir.delete(recursive: true);
+        }
       }
+      return null;
+    }
+
+    Summary summary = await analyze();
+    if (summary == null || (summary.toolProblems?.isNotEmpty ?? false)) {
+      _logger.info('Retrying $task...');
```
mkustermann: Ideally I would think this is the exceptional case and we emit it as a warning instead of an info. How often do you think, does this happen?

isoos: No idea about the frequency. There are two kind of packages that will fall into this: the really outdated ones, which will have the issues even on the retry, and the ones that have only a transient issue, like `pub update` failed. I think there is no reason to call warning here, because there is a lot of old packages that would trigger it.

PR #486, opened 2017-10-27
Add side package to handle popularity format

kevmoo:
Allows an easy dependency (using git) without publishing something new


kevmoo:
FYI – this does change the format. Numbers are now numbers 😄

kevmoo:
Thoughts, @isoos ?

isoos:
Is it in a side project because the export script references it? If that script is independent, I'd rather have it in the normal project structure, otherwise looks ok for me.

.gitignore
```diff
@@ -14,3 +14,6 @@ packages
 
 # Or local files with secrets
 key.json
+
+# Just for pkg/popularity
+popularity/pubspec.lock
```
mkustermann: Why not `pkg/popularity/pubspec.lock` then?

app/lib/search/backend.dart
```diff
@@ -15,6 +15,7 @@ import 'package:gcloud/db.dart';
 import 'package:gcloud/storage.dart';
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:json_annotation/json_annotation.dart';
+import 'package:_popularity/popularity.dart';
```
mkustermann: If you use an underscore here, please also do it in the directory name.

app/lib/search/backend.dart
```diff
@@ -187,14 +188,12 @@ class PopularityStorage {
 
   void _updateLatest(Map raw) {
     final Map<String, int> rawTotals = {};
+    final popularity = new PackagePopularity.fromJson(raw);
     final List items = raw['items'];
```
mkustermann: Remove this `items`, since we already have `popularity.items` now (use it below in the logging message)?

app/pubspec.yaml
```diff
@@ -5,6 +5,8 @@ description: The pub.dartlang.org website.
 environment:
   sdk: ">=1.25.0-dev.9.0 <2.0.0"
 dependencies:
+  _popularity:
+    path: ../popularity
```
mkustermann: Could we be consistent regarding:
  * package name and directory name, use `_popularity` everywhere
  * comments and location, use `pkg/_popularity` everywhere (and move it into subfolder, it's better if all side packages are in `pkg/`)

popularity/.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: dart
```
mkustermann: Since you have modified `.travis.yml` above, I don't think we need this file.

popularity/lib/popularity.dart
```diff
@@ -0,0 +1,44 @@
...
+library package_popularity;
+
+import 'package:json_annotation/json_annotation.dart';
+
+part 'popularity.g.dart';
+
+@JsonSerializable()
+class PackagePopularity extends Object with _$PackagePopularitySerializerMixin {
+  @JsonKey(name: 'date_first')
+  final DateTime dateFirst;
```
mkustermann: Wouldn't be `firstDate` and `lastDate` or maybe `from` and `to` be better?

popularity/lib/popularity.dart
```diff
@@ -0,0 +1,44 @@
...
+
+  @JsonKey(name: 'date_last')
+  final DateTime dateLast;
+
+  final List<PackageInfo> items;
```
mkustermann: Since it's assumed we get exactly one `PackageInfo` per package name, which this `List` doesn't enforce, you could make it a `Map` instead with the keys being the package name.

popularity/tool/build.dart
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env dart --checked
```
mkustermann: Maybe update the `README.md` file and update it to include instructions for re-building the generated code here.

kevmoo:
@isoos 

> Is it in a side project because the export script references it? If that script is independent, I'd rather have it in the normal project structure, otherwise looks ok for me.

The script references a bit of internal infrastructure, so I'm keeping it internal...

.gitignore
kevmoo: If a package is meant to be consumed – not an app – we generally don't include the lock file.

...the pubspec.lock file we really care about is in the pub site.

app/lib/search/backend.dart
kevmoo: ack

popularity/.travis.yml
kevmoo: This file is used by `mono_repo` to generate the root `.travis.yml` and the associated script.

popularity/lib/popularity.dart
kevmoo: I hear ya. This file is effectively straight out of BigQuery...trying to minimize processing on our end. Will add a to-do

popularity/lib/popularity.dart
kevmoo: Eh...I like 'em how they are 🤷‍♂️

popularity/lib/popularity.dart
kevmoo: ...on second thought, I'll fix this.

app/pubspec.yaml
kevmoo: Done

kevmoo:
@mkustermann @isoos PTAL

PR #487, opened 2017-10-27
Use pathSegments and Request.change

kevmoo:


app/lib/analyzer/handlers.dart
```diff
@@ -16,15 +16,14 @@ import 'models.dart';
...
     '/debug': debugHandler,
-  }[path];
+  }[request.requestedUri.path];
 
   if (handler != null) {
     return handler(request);
-  } else if (path.startsWith('/packages/')) {
-    return packageHandler(request);
+  } else if (request.url.pathSegments.first == 'packages') {
+    return packageHandler(request.change(path: 'packages'));
```
isoos: This seems to clear the `package/version` part from the path, is that the intention?

kevmoo: Yup! Let's the handler live it's in sub-path world without caring about the global path mapping

isoos: If we won't use the `path`, why change it?

kevmoo: It removes it from the URL – so the handler doesn't have to think about it.

See the code below that changed. You no longer have to ignore the first package segment – it's just not there.

isoos: I see that below, but I find it confusing that we are changing the path here, as if there is a reason or use for it.

kevmoo:
`path` is weird here. It's not saying "change the path to this" so much as
updating the request.url to be relative to `path` – so it's easier to
reason about the items underneath.

kevmoo:
Not a huge deal, @isoos – can close.

String-based path munging makes me 🤢

PR #500, opened 2017-10-31
Split out flutter scores

kevmoo:
Addresses the first part of https://github.com/dart-lang/pub-dartlang-dart/issues/466

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -190,9 +190,7 @@ class PopularityStorage {
     final Map<String, int> rawTotals = {};
     final popularity = new PackagePopularity.fromJson(raw);
     popularity.items.forEach((pkg, item) {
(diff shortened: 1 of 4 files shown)


kevmoo:
Popularity data has been uploaded to prod and dev buckets under `v3`

mkustermann:
Guess this means there is a bias of flutter users to certain packages and other users to certain packages.

Looks fine, but by generalization one would assume there is a similar bias between web users and command line users, it's just we can't distinguish them atm.

kevmoo:
> Guess this means there is a bias of flutter users to certain packages and other users to certain packages.

yup!

> Looks fine, but by generalization one would assume there is a similar bias between web users and command line users, it's just we can't distinguish them atm.

We need tooling hints for that. Maybe, someday

PR #502, opened 2017-11-02
Lightweight analysis extract for better caching in user-facing views.

isoos:
While an analysis data object could be pretty large (depending on the number of libraries and their errors), this is only the strictly required data for the list views. Less data to push through the network, less to parse.

### DIFF
--- a/app/lib/shared/analyzer_client.dart
+++ b/app/lib/shared/analyzer_client.dart
@@ -43,6 +43,30 @@ class AnalyzerClient {
     return new AnalysisView(await getAnalysisData(key));
   }
 
+  Future<List<AnalysisExtract>> getAnalysisExtracts(
+      Iterable<AnalysisKey> keys) {
+    return Future.wait(keys.map(getAnalysisExtract));
+  }
+
+  Future<AnalysisExtract> getAnalysisExtract(AnalysisKey key) async {
+    if (key == null) return null;
+    final String cachedExtract =
+        await analyzerMemcache?.getExtract(key.package, key.version);
+    if (cachedExtract != null) {
+      return new AnalysisExtract.fromJson(JSON.decode(cachedExtract));
+    }
+    final view = await getAnalysisView(key);
+    final extract = new AnalysisExtract(
+      // TODO: set popularity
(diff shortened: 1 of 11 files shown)


app/lib/shared/analyzer_memcache.dart
```diff
@@ -40,10 +47,23 @@ class AnalyzerMemcache {
     return _data.setText(_dataKey(package, version, panaVersion), content);
   }
 
+  Future<String> getExtract(String package, String version) {
+    return _data.getText(_extractKey(package, version));
```
mkustermann: Shouldn't this be `_extract.getText()`? (ditto below)

app/lib/shared/analyzer_client.dart
```diff
@@ -43,6 +43,30 @@ class AnalyzerClient {
...
+
+  Future<AnalysisExtract> getAnalysisExtract(AnalysisKey key) async {
+    if (key == null) return null;
+    final String cachedExtract =
+        await analyzerMemcache?.getExtract(key.package, key.version);
+    if (cachedExtract != null) {
+      return new AnalysisExtract.fromJson(JSON.decode(cachedExtract));
+    }
+    final view = await getAnalysisView(key);
+    final extract = new AnalysisExtract(
```
mkustermann: The CL description says: `Less data to push through the network, less to parse`, even though this is actually pulling the entire content from the analysis service.

It's just pruning it on the client side and store/load the reduced data from memcache. It's fine with me, but we might as well do that on the server side.

app/lib/shared/analyzer_memcache.dart
isoos: good catch, thanks!

app/lib/shared/analyzer_client.dart
isoos: The difference comes from the amount of data: in the previous case, we pulled the entire analysis data from memcache or from the network. With this, there is a high chance that after a while, the most frequently accessed ones will be in the cache, with the minimal data amount. Otherwise the mechanism is just the same.

PR #512, opened 2017-11-02
Upgrade markdown to 1.0.0.

isoos:


app/lib/shared/markdown.dart
```diff
@@ -8,8 +8,10 @@ import 'package:path/path.dart' as p;
 
 final Logger _logger = new Logger('pub.markdown');
 
-final List<m.BlockSyntax> _blockSyntaxes =
```
kevmoo: Could/should we just use the GitHub extension set directly – then add the Header/Setext things?

app/lib/shared/markdown.dart
```diff
@@ -8,8 +8,10 @@ import 'package:path/path.dart' as p;
...
-    m.ExtensionSet.gitHub.blockSyntaxes.toList();
+final List<m.BlockSyntax> _blockSyntaxes = []
+  ..addAll(m.ExtensionSet.gitHub.blockSyntaxes)
+  ..add(const m.HeaderWithIdSyntax())
```
isoos: The `markdown` lib 1.0 removed these from the github extension set, but I think github does generate the header ids and people rely on it for page-internal links, so I added those back here.

app/lib/shared/markdown.dart
isoos: We do remove the `InlineHtmlSyntax` from the github extension set. Not sure the reason, maybe to prevent HTML injection?

kevmoo:
...but we don't do it on the dartdoc site, though.

*Sigh*

isoos:
> ...but we don't do it on the dartdoc site, though.

I'm all for consistency, but I'd prefer to be on the (HTML-)safe side on this one.

kevmoo:
Yup. Let's go w/ this for now...

PR #513, opened 2017-11-03
Unify search and package listing.

isoos:
There are two minor follow-up tasks with the URL handling (added TODO for them), but I felt doing them in the same PR will make it too bloated.


app/lib/frontend/handlers.dart
```diff
@@ -446,19 +435,30 @@ Future<shelf.Response> _packagesHandlerHtmlV2(
     shelf.Request request, String platform) async {
   final int page = _pageFromUrl(request.url);
   final int offset = PackageLinks.resultsPerPage * (page - 1);
+  final queryText = request.requestedUri.queryParameters['q'] ?? '';
+  // TODO: handle package-prefix
+
+  // TODO: handle sort order
```
kevmoo: Yeah, this makes this pretty broken

app/test/frontend/golden/v2/web_landing_page.html
```diff
@@ -60,9 +60,8 @@ <h1 class="_visuallyhidden">Dart pub</h1>
     <main class="home-banner">
       <h2 class="_visuallyhidden">Dart package manager</h2><img class="logo" src="/static/v2/img/site-logo.svg" alt="Dart package manager">
       <p class="text">Find and use packages to build <a href="/experimental/flutter">Flutter</a>, <a href="/experimental/web">web</a> and <a href="/experimental/server">server</a> apps with <a target="_blank" href="https://www.dartlang.org">Dart</a>.</p>
-      <form class="search-bar" action="/experimental/search">
+      <form class="search-bar" action="/experimental/server/packages">
```
kevmoo: shouldn't the web landing page point to web packages?

app/test/frontend/golden/v2/pkg_show_page_flutter_plugin.html
```diff
@@ -59,9 +59,8 @@ <h1 class="_visuallyhidden">Dart pub</h1>
 
   <div class="_banner-bg">
     <main class="package-banner">
-      <form class="search-bar" action="/experimental/search">
+      <form class="search-bar" action="/experimental/packages">
```
kevmoo: /experimental/flutter/packages?

app/lib/frontend/handlers.dart
isoos: Yeah, the default order is context-dependent.

app/test/frontend/golden/v2/pkg_show_page_flutter_plugin.html
isoos: This is a package show page, and we don't have good platform to chose (especially if it is not flutter-only).

app/test/frontend/golden/v2/web_landing_page.html
isoos: good catch, bad input in test, fixing.

PR #515, opened 2017-11-03
Rank by platform specificity.

isoos:
This is a symmetrical score for all of the platforms.


app/test/search/platform_rank_test.dart
```diff
@@ -0,0 +1,87 @@
...
+        platforms: ['flutter', 'server'],
+      ));
+      await index.addPackage(new PackageDocument(
+        package: 'json_3',
+        platforms: KnownPlatforms.all,
+      ));
+      await index.merge();
+    });
+
+    test('text search', () async {
```
kevmoo: Make this two tests? Easier to parse...

PR #519, opened 2017-11-06
Handle sort param in v2 listing page.

isoos:


app/lib/frontend/handlers.dart
```diff
@@ -435,12 +435,24 @@ Future<shelf.Response> _packagesHandlerHtmlV2(
...
+  String queryText = request.requestedUri.queryParameters['q'] ?? '';
+  String packagePrefix = request.url.queryParameters['pkg-prefix'];
+
+  if (queryText != null) {
+    final Match pkgMatch = _packageRegexp.firstMatch(queryText);
+    if (pkgMatch != null) {
+      // only if query attribute was not specified
+      if (packagePrefix == null) {
+        packagePrefix = pkgMatch.group(1).trim();
+        if (packagePrefix.isEmpty) packagePrefix = null;
```
mkustermann: The regexp guarantees `packagePrefix` will never be empty.

app/lib/frontend/handlers.dart
```diff
@@ -435,12 +435,24 @@ Future<shelf.Response> _packagesHandlerHtmlV2(
...
   final int offset = PackageLinks.resultsPerPage * (page - 1);
-  final queryText = request.requestedUri.queryParameters['q'] ?? '';
-  // TODO: handle package-prefix
+  String queryText = request.requestedUri.queryParameters['q'] ?? '';
+  String packagePrefix = request.url.queryParameters['pkg-prefix'];
+
+  if (queryText != null) {
+    final Match pkgMatch = _packageRegexp.firstMatch(queryText);
+    if (pkgMatch != null) {
+      // only if query attribute was not specified
```
mkustermann: Maybe remove this comment and say rather the following (at line 441):
```
// If queryText is of the form "a package:foo/a/b/c.dart z" we try to set
//     pkg-prefix=foo
//     queryText= a /a/b/c.dart z
```
Even though I'm not quite sure if that's what's happening further down with `queryText.replaceFirst`, since the regexp is defined in a weired way:

```dart
final _packageRegexp = new RegExp('package:([_a-z0-9]+)\\*?', caseSensitive: false);
```

Does this mean that `package:foo` must be followed by 0 or more backslashes, which are then marked as optional with the question mark? Somehow I think this should be rather just `package:([_a-z0-9]+)`.

isoos:
Good points about pkg-prefix, fixing...

Thinking about it a bit more, it may be better to never use the `pkg-prefix` parameter, and extract the prefix filtering only at the time of the search, keeping the data in query text. (I'd do that in a follow-up change.)

app/lib/frontend/handlers.dart
```diff
@@ -250,11 +250,7 @@ Future<shelf.Response> searchHandler(shelf.Request request) async {
 
   final Match pkgMatch = _packageRegexp.firstMatch(queryText);
   if (pkgMatch != null) {
-    // only if query attribute was not specified
-    if (packagePrefix == null) {
-      packagePrefix = pkgMatch.group(1).trim();
-      if (packagePrefix.isEmpty) packagePrefix = null;
-    }
+    packagePrefix = pkgMatch.group(1).trim();
```
mkustermann: The trim here is not strictly necessary, since it's guaranteed to have only `[_a-z0-9]` characters (and is also not empty, since it's guaranteed to have at least one).

app/lib/frontend/handlers.dart
```diff
@@ -435,12 +431,20 @@ Future<shelf.Response> _packagesHandlerHtmlV2(
...
   final int offset = PackageLinks.resultsPerPage * (page - 1);
-  final queryText = request.requestedUri.queryParameters['q'] ?? '';
-  // TODO: handle package-prefix
+  String queryText = request.requestedUri.queryParameters['q'] ?? '';
+  String packagePrefix = request.url.queryParameters['pkg-prefix'];
+
+  if (queryText != null) {
+    final Match pkgMatch = _packageRegexp.firstMatch(queryText);
+    if (pkgMatch != null) {
+      packagePrefix = pkgMatch.group(1).trim();
```
mkustermann: Same here.

PR #521, opened 2017-11-06
Remove pkg-prefix from user-facing URLs

isoos:
With this change, the user-facing URLs won't have `pkg-prefix` attribute handling, and if there is such value, it will always be part of the query text. The parsing of the query text happens in the frontend, and it will use separate attributes to pass the values to the service.

It also simplifies the `platform` parsing.

### DIFF
--- a/app/lib/shared/search_service.dart
+++ b/app/lib/shared/search_service.dart
@@ -134,6 +134,9 @@ String serializeSearchOrder(SearchOrder order) {
   return order.toString().split('.').last;
 }
 
+final RegExp _packageRegexp =
+    new RegExp('package:([_a-z0-9]+)', caseSensitive: false);
+
 class SearchQuery {
   final String text;
   final PlatformPredicate platformPredicate;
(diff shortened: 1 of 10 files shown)


app/lib/shared/search_service.dart
```diff
@@ -134,6 +134,9 @@ String serializeSearchOrder(SearchOrder order) {
   return order.toString().split('.').last;
 }
 
+RegExp _packageRegexp =
```
kevmoo: final?

app/lib/shared/search_service.dart
```diff
@@ -217,6 +247,19 @@ class SearchQuery {
   bool get hasPackagePrefix =>
       packagePrefix != null && packagePrefix.isNotEmpty;
 
+  bool get hasCompositeText => hasText || hasPackagePrefix;
+
+  String get compositeText {
+    final List<String> parts = <String>[];
+    if (hasText) parts.add(text);
+    if (hasPackagePrefix) parts.add('package:$packagePrefix');
+    if (parts.length == 1) {
```
kevmoo: This could just be `return parts.join(' ');` – the length == 1 case is subsumed

app/lib/shared/search_service.dart
```diff
@@ -243,17 +286,18 @@ class SearchQuery {
...
-    }
     if (order != null) {
       final String paramName = v2 ? 'sort' : 'order';
       params[paramName] = serializeSearchOrder(order);
     }
     if (page != null) {
       params['page'] = page.toString();
     }
-    return new Uri(path: path, queryParameters: params).toString();
+    if (params.isEmpty) {
```
kevmoo: Isn't the empty params cased subsumed in the general case below?

Or is that a hanging `?`?

isoos: Yeah, the hanging `?` is the issue. Either the map should become `null`, or returning the `path` directly .

kevmoo: kk

PR #523, opened 2017-11-06
Link license file to the package repository (for github links)

isoos:
Decided to push further URL-handling logic to pana, which should also check for existence at the time of the analysis: https://github.com/dart-lang/pana/issues/120

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -10,6 +10,7 @@ import 'dart:math';
 
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:mustache/mustache.dart' as mustache;
+import 'package:path/path.dart' as path;
 
 import '../shared/analyzer_client.dart';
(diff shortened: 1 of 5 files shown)


kevmoo:
Why this? Seems it could be brittle of the source is moved, etc...

isoos:
> Why this? Seems it could be brittle of the source is moved, etc...

It is brittle, but there was a discussion around it a few months back, that it would be nice to link directly to the license file, and this covers the low-hanging fruits. After pana will create + check the URL, it will be more reliable to link to the file (at worst moving URLs will be corrected in the next analysis, no later than a month).

Btw. this is similar to the URL rewrite we have in our [markdown processing](https://github.com/dart-lang/pub-dartlang-dart/blob/master/app/lib/shared/markdown.dart): in readme.md there may be relative file links which we rewrite and point to the homepage_url + relative url.

mkustermann:
I'll leave this for @kevmoo

PR #532, opened 2017-11-07
Use memcached inside the containers again

mkustermann:
We need to **force-downgrade package:appengine** in order for this to work,
since package:appengine will first respect the environment variables
(given from GAE) and only later fall back to the local memcached (if
available)

We also reduce the expiration dates so all serving instances are
up-to-date after 1 minute.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,11 @@
 FROM google/dart-runtime-base:1.25.0-dev.9.0
 
- # After install we remove the apt-index again to keep the docker image diff small.
- RUN apt-get update && \
-     apt-get upgrade -y && \
-     apt-get install -y git unzip && \
-     rm -rf /var/lib/apt/lists/*
+# We install memcached and remove the apt-index again to keep the
+# docker image diff small.
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install -y git memcached unzip && \
+    rm -rf /var/lib/apt/lists/*
 
 # Let the pub server know that this is not a "typical" pub client but rather a bot.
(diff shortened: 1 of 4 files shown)


mkustermann:
Since the issue has been fixed within 3 hours we could decide to continue to use the memcache service. So we don't have to land this PR.

Let me know what you think.

isoos:
> Let me know what you think.

Please check the email I've sent a couple of hours ago. I don't think that the root cause of the outage was (a) memcache.get() returning errors as ligitimate content and (b) we weren't protected against it. While most of the memcache use we had was HTML in and HTML out, this wasn't an issue we'd detect, but since Analysis JSON is in memcache, we did fail on parsing the bad content.

I don't think the timeout decrease is necessary, and I'd rather use one consistent memcache or no memcache at all (assuming that the infrastructure will be back in a day at most). Inconsistency across local memcache instances was a hard problem to solve, and I'd rather trade temporary slowness for it.

mkustermann:
This PR is due to the memcache service we use being unreliable, not to bugs in code. More than a week ago I've already mentioned offline that we have plenty of errors in the logs due to that very same memcache service.

So the question is whether we are happy with this "alpha version" with occasional issues, or we want to have 100% caching reliability (which we can get with this CL - with the downside of 1 minute eventual consistency).

kevmoo:
I hate the eventual consistency thing. I causes issues w/ Travis and other places where folks expect to see the latest state.

I'd rather we be resilient in the face of memcache issues

mkustermann:
Then let's close it for now - Though the fact that our logs are spammed with nonworking memcache calls is very suboptimal

PR #534, opened 2017-11-07
Don't store Analysis regression for a short period.

isoos:
Before:
- During the 3-hour outage, a few Analysis may have completed with tool failures, and we store them as aborted or failed. The task scheduler will pick these up after a day or two at most, and re-analysis will make them succeed.
- Worst case: a good package will be treated as bad for a few days at most.

After:
- If there is an Analysis that is a regression (e.g. it shows tool failure, while the previous one doesn't have failures), we won't store it for a set amount of days. Instead, the task scheduler will schedule it again, and we analyze it the next day, and the next. We keep doing it until either it is successful, or the age threshold is over, either way storing the result.
- Worst case: a bad package will be treated as good for two weeks longer than it usually would.  Unlikely though, because dependency changes don't trigger tool failures, unless we delete one of its dependencies.

### DIFF
--- a/app/lib/analyzer/backend.dart
+++ b/app/lib/analyzer/backend.dart
@@ -10,6 +10,7 @@ import 'package:logging/logging.dart';
 
 import '../frontend/models.dart';
 import '../shared/analyzer_memcache.dart';
+import '../shared/analyzer_service.dart';
 import '../shared/utils.dart';
 
 import 'models.dart';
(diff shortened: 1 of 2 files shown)


app/lib/analyzer/backend.dart
```diff
@@ -111,19 +118,22 @@ class AnalysisBackend {
         inserts.add(package);
       }
 
-      final DateTime prevTimestamp = version?.analysisTimestamp;
+      final Duration ageDiff = version == null
+          ? null
+          : version.analysisTimestamp.difference(version.analysisTimestamp);
```
mkustermann: I think you might have a typo here, since this is basically doing `version.analysisTimestamp - version.analysisTimestamp` which is always `0`.

isoos: yeah, fixing...

isoos:
> Isn't 45 days a very long time for this?

The worst case between subsequent analysis is roughly 31-35 days (30+ to get triggered, and if all of the packages are in the queue for some reason at the same time it may take days). Anything above (the maximum task scheduling threshold + a few days) could reasonable, 45 is as good as 40 or 50.

PR #535, opened 2017-11-07
Upgrade package dependencies to pick-up fix for package:memcache

mkustermann:


app/pubspec.lock
```diff
@@ -470,4 +470,4 @@ packages:
     source: hosted
     version: "2.1.13"
 sdks:
-  dart: ">=1.25.0-dev.9.0 <=2.0.0-dev.6.0"
+  dart: ">=1.25.0-dev.9.0 <=2.0.0-edge.74f4c4f3e6a318889107d152c54fbb4fa50dbafc"
```
isoos: this looks weird

mkustermann: I've done a `docker build .` to ensure the container builds, which includes running `pub get`. So it's fine.

The reason it looks weird is that the `pub` I happened to have in `PATH` is one which was built from the bleeding edge branch.

PR #538, opened 2017-11-08
Handle more author format.

isoos:


mkustermann:
Would you mind adding a test for this? It would allow all of us to see in the test which kinds of author formats we support.

isoos:
> Would you mind adding a test for this? It would allow all of us to see in the test which kinds of author formats we support.

Sure, but in this change it will be a bit weird. I'm planning to extract the author parsing into pubspec (and eventually move it into pana), and that would be a much better place for it. Can we merge this to unblock the rendering and do the refactor + testing in a subsequent change?

PR #539, opened 2017-11-08
'More packages...' links at the bottom of the landing pages.

isoos:
This adds the "Top packages..." and "More packages..." links, I'm not sure what to do with the "Newest" column there, because the site polish doc says only the 'top' and 'recent' sort method should be valid.

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -568,7 +568,15 @@ class TemplateService {
     String newestHtml,
     String platform,
   ) {
+    final String platformName = _formattedPlatformName(platform);
     final values = {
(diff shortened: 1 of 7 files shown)


app/lib/frontend/templates.dart
```diff
@@ -984,5 +992,15 @@ final Map<String, Map> _logoData = const {
   },
 };
 
+String _formattedPlatformName(String platform) {
```
isoos: This is also under-specified. An alternative could be to uppercase the platform names by default.

mkustermann: That's something for @kevmoo

PR #540, opened 2017-11-08
Fast local cache of analysis extracts.

isoos:
This is a micro-optimization, but it helps with (currently) non-cached listing pages a lot.

### DIFF
--- a/app/lib/shared/analyzer_client.dart
+++ b/app/lib/shared/analyzer_client.dart
@@ -15,6 +15,7 @@ import 'package:pub_dartlang_org/analyzer/versions.dart';
 
 import 'analyzer_memcache.dart';
 import 'analyzer_service.dart';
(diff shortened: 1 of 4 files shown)


app/lib/shared/analyzer_client.dart
```diff
@@ -33,6 +34,7 @@ final Logger _logger = new Logger('pub.analyzer_client');
 /// Client methods that access the analyzer service and the internals of the
 /// analysis data. This keeps the interface narrow over the raw analysis data.
 class AnalyzerClient {
+  final Map<AnalysisKey, AnalysisExtract> _extractCache = {};
```
kevmoo: How do we make sure this doesn't grow unbounded?

isoos: good point, added a max limit

PR #541, opened 2017-11-08
Extract Author parsing + tests.

isoos:


isoos:
At first I wanted to have the Pubspec class return only the Author type, but then the `PackageView` needs to use this type too, and I'd rather keep it simple for now. Maybe later, when I'll have some time for the pubspec-consolidation.

app/lib/frontend/model_properties.dart
```diff
@@ -156,3 +156,39 @@ class FileObject {
 
   FileObject(this.filename, this.text);
 }
+
+final _authorsRegExp = new RegExp(r'^\s*(.+)\s+<(.+)>\s*$');
```
kevmoo: nit: if this is only used in the class, I prefer static – but not a biggy

Issue #341, opened 2017-09-11
Search service should not depend on analyzer


isoos:
A [recent PR](https://github.com/dart-lang/pub-dartlang-dart/pull/330) triggered a discussion and the plan is to refactor how search works and remove its dependency on analyzer.

My first thought is to:
- Introduce a `PackageSearch` object in Datastore, named similarly to `PackageAnalysis`.
- It would contain the data that needs to be indexed (`PackageDocument`) for each package.
- It would be populated and updated by the `analyzer` service, every time it completes analysis on the latest stable version.
- It would be polled by the `search` service roughly the same way the current pollers work, looking for recent updates via timestamp.
- The `search` service's snapshot is still worth to keep, so there is no change there.

I don't see a reliable alternative using the new 'global' memcache, because there is always a chance that a new version data gets evicted before any of the `search` instances pick it up, and we may wait for a month for a new analysis (if nothing else triggers it).

WDYT @mkustermann @kevmoo?

isoos:
Wrt. the [trigger re-analysis of packages that depend on a new version](https://github.com/dart-lang/pub-dartlang-dart/issues/317), with the proposed solution above, the service dependencies would be these:
- `frontend` -> `search`, `analyzer`, `dartdoc` (trigger only)
- `analyzer` -> `search` (query dependees), `dartdoc` (trigger only)
- `search` (no service deps)
- `dartdoc` (no service deps)

kevmoo:
SGTM

mkustermann:
Firstly, we know for sure that there is some kind of `search -> analyzer` interaction, since search uses analyzer results for ranking. It can be a weak dependency, but it is there!

Yet, I still don't completely understand why there should be any `analyzer -> search` interaction at all. Analyzer doesn't need any full-text-search information (ever).

Generally speaking, it's better to not introduce a dependency, just because the other thing happens to have some information that could be useful somewhere else. This leads quickly to spaghetti code, which is even worse on bigger components. Clear separation of concerns seems much better to me and much more maintainable.

[...]

PR #543, opened 2017-11-08
Search service tracks new Analysis instances to update index.

isoos:
Related to the discussions in #341. After this change, we could remove the notications entirely, reducing the direct dependencies between the services.

### DIFF
--- a/app/bin/service/search.dart
+++ b/app/bin/service/search.dart
@@ -17,6 +17,7 @@ import 'package:pub_dartlang_org/shared/analyzer_memcache.dart';
 import 'package:pub_dartlang_org/shared/configuration.dart';
 import 'package:pub_dartlang_org/shared/task_client.dart';
 import 'package:pub_dartlang_org/shared/task_scheduler.dart';
+import 'package:pub_dartlang_org/shared/task_sources.dart';
 
 import 'package:pub_dartlang_org/search/backend.dart';
 import 'package:pub_dartlang_org/search/handlers.dart';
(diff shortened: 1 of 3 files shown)


isoos:
The removal of the index update threshold is important, because `PackageDocument` creation is relatively cheap, and we don't want the index to lag behind too much.

app/bin/service/search.dart
```diff
@@ -75,6 +76,12 @@ void _main(int isolateId) {
       [
         new ManualTriggerTaskSource(taskReceivePort),
         new IndexUpdateTaskSource(db.dbService, batchIndexUpdater),
+        new DatastoreVersionsHeadTaskSource(
+          db.dbService,
+          TaskSourceModel.analysis,
+          sleep: const Duration(minutes: 10),
```
kevmoo: So every 10 minutes, what happens?

isoos: we poll the datastore to see if there are any new `Analysis` instances created

app/lib/analyzer/pana_runner.dart
```diff
@@ -89,11 +89,6 @@ class PanaRunner implements TaskRunner {
       // TODO: trigger re-analysis of packages depending on this one
     }
 
-    if (backendStatus.isLatestStable) {
-      // Do not await on the notification.
-      notificationClient.notifySearch(analysis.packageName);
```
isoos: update: I'm removing this in the same PR

PR #553, opened 2017-11-09
Search ranking with IDF

isoos:
Better cutoff for frequent terms, while still keeping the ability of typo-resistant search.


mkustermann:
From now on we should keep the ranking document up-to-date when making changes to the ranking. From the code alone it seems difficult to see. Would you mind updating the document?

isoos:
Sure, updated.

mkustermann:
Thanks.

PR #563, opened 2017-11-13
Apply new package name policies

isoos:
The initial `refresh()` takes a couple of seconds at most, the subsequent ones just a few hundred milliseconds.  I've skipped tests, because the logic to test is ~4 lines, and straightforward, while adding it to the mocks will be a lot, without much additional value. If the policy goes ahead, we'd likely transform the uppercased-ones to lowercase and this code will be removed.

### DIFF
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -42,6 +42,14 @@ void registerBackend(Backend backend) => ss.register(#_backend, backend);
 /// The active backend service.
 Backend get backend => ss.lookup(#_backend);
 
+/// Sets the existing package validator service.
+void registerExistingPackageValidator(ExistingPackageValidator validator) =>
+    ss.register(#_existingPackageValidator, validator);
+
+/// The existing package validator service.
+ExistingPackageValidator get existingPackageValidator =>
+    ss.lookup(#_existingPackageValidator);
+
 /// A callback to be invoked after an upload of a new [PackageVersion].
 typedef Future FinishedUploadCallback(models.PackageVersion pv);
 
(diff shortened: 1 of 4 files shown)


app/lib/frontend/backend.dart
```diff
@@ -695,6 +706,58 @@ Future<models.PackageVersion> parseAndValidateUpload(
   return version;
 }
 
+class ExistingPackageValidator {
+  final DatastoreDB _db;
+  final Set<String> _fullNames = new Set();
```
kevmoo: Inference is your friend. Put all the types on the right.

kevmoo:
@mkustermann ?

mkustermann:
I'm not a big fan of it:
   * Don't like the "seconds" or "hundered milliseconds" added in latency (in particular it's done lazily and not eagerly, i.e. it will have slow first upload)
   * Don't like that we need to scan the entire DB (e.g. what if an exception occurs during scanning of > 1 thousands datastore entities)

How about we just hardcode the set of upper case letter packages we have atm into the source and check whether the name is lowercase? (After a deployment of that, nobody can upload a new package with upper case names and the time between commit and deployment is negligible.)

This simplifies to logic to 5 lines of code!

isoos:
> How about we just hardcode the set of upper case letter packages we have atm into the source and check whether the name is lowercase?

Sounds good. Could you please send me the list of packages with uppercase? I can write a script for that if it helps, but I have access to prod.

isoos:
Closing in favor of https://github.com/dart-lang/pub-dartlang-dart/pull/575.

PR #566, opened 2017-11-13
Edited README; fix formatting and upgraded instructions to run locally

luanpotter:
Few minor fixes:

 * Added protocol so remote URL would work;
 * Fixed a few typos;
 * Updated instructions on how to build locally.

The current instructions were outdated, so I updated them to reflect how I was able to run locally.  If that's not the proper way, please let me know!

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 The server for hosting pub packages on [pub.dartlang.org](https://pub.dartlang.org)
 is implemented using AppEngine Custom Runtimes with Flexible environment
-(see [package:appengine](github.com/dart-lang/appengine) for more information about
+(see [package:appengine](https://github.com/dart-lang/appengine) for more information about
 Dart support for AppEngine).
 
 ## Local Development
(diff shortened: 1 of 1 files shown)


luanpotter:
I signed it!

kevmoo:
LGTM – will have @isoos take a look

kevmoo:
@luanpotter – may I ask why you're looking at this code?

luanpotter:
Sure, @kevmoo! :)

I was investigating [this issue](https://github.com/dart-lang/pub-dartlang-dart/issues/562).

I was trying to work out exactly what steps `pub-dartlang` uses to verify packages. Turns out, it runs an equivalent of the command `dartanalyzer .`, as @isoos pointed out, which indeed gives an error for [flame](https://github.com/luanpotter/flame); but that's because the supermixin flag is disabled (if you run `dartanalyzer . --supermixin`, it works).

So I was looking into it to see if it was possible for `pub-dartlang` to run its checks with `---supermixin` enabled for packages from `1.13` or higher, and maybe contribute with a fix. For that, I needed to run the whole thing locally.

kevmoo:
@luanpotter – I'd start with pana - https://pub.dartlang.org/packages/pana

MUCH easier.

isoos:
AFAICT supermixin does not work in dart2js or in dev compiler [link](https://www.dartlang.org/articles/language/mixins). If we would enable it, packages may be classified to work on the web, but in reality they would cause compiler errors.

README.md
```diff
@@ -16,29 +16,57 @@ Generally speaking it is very easy to run the entire application locally
...
-to be modified:
+To use a different Google Cloud Project, the `app/lib/shared/configuration.dart` needs
+to be modified. First, create a new `Configuration`, copying `Configuration._dev()`,
+but replacing `dartlang-pub-dev` with the name of a new Google Cloud project
+you will need to create:
+
+```dart
+  Configuration._localdev()
+      : projectId = '<gcloud-project>',
+        packageBucketName = '<gcloud-project>--pub-packages',
```
isoos: While this is correct, I think the better way would be to have a static initializer method that puts the <gcloud-project> into the Strings. Would you be interested to change that part in `configuration.dart` too?

luanpotter: Great idea, I will try to make those changes right away. Thanks for the feedback :)

PR #567, opened 2017-11-14
Simpler configuration

luanpotter:
Made local configuration easier, putting the project id into the strings, as suggested.

Also updated the README to reflect the new method for local development.

I'm not sure that's the approach you guys had in mind, so please let me know how to improve it!

### DIFF
--- a/app/lib/shared/configuration.dart
+++ b/app/lib/shared/configuration.dart
@@ -66,17 +66,18 @@ class Configuration {
         popularityDumpBucketName = PackagePopularity.bucketName(false),
         searchSnapshotBucketName = 'dartlang-pub--search-snapshot';
 
-  /// Create a configuration for development/staging deployment.
-  Configuration._dev()
-      : projectId = 'dartlang-pub-dev',
-        packageBucketName = 'dartlang-pub-dev--pub-packages',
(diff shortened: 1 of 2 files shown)


app/lib/shared/configuration.dart
```diff
@@ -66,17 +66,18 @@ class Configuration {
...
-            'https://dartdoc-dot-dartlang-pub-dev.appspot.com',
-        searchServicePrefix = 'https://search-dot-dartlang-pub-dev.appspot.com',
+  /// Base configuration for local development
+  Configuration._local(String projectId)
+      : projectId = projectId,
+        packageBucketName = '$projectId--pub-packages',
+        analyzerServicePrefix = 'https://analyzer-dot-$projectId.appspot.com',
+        dartdocServicePrefix = 'https://dartdoc-dot-$projectId.appspot.com',
+        searchServicePrefix = 'https://search-dot-$projectId.appspot.com',
         popularityDumpBucketName = PackagePopularity.bucketName(true),
```
isoos: This is in pkg/_popularity/lib/popularity.dart, and if you change that too, both the _dev() and the _prod() constructor can become just a call to _local(<project-id>)

PR #568, opened 2017-11-14
Rename V2 package tab IDs

isoos:
- analysis tab's name (`score`) was not following the same pattern
- `pub-tab-*` was hard to read, `-*-tab-` is better, and seems to be unique enough that we won't have name collisions from user-provided content.


mkustermann:
Seems kind of weird, they start with a dash, but ok.

PR #569, opened 2017-11-14
Show detailed package dependencies on the analysis tab.

isoos:


kevmoo:
@isoos – looks like you'll need to rebase

kevmoo:
@isoos – I like this, but we should do work to make it look...cleaner. Can dig in later...

isoos:
> we should do work to make it look...cleaner. Can dig in later...

Agreed. no feature < raw look < nice look.

PR #571, opened 2017-11-14
Format overall score in package listing.

isoos:
Notes:
- I've removed the label for now, TBD if we really need it.
- Colors are not "rottentomato" yet, let's see how a 4-color version goes (2 good-ish, 1 meh, 1 for the unknown analysis).
- Analysis tab logo needs to get the AnalysisExtract to display the same number (will be in a follow-up PR)
- Search does have all the scores, but frontend not yet (pending on new pana version which will provide the maintenance score as part of the analysis).
- Scoring is a bit weird, see https://github.com/dart-lang/pub-dartlang-dart/issues/570

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -129,6 +129,7 @@ class TemplateService {
     final packagesJson = [];
     for (int i = 0; i < packages.length; i++) {
       final view = packages[i];
+      final overallScore = view.overallScore;
       packagesJson.add({
         'url': '/packages/${view.name}',
         'name': view.name,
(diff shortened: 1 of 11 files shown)


app/lib/frontend/templates.dart
```diff
@@ -849,6 +852,20 @@ String _getAuthorsHtml(List<String> authors, {bool clickableName: false}) {
   }).join('<br/>');
 }
 
+String _formatScore(double value) {
+  if (value == null) return '--';
```
mkustermann: Isn't it a bug in the code if the score is not within `[0, 1]`?

isoos: I'd have the '--' as an indicator for packages that have no analysis (e.g. because they were uploaded recently). Out of range of `[0, 1]` would be a bug, yes.

PR #573, opened 2017-11-14
Notification cleanup

### LINKED ISSUE #341: Search service should not depend on analyzer

isoos:
- Fixes early return bug in analyzer's dependency notification.
- Moved notification methods to their appropriate client (analyzer and search).
- Moved notification header-related methods to notification.dart.
- More attempt to notify search service for reindexing.
- Fixes #341.


app/bin/tools/notify_service.dart
```diff
@@ -22,16 +24,18 @@ Future main(List<String> args) async {
...
+      await analyzerClient.triggerAnalysis(args[1], args[2], new Set<String>());
     } else if (service == 'search' && args.length == 2) {
-      await notificationClient.notifySearch(args[1]);
+      await searchClient.triggerReindex(args[1]);
     } else {
       _printHelp();
     }
-    await notificationClient.close();
+    await analyzerClient.close();
+    await searchClient.close();
```
mkustermann: Normally it is encouraged to use `registerScopeExitCallback(analyzerClient.close)` after `registerAnalyzerClient` instead of doing it yourself, since that will do it also in the case of exceptions and does it precisely when leaving the service scope.

But since this is a tool, it's fine.

app/lib/shared/search_client.dart
```diff
@@ -50,6 +51,19 @@ class SearchClient {
...
   }
 
+  /// Search service maintains a separate index in each of the running instances.
+  /// At the moment we cannot guarantee that each of them will receive the
+  /// notification.
+  Future triggerReindex(String package) {
+    // We have a good chance that Appengine's round-robin will send the requests
+    // to separate service instances.
+    final List<Future> requests = new List.generate(
+        10,
```
mkustermann: :-/  Please add a TODO here. It is possible to query the admin api to get the list of instances.

isoos: While the only caller is the tool that sends notifications manually, I think this is not an urgent issue. I'll add a TODO in a follow-up.

PR #576, opened 2017-11-15
Single configuration initialization for both prod and dev.

isoos:


pkg/_popularity/lib/popularity.dart
```diff
@@ -12,9 +12,6 @@ part 'popularity.g.dart';
 class PackagePopularity extends Object with _$PackagePopularitySerializerMixin {
   static const int version = 3;
 
-  static String bucketName(bool dev) =>
```
kevmoo: I was actually using this in my *other* tool to make sure we stay in sync. I'll figure it out...

PR #577, opened 2017-11-15
Analysis tab's score box use the same rendering as listing.

isoos:


kevmoo:
@isoos – I'd love to get rid of the TBDs so the scores are more clear. But this is a good start...

isoos:
>  I'd love to get rid of the TBDs so the scores are more clear.

Blocked on pana 0.6, but I think we'll get rid of it by the end of the week, early next week at latest, depending on the review and release schedule.

PR #579, opened 2017-11-15
Filter on exact phrases.

isoos:
My local test queries were fast enough (worst around 50ms, usually <10ms). We could create a slightly faster filtering with a more upfront work, but I don't think it is worth the complexity.

### DIFF
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -21,6 +21,7 @@ void registerPackageIndex(PackageIndex index) =>
 
 class SimplePackageIndex implements PackageIndex {
   final Map<String, PackageDocument> _packages = <String, PackageDocument>{};
+  final Map<String, String> _normalizedPackageText = <String, String>{};
   final TokenIndex _nameIndex = new TokenIndex(minLength: 2);
   final TokenIndex _descrIndex = new TokenIndex(minLength: 3);
   final TokenIndex _readmeIndex = new TokenIndex(minLength: 3);
@@ -64,8 +65,12 @@ class SimplePackageIndex implements PackageIndex {
     await removePackage(doc.package);
     _packages[doc.package] = doc;
     _nameIndex.add(doc.package, doc.package);
-    _descrIndex.add(doc.package, compactDescription(doc.description));
-    _readmeIndex.add(doc.package, compactReadme(doc.readme));
+    _descrIndex.add(doc.package, doc.description);
+    _readmeIndex.add(doc.package, doc.readme);
+    final String allText = [doc.package, doc.description, doc.readme]
+        .where((s) => s != null)
+        .join(' ');
+    _normalizedPackageText[doc.package] = normalizeBeforeIndexing(allText);
   }
 
   @override
@@ -82,6 +87,7 @@ class SimplePackageIndex implements PackageIndex {
     _nameIndex.remove(package);
     _descrIndex.remove(package);
     _readmeIndex.remove(package);
+    _normalizedPackageText.remove(package);
   }
 
   @override
@@ -231,7 +237,25 @@ class SimplePackageIndex implements PackageIndex {
         return Score.max([name, descr, readme]).removeLowValues(
             fraction: 0.01, minValue: 0.001);
       }).toList();
-      return Score.multiply(wordScores);
+      Score score = Score.multiply(wordScores);
+
+      // filter results based on exact phrases
+      final List<String> phrases =
+          extractExactPhrases(text).map(normalizeBeforeIndexing).toList();
+      if (phrases.isNotEmpty) {
+        final Map<String, double> matched = <String, double>{};
+        for (String package in score.getKeys()) {
+          final allText = _normalizedPackageText[package];
+          final bool matchedAllPhrases =
+              phrases.every((phrase) => allText.contains(phrase));
+          if (matchedAllPhrases) {
+            matched[package] = score[package];
+          }
+        }
+        score = new Score(matched);
+      }
+
+      return score;
     }
     return null;
   }
--- a/app/lib/search/text_utils.dart
+++ b/app/lib/search/text_utils.dart
@@ -6,6 +6,7 @@ final RegExp _separators =
     new RegExp(r'[_\.,;=\(\)\>\<\[\]\{\}\|\?\!\/\+\-\*]|\s');
 final RegExp _nonCharacterRegExp = new RegExp('[^a-z0-9]');
 final RegExp _multiWhitespaceRegExp = new RegExp('\\s+');
+final RegExp _exactTermRegExp = new RegExp(r'"([^"]+)"');
 
 String compactText(String text, {int maxLength: -1}) {
   if (text == null) return '';
@@ -33,3 +34,6 @@ Iterable<String> splitForIndexing(String text) {
   if (text == null || text.isEmpty) return new Iterable.empty();
   return text.split(_separators).where((s) => s.isNotEmpty);
 }
+
+List<String> extractExactPhrases(String text) =>
+    _exactTermRegExp.allMatches(text).map((m) => m.group(1)).toList();
--- a/app/lib/search/README.md
+++ b/app/lib/search/README.md
@@ -110,6 +110,11 @@ and accumulate a score based on the following calculation:
   - `doc.size = the size of the document`, calculated as `sqrt(number of tokens)`.
   - `score(doc) = (matched token weights) / (query weight * doc.size)`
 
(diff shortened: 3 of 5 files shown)


app/lib/search/text_utils.dart
```diff
@@ -6,6 +6,7 @@ final RegExp _separators =
     new RegExp(r'[_\.,;=\(\)\>\<\[\]\{\}\|\?\!\/\+\-\*]|\s');
 final RegExp _nonCharacterRegExp = new RegExp('[^a-z0-9]');
 final RegExp _multiWhitespaceRegExp = new RegExp('\\s+');
+final RegExp _exactTermRegExp = new RegExp('"(.*?)"');
```
mkustermann: Instead of using a non-greedy operator (which is super hard to spot in the first place), maybe we can use this:

```dart
final RegExp _exactTermRegExp = new RegExp(r'"([^"]+)"')
```

app/lib/search/text_utils.dart
```diff
@@ -33,3 +34,12 @@ Iterable<String> splitForIndexing(String text) {
   if (text == null || text.isEmpty) return new Iterable.empty();
   return text.split(_separators).where((s) => s.isNotEmpty);
 }
+
+List<String> extractExactPhrases(String text) => _exactTermRegExp
+    .allMatches(text)
```
mkustermann: For a search string like:
```
before "hello world" middle "greet from world" after
```
This will probably give 3 matches:
```
hello world
 middle 
greet from world
```

I think it would be easier to not use regexp in the first place and just change the tokenization mechanism of the text.

app/lib/search/text_utils.dart
```diff
@@ -33,3 +34,12 @@ Iterable<String> splitForIndexing(String text) {
...
+    .map((m) => m.group(1).trim())
+    .where((s) => s.isNotEmpty)
+    .toList();
+
+bool matchesPhrase(String text, String phrase) =>
+    normalizeBeforeIndexing(text).contains(normalizeBeforeIndexing(phrase));
```
mkustermann: This is an `O(3 * n)` for normalization plus another `O(n)` for `contains`. For larger README files this is really not good!

app/lib/search/index_simple.dart
```diff
@@ -231,7 +231,29 @@ class SimplePackageIndex implements PackageIndex {
...
+      final List<String> exactPhrases = extractExactPhrases(text);
+      if (exactPhrases.isNotEmpty) {
+        final Map<String, double> matched = <String, double>{};
+        for (String package in score.getKeys()) {
+          final doc = _packages[package];
+          bool matchedAllPhrases = true;
+          for (String phrase in exactPhrases) {
+            final bool matchedPhrase = matchesPhrase(doc.package, phrase) ||
+                matchesPhrase(doc.description, phrase) ||
+                matchesPhrase(doc.readme, phrase);
```
mkustermann: Given the cost of this, I would prefer if we do not do this on the `readme` file.

isoos: We limit the readme length before it even reaches the index: https://github.com/dart-lang/pub-dartlang-dart/blob/master/app/lib/search/backend.dart#L101

app/lib/search/text_utils.dart
isoos: Something like that is addressed in this test: https://github.com/dart-lang/pub-dartlang-dart/pull/579/files#diff-fdba8841ec20cef640d3f49523edcaf9R34

I think a `split` or `indexOf`-based phrase extraction will be more complex than the regexp. I'll change to the non-greedy version though, it may be faster.

app/lib/search/text_utils.dart
isoos: I'll add a pre-processing step to reduce the time of this. For this check, we could concatenate the title, description and readme (strictly speaking there is a chance for false-positives, but that is probably not an issue in practice), and do the normalization at indexing time.

isoos:
PTAL

kevmoo:
@mkustermann – thoughts?

app/lib/search/text_utils.dart
```diff
@@ -6,6 +6,7 @@ final RegExp _separators =
     new RegExp(r'[_\.,;=\(\)\>\<\[\]\{\}\|\?\!\/\+\-\*]|\s');
 final RegExp _nonCharacterRegExp = new RegExp('[^a-z0-9]');
 final RegExp _multiWhitespaceRegExp = new RegExp('\\s+');
+final RegExp _exactTermRegExp = new RegExp(r'"([^"]*)"');
```
mkustermann: The `*` should be a `+` to ensure we match at least one character

app/lib/search/index_simple.dart
```diff
@@ -231,7 +237,28 @@ class SimplePackageIndex implements PackageIndex {
         return Score.max([name, descr, readme]).removeLowValues(
             fraction: 0.01, minValue: 0.001);
       }).toList();
-      return Score.multiply(wordScores);
+      Score score = Score.multiply(wordScores);
+
+      // filter results based on exact phrases
+      final List<String> exactPhrases = extractExactPhrases(text);
```
mkustermann: Please update the documentation about this change.

app/lib/search/index_simple.dart
```diff
@@ -231,7 +237,28 @@ class SimplePackageIndex implements PackageIndex {
...
+      if (exactPhrases.isNotEmpty) {
+        final Map<String, double> matched = <String, double>{};
+        for (String package in score.getKeys()) {
+          final allText = _normalizedPackageText[package];
+          bool matchedAllPhrases = true;
+          for (String phrase in exactPhrases) {
+            final String normalizedPhrase = normalizeBeforeIndexing(phrase);
+            final bool matchedPhrase = allText.contains(normalizedPhrase);
+            matchedAllPhrases = matchedAllPhrases && matchedPhrase;
```
mkustermann: We can actually break out of the loop as soon as `matchedPhrase == false`.
e.g.

```dart
next_package:
for (String package in score.getKeys()) {
  final allText = _normalizedPackageText[package];
  for (String phrase in exactPhrases) {
    if (!allText.contains(phrase)) {
      continue next_package;
    }
    matched[package] = score[package];
  }
}

```

app/lib/search/text_utils.dart
```diff
@@ -33,3 +34,9 @@ Iterable<String> splitForIndexing(String text) {
   if (text == null || text.isEmpty) return new Iterable.empty();
   return text.split(_separators).where((s) => s.isNotEmpty);
 }
+
+List<String> extractExactPhrases(String text) => _exactTermRegExp
+    .allMatches(text)
+    .map((m) => m.group(1).trim())
+    .where((s) => s.isNotEmpty)
+    .toList();
```
mkustermann: Add a `.map(normalizeTextBeforeIndexing)` here and remove it on the usage site (which does it repeatedly atm).

mkustermann:
Well as I commented before, this is doing `O(#num-packages * avg-readme-size * #num-exact-phrases)` computation for one search query. This is just not good (whether we have a maximum size of the readme we look at or not). Search is usually implemented by using inverted indexes instead of doing linear searching.

If you want, you can of course land it, but I am not voting for it.

app/lib/search/text_utils.dart
isoos: I've put the `.map` at the usage point, it keeps the extraction and the indexing/matching of it separate.

isoos:
> Well as I commented before, this is doing O(#num-packages * avg-readme-size * #num-exact-phrases) computation for one search query.

We do have O(#num-filtered-packages * #num-exact-phrases) String.contains() calls, but that's not as bad as you think. E.g.

- an exact phrase of "dar" yields 2217 candidates on staging, and the median ngram-filtering time is roughly 3.8 ms on my notebook, while the exact phrase matching is an additional 4.1ms.
- an exact phrase of "game" yields 218 candidates, ngram-filtering is ~0.4ms, exact phrase matching is an additional 0.9ms.
- an exact phrase of "compression" yields 19 candidates, ngram-filtering is ~0.15ms, exact phrase matching is an additional 0.2ms.

> Search is usually implemented by using inverted indexes instead of doing linear searching.

We could collect word positions and do the matching against the phrases, but that would (a) make the index's memory size larger, (b) doesn't seem to be urgent, given the query times above.

kevmoo:
@mkustermann ?

isoos:
@mkustermann: I believe with #623 we can at least monitor whether any part of search gets slow, and re-do the algorithm here with the more elaborate positional indexing.

PR #580, opened 2017-11-16
Upgrade pana to 0.6.0

isoos:
Fixing immediate breaking changes and easy-to-do substitution (`AnalysisView.health`). Maintenance score will be done in a subsequent PR.


isoos:
I've also check the `analyzer` service on staging, looks ok so far. `search` will be tested with the maintenance score change.

PR #581, opened 2017-11-16
Store maintenance score in Analysis and serve it to both search and frontend.

isoos:
Maintenance score depends on the publish date, and either we apply it in `AnalyzerClient` by querying the Datastore at the time of the query, or we store the score alongside the `Analysis` instance. I think the later is much better latency-wise, and we don't lose too much if the score drops only at regular analysis intervals (~ a month). 

Note: some of the staging data does not contain this field yet, because the version I've deployed in the morning didn't store it.


isoos:
PTAL

isoos:
> PTAL
^ wrong PR, sorry

PR #583, opened 2017-11-16
Package name policy: add missing check and address review comments.

isoos:


app/lib/shared/utils.dart
```diff
@@ -174,11 +183,14 @@ void validatePackageName(String name) {
...
   final bool isLower = name == name.toLowerCase();
-  final bool matchesMixedCase =
-      knownMixedCasePackages.any((s) => s.toLowerCase() == name.toLowerCase());
+  final bool matchesMixedCase = knownMixedCasePackages.contains(name);
   if (!isLower && !matchesMixedCase) {
     throw new Exception('Package name must be lowercase.');
   }
+  if (isLower && blockedLowerCasePackages.contains(name)) {
+    throw new Exception(
+        'Name collision with mixed-case package. Please open new issue at: https://github.com/dart-lang/pub-dartlang-dart/issues/new');
```
mkustermann: So we prevent the uploader of `babylon` to make new uploads, but we do allow new uploads to `Babylon`?


Shouldn't we do it the other way around and ask the author if we can remove `Babylon` entirely?

isoos: We accept both `Babylon` and `babylon`, but you are right, we shall reject `Babylon` in this case.

isoos:
PTAL, updated tests to distinguish the two major use case: when authors did contact vs. not.

Issue #368, opened 2017-09-20
Add keywords field to pubspec.yaml


nex3:
_From @anders-sandholm on August 11, 2017 13:56_

_From @mnordine on August 11, 2017 12:59_

Search results on https://pub.dartlang.org/ could be improved with the addition of keywords for packages. It could also improve the home page by listing packages via popular keywords. For prior art, see https://crates.io/ and http://doc.crates.io/manifest.html

From the previous link:
```toml
# This is a list of up to five keywords that describe this crate. Keywords
# are searchable on crates.io, and you may choose any words that would
# help someone find this crate.v
keywords = ["...", "..."]
```

_Copied from original issue: dart-lang/sdk#30418_

_Copied from original issue: dart-lang/pub#1686_

isoos:
We already index description and the first part of readme. I'm not sure what another free-form text field could add to this, unless we want to promote these keywords somehow else. Thoughts?

Ali-Azmoud:
> We already index description and the first part of readme.  

No you don't. for example, you can't find [Flutter Xlider](https://pub.dartlang.org/packages/flutter_xlider) by keywords `xlider`, flutter xlider`, `(Flutter Xlider)` on google, while these words are inside the first part of readme. description as well

isoos:
@Ali-Azmoud: When I was say we index description and the first part of the readme, I was referring to the pub site. The pub site searches for [xlider](https://pub.dartlang.org/flutter/packages?q=xlider) and [flutter xlider](https://pub.dartlang.org/flutter/packages?q=flutter+xlider) bring up the package as the first hit. 

We also do expose lots of metadata for Google search to index the site, but we don't have direct influence on when, how and with what relevance it will happen.

mit-mit:
Closing as a duplicate of https://github.com/dart-lang/pub-dev/issues/5219 (that uses a slightly different mechanism for specifying the tags, but logically it's similar).

PR #602, opened 2017-11-17
 Custom package description extensions.

isoos:
Definitely improved angulardart -> angular result, but I wasn't able to check it on prod index.

### DIFF
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -64,8 +64,9 @@ class SimplePackageIndex implements PackageIndex {
     await removePackage(doc.package);
     _packages[doc.package] = doc;
     _nameIndex.add(doc.package, doc.package);
-    _descrIndex.add(doc.package, compactDescription(doc.description));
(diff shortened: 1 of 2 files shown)


mkustermann:
I really don't know if this is the right approach. I mean every package author could say: I'd like package X rank higher when searched for Y. So we would continue to add more terms to the description.

How about just publishing a new angular package with the desription in it? Or adding (with coordination with the pub client devs) a `keywords: ` to the `pubspec.yaml`?

isoos:
> How about just publishing a new angular package with the description in it?

+1

> Or adding (with coordination with the pub client devs) a keywords: to the pubspec.yaml?

What would you put in `keywords` which you wouldn't put in `description`? Related request: https://github.com/dart-lang/pub-dartlang-dart/issues/368

mkustermann:
> What would you put in keywords which you wouldn't put in description? Related request: #368

I would consider `description` like human readable text (i.e. sentences) whereas `keywords` as  structured data (not natural language), i.e .machine-readable.

isoos:
Closing PR, I plan another way of addressing this.

PR #609, opened 2017-11-20
Ui tweaks

kevmoo:


kevmoo:
Gah! I hadn't looked at the sizes. We should shrink those to the max
expected size, no?

isoos:
> We should shrink those to the max expected size, no?

4x of the expected is ok, depending on the dpi it might help with the smooth edges, but the 12x may be too much.

kevmoo:
Oh...uh...so we should ignore the CL I just sent....

Hrm...

PR #610, opened 2017-11-20
Resize header images to 1040px wide

kevmoo:


kevmoo:
@isoos – went with 1040px. Not worried about 4x scaling here. I want initial load to be snappy

isoos:
wrong close

mkustermann:
Please add the image scaling and compression command to readme (be it using imagemagic's `convert` or whatever else)

PR #614, opened 2017-11-21
Alternative search query: remove common and platform-specific phrases.

isoos:


isoos:
While this is a lucky, almost-coincidence that it helps https://github.com/dart-lang/pub-dartlang-dart/issues/595, the query rewrite helps packages in a generic way.

kevmoo:
Help me understand this a bit better before we land it?

isoos:
> Help me understand this a bit better before we land it?

Assume you are on Flutter, and search for `Flutter datastore`. Right now we will serve you with one result. With this patch, we'll do the same `Flutter datastore` query, but also try `datastore` as well, and combine its results with the original query's results.

The phrases we try to remove from the query are common (e.g. `dart`) and also platform-specific, in this case `Flutter`. The alternative query's weight decreases with the more phrases we are removing, e.g. `flutterdart database` will remove both `flutter` and `dart` from the query text, and it will weight much less than if only one of them could have been removed.

I guess I'll need to update the readme with this :)

isoos:
Updated readme.

isoos:
I'm having a second thought on this. Please postpone review for now.

isoos:
While it helps for the `AngularDart` query, it seems to put packages like `dartson` (with the `dartson` as query text) at disadvantage, causing it to rank further down than their original position.

I've also sampled the indexed terms and the past week's history of searches I don't think that the complexity here is worth it.

PR #617, opened 2017-11-21
Log request X-Cloud-Trace-Context headers and pass them to services

kevmoo:
Fixes https://github.com/dart-lang/pub-dartlang-dart/issues/551

### DIFF
--- a/app/lib/shared/trace_context.dart
+++ b/app/lib/shared/trace_context.dart
@@ -1,0 +1,42 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:gcloud/service_scope.dart' as ss;
+import 'package:logging/logging.dart';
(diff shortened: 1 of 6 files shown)


kevmoo:
Now we'll see results like `X-Cloud-Trace-Context : <trace id>/<span_id>` in the logs

We can search for `<trace id>` and see all requests across services for that trace ID

We can even explicitly set a trace_id when hitting a URL 

```
curl "http://www.example.com" --header "X-Cloud-Trace-Context:
  105445aa7843bc8bf206b120001000/0;o=1"
```

So we can debug issues.

Is nice!

app/lib/shared/trace_context.dart
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
isoos: If we don't expect that this will grow further, the code here could go into `shared/handlers.dart`.

kevmoo: I looked around for a good spot for it. I'd have to add an import to `shared/handlers.dart` to every entry point, too.

I kinda like keeping it self-contained.

Although centralizing error handling, etc could be interesting in the future

PR #622, opened 2017-11-22
Popularity score: use percentiles.

isoos:
I've started off with a helper class that takes the entries and creates a reverse index of the percentiles, but in the end this version ended up simple enough.


kevmoo:
Willing to take a look here. I could imagine that we want the long tail to suffer a bit – but let's see how it looks.

PR #647, opened 2017-11-27
Making sure both the landing images have the same dimensions.

isoos:


isoos:
I've added 2 extra transparent pixel rows to the top, and compressed the result image.

PR #675, opened 2017-11-30
Indexing dependencies and searching using them.

isoos:
I'm not sure in what format we want to expose this, but seems to be useful in general. Some possible alternatives (for later refinement):
- we could use different keywords for direct- dev- or transitive dependencies (e.g. `devDependency:packageName`)
- we could default to search only `direct` and `dev` dependencies, and include all if: `dependency*:package` or `*dependency:package` is specified

Besides search, we could put a link on each package page: "Who is using this?"


kevmoo:
Nice!

kevmoo:
@isoos – you made travis 😭

PR #680, opened 2017-11-30
Docker silly

kevmoo:
Figured out why we were running `2.0.0-dev` w/ `1.25` was called out explicitly

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,14 @@
 FROM google/dart-runtime-base:1.25.0-dev.9.0
 
- # After install we remove the apt-index again to keep the docker image diff small.
- RUN apt-get update && \
-     apt-get upgrade -y && \
-     apt-get install -y git unzip && \
-     rm -rf /var/lib/apt/lists/*
+# `apt-mark hold dart` ensures that Dart is not upgraded with the other packages
+#   We want to make sure SDK upgrades are explicit.
+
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,10 +1,9 @@
-FROM google/dart-runtime-base:1.25.0-dev.9.0
+FROM google/dart-runtime-base:2.0.0-dev.8.0
 
- # After install we remove the apt-index again to keep the docker image diff small.
- RUN apt-get update && \
-     apt-get upgrade -y && \
-     apt-get install -y git unzip && \
-     rm -rf /var/lib/apt/lists/*
+# After install we remove the apt-index again to keep the docker image diff small.
+RUN apt-get update && \
```
isoos: why remove upgrade?

kevmoo: Upgrade upgrades **everything* that's installed, including `Dart`.

We're set up to use a recent Dart image – which uses a recent Google Cloud image. So I think we're good!

Dockerfile
```diff
@@ -1,10 +1,11 @@
...
+    apt-mark hold dart && \
+    apt-get upgrade -y && \
```
kevmoo: @isoos – was able to keep upgrade by `holding` dart. Nice!

isoos: That is much better. We'll get the latest security patches on creation while the dart version is under control, nice!

PR #683, opened 2017-12-01
String interning for package documents.

isoos:
I don't think this is strictly required at the moment, but we may need it later. No measurable speed drawback during indexing, and at the moment only a little memory benefit on staging (hard to tell, lots of memory size fluctuations).

### DIFF
--- a/app/lib/search/text_utils.dart
+++ b/app/lib/search/text_utils.dart
@@ -37,3 +37,15 @@ Iterable<String> splitForIndexing(String text) {
 
 List<String> extractExactPhrases(String text) =>
     _exactTermRegExp.allMatches(text).map((m) => m.group(1)).toList();
+
+class StringInternPool {
+  final Map<String, String> _values = <String, String>{};
+
+  String intern(String value) => _values.putIfAbsent(value, () => value);
+
+  void checkUnboundGrowth() {
+    if (_values.length > 100000) {
+      _values.clear();
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


app/lib/search/text_utils.dart
```diff
@@ -37,3 +37,15 @@ Iterable<String> splitForIndexing(String text) {
 
 List<String> extractExactPhrases(String text) =>
     _exactTermRegExp.allMatches(text).map((m) => m.group(1)).toList();
+
+class StringInternPool {
```
mkustermann: There is already a class for this. Also as mentioned before, please use "canonicalize" instead of "interning" (which nobody knows what it means).

app/lib/shared/deps_graph.dart:15:class StringCanonicalization {

isoos: Ah, good point about that class, it shall be a single one of it.

However, I disagree with the name, as both Java and C# are using it for a long time:
https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#intern()
https://msdn.microsoft.com/en-us/library/system.string.intern.aspx

Canonicalization changes the date (like in a path `/a/../b/` -> `/b/`), while interning does not change it. There are more examples here: https://en.wikipedia.org/wiki/Canonicalization

isoos:
Updated this with a fresh rebase. @mkustermann: I'm open to consider another term beside interning, but canonical doesn't feel the right one.

mkustermann:
Even though I think "canonicalize" would be much better than "intern", I see there is little chance of convincing you, so I'll just merge it.

The clearing of the values is IMHO not necessary, since we only use it to canonicalize certain strings, e.g. package names, platforms which we have in memory anyway. It seems weired that you added this clearing method.

PR #685, opened 2017-12-01
Better styling for <code> blocks.

isoos:
I've found myself too annoyed that the inline code blocks are not really separated from the normal flow of text.


kevmoo:
Love it!

PR #686, opened 2017-12-01
[WIP] Create and use a pre-built Docker image

kevmoo:
Improves the container build time by ~1.5 minutes
Removes the `setting up flutter` during analyzer instance startup

Fixes https://github.com/dart-lang/pub-dartlang-dart/issues/593

### DIFF
--- a/app/bin/service/analyzer.dart
+++ b/app/bin/service/analyzer.dart
@@ -21,15 +21,14 @@ import 'package:pub_dartlang_org/analyzer/handlers.dart';
 import 'package:pub_dartlang_org/analyzer/pana_runner.dart';
 import 'package:pub_dartlang_org/analyzer/task_sources.dart';
 
-final Logger logger = new Logger('pub.analyzer');
+final Logger _logger = new Logger('pub.analyzer');
 
 Future main() async {
   useLoggingPackageAdaptor();
 
   withAppEngineServices(() async {
-    _initFlutterSdk().then((_) async {
-      startIsolates(logger, _runSchedulerWrapper);
-    });
+    // TODO(kevmoo): check that the flutter versions align
+    startIsolates(_logger, _runSchedulerWrapper);
     _registerServices();
     await runHandler(analyzerServiceHandler);
   });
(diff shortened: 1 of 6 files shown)


isoos:
This doesn't remove setup flutter, it does setup it in each of the services. (Not a bad thing, but PR description would suggest the removal).

Is there a change in the usual deploy process? Could you please update the README with it?

kevmoo:
Not excited to run w/ this now.

This will cause issues w/ the dev version, for instance –– since that project won't have access to these images.

We should revisit later...

mkustermann:
Reducing the deploy time is an optimization but there is imho a more important use-case where this would make a lot of sense:

Currently we cannot make new deployments if the server is down, because building requires `pub get`. So having a base image with all the transitive dependencies in it would allow us to build & deploy any time, even if prod version is down/broken, since we can use `pub get --offline` with the packages in the pub-cache (from a base image).
-> That would be very helpful!

kevmoo:
@mkustermann – agreed!

We'd have to do some dancing around to pre-populate the pkg cache, but that should be doable

app/bin/service/analyzer.dart
```diff
@@ -21,15 +21,14 @@ import 'package:pub_dartlang_org/analyzer/handlers.dart';
...
+final Logger _logger = new Logger('pub.analyzer');
 
 Future main() async {
   useLoggingPackageAdaptor();
 
   withAppEngineServices(() async {
-    _initFlutterSdk().then((_) async {
-      startIsolates(logger, _runSchedulerWrapper);
-    });
+    // TODO(kevmoo): check that the flutter versions align
```
mkustermann: Which flutter version needs to be in sync? Should this comment be addressed before merging the PR?

I thought we had the git hash in some shell script or something.

kevmoo: We could check that the environment variable is set correctly – and aligns w/ what's in the Dart source...

docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM google/dart-runtime-base:2.0.0-dev.8.0
+
+# `apt-mark hold dart` ensures that Dart is not upgraded with the other packages
```
mkustermann: If you keep this file one level up we can get the transitive packages in as well. Just use a different name, like `Dockerfile.pub_site_base`. `docker build -f  Dockerfile.pub_site_base` will build it.

Also add the `pubspec.yaml` and the `pub get` in here.

kevmoo:
I'm going to close this pull request, but leave the branch. I think it'd be great if someone ran w/ this.

PR #688, opened 2017-12-01
Link to dependent search on the package page

isoos:
screen shot 2017-12-01 at 7 08 55 pm

### DIFF
--- a/app/views/v2/pkg/show.mustache
+++ b/app/views/v2/pkg/show.mustache
@@ -128,6 +128,9 @@ import 'package:{{package}}/{{library}}';
       <h3 class="title">License</h3>
       <p>{{& package.license_html}}</p>
     {{/package.license_html}}
+
+    <h3 class="title">More</h3>
+    <p><a href="{{{search_deps_link}}}">Packages that depend on {{package.name}}</a></p>
   </aside>
 </div>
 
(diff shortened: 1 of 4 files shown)


app/views/v2/pkg/show.mustache
```diff
@@ -128,6 +128,9 @@ import 'package:{{package}}/{{library}}';
       <h3 class="title">License</h3>
       <p>{{& package.license_html}}</p>
     {{/package.license_html}}
+
+    <h3 class="title">More</h3>
+    search <a href="{{search_deps_link}}">dependents</a>
```
kevmoo: Make the link content `Packages that depend on <pkgname>` – no `search`

app/views/v2/pkg/show.mustache
```diff
@@ -128,6 +128,9 @@ import 'package:{{package}}/{{library}}';
       <h3 class="title">License</h3>
       <p>{{& package.license_html}}</p>
     {{/package.license_html}}
+
+    <h3 class="title">More</h3>
+    Packages that depend on <a href="{{{search_deps_link}}}">{{package.name}}</a>
```
kevmoo: I meant make the whole line a link. Or just `Packages`, maybe? Having just the pkg name as the link content is a bit weird. We're not linking to `pkg/foo`

isoos: I'm not really fan of overtly long links, but let's give it a try...

Issue #694, opened 2017-12-01
Add link to all packages by an author


DanTup:
https://pub.dartlang.org/experimental/packages/test/versions/0.12.2

This is more of a request for something new than just feedback, but it surprised me when I clicked "Dart Team" on that page that it was an email link. I think it'd be useful to have a link to a page for that author that lists all of their packages.

It's also slightly strange (ironic?) that the email addresses further down the page (which you might expect to be mailto links) are not clickable (it also seems a little strange to list email addresses with no names).

isoos:
This is now done and deployed.

DanTup:
LGTM! 👍

PR #702, opened 2017-12-04
New template for author and uploader rendering.

isoos:
The idea here is to:
- have the same template and UX for authors (name and/or email) and uploaders (e-mail only)
- now it is only e-mail, but in the near future we could add `packages` there, which would link to the search that lists all of the packages that the email-address is associated to

screen shot 2017-12-04 at 3 41 49 pm

Related to #694

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -445,6 +445,8 @@ class TemplateService {
           selectedVersion.pubspec.getAllAuthors(),
           clickableName: true,
         ),
+        'authors_v2_html':
+            _getAuthorsHtmlV2(selectedVersion.pubspec.getAllAuthors()),
         'homepage': selectedVersion.homepage,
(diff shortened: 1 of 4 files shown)


kevmoo:
Not a fan of this design. Having (xyz) after the name is noisey. Adding (packages) will be even worse. It's especially weird since the Uploader content is *always* an email. Not just linking the email feels very non-webby.

isoos:
Any suggestion how to present if/when we'll have two clickable thing for it: e-mail link and search for e-mail in packages?

isoos:
I was also thinking about using icons for e-mail (letter or `@`) and packages (e.g. the dart icon), but both of these are a bit on the obscure side (it is not visually apparent that they are clickable).

isoos:
Closing this for now. I have an idea which I'll flush out in combination with indexing the authors.

PR #704, opened 2017-12-04
Improve GitHub issues links

kevmoo:
Also moved the issue report link to the bottom – prepare for release


isoos:
I think there is no need for the static/v2/js/script.dart.js.info.json file

kevmoo:
> I think there is no need for the static/v2/js/script.dart.js.info.json file

Strictly, no. But it does let us easily track how the size of the output evolves. And it has zero impact on end users.

PR #716, opened 2017-12-05
Remove mock scores

kevmoo:
Better zeros across the board that outdated noise


mkustermann:
Agree, we shouldn't use them in production ever, and we should always have ones in the bucket.

PR #733, opened 2017-12-06
Index + search e-mails (author, uploader).

isoos:
I am undecided whether this should be the final format. A few things to consider:
- search query text could be `author:user@domain.com` instead of `email:user@domain.com`
- authors and uploaders could be indexed and searched separately
- author names could be indexed and search for (more complex indexing and search lookup)


isoos:
@mkustermann @kevmoo: any take on this?

isoos:
All things considered, I think the `email:`-based search should be enough, and covers the most important concerns. Let's see it in practice :)

PR #734, opened 2017-12-06
Temp cache namespace to differentiate between running versions.

isoos:


kevmoo:
Uh...but uploads to the old site won't invalidate these new keys, right?

isoos:
> Uh...but uploads to the old site won't invalidate these new keys, right?

Ah, that's true. Shall I revert the memcache-related changes instead?

PR #743, opened 2017-12-06
V2 release: move views and update template handling.

isoos:
Moved views and test golden files from `v2/` dirs to normal dirs and removed some methods that are now deprecated. Renamed methods ending with V2 to normal name.

A few more follow-ups remain, but this should unblock other fixes.


isoos:
I've also started to deploy this to staging (version will be 20171206t232028).

PR #745, opened 2017-12-06
copy edit: /help

kwalrath:
Some notes:

* I'd like to use `<dl>`s for some of the lists, but didn't because I wasn't sure how it'd be formatted.
* We could be more specific in many places. For example:
  * What is searched? The package source code, the pubspec, what you can actually see on pub.dartlang.org, ...?
  * Do we support combining multiple search expressions?
* We should show some examples of searches.


kevmoo:
@kwalrath – we just made your life hard. @isoos just moved everything to new directories.

Should just be a copy-paste when you rebase

kwalrath:
Contributes to #745.

isoos:
> Do we support combining multiple search expressions?

We do.

app/views/help.mustache
```diff
@@ -5,35 +5,47 @@
...
+    from scores for the package's <a href="#popularity">popularity</a>,
+    <a href="#health">health</a>, and <a href="#maintenance">maintenance</a>.
   </p>
 
   <h3 id="popularity">Popularity</h3>
   <p>
-    Popularity score is good proxy for a package's use, derived from download statistics.
-    It is based on actual download counts, but compensated for automated tools (like
-    continuous builds fetching the package on each change request).
+    The popularity score—representing how often a package is used—is derived
```
isoos: `used—is` - this looks strange

isoos:
@kwalrath: thanks, this is really appreciated!

kevmoo:
@kwalrath - please squash and merge!

app/views/help.mustache
kwalrath: I'll check the appearance once it's up... it might not be so odd once it's displayed using better fonts.

PR #755, opened 2017-12-07
Error header

kevmoo:
I think the first commit is solid.
The 2nd is open for discussion. I think it's nice to be able to force a 500 – thoughts?


kevmoo:
thoughts, @isoos ?

isoos:
I'm a bit undecided on the forced exception throwing. For how long would you keep that in the source code?

kevmoo:
Well, let me land the response header thing. The error bit I thought would be nice for testing...

kevmoo:
@isoos – down to just the response header now...PTAL

PR #756, opened 2017-12-07
Fix page titles

kevmoo:
Flips variables from https://github.com/dart-lang/pub-dartlang-dart/commit/fbba3ae15183edb9d6c5cda0245e8e44db326da0


isoos:
Ah, the rush.. :(

PR #765, opened 2017-12-07
Show package dependencies in the sidebar.

### LINKED ISSUE #721: Show package dependencies on package details
This is a request to show all packages that a package depends on under package details. I.e., something like this:

**Uploader**
zarah@google.com
goderbauer@google.com

**License**
unknown (LICENSE)

**Dependencies**
[flutter](https://pub.dartlang.org/experimental/packages/flutter), [url_launcher](https://pub.dartlang.org/experimental/packages/url_launcher) 

**More**

isoos:
This fixes #721. There is no trim of the list yet (e.g. if the list is larger than 20 items), because ideally we'd like to select the most popular ones, and that would be an expensive operation at the time of the package show page. We could implement showing the first 20 alphabetically but then selecting them at random would be more fair. Let's try it without trim, and fix it once we have an outlier that has too many dependencies.

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -375,6 +375,7 @@ class TemplateService {
             _renderLicenses(selectedVersion.homepage, analysis?.licenses),
         'score_box_html': _renderScoreBox(extract?.overallScore,
             isNewPackage: package.isNewPackage()),
+        'dependencies_html': _renderDependencyList(analysis),
         'analysis_html': renderAnalysisTab(extract, analysis),
       },
       'version_table_rows': versionTableRows,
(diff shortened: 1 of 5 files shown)


kevmoo:
I'm not sure I'm a fan of this change. Let's hold off and see what @mit-mit says in the bug.

mit-mit:
@isoos any easy way to query all packages to determine how many have "lots of dependencies" (say, more than 10)?

mkustermann:
We have an in-memory dependency graph which can answer questions about: direct dependencies and  transitive dependencies.

It is super simple doing the reverse, i.e. answering questions about which packages directly depend on a given package (or do so transitively).

It is a conservative estimate, since it marginalizes over all versions of a package to determine the deps graph.

Not sure if this is what @mit-mit is looking for though.

isoos:
> we'd like to select the most popular ones

On a second thought, selecting the most popular ones isn't that expensive (popularity score is cheap, overall score is not).

> any easy way to query all packages to determine how many have "lots of dependencies" (say, more than 10)?

5 packages have 20+: `angular, angular2, build_runner, repub, test`
45 packages have 10+: `analyzer, angel_cli, angel_common, angel_framework, angular, angular2, angular_components, angular_test, aqueduct, atlassian_connect_host, atlassian_connect_shelf, build_runner, build_test, dart_dev, dartabase_migration, dartdoc, den, dev_compiler, flare, fun_components, generator_plus, gex_webapp_kit, graphql_fetch_generator, jaguar, jaguar_cli, jefe, l10n, mdl, mojito, observe, omnitest, over_react, pana, peg, pen, polyce, polymer, polymerize, polymorphic_bot, repub, shelf_serve, sitegen, soniq, test, vue2`

mit-mit:
That's a fairly small set of packages with many dependencies, so I'd say go ahead and just show the full list

isoos:
Can we decide on this? @mit-mit @kevmoo

PR #770, opened 2017-12-07
Add a special package name that causes a 500 response – useful for testing

kevmoo:
CC @nex3 – would this be useful? If we wire up the pub client for such logic, It'd be good to make sure it works consistently

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -27,6 +27,12 @@ import 'models.dart';
 import 'search_service.dart';
 import 'templates.dart';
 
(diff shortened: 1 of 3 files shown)


isoos:
Is the big change in `app/test/frontend/handlers_test.dart` indenting? Hard to tell from the diff.

kevmoo:
> Is the big change in app/test/frontend/handlers_test.dart indenting? Hard to tell from the diff.

Sorry to blend commits. Look at the 2nd commit. Easier to see.

nex3:
I don't know that this would provide us much value... we test against local servers, not against the live site, and we can always set those up to return whatever status codes we want.

kevmoo:
ack!

Issue #739, opened 2017-12-06
Copy edit needed


kwalrath:
URL: https://pub.dartlang.org/help

This page and others need a bit of a copy edit. I'd be happy to do it.

kevmoo:
@kwalrath – pull requests welcome! https://github.com/dart-lang/pub-dartlang-dart/blob/master/app/views/v2/help.mustache

kwalrath:
Working on it now...

isoos:
The change was merged and deployed.

PR #772, opened 2017-12-08
Copy edit of site footer

kwalrath:
Used sentence case, shortened links to Terms & Privacy to match the
links in the footer of www.dartlang.org.

Contributes to #739.


isoos:
@kwalrath: you'll need to update the template tests. In `templates_test.dart` change the `final _regenerateGoldens = false;` to true, re-run it, then change it back.

PR #773, opened 2017-12-08
Host-based frontend memcache keys with invalidations on pub.dartlang.org

### LINKED ISSUE #768: Create some model to selectively skip memcache
So we can validate a deployment...

isoos:
Closes #768

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -154,7 +154,8 @@ Future<shelf.Response> _indexHandler(
     return redirectResponse(
         request.requestedUri.replace(path: newPath).toString());
   }
-  String pageContent = await backend.uiPackageCache?.getUIIndexPage(platform);
+  String pageContent = await backend.uiPackageCache
+      ?.getUIIndexPage(request.requestedUri.host, platform);
   if (pageContent == null) {
     Future<String> searchAndRenderMiniList(SearchOrder order) async {
       final count = 15;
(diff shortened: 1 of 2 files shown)


mkustermann:
Couldn't we just prefix the memcache entries with the deployment version (which is always the same for a given instance) instead of deriving the memcache prefix from the incoming requests?

isoos:
> Couldn't we just prefix the memcache entries with the deployment version (which is always the same for a given instance) instead of deriving the memcache prefix from the incoming requests?

But then an upload that happens on a non-migrated or a partially migrated instance wouldn't clear the cache on all of the prod instances.

kevmoo:
> Couldn't we just prefix the memcache entries with the deployment version (which is always the same for a given instance) instead of deriving the memcache prefix from the incoming requests?

@mkustermann – I had a similar idea, but we'd have to get pretty crazy w/ our cache invalidation logic.

This is a simple first step so we can sanity check a deploy before making it default.

mkustermann:
Well, I'm not a big fan of this. Our invalidation logic will still be unable to invalidate all keys in memcache.

We could just override the memcache service in the service scope to be a NOP cache if the URL has a specific get parameter or if the requested url is not `pub.dartlang.org`. Then we'll always get up-to-date versions of the page.

isoos:
> We could just override the memcache service in the service scope to be a NOP cache if the URL has a specific get parameter or if the requested url is not pub.dartlang.org.

I've considered not caching when the URL is not `pub.dartlang.org`, but then if there is an issue with the caching logic, we wouldn't notice it during validation of a new release.

kevmoo:
Let's punt on this for now, given the alternative...

PR #774, opened 2017-12-08
Search order dropdown.

isoos:
- fixes default handling in `search` (requires deploy)
- the non-JS visible label is now in-sync with the search order
- the JS will replace the label with a dropdown, and on change it submits the form with a new order
- Not a stellar design, but functional:

screen shot 2017-12-08 at 6 47 21 pm


isoos:
> I'm confused by the text change between pkg_index_page and search_page

`search_page` test has `query: 'foobar', order: SearchOrder.top`, the other has no set order (hence the `default` text). The text content is coming from `template_consts.dart`

app/test/frontend/golden/pkg_index_page.html
```diff
@@ -78,9 +78,10 @@ <h1 class="_visuallyhidden">Dart pub</h1>
   
 <div class="listing-sort-header">
   <div class="tooltip-base">
-    <span class="tooltip-dotted">Sorted by:</span> overall score
+    <span class="tooltip-dotted">Sorted by:</span>
+    <span id="sort-control" data-sort="default">default</span>
     <div class="tooltip-content tooltip-bottom-left">
-      Packages are sorted by the overall score. More information on <a href="/help#ranking">ranking</a>.
+      Packages are sorted by the combination of text match, overall score and platform context. More information on <a href="/help#ranking">ranking</a>.
```
kevmoo: The pkg page shouldn't be sorted by text match, right?

app/test/frontend/golden/search_page.html
```diff
@@ -78,9 +78,10 @@ <h1 class="_visuallyhidden">Dart pub</h1>
   
 <div class="listing-sort-header">
   <div class="tooltip-base">
-    <span class="tooltip-dotted">Sorted by:</span> relevance
+    <span class="tooltip-dotted">Sorted by:</span>
+    <span id="sort-control" data-sort="top">overall score</span>
     <div class="tooltip-content tooltip-bottom-left">
-      Packages are sorted by the combination of text match and overall score. More information on <a href="/help#ranking">ranking</a>.
+      Packages are sorted by the overall score. More information on <a href="/help#ranking">ranking</a>.
```
kevmoo: The search page *should* be ordered by text match, right?

static/v2/js/script.dart
```diff
@@ -151,6 +152,52 @@ void _setEventForSearchInput() {
   });
 }
 
+_setEventForSortControl() {
```
kevmoo: void?

app/test/frontend/golden/search_page.html
isoos: normally yes, but in the template test it is rendered with another ordering

app/test/frontend/golden/pkg_index_page.html
isoos: Text match will be applied only if there is text and ordering is either `default` or `text`. I've kept it in the tooltip because it was easier to have one tooltip for both cases that handling the case where we are on `default` with text and without text.

app/lib/frontend/template_consts.dart
```diff
@@ -86,3 +86,41 @@ final Map<String, String> _landingBlurbs = const {
...
+
+  const SortDict({this.label, String tooltip})
+      : this.tooltip =
+            '$tooltip More information on <a href="/help#ranking">ranking</a>.';
+}
+
+// Synchronize with `script.dart`'s dropdown.
+final _sortDicts = const <String, SortDict>{
+  'default': const SortDict(
+      label: 'default',
```
kevmoo: Instead of `default` -> `Text match and score`

isoos:
> When I land on this page w/ no text in the query, I'd assume the selected sort option would be 'overall score'. Only when there is text, would I expect to see 'default' (or my suggestion "Text match and score").

Without search text, default = overall score + platform specificity. default never equals overall score, that is a separate sort order

kevmoo:
Sure. Default is not descriptive. Can we make it more so?

isoos:
PTAL. I've ended up with separating the two kind of default into "listing relevance" and "search relevance".

PR #780, opened 2017-12-11
Clickable hover menus (touch support where :hover is not present).

isoos:


kevmoo:
Looks like we need a rebase...

kevmoo:
@isoos – does this address a bug? Something about mobile?

isoos:
> does this address a bug? Something about mobile?

Yeah, I wanted to update the issue after this has been deployed

PR #784, opened 2017-12-12
Phrase highlight for search results.

### LINKED ISSUE #465: To discuss: highlight "exact matches" in search
See https://crates.io/search?q=winapi

If the search string exactly matches a package, highlight it on top

Wondering if this would be useful?

isoos:
Closes #465.

screen shot 2017-12-12 at 6 02 01 pm

### DIFF
--- a/app/lib/frontend/highlight.dart
+++ b/app/lib/frontend/highlight.dart
@@ -1,0 +1,48 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:math' as math;
+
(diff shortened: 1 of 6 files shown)


kevmoo:
Now that I see this, I'm not sure I like it.

@mit-mit ?

mit-mit:
I think that looks quite busy indeed. Did search engines used to do that, but removed it as it was too distracting. I usually just do a find in the page myself...

isoos:
The style can be more subtle and faded, but I'm happy to close this if that is the consensus.

isoos:
screen shot 2017-12-14 at 9 39 21 am

Alternative style, less busy. @mit-mit @kevmoo ?

mit-mit:
I'm still not convinced it's worth it, still looks busy and (subjectively) a bit "old school". If a customer wants highlighting, they can just use the browser 'search in page' feature.

PR #789, opened 2017-12-13
Update dependencies

kevmoo:


kevmoo:
pkg/build now uses code_builder ... which uses built

PR #791, opened 2017-12-14
'new' label template and style on overall score

isoos:
Template for the new label:

screen shot 2017-12-13 at 10 14 24 pm

The setting of the flag needs a bit thought, the I did implement it in one way, but I'd rather do it differently.


kevmoo:
@sethladd – thoughts?

@isoos – guessing we'll add the `isNew` argument later...

isoos:
> guessing we'll add the isNew argument later

yeah, that's the plan

sethladd:
love it, let's try it!

PR #792, opened 2017-12-14
Remove only-meta parameter from analyzer.

isoos:
We never use it, and makes the next planned step simpler.


kevmoo:
> We never use it, and makes the next planned step simpler.

@isoos – what's the next planned step?

isoos:
> what's the next planned step?

I'm planning to refactor how the `analyzer` service returns the `AnalysisData` object, most importantly: how to attach additional data to it (like package's created and updated timestamp). One way to do it is to store the `AnalysisData` instead of storing pana's `Summary` plus a few fields.

PR #794, opened 2017-12-14
Memcache only on prod domain.

### LINKED ISSUE #768: Create some model to selectively skip memcache

isoos:
An alternative to #773, closes #768.

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -154,7 +154,9 @@ Future<shelf.Response> _indexHandler(
     return redirectResponse(
         request.requestedUri.replace(path: newPath).toString());
   }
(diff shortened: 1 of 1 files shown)


app/lib/frontend/handlers.dart
```diff
@@ -154,7 +154,9 @@ Future<shelf.Response> _indexHandler(
     return redirectResponse(
         request.requestedUri.replace(path: newPath).toString());
   }
-  String pageContent = await backend.uiPackageCache?.getUIIndexPage(platform);
+  final isProd = _isProd(request);
```
kevmoo: Idea: pondering if some sort of `withCache(..., Future<String> action(), {bool skipCache})`  helper would be useful here to keep us honest about get/set alignment, etc.

Issue #785, opened 2017-12-12
Explain the interaction between PannerRunner.hasCompletedRecently and AnalysisBackend.isValidTarget


kevmoo:
I just updated https://pub.dartlang.org/packages/bot_io which is used by https://pub.dartlang.org/packages/hop

I'd assume that updating bot_io would start another analysis of hop – but the log just lists a bunch of 'skips'.

Maybe `isValidTarget` could log more?
Maybe it could return something "richer" than `bool` – something `PannerRunner.hasCompletedRecently` could use to log it's decision.

etc etc

isoos:
I haven't found much in the logs, but opened #796 for more logging.

kevmoo:
Closed by https://github.com/dart-lang/pub-dartlang-dart/commit/156e66c4f1bcdcb2c604d9455de6fe2165836b2d

PR #796, opened 2017-12-14
More logging for analyzer task decisions.

isoos:
#785

### DIFF
--- a/app/lib/analyzer/backend.dart
+++ b/app/lib/analyzer/backend.dart
@@ -154,12 +154,15 @@ class AnalysisBackend {
     });
   }
 
-  /// Whether [packageName] with [packageVersion] exists, and
+  /// Returns the task target status based on:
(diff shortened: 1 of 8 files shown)


app/lib/analyzer/backend.dart
```diff
@@ -216,15 +225,16 @@ class AnalysisBackend {
...
     if (age > reanalyzeThreshold) {
-      return true;
+      return new TaskTargetStatus.ok();
     }
 
     // Is it a newer analysis than the trigger timestamp?
     if (versionAnalysis.analysisTimestamp.isAfter(updated)) {
-      return false;
+      return new TaskTargetStatus.skip(
+          'Previous analysis completed after task\'s trigger timestamp.');
```
kevmoo: nite: I'd rather use `"` here instead of escaping `'` – but no biggy

PR #797, opened 2017-12-14
Misc cleanup for version-table

isoos:
Cleanup in handler, templates, and also sharing the template between the index page and the versions tab.


app/test/frontend/golden/pkg_versions_page.html
```diff
@@ -0,0 +1,114 @@
+
```
kevmoo: Where does this come from? A new URL?

isoos: A new test, we didn't have a golden template for /packages/[pkg]/versions

PR #804, opened 2017-12-15
Clear memcache in uploader tool.

isoos:


kevmoo:
Ah! realized this was a bit silly. Thanks for doing it, though!

Issue #708, opened 2017-12-04
Add structured data to site


kevmoo:
See https://search.google.com/structured-data/testing-tool#url=github.com%2Fdart-lang%2Ftest as an example

For packages, we could populate: http://schema.org/SoftwareSourceCode
A bunch of easy things here

For the main site:
https://search.google.com/structured-data/testing-tool#url=npms.io

Wiring up the search action - http://schema.org/SearchAction – would be goodness to give folks access to our search engine

kevmoo:
@isoos per SearchAction – missing something https://search.google.com/structured-data/testing-tool/u/0/#url=https%3A%2F%2Fpub.dartlang.org%2F

isoos:
While the Google Search page doesn't show the sitemap's searchbar yet, Chrome does use the autocompletion (type pub.dartlang.org and hit tab, it will switch to search mode and will use the `/packages?q=` url.

PR #806, opened 2017-12-15
Structured data: schema.org/SearchAction

isoos:
Fist step in #708.

### DIFF
--- a/app/views/layout.mustache
+++ b/app/views/layout.mustache
@@ -120,5 +120,8 @@
   <a class="link github_issue" href="https://github.com/dart-lang/pub-dartlang-dart/issues/new">Report an issue</a>
 </footer>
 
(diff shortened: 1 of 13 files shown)


app/views/layout.mustache
```diff
@@ -120,5 +120,17 @@
   <a class="link github_issue" href="https://github.com/dart-lang/pub-dartlang-dart/issues/new">Report an issue</a>
 </footer>
 
+<script type="application/ld+json">
```
kevmoo: squash this down a bit? Remove indent at least – since it's on every page.

isoos: updated: generating JSON in templates.dart, compact, yet keeping it readable in the source

kevmoo: Nice!

PR #811, opened 2017-12-16
Use new part syntax for generated files

kevmoo:
Since we're using Dart 2.0.0-dev...

### DIFF
--- a/app/lib/shared/search_service.g.dart
+++ b/app/lib/shared/search_service.g.dart
@@ -6,7 +6,7 @@
 
 // ignore_for_file: prefer_final_locals
 
-part of pub_dartlang_org.shared.search_service;
+part of 'search_service.dart';
 
 // **************************************************************************
 // Generator: JsonSerializableGenerator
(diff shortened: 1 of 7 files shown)


app/lib/shared/search_service.g.dart
```diff
@@ -33,6 +33,7 @@ PackageDocument _$PackageDocumentFromJson(Map<String, dynamic> json) =>
         dependencies: json['dependencies'] == null
             ? null
             : new Map<String, String>.from(json['dependencies'] as Map),
+        emails: (json['emails'] as List)?.map((e) => e as String)?.toList(),
```
kevmoo: Oops! Looks like the generator didn't get run! @isoos

kevmoo: https://github.com/dart-lang/pub-dartlang-dart/issues/812

isoos: oops :(

PR #824, opened 2017-12-18
Schema.org metadata for packages

isoos:
#708

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -375,6 +375,8 @@ class TemplateService {
             isNewPackage: package.isNewPackage()),
         'dependencies_html': _renderDependencyList(analysis),
         'analysis_html': renderAnalysisTab(extract, analysis),
+        'schema_org_pkgmeta_json':
+            JSON.encode(_schemaOrgPkgMeta(package, selectedVersion, analysis)),
       },
       'version_table_rows': versionTableRows,
       'show_versions_link': totalNumberOfVersions > versions.length,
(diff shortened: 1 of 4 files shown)


kevmoo:
Was thinking http://schema.org/SoftwareSourceCode was a better fit.

Thoughts?

isoos:
> Was thinking http://schema.org/SoftwareSourceCode was a better fit.

Changed. We are using only the basic attributes anyway.

app/lib/frontend/templates.dart
```diff
@@ -888,3 +890,11 @@ const _schemaOrgSearchAction = const {
     'query-input': 'required',
   },
 };
+
+Map _schemaOrgPkgMeta(PackageVersion pv) => {
```
kevmoo: Happy to start small, but would you add TODOs for

http://schema.org/programmingLanguage - could start w/ text, but have fun with http://schema.org/ComputerLanguage
http://schema.org/codeRepository - if the homepage is github.com
http://schema.org/dateModified - last published
http://schema.org/dateCreated - first published
http://schema.org/license
http://schema.org/version

isoos: Updated these except for the codeRepository (added a TODO for it). I have a few plans around the github / gitlab link handling, it will be done eventually...

app/lib/frontend/templates.dart
```diff
@@ -874,3 +876,25 @@ const _schemaOrgSearchAction = const {
...
+Map _schemaOrgPkgMeta(Package p, PackageVersion pv, AnalysisView analysis) {
+  final Map map = {
+    '@context': 'http://schema.org',
+    '@type': 'SoftwareSourceCode',
+    'name': pv.package,
+    'version': pv.version,
+    'description': '${pv.package} - ${pv.pubspec.description}',
+    'url': 'https://pub.dartlang.org/packages/${pv.package}',
+    'dateCreated': p.created.toIso8601String(),
+    'dateUpdated': pv.created.toIso8601String(),
```
kevmoo: Oops! Should be 'dateModified', @isoos !

PR #830, opened 2017-12-19
More packages link on the landing page.

isoos:
I couldn't decided on where to put these links, and until a better one emerges, these should be fine.

screen shot 2017-12-19 at 5 03 37 pm

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -515,9 +515,12 @@ class TemplateService {
     String platform,
   ) {
     final platformDict = getPlatformDict(platform);
+    final packagesUrl = platform == null ? '/packages' : '/$platform/packages';
     final values = {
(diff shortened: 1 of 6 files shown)


kevmoo:
I'm not a huge fan of this. I find it a tiny bit confusing. I think having the drop-down on the results page is sufficient.

@mit-mit ?

isoos:
> I find it a tiny bit confusing. 

Agreed.

> I think having the drop-down on the results page is sufficient.

There is some demand for something like this: https://gitter.im/dart-lang/TALK-general?at=5a38d34b0163b02810545a86

isoos:
any decision yet on how to proceed with this?

kevmoo:
@mit-mit ?

mit-mit:
I don't think we need to complicate the front page given this is available on the general sort page.

PR #832, opened 2017-12-19
Static file handling refactoring

isoos:
Main reason: I want to increase the amount of time the browser can cache the static files. To do that, each new release should include a release-specific version in the `/static` path e.g. `/static/1234abcd/v2/style.css`, otherwise at the next load there could be a mismatch between the HTML file and CSS (or JS).

This change allows us to change the `/static/` URL prefix to whatever we want it to, e.g. `/abc` will work just as good as `/static/1234abcd`. I'll have a simple follow-up on this to hash the static files and generate the `1234abcd` part based on that overall hash.


app/lib/frontend/handlers.dart
```diff
@@ -221,36 +219,29 @@ Future<shelf.Response> packagesHandler(shelf.Request request) async {
...
+    if (ifModifiedSince != null &&
+        !staticFile.lastModified.isAfter(ifModifiedSince)) {
+      return new shelf.Response.notModified();
     }
+    return new shelf.Response.ok(
+      staticFile.bytes,
+      headers: {
+        HttpHeaders.CONTENT_TYPE: staticFile.contentType,
+        HttpHeaders.CONTENT_LENGTH: staticFile.bytes.length.toString(),
+        HttpHeaders.LAST_MODIFIED: formatHttpDate(staticFile.lastModified),
```
kevmoo: If we're going through all of this work, we should switch to etags based on file hashes...

make it a todo for now, I guess

isoos: `ETag` over the currently in-use `If-Modified-Since` does improve a bit, but not too much. Added TODOs though, because it will be easy with the hashed content.

kevmoo:
@isoos – want to squash this or land as separate commits?

isoos:
> but they are the same now, right?

Yeah, right now they are the same.

> want to squash this or land as separate commits?

squash

PR #840, opened 2017-12-20
ETag support for static files.

isoos:


kevmoo:
Shouldn't we drop the time header then?

Reason: every deploy the time will change, but if the content is the same the etag will not

app/lib/frontend/handlers.dart
```diff
@@ -233,14 +233,17 @@ Future<shelf.Response> staticsHandler(shelf.Request request) async {
...
+    }
     return new shelf.Response.ok(
       staticFile.bytes,
       headers: {
         HttpHeaders.CONTENT_TYPE: staticFile.contentType,
         HttpHeaders.CONTENT_LENGTH: staticFile.bytes.length.toString(),
         HttpHeaders.LAST_MODIFIED: formatHttpDate(staticFile.lastModified),
-        // TODO: emit ETag header
+        HttpHeaders.ETAG: staticFile.etag,
         HttpHeaders.CACHE_CONTROL: 'max-age: ${_staticMaxAge.inSeconds}',
```
kevmoo: I think this should be `'public, max-age: ...`

app/lib/frontend/handlers.dart
```diff
@@ -233,14 +233,17 @@ Future<shelf.Response> staticsHandler(shelf.Request request) async {
...
+    final ifNoneMatch = request.headers[HttpHeaders.IF_NONE_MATCH];
+    if (ifNoneMatch != null && ifNoneMatch == staticFile.etag) {
+      return new shelf.Response.notModified();
+    }
     return new shelf.Response.ok(
       staticFile.bytes,
       headers: {
         HttpHeaders.CONTENT_TYPE: staticFile.contentType,
         HttpHeaders.CONTENT_LENGTH: staticFile.bytes.length.toString(),
         HttpHeaders.LAST_MODIFIED: formatHttpDate(staticFile.lastModified),
```
kevmoo: Remove this one, then. Times may change based on deploys...but etags are valid as long as the content stays the same

isoos:
> Shouldn't we drop the time header then?

They work in tandem: if either of them identifies that the content hasn't changed, we return the HTTP 304 status code. It is cheap to keep both.

app/lib/frontend/handlers.dart
isoos: AFAICT `public` refers to proxies and other caching between the browser and the server. I'd keep those from caching.

kevmoo:
> They work in tandem: if either of them identifies that the content hasn't changed, we return the HTTP 304 status code. It is cheap to keep both.

Cool...

PR #847, opened 2017-12-21
Search: scale down document size, let the overall score do the ranking.

isoos:
Based on top of #846.

Previously, when an expression e.g. `haversine` was present in a much longer text, the ranking gave it a lower score and the end result was further down, but it may have been only because of the more detailed example. With this tweak, we still prefer matches on shorter text, but the difference is more subtle.

Another effect: searching for `AngularDart` will not rank the `angular` package that behind, it will be likely on the first page.


kevmoo:
@isoos – you'll have to rebase, it seems...

isoos:
> you'll have to rebase, it seems...

done.

PR #853, opened 2017-12-22
A bunch of cleanup to site metadata, titles, etc

kevmoo:


kevmoo:
> This is a true winner of the "Unrelated commits in a single PR in 2017" award.

😜

PR #856, opened 2017-12-22
Feedback on unsupported search qualifier.

### LINKED ISSUE #826: Be clear if someone tries to use a search qualifier that isn't supported.
If I put in `mail:bob@gmail.com` or something, tell me "`mail` is not a supported qualifier see http://XYZ for the list of supported qualifiers".

This info should be sent back from the search service – so it's based on what the currently deployed service supports, etc.

isoos:
Fixes #826. Ugly, but can be improved later.

screen shot 2017-12-22 at 11 16 31 am

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -107,6 +107,8 @@ class TemplateService {
 
     final PlatformDict platformDict = getPlatformDict(currentPlatform);
     final isSearch = searchQuery != null && searchQuery.hasQuery;
+    final unsupportedQualifier =
+        isSearch && searchQuery.parsedQuery.text.contains(':');
     final String sortValue = serializeSearchOrder(searchQuery?.order) ??
         (isSearch ? 'search_relevance' : 'listing_relevance');
     final SortDict sortDict = getSortDict(sortValue);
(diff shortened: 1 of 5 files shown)


app/lib/frontend/templates.dart
```diff
@@ -104,6 +104,8 @@ class TemplateService {
 
     final PlatformDict platformDict = getPlatformDict(currentPlatform);
     final isSearch = searchQuery != null && searchQuery.hasQuery;
+    final unsupportedQualifier =
+        isSearch && searchQuery.parsedQuery.text.contains(':');
```
kevmoo: I'm guessing we should support a literal `"cool:beans"` we want to support that, but it'd be caught here?

...part of making this sarter?

isoos: The `parsedQuery.text` is the remainder, it is whatever is left over processing the other known `X:Y` patterns. Right now `"explicit phrases"` are not part of that processing (although they should), but it could be addressed in a different/unrelated PR.

kevmoo: kk

kevmoo:
Travis is sad 😦

isoos:
> Travis is sad 

Fixed. PTAL

PR #859, opened 2017-12-22
Use hash on script.dart.js for longer caching

isoos:
- Makes ETag shorter (half in size).
- Changes the URL structure of static resources (only the `script.dart.js` for now, if we like the pattern then rest can be migrated to it).
- Increased the cache period for resources with matching hash (with protection agains version mismatches).

### DIFF
--- a/app/lib/frontend/static_files.dart
+++ b/app/lib/frontend/static_files.dart
@@ -48,7 +48,9 @@ class StaticsCache {
     final bytes = file.readAsBytesSync();
     final lastModified = file.lastModifiedSync();
     final String relativePath = path.relative(file.path, from: rootPath);
-    final String etag = crypto.sha256.convert(bytes).toString();
+    final digest = crypto.sha256.convert(bytes);
+    final String etag =
+        digest.bytes.map((b) => (b & 31).toRadixString(32)).join();
     return new StaticFile(relativePath, contentType, bytes, lastModified, etag);
   }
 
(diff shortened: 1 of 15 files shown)


app/lib/frontend/static_files.dart
```diff
@@ -96,4 +99,24 @@ class StaticUrls {
...
+
+/// Returns the URL of a static resource
+String _getCacheableStaticUrl(String relativePath) {
+  if (!relativePath.startsWith('/')) {
+    relativePath = '/$relativePath';
+  }
+  final String requestPath = '${staticUrls.staticPath}$relativePath';
+  final file = staticsCache.getFile(requestPath);
+  if (file == null) {
+    throw new Exception('Static resource not found: $relativePath');
```
jakobr-google: I don't think we want to throw here, since this is called when initializing `_assets` lazily, by renderLayoutPage(), which (I'm guessing) would result in a 500 error message thrown back to users.

Maybe log an error instead, and fall back to `requestPath`? Then they should get a nice 404 when they follow the src href instead.

isoos: This exception is thrown mostly because it can act as an early-warning in tests: if we reference something that does not exists anymore, we'll get notified by the failed test.

jakobr-google: True, we'll pick it up early if this happens.

PR #864, opened 2018-01-01
Add /sitemap.txt

kevmoo:
Fixes https://github.com/dart-lang/pub-dartlang-dart/issues/860


app/lib/frontend/handlers.dart
```diff
@@ -101,6 +101,7 @@ const _handlers = const <String, shelf.Handler>{
   '/web/packages': webPackagesHandlerHtml,
   '/debug': debugHandler,
   '/feed.atom': atomFeedHandler,
+  '/sitemap.txt': _siteMapHandler,
```
isoos: Why not an xml sitemap? IIRC that is more widely adopted by search engines other than Google.

kevmoo:
It takes ~3s to load. Not to bad.

PR #870, opened 2018-01-03
Fix(?) sitelink search box metadata

isoos:
#708 was deployed weeks ago, yet, Google doesn't display the search box in the sitelinks section. As most online examples (including the official one) uses the `search_term_string` expression, I'd rather use that too, just to make sure we are doing it by the book.

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -549,6 +549,7 @@ class TemplateService {
         ? null
         : serializeSearchOrder(searchQuery.order);
     final platformDict = getPlatformDict(platform);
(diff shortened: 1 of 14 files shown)


kevmoo:
https://developers.google.com/search/docs/data-types/sitelinks-searchbox

"Add this markup only to the homepage, not to any other pages."

Maybe this is the issue?

kevmoo:
Er, sorry. I'd update this PR to put the search info on the root page, too.

Then LGTM

isoos:
> "Add this markup only to the homepage, not to any other pages."

Ah! Will update it.

isoos:
Updated. PTAL.

PR #873, opened 2018-01-04
Reject robots unless it is the app on pub.dartlang.org

### LINKED ISSUE #872: When the host is not pub.dartlang.org, have robots.txt block the whole domain
We're getting a crazy number of search hits against our non-primary domain.

We should just block them all.

If a request for `robots.txt` comes in and it's not for the `pub.dartlang.org` host, send a go-away response.

```
Disallow: /
```

isoos:
Fixes #872

### DIFF
--- a/app/lib/dartdoc/handlers.dart
+++ b/app/lib/dartdoc/handlers.dart
@@ -17,6 +17,8 @@ Future<shelf.Response> dartdocServiceHandler(shelf.Request request) async {
   final path = request.requestedUri.path;
   final handler = {
     '/debug': debugHandler,
+    // TODO: have a proper robots.txt after we are serving content
+    '/robots.txt': rejectRobotsHandler,
   }[path];
 
   if (handler != null) {
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -79,6 +79,8 @@ Future<shelf.Response> appHandler(
     return _packageHandler(request);
   } else if (path.startsWith('/doc')) {
     return _docHandler(request);
+  } else if (path == '/robots.txt' && !_isProd(request)) {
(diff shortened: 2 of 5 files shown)


app/lib/frontend/handlers.dart
```diff
@@ -82,8 +82,13 @@ Future<shelf.Response> appHandler(
   } else if (path.startsWith(staticUrls.staticPath)) {
     return _staticsHandler(request);
   } else if (staticRootFiles.contains(path)) {
-    return _staticsHandler(request,
-        pathOverride: '${staticUrls.staticPath}$path');
+    print(path);
```
jakobr-google: Remove the print statements, or use debug logging?

app/lib/frontend/handlers.dart
```diff
@@ -82,8 +82,13 @@ Future<shelf.Response> appHandler(
...
+    var pathOverride = '${staticUrls.staticPath}$path';
+    if (path == '/robots.txt' && !_isProd(request)) {
+      pathOverride = '${staticUrls.staticPath}/no_robots.txt';
```
jakobr-google: Could this be
```dart
return rejectRobotsHandler(request);
```
instead? (So we only have the reject robots.txt contents in one place.)

app/lib/dartdoc/handlers.dart
```diff
@@ -17,6 +17,7 @@ Future<shelf.Response> dartdocServiceHandler(shelf.Request request) async {
   final path = request.requestedUri.path;
   final handler = {
     '/debug': debugHandler,
+    '/robots.txt': rejectRobotsHandler,
```
kevmoo: We might want to add a todo here about removing when this app has content...right?

kevmoo:
@isoos – travis be sad

PR #876, opened 2018-01-04
Sitemap improvements

kevmoo:


static/robots.txt
```diff
@@ -1,2 +1,3 @@
 User-agent: *
 Disallow: /experimental/
+sitemap: https://pub.dartlang.org/sitemap.txt
```
kevmoo: See https://developers.google.com/search/reference/robots_txt#sitemap

app/lib/frontend/backend.dart
```diff
@@ -77,9 +77,12 @@ class Backend {
   }
 
   /// Retrieves the names of all packages, ordered by name.
-  Stream<String> allPackageNames() {
+  Stream<String> allPackageNames({DateTime updatedSince}) {
     final query = db.query(models.Package);
-    return query.run().map((p) => (p as models.Package).name);
+    return query.run().where((p) {
+      return updatedSince == null ||
```
isoos: why not put this into the query?

kevmoo: Oh!...uh...I guess I could. Wasn't sure how...

kevmoo: Done!

Issue #888, opened 2018-01-06
`outdated` package showing up as `[awaiting]` on package site


kevmoo:
See https://dartlang-pub.appspot.com/packages/shelf_rpc

and https://analyzer-dot-dartlang-pub.appspot.com/packages/shelf_rpc/0.0.3+2

Oops?

isoos:
We have more than one issue here:
- the text that "We analyzed this package...", while we are not doing it at all
- the `[awaiting]` flag on the package extract and the package show page
- there is a bug in the logic that decides not to analyze too old packages and we skip the latest stable ones too

I believe that the `[awaiting]` will be resolved if we fix the later one, because the awaiting is coming from the data of the latest stable.

isoos:
Most of this is fixed now, there is still an issue with the `[awaiting]` label on pages like https://dartlang-pub.appspot.com/packages/shelf_rpc/versions/0.0.3

mit-mit:
@isoos @jakobr-google looks like this is fixed now?

isoos:
yeah, fixed.

PR #894, opened 2018-01-08
Fix analysis tab when package version is outdated.

isoos:
When a package version is too old, we provide a link to the latest stable's analysis.

#888

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -145,7 +145,8 @@ class TemplateService {
   }
 
   /// Renders the `views/pkg/analysis_tab.mustache` template.
-  String renderAnalysisTab(AnalysisExtract extract, AnalysisView analysis) {
+  String renderAnalysisTab(
+      String package, AnalysisExtract extract, AnalysisView analysis) {
     if (analysis == null || !analysis.hasAnalysisData) return null;
 
     String statusText;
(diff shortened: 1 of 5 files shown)


app/lib/frontend/templates.dart
```diff
@@ -191,6 +192,9 @@ class TemplateService {
         devDeps.isNotEmpty;
 
     final Map<String, dynamic> data = {
+      'package': package,
```
jakobr-google: Tiny nit: We also use 'package' in the dependencies lists. I know it works due to scoping, but maybe use a different name here?

app/lib/frontend/templates.dart
```diff
@@ -191,6 +192,9 @@ class TemplateService {
...
+      'show_outdated': analysis.analysisStatus == AnalysisStatus.outdated,
```
jakobr-google: Nit: Could also use `{{^show_content}}` in the template to avoid passing in two booleans for this state.

isoos: Updated, although with the other flag (using `{{^show_outdated}}` in the template)

app/lib/frontend/templates.dart
isoos: I kind of like the compact name and rely on the scope, but renaming can also work. Any suggestion on how they should be different?

jakobr-google: It's fine as is. I just stumbled across it, and had to double-check how mustache worked.

PR #895, opened 2018-01-08
Fix score box title when version is outdated.

isoos:
This is visible only on the package show page's tab, when hovering on the score box with the outdated version.

#888


isoos:
> Shouldn't we have a test for it, though?

I've realized that we'll need to change the tags section too, I'll update this shorty with the tests.

isoos:
PTAL. I've updated the platform tags and added test too.

isoos:
Fixed travis.

PR #899, opened 2018-01-08
Extended package meta description

### LINKED ISSUE #886: Add more to the default description of each package
Instead of 
`<pkg name> [<version>] - <pkg description`

Go with
`<pkg name> [<version>] [Flutter and] Dart Package - <pkg description`

Where `[Flutter and]` is included if the package is flutter-only!

isoos:
Fixes #886.

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -463,11 +463,18 @@ class TemplateService {
     );
     final content = _renderTemplate('pkg/show', values);
     final versionString = isVersionPage ? '${selectedVersion.id} ' : '';
+    var pageDescription =
+        '${selectedVersion.package} ${selectedVersion.version}';
+    if (isFlutterPlugin) {
+      pageDescription += ' Flutter and Dart package';
+    } else {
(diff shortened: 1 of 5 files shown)


app/lib/frontend/templates.dart
```diff
@@ -463,11 +463,18 @@ class TemplateService {
     );
     final content = _renderTemplate('pkg/show', values);
     final versionString = isVersionPage ? '${selectedVersion.id} ' : '';
+    var pageDescription =
+        '${selectedVersion.package} ${selectedVersion.version}';
+    if (isFlutterPlugin) {
```
jakobr-google: Nit: This name is a bit misleading. What we want to capture (and do) are Flutter packages -- ie. packages that depend on Flutter. A Flutter plugin is a special kind of Flutter package that also contains native code. `isFlutterPackage` would be a better name for this variable.

isoos: I'll update that in a subsequent PR. It has a bit of history, and both the name and its initialization could be updated.

PR #903, opened 2018-01-08
Metadata: Twitter card and OpenGraph.

### LINKED ISSUE #885: Should add OpenGraph items to pages, too
I hate duplicating content, but it seems folks like Twitter want it

https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started

isoos:
Fixes #885.

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -473,14 +473,25 @@ class TemplateService {
     } else {
       pageDescription += ' Dart package';
     }
+    final pageTitle =
+        '${package.name} $versionString| ${isFlutterPackage ? 'Flutter' : 'Dart'} Package';
     pageDescription += ' - ${selectedVersion.ellipsizedDescription}';
-    return renderLayoutPage(PageType.package, content,
-        title:
-            '${package.name} $versionString| ${isFlutterPackage ? 'Flutter' : 'Dart'} Package',
-        pageDescription: pageDescription,
-        faviconUrl: isFlutterPackage ? staticUrls.flutterLogo32x32 : null,
-        canonicalUrl:
-            isVersionPage ? _canonicalUrlForPackage(package.name) : null);
+    final canonicalUrl =
+        isVersionPage ? _canonicalUrlForPackage(package.name) : null;
+    final headerHtml = _renderTemplate('pkg/meta_header', {
+      'title': pageTitle,
+      'description': pageDescription,
+      'canonicalUrl': canonicalUrl,
+    });
+    return renderLayoutPage(
+      PageType.package,
+      content,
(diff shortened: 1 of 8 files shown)


app/test/frontend/golden/pkg_show_page.html
```diff
@@ -22,6 +22,13 @@
       ga('send', 'pageview');
     </script>
     <script async="" defer="" src="//www.google.com/insights/consumersurveys/async_survey?site=5wdvu4vc5lhop3xbatxof6lzcm"></script>
+    <meta name="twitter:card" content="summary" />
```
kevmoo: If I ask you to fix the indent, will you be mad @isoos ?

isoos: Either the template or the output sucks. If you are fine with that tradeoff, it can be indented.

app/views/pkg/meta_header.mustache
```diff
@@ -0,0 +1,12 @@
+{{! Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+    for details. All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file. }}
+<meta name="twitter:card" content="summary" />
+<meta name="twitter:title" content="{{title}}" />
```
kevmoo: See https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started#opengraph – at the bottom

Don't duplicate title/description w/ twitter:

Also, `<meta name="twitter:site" content="@dart_lang" />`

app/lib/frontend/templates.dart
```diff
@@ -473,14 +473,25 @@ class TemplateService {
...
+        isVersionPage ? _canonicalUrlForPackage(package.name) : null;
+    final headerHtml = _renderTemplate('pkg/meta_header', {
+      'title': pageTitle,
+      'description': pageDescription,
+      'canonicalUrl': canonicalUrl,
+    });
+    return renderLayoutPage(
+      PageType.package,
+      content,
+      title: pageTitle,
```
isoos: I've started this with the concept of having different title and description here, but all the guides I check say that it is better if they are the same across the different metadata definitions. Shall I move these to the `layout.mustache` (now or in a follow-up)?

kevmoo: Follow-up is fine. Keeping them the same is good...

kevmoo: @isoos – should we have a bug open for the follow-up?

isoos: Prepared it: https://github.com/dart-lang/pub-dartlang-dart/pull/907

Issue #842, opened 2017-12-20
Separate homepage and repository fields for a package


tejainece:
Right now, Dart only supports specifying homepage field.

Right now, people are using their homepage URL in that field. This makes it harder to reach the actual project repository.

Would be nice if this homepage field is split into two:

+ Repository
+ Homepage

Here is an example of one of my projects:

![image](https://user-images.githubusercontent.com/2513673/34222521-21d0616c-e5bc-11e7-8d17-9230c17ff009.png)

I would like to have:

+ Repository: https://github.com/Jaguar-dart/jaguar
+ Homepage: https://jaguar-dart.github.io/

isoos:
@kevmoo @mkustermann

kevmoo:
Interesting idea!

xqwzts:
Fortuitous timing, I was trying to see how one might implement links to source files in dartdoc [for https://github.com/dart-lang/dartdoc/issues/73], this would definitely help.

kevmoo:
Nice! We'd want to align w/ the right commit, etc etc.

We'll take small steps...

isoos:
Closing this, as the use case is considered too rare.

My personal suggestion is that you fill the `homepage:` field with the github repo URL, and in the settings of that repo you can add the URL of the "traditional" homepage.

PR #904, opened 2018-01-09
Separate homepage and repository for a package

isoos:
- introduces a new `repository` field in `pubspec.yaml`
- if only `homepage` is specified and it is a known repository url, it will be displayed as "Repository" and not as "Homepage"
- if both are specified, both will be displayed

Deployed in staging, a few examples:
- https://dartlang-pub-dev.appspot.com/packages/http
- https://dartlang-pub-dev.appspot.com/packages/jaguar

### DIFF
--- a/app/lib/frontend/model_properties.dart
+++ b/app/lib/frontend/model_properties.dart
@@ -78,6 +78,11 @@ class Pubspec {
     return _asString(_json['homepage']);
   }
 
+  String get repository {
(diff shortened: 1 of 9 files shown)


isoos:
Fixes #842

kevmoo:
Chatted w/ @mit-mit about this. There is no simple fix here. Not a high priority now.

Issue #712, opened 2017-12-04
Never explain what '100' is


kevmoo:
Not mentioned on the help page. It should.

kevmoo:
Per @lukechurch : user should never have to go to a help page to understand this.

isoos:
I've experimented with a few things here, but couldn't really decide what would work best. E.g. a simple change would be to add the "/ [maximum value]" and/or a text-based rating. The following screenshot has them as alternatives:

<img width="254" alt="screen shot 2018-01-09 at 5 19 35 pm" src="https://user-images.githubusercontent.com/4778111/34731041-66c5d158-f561-11e7-9b23-82713084285e.png">

We could do a bit more graphical representation like [pie chart](https://cloud.netlifyusercontent.com/assets/344dbf88-fdf9-42bb-adb4-46f01eedd629/24ff6c2c-378e-4880-9653-3fc9351cbf4d/pie-chart-08.png) or some progress [bar-like design](https://i.pinimg.com/originals/f7/2a/86/f72a860b2a6fb62f3ed1c6c5c3d5a986.jpg).

@kevmoo: any preference?

kevmoo:
We should chat with @mit-mit

mit-mit:
The simple `44 / 100` option SGTM. I doubt users have trouble understanding 0 is bad and 100 is good.

mit-mit:
This was fixed

PR #906, opened 2018-01-10
Score range indicator on the analysis tab.

isoos:
screen shot 2018-01-10 at 10 38 42 am

- I think the overall score doesn't need the same indicator, it feels less crowded without it.
- Using a new cell in the table layout increases the gap between the actual value and the "/ 100", but keeps the style and layout simple.


isoos:
Fixes #712

kevmoo:
Thoughts, @mit-mit ?

mit-mit:
UI LGTM!

Over to cc @jakobr-google for review.

PR #912, opened 2018-01-10
Separate stable and dev versions on the all versions page.

### LINKED ISSUE #693: Sorting of package versions could be more clear
https://pub.dartlang.org/experimental/packages/test/versions

This doesn't seem to be sorted as expected either by the date or the version:

screen shot 2017-12-01 at 18 52 52

isoos:
Fixes #693

### DIFF
--- a/app/test/frontend/golden/pkg_versions_page.html
+++ b/app/test/frontend/golden/pkg_versions_page.html
@@ -74,7 +74,7 @@
 
 <main>
   
-<h1>All versions of foobar</h1>
+<h1>Stable versions of foobar</h1>
 <table class="version-table">
   <thead>
   <tr>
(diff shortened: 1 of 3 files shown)


app/test/frontend/golden/pkg_versions_page.html
```diff
@@ -74,7 +74,7 @@ <h1 class="_visuallyhidden">Dart pub</h1>
 
 <main>
   
-<h1>All versions of foobar</h1>
+<h1>Stable versions of foobar</h1>
```
kevmoo: Could you post a screen shot? Or update the template sample to render both flavors?

isoos: deploying it to staging, check the following page in ~10 minutes: https://dartlang-pub-dev.appspot.com/packages/test/versions

kevmoo:
Would love an option to see these merged w/ a checkbox – but we can do that later.

PR #918, opened 2018-01-15
Fix flutter version tag.

jakobr-google:
The Flutter repo uses 'vX.Y.X' as the format for version tags.


isoos:
You'll need to update the `analyzer_version_test.dart` too to make it passing.

PR #920, opened 2018-01-15
Link to Flutter-specific search expression on the Flutter landing page.

### LINKED ISSUE #764: How to filter flutter only packages in the new UI?
URL: https://pub.dartlang.org/flutter/packages

I find out the webpage was updated in recent two days. The new UI is pretty cool. I just have one question what if I want to filter and check flutter only packages? Is there a way to achieve that?

I remembered it was like 7 or 8 pages of flutter packages in the old days, but with the new UI now, nearly all packages will have the `Flutter` tag. I know the packages worked in the web and server will also work in flutter. But I just want to check a list of flutter only packages so that I can understand what flutter can provide now, for example: camera, bluetooth access, sqlite, shared references etc.. They are related to flutter only. 

Thanks a lot.

isoos:
Closes #764

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -548,9 +548,16 @@ class TemplateService {
     String platform,
   ) {
     final platformDict = getPlatformDict(platform);
+    final packagesUrl = platform == null ? '/packages' : '/$platform/packages';
+    final morePackages = 'More ${platformDict.name} packages...';
+    final links = <String>[
+      '<a href="$packagesUrl">${_htmlEscaper.convert(morePackages)}</a>'
+    ];
+    if (platform == KnownPlatforms.flutter) {
+      links.add(flutterSpecificPackagesHtml);
+    }
     final values = {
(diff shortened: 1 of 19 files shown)


app/lib/frontend/templates.dart
```diff
@@ -543,9 +543,20 @@ class TemplateService {
...
+    final morePackages = 'More ${platformDict.name} packages...';
+    final moreLinks = <String>[
+      '<a href="$packagesUrl">${_htmlEscaper.convert(morePackages)}</a>'
+    ];
+    if (platform == KnownPlatforms.flutter) {
+      final url = new Uri(
+          path: '/packages', queryParameters: {'q': 'dependency:flutter'});
+      final flutterOnly = 'Flutter-specific packages...';
+      moreLinks.add(
+          '<a href="${_attrEscaper.convert(url.toString())}">${_htmlEscaper.convert(flutterOnly)}</a>');
```
jakobr-google: Could inline $flutterOnly here. There's not much reason to escape it, since it's our own constant.

We could technically make the entire string a constant, since the url is also hard-coded.

Issue #858, opened 2017-12-22
How to resolve missed platform classifications?


isoos:
We have a couple of interesting cases open:

- https://github.com/dart-lang/pub-dartlang-dart/issues/598, https://github.com/dart-lang/pub-dartlang-dart/issues/634 - build tools for apps targeting the `web`, classified as `server`.
- https://github.com/dart-lang/pub-dartlang-dart/issues/790, https://github.com/dart-lang/pana/issues/188 build tools for all kind of apps, classified as `server`
- https://github.com/dart-lang/pub-dartlang-dart/issues/709 "server" library for `NodeJs` apps, to be used with `dart2js`, classified as `web`

The common theme in all of these is the difference between the use of the library and the target platform of the app that is built using it.

I can see a couple of ways to go ahead:

**1. No action in the classification.** 

There are packages that do provide build and code generation utilities, and are not mis-categorized (e.g. [pwa](https://pub.dartlang.org/packages/pwa) generates code, but all of that lives in `bin/`, so there is no hint that it would be a `server` package).

If we identify these patterns, and give guidance to package authors, eventually most of these will be corrected.

Related questions:
- What is the ideal pattern (besides hiding the dart:io code in `bin/`)?
- Shall we add a suggestion for it in `pana` (e.g. detecting somebody relies on `source_gen`)?

**2. Metadata from `pubspec.yaml`**.

Keywords (https://github.com/dart-lang/pub-dartlang-dart/issues/368), categories (https://github.com/dart-lang/pub-dartlang-dart/issues/367) or a property for platforms could be coming from `pubspec.yaml`, and the UI could display it along the ones we are detecting.

Related questions:
- How to display them if they are not the same? (colors, format, order, icon.. ?)
- Which of it takes preference (e.g. in search)?
- Should this affect the score or ranking in any way?

**3. Hand-coded exception in the `pana` analysis.**

We could assume that if a library *uses* certain helper libraries (through transitive dependencies), than it is a `build-utility` or a `nodejs` library. If/when `dart:io` is used only through them, the package could be classified as generic, unless other API use restrict it.

Related questions:
- Is it the case really that if something is a build tool, it can be applied to every platform? I can't think of a counter-example, but there may be one.
- Shall we put NodeJs support in a different category? It is an outlier compared to the build tools. 

**4. New platform for `build` utilities.**

[...]

PR #923, opened 2018-01-16
Removing 'Server' platform from the UI.

isoos:
This is the first part of the short-term fix around the platform classifications (#858). We cannot provide the same level of classification consistency for 'Server' (including build tools and Node.js) as for Web or Flutter, and in many cases this results in a wrong classification. This removes:
- the landing blurb text
- the platform selection tab
- the package platform tag
- the listing page

/cc @mit-mit

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -136,8 +136,8 @@ Future<shelf.Response> _flutterLandingHandler(shelf.Request request) =>
     _indexHandler(request, KnownPlatforms.flutter);
 
 /// Handles requests for /server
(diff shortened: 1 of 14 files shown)


mit-mit:
I'm reviewing this on the product side also; please hold a bit until I give a green light!

mit-mit:
Clear on the product side, we can remove the Server landing page, https://pub.dartlang.org/server.

isoos:
> Removing 'Server' makes me sad.

Me too, but I'm getting optimistic with new platform components.
Unfortunately there is this period where the new is not ready yet, but we already deprecated the old :[

PR #924, opened 2018-01-16
Hardened GC of obsolete packages.

isoos:
- Moved GC to the beginning of the analysis, which should reduce the risk of eventual consistency issues (within the same analyzer session).

- Made sure that the referenced Analysis is not deleted (instead of keeping the latest one, which is the likely cause of our current data inconsistency).

### DIFF
--- a/app/lib/analyzer/backend.dart
+++ b/app/lib/analyzer/backend.dart
@@ -275,14 +275,15 @@ class AnalysisBackend {
   /// - it is **not** the latest for the pana version, or
   /// - it is older than 6 months (except if it is the latest one).
   Future deleteObsoleteAnalysis(String package, String version) async {
+    final pvaKey = db.emptyKey
+        .append(PackageAnalysis, id: package)
+        .append(PackageVersionAnalysis, id: version);
+    final PackageVersionAnalysis pva = (await db.lookup([pvaKey])).single;
+    if (pva == null) return;
+
     final DateTime threshold =
         new DateTime.now().toUtc().subtract(obsoleteThreshold);
-    final Query scanQuery = db.query(
-      Analysis,
-      ancestorKey: db.emptyKey
-          .append(PackageAnalysis, id: package)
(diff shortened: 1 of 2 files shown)


app/lib/analyzer/backend.dart
```diff
@@ -302,9 +307,12 @@ class AnalysisBackend {
...
     }
 
-    // sanity check that we keep the latest, even if it is too old
-    if (obsoleteKeys.length == existingAnalysis.length) {
-      obsoleteKeys.removeLast();
+    // sanity check that we don't delete everything
+    if (obsoleteKeys.isNotEmpty &&
+        obsoleteKeys.length == existingAnalysis.length) {
+      _logger.warning(
+          'Inconsitency: all Analysis entries are marked as obsolete for $package $version.');
```
jakobr-google: s/Inconsitency/Inconsistency/

app/lib/analyzer/backend.dart
```diff
@@ -275,21 +275,26 @@ class AnalysisBackend {
   /// - it is **not** the latest for the pana version, or
   /// - it is older than 6 months (except if it is the latest one).
   Future deleteObsoleteAnalysis(String package, String version) async {
+    final pvaKey = db.emptyKey
+        .append(PackageAnalysis, id: package)
+        .append(PackageVersionAnalysis, id: version);
+    final PackageVersionAnalysis pva = (await db.lookup([pvaKey])).single;
```
jakobr-google: `.single` will throw an exception if there's not exactly one element in the list. I don't think that's what we want here.

isoos: The `db.lookup` will return with a List the same size as the input list, and `.single` is just as good as `.first` or `[0]`. I can change it to either of them.

jakobr-google: In that case, `.single` should be fine. Thanks!

PR #925, opened 2018-01-17
Rewrite relative URLs of both links and images.

### LINKED ISSUE #850: How to handle README/CHANGELOG files w/ relative links
See https://pub.dartlang.org/packages/video_player/versions/0.0.7

Works great here: https://github.com/flutter/plugins/blob/8f8d0ae1d6d026bb287ceb14770c263a4ec20b8a/packages/video_player/README.md

Because this link is relative to assets on github
```markdown
![The example app running in iOS](doc/demo_ipod.gif)
```

Solution ideas:

1. Warn in pana if README/CHANGELOG items use relative links – encourage absolute links
2. Get clever about the homepage URL and try to figure out the best thing?

isoos:
Fixes #850.

Unfortunately having more then one InlineSyntax in the inlineSyntaxes list didn't work out as expected: either the links or the images were updated, but never both. Instead, doing the transformation on the tree output of the parser.

I've also fixed the test with the root URLs (e.g. '/prefixed/with/slash' should be kept intact).

### DIFF
--- a/app/lib/shared/markdown.dart
+++ b/app/lib/shared/markdown.dart
@@ -4,7 +4,6 @@
 
 import 'package:logging/logging.dart';
 import 'package:markdown/markdown.dart' as m;
-import 'package:path/path.dart' as p;
 
 import 'package:pana/src/download_utils.dart' show getRepositoryUrl;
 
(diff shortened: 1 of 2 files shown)


isoos:
Updated with rebase for the latest pana, now it does use the raw urls. I'm not sure why travis doesn't seem to do any work...

isoos:
> absolute paths won't really work

I think they do. If you are verifying locally, could you check please `/packages/video_player/versions/0.0.7`?

jakobr-google:
That one works, but it's also using an absolute URI, AFAICT. I meant relative URIs with absolute paths, ie. '/example/video.png'.

isoos:
> That one works, but it's also using an absolute URI, AFAICT. I meant relative URIs with absolute paths, ie. '/example/video.png'.

Ah, good point. I was planning to change the code in `pana` for it, but I realized I can do it here first, and in `pana` we don't really have a use case for the absolute urls as an input. PTAL.

PR #930, opened 2018-01-18
Upgrade to latest build_runner and json_serializable

kevmoo:


app/build.yaml
```diff
@@ -0,0 +1,19 @@
...
+  pub_dartlang_org:
+    sources:
+      - 'lib/shared/*.dart'
+      - 'lib/search/backend*.dart'
+    builders:
+      json_serializable:
+        options:
+          require_library_directive: false
+          header: |+
+           // Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
```
jakobr-google: 2018?

app/pubspec.lock
```diff
@@ -604,4 +590,4 @@ packages:
     source: hosted
     version: "2.1.13"
 sdks:
-  dart: ">=2.0.0-dev <=2.0.0-dev.15.0"
+  dart: ">=2.0.0-dev <=2.0.0-dev.17.0"
```
jakobr-google: We'll need to update the base image in Dockerfile too, then, right?

app/build.yaml
```diff
@@ -0,0 +1,19 @@
+# Read about `build.yaml` at https://pub.dartlang.org/packages/build_config
+# To update generated code, run `pub run build_runner build`
+targets:
+  pub_dartlang_org:
+    sources:
+      - 'lib/shared/*.dart'
+      - 'lib/search/backend*.dart'
+    builders:
+      json_serializable:
```
jakobr-google: So, do `json_serializable` do both `JsonSerializableGenerator` and `JsonLiteralGenerator`?

app/pubspec.lock
isoos: I think that can be independent of this upper bound constraint. I had an issue with dev.17 running a local instance, not yet sure what was causing it.

app/build.yaml
kevmoo: I forget – but we don't need the LiteralGenerator any more

app/build.yaml
kevmoo: I left it like this to keep the other files unchanged – when we change something else, we can bump those...

PR #938, opened 2018-01-22
Fix NPE.

jakobr-google:
Memcache temporarily has analysis extracts with platform==server, which we don't have a platformDict for.


jakobr-google:
(fix already deployed to prod)

PR #943, opened 2018-01-22
Prepare dartdoc service for serving the same URLs as dartdocs.org.

isoos:


app/lib/dartdoc/handlers.dart
```diff
@@ -65,3 +67,51 @@ Future<shelf.Response> packageHandler(shelf.Request request) async {
...
+DocFilePath parseRequestUri(Uri uri) {
+  final pathSegments = uri.pathSegments;
+  if (pathSegments.length < 3) {
+    return null;
+  }
+  if (pathSegments[0] != 'documentation') {
+    return null;
+  }
+  final String package = pathSegments[1];
+  final String version = pathSegments[2];
```
jakobr-google: The comment on `documentationHandler` claims we also support `/documentation/<package>`, so shouldn't `version` be optional?

isoos: on the current dartdoc site there is no redirect from `/documentation/<package>` to `/documentation/<package>/<latest_version>`, and with that, the version is not optional. Fixed the documentation, and later we can add the redirect.

PR #951, opened 2018-02-01
Upgrade dependencies

isoos:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM google/dart-runtime-base:2.0.0-dev.15.0
+FROM google/dart-runtime-base:2.0.0-dev.20.0
```
kevmoo: Of course I think of this *after* I hit merge.

We've started to add members to the Dart SDK that may cause non-bleeding edge packages to break – I think.

Should be careful to watch errors as we roll this out.

CC @mit-mit

PR #964, opened 2018-02-06
Make dartdoc_runner actually running.

isoos:
Various fixes:
- store build info in `_build_log.txt`
- store metadata in the output dir and not in the pkg dir, also renamed to `_metadata.json`
- pub cache dir needs to be created (not sure why)
- fixed dartdoc version parsing

There is an issue with the pub cache dir not being populated (and probably the system-level dir is used instead), but I couldn't figure out that yet, the rest seems to be working regardless.

### DIFF
--- a/app/lib/dartdoc/dartdoc_runner.dart
+++ b/app/lib/dartdoc/dartdoc_runner.dart
@@ -11,12 +11,14 @@ import 'package:pana/pana.dart';
 import 'package:pana/src/download_utils.dart';
 import 'package:path/path.dart' as p;
 
+import '../shared/configuration.dart' show envConfig;
 import '../shared/task_scheduler.dart' show Task, TaskRunner;
 
 final Logger _logger = new Logger('pub.dartdoc.runner');
 final String _hostedUrl = 'https://www.dartdocs.org';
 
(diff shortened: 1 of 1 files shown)


isoos:
Found culprit of the pub cache dir use: https://github.com/dart-lang/pana/pull/245

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -35,51 +37,75 @@ class DartdocRunner implements TaskRunner {
...
-      ['--output', outputDir],
-      workingDirectory: pkgPath,
-    );
+  Future _resolveDependencies(
+      PubEnvironment pubEnv, Task task, String pkgPath) async {
+    final pr = await pubEnv.runUpgrade(pkgPath, /* isFlutter */ false);
     if (pr.exitCode != 0) {
-      _logger.severe('Error while running dartdoc for $task.\n'
+      print(pr.stdout);
+      print(pr.stderr);
```
jakobr-google: Debug leftovers?

isoos: Yeah, thanks for spotting, removed.

Issue #289, opened 2017-08-08
Allow marking a package as "discontinued"


matanlurey:
Pub currently ranks packages based on the `updated` column:

<img width="939" alt="screen shot 2017-08-08 at 8 32 43 am" src="https://user-images.githubusercontent.com/168174/29080258-2d8b3480-7c14-11e7-95f7-9336b2deb19e.png">

This works great, until it doesn't. For the `angular` package, there is a historic, unsupported/maintained `1.x` version that is presented instead of the newest `4.0.0-alpha` version:

<img width="936" alt="screen shot 2017-08-08 at 8 33 36 am" src="https://user-images.githubusercontent.com/168174/29080310-4fc56c14-7c14-11e7-8cd3-78eaaf579da7.png">
<img width="868" alt="screen shot 2017-08-08 at 8 33 42 am" src="https://user-images.githubusercontent.com/168174/29080311-4fcc6c26-7c14-11e7-92e6-25601d29bc64.png">

Under normal circumstances, this might be fine, i.e. ranking `stable` releases higher than `unstable` for viewing preference. However, package authors should be given an option (perhaps in an admin, or manually via a ticket system for the time being) to _deprecate_ a version range.

In this case, imagine the following metadata was attached to the `angular` package:

```json
{
  "deprecated": [
    "<4.0.0-alpha",
  ],
}
```

And the ranking was instead `stable` > `unstable` > `deprecated`.

kwalrath:
Yes, please. This also affects https://pub.dartlang.org/packages/dev_compiler. See dart-lang/sdk#29911 for details.

isoos:
Nit: What you are referring to is not the search ranking, it is the package index ordering.

I see a few conflicts with deprecated as it is proposed:
- version 2.0 could deprecate 1.0, but what if 2.1 un-deprecates everything?
- version 2.3 could deprecate 2.3, and it could be just as valid as not allowing it

I think at one point the package author needs to answer the question: what version do I want to be used by others? And if I want to use the newer, why don't I release it as stable?

If you deprecate everything that was `stable`, then what does it mean to be `unstable`?

isoos:
Another thing is that we could consider is to use `discontinued` instead of `deprecated`, because deprecation is the intent of the original developer, but others using it may consider it just fine a keep going and use it. It also simplifies the tracking, because a package is discontinued if and only if the last published version (stable or otherwise) has this flag set to true.

[...]

PR #967, opened 2018-02-07
Introducing the deprecated flag on the backend.

isoos:
#289.

- tool to get/set/clear the flag in the DataStore
- `search` skips adding it to the index when the flag is set
- `analyzer` skips analysis if the flag is set
- package tag and score is updated to handle the status

Disclaimer: I haven't tested it on staging yet.

### DIFF
--- a/app/bin/tools/deprecate.dart
+++ b/app/bin/tools/deprecate.dart
@@ -1,0 +1,83 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:appengine/appengine.dart';
+import 'package:gcloud/db.dart';
+
+import 'package:pub_dartlang_org/frontend/models.dart';
+import 'package:pub_dartlang_org/frontend/service_utils.dart';
+import 'package:pub_dartlang_org/shared/analyzer_client.dart';
+import 'package:pub_dartlang_org/shared/package_memcache.dart';
+
+void _printUsage() {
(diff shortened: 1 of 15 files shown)


app/bin/tools/deprecate.dart
```diff
@@ -0,0 +1,78 @@
...
+    }
+  });
+}
+
+Future _read(String packageName) async {
+  final Package p = (await dbService
+          .lookup([dbService.emptyKey.append(Package, id: packageName)]))
+      .single;
+  if (p == null) {
+    throw new Exception('Package $packageName does not exists.');
```
jakobr-google: s/exists/exist/

app/lib/search/updater.dart
```diff
@@ -55,7 +55,8 @@ class BatchIndexUpdater implements TaskRunner {
         final int count = _snapshot.documents.length;
         _logger
             .info('Got $count packages from snapshot at ${_snapshot.updated}');
-        await packageIndex.addPackages(_snapshot.documents.values);
+        await packageIndex.addPackages(
+            _snapshot.documents.values.where((pd) => pd.isDeprecated != true));
```
jakobr-google: *Maybe* add a little note here, so some helpful person doesn't accidentally change it to `!pd.isDeprecated`.

app/lib/frontend/templates.dart
```diff
@@ -224,7 +227,10 @@ class TemplateService {
 
     final Map<String, dynamic> data = {
       'package': package,
+      'show_depreacted': analysisStatus == AnalysisStatus.deprecated,
```
jakobr-google: s/depreacted/deprecated/ (here, and in the template)

isoos:
I'll test this tomorrow on staging before merging.

PR #969, opened 2018-02-09
Name tracker to reject package names that are too similar to another one.

### LINKED ISSUE #929: Prevent punctuation variations of package names
Prior art is NPM: https://twitter.com/jawache/status/953794118483922944

i.e. you could not publish `package:codebuilder` if there is already a `package:code_builder`.

isoos:
Closes #929

### DIFF
--- a/app/lib/shared/name_tracker.dart
+++ b/app/lib/shared/name_tracker.dart
@@ -1,0 +1,84 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:logging/logging.dart';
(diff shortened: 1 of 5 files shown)


app/lib/shared/name_tracker.dart
```diff
@@ -0,0 +1,82 @@
...
+    _reducedNames.add(_reduce(name));
+  }
+
+  bool hasPackage(String name) => _names.contains(name);
+
+  bool hasConflict(String name) => _reducedNames.contains(_reduce(name));
+
+  bool accept(String name) => hasPackage(name) || !hasConflict(name);
+
+  String _reduce(String name) => name.replaceAll('_', '').toLowerCase();
```
jakobr-google: We only remove `_` because we already don't allow other punctuation, right?

isoos: Correct, added a comment for it.

Issue #970, opened 2018-02-09
Update to analyzer v0.31.0


kevmoo:
Let's you remove this override

https://github.com/dart-lang/pub-dartlang-dart/blob/e93a423e893b6caddf7224b20f9120abb863b795/app/pubspec.yaml#L48

Blocked on https://github.com/dart-lang/pana/issues/246

PR #971, opened 2018-02-09
Upgrade package versions.

isoos:
Related to #970, but it does not closes it yet.

### DIFF
--- a/app/pubspec.yaml
+++ b/app/pubspec.yaml
@@ -10,7 +10,7 @@ dependencies:
   args: '>=0.13.7 <2.0.0'
   crypto: '^2.0.2'
   gcloud: '>=0.4.0 <0.5.0'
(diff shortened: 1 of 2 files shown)


app/pubspec.yaml
```diff
@@ -45,4 +45,4 @@ dev_dependencies:
 dependency_overrides:
   # Fixes https://github.com/dart-lang/pana/issues/138
   # Required until we have a 'gold' release
-  analyzer: 0.31.0-alpha.2
+  analyzer: 0.31.1
```
kevmoo: Can't you just remove the dependency_override entirely, @isoos ?

isoos: Only after we've upgraded with the appropriate pana version (it will be in 0.10.2).

kevmoo:
ack

PR #980, opened 2018-02-13
Dartdoc backend implementation on Bucket and files.

isoos:
The main idea here is that we can store and serve the generated files without going to Datastore, working only on files in a bucket.

- Each dartdoc run is associated with an uuid.
- `/$pkg/$version/content/$uuid/` will contain the generated files.
- `/$pkg/$version/in-progress/$uuid.json` indicates that an upload is in progress (or aborted)
- `/$pkg/$version/entry/$uuid.json` indicates a successful upload.
- The `../entry/` and the `.../in-progress/` are organized in a separate directories, in order to make their listing fast.

I'm planning to add memcache for the `entry/` listing, the uuids and the content serving too.

The alternative would be to introduce a similar hierarchy than we have in `PackageAnalysis`, `PackageVersionAnalysis` and `Analysis`, but thinking about it more, we could simplify that model using files like here. Both system could rely on the bucket to delete old entries automatically, sparing some worry about GC-ing orphaned entries.

### DIFF
--- a/app/build.yaml
+++ b/app/build.yaml
@@ -3,6 +3,7 @@
 targets:
   pub_dartlang_org:
     sources:
+      - 'lib/dartdoc/models.dart'
       - 'lib/shared/*.dart'
       - 'lib/search/backend*.dart'
     builders:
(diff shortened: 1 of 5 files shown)


kevmoo:
So excited!

PR #985, opened 2018-02-16
Fix deprecated handling: do not remove entirely from the index.

isoos:
I believe removing the package from the index was confusing, especially when the package name is rare, and I know it exists. Introducing three different handling inside the index:

- default ranking (displayed on landing page and platform listings) and text-based ranking will keep the deprecated package, but will list it at the end.
- popularity ranking will keep the package's original popularity and list it along other packages (total score will be a gray `--` though)
- updated and created ranking will keep the package as it was
- health and maintenance ranking will remove it from the results (as both the health and maintenance is considered too bad for displaying it there).

#289


isoos:
Btw. I've now tested this on staging, with the fix here #289 could be considered completed.

PR #990, opened 2018-02-19
Connecting dartdoc components

isoos:
Main changes:
- task sources to generate tasks based on the metadata in the bucket
- fixed path issue: bucket objectpath shouldn't start with `/`
- fixed requestUri parsing that started some relative paths with `/`
- added content-type for both the stored entry and on serving the file

Tested this locally, and seems to work, but I'm not satisfied with it:
- even with the concurrent uploads, uploading large packages takes long time
- with a large API surface, we'll generate an upload request per method, which will possibly ramp up storage costs.

In an upcoming PR I'll rewrite the backend to store a .tar.gz of the generated directory and serve content from a local filesystem cache, but I wanted to submit this as this is a working version.

### DIFF
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -10,7 +10,8 @@ import 'package:gcloud/storage.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as p;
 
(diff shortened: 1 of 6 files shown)


isoos:
Per our discussion: I've added a TODO to add the `dartdocVersion` prefix in a subsequent PR.

PR #999, opened 2018-02-20
Additional context versions for dartdoc entry.

isoos:
- Consolidated "should we update it" logic in `DartdocEntry.requiresNewRun()`
- Added `usesFlutter` and `flutterVersion` to trigger the re-run of packages when there is a flutter version upgrade (applies only on packages with flutter).
- Added `customizationVersion` to trigger the re-run of the packages when there is an update in our customization code.

### DIFF
--- a/app/lib/dartdoc/dartdoc_runner.dart
+++ b/app/lib/dartdoc/dartdoc_runner.dart
@@ -14,7 +14,7 @@ import 'package:uuid/uuid.dart';
 import '../shared/configuration.dart' show envConfig;
 import '../shared/task_scheduler.dart' show Task, TaskRunner;
 import '../shared/utils.dart' show redirectDartdocPages;
-import '../shared/versions.dart';
+import '../shared/versions.dart' as versions;
 
 import 'backend.dart';
 import 'customization.dart';
@@ -33,7 +33,7 @@ class DartdocRunner implements TaskRunner {
     if (redirectDartdocPages.containsKey(task.package)) {
       return true;
     }
-    final shouldRun = await dartdocBackend.shouldRunTask(task, dartdocVersion);
+    final shouldRun = await dartdocBackend.shouldRunTask(task);
     return !shouldRun;
   }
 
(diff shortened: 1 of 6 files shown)


app/lib/dartdoc/models.dart
```diff
@@ -59,6 +74,23 @@ class DartdocEntry extends Object with _$DartdocEntrySerializerMixin {
   String get contentPrefix => '$packageName/$packageVersion/content/$uuid';
 
   List<int> asBytes() => UTF8.encode(JSON.encode(this.toJson()));
+
+  bool requiresNewRun() {
+    if (isNewer(_semanticDartdocVersion, versions.semanticDartdocVersion)) {
+      return true;
+    } else if (dartdocVersion != versions.dartdocVersion) {
```
jakobr-google: This one took me a couple of seconds to understand, so maybe add a comment. I understood it as:
* If our dartdoc version is newer, regenerate (checked above)
* If dartdoc versions are different, *our version must be older* - so skip generating
* Otherwise, dartdoc versions are the same, so we only regenerate if flutter or customization is newer.

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -75,9 +78,14 @@ class DartdocRunner implements TaskRunner {
     return false; // no race detection
   }
 
-  Future _resolveDependencies(
-      PubEnvironment pubEnv, Task task, String pkgPath) async {
-    final pr = await pubEnv.runUpgrade(pkgPath, /* isFlutter */ false);
+  Future<bool> _usesFlutter(String pkgPath) async {
+    // TODO: detect if package is Flutter
```
jakobr-google: I assume you're tracking this one somewhere.

app/lib/dartdoc/models.dart
isoos: That is correct, adding docs.

app/lib/dartdoc/dartdoc_runner.dart
isoos: yeah, I'll do this right after this will be merged.

PR #1006, opened 2018-02-21
Deprecated -> Discontinued

isoos:
- renames fields, labels and template variables
- apply highly visible style
- removes discontinued items from the search index

### DIFF
--- a/app/lib/analyzer/model_properties.dart
+++ b/app/lib/analyzer/model_properties.dart
@@ -25,8 +25,8 @@ class AnalysisStatusProperty extends StringProperty {
           return 'failure';
         case AnalysisStatus.success:
           return 'success';
-        case AnalysisStatus.deprecated:
-          return 'deprecated';
+        case AnalysisStatus.discontinued:
+          return 'discontinued';
         case AnalysisStatus.outdated:
           return 'outdated';
         default:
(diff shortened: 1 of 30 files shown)


app/lib/analyzer/model_properties.dart
```diff
@@ -47,8 +47,8 @@ class AnalysisStatusProperty extends StringProperty {
           return AnalysisStatus.failure;
         case 'success':
           return AnalysisStatus.success;
-        case 'deprecated':
```
jakobr-google: It won't be a problem in prod, but do we have entries with status=='deprecated' stored in dev?

isoos: We have, but it is unlikely to cause problems.

jakobr-google: It'll cause an exception to be thrown here, right?

isoos: It will. But it will happen only if the most-recent analysis gets deleted, and only the older `'deprecated'` Analysis instance is available. The automatic removal of old Analysis instances will remove that soon and it won't cause issues.

jakobr-google: Right. No worries, then. :)

Issue #1011, opened 2018-02-21
Smart SEO for discontinued packages


kevmoo:
- [ ] Don't include them in the sitemap
- [ ] Consider excluding them robots.txt

mit-mit:
cc @isoos @jakobr-google WDYT?

isoos:
- Sitemap is easy (#1013 out for review).
- robots.txt is a bit tricky. If other pages link to it, it may still show up in search results. Instead, I'd rather use the `noindex` [meta header](https://support.google.com/webmasters/answer/93710?hl=en)

kevmoo:
`noindex` SGTM – we want robots.txt to be fast.

PR #1013, opened 2018-02-22
Remove discontinued packages from sitemap.

isoos:
#1011

### DIFF
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -78,14 +78,22 @@ class Backend {
   }
 
   /// Retrieves the names of all packages, ordered by name.
-  Stream<String> allPackageNames({DateTime updatedSince}) {
+  Stream<String> allPackageNames(
+      {DateTime updatedSince, bool excludeDiscontinued: false}) {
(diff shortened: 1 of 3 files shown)


app/lib/frontend/backend.dart
```diff
@@ -78,14 +78,22 @@ class Backend {
...
+  Stream<String> allPackageNames(
+      {DateTime updatedSince, bool excludeDiscontinued: false}) {
     final query = db.query(models.Package);
 
     if (updatedSince != null) {
       query.filter('updated >', updatedSince);
     }
 
-    return query.run().map((p) => (p as models.Package).name);
+    bool isExcluded(models.Package p) =>
```
kevmoo: rename this `shouldInclude`, flip the bool – then you can pass it as a lambda into `where` below.

isoos: I think flipping the bool makes `!excludeDiscontinued || p.isDiscontinued != true` to be harder to read, and passing it as a lambda is not worth changing it.

kevmoo:
Fair...

PR #1024, opened 2018-02-26
Use pool for continuous and concurrent upload and delete.

isoos:


app/lib/dartdoc/backend.dart
```diff
@@ -180,9 +177,15 @@ class DartdocBackend {
   Future _deleteAll(DartdocEntry entry) async {
     var page = await _storage.page(prefix: entry.contentPrefix);
     for (;;) {
+      final deletePool = new Pool(concurrentDeletes);
```
jakobr-google: We could use this pool for the entire delete, right? (Probably won't do much, though, since you're also awaiting the deleteFutures for each loop.)

isoos: Sure, we can use it, moved it to the outside. I'd still `await` the `deleteFutures`, but I don't think we'd lose much if we don't do that.

PR #1043, opened 2018-02-28
Create a tar archive of the documentation and upload it along the rest of the files.

isoos:


app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -179,4 +184,18 @@ class DartdocRunner implements TaskRunner {
     buffer.write('STDOUT:\n${pr.stdout}\n\n');
     buffer.write('STDERR:\n${pr.stderr}\n\n');
   }
+
+  Future _tar(Task task, String tmpDir, String tarDir, String outputDir,
+      StringBuffer logFileOutput) async {
+    logFileOutput.write('Running tar:\n');
+    final archive = 'archive.tar.gz';
```
jakobr-google: The old site has `package.tar.gz`, so if we're changing it, we'll definitely need to provide a link to the new name somewhere.

isoos: Good catch, fixing it.

PR #1050, opened 2018-03-02
Use memcache for FileInfo objects.

isoos:
This helps repeatedly serving the static content while browsing the documentation.


isoos:
(please hold off, I'll need to change this a bit)

isoos:
(fixed, PTAL)

Issue #1032, opened 2018-02-26
Sync to dartdoc 0.17.0 – SDK -dev.32


kevmoo:
...when it's published: https://github.com/dart-lang/dartdoc/pull/1616

the strong mode changes are awesome...

isoos:
Auxiliary: We now rely on the `dartdoc` provided by the SDK. At minimum we'll need to switch to depend on it as package and runt `pub run dartdoc`.

kevmoo:
In the short term, it'll be easier to just use what's in the SDK.

The update landed here: https://github.com/dart-lang/sdk/commit/7df5055b0679748e9800f73defb9ef3ff61ad3d9

Since we've already branched for -dev.31, this will almost certainly land in -dev.32

If ya'll end up looking at https://github.com/dart-lang/pub-dartlang-dart/issues/871 (which I think would rock) we'd have to look depending on the package directly – but that can wait.

kevmoo:
FYI @isoos – the dev release w/ these updates has landed

jakobr-google:
Fixed in #1062.

PR #1052, opened 2018-03-02
Upgrade SDK version in CI and docker

isoos:


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Keep version in-sync with .travis.yml
```
jakobr-google: Nit: Should we add 'and app/.travis.yml' ?

isoos: Not sure if `app/.travis.yml` is used for anything. I wanted to check it and remove, but never got around it.

jakobr-google: :)

kevmoo: FYI - https://github.com/dart-lang/mono_repo

But that tool has been updated to use a new kind of file.

I'll fix...

isoos:
> Or .32? #1032 

I don't feel comfortable switching the dartdoc version right before our release, without any time for testing it. The current version has been checked a few times, I'd give a bit more time for the next version.

kevmoo:
ack

Issue #1055, opened 2018-03-03
Add Package.discontinuedMessage


kevmoo:
Why: let authors (or site maintainers) provide "Use this instead..." messaging without pushing a new version.

Then map `isDiscontinued` => `discontinuedMessage != null`

If there is no message, just store empty string...or something.

https://github.com/dart-lang/pub-dartlang-dart/blob/a5a1c03ab71ed70ae1dd3e859de38009a9215748/app/lib/frontend/models.dart#L53

kevmoo:
CC @mit-mit

mit-mit:
For now the recommendation is to update pubspec.yaml

PR #1057, opened 2018-03-04
Store discontinued message instead of a flag.

isoos:
First part of #1055 - changes the model to store a message. We'll still need to figure where to put it and how to display it (e.g. is it a markdown-formatted text)?

### DIFF
--- a/app/bin/tools/discontinue.dart
+++ b/app/bin/tools/discontinue.dart
@@ -16,7 +16,7 @@ import 'package:pub_dartlang_org/shared/package_memcache.dart';
 void _printUsage() {
   print('Usage:');
   print('    ${Platform.script} read  <package>');
-  print('    ${Platform.script} set   <package>');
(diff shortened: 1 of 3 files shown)


app/bin/tools/discontinue.dart
```diff
@@ -67,7 +67,7 @@ Future _set(String packageName, bool value) {
     if (p == null) {
       throw new Exception('Package $packageName does not exist.');
     }
-    p.isDiscontinued = value;
+    p.discontinuedMessage = 'Package is discontinued.';
```
kevmoo: Perhaps a TODO here about allowing one to provide the message?

isoos: Fixed with taking the rest of the arguments as a message.

isoos:
@jakobr-google: shall we scrap this?

jakobr-google:
Let's. And then revisit it later with a proper self-service solution for package authors.

Issue #1073, opened 2018-03-07
Ability to see number of packages failing analysis


mit-mit:
I would like page that shows the number of packages failing analysis. I'm thinking at least a daily update, and the ability to see last seven days.

The goal is to enable us to catch systemic analysis issues, as as the current issues on:
https://pub.dartlang.org/packages/url_launcher#-readme-tab-

```
Fix dependencies in pubspec.yaml.

Running flutter packages pub upgrade failed with the following output:

ERR: Invalid argument(s): Minimum version ("2.0.0-dev.23.0") must be less than maximum ("2.0.0-dev.16.0").
ERR: package:pub_semver/src/version_range.dart 59         new VersionRan
```

isoos:
There are several reasons that old packages fail. If we are looking for alerting on mass-failures, I'd focus on two groups:
- packages that have new version published in the past X months (X = 6?)
- packages that are in the top N-th percentile wrt. popularity (N = 80?)

For the reporting part, I'm not yet sure if we can do this within the timeout limits of appengine (60 seconds to serve any request), but could start with a tool and refactor the code to a web page if it is fast enough.

jakobr-google:
I think we just need to tighten up our logging, so we can easily grep for this. We should then be able to have Stackdriver generate the reports.

isoos:
This may be lost in the noise of logging (e.g. we care much more of the analysis of the latest version than anything else, while the log may have much more failures). I'm putting together a small tool that will give us at least a minimal statistics.

jakobr-google:
![Why don't we have both](http://i0.kym-cdn.com/photos/images/original/000/538/731/0fc.gif)

Having a tool to query this is a good idea, especially if it can provide a bit more insight.

Specifically for this, as you said, we're mostly interested in failures for the latest version of a package, so if we add a log line like:

```
Analysis failed for latest version of $package ($version): $error
```

then we can create a metric in Stackdriver, get a report showing the trends, and be alerted if we cross a set threshold. This should've alerted us to the issue @mit-mit is referring to.

Metrics for total number of analysis failures would also give us a clue if things go south.

mit-mit:
Closing in favor of https://github.com/dart-lang/pub-dartlang-dart/issues/1122

PR #1074, opened 2018-03-07
Task stats for analyzer and dartdoc failures.

isoos:
#1073

### DIFF
--- a/app/bin/tools/task_stats.dart
+++ b/app/bin/tools/task_stats.dart
@@ -1,0 +1,128 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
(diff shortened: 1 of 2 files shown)


app/bin/tools/task_stats.dart
```diff
@@ -0,0 +1,127 @@
...
+  final Map report = {};
+
+  await withProdServices(() async {
+    registerAnalyzerMemcache(new AnalyzerMemcache(memcacheService));
+    registerDartdocMemcache(new DartdocMemcache(memcacheService));
+    registerAnalyzerClient(new AnalyzerClient());
+    registerDartdocClient(new DartdocClient());
+
+    final statFutures = <Future<_Stat>>[];
+    final twoYearsAgo = new DateTime.now().subtract(new Duration(days: age));
```
jakobr-google: `ageLimit`? It's not really two years now.

isoos: Good catch. Using `maxAgeDays` and `updatedAfter` instead.

Issue #1079, opened 2018-03-08
New Task allocation


isoos:
I'm planning to rewrite the task scheduler and allocation. Right now each service instance maintains a slightly randomized queue of tasks and hopes that there won't be too much race. We also track many different input versions separately (SDK, Flutter, pana...), and check them separately for a change.

Instead, we shall have a central queue of tasks, and "workers" could pull from these, with a simplified version to indicate the need for update.

The proposal is to create a Datastore entry: `Job`, with the following fields:
- `id`: composition of `service/packageName/packageVersion`
- `service`: e.g. `analyzer` or `dartdoc`
- `packageName`
- `packageVersion`
- `runtimeVersion`:
  - this would be a composite version that we increment each time one of the following changes: SDK, Flutter, pana, dartdoc, our customizations. 
  - it could be in the format of `year.month.day+patch`, e.g. `2018.03.08+1`
  - a "producer" may override this value only if it has a matching or newer version
  - a "worker" may process the job only if it has a matching or newer version
- `priority`:
  - 0 for freshly uploaded packages
  - 10 for latest stable versions
  - 100 + number of months passed for non-latest stable versions
- `lockedKey`: a random id to prevent parallel updates
- `lockedUntil`: a date until the job is locked by a worker

A "producer" would work like the `TaskSources`, but instead of emitting a `Task`, they would go to the Datastore, check if there is an entry, create it if missing, update it if allowed. 

A "consumer" would work much like the `TaskScheduler`, but instead of maintaining its internal queue of `Task`s, it would poll the Datastore for `Job`s, fetch the first 200 in priority order. It would select one preferring the beginning of the queue, and update its `lockedUntil` property with a reasonable default. After the processing is done the "worker" can delete the `Job` entry, assuming nothing changed (`lockedKey`).

/cc @jakobr-google WDYT?

jakobr-google:
I think having an actual task queue (that we can monitor) is a great idea, and this proposal should do the trick!

AppEngine does have a 'task queue' system (https://cloud.google.com/appengine/docs/standard/java/taskqueue) that we could potentially also use, but it doesn't look like it'd be easy to implement the prioritization there. Maybe with a pull queue, we could have different tags for the 3 buckets of tasks?

I can't tell from the docs how mature this system is, though, so rolling our own that fits our use-case exactly seems like a good idea, especially when it's this simple.

isoos:
[...]

PR #1080, opened 2018-03-08
Tracking runtimeVersion and using it in dartdoc.

isoos:
#1079
Analyzer requires a bit more work, because I'd like to add the `usesFlutter` flag in that model too.


app/lib/shared/versions.dart
```diff
@@ -6,6 +6,13 @@ import 'package:pub_semver/pub_semver.dart';
 
 import 'utils.dart' show isNewer;
 
+final String runtimeVersion = '2018.3.8';
```
jakobr-google: Maybe add a comment here, to update this whenever one of the other versions change?

app/lib/dartdoc/models.dart
```diff
@@ -84,30 +83,30 @@ class DartdocEntry extends Object with _$DartdocEntrySerializerMixin {
...
+    final age = new DateTime.now().difference(timestamp).abs();
+    if (age > entryUpdateThreshold) {
       return new TaskTargetStatus.ok();
     }
 
-    if (isNewer(
-        _semanticCustomizationVersion, versions.semanticCustomizationVersion)) {
-      // our customization version is newer -> regenerate
+    if (runtimeVersion == null) {
+      // legacy data -> regenerate
```
jakobr-google: Could we somehow skip the regeneration and just update the datastore entry, if hasContent and the versions are the same as current?

isoos: Good point, but instead of checking each version, I've opted for a hardcoded `?? '2018.3.8'` for now. PTAL.

Issue #948, opened 2018-01-31
SEO seems lacking


mit-mit:
I wonder if there is anything we can do to promote Pub pages more in searches?

For example, googling for 'flutter send sms' shows pub as the fourth result, way after github:
https://www.google.dk/search?q=flutter+send+sms

isoos:
Based on the Moz OSE, I think we have very few inbound links for that package, e.g. [url_launcher](https://moz.com/researchtools/ose/links?site=https%3A%2F%2Fpub.dartlang.org%2Fpackages%2Furl_launcher&filter=&source=external&target=page&group=0&page=1&sort=page_authority&anchor_id=&anchor_type=&anchor_text=&from_site=) vs [angular](https://moz.com/researchtools/ose/links?site=https%3A%2F%2Fpub.dartlang.org%2Fpackages%2Fangular&filter=&source=external&target=page&group=0&page=1&sort=page_authority&anchor_id=&anchor_type=&anchor_text=&from_site=) (note: I think there is a daily limit on how many URLs the OSE analyses).

Opportunities for improvements outside of pub site:

- GitHub repository can have one URL that points to the package, e.g. in the `pana` repo:
  <img width="789" alt="screen shot 2018-02-01 at 9 19 16" src="https://user-images.githubusercontent.com/4778111/35668241-06e12738-0731-11e8-922a-8241e99cb150.png">
  Unfortunately this is not great for flutter where the packages live in the single big repo.

- StackOverflow answers should be extended with links to the package page on pub.

- Our internal guides and tutorials should link to the package pages whenever it is natural to do so.

Opportunities for us:

- Increase the number of packages on the listing/search pages. This will reduce the distance from the index page for packages and hopefully increase their score.

- Add some attributes for the indexing pages that keeps the out of the index and passes their pagerank to the linked packages.

- With documentation on a separate domain, we can link back to the package, and provide one additional linkback from external domains.

kevmoo:
I've pushed through a number of changes to help with https://www.google.com/webmasters/tools/html-suggestions?hl=en&authuser=0&siteUrl=https://pub.dartlang.org/

(I think only @mit-mit will have access)

kevmoo:
This is annoying: `dart test package`

- test package homepage - solid
- Dart testing page – solid
- test_runner: very deprecated
- unittest: very deprecated
- matcher: related-ish
- dart_dev: related-ish

[...]

PR #1083, opened 2018-03-09
Update default page description.

isoos:
#948


app/lib/frontend/template_consts.dart
```diff
@@ -139,5 +141,9 @@ SortDict getSortDict(String sort) {
   );
 }
 
+final String defaultPageDescriptionEscaped = HTML_ESCAPE.convert(
+    'Pub is the packager for the Dart programming language, containing reusable '
```
jakobr-google: s/packager/package manager/ ? @mit-mit

mit-mit: Yes, please

PR #1091, opened 2018-03-12
Control dartdoc serving through runtimeVersion.

isoos:
#1079


app/test/shared/versions_test.dart
```diff
@@ -61,41 +61,16 @@ void main() {
...
-    test('max versions are serving', () {
-      expect(
-          shouldServeDartdoc(
-            dartdocVersion,
-            flutterVersion,
-            customizationVersion,
-          ),
-          isTrue);
+    test('current version is serving', () {
+      expect(shouldServeDartdoc(dartdocVersion), isTrue);
```
jakobr-google: This should be the current `runtimeVersion`, right?

isoos: yes, this is the current

jakobr-google: It's the current dartdoc version ('0.17.0'), but we're comparing it against the current `_dartdocServingRuntime` (2018.3.8), if I'm reading the code right.

isoos: ah, yeah, my mistake, fixing

PR #1114, opened 2018-03-15
New task/job queue processing.

isoos:
#1079

**Special deploy**: `index.yaml` changed, the following command needs to be run before deploying in prod:
```
CLOUDSDK_CORE_PROJECT=dartlang-pub gcloud app deploy index.yaml
```
The new method creates a `Job` for each `service` + `package` + `version`, and ensures that only the latest runtime version is being processed. It tracks `JobState`: `available` (for processing) -> `processing` -> `idle` (waiting for the specified time before considering new processing) -> `available`.

On the `https://dartdoc-dot-dartlang-pub-dev.appspot.com/debug` url one can check the current state for `stable` and `all` package versions, but we don't have report on the progress or on the ETA yet. (will add in a subsequent change)

@jakobr-google: let me know if you'd prefer to split this up a bit more, I can submit it in a few separate stages.


app/lib/job/backend.dart
```diff
@@ -0,0 +1,275 @@
...
+      });
+    }
+
+    final query = _db.query(Job)
+      ..filter('service =', service)
+      ..filter('state =', JobState.processing)
+      ..filter('lockedUntil <', new DateTime.now().toUtc());
+    await for (Job job in query.run()) {
+      try {
+        if (job.runtimeVersion != versions.runtimeVersion) continue;
```
jakobr-google: Why don't we want to unlock stale jobs from a different runtime version?

app/lib/job/backend.dart
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  Future complete(Job job, JobStatus status, {Duration extendDuration}) async {
+    return _db.withTransaction((tx) async {
+      final items = await tx.lookup([_db.emptyKey.append(Job, id: job.id)]);
+      final Job selected = items.single;
+      if (selected != null && selected.processingKey == job.processingKey) {
+        _logger.info('Updating $job with $status');
+        final errorCount =
+            status == JobStatus.success ? 0 : selected.errorCount + 1;
```
jakobr-google: Why do we keep jobs that succeeded in the job queue? If I'm reading this right, the Job db will end up containing an entry for each PackageVersion, and we'll keep running through them and extending their lock until we finally decide we need to process them.

Wouldn't it be better to only have jobs that we should process in the queue?

It also looks like we don't remove jobs from old runtime versions?

app/lib/job/job.dart
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  Future _updateFromTask(Task task) async {
+    final pKey = _db.emptyKey.append(Package, id: task.package);
+    final pvKey = pKey.append(PackageVersion, id: task.version);
+    final list = await _db.lookup([pKey, pvKey]);
+    final Package p = list[0];
+    final PackageVersion pv = list[1];
+
+    final isLatestStable = p == null || p.latestVersion == task.version;
```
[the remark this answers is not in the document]
isoos: I thought that here we are dealing with an eventually consistent request, but I've just realized that all lookup by key will be using strong consistency. Fixing it.

app/lib/job/job.dart
```diff
@@ -0,0 +1,153 @@
...
+        if (!packages.containsKey(pv.package)) {
+          final list = await _db.lookup([pv.packageKey]);
+          final Package p = list.single;
+          packages[pv.package] = p.latestVersion;
+        }
+        final bool isLatestStable = packages[pv.package] == pv.version;
+        final shouldProcess =
+            await _shouldProcess(pv.package, pv.version, pv.created);
+        await jobBackend.createOrUpdate(_service, pv.package, pv.version,
+            isLatestStable, pv.created, shouldProcess);
```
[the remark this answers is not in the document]
isoos: That was on the eventual consistency assumption, removing from `_updateFromTask`

app/lib/job/backend.dart
isoos: removing...

app/lib/job/backend.dart
isoos: > It also looks like we don't remove jobs from old runtime versions?

When a new service with a newer runtime comes alive, it will rewrite the `Job` entries with the new `runtimeVersion` using `jobMaintenance.syncDatastoreHistory`.

> Wouldn't it be better to only have jobs that we should process in the queue?

I've started with that. The current version gives the benefit of having the last status there, which in theory could be useful, although I can't really point out any immediate use of it. Another potentially useful thing would be to have an "audit trace" in the Datastore, e.g. `Job` would be a parent entity, while a `JobExecution` could contain the main logs and execution time of the given run. Neither is really important or useful for us, shall I remove the entries?

isoos:
Another item to consider here: `Job` could be used to periodically generate `sitemap.txt` or other aggregation-related / non-package-related tasks. It requires a slightly different `id` generation and maybe another field, but otherwise would be robust to use it.

app/lib/job/backend.dart
isoos: One thing that is benefit here: if the successfully generated documentation sets the `lockedUntil` for 90 days, none of the instances need to check whether it needs a new run until it times out. With deleting the entry, it will be checked at least once a day in each running instance.

jakobr-google: True, let's keep it. I'd missed the update of `runtimeVersion`, so we should be good.

So this is not really a queue, but more of an extension to the PackageVersion table, holding the job statuses for the version. Cool!

isoos: > So this is not really a queue, but more of an extension to the PackageVersion table, holding the job statuses for the version.

Yeah, that is kind of the role. I am not sure whether the name describes the intent properly. Would you use a different naming for it?

jakobr-google: `Job` is fine, since it records the jobs and when we need to look at them again. I just had queues in my head.

PR #1134, opened 2018-03-21
Tune job locking period

isoos:
- increased job locked time as error count increases
- don't force retry failed jobs on each startup


app/lib/job/backend.dart
```diff
@@ -280,4 +280,11 @@ class JobBackend {
       },
     };
   }
+
+  DateTime _lockTimestamp(int errorCount, Duration lockDuration) {
```
kevmoo: use named param if it's optional

isoos:
Updated with further tuning, PTAL. Deploying in staging for further monitoring.

PR #1142, opened 2018-03-23
Fix bad git merge.

isoos:


isoos:
This is a first for me: I haven't seen a bad merge like this one.

Issue #1101, opened 2018-03-14
Catch all unhandled exceptions to prevent the entire service going down


isoos:
Follow-up to https://github.com/dart-lang/pub-dartlang-dart/issues/1098
- [x] frontend
- [x] analyzer
- [x] search
- [x] dartdoc

isoos:
With #1148, the following parts are inside isolates that are automatically restarted on uncaught exceptions (after a 1-minute delay):
- `analyzer`'s job processing
- `search` (both the indexing and http serving)
- `dartdoc`'s job processing

1. Shall we put all the other http serving parts (`frontend`, `analyzer`, `dartdoc`) in a separate, restartable isolate?
2. How much delay should we use for http-serving isolate restarts (if at all)?

I think restartable isolates could provide a good catch-all handler, and if there is too many of them (e.g. 3+ in a minute), we could just abort/exit the DartVM and let AppEngine do its thing. WDYT @jakobr-google ?

jakobr-google:
Running the http handlers in a separate isolate from the job runners sgtm. I don't think I'd put a delay on restarting the http handler isolate, at least not the first time - but maybe limit it to 1 restart per 10-20 seconds? And then abort completely on too many failures.

isoos:
I was checking `package:isolate` if it can help us, but I don't think it is a good match for our use-case. Pinging @kevmoo as he contributed to that package to confirm that it is more for running short-lived tasks through a "main" controller than our "managed process" case.

isoos:
Reopening, because local development didn't show frontend isolate error, and it was confusing what is happening.

PR #1148, opened 2018-03-26
Use the shared isolate handling in search service.

isoos:
#1101

### DIFF
--- a/app/bin/service/search.dart
+++ b/app/bin/service/search.dart
@@ -29,26 +29,18 @@ import 'package:pub_dartlang_org/search/updater.dart';
 final Logger _logger = new Logger('pub.search');
 
 Future main() async {
-  final ReceivePort errorPort = new ReceivePort();
-  errorPort.listen((error) {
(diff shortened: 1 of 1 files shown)


app/bin/service/search.dart
```diff
@@ -29,27 +29,19 @@ import 'package:pub_dartlang_org/search/updater.dart';
...
 
 // Remove after https://github.com/dart-lang/sdk/issues/30755 lands.
 void _mainWrapper(message) => _main(message);
 
-void _main(int isolateId) {
+void _main(List<SendPort> sendPorts) {
   useLoggingPackageAdaptor();
 
+  final SendPort mainSendPort = sendPorts[0];
+  final ReceivePort taskReceivePort = new ReceivePort();
```
jakobr-google: We're creating another `taskReceivePort` in the code below. Should it use this port instead?

isoos: good catch, fixed. (although in search case it is not used at the moment)

PR #1153, opened 2018-03-27
Cleanup after worker isolate exited.

isoos:
#1101

- Closing ports after an isolate exists.
- Unregistering related sendPort for task delegation.
- Handle more than one task sendPort (when multiple workers are started).

### DIFF
--- a/app/lib/shared/task_client.dart
+++ b/app/lib/shared/task_client.dart
@@ -6,28 +6,41 @@ import 'dart:isolate';
 
 import 'task_scheduler.dart' show Task;
 
-SendPort _taskSendPort;
+final _taskSendPorts = <SendPort>[];
+int _taskSendIndex = 0;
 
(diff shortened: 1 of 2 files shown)


app/lib/shared/task_client.dart
```diff
@@ -6,28 +6,41 @@ import 'dart:isolate';
...
+int _taskSendIndex = 0;
 
 // Temporary queue until the isolate comes online.
-List<Task> _taskQueue = [];
+final List<Task> _taskQueue = [];
 
 void registerTaskSendPort(SendPort taskSendPort) {
-  _taskSendPort = taskSendPort;
+  _taskSendPorts.add(taskSendPort);
   if (_taskQueue != null) {
```
jakobr-google: `if (_taskQueue.isNotEmpty)` ? It's never `null` now.

Issue #1088, opened 2018-03-09
Use highlight.js and github.css from pkg:dartdoc


kevmoo:
This is on the way to https://github.com/dart-lang/pub-dartlang-dart/issues/871

Once https://github.com/dart-lang/pub-dartlang-dart/pull/1070 lands, you have access to https://github.com/dart-lang/dartdoc/tree/master/lib/resources

Could look to remove https://github.com/dart-lang/pub-dartlang-dart/tree/master/static/highlight – and just use the content in dartdoc

isoos:
#1070 don't include it as pubspec dependency, it'll be only a `global activate` util. nevertheless, we can and shall adopt the same styles.

isoos:
I've checked, the files in https://github.com/dart-lang/pub-dartlang-dart/tree/master/static/highlight are the same as in dartdoc. Nothing to do this time?

kevmoo:
I was hoping for a package dependency – so it just "happens" – but no biggy.

Figuring out how to align the custom CSS would also be great

isoos:
On thing that I've noticed with the current markdown table issue is that dartdoc uses a lot of styling in their `styles.css` that we may also use, but can't directly, because the rest of the styles are for a full dartdoc page, and not for our package page.

Not sure how much effort we should put in it, but separating some of these reusable styles could work for us.

kevmoo:
Thoughts, @jcollins-g ?

jcollins-g:
No objections to reorganizing our CSS in a way that works better for you.  If you do so though, please check that Flutter in particular still works as they also modify the style.

isoos:
As far as I can tell (checking docs.flutter.io) flutter does not override the `static-assets/styles.css`, it just adds extra `<style>` blocks that does the override, e.g. here:
https://github.com/flutter/flutter/blob/master/dev/tools/dartdoc.dart#L124

My 'split' suggestion would preserve the same semantics that `styles.css` has, but it would contain the rules in two files instead of one, and the dartdoc template would include both.

isoos:
This was done (~ a year ago?).

PR #1157, opened 2018-03-28
Test whether dartdoc assets here match the dartdoc repository.

isoos:
#1088


isoos:
This is a presubmit-only test, which I hope will be run only on travis? (@kevmoo is that true?)

kevmoo:
I believe so - https://github.com/dart-lang/pub-dartlang-dart/blob/master/app/dart_test.yaml

PR #1160, opened 2018-04-03
Move frontend handler wrappers to shared.

isoos:
- URL sanitizer is only for frontend, because documentation handles it differently
- removed the extra `try {} catch (e, st) {}` block around the logger wrapping
- user auth parsing seems to be something we don't use for dartdoc or other services, but won't hurt either

### DIFF
--- a/app/lib/shared/handler_helpers.dart
+++ b/app/lib/shared/handler_helpers.dart
@@ -7,10 +7,20 @@ import 'dart:io';
 
 import 'package:appengine/appengine.dart';
 import 'package:logging/logging.dart';
+import 'package:path/path.dart' as path;
 import 'package:shelf/shelf.dart' as shelf;
 import 'package:shelf/shelf_io.dart' as shelf_io;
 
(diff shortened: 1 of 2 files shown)


app/lib/shared/handler_helpers.dart
```diff
@@ -7,11 +7,21 @@ import 'dart:io';
...
 import 'package:shelf/shelf_io.dart' as shelf_io;
 
-Future runHandler(Logger logger, shelf.Handler handler, {bool shared: false}) {
+import '../frontend/service_utils.dart';
+
+import 'utils.dart' show fileAnIssueContent;
+
+Future runHandler(Logger logger, shelf.Handler handler,
+    {bool sanitize: false, bool shared: false}) {
   handler = redirectToHttpsWrapper(handler);
```
jakobr-google: Are these applied in the right order? Ie. should we redirect to https before parsing auth/sanitizing the request?

isoos: I don't see issues either way, reordered.

Issue #1156, opened 2018-03-28
`+` need not be encoded in URIs (unless it's in the query)


kevmoo:
See
https://pub.dartlang.org/packages/angular/versions/5.0.0-alpha%2B8
vs
https://pub.dartlang.org/packages/angular/versions/5.0.0-alpha+8

and
https://pub.dartlang.org/documentation/angular/5.0.0-alpha%2B8/
vs
https://pub.dartlang.org/documentation/angular/5.0.0-alpha+8/

Reference: https://stackoverflow.com/a/1006074/39827

`+` is used in a lot of package versions – it'd be great if these URLs were prettier!

isoos:
This should be coming along with https://github.com/dart-lang/pub-dartlang-dart/blob/master/app/lib/shared/urls.dart getting done.

PR #1172, opened 2018-04-05
Move more url methods to shared/urls.dart

isoos:
#1156

### DIFF
--- a/app/lib/shared/urls.dart
+++ b/app/lib/shared/urls.dart
@@ -4,8 +4,9 @@
 
 import 'package:path/path.dart' as p;
 
-import 'utils.dart' show siteRoot;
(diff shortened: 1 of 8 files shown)


app/lib/shared/urls.dart
```diff
@@ -19,12 +20,16 @@ String niceUrl(String url) {
   return url;
 }
 
-String pkgPageUrl(String package, {String version, bool includeHost: false}) {
+String pkgPageUrl(String package,
+    {String version, bool includeHost: false, String hash}) {
```
jakobr-google: Nit: 'fragment' instead of 'hash'? Hash sounds like it's a hash of something.

isoos: Good point, fixed.

PR #1173, opened 2018-04-05
Restructure notification URLs

isoos:
This is auxiliary to the URL refactoring, and makes notification API separate from the usual package-data serving paths. `POST /packages/[pkg]/[version]` becomes: `POST /api/notification` with a JSON payload.

On deploy there is a short period while the old version wants to notify the new or the new version wants to notify the old API. We'll have some entries in the logs wrt. unsuccessful notification, but otherwise there is no break in functionality, as datasource polling should pick up the same task in 10-15 minutes at most.

### DIFF
--- a/app/lib/analyzer/handlers.dart
+++ b/app/lib/analyzer/handlers.dart
@@ -4,22 +4,19 @@
 
 import 'dart:async';
 
-import 'package:logging/logging.dart';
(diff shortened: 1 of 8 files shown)


isoos:
> should fix the failing test, though.

Fixed.

PR #1180, opened 2018-04-09
No encoding for + in version urls

isoos:
#1156


isoos:
I'll add `+1` versions for tests in a follow-up PR

PR #1197, opened 2018-04-12
Fixing dartdoc canonical urls

### LINKED ISSUE #1017: SEO for dartdoc - consider canonical pages, etc
https://github.com/dart-lang/pub-dartlang-dart/pull/1016 made me think of this

Search engines like a clear notion of "the best" page where there is a sea of very similar pages.

When docs are indexed, I'm worried that the best page for `pkg` will change from  `/documentation/pkg/1.0.0` to `/documentation/pkg/1.0.1` to ``/documentation/pkg/1.02`

I think doc URLs should be like package URLs

`https://pub.dartlang.org/packages/pana` does not redirect to `https://pub.dartlang.org/packages/pana/versions/0.10.2`

Instead `https://pub.dartlang.org/packages/pana/versions/0.10.2` has a cononical tag that points search engines back to the "main" page

```html
<link rel="canonical" href="https://pub.dartlang.org/packages/pana"/>
```

I think we should do the same w/ docs.

Thoughts?

### LINKED ISSUE #1182: Canonical URL for "latest" page wrong?
https://pub.dartlang.org/documentation/angular/latest/

Has a canonical URL of `https://pub.dartlang.org/documentation/angular/4.0.0+2//index.html`

1 - Extra slash is sad
2 - Doesn't feel like it aligns w/ discussion here https://github.com/dart-lang/pub-dartlang-dart/issues/1017#issuecomment-378746468

We want the canonical URL to be stable – https://pub.dartlang.org/documentation/angular/latest/ – so search engines aren't constantly linking to different pages as the content updates

isoos:
- Closes #1182 by removing extra slash at the end.
- Closes #1017 by using `/latest/` as the version for latest stable releases.


app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -158,6 +158,12 @@ class DartdocJobProcessor extends JobProcessor {
     StringBuffer logFileOutput,
   ) async {
     logFileOutput.write('Running dartdoc:\n');
+    final canonicalVersion = job.isLatestStable ? 'latest' : job.packageVersion;
+    String canonicalUrl = pkgDocUrl(job.packageName,
```
kevmoo: Wouldn't you want to fix `pkgDocUrl` here?

Or maybe use path.join?

isoos: Added `omitTrailingSlash` to `pkgDocUrl`. Unfortunately, due to the dartdoc output's use of base href, we need to have a trailing slash when serving the content. We need to either add there (template output, /latest/ redirects) or remove here.

Issue #1018, opened 2018-02-23
Have an 'audit log' for each package


kevmoo:
Something like https://help.github.com/articles/reviewing-the-audit-log-for-your-organization/

- Version X uploaded by y@example.com
- Uploader foo@ was added by bar@
- Uploader foo@ was removed by bar@
- Analysis succeeded/failed for vX.Y – link to log
- Doc generation succeeded/failed for vX.Y - link to log
- etc

Would be great for a package maintainer

Since all of the emails are public anyway, I'm not sure we'd need to put this behind a login – should discuss.

Related to https://github.com/dart-lang/pub-dartlang-dart/issues/993

isoos:
I think we should separate two major groups here:
- Package-level messages, which are usually initiated by the user (e.g. upload, uploader maintenance)
- PackageVersion-level messages, which are usually coming from our services

The PackageVersion-level ones will be more frequent and larger in numbers, it may be reasonable to separate their display (and maybe their database representation too).

The straightforward use case is relatively simple: 1 event -> 1 database record, which we can list in reverse chronological order. However we have multiple edge cases:
- A package may be analyzed multiple times a day (e.g. several of its dependencies get an update, and we finish its analysis before the next update happens).
- A failed analysis may be retried multiple times over a week, and it may fail each time.

In such cases I think it would be better to aggregate multiple events in a single record, e.g. "We have tried to analyze `pkg` `v0.1.0` 4 times between `May 23, 2018 and May 30, 2018`, but failed".

Alternatives:
- do such aggregation at the time we are displaying the audit log (seems harder when if/when we need to do pagination)
- do such aggregation into of a separate table that gets displayed on the UI

WDYT?

kevmoo:
I like the idea of splitting out the user actions and the auto actions
I get the point about too many actions happening at once
Aggregation sounds nice, but it seems like something we should look at later.

Are we talking about dozens of possible events for one package in a day?

isoos:
> Are we talking about dozens of possible events for one package in a day?

I'd guesstimate it to be 1 event for each package-version per day. If we are displaying "auto" events separately for each version, it shouldn't be a problem.

We could start with immutable user-initiated events, and maybe-mutable auto-events. The aggregation can be added later.

isoos:
Right now (in staging) we have `History` entries enabled, and storing:
- uploader invites
- uploader changes (add/remove)
- package uploads

[...]

PR #1204, opened 2018-04-13
More history preparation

isoos:
#1018

- initializing backend for services,
- type-based serialization and lookup,
- example event with formatting.

### DIFF
--- a/app/lib/history/models.dart
+++ b/app/lib/history/models.dart
@@ -5,9 +5,12 @@
 import 'dart:convert';
 
 import 'package:gcloud/db.dart' as db;
+import 'package:json_annotation/json_annotation.dart';
+
+part 'models.g.dart';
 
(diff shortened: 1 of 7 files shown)


app/lib/history/models.dart
```diff
@@ -46,4 +52,61 @@ class History extends db.ExpandoModel {
...
+  String getType();
+
+  Map<String, dynamic> toJson();
+
+  String formatMarkdown(HistoryData data);
+}
+
+typedef HistoryEvent HistoryEventFromJson(Map<String, dynamic> json);
+
+final _events = <String, HistoryEventFromJson>{
```
jakobr-google: `_eventDeserializers`? I originally thought this held the actual events, which didn't really make sense to me. :)

isoos: good point, fixed

PR #1206, opened 2018-04-16
Move search url handling to shared.

isoos:
It also improves #1156 in the side block's dependency search link.

### DIFF
--- a/app/lib/shared/handlers.dart
+++ b/app/lib/shared/handlers.dart
@@ -24,7 +24,7 @@ const staticLongCache = const Duration(hours: 24);
 shelf.Response redirectResponse(String url) => new shelf.Response.seeOther(url);
 
 shelf.Response redirectToSearch(String query) {
-  final uri = new Uri(path: '/packages', queryParameters: {'q': query});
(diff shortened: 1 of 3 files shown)


app/lib/shared/handlers.dart
```diff
@@ -24,7 +24,7 @@ const staticLongCache = const Duration(hours: 24);
 shelf.Response redirectResponse(String url) => new shelf.Response.seeOther(url);
 
 shelf.Response redirectToSearch(String query) {
-  final uri = new Uri(path: '/packages', queryParameters: {'q': query});
+  final uri = searchUrl(q: query);
   return redirectResponse(uri.toString());
```
jakobr-google: `uri` is now a `String`, so we don't really need to call `toString()` on it. Doesn't hurt, though.

isoos: Starting which SDK? I don't see it with dev-46.

jakobr-google: Sorry, I meant the `uri` variable. `searchUrl()` returns a `String`.

Issue #1179, opened 2018-04-09
Extend search to also cover dartdoc contents


mit-mit:
Likely just starting classes and members

isoos:
Related update: https://github.com/dart-lang/pana/issues/269#issuecomment-393098611

PR #1215, opened 2018-04-17
Index and search public API symbols (when provided).

isoos:
First part of #1179, if we are going for the simpler search result page.


isoos:
I'm having second thoughts about this design. It may be better to go with complex version from the start (where we preserve some path information, and later on we may be able to display it as result).

Issue #1213, opened 2018-04-16
Make installing tab "smarter" about packages w/ binaries


kevmoo:
URL: https://pub.dartlang.org/packages/webdev#-installing-tab-

`webdev` has no libraries to use. It does have an executable defined in `pubspec.yaml`, though.

Consider adding `pub global activate` instructions for such packages.

See also
https://pub.dartlang.org/packages/peanut
https://pub.dartlang.org/packages/pubviz
https://pub.dartlang.org/packages/stagehand

isoos:
This seemed to be a low-hanging fruit, but it turns out it has a lot of edge-cases, with the following parameters:

- with/without libraries
- with/without executables
- runs on Flutter/runs only on Flutter

How would you display the install tab when there are libraries and executables? What if it is Flutter-only?

kevmoo:
> How would you display the install tab when there are libraries and executables? What if it is Flutter-only?

🤷‍♂️ 

Maybe just show both? (When we talk about executables, we mean packages with an `executables:` section in their pubspec, right?

isoos:
> Maybe just show both?

Yeah, but what should be the wording? We already have a section of "depend on it", and also "install it", while global activate would be another installing, sometimes with library use, sometimes without it.

> When we talk about executables, we mean packages with an executables: section in their pubspec, right?

Yes.

jakobr-google:
Maybe just show both, using sub-headings:

'Use this package as a library'

'Use this package as an executable'

PR #1238, opened 2018-04-25
 Install tab with global activate for binary-only packages.

isoos:
One step closer to #1213:
- The first commit is a refactor of the template for simpler structure and narrow variable names.
- The second introduces a new template part for packages that are binary-only.

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -447,6 +447,11 @@ class TemplateService {
       }).toList();
     }
 
+    final executables = selectedVersion.pubspec.executables?.keys?.toList();
+    executables?.sort();
+    final hasExecutables = executables != null && executables.isNotEmpty;
+    final onlyGlobalInstall = hasExecutables && importExamples.isEmpty;
+
     final exampleVersionConstraint = '"^${selectedVersion.version}"';
 
     final bool usePubGet = !isFlutterPackage ||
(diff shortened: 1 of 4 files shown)


app/lib/frontend/templates.dart
```diff
@@ -447,38 +447,43 @@ class TemplateService {
...
 
-    String toolHtml;
-    if (renderGeneric && renderFlutter) {
-      toolHtml = '<code>pub get</code> or <code>packages get</code>';
-    } else if (renderFlutter) {
-      toolHtml = '<code>packages get</code>';
+    String editorSupportedToolHtml;
+    if (usePubGet && useFlutterPackagesGet) {
+      editorSupportedToolHtml =
+          '<code>pub get</code> or <code>packages get</code>';
```
jakobr-google: Should the last one be `<code>flutter packages get</code>` ?

app/lib/frontend/templates.dart
```diff
@@ -447,38 +447,43 @@ class TemplateService {
...
+    } else if (useFlutterPackagesGet) {
+      editorSupportedToolHtml = '<code>packages get</code>';
```
jakobr-google: Same here.

app/lib/frontend/templates.dart
isoos: yeah, it should, fixing...

PR #1245, opened 2018-04-27
Log History entry when a PackageVersion is uploaded.

isoos:
Part of #1018

- Disabled by default (enabled through the `HISTORY_ENABLED=1` environment variable).
- Started to write a backfill script.

Issues to resolve (in a subsequent work):
- The "log after upload" and the backfill is not using the same code, but they should.
- The object creation and use is a bit strange, should be simpler.
- Not yet sure how the scopes will be handled.

### DIFF
--- a/app/bin/tools/backfill_history.dart
+++ b/app/bin/tools/backfill_history.dart
@@ -1,0 +1,62 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:appengine/appengine.dart';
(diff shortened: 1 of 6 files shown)


app/bin/tools/backfill_history.dart
```diff
@@ -0,0 +1,60 @@
...
+      await for (Package p in dbService.query(Package).run()) {
+        await _backfillPackage(p.name);
+      }
+    }
+  });
+}
+
+Future _backfillPackage(String package) async {
+  print('Backfill history in: $package');
+  await for (PackageVersion pv in dbService.query(PackageVersion).run()) {
```
jakobr-google: Shouldn't we limit this query to the package we want?

isoos: oops... good catch! fixed.

Issue #1249, opened 2018-05-02
Update icon


mit-mit:
Dart has a new icon, so we should update the Pub site. If possible, we should also tweak the color scheme to match the new colors.

anders-sandholm:
Looks like this has been fixed, landed, and deployed. 
Woohoo.
Thanks @isoos and @jakobr-google

mit-mit:
Re-opening: there is still an issue with font sizes we are tracking down, so this is not entirely fixed

PR #1261, opened 2018-05-04
Update favicon and icon reference with new Dart logo.

isoos:
#1249


isoos:
I'm re-doing this with the new, larger icon's I've got on the weekend.

Issue #1248, opened 2018-04-30
pkg:english_words – support excluding some packages from popularity


kevmoo:
URL: https://pub.dartlang.org/packages/english_words

@filiph did a great job, but I'm pretty sure this is popular because there's a getting started app that uses it.

We should save space on our landing page for packages more likely to be generally useful...

filiph:
![rxdlq](https://user-images.githubusercontent.com/919717/39447306-315134ec-4c76-11e8-9460-a101367eeb69.gif)

isoos:
> sure this is popular because there's a getting started app that uses it.

Why would we pick on the reasons why something is popular? It is either popular or not popular, it got its fair score.

Now, if you want to have a curated list of packages on the landing page, that is an entirely different request, but I wouldn't dilute the meaning of popularity...

kevmoo:
Nothing but love for the work from @filiph – I feel it's okay for us to put a handicap on packages "we" own.

isoos:
The package in question is not a full-blown NLP package, but it doesn't need to be one to provide a value on its own, outside of the tutorial's scope. I think it not fair to penalize it on the ground that @filiph is a Googler.

filiph:
I'm with Kevin here. I bet the package gets 99% of usage from newcomers using the initial flutter codelab. 

Maybe just prevent it from showing in the top packages. I think that list should be, conceptually, "packages others use because they start a new non-beginner project and add them manually".

In the perfect world, we would have the ability to ignore those codelab/firstcomer downloads - not just for english_words, but for other packages as well (angular_components, angular itself, etc.), which would give us much more realistic data. But since that's unlikely to ever happen, I'm fine with a little bit of weeding.

isoos:
The quick fix was deployed, moving the issue to M7, to track the introduction of "do-not-advertise-me" flag.

isoos:
@jakobr-google: could you please set the flag for the package?

PR #1264, opened 2018-05-04
Introducing the do-not-advertise flag.

isoos:
Part of #1248, will be set with the refactored script from https://github.com/dart-lang/pub-dartlang-dart/pull/1260


isoos:
Updated the flag.dart script too.

Issue #1210, opened 2018-04-16
Cleanup orphaned records


isoos:
It just occurred to me that after deleting a package, we have a few records that are orphaned:
- [ ] `PackageAnalysis`, `PackageVersionAnalysis`, `Analysis`
- [ ] `Job` entries
- [ ] `History` entries (soon)
- storage on `dartdoc`'s bucket

jakobr-google:
Yep. I've modified the remove_package tool to clean up the analysis-related records. I'll clean that up and send for review tomorrow.

I removed the Job entries and dartdoc storage manually for the two packages I've removed so far.

PR #1265, opened 2018-05-04
Delete Job entries on package delete

isoos:
#1210

### DIFF
--- a/app/bin/tools/remove_package.dart
+++ b/app/bin/tools/remove_package.dart
@@ -7,14 +7,15 @@ import 'dart:io';
 
 import 'package:gcloud/db.dart';
 import 'package:gcloud/storage.dart';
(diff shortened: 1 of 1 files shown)


app/bin/tools/remove_package.dart
```diff
@@ -120,6 +119,10 @@ Future removePackage(String packageName) async {
 
     print('Package "$packageName" got successfully removed.');
     print('WARNING: Please remember to clear the AppEngine memcache!');
+
+    print('Committing changes to DB ...');
+    T.queueMutations(deletes: deletes);
+    await T.commit();
```
jakobr-google: Shouldn't we do this before we claim the package has been successfully removed?

isoos: Good point, moving the `print` out after the transaction commits.

PR #1270, opened 2018-05-07
API docs link update: no domain, handle latest

isoos:
- remove domain part from the URL, to make it easier to test it on staging
- use `/latest/` as a version if the package version is the latest stable on package page (both the sidebar and the versions tab)


isoos:
On a second thought I've updated this to change the URL only in the sidebar, because it makes more sense to link to the version-specific documentation on the versions tab/versions page

PR #1275, opened 2018-05-09
Remove History entries on package removal

isoos:
#1210 

I'm a bit worried if the `deletes` list gets too large for the transaction, removing some entries in separate transactions. This could also go in another direction: creating the single `deletes` list, slice it in 100-key chunks, and execute the delete transaction per chunk.

### DIFF
--- a/app/bin/tools/remove_package.dart
+++ b/app/bin/tools/remove_package.dart
@@ -70,21 +70,20 @@ Future listPackage(String packageName) async {
   }
 }
 
-Future removePackage(String packageName) async {
+Future _deleteWithQuery(Query query) async {
   final deletes = <Key>[];
-
-  final jobQuery = dbService.query(Job)..filter('packageName =', packageName);
-  await for (Job job in jobQuery.run()) {
-    deletes.add(job.key);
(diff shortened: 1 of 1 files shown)


app/bin/tools/remove_package.dart
```diff
@@ -69,13 +70,25 @@ Future listPackage(String packageName) async {
...
   }
+  await dbService.withTransaction((tx) async {
+    tx.queueMutations(deletes: deletes);
+    await tx.commit();
+  });
+}
+
+Future removePackage(String packageName) async {
+  await _deleteWithQuery(
+      dbService.query(Job)..filter('packageName =', packageName));
```
jakobr-google: Should we do this after deleting the Package? There's a tiny window here where our scheduling job might pick up the package and add new Jobs for it, after we've deleted the Jobs, but before we delete the Package/PackageVersion.

isoos: Moved both to the end of the method. I believe there is always a window that could introduce artifacts, e.g. when there is an ongoing analysis and we delete the records while it is still doing its stuff. The script can be run multiple times though, cleaning these up later.

PR #1284, opened 2018-05-11
Remove Stream randomizer

isoos:
No longer needed (`Job` processing).


isoos:
Putting this on hold, as the lack of randomization would increase the number of aborted Job update transactions during a new release.

Issue #1281, opened 2018-05-10
Updated packages seems behind (3 days?)


kevmoo:
Taken at ~5pm on Thursday.

![screen shot 2018-05-10 at 4 58 42 pm](https://user-images.githubusercontent.com/17034/39899936-72143402-5473-11e8-9be9-681b589b65d0.png)

csslib has been published twice today – but it's not showing up 😢

isoos:
@kevmoo @jakobr-google: a reboot of the instances would mitigate the immediate problem again, but I'd also like to have that extra logging in a deployed version. Would it be possible?

isoos:
I think the regression was caused by the new code that is now partially disabled: we do fetch the dartdoc-generated `index.json` from the `dartdoc` service, but don't use it in the search index. I think the symptoms in the log are consistent with a hanging `Future` where the `HTTP GET ... index.json` did not complete. https://github.com/dart-lang/pub-dartlang-dart/pull/1288 to fix that with a timeout.

isoos:
#1288 seems to have fixed this. Package updates are now picked up in ~10 minutes or so.

PR #1288, opened 2018-05-14
Add timeout for fetching dartdoc-generated content.

isoos:
This seems to be a likely candidate that (a) explains the regression of #1281 and (b) adds a fix for it.

### DIFF
--- a/app/lib/search/updater.dart
+++ b/app/lib/search/updater.dart
@@ -106,6 +106,9 @@ class BatchIndexUpdater implements TaskRunner {
     }
     if (_batch.isEmpty) return;
 
+    final Timer timer = new Timer(const Duration(minutes: 5), () {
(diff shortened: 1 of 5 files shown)


app/lib/search/updater.dart
```diff
@@ -106,6 +106,9 @@ class BatchIndexUpdater implements TaskRunner {
     }
     if (_batch.isEmpty) return;
 
+    final Timer timer = new Timer(const Duration(minutes: 5), () {
+      _logger.severe('Batch updated failed to complete in 5 minutes.');
```
jakobr-google: Do we want to complete the future here as well, possibly with an error? Or would that not help anything?

isoos: First I'd like to see if this is really the case. Completing with an error would make sense, but would introduce additional error handling too. Let's see the logs first.

PR #1311, opened 2018-05-22
Update a number of dependencies

kevmoo:


app/pubspec.yaml
```diff
@@ -33,13 +33,13 @@ dependencies:
   yaml: '^2.1.12'
   # pana version to be pinned
   # keep in-sync with app/lib/shared/versions.dart
-  pana: '0.11.1'
+  pana: '0.11.2'
```
kevmoo: Oops! Uh...I'll let you run w/ this @isoos

isoos: You'll need to change the `shared/versions.dart` too, and with that, update the runtimeVersion and the related test. Feel free to leave it out, I'll upgrade it (but wanted to do it with another change included, but it is not that important to bundle it).

kevmoo:
kk

kevmoo:
Honestly, I'll abandon this. Feel free to copy-paste the bits you want and
delete the branch.

Issue #992, opened 2018-02-19
Lower score for packages that fail dartdocs generation


mit-mit:
Once we get rid of the system issues with dartdoc generation, we should lower the score for individual packages that have dartdoc generation issues.

mit-mit:
Moving to M6 for consideration. But need to confirm that we are not failing anymore due to system issues.

isoos:
We are able to run `dartdoc` as part of the analysis, but there are a couple of reasons why we should keep them separate:

- `dartdoc` is run on all the versions of a package, not only on the ones that were published in the past two years. With the new direct linking between dependent packages, it would be important to keep those links alive.

- The coordination between `dartdoc` and `analyzer` is hard, especially when there is a new release, and both service tries to go through the new active jobs as fast as they can. If `analyzer` would produce a new `dartdoc` upload, it would either complicate things, or most likely duplicate (high-latency) work.

- We can update an `Analysis` much faster than upload a generated `dartdoc` content. We could re-do the analysis easily after each `dartdoc` run.

I think the better solution forward is to:
- keep the two job separate,
- fetch the latest dartdoc status for augmenting the analysis results,
- trigger a new analysis after a dartdoc upload completes.

(/cc @jakobr-google)

PR #1317, opened 2018-05-23
Apply platform override in a typed way.

isoos:
Trying to make the code less reliant on the stored data format, in preparation for the bigger data-handling update of #992.


isoos:
will redo this with a few additional changes.

PR #1319, opened 2018-05-23
Fix version upgrades

isoos:
Should fix CI


isoos:
This code is not affected by the same-name-variable bug, only `pana`.

Issue #1324, opened 2018-05-24
Update SDK to -dev.55+


kevmoo:
`shelf` and `shelf_static` are already broken. pkg:angular and friends are coming out today and they also depend on 55+

kevmoo:
Thanks!

PR #1325, opened 2018-05-25
Upgrade versions: SDK, pana, dependencies.

isoos:
#1324


isoos:
Deploying to staging.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Keep version in-sync with .travis.yml and app/lib/shared/versions.dart
```
jakobr-google: Should we update .travis.yml too? And should we add the .mono_repo.yml files to this comment as well?

isoos: Updated.

PR #1328, opened 2018-05-25
Add results from dartdoc service into the analysis results.

isoos:
#992

- `analyzer` will request the latest status of the dartdoc generation, and will report it as part of the maintenance score if it failed.
- `dartdoc` will notify `analyzer` that the given package should be re-analyzed (and updated with the new dartdoc results).


isoos:
This looks good in staging.

PR #1330, opened 2018-05-28
Search index rewrite: two-step token matching.

### LINKED ISSUE #1329: Relevant search results for short search terms
One example that was asked for: "REST API", which currently yields `cloud_firestore` as the first result, mostly because `rest` is part of the name.

isoos:
The new index will store only the tokens, and won't store the sub-strings (n-grams) anymore.

The lookup time is a bit longer, as we are trying to match a subset of tokens with the search terms, but in practice I haven't found slow lookups.

Closes #1329.


isoos:
PR is not ready yet, I'm still figuring out some edge-cases.

isoos:
PTAL. Update PR and also deployed to staging.

PR #1337, opened 2018-05-31
Disable implicit casts

isoos:
Seems like we don't have any potential bug, but nevertheless...

### DIFF
--- a/app/bin/service/search.dart
+++ b/app/bin/service/search.dart
@@ -78,7 +78,7 @@ void _main(FrontendEntryMessage message) {
     final scheduler = new TaskScheduler(
       batchIndexUpdater,
       [
-        new ManualTriggerTaskSource(taskReceivePort),
+        new ManualTriggerTaskSource(taskReceivePort.cast<Task>()),
         new IndexUpdateTaskSource(db.dbService, batchIndexUpdater),
         new DatastoreHeadTaskSource(
           db.dbService,
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -31,20 +31,21 @@ final Logger _logger = new Logger('pub.cloud_repository');
 void registerLoggedInUser(String user) => ss.register(#_logged_in_user, user);
 
 /// The active logged-in user. This is used for doing authentication checks.
-String get _loggedInUser => ss.lookup(#_logged_in_user);
+String get _loggedInUser => ss.lookup(#_logged_in_user) as String;
 
 /// Sets the active tarball storage
 void registerTarballStorage(TarballStorage ts) =>
     ss.register(#_tarball_storage, ts);
 
 /// The active tarball storage.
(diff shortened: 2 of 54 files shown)


jakobr-google:
Another good thing to check is to run the tests in Dart 2 mode. I've been using
```
find test -name \*_test.dart -exec dart --preview-dart-2 {} \;
```
which catches a bunch of these, and possibly more.

I've been fixing Dart 2 issues in the libraries that Pub uses, so we should also make sure to upgrade to the latest versions of appengine, gcloud, googleapis, and memcache to get the Dart 2 runtime fixes.

app/bin/service/search.dart
```diff
@@ -78,7 +78,7 @@ void _main(FrontendEntryMessage message) {
     final scheduler = new TaskScheduler(
       batchIndexUpdater,
       [
-        new ManualTriggerTaskSource(taskReceivePort),
+        new ManualTriggerTaskSource(taskReceivePort as Stream<Task>),
```
jakobr-google: Are you sure this works in Dart 2 mode? If I'm reading things right, a `ReceivePort` is a `Stream<dynamic>`, so this cast should fail at runtime.

app/lib/frontend/backend.dart
```diff
@@ -191,7 +198,7 @@ class GCloudPackageRepository extends PackageRepository {
         onPause: () => subscription.pause(),
         onResume: () => subscription.resume(),
         onCancel: () => subscription.cancel());
-    return controller.stream;
+    return controller.stream.cast<PackageVersion>();
```
jakobr-google: Instead, you could make `controller` a `StreamController<PackageVersion>`:
```dart
  final controller = new StreamController<PackageVersion>(...);
```
Then you wouldn't need to cast here.

app/lib/frontend/backend.dart
```diff
@@ -126,27 +130,30 @@ class Backend {
     final packageVersionKey = db.emptyKey
         .append(models.Package, id: package)
         .append(models.PackageVersion, id: version);
-    return (await db.lookup([packageVersionKey])).first;
+    return (await db.lookup([packageVersionKey]))
+        .cast<models.PackageVersion>()
```
jakobr-google: Using `cast()` wraps the list in a new object that does the cast on every access. It doesn't matter so much here, but I believe the recommendation is to cast the object returned by `first` instead:
```dart
return (await db.lookup(...)).first as PackageVersion;
```
https://www.dartlang.org/guides/language/effective-dart/usage#avoid-using-cast

isoos:
Fixed most things, but lot of the dependent libraries are not dart2-ready yet.

jakobr-google:
I've done a lot of Dart 2 readiness updates to package:appengine, gcloud, googleapis, and memcache recently, so make sure you get the latest available versions of those.

isoos:
Yeah, thanks for noting that. I've noticed that we have a few potential upgrades, but it wasn't picked up yet, will take a look.

Issue #1338, opened 2018-05-31
Forbid packages w/ both author and authors fields


kevmoo:
Are there any existing packages that fit this bill, @isoos ?

If not, it'd be great to restrict it!

isoos:
In the staging database and on the latest versions, we have only a single package: `crclib` that has both. I think it should be ok to restrict it.

isoos:
This has been deployed.

PR #1340, opened 2018-06-01
Prevent both author and authors in pubspec.yaml

isoos:
#1338


isoos:
> The nearby code sometimes throws a string, sometimes throws an exception. Is this string presented to the user? How about making a particular kind of exception type for pubspec errors that can be presented to the user, and not presenting the user with any other exception details?

The thrown object's .toString() is presented to the user, while the exception details go to the logs. I have no idea if a separate exception type would be beneficial, but it can be done.

I'll deal with the thrown String in a separate PR, as there seems to be happening in a lot of other places (link rule was not turned on).

sortie:
sgtm

Issue #1333, opened 2018-05-30
Simplify dartdoc storage


isoos:
Right now we upload all the generated files separately to the bucket and serve / proxy them from there. Uploading them one-by-one takes a long time, way more than generating them.

I'm wondering if it would be less wasteful on resources if we were uploading only the archive, and when the dartdoc service gets a request to something inside, it would use a disk-based cache to serve these (and download the package archive if it is not in the cache).

@jakobr-google wdyt?

jakobr-google:
I don't know how much local storage we have available, but it sound like it'd be worth exploring.

It would mean that first access to a package's doc would be slower, potentially across all of the (max 8) instances, though, so I'm not sure how well it'll perform.

isoos:
We have 8 instance only because doc generation (and upload) takes a long time. Once it is done, it will go back to 2 instances.

We could also pre-fetch the latest archives of the most popular packages, making their access fast.

isoos:
Based on a random sampling, I think the documentation of the latest stable versions compressed should be around 1-1.4 GB, but uncompressed size would be around 10-15 GB. However the access would not always on the latest, as cross-documentation links are versioned to the resolved package version.

The default instance that we are using has ~4 GB free disk space for our use. We can request more resources, but it may be cheaper to request memory and/or use memcache more aggressively for this.

Before going ahead with the actual implementation of the local-disk-based serving, I think I would like to do a bit of tracking first:
- storing the compressed and uncompressed size of the archive
- track the network traffic with the current direct-to-bucket requests
- emulate as if we were using an LRU-based local disk cache: what is the new disk and network usage

/rfc @kevmoo @sortie

kevmoo:
Leaving it as an engineering decision. I support measuring first.

Have you looked at: https://cloud.google.com/storage/docs/best-practices#uploading

Also: I wonder if "shelling out" to gsutil would be faster – something to investigate?

Or maybe create a gRPC client? https://github.com/googleapis/googleapis/tree/master/google/storagetransfer/v1

isoos:
> Have you looked at: https://cloud.google.com/storage/docs/best-practices#uploading
> Also: I wonder if "shelling out" to gsutil would be faster – something to investigate?

My (still not in-depth) search shows the following:
- there is no bulk upload api,
- `gsutil` uses the same api endpoints as we do,
- "storage transfer" is a service for transferring between buckets (e.g. from AWS S3, or between Goole Buckets)

[...]

PR #1342, opened 2018-06-01
Track and expose storage requirement estimates for locally cached dartdoc archives.

isoos:
#1333

### DIFF
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -321,3 +321,62 @@ class DartdocBackend {
     await deletePool.close();
   }
 }
+
+// Tracks the disk space and network requirements of downloading the dartdoc
+// package archive to the local disk and serving the content from there.
+// TODO: remove after tracking confirmed size requirements
+class ArchiveTracker {
+  final _pvMap = <String, _TrackedEntry>{};
+  final _sampleLines = <String>[];
+
+  ArchiveTracker() {
+    new Timer.periodic(const Duration(minutes: 10), (_) => _cleanup());
+    new Timer.periodic(const Duration(minutes: 45), (_) => _logSample());
+  }
+
+  List<String> get samples => _sampleLines;
+
+  void track(DartdocEntry entry) {
+    final key = '${entry.packageName}/${entry.packageVersion}';
+    final tracked = _pvMap.putIfAbsent(key, () => new _TrackedEntry());
+    tracked.archiveSize = entry.archiveSize ?? 0;
+    tracked.totalSize = entry.totalSize ?? 0;
+    tracked.lastAccess = new DateTime.now().toUtc();
+  }
+
+  void _cleanup() {
+    final now = new DateTime.now().toUtc();
+    _pvMap.removeWhere((key, value) {
+      final age = now.difference(value.lastAccess);
+      return age.inMinutes > 90;
+    });
+  }
+
+  void _logSample() {
+    final now = new DateTime.now().toUtc();
+    final count = _pvMap.length;
+
+    final total = _pvMap.values.fold(0, (sum, tracked) {
+      return sum + tracked.totalSize;
+    });
+    final totalMB = total ~/ _mb;
+    final archive = _pvMap.values.fold(0, (sum, tracked) {
+      return sum + tracked.archiveSize;
+    });
+    final archiveMB = archive ~/ _mb;
+    // roughly one and half days of samples
+    if (_sampleLines.length > 48) {
+      _sampleLines.removeAt(0);
+    }
+    _sampleLines.add(
+        '${now.toIso8601String()} count: $count, total: $totalMB MB, archive: $archiveMB MB');
+  }
+}
+
+class _TrackedEntry {
+  int totalSize;
+  int archiveSize;
+  DateTime lastAccess;
+}
+
+const _mb = 1024 * 1024;
--- a/app/lib/dartdoc/dartdoc_runner.dart
+++ b/app/lib/dartdoc/dartdoc_runner.dart
@@ -28,8 +28,10 @@ final Logger _logger = new Logger('pub.dartdoc.runner');
 final Uuid _uuid = new Uuid();
 
 const statusFilePath = 'status.json';
+const _archiveFilePath = 'package.tar.gz';
 const _buildLogFilePath = 'log.txt';
 const _dartdocTimeout = const Duration(minutes: 10);
+final Duration _twoYears = const Duration(days: 2 * 365);
 
 class DartdocJobProcessor extends JobProcessor {
   DartdocJobProcessor({Duration lockDuration})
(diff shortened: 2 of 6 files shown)


isoos:
Updated this with the latest insight into the problem: tracking 'obsolete' package access separately, as it seems we have the biggest low-risk gains there.

app/lib/dartdoc/backend.dart
```diff
@@ -320,3 +320,58 @@ class DartdocBackend {
     await deletePool.close();
   }
 }
+
+class ArchiveTracker {
```
sortie: There's no tests of this class, but perhaps this is temporary code? There's no documentation for this class. It's a bit unclear at first exactly what it is measuring.

app/lib/dartdoc/backend.dart
```diff
@@ -320,3 +320,58 @@ class DartdocBackend {
...
+  final _pvMap = <String, _TrackedEntry>{};
+  final _sampleLines = <String>[];
+
+  ArchiveTracker() {
+    _logger.info('Tracking: init.');
```
sortie: Suggestion: This message could be more specific and mention ArchiveTracker? It's also not in the same format as the  lines printed out by _logSamples. It doesn't say which tracker it is.

app/lib/dartdoc/backend.dart
```diff
@@ -320,3 +320,58 @@ class DartdocBackend {
...
+    new Timer.periodic(const Duration(minutes: 10), (_) => _cleanup());
+    new Timer.periodic(const Duration(minutes: 45), (_) => _logSample());
```
sortie: Consider: There's no way to prevent these from being turned on immediately if something wanted to construct such an object but not use it yet (pass it to something else). There's no way to turn them off when the program is done using the object. This may well be fine and I do the same in some of my code. Since this class is constructor as global objects, there's no way to import the files in question without automatically starting these side effects, making it harder to reuse the code for testing and other purposes.

The cleanup should happen as part of the logging samples. That way the intervals of cleanup and logging, which doesn't align, doesn't interfere with each other.

app/lib/dartdoc/backend.dart
```diff
@@ -320,3 +320,58 @@ class DartdocBackend {
...
+  }
+
+  List<String> get samples => _sampleLines;
+
+  void track(DartdocEntry entry) {
+    final key = '${entry.packageName}/${entry.packageVersion}';
+    final tracked = _pvMap.putIfAbsent(key, () => new _TrackedEntry());
+    tracked.archiveSize = entry.archiveSize ?? 0;
+    tracked.totalSize = entry.totalSize ?? 0;
+    tracked.lastAccess = new DateTime.now();
```
sortie: This is in local time, not UTC. New dart has DateTime.utc() IIRC. You cannot reliably do calculations with local time (like you do below) because of time zones, etc. It should be reasonably possible to do such calculations in UTC. Otherwise consider using a StopWatch and then measure its value on each use and use that to determine the age of objects, as stopwatches provide monotonic linear time.

app/lib/dartdoc/backend.dart
```diff
@@ -320,3 +320,58 @@ class DartdocBackend {
...
+  }
+
+  void _cleanup() {
+    final now = new DateTime.now();
+    _pvMap.removeWhere((key, value) {
+      final age = value.lastAccess.difference(now).abs();
```
sortie: No need for abs() if you use monotonic time (where you know now is not before lastAccess) and you do "now.difference(lastAccess)".

app/lib/dartdoc/backend.dart
```diff
@@ -320,3 +320,58 @@ class DartdocBackend {
...
+      return age.inMinutes > 90;
+    });
```
sortie: This is a linear scan and doesn't scale well if there are a lot of entries. You could use a Queue that lets you efficiently dequeue the oldest entries. However, I don't think Dart's Queue class lets you efficiently find and remove an object from the middle (so it can be inserted at the end). Perhaps there's a priority queue class out there that could be used. But if the cache isn't too big, this can be fine.

What is being measured here? This cleanup seems detached from other stuff, like some cache whose size is being measured. Is it an attempt at a moving window statistic? You could have buckets with per-minute or per-15-minutes statistics that you just add to (though you'll probably want to keep track of which entries have been added to those buckets).

app/lib/dartdoc/backend.dart
```diff
@@ -320,3 +320,58 @@ class DartdocBackend {
...
+  }
+
+  void _logSample() {
+    // roughly one and half days of samples
+    if (_sampleLines.length > 48) {
+      _sampleLines.removeAt(0);
```
sortie: Suggestion: This is an inefficient ring buffer, but since it's bounded to 48 entries every 45 minutes, this is fine. I suggest moving this code down to just before the new sample line is inserted.

app/lib/dartdoc/backend.dart
```diff
@@ -320,3 +320,58 @@ class DartdocBackend {
...
+    }
+    final now = new DateTime.now();
+    final count = _pvMap.length;
+
+    final totalMB = _pvMap.values.fold(0, (sum, tracked) {
+      return sum + tracked.totalSize ~/ _mb;
```
sortie: You should sum it together as bytes and then convert to the right magnitude at the end. This code rounds down to the previous megabyte, losing a lot of accuracy.

app/lib/dartdoc/backend.dart
```diff
@@ -320,3 +320,58 @@ class DartdocBackend {
...
+    });
+    final archiveMB = _pvMap.values.fold(0, (sum, tracked) {
+      return sum + tracked.archiveSize ~/ _mb;
+    });
+    _sampleLines.add(
+        '${now.toIso8601String()} count: $count, total: $totalMB MB, archive: $archiveMB MB');
```
sortie: It doesn't say which tracker it is.

Suggestion: I suggest having a utility function that can pretty print an arbitrary byte quantity with the right unit. That will handle if the cache is very large of very small. I did that for another server side dart project and I like it.

app/lib/dartdoc/backend.dart
isoos: Yes, only temp code here. It just needs to give us a rough estimate on how much local disk space we'd need, and be able to estimate roughly how much network latency we would incur when downloading the package archives. Adding comment.

app/lib/dartdoc/backend.dart
isoos: I thought that the cleanup should be more frequent, similar to what I'd do if there would be an actual disk-based cache. The log intervals are more wide, with the goal having a good granularity over a day's use, but not too detailed information.

Re: side effects and reusability: I'd throw this class out once we have the info. I see no reason to make it robust and testable, but I can certainly make it so if needed.

app/lib/dartdoc/backend.dart
isoos: Good point, fixing to `new DateTime.now().toUtc()`. Btw `DateTime.diff` works the same way, regardless of time zones.

app/lib/dartdoc/backend.dart
isoos: Fixing. Bad habit: I don't want to think about which should be the later timestamp :)

app/lib/dartdoc/backend.dart
isoos: The thought process behind the cleanup is the following: when we have an access to a `DartdocEntry`, it would mean that we either download its related package archive to local disk, or we have it already there. After 90 minutes of inactivity we would delete it from the local disk. The cleanup is done in 10 minutes intervals, making sure we don't stretch too much over the 90 minutes, but the sampling of the required space is not important to be precise.

app/lib/dartdoc/backend.dart
isoos: Now that you mention it, I shall have it somewhere too. Not sure if it is that important here: for the storage anything less than 100 MB is essentially the same. May be useful for the compressed archive length, but based on the staging datasets, MB range still gives us useful insights (e.g. total archive size is ~7MB for 120 packages -> ~58K per package on average).

isoos:
PTAL

app/lib/dartdoc/backend.dart
sortie: Not sure it does, because of daylight savings time and other trickeries. UTC is at least stable.

app/lib/dartdoc/backend.dart
sortie: Yeah I always look at up as well.

isoos:
> I can't get github to show me the differences between your first patchset and the latest patchset, like gerrit can. Is there a way? It was greyed out for me.

I believe it is not possible. Next time I can do it in a separate commit, but I usually squash the trivial and direct followup-up updates early.

sortie:
Alright that's fine. It's easy to compare patchsets on gerrit, it's unfortunate github doesn't support that. I can diff commits locally as a workaround.

PR #1344, opened 2018-06-02
Use FLUTTER_ROOT in env variables.

isoos:
https://github.com/dart-lang/dartdoc/issues/1705


isoos:
> lib/shared/configuration.dart should be app/lib/shared/configuration.dart.

Fixed.

> I assume no change needs to be done to app/script/setup-flutter.sh and app/lib/shared/configuration.dart.

Yeah, those are fine.

PR #1345, opened 2018-06-04
Update window.hash when using custom scrollTo

### LINKED ISSUE #1343: pub package page hijacks location
Steps to recreate:

1. I want to share a link to the "Asynchronous Tests" section of [test's README](https://pub.dartlang.org/packages/test)
2. I click the link at the top, in the table of contents, the link to: https://pub.dartlang.org/packages/test#asynchronous-tests
3. I confirm, yes, this is the content I want to link.
4. I go to the location bar, expecting an anchor in the URL, but it is still just https://pub.dartlang.org/packages/test.

Expected result:

Like a normal website, the anchor should be in the URL. It should read: https://pub.dartlang.org/packages/test#asynchronous-tests

isoos:
Closes #1343.


isoos:
@sortie: The script change itself is not much, but on top of that:
- We do check in the compiled static js.
- We do check in the compiled static js's metadata (@kevmoo: do we ever use this?)
- We need to update the templates test, as the script (and the main css style) is referenced with a content-based hash, and these need to be updated when the script or the css changes.

kevmoo:
> We do check in the compiled static js's metadata

Wha? 🤷‍♂️

isoos:
>> We do check in the compiled static js's metadata
> Wha? 🤷‍♂️

I'm happy to remove it if we don't use it :)

kevmoo:
I don't know what "it" is...

isoos:
> I don't know what "it" is...

`static/js/script.dart.js.info.json`

I'm not sure what use we have for it.

kevmoo:
Ah! It's useful to easily track compiled code size between builds...

Doesn't hurt anything – but doesn't offer much value...

isoos:
> Can we stop checking in the compiled script? Could it be compiled on deployment?

The template tests do rely on it being present and make its hash to be part of the template output.

Nevertheless, I would merge this PR and we can take it out in a subsequent change, and create a build process around it. The template part can be mocked.

Issue #1347, opened 2018-06-04
Log entry with size 277.4K exceeds maximum size of 110.0K


isoos:
I've noticed the following error in the `dartdoc` logs:

````
An error occured while writing log entries:
Error:GrpcError: "Log entry with size 277.4K exceeds maximum size of 110.0K" (code: 3)
#0      Client.invoke (package:appengine/src/grpc_api_impl/grpc.dart:216:9)
#1      Channel.invoke (package:appengine/src/grpc_api_impl/grpc.dart:45:19)
#2      LoggingServiceV2Api.writeLogEntries (package:appengine/src/grpc_api/dart/google/logging/v2/logging.pb.dart:393:20)
#3      SharedLoggingService.flush (package:appengine/src/grpc_api_impl/logging_impl.dart:302:17)
````

It looks like we have similar messages in `analyzer` logs, but none in the `app` and `search`. 

Reducing the number of log messages seems to be a good idea for now.

sortie:
The error suggests to me it's a single log entry that's too large, if that's the case then doing fewer small log messages won't make a difference? Or perhaps this is after a bunch of log messages have been aggregated?

isoos:
I think [this code](https://github.com/dart-lang/appengine/blob/master/lib/src/grpc_api_impl/logging_impl.dart#L298) prepares a store request for all the entries enqueued, and it seems that flush is called either after 3 seconds or reaching 25 items, whichever is first.

The text which we put inside the log messages are not long (should be under 1K), but I guess the appengine library stores additional meta-information there too, and 20+ of them could easily push it over the 110k limit.

Making our texts shorter may help, but I'd also use the opportunity to remove some of the debug information we don't need anymore.

isoos:
An update: `dartdoc` service is still affected by this, and `analyzer` too for a lesser extent. `app` and `search` seems to be doing fine.

isoos:
Update: both `dartdoc` and `analyzer` remains affected, but many log-affecting changes were merged after the current release.

isoos:
It is still affecting the new release (`analyzer` and `dartdoc` services), and it seems that the frequency and the exceed amount has increased.

isoos:
[...]

PR #1349, opened 2018-06-04
Fewer log message during dartdoc upload.

isoos:
#1347


isoos:
> Could it be that the error is that a single log write is too large?

It could be, but I don't see which of our messages would be too long (and too many at the same time).

PR #1350, opened 2018-06-05
Update customization test with dartdoc 0.20.0 outputs.

isoos:
Changes are only in the generator metadata and in the links to the SDK documentation due to version change.


isoos:
> Where is the matching update to dartdoc 0.20? Did that happen as part of another PR?

Yeah, it did happen in #1351, but this test is kind of a pre-requisite for that to happen. Not necessarily in the merge order, but we should upgrade the version unless the test is upgraded.

PR #1358, opened 2018-06-11
Add API page references to the search results.

isoos:
There is a bit more processing going on inside the search index, but I haven't noticed much slowdown during regular searches. The current changes are non-breaking, either of the services can be upgraded without the other.

An early version of the code is still deployed to staging, but the style changes have not been deployed there. The result looks like this:

screen shot 2018-06-11 at 14 19 04

Later on the visible text of the links should become `ecb_fx library` or `EcbFxClient class`, but that will be done only with the new dartdoc extraction code PR.

### DIFF
--- a/app/lib/shared/search_service.dart
+++ b/app/lib/shared/search_service.dart
@@ -441,11 +441,37 @@ class PackageScore extends Object with _$PackageScoreSerializerMixin {
   @JsonKey(includeIfNull: false)
   final double score;
 
+  @JsonKey(includeIfNull: false)
+  final List<ApiPageRef> apiPages;
+
   PackageScore({
     this.package,
     this.score,
+    this.apiPages,
   });
 
   factory PackageScore.fromJson(Map<String, dynamic> json) =>
       _$PackageScoreFromJson(json);
+
+  PackageScore change({
+    double score,
+    List<ApiPageRef> apiPages,
+  }) {
+    return new PackageScore(
+      package: package,
+      score: score ?? this.score,
+      apiPages: apiPages ?? this.apiPages,
+    );
+  }
+}
+
+@JsonSerializable()
+class ApiPageRef extends Object with _$ApiPageRefSerializerMixin {
+  final String title;
+  final String path;
+
+  ApiPageRef({this.title, this.path});
+
+  factory ApiPageRef.fromJson(Map<String, dynamic> json) =>
+      _$ApiPageRefFromJson(json);
 }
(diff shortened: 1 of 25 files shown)


mit-mit:
Looks great!

app/lib/shared/search_service.dart
```diff
@@ -441,11 +441,37 @@ class PackageScore extends Object with _$PackageScoreSerializerMixin {
...
       _$PackageScoreFromJson(json);
+
+  PackageScore change({
+    double score,
+    List<ApiPageRef> apiPages,
+  }) {
+    return new PackageScore(
+      package: package,
+      score: score ?? this.score,
+      apiPages: this.apiPages ?? apiPages,
```
sortie: Did you swap the order here?

app/pubspec.lock
```diff
@@ -576,4 +576,4 @@ packages:
     source: hosted
     version: "2.1.14"
 sdks:
-  dart: ">=2.0.0-dev.55.0 <=2.0.0-dev.59.0"
+  dart: ">=2.0.0-dev.55.0 <=2.0.0-dev.60.0"
```
sortie: Shouldn't this have happened in another PR?

app/test/frontend/golden/authorized_page.html
```diff
@@ -24,7 +24,7 @@
   <meta property="og:description" content="Pub is the package manager for the Dart programming language, containing reusable libraries &amp; packages for Flutter, AngularDart, and general Dart programs." />
   <link rel="alternate" type="application/atom+xml" title="Updated Packages Feed for Pub" href="/feed.atom" />
   <link href="/static/highlight/github.css" rel="stylesheet" />
-  <link href="/static/css/style.css?hash=18js5k8h2d638mkc6lerk5gr5g5hjavp" rel="stylesheet" type="text/css" />
+  <link href="/static/css/style.css?hash=cc69ncs8qti1ldoshp4tnusgh051fi9j" rel="stylesheet" type="text/css" />
```
sortie: Why is this updated like this? Could a http style etag or something work instead?

app/lib/shared/search_service.dart
isoos: good catch, fixing.

app/test/frontend/golden/authorized_page.html
isoos: We have both: ETag support (+ IfModifiedSince support), and an additional hash in the URL. The later is used because that way we can use a really high value for browser-side cache timeout, otherwise we would have a window where we are already serving the new version with e.g. a new layout, but the browser may not even attempt to request the resource from the server. It is used only for the two critical resource for use: the main CSS and the main JS file.

PR #1361, opened 2018-06-13
Refactoring static files handling for tests and auto-generated assets

isoos:
This is a follow-up PR to prepare the removal of the checked-in script.dart.js file. I've restructured the handling of static files to enable mocking in templates, and also added a pre-submit check to make sure that the mocked files do exists.

### DIFF
--- a/app/lib/frontend/static_files.dart
+++ b/app/lib/frontend/static_files.dart
@@ -110,18 +110,18 @@ class StaticUrls {
       'style_css': _getCacheableStaticUrl('/css/style.css'),
     };
   }
-}
 
-/// Returns the URL of a static resource
-String _getCacheableStaticUrl(String relativePath) {
-  if (!relativePath.startsWith('/')) {
-    relativePath = '/$relativePath';
-  }
-  final String requestPath = '${staticUrls.staticPath}$relativePath';
-  final file = staticsCache.getFile(requestPath);
-  if (file == null) {
-    throw new Exception('Static resource not found: $relativePath');
-  } else {
-    return '$requestPath?hash=${file.etag}';
+  /// Returns the URL of a static resource
+  String _getCacheableStaticUrl(String relativePath) {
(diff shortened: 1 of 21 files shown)


app/lib/frontend/static_files.dart
```diff
@@ -27,25 +29,27 @@ String _resolveStaticDirPath() {
 }
 
 class StaticsCache {
```
sortie: Where is the documentation?

app/lib/frontend/static_files.dart
```diff
@@ -27,25 +29,27 @@ String _resolveStaticDirPath() {
...
-  final String staticPath;
+  final String staticPath = _defaultStaticPath;
```
sortie: Now it's no longer possible to override this?

app/lib/frontend/static_files.dart
```diff
@@ -78,19 +82,25 @@ class StaticFile {
   );
 }
 
-final staticUrls = new StaticUrls(staticsCache.staticPath);
+final staticUrls = new StaticUrls();
```
sortie: Suggestion: This is a global object that scans the filesystem in its constructor. That may be fine? Perhaps an initializer method or initializing on first use may be preferable?

app/lib/frontend/static_files.dart
isoos: adding

app/lib/frontend/static_files.dart
isoos: I'm not sure if there is a need for the override, and if that arises, it should be easy to do.

app/lib/frontend/static_files.dart
isoos: Fair point, I was already considering to make this more similar to the other `registerXYZ` initialization we have in other parts of the codebase, will do it in a follow-up.

Issue #1363, opened 2018-06-13
Update search logging and re-indexing


isoos:
- ~[ ] The 200ms threshold is interesting only in periods where there is no parallel re-indexing running.~
- ~[ ] When re-indexing is running, it should be increased to 500ms.~
- [ ] Should consider lowering indexing batch size.
- [ ] Should consider package hash that would make re-indexing skip identical packages.

sortie:
The log warning just contained the search URL. It should be prefixed with "Slow search" so it's more visible what it is at a glance without expanding it.

isoos:
It turns out that the increased number of warnings did indicate a slowdown, making most if the task list obsolete. As #1366 fixes a huge chunk of the latency, I'm closing the rest, keeping the warning at 200ms seems reasonable, and it turned out to be useful for this time.

isoos:
> Is there any monitoring of latency for different APIs, keeping track of different percentiles?

No, I think earlier the dashboard and tools that google cloud console provided were enough. I've added this to keep an eye on search internals, and we have another metric that exposes the event loop latency on the /debug url, but it is not monitored either.

PR #1366, opened 2018-06-13
Make ngram similarity calculation faster.

isoos:
#1363

After profiling it a bit, it turns out that `Set.addAll()` was the most time-consuming and dominating factor for a search request, and rewriting this small method should provide a quick win for the search latencies.


isoos:
> How much faster is this?

On my Macbook, a random, non-representative query took 70ms with the old version, and 11ms with the new version.

PR #1367, opened 2018-06-14
Fix a few Dart2 type check errors.

isoos:
Just a few low-hanging fruit to fix, a couple more remains.

### DIFF
--- a/app/lib/frontend/model_properties.dart
+++ b/app/lib/frontend/model_properties.dart
@@ -9,6 +9,15 @@ import 'dart:convert';
 import 'package:gcloud/db.dart';
 import 'package:yaml/yaml.dart';
 
+Map<String, dynamic> _loadYaml(String yamlString) {
+  final Map map = loadYaml(yamlString);
+  // TODO: remove this part after yaml returns a proper map
+  if (map is YamlMap) {
+    return json.decode(json.encode(map)) as Map<String, dynamic>;
+  }
+  return map as Map<String, dynamic>;
+}
+
 class Pubspec {
   final String jsonString;
   Map _json;
@@ -20,7 +29,7 @@ class Pubspec {
         _json = jsonMap;
 
   factory Pubspec.fromYaml(String yamlString) =>
-      new Pubspec.fromJson(loadYaml(yamlString) as Map<String, dynamic>);
+      new Pubspec.fromJson(_loadYaml(yamlString));
 
   Map get asJson {
     _load();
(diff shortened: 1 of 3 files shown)


app/lib/frontend/model_properties.dart
```diff
@@ -20,7 +29,7 @@ class Pubspec {
         _json = jsonMap;
 
   factory Pubspec.fromYaml(String yamlString) =>
-      new Pubspec.fromJson(loadYaml(yamlString) as Map<String, dynamic>);
+      new Pubspec.fromJson(_loadYaml(yamlString));
```
sortie: .fromYaml accepts a string while .fromJson accepts a map? That's a bit inconsistent. I prefer to have .fromJson do JSON decoding and having a .fromMap that constructs the object from a decoded object.

app/lib/frontend/model_properties.dart
```diff
@@ -9,6 +9,15 @@ import 'dart:convert';
 import 'package:gcloud/db.dart';
 import 'package:yaml/yaml.dart';
 
+Map<String, dynamic> _loadYaml(String yamlString) {
+  final Map map = loadYaml(yamlString);
+  // TODO: remove this part after yaml returns a proper map
+  if (map is YamlMap) {
```
sortie: Can yaml contain a list that contains a map? If so we also want to do it in that case.

app/lib/search/index_simple.dart
```diff
@@ -645,9 +645,9 @@ class TokenIndex {
     int sumFn(int sum, String str) =>
         sum + math.min(100, str.length * str.length);
 
-    final intersectionWeight = intersection.fold(0, sumFn);
+    final intersectionWeight = intersection.fold<int>(0, sumFn);
```
sortie: Oh I thought this would be deduced.

app/lib/frontend/model_properties.dart
isoos: Yeah, that is unfortunate, but all of the JSON serialization code is using `factory ClassName.fromJson(Map<String, dynamic)` as their deserialization constructor, and I'm not sure if the [code generator](https://pub.dartlang.org/packages/json_serializable) would work with a different signature. (Also they use `Map toJson()` instead of `toMap()` and `String toJson()` :( )

app/lib/frontend/model_properties.dart
isoos: I believe it can never be the case with `pubspec.yaml`, and the `pub` client validation and/or the upload will fail if it is not a Map.

app/lib/frontend/model_properties.dart
isoos: Although, `fromJson(Map)` may be acceptable if its pair is `parseJson(String)`...

PR #1374, opened 2018-06-15
Upgrade versions

isoos:
Upgrading: Dart SDK, Flutter, dependencies. I'm going to deploy this to staging soon.

Note: the upgrade does not have much risk, as the new`pana` (with `package:pubspec_parse`) is not yet used.


isoos:
@sortie: staging looks good, I'll prepare a few small changes based on the logs, but nothing prevents a new deploy.

isoos:
> Is there a list of steps you go through to update this dependency?

I'll write a follow-up PR in the README about it.

PR #1380, opened 2018-06-18
Fix use of the dependency graph.

isoos:
I was chasing log entries, and it turns out that we have never really called the cascading analysis of the dependent packages when a new version is uploaded. We've stored the callback in the backend, and never called it or passed it to the real place where it is getting called.

I'm also changing the `await` logic of the notification: if there is an upload happening right after the instance is started, it could wait a minute on the graph to be built, before confirming to the upload.


isoos:
Tested with an upload in staging, it looks like working as expected.

PR #1382, opened 2018-06-18
Remove the need of syncing Flutter version between scripts.

isoos:
One less item to keep in mind and document :)

### DIFF
--- a/app/script/setup-flutter.sh
+++ b/app/script/setup-flutter.sh
@@ -22,9 +22,14 @@ then
   exit 1
 fi
 
-git clone -b v0.5.4 --single-branch https://github.com/flutter/flutter.git $FLUTTER_SDK
+if [[ -z "$1" ]];
+then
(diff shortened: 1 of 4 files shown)


app/script/setup-flutter.sh
```diff
@@ -22,9 +22,14 @@ then
...
 fi
 
-git clone -b v0.5.4 --single-branch https://github.com/flutter/flutter.git $FLUTTER_SDK
+if [[ -z "$1" ]];
+then
+  echo "Version argument is missing."
+  exit 1
+fi
+
+echo git clone -b $1 --single-branch https://github.com/flutter/flutter.git $FLUTTER_SDK
```
sortie: "echo" seems like leftover debug code.

isoos: Indeed, thanks for catching it!

isoos:
This looks good in staging, merging.

PR #1383, opened 2018-06-18
Initialize pkg/pub_dartdoc along the global dartdoc intialization.

isoos:
I'd like to keep the two versions side-by-side and test it before removing the global one.

### DIFF
--- a/app/lib/shared/service_utils.dart
+++ b/app/lib/shared/service_utils.dart
@@ -281,7 +281,9 @@ Future initGlobalDartdoc(Logger logger) async {
 Future initPkgDartdoc(Logger logger) async {
   logger.info('Initializing pkg/pub_dartdoc');
   final dir = Platform.script.resolve('../../pkg/pub_dartdoc').toFilePath();
-  final pr = await Process.run('pub', ['get'], workingDirectory: dir);
+  final pr = await Process
+      .run('pub', ['get'], workingDirectory: dir)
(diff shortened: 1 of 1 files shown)


app/lib/shared/service_utils.dart
```diff
@@ -272,6 +278,17 @@ Future initDartdoc(Logger logger) async {
   }
 }
 
+Future initPkgDartdoc(Logger logger) async {
+  logger.info('Initializing pkg/pub_dartdoc');
+  final dir = Platform.script.resolve('../../pkg/pub_dartdoc').toFilePath();
+  final pr = await Process.run('pub', ['get'], workingDirectory: dir);
```
sortie: Perhaps have a time out if the command doesn't halt.

isoos:
> What exactly is happening here?

Instead of using the SDK's `dartdoc` (or the globally activated dartdoc), we want to use `pkg/pub_dartdoc`, which extends it with further data extraction (for search and scoring). We could depend on that project as a regular library, but that would tie us to the dependencies of `dartdoc`, and we could be on a different upgrade schedule. Instead, we run it as it were a separate app, and for that, we need to fetch the dependencies before calling it.

app/lib/shared/service_utils.dart
isoos: Good point, adding it.

isoos:
Looks good in staging, merging.

PR #1385, opened 2018-06-19
Simplified static file cache handling.

isoos:
A follow-up to a previous comment:
- Enables explicit override of the static file cache (for tests).
- Simplifies request handler and templates.

### DIFF
--- a/app/lib/frontend/static_files.dart
+++ b/app/lib/frontend/static_files.dart
@@ -9,14 +9,21 @@ import 'package:mime/mime.dart' as mime;
 import 'package:path/path.dart' as path;
 
 const String _defaultStaticPath = '/static';
+const _staticRootPaths = const <String>['favicon.ico', 'robots.txt'];
 
-/// Stores binary data for /static
-final StaticsCache staticsCache = new StaticsCache();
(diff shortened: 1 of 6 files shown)


app/lib/frontend/static_files.dart
```diff
@@ -9,14 +9,21 @@ import 'package:mime/mime.dart' as mime;
...
-  '/robots.txt',
-]);
+/// The static file cache. If no cache was registered before the first access,
+/// the default instance will be created.
+StaticFileCache get staticFileCache =>
+    _cache ??= new StaticFileCache.withDefaults();
+
+/// Register the static file cache.
+/// Can be called only once, before the static file cache is set.
+void registerStaticFileCache(StaticFileCache cache) {
```
sortie: Or void set staticFileCache? But this is fine.

isoos: Works for me either way, but we have `registerXYZ` calls for similar stuff (although it is for scoped appengine stuff), and I'd  keep the same pattern here.

sortie: I think it also looks more readable as a singleton

Issue #1348, opened 2018-06-04
Consider using pkg:pubspec_parse 


kevmoo:
https://github.com/dart-lang/pub-dartlang-dart/blob/6a3779161dddd987079cfe7c107697850ab24855/app/lib/frontend/model_properties.dart#L12

Note: pubspec_parse is strict and not lazy, so it may crash on pubspecs that we've allowed.

So...be careful

isoos:
I'd bundle this with the next `pana` release that uses it.

kevmoo:
Sure. Make sure to test over a large corpus first, though...

isoos:
@kevmoo: on the staging database, #1388 revealed the following issues:
- Illegal scheme character
- Unrecognized keys: [git]; supported keys: [version, hosted]
- Unsupported value for `url`.

The following packages were affected:

````
FileTeCouch 0.6.0
FileTeCouch 0.6.1
FileTeCouch 0.6.2
bwu_utils 0.1.0-dev.1
dartemis_toolbox 0.4.0
dependency_injection 0.1.7
dev_test 0.10.0
dev_test 0.10.1
fs_shim 0.6.4
fs_shim 0.6.5
fs_shim 0.6.7
fs_shim 0.7.0
fs_shim 0.7.1
graph 0.0.2
guava_event_bus 0.0.6
guava_event_bus 0.1.0
idb_shim 1.3.7
idb_shim 1.4.0
lightning 2.1.3
material2_dart 0.4.0
material2_dart 0.4.1
material2_dart 0.4.2
material2_dart 0.4.3
material2_dart 0.4.4
mojito 0.6.0
opensubtitles_api 0.1.0
polymer_dependency_injection 0.0.14
redstone_jwt_plugin 0.0.1
redstone_spa_plugin 0.0.1
rethinkdb 0.1.0
sembast 1.3.5
sembast 1.3.6
sembast 1.3.7
sembast 1.3.9
sembast 1.4.0
sembast 1.5.0
sembast 1.6.0
stomp_sockjs 0.2.0
web 0.0.1
web_server 2.0.0
web_server 2.0.0+1
web_server 2.0.0+2
web_server 2.0.0+3
````

This also blocks their `pana` analysis.

kevmoo:
Let me see if I can fix that...

kevmoo:
https://github.com/dart-lang/pubspec_parse/issues/13

kevmoo:
[...]

PR #1386, opened 2018-06-19
Use package:pubspec_parse

isoos:
#1348

Will be tested on staging before merging it.

/cc @kevmoo


app/lib/frontend/model_properties.dart
```diff
@@ -7,6 +7,7 @@ library pub_dartlang_org.model_properties;
 import 'dart:convert';
 
 import 'package:gcloud/db.dart';
+import 'package:pubspec_parse/pubspec_parse.dart' as pubspek show Pubspec;
```
sortie: pubspek? Intentional?

kevmoo:
unblocked now. Use `^0.1.2` of pubspec_parse

isoos:
@kevmoo: thanks, that fixes the parsing issues we've got in the existing database. I'll do another round of testing before merging this.

PR #1387, opened 2018-06-19
Upgrade pana to 0.11.4

isoos:
#1348

As the new `pana` version is already using `pubspec_parse` this change will be coupled with #1386 and will be tested on staging before merging it.


app/lib/shared/versions.dart
```diff
@@ -7,15 +7,15 @@ import 'package:pub_semver/pub_semver.dart';
 import 'utils.dart' show isNewer;
 
 // update this whenever one of the other versions change
-final String runtimeVersion = '2018.6.15';
+final String runtimeVersion = '2018.6.19';
```
sortie: Can this use a single source of truth?

isoos: In a sense, this is the source of truth... We started out pinning the Flutter and the `pana` versions, but eventually realized that we need to pin the SDK, the dartdoc version, and we should be able to mark our own version-independent updates too. Those are merged into this `runtimeVersion` and it annotates a specific configuration of all these variables.

PR #1396, opened 2018-06-21
 Make sure pub_dartdoc_data.dart is synced between the projects.

isoos:
- `app` should not directly depend on `pkg/pub_dartdoc` (or the other way around), because it helps to keep the dependency upgrade paths separate.
- dynamically copying the files won't work either, as copy should not happen after the given project is started.
- added test to make sure they are in-sync.


isoos:
The data file will be read in `search` and in `dartdoc` (or in the `analyzer`) service, both need to have full parse of it.

app/test/dartdoc/pub_dartdoc_data_sync_test.dart
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:test/test.dart';
+
+void main() {
+  test('pub_dartdoc_data.dart should be the same with pkg/pub_dartdoc', () {
```
sortie: s/same with/same as/g

Issue #1352, opened 2018-06-05
New model (for unified score and more)


isoos:
The [discussion around the dartdoc score](https://github.com/dart-lang/pana/issues/269) identified the need for a new model on how we store the score(s). Until now we have stored it in `PackageVersionAnalysis`, which is tightly-coupled with storing the result and serialization of `pana`. 

The model also introduced some operational complexity, as we built the UI around the internals of `pana`, and changes to it require consideration on how we roll out a release. We've also developed a few things to cache the quick summary of the analysis (`AnalysisExtract`) in order to display it along the various UI places.

Considering the experiences on how we use the data, and a few things that are in the pipeline, I'd suggest the following:

- Let's introduce a Datastore entity for storing the high-level analysis data (essentially `AnalysisExtract` + the new dartdoc score), but decoupled from the `pana`-related work.

- Let's introduce a Datastore entity and a pub-site-only data model for the output of `pana` and `dartdoc` analysis. This should contain the information on the analysis tab, plus the collected suggestions from both pana and dartdoc. For each new serialization change of `pana` we shall transform it to the pub site internals. This would add a bit of work when there is a breaking change in pana, but it would reduce the deploy worries and complexities.

- If/when we'll have a facility for runtime analysis, it would be easy to adapt its result into the above items.

As far as I can tell, this should be possible in an incremental way, and we may do a full migration in a couple of deploys.

/cc @sortie @kevmoo @mit-mit

isoos:
It looks like that the model can't be as simple as I've originally suggested, because it lacks the ability of handling the rollback of a partial rollout. We'll need to extend it with the `runtimeVersion` as part of the key, but at that point it could also merge with the current job processing (which has the same issues with the rollback, for the very same reasons).

The updated suggestion:

- `ScoreCard` and their detailing `ScoreCardReport` ids should be in the form of `package/version/runtimeVersion`.
- Each deployed runtime should keep track of or if needed update their versioned entries, independent of other runtimes. That way if there is a rollback, the new entries created by the new runtime will never interfere with the old ones.
- A GC process could periodically scan the table and remove entries that were not updated for more then a few months.
- Each deployed runtime should try to load their own version first, and if needed, fallback to earlier runtime versions, but never load later runtime versions.

isoos:
[...]

PR #1398, opened 2018-06-21
Proposal: ScoreCard model

isoos:
This is the first step for #1352:

`ScoreCard` would be the new all-unifying Datastore entity that holds PackageVersion score and other details (the current `AnalysisExtract`), while `ScoreCardReport` would be the new place for detailed reports for: `pana` (the current `AnalysisView`), `dartdoc` (coverage report and score), and possible further ones like runtime analysis.

I've opted to not having a two-level structure (like `Package` and `PackageVersion`), rather having a single entry for a given combination of that.

A long-term aspect: this should make the items in `lib/analysis/models.dart` obsolete.

### DIFF
--- a/app/lib/scorecard/models.dart
+++ b/app/lib/scorecard/models.dart
@@ -1,0 +1,118 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:gcloud/db.dart' as db;
+import 'package:meta/meta.dart';
+
+import 'package:pub_dartlang_org/search/scoring.dart'
+    show calculateOverallScore;
+
+import '../shared/model_properties.dart';
+
+String _id(String packageName, String packageVersion) =>
+    '$packageName/$packageVersion';
+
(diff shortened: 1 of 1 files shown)


app/lib/scorecard/models.dart
```diff
@@ -0,0 +1,118 @@
...
+  DateTime packageVersionCreated;
+
+  @db.BoolProperty()
+  bool isDiscontinued;
+
+  @CompatibleStringListProperty()
+  List<String> panaPlatformTags;
+
+  @db.DoubleProperty()
+  double documentationScore;
```
sortie: Document the fields in this key data class? For instance, is the values in documentationScore between 0.0 and 1.0, or 0.0 and 100.0?

app/lib/scorecard/models.dart
```diff
@@ -0,0 +1,118 @@
...
+    @required this.packageVersion,
+    @required this.packageCreated,
+    @required this.packageVersionCreated,
+  }) {
+    id = _id(packageName, packageVersion);
+  }
+
+  void updateOverallScore() {
+    // TODO: use documentationScore too
+    overallScore = calculateOverallScore(
```
sortie: Suggestion: Can overallScore be a getter that dynamically returns the score? Or can healthScore, maintenanceScore, popularityScore, and documentationScore be setters that also automatically regenerates the score?

PR #1403, opened 2018-06-22
Make dartdoc test compatible with dart 2 preview.

isoos:
Not sure why, but the dart 2 preview tests produce an additional newline in the output. Removing those lines from the test matching should not cause issues in the markup.


isoos:
> This also ignores unexpected blank lines in the middle of the output.

Yes. However, I think, this should not really matter for most practical purposes for us.

Issue #1409, opened 2018-06-25
issue with notifications


isoos:
- In some cases notifications are not sent (https://github.com/dart-lang/pub-dartlang-dart/issues/1399).
- In staging I've tested the same, and while the notification was received, it was not processed (Job was not updated).

PR #1410, opened 2018-06-25
Use JobBackend instead of HTTP-based notification trigger.

isoos:
#1409

The direct use of the `JobBackend` was long overdue, the http-based notification predates it and it is less reliable. It also allows a cleanup in the isolate initialization and handling, as less coordination is needed at that point.


isoos:
> Is there any documentation for how this job system works?

Unfortunately no, but I've added a tracking for it:
https://github.com/dart-lang/pub-dartlang-dart/issues/1417

sortie:
sgtm

Issue #1368, opened 2018-06-14
SEO for dartdoc content


isoos:
- [x] `sitemap.txt` should contain links to the `/documentation/[pkg]/latest/` page.
- [x] Redirect to the `/latest/` url if the `/version/` in the url is the latest.
- [x] Set `noindex` header in non-latest versions.
- [x] Link to the latest documentation in non-latest versions.
- [x] In latest packages, set `noindex` only for top level files (library, class).

isoos:
Considering the pros and cons, I think we should remove a lot of pages from the search index, as most of them provide low information density and I fear that the overall site quality would be affected by it.

For that effect, we should use `noindex` by default, and enable indexing only if:
- the package is the latest, and
- the page is one of these:
   - documentation index, or
   - library page, or
   - class page.
 
Non-latest packages should have a link to the latest version of the documentation (there is a good place for it in the footer), and all of them should be `noindex`.

mit-mit:
@isoos can update the checklist at the top with what was actually completed, please?

And can we resolve this issue?

PR #1414, opened 2018-06-26
Set documentation link to the canonical doc root.

isoos:
#1368 

Otherwise each page would point to the `/documentation/pkg/version/index.html`, which we don't want.


isoos:
I've also updated the customization test, as we'll have more of the latest vs. non-latest customization, but keeping only a very simplified diff around.

PR #1419, opened 2018-06-28
Visual update of scores on analysis tab.

isoos:
Changed to "progress bar"-like visual layout, intentionally with muted colors (choosing colors is hard). Preview at the end of the description, and also deployed to staging, a few examples: [100%](https://dartlang-pub-dev.appspot.com/packages/firebase_auth#-analysis-tab-), [77%](https://dartlang-pub-dev.appspot.com/packages/executor#-analysis-tab-), [50%](https://dartlang-pub-dev.appspot.com/packages/ecb_fx#-analysis-tab-), [22%](https://dartlang-pub-dev.appspot.com/packages/riak_client#-analysis-tab-), [outdated](https://dartlang-pub-dev.appspot.com/packages/riak_client/versions/0.5.1#-analysis-tab-).

- Removed a few headers that were not useful ("Analysis", "Scores").
- Moved the "We analyzed this package" below the scores.
- Moved the "Learn more.." label to the right and made is smaller.
- The colors use interpolation/shading: lower values are darker, and the overall score matches the color of the score circle.

This could go in multiple directions:
-  (1)The bars could be of different colors (e.g. popularity - red, code health - green, maintenance - yellow, documentation - blue), or:
- (2) The bars could be the same, following a consistent pattern (low values: red, middle: yellow/orange, high values: green).

In the first case, we could restructure the score circle to have similar color indicator bars, in the second I believe the circle should match the color of the overall score's color here.

/cc @mit-mit @sortie: how shall we balance the visuals with the new icons?

screen shot 2018-06-28 at 16 43 59


mit-mit:
I really like the bar-type visualization! However, the bars themselves seems a bit too "fat", and I think would benefit from a more light visualization. I also don't think we use rounded corners anywhere else, so that looks a bit off.

I think the overall score bar should be colored to reflect the score, like in thee examples, but also think we can use the same color for the other three bars.

isoos:
>I think would benefit from a more light visualization

Any suggestion on the color (or combination of them)?

isoos:
@mit-mit PTAL. Attached two image, and currently deploying to staging (should be available in ~10-20 minutes).

screen shot 2018-07-02 at 10 18 28

screen shot 2018-07-02 at 10 18 35

mit-mit:
Can you try the slimmer bar layout I mentioned above? Something similar to what Material specs:
https://materialdoc.com/components/progress/#linear

That would also allow us to use a regular font for the score (it would be placed either above or after the bar)

mit-mit:
Oh, and a nit: Using the theme (blue) color for the tab selection -- the blue line under (22) -- looks odd now that we use use colors for the score. Can we try to use a black/grayscale color for that?

isoos:
> Can you try the slimmer bar layout I mentioned above? 

@mit-mit: I was already considering [these pointers](https://www.jqueryscript.net/images/Responsive-Animated-Progress-Bar-With-jQuery-CSS3-Barfiller-js.jpg), but there is a question with them: if the score is high, the label above the progressbar can fit in the same line as the score title. However, if the value is low, it will either:
- push the entire bar down (line 1: "Popularity", line 2: "12", line 3: "progress bar 12% filled"), or
- the label will below it (line 1: "Popularity", line 2: "progress bar 12% filled", line 3: "12"), or
- the the labels can be in the same line, and the progress bar will only start after the title.

I couldn't really decide on those, that's why I've opted for the current bar. Any preference on the above?

isoos:
What do you think about this one?

screen shot 2018-07-02 at 20 54 50

mit-mit:
That looks great!

One small tweak: I think we should remove the small lines you have at each end of the bar. The Material spec doesn't use those.

mit-mit:
That looks great, design LGTM

@sortie can you review the code?

app/lib/frontend/color.dart
```diff
@@ -0,0 +1,98 @@
...
+final _blackShadow = Color.black.change(a: 0.5);
+final _scoreBoxMissing = new Color(204, 204, 204);
+final _scoreBoxSolid = new Color(1, 117, 194);
+final _scoreBoxGood = new Color(0, 196, 179);
+final _scoreBoxRotten = new Color(187, 36, 0);
+
+final _defaultBrush = new Brush(
+    background: _scoreBoxMissing, color: Color.white, shadow: _blackShadow);
+
+Brush overallScoreBrush(double score) {
```
sortie: Perhaps add some dart doc. The above is plain old data but now we're in logic code.

app/lib/frontend/templates.dart
```diff
@@ -420,6 +418,34 @@ class TemplateService {
     return values;
   }
 
+  Map<String, dynamic> _renderScoreBars(AnalysisExtract extract) {
+    String renderScoreBar(double score, Brush brush) {
+      return _renderTemplate('pkg/score_bar', {
+        'percent': _formatScore(score ?? 0.0),
```
sortie: A skipped package will have a 0 score rather than a N/A score?

app/lib/frontend/templates.dart
```diff
@@ -835,10 +861,12 @@ String _getAuthorsHtml(List<String> authors) {
   }).join('<br/>');
 }
 
+bool _isAnalysisSkipped(AnalysisStatus status) =>
```
sortie: Maybe skipped is the wrong word here? I was thinking of packages that are brand new (no data yet), or there's a minor problem why the data isn't there right now; not so much that it's packages that are obsolete.

app/lib/frontend/color.dart
```diff
@@ -0,0 +1,98 @@
...
+    );
+  }
+
+  @override
+  String toString() => a == 1.0
+      ? 'rgb($r, $g, $b)'
+      : 'rgba($r, $g, $b, ${a.toStringAsFixed(4)})';
+}
+
+class Brush {
```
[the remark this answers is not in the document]
isoos: I think this file should become a go-to place for all color-related code, I don't intend to introduce a full-blown color-handling library for other clients. Adding dartdoc to make the intent better on these classes.

app/lib/frontend/templates.dart
isoos: The `percent` will be used only to render the width of the progress bar, while the `score` one line below will be used to render the actual text label. Btw. great catch, because I haven't used the `score` to do so.

app/lib/frontend/templates.dart
isoos: We don't have an `AnalysisStatus` value for the pending ones (e.g. the one that is scheduled but not done yet), only for the ones that have been done. This way if a status is missing (==null), then it is by default pending (and should be scheduled), otherwise it is done.

We 'skip' two kinds of packages in the analysis: one that is older than two years, and one that is flagged as discontinued. We shall analyse everything else.

Ideally https://github.com/dart-lang/pub-dartlang-dart/issues/1352 could also track the scheduled but not done yet status too.

PR #1421, opened 2018-06-29
Upgrade Dart and Flutter SDK versions

isoos:
The SDK upgrade is high risk now, so I'm keeping the changes to the bare minimum, and only added Flutter SDK upgrade to the mix. Deploying to staging soon to validate.


isoos:
The deploy was unsuccessful, even with couple of additional patch releases. I'm closing this for now, will submit the fixes I've found so far in a separate PR and will aim for dev.66 upgrade.

PR #1424, opened 2018-07-02
Greyscaled tab indicator.

isoos:
screen shot 2018-07-02 at 13 54 09

Black: active tab
Gray: on-hovering

/cc @mit-mit

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -858,7 +858,8 @@ pre > code {
     list-style: none
 }
 
-.package-tabs > .tab{
+.package-tabs > .tab {
+    color: #666666;
     display: inline-block;
     margin-right: 35px;
     padding: 10px 2px;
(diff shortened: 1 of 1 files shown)


mit-mit:
It looks a bit odd with the blue text and black highlight bar. Can you try to make the text black?

isoos:
>  Can you try to make the text black?

The text is already black, but we can make the rest of the tabs lighter, and the selected one bold. What about this?

screen shot 2018-07-02 at 14 05 51

isoos:
Although the bolding makes the line a bit to "move" on tab switching, and it is not bad without bold either:

screen shot 2018-07-02 at 14 08 57

mit-mit:
Last example (black, no bold) LGTM!

isoos:
> The indention style looks a tiny bit inconsistent in this stylesheet.

It is, as it is a mixture of an old style, a new design which was handed to us, and our additional updates. I'm trying to make it more consistent (two tabs), but looks like I've missed it this time, will fix.

PR #1428, opened 2018-07-03
Use extracted dartdoc texts in search.

isoos:
#1179

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -92,7 +92,14 @@ class SearchBackend {
       final double popularity = popularityStorage.lookup(pv.package) ?? 0.0;
 
       final List<int> pubDataContent = pubDataContents[i];
-      final apiDocPages = _apiDocPagesFromPubData(pubDataContent);
+      List<ApiDocPage> apiDocPages;
+      if (pubDataContent != null) {
+        try {
+          apiDocPages = _apiDocPagesFromPubData(pubDataContent);
+        } catch (e, st) {
+          _logger.severe('Parsing pub-data.json failed.', e, st);
+        }
+      }
 
       results[i] = new PackageDocument(
         package: pv.package,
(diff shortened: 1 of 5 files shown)


app/lib/search/backend.dart
```diff
@@ -135,41 +136,60 @@ class SearchBackend {
...
+          relativePath: path,
+          symbols: symbols,
+          textBlocks: docMap[key],
+        );
       }).toList();
       results.sort((a, b) => a.relativePath.compareTo(b.relativePath));
       return results;
     } catch (e, st) {
-      _logger.warning('Parsing dartdoc index.json failed.', e, st);
+      _logger.warning('Parsing pub-data.json failed.', e, st);
```
sortie: If pub-data.json is made by us and not by users, then that sounds like an error to me (we'll certainly want to know about it). What makes this case plausible? The return null can be moved into the catch body.

Maybe the caller should have this try block instead of wrapping the whole function? This function also doesn't know the data came from 'pub-data.json'.

app/lib/search/backend.dart
```diff
@@ -70,15 +71,15 @@ class SearchBackend {
         versionList.where((pv) => pv != null),
         key: (pv) => (pv as PackageVersion).package);
 
-    final indexJsonFutures = Future.wait(packages.map((p) =>
-        dartdocClient.getContentBytes(p.name, 'latest', 'index.json',
+    final pubDataFutures = Future.wait(packages.map((p) =>
+        dartdocClient.getContentBytes(p.name, 'latest', 'pub-data.json',
```
sortie: Does pub-data.json already exist?

isoos: Prod: not yet, but it will with the next release.

app/lib/search/backend.dart
isoos: The parse error could happen if we change the format in a breaking way, and an older client reads the new format, or a newer client reads the old format, and they are not prepared for some kind of new structure.

PR #1438, opened 2018-07-09
Adjust health score and suggestions label

isoos:


app/lib/frontend/templates.dart
```diff
@@ -230,6 +230,8 @@ class TemplateService {
         'suggestion_help_html': getSuggestionHelpMessage(suggestion.code),
       };
     })?.toList();
+    final hasIssues = suggestions.isNotEmpty &&
+        analysis.suggestions.any((s) => s.isError || s.isWarning);
```
sortie: .any is false for the empty array, so suggestions.isNotEmpty isn't needed.

app/lib/shared/analyzer_client.dart
```diff
@@ -232,7 +233,17 @@ class AnalysisView {
...
-  double get health => _summary?.fitness?.healthScore ?? 0.0;
+  double get health {
+    final oldFitnessScore = _summary?.fitness?.healthScore ?? 0.0;
+    if (_summary?.maintenance == null) {
+      return oldFitnessScore;
+    }
+    final newFitnessScore = 1.0 *
+        math.pow(0.8, _summary.maintenance.errorCount ?? 0) *
+        math.pow(0.95, _summary.maintenance.warningCount ?? 0) *
+        math.pow(0.999, _summary.maintenance.hintCount ?? 0);
```
sortie: Has this been tested with absurdly large exponents? It should give 0, so it should be safe.

It would be interesting to know how how the old and new scores compare, and how often the new score is higher than the old one. Can we log/store pairs of these values ("$oldFitnessScore -> $newFitnessScore") so we can do some number crunching?

isoos: Large exponents (tested with 0.999^1000000000) will give `0.0`.

I'll soon (~ in an hour or so) send a simple report on the change these numbers give us, and also two alternatives.

isoos:
@sortie PTAL. I've updated the scores to be a bit stricter and also added a quick patch for https://github.com/dart-lang/pana/issues/369

PR #1446, opened 2018-07-12
Increase dartdoc init timeout.

isoos:


app/lib/shared/service_utils.dart
```diff
@@ -224,8 +224,7 @@ void _wrapper(List fnAndMessage) {
 Future initDartdoc(Logger logger) async {
   logger.info('Initializing pkg/pub_dartdoc');
   final dir = Platform.script.resolve('../../pkg/pub_dartdoc').toFilePath();
-  final pr = await Process.run('pub', ['get'], workingDirectory: dir)
-      .timeout(const Duration(minutes: 1));
+  final pr = await runProc('pub', ['get'], workingDirectory: dir, timeout: const Duration(minutes: 2));
   if (pr.exitCode != 0) {
     final message = 'Failed to initialize pkg/pub_dartdoc';
     logger.shout(message);
```
sortie: Can we also log $pr here, so we have data on what happened? Does runProc modify pr so it mentions timing out?

isoos: I don't think there is a mention, but exitCode will be 15 == sigkill. I'm adding it to the log message.

PR #1452, opened 2018-07-13
Catch oath2 request issues.

isoos:


app/lib/frontend/oauth2_service.dart
```diff
@@ -36,9 +40,11 @@ class OAuth2Service {
     try {
       final api = new oauth2_v2.Oauth2Api(authClient);
       info = await api.userinfo.get();
+    } catch (e, st) {
+      _logger.log(Level.INFO, 'Bearer token lookup failed.', e, st);
```
sortie: Can we catch a known exception, such as access denied? Then we can log that as info, while unknown exceptions can be logged as errors / warnings?

app/lib/frontend/oauth2_service.dart
```diff
@@ -36,9 +40,11 @@ class OAuth2Service {
...
     } finally {
       authClient.close();
     }
-    return info.email;
+    return info?.email;
```
sortie: This could be moved into the try block, and the catch clause could return null.

isoos:
PTAL

app/lib/frontend/oauth2_service.dart
```diff
@@ -36,9 +40,15 @@ class OAuth2Service {
     try {
       final api = new oauth2_v2.Oauth2Api(authClient);
       info = await api.userinfo.get();
+      return info.email;
+    } on AccessDeniedException catch (e, st) {
+      _logger.log(
+          Level.INFO, 'Bearer token lookup failed with access denied.', e, st);
```
sortie: I would rephrase that as 'Access denied for bearer token'. What's a bearer token btw, maybe mention oauth or something?

PR #1453, opened 2018-07-13
Don't try to serve documentation API for redirected packages.

isoos:


isoos:
> Could we redirect to the proper location instead?

This is our `/api/documentation` which serves info to the package page (to display whether a documentation has failed or is ready) and maybe others are using it. When you access it through `/documentation/[package]` it will redirect.

PR #1458, opened 2018-07-13
Tool to update the runtimeVersion of the Job table.

isoos:
Just in case...

### DIFF
--- a/app/bin/tools/update_job_runtime.dart
+++ b/app/bin/tools/update_job_runtime.dart
@@ -1,0 +1,63 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:args/args.dart';
+import 'package:gcloud/db.dart';
+import 'package:pool/pool.dart';
+import 'package:pub_semver/pub_semver.dart';
+
+import 'package:pub_dartlang_org/frontend/service_utils.dart';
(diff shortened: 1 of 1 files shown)


app/bin/tools/update_job_runtime.dart
```diff
@@ -0,0 +1,63 @@
...
+import 'package:pub_dartlang_org/job/backend.dart';
+
+Future main(List<String> args) async {
+  final ArgParser parser = new ArgParser()
+    ..addOption('runtime-version', help: 'The runtime version to set.')
+    ..addOption('concurrency',
+        defaultsTo: '10', help: 'The concurrency of the updates.')
+    ..addFlag('reset',
+        defaultsTo: false, help: 'Trigger re-processing of the jobs.');
+  final ArgResults argv = parser.parse(args);
```
sortie: argv is the argument vector itself, what you call args above. This is more accurately called 'options'.

app/bin/tools/update_job_runtime.dart
```diff
@@ -0,0 +1,63 @@
...
+  final String runtimeVersion = argv['runtime-version'];
+  final int concurrency = int.parse(argv['concurrency'] as String);
+  final bool reset = argv['reset'];
+
+  if (runtimeVersion == null || runtimeVersion.isEmpty) {
+    throw new Exception('--runtime-version must be set');
+  }
+  // check whether it can be parsed
+  new Version.parse(runtimeVersion);
```
sortie: Does the empty runtime version parse? Otherwise --runtime-version was set, but the empty value is just another invalid runtime version.

isoos: Checking it in the previous block.

PR #1462, opened 2018-07-16
Upgrade versions

isoos:
Brings in the fixed `appengine` and `http2` packages.


isoos:
Deployed to staging yesterday, it looks good.

PR #1463, opened 2018-07-18
Upgrade dependencies for pkg/_popularity.

isoos:


isoos:
I've merged this because it shouldn't affect the current latency issue.

PR #1465, opened 2018-07-18
Add RSS link to the footer.

### LINKED ISSUE #1426: Add RSS/Atom Feed link to Pub web page
To make it easier to subscribe to the RSS feed, please add a link from the pub web page.

A standard feed icon would be a plus too.

Much more user friendly than doing 'View Source'.

Thanks!

isoos:
Fixes #1426

The footer becomes the following:

screen shot 2018-07-18 at 11 14 54

It is subtle, yet if somebody is looking for it they will find it. I opted to not include any text, because "RSS" is the common term, but we emit the "Atom" format...


isoos:
Merging this because it shouldn't affect the current latency issue.

PR #1489, opened 2018-07-25
Fix anchor handling and scrolling

isoos:
- switch to a different tab if the anchor is there (e.g. `#100` could switch to the changelog's tab and scroll to the version header of `1.0.0`)
- scroll was jumpy in many cases, fixing it

https://github.com/dart-lang/pub-dartlang-dart/issues/1429

### DIFF
--- a/static/js/script.dart
+++ b/static/js/script.dart
@@ -220,11 +220,15 @@ void _setEventForHashChange() {
 
 Future _scrollTo(Element elem) async {
   final int stepCount = 30;
-  final int offsetTop = elem.offsetTop - 24;
-  final int scrollY = window.scrollY;
-  final int diff = offsetTop - scrollY;
   for (int i = 0; i < stepCount; i++) {
     await window.animationFrame;
+    final int offsetTop = elem.offsetTop - 12;
+    final int scrollY = window.scrollY;
+    final int diff = offsetTop - scrollY;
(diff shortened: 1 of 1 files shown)


mkustermann:
I'm not very familiar with client things. Moving to @kevmoo

kevmoo:
I'm not very familiar, either. I'd love to see this staged to understand what it does...

isoos:
Deployed to staging. Try the following links, switching back and forth by rewriting the address:
- https://dartlang-pub-dev.appspot.com/packages/w_flux#100
- https://dartlang-pub-dev.appspot.com/packages/w_flux#store

sortie:
Why are we messing with scrolling instead of letting the browser do this on its own, like browsers have always done?

isoos:
> Why are we messing with scrolling instead of letting the browser do this on its own, like browsers have always done?

Two reasons:

1. There is a styling thing where in some cases the browser puts the label partially out of the view after jumping to it. That's the reason we have a 12px offset on each anchor: https://github.com/dart-lang/pub-dartlang-dart/blob/master/static/js/script.dart#L225

2. There was an edge case where we needed to capture the event and needed to do manual scrolling. 

Maybe both of these issues have been only with the old design, and the new one is no longer affected. I have no recall when these originally happened and whether they still apply.

PR #1498, opened 2018-08-02
Implement ScoreCard's model and backend

isoos:
Many steps closer to https://github.com/dart-lang/pub-dartlang-dart/issues/1352

- `ScoreCardData` will replace `AnalysisExtract` with efficient storage and caching.
- `ScoreCard` stores a lot of relevant information that is required for replacing `Job` scheduling.
- `ScoreCardReport` will replace `Analysis` and at the same time will allow `dartdoc` and other reports reports to be stored and handled the same way.

### DIFF
--- a/app/lib/scorecard/backend.dart
+++ b/app/lib/scorecard/backend.dart
@@ -7,6 +7,7 @@ import 'dart:async';
 import 'package:gcloud/db.dart' as db;
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
 
 import '../frontend/models.dart' show Package, PackageVersion;
 import '../shared/popularity_storage.dart';
@@ -15,6 +16,7 @@ import '../shared/versions.dart' as versions;
 
 import 'helpers.dart';
 import 'models.dart';
+import 'scorecard_memcache.dart';
 
 export 'models.dart';
 
@@ -32,12 +34,28 @@ class ScoreCardBackend {
   final db.DatastoreDB _db;
   ScoreCardBackend(this._db);
 
-  Future<ScoreCard> latestScoreCard(
-      String packageName, String packageVersion) async {
+  Future<ScoreCardData> getScoreCardData(
+    String packageName,
+    String packageVersion, {
+    @required bool onlyCurrent,
+  }) async {
+    final cached = await scoreCardMemcache.getScoreCardData(
+        packageName, packageVersion, versions.runtimeVersion,
+        onlyCurrent: onlyCurrent);
+    if (cached != null) {
+      return cached;
+    }
+
     final key = scoreCardKey(packageName, packageVersion);
     final currentList = await _db.lookup([key]);
     if (currentList.first != null) {
-      return currentList.first as ScoreCard;
+      final data = (currentList.first as ScoreCard).toData();
+      await scoreCardMemcache.setScoreCardData(data);
+      return data;
+    }
+
+    if (onlyCurrent) {
+      return null;
     }
 
     final query = _db.query(ScoreCard, ancestorKey: key.parent)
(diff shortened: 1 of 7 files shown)


isoos:
@sortie this has been updated with the latest 2.0 changes, and can be reviewed again

app/lib/scorecard/backend.dart
```diff
@@ -0,0 +1,178 @@
...
+        .run()
+        .where((sc) =>
+            // sanity check to not rely entirely on the lexicographical order
+            isNewer(sc.semanticRuntimeVersion, versions.semanticRuntimeVersion))
+        .toList();
+    if (all.isEmpty) {
+      return null;
+    }
+    all.sort((a, b) =>
+        isNewer(a.semanticRuntimeVersion, b.semanticRuntimeVersion) ? -1 : 1);
```
sortie: It's faster to search for the latest rather than sort and then take the last.

app/lib/scorecard/backend.dart
```diff
@@ -0,0 +1,178 @@
...
+
+/// Sets the active scorecard backend.
+void registerScoreCardBackend(ScoreCardBackend backend) =>
+    ss.register(#_scorecard_backend, backend);
+
+/// The active job backend.
+ScoreCardBackend get scoreCardBackend =>
+    ss.lookup(#_scorecard_backend) as ScoreCardBackend;
+
+class ScoreCardBackend {
```
sortie: Where is the documentation on score cards? Can add you add some words on the data model and the operations done here?

Should there be tests for this code and score cards?

app/lib/scorecard/backend.dart
```diff
@@ -0,0 +1,178 @@
...
+      }
+      tx.queueMutations(inserts: [report]);
+      await tx.commit();
+    });
+  }
+
+  Future<Map<String, ReportData>> loadReports(
+      String packageName, String packageVersion,
+      {List<String> reportTypes}) async {
+    reportTypes ??= [ReportType.pana, ReportType.dartdoc];
```
sortie: These are static const, can you set a default parameter value for reportTypes instead, if the list can be const?

app/lib/scorecard/models.dart
```diff
@@ -6,18 +6,42 @@ import 'dart:convert';
...
+  static const String dartdoc = 'dartdoc';
+}
+
+abstract class ReportStatus {
+  static const String success = 'success';
+  static const String failed = 'failed';
+  static const String aborted = 'aborted';
+}
+
 /// Summary of various reports for a given PackageVersion.
```
sortie: This documentation seems a little sparse?

app/lib/scorecard/models.dart
```diff
@@ -57,6 +75,18 @@ class ScoreCard extends db.ExpandoModel {
   @db.DoubleProperty()
   double popularityScore;
 
+  /// The platform tags (flutter, web, other).
+  @CompatibleStringListProperty()
+  List<String> platformTags;
+
+  /// The flags for the package, version or analysis.
```
sortie: Maybe list a few?

app/lib/scorecard/models.dart
```diff
@@ -117,4 +200,150 @@ class ScoreCardReport extends db.ExpandoModel {
...
+
+  final double healthScore;
+
+  final double maintenanceScore;
+
+  /// The platform tags (flutter, web, other).
+  @CompatibleStringListProperty()
+  List<String> platformTags;
+
+  final String platformReason;
```
sortie: What is this?

isoos:
I've updated the documentation, but looking at it after a month or so, I've found the field duplicates in the `ScoreCard` vs `ScoreCardData` a bit confusing. Maybe `ScoreCardData` should be stored as a serialized binary in `ScoreCard` like we do with the `ReportData`? That way the fields in the DataStore records won't be duplicates (it is unlikely that we'd want to filter any of these, as they would also require the `runtimeVersion` in the id and in the index).

sortie:
lgtm

PR #1513, opened 2018-08-09
Fast fail for dartdoc if the pubspec supports only legacy SDK.

isoos:
Part of https://github.com/dart-lang/pub-dartlang-dart/issues/1504

### DIFF
--- a/app/lib/frontend/model_properties.dart
+++ b/app/lib/frontend/model_properties.dart
@@ -75,8 +75,9 @@ class Pubspec {
     return _asString(environment['sdk']);
   }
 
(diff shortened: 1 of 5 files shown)


app/lib/frontend/model_properties.dart
```diff
@@ -73,7 +75,11 @@ class Pubspec {
     return _asString(environment['sdk']);
   }
 
-  VersionConstraint get sdkVersionConstraint => _inner.environment['sdk'];
```
sortie: Why is this getter removed? Seems useful?

PR #1525, opened 2018-08-10
More insight into the Job processing stats.

isoos:
- `total` keeps the total count
- `owned` keeps the count of jobs that are in the scope of the current version (accounts for the delay in the scheduling phase)
- the rest is filtered only for the "owned" jobs

### DIFF
--- a/app/lib/job/backend.dart
+++ b/app/lib/job/backend.dart
@@ -322,6 +322,7 @@ class _Stat {
   final bool _collectFailed;
   final _failedPackages = new Set<String>();
   int _totalCount = 0;
+  int _ownedCount = 0;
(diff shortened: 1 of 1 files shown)


app/lib/job/backend.dart
```diff
@@ -331,6 +332,11 @@ class _Stat {
 
   void add(Job job) {
     _totalCount++;
```
sortie: Why do we count jobs in the total if they have the wrong runtime version?

isoos: Sanity check, in case some counter is 0, we can see that it is indeed counted some `total` number of rows, but the filtering didn't match any of it.

PR #1526, opened 2018-08-10
Collect statistics about keys in pubspec.yaml files.

isoos:


app/bin/tools/pubspec_stats.dart
```diff
@@ -0,0 +1,36 @@
...
+
+    int index = 0;
+    final query = dbService.query(PackageVersion);
+    await for (PackageVersion pv in query.run().cast()) {
+      index++;
+      if (index % 100 == 0) {
+        print('$index - ${pv.package} ${pv.version}');
+      }
+      pv.pubspec.asJson.keys.forEach((key) {
+        final k = key.toString();
```
sortie: k vs key is a bit bad naming but it's alright

PR #1541, opened 2018-08-22
Upgrade runtime SDK to 2.1-dev

isoos:
- updated generated code
- updated API changes for gcloud/appengine
- upgraded Flutter to 0.6.2, dartdoc to 0.20.3, pana to 0.11.9

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Keep version in-sync with .travis.yml, .mono_repo.yml and app/lib/shared/versions.dart
-FROM google/dart-runtime-base:2.0.0-dev.63.0
+FROM google/dart-runtime-base:2.1.0-dev.1.0
 
 # `apt-mark hold dart` ensures that Dart is not upgraded with the other packages
 #   We want to make sure SDK upgrades are explicit.
--- a/app/analysis_options.yaml
+++ b/app/analysis_options.yaml
@@ -1,7 +1,5 @@
 # https://www.dartlang.org/guides/language/analysis-options
 analyzer:
-  language:
-    enablePreviewDart2: true
   strong-mode:
     implicit-casts: false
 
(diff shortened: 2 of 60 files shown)


isoos:
I was able to remove the larger timeout, but the parallelism is still limited at 1, I'm not sure why it fails more when it is run in parallel.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Keep version in-sync with .travis.yml, .mono_repo.yml and app/lib/shared/versions.dart
```
sortie: Can we reduce the number of sources of truth?

app/analysis_options.yaml
```diff
@@ -43,7 +41,8 @@ linter:
     - package_prefixed_library_names
 #    - parameter_assignments
     - prefer_final_fields
-    - prefer_final_locals
+# TODO: enable after generated code passes
```
sortie: Is a bug filed? How about the other commented out elements here?

app/pubspec.yaml
```diff
@@ -2,21 +2,21 @@ name: pub_dartlang_org
 author: Dart Team <misc@dartlang.org>
 description: The pub.dartlang.org website.
 environment:
-  sdk: ">=1.25.0-dev.9.0 <2.0.0"
+  sdk: ">=2.0.0 <2.2.0"
```
sortie: 2.2.0 hasn't even been released yet.

Did you mean <3.0.0?

app/pubspec.yaml
```diff
@@ -34,13 +34,13 @@ dependencies:
   yaml: '^2.1.12'
   # pana version to be pinned
   # keep in-sync with app/lib/shared/versions.dart
-  pana: '0.11.8'
+  pana: '0.11.9'
```
sortie: Pana bump is intentional?

isoos: Yes, it is a must, otherwise versions don't resolve.

Dockerfile
isoos: I'm not sure about `.mono_repo.yaml` (if it could infer things), but Dockerfile and `.travis.yaml` seem to be independent of anything and can't take it as an external variable.

app/analysis_options.yaml
isoos: https://github.com/dart-lang/json_serializable/issues/290

I think we haven't really revised the other elements in ages, it would be worth to take a fresh look (e.g. check if new rules could be applied).

PR #1550, opened 2018-08-23
Make redirect to pub site available for all handlers.

isoos:
Part of https://github.com/dart-lang/pub-dartlang-dart/issues/1534

### DIFF
--- a/app/lib/shared/handler_helpers.dart
+++ b/app/lib/shared/handler_helpers.dart
@@ -28,6 +28,7 @@ Future runHandler(Logger logger, shelf.Handler handler,
     handler = _sanitizeRequestWrapper(handler);
   }
   handler = _redirectToHttpsWrapper(handler);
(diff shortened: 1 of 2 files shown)


app/lib/shared/handler_helpers.dart
```diff
@@ -85,6 +86,20 @@ shelf.Handler _userAuthParsingWrapper(shelf.Handler handler) {
   };
 }
 
+shelf.Handler _redirectToPubSiteWrapper(shelf.Handler handler) {
+  return (shelf.Request request) async {
+    final host = request.requestedUri.host;
+    final shouldRedirect =
+        (host == 'www.dartdocs.org') || (host == 'dartdocs.org');
```
sortie: ()'s are not needed and don't add clarity.

isoos:
> But seems you dropped this CL?

It was less work to make the other CL work (with tests) than fix these tests.

PR #1551, opened 2018-08-23
Serve /documentation URLs through default service (frontend).

isoos:
Part of https://github.com/dart-lang/pub-dartlang-dart/issues/1534

- tests and the redirect logic is copied almost verbatim
- the serving handler is referenced
- tweak added to preserve the original request after sanitization, and work on that if it is in the context


isoos:
> Though I am not quite certain how this stops redirecting to the dartdoc instance and serves itself instead.

Oh, true, that's the `dispatch.yaml`. I'm not yet sure if we should remove it (but then how to deploy it), make it empty, or override the `dartdoc` references to `default` there.

Issue #1553, opened 2018-08-27
Search could be more 'fuzzy'


mit-mit:
The search term is matched as-typed to the title, so for example a search for 'location' as an attempt to get packages that perform geolocation ends up showing a bunch of poorly scored packages:
https://pub.dartlang.org/packages?q=location

I think we should include hits in the package description and API surface to a higher degree.

isoos:
On that search result page, which packages should be promoted up/down? Maybe it is not necessary the description or API surface that needs to be promoted more, rather the exact word matches vs. part-of-the-bigger-word matches?

mit-mit:
I would have expected `geolocated` with score 96 and `firestore_helpers` with score 90 to be higher in the sorted results, given their high scores and the fact that they have `location` in their descriptions.

isoos:
I've tracked this down to an implementation detail of our `TokenIndex.lookupTokens`, which uses Inverse Document Frequency to weight tokens that we look up. In this specific case, it will weight the term `geolocation` about 60% more than the exact query `location`, because the former is less frequently used in the corpus of text that are in the index.

I'm a bit undecided how to fix that, will need to experiment a bit with different tuning.

isoos:
FYI: The above change has been landed and deployed, but the result page didn't improve too much. An internal detail shows that the text-only search has very low score for e.g. `firestore_helpers`, I'm going to debug it further why it can be the case.

isoos:
The change has been deployed, and the results seem to be a bit better. wdyt @mit-mit ?

mit-mit:
Yes, this looks OK now, closing
https://pub.dartlang.org/flutter/packages?q=location&api=0

PR #1558, opened 2018-08-28
Remove IDF weight from token weights.

isoos:
- fixes the `location` vs. `geolocation` issue in https://github.com/dart-lang/pub-dartlang-dart/issues/1553
- other search results seem to be fine with the change


sortie:
Okay, I understand a bit better after reading #1553 more carefully.

PR #1569, opened 2018-08-30
Don't use dartdoc (HTTP) client in frontend.

isoos:
As we are serving dartdoc files from the frontend, there is no reason to query the dartdoc service from there. I think the other services could also need to get migrated to direct backend use.


isoos:
> What other services?

`search` uses it to get the API/dartdoc content for indexing.
`analyzer` uses it to get the dartdoc status that will be included in the analysis results. (should be replaced once the scorecard model gets done)

PR #1581, opened 2018-09-05
Generate, index and display Dart SDK API results in pub search.

isoos:
https://github.com/dart-lang/pub-dartlang-dart/issues/181

### DIFF
--- a/app/bin/service/search.dart
+++ b/app/bin/service/search.dart
@@ -80,9 +80,9 @@ Future _main(FrontendEntryMessage message) async {
     registerPackageIndex(new SimplePackageIndex());
     registerTaskSendPort(taskReceivePort.sendPort);
 
-    // don't block on SDK index updates, as it may take several minutes before
+    // Don't block on SDK index updates, as it may take several minutes before
     // the dartdoc service produces the required output.
-    _updateDartSdkIndex();
+    _updateDartSdkIndex().whenComplete(() {});
 
     final BatchIndexUpdater batchIndexUpdater = new BatchIndexUpdater();
     await batchIndexUpdater.initSnapshot();
--- a/app/bin/service/search.dart
+++ b/app/bin/service/search.dart
@@ -122,7 +122,7 @@ Future _updateDartSdkIndex() async {
       _logger.warning('Error loading Dart SDK index.', e, st);
     }
     if (i % 10 == 0) {
-      _logger.warning('Unable to load Dart SDK index. Cycle: $i');
+      _logger.warning('Unable to load Dart SDK index. Attempt: $i');
     }
     await new Future.delayed(const Duration(minutes: 1));
   }
--- a/app/bin/service/search.dart
+++ b/app/bin/service/search.dart
@@ -11,6 +11,7 @@ import 'package:gcloud/service_scope.dart';
 import 'package:gcloud/storage.dart';
 import 'package:logging/logging.dart';
 
+import 'package:pub_dartlang_org/dartdoc/backend.dart';
 import 'package:pub_dartlang_org/shared/analyzer_client.dart';
 import 'package:pub_dartlang_org/shared/analyzer_memcache.dart';
 import 'package:pub_dartlang_org/shared/configuration.dart';
@@ -23,6 +24,8 @@ import 'package:pub_dartlang_org/shared/service_utils.dart';
 import 'package:pub_dartlang_org/shared/task_client.dart';
 import 'package:pub_dartlang_org/shared/task_scheduler.dart';
 import 'package:pub_dartlang_org/shared/task_sources.dart';
+import 'package:pub_dartlang_org/shared/versions.dart';
+import 'package:pub_dartlang_org/shared/urls.dart';
 
 import 'package:pub_dartlang_org/search/backend.dart';
 import 'package:pub_dartlang_org/search/handlers.dart';
@@ -56,6 +59,9 @@ Future _main(FrontendEntryMessage message) async {
     registerAnalyzerClient(analyzerClient);
     registerScopeExitCallback(analyzerClient.close);
 
+    final Bucket dartdocBucket = await getOrCreateBucket(
+        storageService, activeConfiguration.dartdocStorageBucketName);
+    registerDartdocBackend(new DartdocBackend(db.dbService, dartdocBucket));
     registerDartdocMemcache(new DartdocMemcache(memcacheService));
     final DartdocClient dartdocClient = new DartdocClient();
     registerDartdocClient(dartdocClient);
@@ -69,8 +75,14 @@ Future _main(FrontendEntryMessage message) async {
         new SnapshotStorage(storageService, snapshotBucket));
 
     final ReceivePort taskReceivePort = new ReceivePort();
+    registerDartSdkIndex(new SimplePackageIndex.sdk(
+        urlPrefix: dartSdkMainUrl(toolEnvSdkVersion)));
     registerPackageIndex(new SimplePackageIndex());
     registerTaskSendPort(taskReceivePort.sendPort);
+
+    // don't block on SDK index updates, as it may take several minutes before
+    // the dartdoc service produces the required output.
+    _updateDartSdkIndex();
 
     final BatchIndexUpdater batchIndexUpdater = new BatchIndexUpdater();
     await batchIndexUpdater.initSnapshot();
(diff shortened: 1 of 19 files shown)


isoos:
Staging is being updated with this new version, should be available in less then an hour.

app/bin/service/search.dart
```diff
@@ -69,9 +75,15 @@ Future _main(FrontendEntryMessage message) async {
...
 
     final ReceivePort taskReceivePort = new ReceivePort();
+    registerDartSdkIndex(new SimplePackageIndex.sdk(
+        urlPrefix: dartSdkMainUrl(toolEnvSdkVersion)));
     registerPackageIndex(new SimplePackageIndex());
     registerTaskSendPort(taskReceivePort.sendPort);
 
+    // don't block on SDK index updates, as it may take several minutes before
+    // the dartdoc service produces the required output.
+    _updateDartSdkIndex();
```
sortie: I believe there is an optional hint about unawaited futures. This code also doesn't look like it creates a future. How about .then((_) {}) or something to show that we do wish to ignore this background future?

app/bin/service/search.dart
```diff
@@ -92,3 +104,26 @@ Future _main(FrontendEntryMessage message) async {
...
+        await dartSdkIndex.addPackages(docs);
+        await dartSdkIndex.merge();
+        _logger.info('Dart SDK index loaded successfully.');
+        return;
+      }
+    } catch (e, st) {
+      _logger.warning('Error loading Dart SDK index.', e, st);
+    }
+    if (i % 10 == 0) {
+      _logger.warning('Unable to load Dart SDK index. Cycle: $i');
```
sortie: Maybe use the word 'Attempt' instead of 'Cycle'

app/lib/dartdoc/backend.dart
```diff
@@ -43,6 +46,60 @@ class DartdocBackend {
 
   DartdocBackend(this._db, this._storage);
 
+  /// Whether the storage bucket has a useable extracted data file.
```
sortie: usable

app/lib/dartdoc/backend.dart
```diff
@@ -43,6 +46,60 @@ class DartdocBackend {
...
+  /// Only the existence of the file is checked.
+  // TODO: decide whether we should re-generate the file after a certain age
+  Future<bool> hasValidDartSdkDartdocData() async {
+    final objectName =
+        storage_path.dartSdkDartdocDataName(shared_versions.runtimeVersion);
+    try {
+      final info = await _storage.info(objectName);
+      return info != null;
+    } catch (_) {
```
sortie: Can we catch a specific exception instead?

app/lib/dartdoc/backend.dart
```diff
@@ -43,6 +46,60 @@ class DartdocBackend {
 
   DartdocBackend(this._db, this._storage);
 
+  /// Whether the storage bucket has a useable extracted data file.
+  /// Only the existence of the file is checked.
+  // TODO: decide whether we should re-generate the file after a certain age
```
sortie: Is it redone when we bump pana or the runtime versions?

app/lib/dartdoc/backend.dart
```diff
@@ -43,6 +46,60 @@ class DartdocBackend {
...
+  /// Upload the generated dartdoc data file for the Dart SDK to the storage bucket.
+  Future uploadDartSdkDartdocData(File file) async {
+    final objectName =
+        storage_path.dartSdkDartdocDataName(shared_versions.runtimeVersion);
+    try {
+      final contentBytes = await file.readAsBytes();
+      await _storage.writeBytes(objectName, _gzip.encode(contentBytes));
+    } catch (e, st) {
+      _logger.warning(
+          'Unable to upload SDK pub dartdoc data file: $objectName', e, st);
```
sortie: Is this likely? Should it be an error? Can we catch unlikely issues here? Can we catch a more specific exception?

app/lib/dartdoc/backend.dart
```diff
@@ -43,6 +46,60 @@ class DartdocBackend {
...
+    }
+  }
+
+  /// Upload the generated dartdoc data file for the Dart SDK to the storage bucket.
+  Future uploadDartSdkDartdocData(File file) async {
+    final objectName =
+        storage_path.dartSdkDartdocDataName(shared_versions.runtimeVersion);
+    try {
+      final contentBytes = await file.readAsBytes();
+      await _storage.writeBytes(objectName, _gzip.encode(contentBytes));
```
sortie: Is there a stream way to write the bytes?

app/lib/dartdoc/backend.dart
```diff
@@ -43,6 +46,60 @@ class DartdocBackend {
...
+      return new PubDartdocData.fromJson(map);
+    }
+
+    for (int i = 0; i < 3; i++) {
+      try {
+        return await load();
+      } catch (e, st) {
+        final message =
+            'Unable to read SDK pub dartdoc data file (attempt #${i + 1}): $objectName ';
+        _logger.info(message, e, st);
```
sortie: Woah. This seems flaky. What's the problem here?

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -40,6 +40,56 @@ class DartdocJobProcessor extends JobProcessor {
...
+        '--no-validate-links',
+      ];
+      if (envConfig.toolEnvDartSdkDir != null) {
+        args.addAll(['--sdk-dir', envConfig.toolEnvDartSdkDir]);
+      }
+      final pr = await runProc(
+        'dart',
+        ['bin/pub_dartdoc.dart']..addAll(args),
+        workingDirectory: _pkgPubDartdocDir,
+        timeout: _dartdocTimeout * 2,
```
sortie: Why times 2?

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -40,6 +40,56 @@ class DartdocJobProcessor extends JobProcessor {
...
+      );
+
+      final pubDataFile = new File(p.join(outputDir, 'pub-data.json'));
+      final hasPubData = await pubDataFile.exists();
+      final isOk = pr.exitCode == 0 && hasPubData;
+      if (!isOk) {
+        _logger.warning(
+            'Error while generating SDK docs.\n\n${pr.stdout}\n\n${pr.stderr}');
```
sortie: Should this be an error? Can this be triggered by poor user code? In that case, can we recognize the difference between bad user input and an infrastructure problem?

app/lib/dartdoc/storage_path.dart
```diff
@@ -49,3 +50,13 @@ String contentObjectName(String packageName, String packageVersion, String uuid,
...
 }
+
+/// ObjectName of an SDK asset.
+String sdkObjectName(String relativePath) {
+  return p.join(_sdkAssetPrefix, relativePath);
+}
+
+/// The ObjectName for the Dart SDK's extracted dartdoc content.
+String dartSdkDartdocDataName(String runtimeVersion) {
+  return sdkObjectName('dart/pub-dartdoc-data/$runtimeVersion.json.gz');
```
sortie: Does anything garbage collect these old files or do they accumulate indefinitely? Does this mean we have to update the runtime version whenever we update the Dart SDK? Should we decouple the Dart SDK used for the API results and the Dart SDK pub's website itself uses?

app/lib/dartdoc/backend.dart
isoos: Yes, it is redone.

app/lib/dartdoc/backend.dart
isoos: On the larger scale of uploading thousands of files of the dartdoc-generated files, it does happen (request timeout, network issue, an occasional error 500 from the AppEngine service), but on a per-file basis, it may be rare. This specific file is larger then the rest, and it may happen more frequently.
I wouldn't make it a hard error, because at worst case we won't be able to search in the API docs for a limited period, which should be rather small (measured in minutes), as eventually another dartdoc instance would try to generate and upload it too.

app/lib/dartdoc/backend.dart
isoos: IIRC it was about the time where we have experienced several timeouts and 500s from the dartdoc's storage bucket both in prod and in staging. We can be more optimistic now, shall I remove?

app/lib/dartdoc/dartdoc_runner.dart
isoos: Generating docs for the SDK takes longer time, and this seemed to be a good number. I'll refactor it into a separate const.

app/lib/dartdoc/dartdoc_runner.dart
isoos: It is most likely an infrastructure problem, making it a hard error.

app/lib/dartdoc/storage_path.dart
isoos: > Does anything garbage collect these old files or do they accumulate indefinitely?

No GC yet, but there will be.

> Does this mean we have to update the runtime version whenever we update the Dart SDK? Should we decouple the Dart SDK used for the API results and the Dart SDK pub's website itself uses?

We do need to update it right now, but we could decouple the runtime SDK version from it. I'll need to think about it if there is any additional risk with the current stack, and I'll file a separate PR.

app/lib/dartdoc/backend.dart
isoos: I recall two distinct kind of error there, but one of them may be related to the outage I have mentioned earlier.

isoos: The other is `DetailedApiRequestError`, and I'm not sure where it is coming from.

app/lib/dartdoc/backend.dart
sortie: If you think it's stable? If we try once here, and it fails, is it later retried at a higher level?

app/lib/dartdoc/backend.dart
sortie: Alright sounds good

app/lib/dartdoc/backend.dart
isoos: Yes, it is retried at a higher level, removing the redundancy.

Issue #1522, opened 2018-08-10
Add an 'Issues' link below the 'Homepage' link


mit-mit:
Make it easier for package consumers to file issues.

Ideas for realization:
  * Add a new field to `pubspec.yaml`
  * If field is unset: Deduce the Issue tracker link for the case where we have a GitHub link

sortie:
Aqueduct, for instance, is hosted on github but has its homepage set to its actual homepage. We should consider adding a 'Repository' field to the `pubspec.yaml` while we're doing this, so the issue tracker can be deduced in such cases.

sortie:
The pub client could at upload time check if there is a git remote, if so, suggest the pubspec.yaml contains the right repository value.

isoos:
One note for #1675: I've prepared the issue tracker link pointing to the list of issues, and not to the "new issue page". Both GitHub and GitLab would ask the user to authenticate, and I thought that it is better UX to point to a page that won't redirect immediately.

isoos:
Known issue trackers are now exposed on the UI. However, IIRC, adding a new keyword to `pubspec.yaml` may require a bit more coordination. @nex3 @kevmoo: what's the process there?

kevmoo:
We should create a (short) doc and socialize it a bit. We should be clear about what this field is for, valid values, etc

It requires updates to (at least) docs (cc @kwalrath ) and https://pub.dartlang.org/packages/pubspec_parse

We might want to consider some minimal validation in the pub client, too – at least during publish

@nex3 – can you think of any other place?

nex3:
Not off the top of my head.

kwalrath:
When do we want to update the docs?

mit-mit:
Two new fields:

`repository`: Optional field to specify the source code repository of the package. Useful when a package has both a home page and a repository.

`report_issue`: Optional field to a web page where developers can report new issues.

kwalrath:
I think I expected issue_tracker, not report_issue. Tha is, I expected a link to the issue tracker (e.g. https://github.com/dart-lang/angular/issues), not specifically a new-issue page (e.g. https://github.com/dart-lang/angular/issues/new).

mit-mit:
Sure, but isn't the 'report' issue link/action pretty clear (and will do the redirect) once you have made it to the issue tracker?

isoos:
[...]

PR #1585, opened 2018-09-06
Display a short message before reporting a pub site issue.

isoos:
You can try this in staging, by clicking on the "report an issue" link at the bottom of the page: https://dartlang-pub-dev.appspot.com/

screen shot 2018-09-06 at 17 37 09

I'm not sure what the best wording would be, /cc @mit-mit

### DIFF
--- a/static/js/script.dart
+++ b/static/js/script.dart
@@ -18,6 +18,7 @@ void main() {
   _setEventForMobileNav();
   _setEventForHashChange();
   _setEventForSearchInput();
+  _guardReportIssue();
   _fixIssueLinks();
   _setEventForSortControl();
   _setEventForCheckboxChanges();
@@ -314,6 +315,18 @@ void _setEventForCheckboxChanges() {
     // TODO: instead of submitting, compose the URL here (also removing the single `?`)
     formElement.submit();
   });
+}
+
+void _guardReportIssue() {
+  for (AnchorElement bugLink in document.querySelectorAll('a.github_issue')) {
+    bugLink.onClick.listen((event) {
+      if (!window.confirm('This link is for reporting issues for the pub site. '
+          'If you would like to report a problem with a package, please visit '
+          'its homepage or contact its developers.')) {
+        event.preventDefault();
+      }
+    });
+  }
 }
 
 void _fixIssueLinks() {


kevmoo:
CC @mit-mit – I really don't like this UX. These types of pop-ups aren't super clear about what clicking "okay" means vs cancel.

I realize it's easier to implement than pulling in orimplementing some other type of pop-up JS lib, though...

isoos:
> I realize it's easier to implement than pulling in orimplementing some other type of pop-up JS lib, though...

Yeah, it is a tradeoff between spending a few minutes with each mis-reported issue vs. displaying something that provides enough info about the correct process (and it is easy to implement).

static/js/script.dart
```diff
@@ -316,6 +317,18 @@ void _setEventForCheckboxChanges() {
   });
 }
 
+void _guardReportIssue() {
+  for (AnchorElement bugLink in document.querySelectorAll('a.github_issue')) {
+    bugLink.onClick.listen((event) {
+      if (!window.confirm('This link is for reporting issues for the pub site. '
+          'If you would like to report a problem with a package, please visit '
+          'its homepage or contact its developers.')) {
+        event.preventDefault();
```
sortie: Not sure if the browser popup is the right UI, maybe a html window of our own making is better. Browser popup dialogues are something of an happening and not the nicest UI.

If we use a custom window, and we're on a package page, we could link people to the package's homepage. I'd love if we did #1522 and could take advantage of that.

The message looks fine to me. If we go with this approach, we could have a tech writer or someone review that the message is clearly understood by people in a hurry that don't understand English as well as we do. I suspect your proposed UX doesn't quite catch the kind of people that don't pay attention to which bug tracker they are on.

isoos: > If we use a custom window, and we're on a package page, we could link people to the package's homepage. I'd love if we did #1522 and could take advantage of that.

Agreed. Shall we land this until that happens, or shall we leave it as it is now?

sortie: I'm OK if this lands as it is, but consider if you'll want to invest some time in improving the UX a bit, or whether we'll file an issue about doing that.

isoos: Tracking in https://github.com/dart-lang/pub-dartlang-dart/issues/1594

mit-mit:
@isoos how often do we actually get package bugs reported here?

isoos:
> how often do we actually get package bugs reported here?

2-5 per week, but the trend is increasing in the past few months.

PR #1586, opened 2018-09-06
Prominent warning for Dart2-incompatible packages.

isoos:

screenshot about the unidentified vs the legacy style:
screen shot 2018-09-06 at 18 30 41

/cc @mit-mit

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -786,6 +786,12 @@ class TemplateService {
         'text': '[outdated]',
         'title': 'Package version too old, check latest stable.',
       });
+    } else if (status == AnalysisStatus.legacy) {
+      tags.add({
+        'status': 'legacy',
+        'text': 'Dart 2 incompatible',
+        'title': 'Package does not support Dart 2.',
+      });
     } else {
       tags.add({
         'status': 'unidentified',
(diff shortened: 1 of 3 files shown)


mit-mit:
Style LGTM!

app/lib/frontend/templates.dart
```diff
@@ -786,6 +786,12 @@ class TemplateService {
         'text': '[outdated]',
         'title': 'Package version too old, check latest stable.',
       });
+    } else if (status == AnalysisStatus.legacy) {
```
sortie: Does this only happen for packages that don't allow Dart 2 in the Dart SDK constraints? It doesn't happen for packages that do allow Dart 2 but their dependencies don't?

That would be alright, we can tighten the feature later.

There's also packages that do allow Dart 2, but were released prior to Dart 2 and have analysis errors. We could mark those packages as Dart 2 incompatible as well. But that doesn't need to be now.

isoos: > Does this only happen for packages that don't allow Dart 2 in the Dart SDK constraints?

This PR: yes.

> It doesn't happen for packages that do allow Dart 2 but their dependencies don't?

It could/will happen in:
https://github.com/dart-lang/pub-dartlang-dart/pull/1587

PR #1587, opened 2018-09-06
Classify analysis as legacy if there is a version resolution failure.

isoos:
Background explained in the comment here:
https://github.com/dart-lang/pub-dartlang-dart/issues/1574#issuecomment-419169354

To be consistent with the `legacy` analysis page's message, the maintenance score is set to `0`.

### DIFF
--- a/app/lib/analyzer/pana_runner.dart
+++ b/app/lib/analyzer/pana_runner.dart
@@ -136,17 +136,28 @@ class AnalyzerJobProcessor extends JobProcessor {
     } else {
       summary = applyPlatformOverride(summary);
       summary = await _expandSummary(summary, packageStatus.age);
-      final bool lastRunWithErrors =
-          summary.suggestions?.where((s) => s.isError)?.isNotEmpty ?? false;
+      final analyzeProcessFailed =
+          summary.health?.analyzeProcessFailed ?? false;
+      final lastRunWithErrors = analyzeProcessFailed ||
+          (summary.suggestions?.where((s) => s.isError)?.isNotEmpty ?? false);
+      final bool isLegacy = lastRunWithErrors &&
+          (summary.suggestions
+                  ?.where((s) =>
+                      s.code ==
+                      SuggestionCode.pubspecDependenciesFailedToResolve)
+                  ?.isNotEmpty ??
+              false);
       if (!lastRunWithErrors) {
         analysis.analysisStatus = AnalysisStatus.success;
         status = JobStatus.success;
+      } else if (isLegacy) {
+        analysis.analysisStatus = AnalysisStatus.legacy;
       } else {
         analysis.analysisStatus = AnalysisStatus.failure;
       }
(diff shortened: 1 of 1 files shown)


app/lib/analyzer/pana_runner.dart
```diff
@@ -136,17 +136,28 @@ class AnalyzerJobProcessor extends JobProcessor {
...
       summary = await _expandSummary(summary, packageStatus.age);
-      final bool lastRunWithErrors =
-          summary.suggestions?.where((s) => s.isError)?.isNotEmpty ?? false;
+      final analyzeProcessFailed =
+          summary.health?.analyzeProcessFailed ?? false;
+      final lastRunWithErrors = analyzeProcessFailed ||
+          (summary.suggestions?.where((s) => s.isError)?.isNotEmpty ?? false);
+      final bool isLegacy = lastRunWithErrors &&
+          (summary.suggestions
+                  ?.where((s) =>
```
sortie: .any

app/lib/analyzer/pana_runner.dart
```diff
@@ -136,17 +136,28 @@ class AnalyzerJobProcessor extends JobProcessor {
     } else {
       summary = applyPlatformOverride(summary);
       summary = await _expandSummary(summary, packageStatus.age);
-      final bool lastRunWithErrors =
-          summary.suggestions?.where((s) => s.isError)?.isNotEmpty ?? false;
+      final analyzeProcessFailed =
+          summary.health?.analyzeProcessFailed ?? false;
+      final lastRunWithErrors = analyzeProcessFailed ||
+          (summary.suggestions?.where((s) => s.isError)?.isNotEmpty ?? false);
+      final bool isLegacy = lastRunWithErrors &&
```
sortie: I don't understand this. How can the package both have analysis errors and have failed to resolve dependencies? I thought failing to pub get means analysis doesn't happen?

I fear this is too broad. Having analysis errors or dependencies failing to resolve doesn't mean it's a Dart 1 package, unless it's a SDK version constraint bug (and even there, could be a package with a higher bound than we run on pub).

isoos: > I don't understand this. How can the package both have analysis errors and have failed to resolve dependencies?

We have several classes of errors. The ones you refer to are coming from `dartanalyzer`, and they are tracked in `Health.analyzerErrorCount`.

Here, it can be any issue coming form the tool-enviroment, e.g. `pub upgrade` failing due to network errors. In such case we have errors, but not in the category that would classify it as a legacy.

isoos: > Having analysis errors or dependencies failing to resolve doesn't mean it's a Dart 1 package, unless it's a SDK version constraint bug (and even there, could be a package with a higher bound than we run on pub).

There is a subset of cases where this could be a legitimate dependency resolution error, that's right.

PR #1588, opened 2018-09-07
Cache pana's ToolEnvironment for a limited amount of access.

isoos:
- Caches the dart and flutter runtime information and doesn't need to run e.g. `dart --version` before each analysis. (related to https://github.com/dart-lang/pub-dartlang-dart/issues/1347)

- Caches the downloaded packages for up to 100 separate analysis/dartdoc runs. Previously all of these were started from a clean state, and this part of the change should improve the `pub upgrade` part because it won't download the common dependencies in subsequent runs. (related to https://github.com/dart-lang/pub-dartlang-dart/issues/1521)

Because we don't want to have an unbounded number of packages being downloaded locally, I've set the arbitrary limit of 100. We could also monitor the total size of the directory, but I think we have enough value out of the simpler approach.

### DIFF
--- a/app/lib/shared/tool_env.dart
+++ b/app/lib/shared/tool_env.dart
@@ -9,6 +9,12 @@ import 'package:pana/pana.dart';
 
 import 'configuration.dart';
 
+/// Subsequent calls of the analyzer or dartdoc job can use the same [ToolEnvRef]
+/// instance [_maxCount] times.
+///
+/// Until the limit is reached, the [ToolEnvRef] will reuse the pub cache
+/// directory for its `pub upgrade` calls, but once it is reached, the cache
+/// will be deleted and a new [ToolEnvRef] with a new directory will be created.
 const _maxCount = 100;
 
 ToolEnvRef _current;
@@ -17,6 +23,10 @@ Completer _ongoing;
 /// Tracks the temporary directory of the downloaded package cache with the
 /// [ToolEnvironment] (that was initialized with that directory), along with its
 /// use stats.
+///
+/// The pub cache will be reused between `pub upgrade` calls, until the
+/// [_maxCount] threshold is reached. The directory will be deleted once all of
+/// the associated jobs complete.
 class ToolEnvRef {
   final Directory _pubCacheDir;
   final ToolEnvironment toolEnv;
@@ -43,7 +53,7 @@ class ToolEnvRef {
 }
 
 /// Gets a currently available [ToolEnvRef] if it is used less than the
-/// configured threshold (_maxCount, currently 100). If it it has already
+/// configured threshold ([_maxCount]). If it it has already
 /// reached the amount, a new cache dir and environment will be created.
 Future<ToolEnvRef> getOrCreateToolEnvRef() async {
   if (_current != null && _current._started < _maxCount) {
--- a/app/lib/shared/tool_env.dart
+++ b/app/lib/shared/tool_env.dart
@@ -56,28 +56,34 @@ class ToolEnvRef {
 /// configured threshold ([_maxCount]). If it it has already
 /// reached the amount, a new cache dir and environment will be created.
 Future<ToolEnvRef> getOrCreateToolEnvRef() async {
-  if (_current != null && _current._started < _maxCount) {
-    _current._aquire();
-    return _current;
-  }
-  if (_ongoing == null) {
+  ToolEnvRef result;
+  while (result == null) {
+    if (_current != null && _current._started < _maxCount) {
+      result = _current;
+      result._aquire();
+      break;
+    }
+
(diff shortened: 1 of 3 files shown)


app/lib/shared/tool_env.dart
```diff
@@ -0,0 +1,52 @@
...
+
+import 'package:pana/pana.dart';
+
+import 'configuration.dart';
+
+const _maxCount = 100;
+
+ToolEnvRef _current;
+
+class ToolEnvRef {
```
sortie: Add some documentation.

This class seems like it would be easy to write a test case for.

app/lib/shared/tool_env.dart
```diff
@@ -0,0 +1,52 @@
...
+
+  void _aquire() {
+    _started++;
+    _active++;
+  }
+
+  Future release() async {
+    _active--;
+    if (_started >= _maxCount && _active == 0) {
+      await _pubCacheDir.delete(recursive: true);
```
sortie: I think this also deletes the directory itself, we could consider just emptying it. I think this will be fine though as it will be remade on the next aquire.

app/lib/shared/tool_env.dart
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+}
+
+Future<ToolEnvRef> createToolEnvRef() async {
```
sortie: This design isn't concurrency safe if we'll ever want to use two different pub caches or analyze two things at once. Although, yes, we'll want some kind of shared object to share the cache.

isoos:
PTAL:
- I've fixed the issue of multiple instance creation.
- I have prepared a test, but it requires Flutter in the CI environment, and getting that installed seems not to worth it (increases the test times a lot).

app/lib/shared/tool_env.dart
```diff
@@ -12,7 +12,11 @@ import 'configuration.dart';
 const _maxCount = 100;
 
 ToolEnvRef _current;
+Completer _ongoing;
 
+/// Tracks the temporary directory of the downloaded package cache with the
+/// [ToolEnvironment] (that was initialized with that directory), along with its
+/// use stats.
```
sortie: Mention the cache eviction feature

app/lib/shared/tool_env.dart
```diff
@@ -12,7 +12,11 @@ import 'configuration.dart';
 const _maxCount = 100;
```
sortie: Add dartdoc for what this value means.

app/lib/shared/tool_env.dart
```diff
@@ -28,17 +32,32 @@ class ToolEnvRef {
...
+      if (_started >= _maxCount || _current != this) {
+        await _pubCacheDir.delete(recursive: true);
+      }
     }
   }
 }
 
-Future<ToolEnvRef> createToolEnvRef() async {
+/// Gets a currently available [ToolEnvRef] if it is used less than the
+/// configured threshold (_maxCount, currently 100). If it it has already
```
sortie: Don't document the current value, refer to the constant so it doesn't go out of date.

app/lib/shared/tool_env.dart
```diff
@@ -28,17 +32,32 @@ class ToolEnvRef {
...
     _current._aquire();
     return _current;
   }
+  if (_ongoing == null) {
+    _ongoing = new Completer();
+  } else {
+    // There may be a race condition, it is simpler to just call the method
+    // again after the calculation is done.
+    await _ongoing.future;
+    return await getOrCreateToolEnvRef();
```
sortie: Unclear how well Dart optimizes this tail call. I recommend using a while loop here instead. Right now you have a thundering herd effect where everything wakes up and races whenever a single work item completes int he queue.

A simple way to do a fair queue is to create a completer. If _ongoing is null, we set it to our future. Otherwise remember the value of _ongoing and replace it with our own future, and then we await that future. We awake and do the work. Then we complete our future. If _ongoing is our future, we set it to null (we were last).

isoos:
> I would imagine that you could mock 'flutter' in a test so we don't actually need it installed.

It would be easy to mock the parts where the command just prints some text on the output (e.g. `--version`), but that more important use case is to call `pub upgrade` / `flutter packages upgrade`, where we'd need to mock not only the `stdout`, but also do the filesystem changes.

The Dart SDK part is already covered via `pana`'s end2end tests, but we don't have Flutter-based test there. I'd rather try to increase the coverage of our end2end tests, both on `pana` and on the pub site, because many of our recent issues were coming from data-assumptions that become invalid, which would be hidden by mocks anyway.

I've created https://github.com/dart-lang/pub-dartlang-dart/issues/1595 to track these, as it will be outside of the scope of this PR.

PR #1592, opened 2018-09-10
Store dartdoc report with ScoreCard too.

isoos:
https://github.com/dart-lang/pub-dartlang-dart/issues/1352


app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -143,6 +149,20 @@ class DartdocJobProcessor extends JobProcessor {
       await tempDir.delete(recursive: true);
     }
 
+    if (!hasContent) {
+      suggestions.add(getDartdocRunFailedSuggestion());
+    }
+    // TODO: calculate coverage score
```
sortie: File an issue?

isoos: https://github.com/dart-lang/pana/issues/269 covers it.

PR #1599, opened 2018-09-12
Delete old ScoreCard and SDK dartdoc data.

isoos:
I'm introducing a new version that tracks the previously deployed runtime versions. The GC process for ScoreCard and SDK data checks the stored data only against this version, and if it is older than this reference, it can be deleted.

It provides us much better control than the rather vague "keep the latest two or the latest successful" rule of the `Analysis` GC (which would become obsolete once the ScoreCard takes over).

### DIFF
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -86,6 +86,23 @@ class DartdocBackend {
     return new PubDartdocData.fromJson(map);
   }
 
+  /// Deletes the old entries that predate [shared_versions.gcBeforeRuntimeVersion].
+  Future deleteOldSdkData() async {
+    final prefix = storage_path.dartSdkDartdocPrefix();
+    await for (BucketEntry entry in _storage.list(prefix: '$prefix/')) {
+      if (entry.isDirectory) {
+        continue;
+      }
+      final name = p.basename(entry.name);
+      final version = name.replaceAll('.json.gz', '');
+      if (version.length == 10 &&
+          version.compareTo('2018.00.00') > 0 &&
+          version.compareTo(shared_versions.gcBeforeRuntimeVersion) < 0) {
+        await _storage.delete(entry.name);
+      }
+    }
+  }
+
   /// Returns the latest stable version of a package.
   Future<String> getLatestVersion(String package) async {
     final list = await _db.lookup([_db.emptyKey.append(Package, id: package)]);
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -95,8 +95,9 @@ class DartdocBackend {
       }
       final name = p.basename(entry.name);
       final version = name.replaceAll('.json.gz', '');
-      if (version.length == 10 &&
-          version.compareTo('2018.00.00') > 0 &&
+      final matchesPattern = version.length == 10 &&
+          shared_versions.runtimeVersionPattern.hasMatch(version);
+      if (matchesPattern &&
           version.compareTo(shared_versions.gcBeforeRuntimeVersion) < 0) {
         await _storage.delete(entry.name);
       }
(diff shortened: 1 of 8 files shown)


app/lib/dartdoc/backend.dart
```diff
@@ -86,6 +86,23 @@ class DartdocBackend {
...
+  Future deleteOldSdkData() async {
+    final prefix = storage_path.dartSdkDartdocPrefix();
+    await for (BucketEntry entry in _storage.list(prefix: '$prefix/')) {
+      if (entry.isDirectory) {
+        continue;
+      }
+      final name = p.basename(entry.name);
+      final version = name.replaceAll('.json.gz', '');
+      if (version.length == 10 &&
+          version.compareTo('2018.00.00') > 0 &&
```
sortie: Why this check? What about older data?

app/lib/dartdoc/backend.dart
```diff
@@ -86,6 +86,23 @@ class DartdocBackend {
...
+          version.compareTo(shared_versions.gcBeforeRuntimeVersion) < 0) {
+        await _storage.delete(entry.name);
```
sortie: This prevents us from easily reverting back beyond gcBeforeRuntimeVersion as the newer version will interfere with the older version's data, which is none of its business.

Is the .json.gz updated continuously by the instances using it? If so, we can check whether it has been updated in the past X days and then keep it. Otherwise, we could store a marker every now and then to signal that this particular older version is currently running, and other versions shouldn't touch its files.

app/lib/shared/versions.dart
```diff
@@ -10,6 +10,16 @@ import 'utils.dart' show isNewer;
...
 
+/// The version which marks the earliest version of the data which we'd like to
+/// keep during various GC processes. Data prior to this version is subject to
+/// delete (unless there is another rule in place to keep it).
+///
+/// Make sure that at least two versions are kept here as the next candidates
+/// when the version switch happens:
+/// - 2018.09.11
+/// - 2018.09.03
+final String gcBeforeRuntimeVersion = '2018.08.27';
```
sortie: Depending on how much data each version consumes, and how much data we store, we could automate this. We could delete all data older than 6 months, for instance, if those versions are not in use. Alternatively, the above code could scan the directory and identity the N newest versions, and keep those and the older versions currently deployed.

isoos:
> I'm concerned about cornering ourselves here where we discover a big problem and we can't roll back to before it was introduced because this window was too narrow

In that case the ScoreCard* for the given runtimeVersion will be re-generated in a matter of hours (depending the number of instances), and until that happens, the site should be still working and downloads should happen.

app/lib/dartdoc/backend.dart
isoos: Sanity/paranoid check, so that we don't have a similarly-looking file that we delete by accident.

sortie: How about checking if it matches the regular expression for a date?

sortie:
Won't those be re-deleted after a couple hours more?

isoos:
> Won't those be re-deleted after a couple hours more?

Assuming that the very new and the very old versions are both running, yeah, they will be re-deleted.
How about keeping the current method (which keeps the delete logic simple and not-too-slow), and also check for being more than 6 months old, while making the re-generation of the report to happen in every month or so?

isoos:
PTAL

app/lib/shared/versions.dart
```diff
@@ -6,10 +6,23 @@ import 'package:pub_semver/pub_semver.dart';
 
 import 'utils.dart' show isNewer;
 
+/// The pattern of [runtimeVersion].
+final RegExp runtimeVersionPattern = new RegExp(r'\d{4}\.\d{2}\.\d{2}');
```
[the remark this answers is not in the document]
isoos: Year 10k will also break lexicographic ordering, and probably other assumptions too. I'd rather stick with this, I'm sure somebody will figure it out before the year 9800 or so, leaving plenty of time for migration :)

PR #1601, opened 2018-09-12
Improve (GitHub) Markdown styles.

isoos:
Addresses the original issue of https://github.com/dart-lang/pub-dartlang-dart/issues/871

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -1239,4 +1239,35 @@ pre > code {
   }
 }
 
+/******************************************************
+  (GitHub) Markdown styles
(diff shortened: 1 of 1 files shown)


isoos:
I'm putting this on hold for now, I've found a way to easily separate all of the markdown-related github styles, but I need to test them a bit so that it does what we expect it to do. I'll submit that as a separate PR if it works out better.

isoos:
Closing if favor of https://github.com/dart-lang/pub-dartlang-dart/pull/1603

PR #1609, opened 2018-09-13
Reduce the weight of the derived tokens.

isoos:
By lowering the weight of derived tokens of `somewhatCasedWords` -> (`somewhat`, `cased`, `words`), it improves full-match ranking in queries like https://github.com/dart-lang/pub-dartlang-dart/issues/1602

### DIFF
--- a/app/lib/search/text_utils.dart
+++ b/app/lib/search/text_utils.dart
@@ -62,3 +62,42 @@ Iterable<String> splitForIndexing(String text) {
 
 List<String> extractExactPhrases(String text) =>
     _exactTermRegExp.allMatches(text).map((m) => m.group(1)).toList();
+
+const int _maxWordLength = 80;
+
+Map<String, double> tokenize(String originalText) {
+  if (originalText == null || originalText.isEmpty) return null;
+  final tokens = <String, double>{};
+
+  for (String word in splitForIndexing(originalText)) {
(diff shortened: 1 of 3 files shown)


app/lib/search/text_utils.dart
```diff
@@ -62,3 +62,42 @@ Iterable<String> splitForIndexing(String text) {
...
     _exactTermRegExp.allMatches(text).map((m) => m.group(1)).toList();
+
+const int _maxWordLength = 80;
+
+Map<String, double> tokenize(String originalText) {
+  if (originalText == null || originalText.isEmpty) return null;
+  final tokens = <String, double>{};
+
+  for (String word in splitForIndexing(originalText)) {
+    if (word.length > _maxWordLength) word = word.substring(0, _maxWordLength);
```
sortie: Is it better to ignore such words entirely?

isoos: This can be a long (maybe generated) method name, and its parts could be still useful. I'm not sure if it ever occurs though, will need to check.

isoos: Wow, checking for such words opened up a few issues that I'm not sure how to attribute easily (e.g. codegen, dartdoc, or the source code), while it also has some surprises, e.g. very long text blocks with Chinese characters. I'll merge this change and track it separately: https://github.com/dart-lang/pub-dartlang-dart/issues/1615

PR #1611, opened 2018-09-14
Classify more packages as legacy.

isoos:

### DIFF
--- a/app/lib/analyzer/pana_runner.dart
+++ b/app/lib/analyzer/pana_runner.dart
@@ -137,16 +137,26 @@ class AnalyzerJobProcessor extends JobProcessor {
       summary = applyPlatformOverride(summary);
       scoreCardSummary = summary;
       summary = await _expandSummary(summary, packageStatus.age);
+      final isLegacy = summary.suggestions?.any((s) =>
+              s.isError &&
+              s.code == SuggestionCode.pubspecDependenciesFailedToResolve &&
+              s.description != null &&
+              s.description.contains('requires SDK version <2.0.0') &&
+              s.description.contains('version solving failed')) ??
+          false;
       final bool lastRunWithErrors =
           summary.suggestions?.where((s) => s.isError)?.isNotEmpty ?? false;
-      if (!lastRunWithErrors) {
+      if (isLegacy) {
+        analysis.analysisStatus = AnalysisStatus.legacy;
+        analysis.maintenanceScore = 0.0;
+      } else if (!lastRunWithErrors) {
         analysis.analysisStatus = AnalysisStatus.success;
         status = JobStatus.success;
       } else {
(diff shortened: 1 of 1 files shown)


app/lib/analyzer/pana_runner.dart
```diff
@@ -137,16 +137,26 @@ class AnalyzerJobProcessor extends JobProcessor {
       summary = applyPlatformOverride(summary);
       scoreCardSummary = summary;
       summary = await _expandSummary(summary, packageStatus.age);
+      final isLegacy = summary.suggestions?.any((s) =>
+              s.isError &&
+              s.code == SuggestionCode.pubspecDependenciesFailedToResolve &&
+              s.description != null &&
+              s.description.contains('requires SDK version <2.0.0') &&
```
sortie: Technically some packages not allowing Dart 2 have other constraints, like <2.0.0-dev.32 or <1.24.3

Does <2.0.0 depend on the Dart VM that we're using for analysis? Or is that the constraint?

Here is a complete list of the SDK constraints for the packages that don't allow dart 2 and that don't say "<2.0.0":

```
<0.0.1
<0.0.1
<0.0.1
0.5.13+1.r23552
>=0.9.0 <0.9.1
>=1.0.0 < 2.0.0-dev.infinity
<=1.10.1
>=1.17.1 < 2.0.0
>=1.19.0 <=2.0.0-dev.58.0
1.19.1
1.2.0
>=1.23.0 <1.25.0
>=1.24.0 <1.25.0
1.3.0
>=1.3.0 <1.7.0
1.3.6
1.4.2
>=1.4.3 < 2.0.0
1.6.0
1.6.0
1.7.2
1.8.0
1.8.0
1.8.0
>=1.8.0 <=2.0.0-dev.8.0
1.8.3
1.8.3
>=2.0.0-dev <=2.0.0-dev
2.0.0-dev.67.0
<=2.0.0-edge.c080951d45e79cd25df98036c4be835b284a269c
```

isoos: Good point, I've updated the pattern matching to cover (most) of these. PTAL.

app/lib/analyzer/pana_runner.dart
```diff
@@ -162,6 +166,21 @@ class AnalyzerJobProcessor extends JobProcessor {
...
 
+  bool _isLegacy(Suggestion s) {
+    final isVersionFailed = s.isError &&
+        s.code == SuggestionCode.pubspecDependenciesFailedToResolve &&
+        s.description != null &&
+        s.description.contains('version solving failed');
+    if (!isVersionFailed) {
+      return false;
+    }
+    return s.description.contains('requires SDK version 0.') ||
```
sortie: Can you add a little comment about exactly what we're trying to compute here and when it applies?

PR #1616, opened 2018-09-17
Additional animation frame before scrolling.

isoos:
Fixes https://github.com/dart-lang/pub-dartlang-dart/issues/1613


isoos:
> But if we didn't try to do the scrolling ourselves, the browser would do it, and it would just work.

Yeah, in this case it would work, however we have other cases to cover (e.g. tab switching or linking to anchors inside the non-default tab). =

srawlins:
Wow thanks for the quick fix!!!

PR #1621, opened 2018-09-17
Reduce logging level for transient job processing issues.

isoos:
https://github.com/dart-lang/pub-dartlang-dart/issues/1604

### DIFF
--- a/app/lib/job/job.dart
+++ b/app/lib/job/job.dart
@@ -6,6 +6,8 @@ import 'dart:async';
 import 'dart:math' as math;
 
 import 'package:gcloud/db.dart' as db;
+import 'package:gcloud/datastore.dart'
+    show TimeoutError, TransactionAbortedError;
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 import 'package:pool/pool.dart';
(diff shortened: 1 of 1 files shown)


app/lib/job/job.dart
```diff
@@ -55,8 +55,12 @@ abstract class JobProcessor {
             _logger.info('$_serviceAsString job completed: $jobDescription');
           } catch (e, st) {
             status = JobStatus.failed;
-            _logger.severe(
-                '$_serviceAsString job error $jobDescription', e, st);
+
+            final str = '$e';
```
sortie: e.toString() might be faster / more direct but this is fine.

app/lib/job/job.dart
```diff
@@ -55,8 +55,12 @@ abstract class JobProcessor {
...
+            final isTransient = str.contains('The transaction was aborted.');
```
sortie: There isn't a way to catch this exception a particular class?

isoos: > There isn't a way to catch this exception a particular class?

Actually, we can :) PTAL.

app/lib/job/job.dart
```diff
@@ -48,13 +50,16 @@ abstract class JobProcessor {
...
+          JobStatus status = JobStatus.failed;
           try {
             status = await process(job);
-            status ??= JobStatus.success;
+            status = JobStatus.success;
             _logger.info('$_serviceAsString job completed: $jobDescription');
+          } on TransactionAbortedError catch (e, st) {
+            _logger.info('$_serviceAsString job error $jobDescription', e, st);
+          } on TimeoutError catch (e, st) {
+            _logger.info('$_serviceAsString job error $jobDescription', e, st);
```
sortie: Maybe add a "(retrying later)" to the message if it is retried, so it's obvious it's harmless?

isoos: I think it will be redundant, because it will be retried in all of the cases, regardless of the error type.

PR #1631, opened 2018-09-19
Keep and use weights of derived tokens during scoring.

isoos:
- Fixes https://github.com/dart-lang/pub-dartlang-dart/issues/1606
- Increases memory use only by 3% (I'm surprised on this)
- Doesn't effect ranking of original tokens, only derived ones.

### DIFF
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -608,11 +608,10 @@ class TokenMatch {
 }
 
 class TokenIndex {
-  final Map<String, String> _textHashes = <String, String>{};
-  final Map<String, Map<String, double>> _inverseIds =
-      <String, Map<String, double>>{};
-  final Map<String, Set<String>> _inverseNgrams = <String, Set<String>>{};
-  final Map<String, double> _docSizes = <String, double>{};
+  final _textHashes = <String, String>{};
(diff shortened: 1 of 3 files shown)


app/lib/search/index_simple.dart
```diff
@@ -609,7 +609,8 @@ class TokenMatch {
 
 class TokenIndex {
   final Map<String, String> _textHashes = <String, String>{};
-  final Map<String, Set<String>> _inverseIds = <String, Set<String>>{};
+  final Map<String, Map<String, double>> _inverseIds =
```
sortie: You don't need to duplicate the generic type, final _inverseIds = <foo, bar>{} will deduce the correct type.

app/lib/search/index_simple.dart
```diff
@@ -634,8 +635,9 @@ class TokenIndex {
       remove(id);
     }
     for (String token in tokens.keys) {
-      final Set<String> set = _inverseIds.putIfAbsent(token, () => new Set());
-      set.add(id);
+      final Map<String, double> weights =
+          _inverseIds.putIfAbsent(token, () => <String, double>{});
+      weights[id] = math.max((weights[id] ?? 0.0), tokens[token]);
```
sortie: I don't believe the inner () is required or helps?

PR #1632, opened 2018-09-19
Refactoring TaskScheduler

isoos:
Right now it is used only for `search`, but I'm planning to extend this for the non-Job-based scheduling and race-prevention.

- removed some of the historical clutter
- re-added proper priority queues

### DIFF
--- a/app/lib/shared/task_scheduler.dart
+++ b/app/lib/shared/task_scheduler.dart
@@ -13,11 +13,8 @@ import 'utils.dart';
 final Logger _logger = new Logger('pub.scheduler');
 
 /// Interface for task execution.
+// ignore: one_member_abstracts
 abstract class TaskRunner {
-  /// Whether the scheduler should skip the task, e.g. because it has been
-  /// completed recently.
-  Future<TaskTargetStatus> checkTargetStatus(Task task);
-
   /// Run the task.
   /// Returns whether a race was detected while the run completed.
   Future<bool> runTask(Task task);
@@ -54,15 +51,7 @@ class TaskScheduler {
     Future runTask(Task task) async {
       final Stopwatch sw = new Stopwatch()..start();
       try {
-        TaskTargetStatus status;
         if (redirectPackagePages.containsKey(task.package)) {
-          status = new TaskTargetStatus.skip('Package is redirected from pub.');
-        }
-        status ??= await taskRunner.checkTargetStatus(task);
-        if (status.shouldSkip) {
-          _logger.info('Skipping scheduled task: $task: ${status.reason}.');
-          _statusTracker.add('skip');
-          _allLatencyTracker.add(sw.elapsedMilliseconds);
           return;
         }
         final bool raceDetected = await taskRunner.runTask(task);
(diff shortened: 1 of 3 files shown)


isoos:
Parts of this should also help with https://github.com/dart-lang/pub-dartlang-dart/issues/1658

app/lib/shared/task_scheduler.dart
```diff
@@ -44,68 +40,78 @@ class TaskScheduler {
   final TaskRunner taskRunner;
   final List<TaskSource> sources;
   final LastNTracker<String> _statusTracker = new LastNTracker();
-  final LastNTracker<num> _allLatencyTracker = new LastNTracker();
-  final LastNTracker<num> _workLatencyTracker = new LastNTracker();
-  int _pendingCount = 0;
+  final LastNTracker<num> _latencyTracker = new LastNTracker();
+  List<Queue<Task>> _queues;
```
sortie: Can you document the task scheduler's relation to task sources?

app/lib/shared/task_scheduler.dart
```diff
@@ -44,68 +40,78 @@ class TaskScheduler {
...
+          continue;
+        }
+      }
+
+      await runTask(task);
+    }
+  }
+
+  Task _pickFirstTask() {
+    for (Queue<Task> queue in _queues) {
```
sortie: This isn't fair. It gives priority to the first queues.

app/lib/shared/task_scheduler.dart
```diff
@@ -44,68 +40,78 @@ class TaskScheduler {
...
+    }
+
+    for (;;) {
+      final task = _pickFirstTask();
+
+      if (task == null) {
+        if (liveSubscriptions == 0) {
+          break;
+        } else {
+          await new Future.delayed(const Duration(seconds: 5));
```
sortie: I like to have a 'bool running' (or a Future / Completer depending on the design) and then set it to true when the queue is being processed. Once the queue is done, I return. When an element is added to the queue, I respawn the worker future if it's not running.

Alternatively, is there a way to merge the streams into a single stream? Then this could be a big single await for. It would also be fair and handle when all the subscriptions are dead.

app/lib/shared/task_scheduler.dart
```diff
@@ -44,68 +40,78 @@ class TaskScheduler {
...
+        (task) {
+          _queues[i].add(task);
+        },
+        onDone: () {
+          liveSubscriptions--;
+        },
+      );
+    }
+
+    for (;;) {
```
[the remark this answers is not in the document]
isoos: When using while(true), one of linter rules says "Boolean expression composed only from literals." I don't have a strict preference, but at least linter doesn't complain for this one.

app/lib/shared/task_scheduler.dart
isoos: The original version of the scheduler was just like that: it processed the messages from the first queues before the later ones. It gives a nice property: we can easily prioritize the sources, e.g. a latest stable comes before the historical versions.

app/lib/shared/task_scheduler.dart
isoos: I've updated the for loop, and I could rewrite this with some Completer-based notification from the stream, but I think the later would not add more clarity on what's going on.

sortie: Well this is a polling loop, those are generally bad (power inefficient and bad reaction time), but okay

PR #1648, opened 2018-09-25
Fix ScoreCard's maintenanceScore value.

isoos:
`getMaintenanceScore` reports it in the `0.0-100.0` range, while everything else is stored and handled in the `0.0-1.0` range.


isoos:
> Should this be fixed in pana with a breaking change?

I'll do it with a non-breaking two-phase upgrade.

PR #1654, opened 2018-09-26
Fix url rewrite NPE.

isoos:
Fixes https://github.com/dart-lang/pub-dartlang-dart/issues/1652


app/lib/shared/markdown.dart
```diff
@@ -61,10 +61,10 @@ class _UrlRewriter implements m.NodeVisitor {
     try {
       String newUrl = url;
       if (baseUrl != null && !_isAbsolute(newUrl)) {
-        newUrl = _rewriteRelativeUrl(newUrl);
+        newUrl = _rewriteRelativeUrl(newUrl) ?? newUrl;
```
sortie: Why does it return null rather than newUrl unchanged?

PR #1655, opened 2018-09-26
Guard against http_parser bug.

isoos:
Fixes https://github.com/dart-lang/pub-dartlang-dart/issues/1651

### DIFF
--- a/app/lib/shared/handlers.dart
+++ b/app/lib/shared/handlers.dart
@@ -5,6 +5,7 @@
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:logging/logging.dart';
 import 'package:shelf/shelf.dart' as shelf;
 
 import 'popularity_storage.dart';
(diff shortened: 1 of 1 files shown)


app/lib/shared/handlers.dart
```diff
@@ -101,16 +104,21 @@ shelf.Response debugResponse([Map<String, dynamic> data]) {
...
+        lastModified != null &&
+        !lastModified.isAfter(ifModifiedSince)) {
+      return true;
+    }
+
+    final ifNoneMatch = request.headers[HttpHeaders.ifNoneMatchHeader];
+    if (ifNoneMatch != null && ifNoneMatch == etag) {
+      return true;
+    }
+  } catch (e, st) {
```
sortie: Can we catch an exact exception? Note: If this is a client error, we should fail 400 rather than 500.

PR #1656, opened 2018-09-26
Treat missing SDK constraint as legacy package and mark Dart 2 incompatibility.

isoos:


isoos:
> What is the user experience here?

If you include it in a pubspec.yaml as a dependency and do `pub get`, these packages won't resolve to any version, as pub client assumes `<2.0.0`.

> Authors of Dart 2 compatible packages, not marked as such (already had a penalty for no SDK constraint), will be encouraged even more to mark their packages as Dart 2 ready.

I'm also increasing the penalty and severity on that:
https://github.com/dart-lang/pana/pull/419

PR #1663, opened 2018-10-01
Backup / update user-provided Datastore entries.

isoos:
This should help both https://github.com/dart-lang/pub-dartlang-dart/issues/1508 and https://github.com/dart-lang/pub-dartlang-dart/issues/1509

Currently it updates `Package` data, and creates new `PackageVersion` entries (no check for updating the PV entry if the underlying data may have changed).

The script has an extra protection against not writing to other store than staging. I have plans to improve it later (e.g. auto-delete versions that are no longer in prod / extra in staging), but first we shall do a clean refresh of staging with this one.

### DIFF
--- a/app/bin/tools/datastore.dart
+++ b/app/bin/tools/datastore.dart
@@ -52,13 +52,12 @@ class BackupCommand extends Command {
         ? Directory.current
         : new Directory(argResults['backup-directory'] as String);
     await backupDir.create(recursive: true);
-    // current timestamp in YYYYMMDD-HHmmss
+    // current timestamp in YYYY-MM-DD-HH-mm-ss
     final ts = new DateTime.now()
         .toUtc()
         .toIso8601String()
-        .replaceAll('-', '')
         .replaceAll('T', '-')
-        .replaceAll(':', '')
+        .replaceAll(':', '-')
         .split('.')
         .first;
     // output files
--- a/app/bin/tools/datastore.dart
+++ b/app/bin/tools/datastore.dart
@@ -265,6 +265,7 @@ class ArchiveLine {
   Map<String, dynamic> toJson() => _$ArchiveLineToJson(this);
 }
 
+// NOTE: Keep in sync with PackageVersion.
 @JsonSerializable()
 class PackageArchive {
   final String name;
@@ -295,6 +296,7 @@ class PackageArchive {
   Map<String, dynamic> toJson() => _$PackageArchiveToJson(this);
 }
 
+// NOTE: Keep in sync with PackageVersion.
 @JsonSerializable()
 class PackageVersionArchive {
   final String version;
--- a/app/bin/tools/datastore.dart
+++ b/app/bin/tools/datastore.dart
@@ -43,25 +43,25 @@ class BackupCommand extends Command {
   String get description => 'Backup user-provided data from Datastore.';
 
   BackupCommand() {
-    argParser.addOption('backup-directory', help: 'The backup directory.');
+    argParser.addOption('packages', help: 'The *-packages.jsonl backup file.');
   }
 
   @override
   Future run() async {
-    final backupDir = argResults['backup-directory'] == null
-        ? Directory.current
-        : new Directory(argResults['backup-directory'] as String);
-    await backupDir.create(recursive: true);
-    // current timestamp in YYYY-MM-DD-HH-mm-ss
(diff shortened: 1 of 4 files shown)


app/bin/tools/datastore.dart
```diff
@@ -0,0 +1,335 @@
...
+    argParser.addOption('backup-directory', help: 'The backup directory.');
+  }
+
+  @override
+  Future run() async {
+    final backupDir = argResults['backup-directory'] == null
+        ? Directory.current
+        : new Directory(argResults['backup-directory'] as String);
+    await backupDir.create(recursive: true);
+    // current timestamp in YYYYMMDD-HHmmss
```
sortie: I prefer YYYY-MM-DD-HH-mm-ss. YYYYMMDD-HHmmss is too hard to read.

app/bin/tools/datastore.dart
```diff
@@ -0,0 +1,335 @@
...
+
+  UpdateCommand() {
+    argParser.addOption('backup-directory', help: 'The backup directory.');
+  }
+
+  @override
+  Future run() async {
+    if (activeConfiguration.projectId != 'dartlang-pub-dev') {
+      print('\nPROJECT_ID is not dartlang-pub-dev, exiting.\n');
+      exit(-1);
```
sortie: exit(1) or exit(255)

app/bin/tools/datastore.dart
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    final backupDir = argResults['backup-directory'] == null
+        ? Directory.current
+        : new Directory(argResults['backup-directory'] as String);
+    // input files
+    final backupPrefix = await _detectLatest(backupDir);
```
sortie: It seems much simpler and more flexible to just have the user provide the correct backup path on the command line?

app/bin/tools/datastore.dart
```diff
@@ -0,0 +1,335 @@
...
+  });
+
+  factory PackageArchive.fromJson(Map<String, dynamic> json) =>
+      _$PackageArchiveFromJson(json);
+
+  Map<String, dynamic> toJson() => _$PackageArchiveToJson(this);
+}
+
+@JsonSerializable()
+class PackageVersionArchive {
```
sortie: These classes need to stay in sync with the main classes. Can we avoid that? If not, add some comments to these classes and their counterparts telling the developer that they must stay in sync.

app/bin/tools/datastore.dart
isoos: I thought that a cron job would be easier if it doesn't need to specify the exact file name. Maybe have it as an optional argument?

app/bin/tools/datastore.dart
isoos: I've added the notes. I could imagine a test that checks if they are in-sync, in one way or the other:
- making a common interface that both of these classes implement
- parsing the source files with analyzer and doing a comparison
- parsing the classes with dart:mirrors, although that may not be a good long-term solution (?)

app/bin/tools/datastore.dart
sortie: In a cron job, it's more useful if the output mechanism lets me supply the output filename, and then it can provide it to the restore step.

isoos: PTAL, I've updated both commands to use explicit files.

app/bin/tools/datastore.dart
```diff
@@ -135,18 +135,14 @@ class UpdateCommand extends Command {
...
-    final packagesFile = new File('$backupPrefix-packages.jsonl');
+    final packagesFile = new File(argResults['packages'] as String);
+    if (!packagesFile.existsSync()) {
+      print('Packages file $packagesFile does not exists.');
+      exit(1);
+    }
 
     await withProdServices(() async {
-      print('Restoring packages from: $packagesFile');
+      print('Updating packages from: $packagesFile');
```
sortie: 'Restoring' might be more accurate than 'Updating'

isoos: I've changed it to make it more consistent with the rest (e.g. command name, printed summary). Should I rename everything to restore?

sortie: 'restore' does make sense in a backup context and also suggests overwriting things. I think it makes sense, up to you.

PR #1667, opened 2018-10-02
Improve search backend's dartdoc content access.

isoos:
Should improve https://github.com/dart-lang/pub-dartlang-dart/issues/1547 and https://github.com/dart-lang/pub-dartlang-dart/issues/1658.

- removes the http indirection of accessing the dartdoc content
- adds generous timeout for accessing the content

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -72,9 +72,12 @@ class SearchBackend {
         versionList.where((pv) => pv != null),
         key: (pv) => (pv as PackageVersion).package);
 
-    final pubDataFutures = Future.wait(packages.map((p) =>
-        dartdocClient.getContentBytes(p.name, 'latest', 'pub-data.json',
-            timeout: const Duration(seconds: 10))));
+    final pubDataFutures = Future.wait<String>(
+      packages.map(
+        (p) => dartdocClient.getTextContent(p.name, 'latest', 'pub-data.json',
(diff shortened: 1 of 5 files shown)


app/lib/search/backend.dart
```diff
@@ -72,9 +72,12 @@ class SearchBackend {
         versionList.where((pv) => pv != null),
         key: (pv) => (pv as PackageVersion).package);
 
-    final pubDataFutures = Future.wait(packages.map((p) =>
-        dartdocClient.getContentBytes(p.name, 'latest', 'pub-data.json',
-            timeout: const Duration(seconds: 10))));
+    final pubDataFutures = Future.wait<String>(
+      packages.map(
+        (p) => dartdocClient.getTextContent(p.name, 'latest', 'pub-data.json',
+            timeout: const Duration(minutes: 1)),
```
sortie: Yes, but isn't it unreasonable that the dartdoc backend is more than 10 seconds slow for these queries? What's going on here?

isoos: What I gathered so far is the following: accessing the storage bucket has limited parallelism, not yet sure where, but I think it is either p=1 or p<=3. It seems to me that even if we start multiple parallel requests, the later one will be blocked on the former ones. It may not be a single request that is taking longer than 10 seconds, rather a bunch of parallel requests being in a queue, and their overall latency gets over 10 seconds.

Issue #1539, opened 2018-08-22
Increase cloud project quota


sortie:
If we are to speed up package analysis and dartdoc generation by having more machines, we will run out of quota. We should request more cloud quota so we can have a higher peak usage, while still using the same amount of CPU hours in total.

sortie:
We can delay this to M10 if we're not investing in speeding up dartdoc/analysis.

jonasfj:
I requested a quote increase, doubling the amount of resources we have...

We can easily ask for more if we need it, turn around was pretty quick.

sortie:
Thanks Jonas, yeah it's pretty quick at least for projects this size

Issue #1521, opened 2018-08-10
Reduce the processing time for new Pub rollouts


mit-mit:
Work on reducing down to a few hours the total time it takes to process all packages (analysis, dartdoc, etc.) when deploying a new Pub site.

sigurdm:
Closing as fixed.

sigurdm:
From a recent talk by @jonasfj we know that currently a full re-analysis is done in around 8-9 hours.

PR #1668, opened 2018-10-02
Remove http overhead of the AnalyzerClient.

isoos:
Related to #1521 and #1539. As we no longer want to tie our content-serving frontend to the service's frontend, data access is using the *Backend service objects.

### DIFF
--- a/app/lib/shared/analyzer_client.dart
+++ b/app/lib/shared/analyzer_client.dart
@@ -139,8 +139,8 @@ class AnalyzerClient {
       return data;
     } catch (e, st) {
       _logger.shout('Analysis request failed: $key', e, st);
+      return null;
     }
-    return null;
   }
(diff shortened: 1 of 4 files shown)


app/lib/shared/analyzer_client.dart
```diff
@@ -134,20 +129,16 @@ class AnalyzerClient {
...
-            key.package, key.version, panaVersion, content);
-        return data;
+            key.package, key.version, panaVersion, json.encode(data.toJson()));
       }
+      return data;
     } catch (e, st) {
-      _logger.shout('Analysis request failed on $uri', e, st);
+      _logger.shout('Analysis request failed: $key', e, st);
     }
     return null;
```
sortie: Move return null into the catch?

PR #1671, opened 2018-10-04
Refactor auth-related functions + CSRF token and cookie matching

isoos:
- https://github.com/dart-lang/pub-dartlang-dart/issues/1642
- the CSRF token matching is a bit "hanging in the air", but will be used later

### DIFF
--- a/app/lib/frontend/auth.dart
+++ b/app/lib/frontend/auth.dart
@@ -1,0 +1,33 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:gcloud/service_scope.dart' as ss;
+import 'package:shelf/shelf.dart' as shelf;
+
+import 'oauth2_service.dart';
+
+/// Sets the active logged-in user.
+void registerLoggedInUser(String user) => ss.register(#_logged_in_user, user);
+
+/// The active logged-in user. This is used for doing authentication checks.
+String get loggedInUser => ss.lookup(#_logged_in_user) as String;
+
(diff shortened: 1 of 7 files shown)


sortie:
@jonasfj Can you review this one?

app/lib/frontend/auth.dart
```diff
@@ -0,0 +1,63 @@
...
+        registerLoggedInUser(email);
+      }
+    }
+  }
+}
+
+/// Whether the request has a valid pair of CSRF tokens in the cookie and the
+/// request headers. This is the validation part of the cookie-to-header pattern
+/// https://en.wikipedia.org/wiki/Cross-site_request_forgery#Cookie-to-header_token
+bool isCsrfCookieAndHeaderValid(shelf.Request request) {
```
jonasfj: Why do we need this? Are we planning to use cookies as a separate authentication mechanism?

We currently use an oauth2 `accessToken` for authorization, can't we keep doing that, and just use a different flow when signing in on the pub site:
https://developers.google.com/identity/protocols/OAuth2UserAgent

Then we'll have an `accessToken` in javascript, and whatever ajax application we build can talk to the same APIs as the `pub` client. Hence, we won't need to implement the APIs twice, we don't need to worry about CSRF, but we still have to worry about XSS :)

I assume this in preparation of being able to login to the pub site.

isoos: I'd rather assume we will use a mix of APIs, some overlapping with the `pub` client.

I believe `CSRF` will be an issue, no matter what kind of access token we are using: the auth cookie is sent on every request, and if an attacker can create a forged request, we won't be able to distinguish it without some kind of extra check. It seems to me that they recommend even the authorization flow to be [protected by CSRF tokens](https://security.stackexchange.com/questions/20187/oauth2-cross-site-request-forgery-and-state-parameter).

isoos:
obsolete for now

Issue #1642, opened 2018-09-21
Login and self-management


mit-mit:
- Allow login on the Pub site
- See list of packages logged in user is an uploader of
- Package self management
  - Hide email
  - Remove uploader
  - Add uploader
  - Discontinue email
  - Notification on upload
  - Require 2FA auth

isoos:
Couple of notes and my current plans for implementing this:

It is very likely that we want to have a single-page app for the account- and package-administration, while at the same time we'd like to support other on-page, smaller interactions (e.g. star-ing or up/down-voting a package page).

At the moment it is not entirely clear if the SPA vs. a traditional form-based app is faster to implement (esp. wrt. testing), but I think in the long term a SPA provides more benefit for us. The default stack choice would be angular + material components.

The auth mechanism should work for both, and it is likely that it will be a combination of:
- auth cookie (JWT or just a random id that is stored in Datastore with the detailed information)
- CSRF-token (cookie value that gets to be added in the request headers)

There is one concern with the CSRF-token approach: any third-party script that is hosted on the domain could be able to read the token value, and pose a CSRF-risk. It is a very low risk, and at the moment the source of such external script could be either our app (e.g. code highlighter) or dartdoc-generated assets (e.g. typeahead.js).

I'm planning to use the `/api/admin/*` and the `/admin/*` URLs for implementing these functions.

The 2FA is a bit unknown to me: not sure how to approach it, e.g. if we are using Google's Oauth2, what does it mean for us to add 2FA on top of it?

mit-mit:
Jonas to write a design doc

PR #1672, opened 2018-10-04
pkg/pub_admin: skeleton Angular web-app

isoos:
#1642

The web app's code is the default skeleton Angular web app. I've removed a few non-necessary items and hooked up with the static file handling we have.


sortie:
@jonasfj Can you review this one?

app/lib/frontend/static_files.dart
```diff
@@ -173,3 +183,46 @@ Future updateLocalBuiltFiles() async {
...
+    if (pubGetResult.exitCode != 0) {
+      final message = 'Unable to compile pub_admin\n\n'
+          'exitCode: ${pubGetResult.exitCode}\n'
+          'STDOUT:\n${pubGetResult.stdout}\n\n'
+          'STDERR:\n${pubGetResult.stderr}';
+      throw new Exception(message);
+    }
+    final pr = await runProc(
+      'pub',
+      ['run', 'webdev', 'build'],
```
jonasfj: I might be new to dart, but why not use `package:build` and run `pub run build_runner build` as part of the commands in the `Dockerfile` ?

isoos: There is a benefit/tradeoff for both: here, it helps the development, as it will do a build on each startup of the app (local development only needs to care about restarting). In the `Dockerfile` it would make the startup of the instances faster.

isoos:
obsolete for now

PR #1674, opened 2018-10-05
Extend stat collection script

isoos:


isoos:
> I don't see what this has to do with the analyzer / front_end issue?

This is to collect the name of the packages that claim to be SDK2-compatible, and are recent, in order to run the pana analysis on them with the SDK at HEAD.

PR #1675, opened 2018-10-05
 Infer repository and issue tracker URL from homepage.

isoos:
https://github.com/dart-lang/pub-dartlang-dart/issues/1522

The current PR does not check the `pubspec.yaml` properties yet, but infers it only from the homepage. If the homepage and the repository url is the same, only the repository label will be displayed.

There is also a special case for flutter plugin packages: when they are coming from the flutter team, the homepage, the repository url, the issue tracker will be all different, and the issue tracker is on a different repository.

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -409,7 +409,6 @@ class TemplateService {
     );
     final repositoryUrl = urls.inferRepositoryUrl(homepageUrl);
     final issueTrackerUrl = urls.inferIssueTrackerUrl(homepageUrl);
-    final repositoryIsHomepage = repositoryUrl == homepageUrl;
 
     final links = <Map<String, dynamic>>[];
     void addLink(String href, String label) {
@@ -417,12 +416,10 @@ class TemplateService {
       links.add(<String, dynamic>{'href': href, 'label': label});
     }
 
-    if (repositoryIsHomepage) {
-      addLink(repositoryUrl, 'Repository');
-    } else {
+    if (repositoryUrl != homepageUrl) {
       addLink(homepageUrl, isGitHubHomepage ? 'Homepage (GitHub)' : 'Homepage');
-      addLink(repositoryUrl, 'Repository');
     }
+    addLink(repositoryUrl, 'Repository');
     addLink(issueTrackerUrl, 'Issue Tracker');
     addLink(documentationUrl, 'Documentation');
     addLink(dartdocsUrl, 'API Docs');
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -400,8 +400,6 @@ class TemplateService {
             documentationUrl.startsWith('http://pub.dartlang.org/'))) {
       documentationUrl = null;
     }
-    final isGitHubHomepage =
-        homepageUrl != null && homepageUrl.startsWith('https://github.com/');
     final dartdocsUrl = urls.pkgDocUrl(
       package.name,
       version: selectedVersion.version,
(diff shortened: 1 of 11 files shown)


app/lib/shared/urls.dart
```diff
@@ -116,3 +118,52 @@ String dartSdkMainUrl(String version) {
...
+    return null;
+  }
+  final uri = Uri.tryParse(homepage);
+  if (uri == null) {
+    return null;
+  }
+  if (uri.scheme != 'http' && uri.scheme != 'https') {
+    return null;
+  }
+  if (uri.host == 'github.com' || uri.host == 'gitlab.com') {
```
sortie: What if the host has a port set? Then we shouldn't do this, or only if 80 or 443?

app/lib/shared/urls.dart
```diff
@@ -116,3 +118,52 @@ String dartSdkMainUrl(String version) {
...
+    final segments = uri.pathSegments.take(2).toList();
+    if (segments.length != 2) {
+      return null;
+    }
+    segments.add('issues');
+    final url = new Uri(
+      scheme: uri.scheme,
+      host: uri.host,
```
sortie: This throws away the port, which happens to be fine for github and gitlab.

app/lib/shared/urls.dart
```diff
@@ -116,3 +118,52 @@ String dartSdkMainUrl(String version) {
...
+    if (segments.length != 2) {
+      return null;
+    }
+    return new Uri(scheme: uri.scheme, host: uri.host, pathSegments: segments)
+        .toString();
+  }
+  return null;
+}
+
+String inferIssueTrackerUrl(String homepage) {
```
sortie: These functions should be documented?

app/test/frontend/golden/pkg_show_page.html
```diff
@@ -333,8 +333,7 @@ <h3 class="title">About</h3>
       <p>my package description</p>
       <p>
           <a class="link" href="http://hans.juergen.com">Homepage</a><br/>
-          
-          <a class="link" href="/documentation/foobar_pkg/latest/">API Docs</a>
+          <a class="link" href="/documentation/foobar_pkg/latest/">API Docs</a><br/>
```
sortie: This adds a trailing `<br/>`. Is that of consequence?

app/lib/frontend/templates.dart
```diff
@@ -402,8 +400,32 @@ class TemplateService {
...
+      links.add(<String, dynamic>{'href': href, 'label': label});
+    }
+
+    if (repositoryIsHomepage) {
+      addLink(repositoryUrl, 'Repository');
+    } else {
+      addLink(homepageUrl, isGitHubHomepage ? 'Homepage (GitHub)' : 'Homepage');
+      addLink(repositoryUrl, 'Repository');
+    }
+    addLink(issueTrackerUrl, 'Issue Tracker');
```
sortie: 'Issue tracker' is maybe not as clear as 'Bug reports' or 'Report a bug in Foo' or something.

For github, we could preformat a message like we do with our own 'report a bug' link that shows the pub URl the person came from.

app/test/shared/urls_test.dart
```diff
@@ -80,4 +80,68 @@ void main() {
       expect(dartSdkMainUrl('2.0.0'), 'https://api.dartlang.org/stable/2.0.0/');
     });
   });
+
+  group('Infer repository URL', () {
```
sortie: These tests are great.

app/test/shared/urls_test.dart
```diff
@@ -80,4 +80,68 @@ void main() {
...
+    test('empty or bad input', () {
+      expect(inferRepositoryUrl(null), isNull);
+      expect(inferRepositoryUrl(''), isNull);
+      expect(inferRepositoryUrl('abc 123'), isNull);
+      expect(inferRepositoryUrl('ftp://github.com/a/b/c'), isNull);
```
sortie: 'package:foo/foo.dart'? :P

app/test/frontend/golden/pkg_show_page.html
isoos: Yeah, also it removes an extra newline. Its from the template change: now the loop adds a <br/> even though it may be the last one. The rendered version is the same, so I kept the template simple and just accepted the extra <br/>

app/lib/shared/urls.dart
isoos: None of the git hosting services provide alternative ports, so I think we won't do much harm if we just ignore it here.

isoos:
> The URL provided might not be normalized, i.e. it could http, or contain www.

I'm not sure if we want to go after every non-standard use and fix it for the authors, unless we also penalize their maintenance score (to give them incentive to fix it for themselves).

isoos:
> Perhaps 'Source Code' instead of the word 'Repository'? Hmm, but 'Repository' is fine probably. You could add (GitHub) if github perhaps.

'Source Code' could be the link to the archived packages sources for the given version, while the repository could have already a different version or a different project even. I'd stick with 'Repository' in this case.

The 'Issue Tracker' vs. 'Report a bug' is interesting. While I like 'Report a bug' better, I think it is a bit misleading, because the link will be on the "issue list" page, and not the "report new issue" page. If it were on the later, the visitors on the site may go there, just to be redirected to the github / gitlab auth page, and it is a bit confusing.

app/lib/shared/urls.dart
isoos: I've added a check for non-standard url - if it has a port set, we don't infer the repo and issue tracker url. I'd rather not overdo this detection/normalization part.

sortie:
We could link both to the issue tracker, and have a Report a Bug link for issue trackers we understand.

isoos:
> We could link both to the issue tracker, and have a Report a Bug link for issue trackers we understand.

That requires a bit more complex url processing: a future pubspec may specify them in any combination, and we shall handle in it a structured way. I'd merge this PR as it is, and implement that in a subsequent one.

PR #1678, opened 2018-10-08
Increase polling delay for name and dependency trackers.

isoos:
I think a one-minute window is still acceptable.

### DIFF
--- a/app/lib/shared/deps_graph.dart
+++ b/app/lib/shared/deps_graph.dart
@@ -60,7 +60,7 @@ class TransitiveDependencyGraph {
 }
 
 class PackageDependencyBuilder {
-  static const Duration pollingInterval = const Duration(seconds: 10);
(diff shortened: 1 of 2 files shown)


isoos:
> Can you explain what's happening here?

The name tracker polls for packages and tracks their name for (a) rejecting names that are too similar to an already existing package (b) will be used for the telemetry api for listing all of the package names.

Dependency tracker polls for package versions and tracks their pubspec dependencies. When there is a new upload of package A, it will trigger a new analysis for every package that depends on A.

isoos:
Worst case scenario for each:
- name tracker: in a small time period people will publish `pkgx` and `pkg_x`, and we don't reject the later one.
- dependency tracker: we don't immediately re-analyze the package that depends on the newly updated one

sortie:
sgtm

PR #1681, opened 2018-10-09
'Report an issue' link

isoos:
- https://github.com/dart-lang/pub-dartlang-dart/issues/1522
- the main change is in the initialization of the PackageLinks constructor: homepage -> repository -> issue tracker -> new issue url (vs. the previous everything comes from the homepage). It would allow us to process half-filled pubspec.yaml files.

### DIFF
--- a/app/lib/frontend/templates.dart
+++ b/app/lib/frontend/templates.dart
@@ -405,8 +405,10 @@ class TemplateService {
       version: selectedVersion.version,
       isLatest: selectedVersion.version == package.latestVersion,
     );
-    final repositoryUrl = urls.inferRepositoryUrl(homepageUrl);
-    final issueTrackerUrl = urls.inferIssueTrackerUrl(homepageUrl);
+    final packageLinks = new urls.PackageLinks.infer(
+      homepageUrl: homepageUrl,
+      documentationUrl: documentationUrl,
+    );
 
     final links = <Map<String, dynamic>>[];
     void addLink(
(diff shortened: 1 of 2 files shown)


app/lib/frontend/templates.dart
```diff
@@ -426,12 +428,15 @@ class TemplateService {
...
-    if (repositoryUrl != homepageUrl) {
+    if (packageLinks.repositoryUrl != packageLinks.homepageUrl) {
       addLink(homepageUrl, 'Homepage');
     }
-    addLink(repositoryUrl, 'Repository', detectServiceProvider: true);
-    addLink(issueTrackerUrl, 'Issue Tracker', detectServiceProvider: true);
-    addLink(documentationUrl, 'Documentation');
+    addLink(packageLinks.repositoryUrl, 'Repository',
+        detectServiceProvider: true);
+    addLink(packageLinks.issueTrackerUrl, 'Issue Tracker');
```
sortie: Shouldn't detectServiceProvider be true here?

app/lib/shared/urls.dart
```diff
@@ -189,3 +192,39 @@ String inferServiceProviderName(String url) {
   }
   return null;
 }
+
+/// The URLs provided by the package's pubspec or inferred from the homepage.
+class PackageLinks {
```
sortie: Should this be placed in the model as part of a PackageVersion?

app/lib/frontend/templates.dart
isoos: I thought that it is a bit boring to have "(GitHub)" label after each of links, and it is better to keep it only on the important one. The actual source code repository seems to be important, and also the report new issue, because the later requires an account on that site.

app/lib/shared/urls.dart
isoos: Agreed, but in a later PR, because we have another `PackageLinks` object for pagination links, and I'd rather have that refactored/renamed/moved before making this as part of PackageVersion. Without the renaming there would be a few name collision.

PR #1687, opened 2018-10-11
Move/migrate analysis target status check to scorecard.

isoos:
As the ScoreCard model will have a separate entry for all of the analysis reports in each new release, the logic to decide whether to analyze something becomes much simpler. We are already storing the scorecard reports in parallel to the "live" records, and this is the first step for relying on that data.


isoos:
> Are there any risks here?

Mostly refactoring and using the same code in a different way. One behaviour change is that we'll now check for the existence of the ScoreCardReport, and not the Analysis records. In practice this shouldn't add much risk.

PR #1689, opened 2018-10-11
 Store pana suggestions separately in scorecard report.

isoos:
This is to follow-up on the frontend changes (and blocks the migration of AnalysisView).


isoos:
> Is this an incompatible change to how the analysis data is stored? Do we need to bump any runtime versions?

I'll bump it with the Flutter or the Dartdoc update that's pending.

PR #1694, opened 2018-10-12
Don't try to set memcache content over 1MB.

isoos:
https://github.com/dart-lang/pub-dartlang-dart/issues/1693


isoos:
> What is causing these huge items to be stored?

That package has 5458 errors, and we are storing them in the analysis instance. We create a summary of that for the displayed suggestions, but still, the data is kept in the Analysis record.

Btw. ScoreCardReport will store only what we display, and with that, we won't have such large objects around.

sortie:
Thanks for the explanations, that sounds good to me.

PR #1705, opened 2018-10-16
Refactor: explicit flags for tag rendering.

isoos:


app/lib/frontend/models.dart
```diff
@@ -272,6 +272,11 @@ class PackageView {
       apiPages: apiPages,
     );
   }
+
+  bool get isAwaiting => analysisStatus == null;
+  bool get isDiscontinued => analysisStatus == AnalysisStatus.discontinued;
+  bool get isLegacy => analysisStatus == AnalysisStatus.legacy;
+  bool get isObsolete => analysisStatus == AnalysisStatus.outdated;
```
sortie: Should this be an enum? Are these mutually exclusive?

PR #1706, opened 2018-10-16
Add rel=noopener to target=_blank links.

isoos:
Fixes https://github.com/dart-lang/pub-dartlang-dart/issues/1692

### DIFF
--- a/app/lib/frontend/template_consts.dart
+++ b/app/lib/frontend/template_consts.dart
@@ -127,11 +127,11 @@ final Map<String, String> _landingBlurbs = const {
       '<p class="text">Find and use packages to build <a href="/flutter">Flutter</a> and '
       '<a href="/web">web</a> apps with <a target="_blank" rel="noopener" href="https://www.dartlang.org">Dart</a>.</p>',
   KnownPlatforms.flutter:
-      '<p class="text"><a href="https://flutter.io/">Flutter<sup><small>↗</small></sup></a> '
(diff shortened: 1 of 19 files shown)


sortie:
@jonasfj I'll give you the honor of reviewing this one, can you let me know if you're able to review & approve?

app/lib/frontend/template_consts.dart
```diff
@@ -125,13 +125,13 @@ String _landingPageTitle(String platform) {
 final Map<String, String> _landingBlurbs = const {
   'default':
       '<p class="text">Find and use packages to build <a href="/flutter">Flutter</a> and '
-      '<a href="/web">web</a> apps with <a target="_blank" href="https://www.dartlang.org">Dart</a>.</p>',
+      '<a href="/web">web</a> apps with <a target="_blank" rel="noopener" href="https://www.dartlang.org">Dart</a>.</p>',
   KnownPlatforms.flutter:
-      '<p class="text"><a href="https://flutter.io/">Flutter<sup><small>↗</small></sup></a> '
+      '<p class="text"><a target="_blank" rel="noopener" href="https://flutter.io/">Flutter<sup><small>↗</small></sup></a> '
```
jonasfj: This wasn't using target `_blank` before, but I assume that change was deliberate too.

isoos: Yes, deliberate change, I think this way it is more consistent with the other links.

PR #1707, opened 2018-10-17
Detect known failure patterns in dartdoc and skip logging.

isoos:
This should greatly improve https://github.com/dart-lang/pub-dartlang-dart/issues/1347, as most of the warnings I've checked in the dartdoc service are large output blocks of dartdoc failures, and those add up to the larger overall content size.

### DIFF
--- a/app/lib/dartdoc/dartdoc_runner.dart
+++ b/app/lib/dartdoc/dartdoc_runner.dart
@@ -306,10 +306,14 @@ class DartdocJobProcessor extends JobProcessor {
     _appendLog(logFileOutput, r.processResult);
 
     if (r.processResult.exitCode != 0) {
-      _logger.warning('Error while running dartdoc for $job.\n'
-          'exitCode: ${r.processResult.exitCode}\n'
+      final mergedOutput = 'exitCode: ${r.processResult.exitCode}\n'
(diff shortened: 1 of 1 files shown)


app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -382,3 +386,13 @@ class DartdocJobProcessor extends JobProcessor {
     _appendLog(logFileOutput, pr);
   }
 }
+
+bool _isKnownFailurePattern(String output) {
+  if (output.contains('Unhandled exception:') &&
+      output.contains('encountered ') &&
+      output.contains(' analysis errors') &&
+      output.contains('Dartdoc.logAnalysisErrors')) {
+    return true;
```
sortie: Can we tighten these patterns? I imagine that legitimate infrastructure issues might well also produce these errors.

sortie: Ah I see the &&. Maybe not so bad.

isoos: `analysis errors` is coming from the exception and `Dartdoc.logAnalysisErrors` is coming from the top of the stack trace. I think this is a tight pattern for filtering out this specific issue.

Issue #1604, opened 2018-09-13
dartdoc DetailedApiRequestError errors and warnings in the log


sortie:
I'm seeing a lot of these noisy warnings and errors in the dartdoc logs.

This one is common:

```
pub.job: dartdoc job error sequence_zip 0.8.5

Error:
    DetailedApiRequestError(status: 404, message: No such object: dartlang-pub--dartdoc-storage/sequence_zip/0.8.5/entry/22c1ab30-b612-45c3-ab76-dc8c65cc9243.json)
```

This one is rare:

```
pub.job: dartdoc job error browser 0.5.12

Error:
    DetailedApiRequestError(status: 500, message: Backend Error)
```

There's also this one, but that is an unrelated issue (can we avoid this race condition or silence it?)

```
pub.job: dartdoc job error overwrite 0.2.10

Error:
    The transaction was aborted.
```

sortie:
`The transaction was aborted` also happens for the analyzer. If it's harmless, it shouldn't be an error. Should I make a separate issue about this error?

isoos:
Slightly related, but since we no longer serve the dartdoc public content on the `dartdoc` service, the `search` service should be moved off of it too.

isoos:
I think we can reduce the log level of the `The transaction was aborted` entries, but I'd keep the rest.

isoos:
The 404 codes are now cleared from the logs, and only a few upload issues remained. Tracking the handling of those in https://github.com/dart-lang/pub-dartlang-dart/issues/1727.

PR #1708, opened 2018-10-17
Ignore 404 status code errors when deleting a file from dartdoc bucket.

isoos:
I've checked the log entries related to #1604 and many seems to happen when a _deleteAll process hits a file that is no longer there (possible due to a race). Ignoring these errors should be safe.

### DIFF
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -440,7 +440,7 @@ class DartdocBackend {
   Future _deleteFile(String path) async {
     try {
       await _storage.delete(path);
-    } on DetailedApiRequestError catch (e, _) {
+    } on DetailedApiRequestError catch (e) {
       if (e.status != 404) {
         rethrow;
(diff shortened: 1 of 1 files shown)


app/lib/dartdoc/backend.dart
```diff
@@ -434,4 +436,14 @@ class DartdocBackend {
     sw.stop();
     _logger.info('$prefix: $count files deleted in ${sw.elapsed}.');
   }
+
+  Future _deleteFile(String path) async {
+    try {
+      await _storage.delete(path);
+    } on DetailedApiRequestError catch (e, _) {
```
sortie: } on DetailedApiRequestError catch (e) {

Or is the _ needed to keep the stack trace for the rethrow?

isoos: I thought it is needed, but experimenting with a mock code, it seems it is not necessary, removing it...

PR #1709, opened 2018-10-17
Versioned search snapshots.

isoos:
There is a small risk that a very old deployment is kept running, it overrides the snapshot, but it misses some data (e.g. API entries or whatever), and a starting new `search` instance will pick it up. There will be a <hour period where we would be serving slightly off results because of that.

The PR will start to save not only the latest snapshot, but also a versioned one, and prioritize the loading of that versioned one. I'm planning to do proper fallback (to the latest existing version) and cleanup (for really old versions)  in a subsequent PR, but until then this one reduces a lot of the mentioned risk.

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -239,15 +239,28 @@ PackageDocument createSdkDocument(PubDartdocData lib) {
 
 class SnapshotStorage {
   final String _latestPath = 'snapshot-latest.json.gz';
+  final String _currentPath = 'snapshot/${versions.runtimeVersion}.json.gz';
   final Storage storage;
   final Bucket bucket;
 
   SnapshotStorage(this.storage, this.bucket);
 
   Future<SearchSnapshot> fetch() async {
+    final current = await _fetch(_currentPath);
+    if (current != null) {
+      return current;
+    }
+    return await _fetch(_latestPath);
+  }
+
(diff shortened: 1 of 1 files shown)


app/lib/search/backend.dart
```diff
@@ -239,32 +239,44 @@ PackageDocument createSdkDocument(PubDartdocData lib) {
 
 class SnapshotStorage {
```
sortie: This code isn't documented?

app/lib/search/backend.dart
```diff
@@ -239,32 +239,44 @@ PackageDocument createSdkDocument(PubDartdocData lib) {
...
   final Bucket bucket;
 
   SnapshotStorage(this.storage, this.bucket);
 
   Future<SearchSnapshot> fetch() async {
+    final current = await _fetch(_currentPath);
+    if (current != null) {
+      return current;
+    }
+    return await _fetch(_latestPath);
```
sortie: Instead of competing to write out snapshot-latest.json.gz, how about doing a directory listing to find the newest snapshot that's still older than versions.runtimeVersion?

isoos: Planning that in a subsequent PR (with proper cleanup).

app/lib/search/backend.dart
isoos: I'll do a follow-up PR on that.

Issue #1717, opened 2018-10-23
Error: Exception: Homepage URL has no valid host: https://www.example.com


sortie:
In a request for `/api/packages/versions/newUploadFinish`, the following exception occurred:

```
pubserver.shelf_pubserver: An error occured while finishing upload.

Error:
    Exception: Homepage URL has no valid host: https://www.example.com

Stack:
    #0      syntaxCheckHomepageUrl (package:pub_dartlang_org/shared/urls.dart:111:5)
    #1      _parseAndValidateUpload (package:pub_dartlang_org/frontend/backend.dart:715:3)
    <asynchronous suspension>
    #2      GCloudPackageRepository._performTarballUpload (package:pub_dartlang_org/frontend/backend.dart:308:15)
    <asynchronous suspension>
    #3      GCloudPackageRepository.finishAsyncUpload.<anonymous closure>.<anonymous closure> (package:pub_dartlang_org/frontend/backend.dart:290:16)
    <asynchronous suspension>
    #4      withTempDirectory.<anonymous closure> (package:pub_dartlang_org/shared/utils.dart:46:16)
```

However, this is a user error (HTTP 400) rather than an event that should be logged as a warning.

sortie:
Interesting, on subsequent retries this happened:

```
Error:
    DetailedApiRequestError(status: 404, message: No error details. HTTP status was: 404.)
```

Could this be the cause of #1680?

isoos:
Oh, very likely as the pubspec has just that URL: https://github.com/jodinathan/input_mask/blob/master/pubspec.yaml

isoos:
I think I know understand the sequence of events:
- On the first attempt to finish the upload, we throw a non-typed exception, and
  - We clean up the uploaded package archive.
  - It gets translated to a status code of 500.
- the `pub` client see the response and retries it another time:
  - The package archive is not there, we fail to read it, and throw a different non-typed exception.
- the `pub` client keeps retrying (5 times total)

Couple of moving parts were causing this effect:
- Prior to May/June, this wasn't an issue, because `pub` client did not retry on 500, only on 502, 503...
- There was also a change in pub_server that made a distinction between client-side problems and server issues. Client side issues get status code 400 and pub won't retry them.
- We have migrated most of our exceptions at that time, but some validation was added later that didn't follow that pattern and/or some validation were not migrated.

To fix this, I'll submit a PR that will use the proper exception for pub_server.

isoos:
[...]

PR #1721, opened 2018-10-24
Use GenericProcessingException in every validation method during upload.

isoos:
#1717

### DIFF
--- a/app/lib/shared/urls.dart
+++ b/app/lib/shared/urls.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:path/path.dart' as p;
+import 'package:pub_server/repository.dart' show GenericProcessingException;
 
 import 'packages_overrides.dart';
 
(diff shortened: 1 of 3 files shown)


app/lib/shared/urls.dart
```diff
@@ -98,17 +99,18 @@ void syntaxCheckHomepageUrl(String url) {
   try {
     uri = Uri.parse(url);
   } catch (_) {
-    throw new Exception('Unable to parse homepage URL: $url');
+    throw new GenericProcessingException('Unable to parse homepage URL: $url');
   }
   if (!uri.hasScheme || !uri.scheme.startsWith('http')) {
```
sortie: This should be a stronger check as it might allow httpfoo://, strenghen it to check for "http://" and "https://". (Doesn't need to happen in this CL)

app/test/frontend/backend_test.dart
```diff
@@ -843,12 +843,12 @@ void main() {
...
-              'Exception: Package name must not be a reserved word in Dart.');
+              'Package name must not be a reserved word in Dart.');
           expect(await fn('123test'),
-              'Exception: Package name must begin with a letter or underscore.');
+              'Package name must begin with a letter or underscore.');
           expect(
               await fn('With Space'),
-              'Exception: Package name may only contain '
+              'Package name may only contain '
               'letters, numbers, and underscores.');
```
sortie: This doesn't need to be line broken?

app/lib/shared/urls.dart
isoos: I'll do it in a subsequent PR, as it will need to have tests too.

isoos:
> I assume the pub server will handle GenericProcessingException as HTTP 400?

Correct.

PR #1722, opened 2018-10-25
Explicit check for http or https url schemes.

isoos:


app/lib/shared/urls.dart
```diff
@@ -101,9 +101,10 @@ void syntaxCheckHomepageUrl(String url) {
   } catch (_) {
     throw new GenericProcessingException('Unable to parse homepage URL: $url');
   }
-  if (!uri.hasScheme || !uri.scheme.startsWith('http')) {
+  final hasValidScheme = uri.scheme == 'http' || uri.scheme == 'https';
+  if (!hasValidScheme) {
     throw new GenericProcessingException(
-        'Use http:// or https:// URL schemes for homepage URL: $url');
+        'Homepage URL has no valid scheme: $url');
```
sortie: Actually I prefer the old exception text, more useful for people

PR #1724, opened 2018-10-25
Load the latest versioned snapshot.

isoos:
As we've deployed one release that uses the versioned snapshot, we can move away form the non-versioned one (no longer reading or writing it). On startup, we'll load the latest versioned snapshot available.
/cc @jonasfj


isoos:
> file an issue to get rid of old snapshots otherwise

https://github.com/dart-lang/pub-dartlang-dart/issues/1725

PR #1733, opened 2018-10-29
Storing search dataflow documentation.

isoos:
I'm a bit undecided whether this should live in the repo or not, let's discuss.

The diagram is made by draw.io client app, and I'd rather keep it somewhere for later modification. I'm also attaching the png version so that we don't need to open it in the app. And yes, I've given up to do this with graphviz.


sortie:
What is this even?

Issue #1417, opened 2018-06-28
Track missing dev API documentation


isoos:
- [ ] version upgrades
- [ ] Job processing and notification

mit-mit:
Can you clarify what this is?

isoos:
> Can you clarify what this is?

During code reviews Jonas rightfully pointed out a few items that have no documentation and they are not evident from the code. I'll need to document either to process or change the code to make it clear what's happening there.

PR #1736, opened 2018-10-30
Document versions and their upgrades.

isoos:
#1417


pkg/pub_dartdoc/pubspec.yaml
```diff
@@ -8,7 +8,7 @@ dependencies:
   logging: '>=0.9.3 <1.0.0'
   meta: ^1.1.5
   path: ^1.6.0
-  # dartdoc version to be pinned
+  # dartdoc version to be pinned, keep in-sync with versions.dart
```
sortie: app/lib/shared/versions.dart

PR #1745, opened 2018-11-05
Retry dartdoc uploads on 502 or 503 errors.

### LINKED ISSUE #1727: Retry dartdoc upload on 503 status code.

isoos:
Fixes #1727.

### DIFF
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -67,8 +67,8 @@ class DartdocBackend {
     final objectName =
         storage_path.dartSdkDartdocDataName(shared_versions.runtimeVersion);
     try {
-      final sink = _storage.write(objectName);
-      await file.openRead().transform(_gzip.encoder).pipe(sink);
+      await _uploadWithRetry(
+          objectName, () => file.openRead().transform(_gzip.encoder));
     } catch (e, st) {
       _logger.warning(
           'Unable to upload SDK pub dartdoc data file: $objectName', e, st);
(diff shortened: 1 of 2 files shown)


app/lib/dartdoc/backend.dart
```diff
@@ -133,10 +133,41 @@ class DartdocBackend {
...
+  bool _shouldRetryUploadError(e) {
+    if (e is DetailedApiRequestError) {
+      return e.status == 502 || e.status == 503;
+    }
+    return false;
+  }
+
+  Future _uploadWithRetry(
+      String objectName, Stream<List<int>> openStream()) async {
+    for (int attempt = 3; attempt > 0; attempt--) {
```
sortie: Don't we already have a function for the retry pattern?

isoos: We have one both in pana and in pub site, slightly different ones. I've updated the code and the utils, and I think it shouldn't be too far out to merge this with the one in pana. It would be beneficial if we had a common utility package for such shared code or contribute them to one of the async package utils.

app/lib/dartdoc/backend.dart
```diff
@@ -133,10 +133,33 @@ class DartdocBackend {
     return versions.map((pv) => pv.version).take(limit).toList();
   }
 
+  Future _uploadWithRetry(
+      String objectName, Stream<List<int>> openStream()) async {
+    await retryAsync(
+      () async {
+        final sink =
+            _storage.write(objectName, contentType: contentType(objectName));
```
jonasfj: please consider supplying the content-length as well

isoos: Will do in a follow-up (but first it will get a bit of refactoring for another PR).

jonasfj:
@isoos, why is this necessary? As far as I understand the `Bucket.write` API it does a resumeable upload if size is unknown or larger than 1MB. Otherwise, it does a single buffered upload...

It never streams the file to the server, so why do we need to warp with a retries? Shouldn't `Bucket.write` automatically retry, if the error is transient...

Should `Bucket.write` do what `gsutil` does? Or something similar, see:
https://cloud.google.com/storage/docs/gsutil/addlhelp/RetryHandlingStrategy

Isn't this a better place to handle transient errors?

isoos:
> Shouldn't Bucket.write automatically retry, if the error is transient...

I'm not sure what it should do, but in practice it does not retry on 502/503 status codes.

> Should Bucket.write do what gsutil does?
> Isn't this a better place to handle transient errors?

It could, but until that happens, we can do it too.

PR #1747, opened 2018-11-06
Refactoring runtime-versioned data files + remove old search snapshots

### LINKED ISSUE #1725: Remove old search snapshots

isoos:
- Fixes #1725.
- refactored most bucket-related utility methods to `shared/storage.dart`
- updated bucket upload with length info
- refactored the runtime-version-based code from dartdoc's sdk data and search's snapshot storages into `VersionedDataStorage`
- old search snapshots are now deleted too

### DIFF
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -6,8 +6,6 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
-import 'package:_discoveryapis_commons/_discoveryapis_commons.dart'
-    show DetailedApiRequestError;
 import 'package:gcloud/db.dart';
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:gcloud/storage.dart';
@@ -19,7 +17,7 @@ import 'package:pub_semver/pub_semver.dart';
 import '../frontend/models.dart' show Package, PackageVersion;
 
 import '../shared/dartdoc_memcache.dart';
-import '../shared/utils.dart' show contentType, retryAsync;
+import '../shared/storage.dart';
 import '../shared/versions.dart' as shared_versions;
 
 import 'models.dart';
@@ -27,7 +25,6 @@ import 'pub_dartdoc_data.dart';
 import 'storage_path.dart' as storage_path;
 
 final Logger _logger = new Logger('pub.dartdoc.backend');
-final _gzip = new GZipCodec();
 
 final Duration _contentDeleteThreshold = const Duration(days: 1);
 final Duration _sdkDeleteThreshold = const Duration(days: 182);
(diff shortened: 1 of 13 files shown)


app/lib/shared/storage.dart
```diff
@@ -16,3 +28,118 @@ Future<Bucket> getOrCreateBucket(Storage storage, String name) async {
...
+    try {
+      final info = await _bucket.info(_objectName());
+      return info != null;
+    } catch (_) {
+      return false;
+    }
+  }
+
+  /// Upload the current data file to the storage bucket.
+  Future uploadDataFile(File file) async {
```
jonasfj: let's raise the level a bit here...

Specifically, the `VersionedDataStorage` class should either be:
 A) a class that you store versioned binary files in (upload from file, download to file)
 B) a class that you store versioned binary bytes in (upload from `List<int>`, download to `List<int>`) 
 C) a class that you store versioned JSON blobs in (upload from JSON, download to JSON)

I don't really care which one we choose... Obviously, you have a plan to reuse it multiple places :)
But I care that we build abstractions that does something consistent.

app/lib/shared/storage.dart
```diff
@@ -0,0 +1,199 @@
...
+    final bytes = utf8.encode(json.encode(map));
+    try {
+      await uploadBytesWithRetry(_bucket, objectName, bytes);
+    } catch (e, st) {
+      _logger.warning('Unable to upload data file: $objectName', e, st);
+    }
+  }
+
+  /// Gets the content of the data file decoded as JSON Map.
+  Future<Map<String, dynamic>> getContentAsJsonMap([String version]) async {
```
jonasfj: Can't we just return `Future<dynamic>` and call it `getJson([String version = versions.runtimeVersion])` ?

app/lib/shared/storage.dart
```diff
@@ -16,3 +28,118 @@ Future<Bucket> getOrCreateBucket(Storage storage, String name) async {
...
+
+/// Uploads content from [bytes] to the [bucket] as [objectName].
+Future uploadBytesWithRetry(
+        Bucket bucket, String objectName, List<int> bytes) =>
+    uploadWithRetry(bucket, objectName, bytes.length,
+        () => new Stream.fromIterable([bytes]));
+
+/// Utility class to access versioned data file that follows the pattern:
+/// "/path-prefix/runtime-version.ext". The extension must use .gz as the
+/// storage uses transparent gzip on the data stream.
```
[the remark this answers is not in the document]
isoos: I'm not sure: what that would mean in our context? The `Bucket` API does not have any option to set either content-encoding or accept headers.

app/lib/shared/storage.dart
isoos: We can restrict the interface to JSON blob store, the only downside is that the upload of the SDK's `pub-data.json` would need to be first deserialized and then serialized again before uploading. Because that is a one-time event, the performance is not critical.

I don't have an immediate reuse for `VersionedDataStorage`. The static assets of the generated dartdoc archives are a bit similar, but that use case is about putting the files in a versioned subdirectory, and there is no case for JSON blobs there.

I'll update the interface to handle only JSON blobs.

app/lib/dartdoc/backend.dart
```diff
@@ -43,72 +40,32 @@ DartdocBackend get dartdocBackend =>
     ss.lookup(#_dartdocBackend) as DartdocBackend;
 
 class DartdocBackend {
-  DatastoreDB _db;
-  Bucket _storage;
+  final DatastoreDB _db;
+  final Bucket _storage;
+  final VersionedDataStorage _sdkStorage;
```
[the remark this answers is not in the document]
isoos: That is correct: the `pkg/pub_dartdoc` project creates an extracted information about the API surface + documentation (`pub-data.json`), and we store and read it with this storage.

isoos:
Rebased to master.

PR #1751, opened 2018-11-07
Match @domain.com email addresses in search.

### LINKED ISSUE #1646: Search Improvement email: *@domain.com
It would be helpful to be able to search for all packages with an uploader from a domain. So, instead of having to specify an exact email match, I could search for `email: @workiva.com` or `email: *@workiva.com`

isoos:
Fixes #1646

### DIFF
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -186,8 +186,13 @@ class SimplePackageIndex implements PackageIndex {
     if (query.parsedQuery.emails.isNotEmpty) {
       packages.removeWhere((package) {
         final doc = _packages[package];
+        if (doc?.emails == null) {
+          return true;
+        }
         for (String email in query.parsedQuery.emails) {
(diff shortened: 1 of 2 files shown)


app/lib/search/index_simple.dart
```diff
@@ -186,8 +186,13 @@ class SimplePackageIndex implements PackageIndex {
     if (query.parsedQuery.emails.isNotEmpty) {
       packages.removeWhere((package) {
         final doc = _packages[package];
+        if (doc?.emails == null) {
```
jonasfj: `doc?.emails is List<String>` might be better?

Or we could do: `doc?.emails?.contains == null`

isoos: Why would that be different?

jonasfj: It might not... only if `doc.emails` contain a non-list.. maybe? still that seems unlikely..

Issue #1691, opened 2018-10-12
Web interface should have Content-Security-Policy headers


jonasfj:
We should have `Content-Security-Policy` on all pages on `pub.dartlang.org`.
We should be very strict about the origin of scripts we allow to run.

But it might also be useful to block mixed content, and if possible disallow inline scripts.

This would mitigate security risks, if there is a bug in the markdown to HTML generator somewhere in dartdoc.
I don't really see any active ways to exploit it right now, but we should harden services.

jonasfj:
Also, we should definitively set `disown-opener`, even if it's not supported yet.
See [mdn](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) for details.

isoos:
Could you please give an example policy setting you'd like to see? I'm not that familiar with those policies and not sure what can be misconfigured.

jonasfj:
I haven't tested this, but we could do something a long the lines of:
```http
Content-Security-Policy: default-src 'none'; block-all-mixed-content
Content-Security-Policy: script-src 'self' 'unsafe-inline' 'https://www.googletagmanager.com'
Content-Security-Policy: frame-src 'self'; img-src 'self'; frame-ancestors 'self'
```

**Warning:** setting this too strict will break the site.

Ideally, we should probably try to get rid of `'unsafe-inline'` and not have inline scripts at all.
IMO, there is not rush to do this right now, but we probably should guard against XSS with CSP at some point, before we allow too many interactive features.

isoos:
I've started to extract the inline scripts from the templates, but we have a conflict with the inlined microdata we provide for Google Search:
https://developers.google.com/search/docs/guides/intro-structured-data

@jonasfj: any chance to reach out to somebody with a deeper knowledge about the content security policies, whether the `type="application/ld+json"` would give us troubles here? I guess it would be ok to leave it there, because worst case the browser probably won't interpret the script at all, and the search engine will index the part, regardless of the content security policy.

jonasfj:
Hmm, this is a good question, I'm not sure if we can continue to use embedded json-ld...

if the browser doesn't complain I would expect that we can.. :)

isoos:
There is an issue with the `block-all-mixed-content` default rule: if a package references a logo that is on `http`, we would block it from displaying. If we want to turn such CSP on, we should first have a `pana` rule that scans the `readme.md` (and changelog.md, and example.md) for such links. wdyt?

[...]

PR #1753, opened 2018-11-07
Remove inlined scripts from core template outputs.

isoos:
part of #1691

I've added tests that make sure we don't have inline scripts (except for the search index metadata JSONs), and also we don't use any JS action attributes (`on*`, eg. `onclick`).

### DIFF
--- a/app/views/layout.mustache
+++ b/app/views/layout.mustache
@@ -5,12 +5,7 @@
 <html lang="en-us">
 <head>
   <script async src="https://www.googletagmanager.com/gtag/js?id=UA-26406144-13"></script>
-  <script>
-    window.dataLayer = window.dataLayer || [];
-    function gtag(){dataLayer.push(arguments);}
-    gtag('js', new Date());
-    gtag('config', 'UA-26406144-13');
-  </script>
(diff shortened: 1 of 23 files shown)


app/views/layout.mustache
```diff
@@ -5,12 +5,7 @@
 <html lang="en-us">
 <head>
   <script async src="https://www.googletagmanager.com/gtag/js?id=UA-26406144-13"></script>
-  <script>
-    window.dataLayer = window.dataLayer || [];
-    function gtag(){dataLayer.push(arguments);}
-    gtag('js', new Date());
-    gtag('config', 'UA-26406144-13');
-  </script>
+  <script src="{{{static_assets.gtag_js}}}"></script>
```
jonasfj: hmm, I wonder if this could be async, it's not obvious to me...

isoos: Good question. Because we tag the current timestamp in the `gtag('js', new Date())` I thought that it is tracking pageload times, and with that it is probably sync, but can change to async...

jonasfj: I can't tell what it'll do in async mode...

If it'll actually work as intended?

PR #1762, opened 2018-11-09
Change 'report an issue' label in the footer.

### LINKED ISSUE #1594: Improve the UX of filing an issue (pub site vs. package's issue tracker)
We are about to land a change where the "Report an issue" link will ask a confirmation from the user that they want to report an issue agains the pub site, and not against a package. In the future, we'd like to provide a nicer, in-site dialog window that asks the same question in a more user-friendly way, and also redirects them to the package's issue tracker (if applicable).

isoos:
Fixes #1594.

@jonasfj: most of the changes are the golden files following the template change.


sortie:
This is a good change :) Let's see if it works (doesn't seem to have been so bad lately actually, maybe the report links on packages are working?)

Issue #1509, opened 2018-08-09
Ensure we have fresh data in the staging site


sortie:
Staging has an old set of packages. Can we move the latest data from production into staging? @isoos says it is done by exporting a large file and importing it in staging. Can this be automated in a reasonable / occasional manner?

@mkustermann Perhaps you know more?

mkustermann:
It's up to you to decide how often you would like to do an export of the data (and import to -dev project). Not sure if it's necessary to have -dev beign always close to the prod version. The export/import can be done via the cloud console under datastore admin.

Regarding automating this process, it is recommended to follow the [instructions here](https://cloud.google.com/datastore/docs/export-import-entities).

isoos:
We also have this: https://github.com/dart-lang/pub-dartlang-dart/blob/master/app/bin/tools/copy_data.dart

@mkustermann: was this ever used for copying?

mkustermann:
Yes, I think I did use it at some point.

isoos:
@sigurdm: would it be possible (and easy) to automate the restore-from-backup procedure you did in staging last week?

sigurdm:
I believe it should be possible - when we have implemented the backup-as-chron-job, it could run the restore-to-staging as part of the process.

sigurdm:
Basically I was running a `gcloud datastore import` and a `gsutil -m cp -r` command. But with the staging site being my project.

isoos:
@sigurdm: do you see any further work with this, or can we close the issue?

sigurdm:
@jonasfj thinks we should only run restore-to-staging manually as the data model might be different between staging and production.
The playbook covers how to restore to staging.
So I guess we can close this.

PR #1766, opened 2018-11-09
New tool: sync_staging.

isoos:
At this point this is only a read-only scanner that will give use the number of packages and versions that are missing from staging, and it can be a validation of the backup/restore process. Because there is an ongoing restore-from-backup on staging, making the updates is not important to implement right away, but I wanted to submit the code. Later on we can finish the missing pieces for the updates. (#1509)

### DIFF
--- a/app/lib/frontend/pub_client.dart
+++ b/app/lib/frontend/pub_client.dart
@@ -1,0 +1,79 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert' as convert;
+
+import 'package:http/http.dart';
+
+/// Client interface for accessing some of pub's public APIs.
+class PubClient {
+  final Client _client;
+  final String _siteRoot;
+
+  PubClient(this._client, {String siteRoot = 'https://pub.dartlang.org'})
+      : _siteRoot = siteRoot;
+
+  /// List the name of packages.
+  Future<List<String>> listPackages() async {
+    final content = await _client.read('$_siteRoot/api/packages?compact=1');
+    final map = convert.json.decode(content) as Map<String, dynamic>;
+    final packages = (map['packages'] as List).cast<String>();
+    return packages;
+  }
+
+  /// Fetch the base package data (versions and uploaders).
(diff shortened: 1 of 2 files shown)


isoos:
The tool reported 1.3k packages and 10k versions missing from staging on Saturday. I guess the update didn't went ahead on Friday?

jonasfj:
> The tool reported 1.3k packages and 10k versions missing from staging on Saturday. I guess the update didn't went ahead on Friday?

I did hit the button, but my screen session seems to be gone... so I can't tell what the result is (weird that it disappeared). But it was running and updating data points..

app/lib/frontend/pub_client.dart
```diff
@@ -0,0 +1,79 @@
...
+import 'dart:convert' as convert;
+
+import 'package:http/http.dart';
+
+/// Client interface for accessing some of pub's public APIs.
+class PubClient {
+  final Client _client;
+  final String _siteRoot;
+
+  PubClient(this._client, {String siteRoot = 'https://pub.dartlang.org'})
```
jonasfj: If this is a http client, why not just give it a default value? Is there no default global http client?

app/lib/frontend/pub_client.dart
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
```
jonasfj: I'm curious why, does this file belong in frontend folder?

isoos: Any suggestion where it should belong? It will use the APIs served by the frontend, that felt the most relevant here. Eventually this could become a standalone library (there was a chatter on gitter requesting something along the lines).

app/lib/frontend/pub_client.dart
isoos: I am not that familiar with the recent changes in the `http` package, and previously, for the `BrowserClient()` to work in the browser, it shouldn't have a codepath to initialize the IO client. I thought I just keep it that way here, keeping the option open for browser-based uses.

app/lib/frontend/pub_client.dart
jonasfj: seems fine for now... maybe `shared/`, but's probably equally bad...

isoos:
resolved conflicts, CI is green, but github doesn't update the status. merging anyway.

PR #1785, opened 2018-11-15
Don't force the job processing on datastore head polling.

### LINKED ISSUE #1744: Repeated dartdoc generation and uploads on recently published packages.
I've noticed today in the logs that `dart_numerics 0.0.2-hotfix` was generated and uploaded 20 times between `2018-11-05 11:48:57.437` and `2018-11-05 11:54:13.129`.

Based on the log entries nothing seems to be wrong with the processing, but it kept repeating it, because it cannot mark the Job to be done. The most likely reason is the nature of our Job status updates: each analyzer instance detects the new entry on their own schedule, and when they see it, they update the related Job entry. However, there is no check (?) whether a processing is already ongoing, and with that, they reset the lock process token, making the running job invalid.

isoos:
Fixes #1744

The root cause in the issue was that each `syncDatastoreHead` triggered a new Job processing unconditionally and on their own schedule, overlapping with concurrent processing. Using the task timestamp allows us to easily filter on duplicate updates.

### DIFF
--- a/app/lib/job/backend.dart
+++ b/app/lib/job/backend.dart
@@ -49,6 +49,7 @@ class JobBackend {
         ],
       ).toString();
 
+  /// Triggers analysis/dartdoc for [package]/[version] if older than [updated].
   Future trigger(JobService service, String package,
       [String version, DateTime updated]) async {
     final pKey = _db.emptyKey.append(Package, id: package);
--- a/app/lib/job/backend.dart
+++ b/app/lib/job/backend.dart
@@ -49,7 +49,8 @@ class JobBackend {
         ],
       ).toString();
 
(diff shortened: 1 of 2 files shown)


app/lib/job/backend.dart
```diff
@@ -49,7 +49,8 @@ class JobBackend {
         ],
       ).toString();
 
-  Future trigger(JobService service, String package, [String version]) async {
+  Future trigger(JobService service, String package,
```
jonasfj: Please give this a single like documentation... that would make me very happy...

```
/// Triggers analysis/dartdoc for [package]/[version] if older than [updated]
```

if that's what it does...

PR #1792, opened 2018-11-16
Fix skipped analysis penalty in ScoreCard.

isoos:
/cc @jayoung-lee:

After this change, your dashboard will probably show a dip in the scores. We have an analysis penalty for packages that are too old or that are not compatible with Dart 2: set them a fixed 0 score for health and maintenance. `ScoreCard` (the underlying data you get on the `.../metrics` URL) did not set it correctly, and the bug came apparent as I am working to migrate the UI to use it.


jayoung-lee:
Thanks for the heads up!

Issue #1786, opened 2018-11-15
Clicking the "api reference" link for a package redirects to the versions tab (only if it fails?)


jakemac53:
I was trying to look at the [analyzer docs today](https://pub.dartlang.org/documentation/analyzer/latest/) for the latest version, but it just redirects you to the versions page without any error message.

I think it would be better to give an error that the docs haven't been built yet or failed to build, and then maybe a link to the versions page to try other versions.

isoos:
This happens when dartdoc fails on the latest version, and the redirect happens so that the visitor can select another documentation version (the failed ones are marked as failed):

<img width="802" alt="screen shot 2018-11-15 at 17 21 55" src="https://user-images.githubusercontent.com/4778111/48566398-fbdf8d00-e8fa-11e8-9349-6e1197a275f6.png">

Would it help if we would mark the documentation links more prominently?

jakemac53:
The issue is that it isn't obvious that there was an error that resulted in the redirect - (also the top 3 versions of analyzer do this redirect but only the first one has the "failed" message).

In my case I just assumed I clicked the wrong link, hit back, clicked it again, repeated, and the filed this issue haha.

A dedicated error page for this would be more appropriate imo, or possibly redirect to the version page but with a very obvious error banner or something explaining why you got redirected there.

isoos:
Right. So far we have one bug and one feature request:
- Bug: `/documentation/analyzer/0.33.3+1/` redirects to the the `/latest/` URL, because the `isLatestStable` flag is not cleared (because the docs failed to generate). Because of this the status on the `/versions` page is displayed wrong.
- Feature request: display an error page, without redirect. Provide link to the `log.txt` and to the `/versions` page.

isoos:
This has been deployed and fixed.

PR #1799, opened 2018-11-20
Clear the DartdocEntry status flags on regression.

isoos:
Handles the first part of the core issue in #1786.

### DIFF
--- a/app/lib/dartdoc/dartdoc_runner.dart
+++ b/app/lib/dartdoc/dartdoc_runner.dart
@@ -192,6 +192,10 @@ class DartdocJobProcessor extends JobProcessor {
           job.packageName, job.packageVersion);
       if (entry.isRegression(oldEntry)) {
         _logger.severe('Regression detected in $job, aborting upload.');
+        // If `isLatest` or `isObsolete` have changed, we still want to update
+        // the old entry, even if the job failed. `isLatest` is used to redirect
+        // latest versions from versioned url to url with `/latest/`, and this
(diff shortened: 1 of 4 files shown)


app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -192,6 +192,10 @@ class DartdocJobProcessor extends JobProcessor {
           job.packageName, job.packageVersion);
       if (entry.isRegression(oldEntry)) {
         _logger.severe('Regression detected in $job, aborting upload.');
+        if (oldEntry.isLatest != entry.isLatest ||
+            oldEntry.isObsolete != entry.isObsolete) {
+          await dartdocBackend.updateOldEntry(oldEntry, entry);
```
jonasfj: So, basically, if `isLatest` or `isObsolete` have changed, you still want to update the old entry, even if the job failed?

Please maybe add a one-line comment stating that..

isoos:
> Isn't the underlying problem here that we have duplicate information in the DartDocEntry, since isLatest and isObsolete can be computed from other places, right?

Correct. Another option is to use an extra Datastore check (+memcache) on each content that is being served, and then we can sunset the flag too. Which would you prefer?

jonasfj:
I have no opinion at this point.. :)

isoos:
I'm merging this with the doc now. We can easily remove this if we want to change the frontend serving part to not use the fields.

PR #1815, opened 2018-11-23
History: union type for serialization.

isoos:
It is not that compact as I have hoped for.

### DIFF
--- a/app/lib/history/models.dart
+++ b/app/lib/history/models.dart
@@ -136,7 +136,7 @@ abstract class HistoryEvent {
 @JsonSerializable()
 class HistoryUnion {
   @JsonKey(includeIfNull: false)
-  final PackageVersionUploaded packageVersionUploaded;
+  final PackageUploaded packageUploaded;
 
   @JsonKey(includeIfNull: false)
   final UploaderChanged uploaderChanged;
@@ -145,7 +145,7 @@ class HistoryUnion {
   final AnalysisCompleted analysisCompleted;
 
   HistoryUnion({
-    this.packageVersionUploaded,
+    this.packageUploaded,
     this.uploaderChanged,
     this.analysisCompleted,
   }) {
@@ -153,8 +153,8 @@ class HistoryUnion {
   }
 
   factory HistoryUnion.ofEvent(HistoryEvent event) {
-    if (event is PackageVersionUploaded) {
-      return new HistoryUnion(packageVersionUploaded: event);
+    if (event is PackageUploaded) {
+      return new HistoryUnion(packageUploaded: event);
     } else if (event is UploaderChanged) {
       return new HistoryUnion(uploaderChanged: event);
     } else if (event is AnalysisCompleted) {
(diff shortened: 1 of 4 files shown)


app/lib/history/models.dart
```diff
@@ -197,22 +229,19 @@ class UploaderChanged implements HistoryEvent {
...
   final bool hasErrors;
 
   final bool hasPlatforms;
 
   AnalysisCompleted({this.hasErrors, this.hasPlatforms});
 
-  @override
-  String getType() => _type;
+  factory AnalysisCompleted.fromJson(Map<String, dynamic> json) =>
+      _$AnalysisCompletedFromJson(json);
```
jonasfj: What do you need this for?

app/lib/history/models.dart
```diff
@@ -225,6 +254,5 @@ class AnalysisCompleted implements HistoryEvent {
     }
   }
 
-  @override
   Map<String, dynamic> toJson() => _$AnalysisCompletedToJson(this);
```
jonasfj: What do we need this for?

app/lib/history/models.dart
```diff
@@ -122,46 +128,72 @@ abstract class HistoryData {
...
+    if (event is PackageVersionUploaded) {
+      return new HistoryUnion(packageVersionUploaded: event);
+    } else if (event is UploaderChanged) {
+      return new HistoryUnion(uploaderChanged: event);
+    } else if (event is AnalysisCompleted) {
+      return new HistoryUnion(analysisCompleted: event);
+    } else {
+      throw new ArgumentError('Unknown type: ${event.runtimeType}');
+    }
+  }
```
jonasfj: Couldn't we just do:
```dart
HistoryUnion.fromPackageVersionUploaded(this.PackageVersionUploaded);
HistoryUnion.fromUploaderChanged(this.UploaderChanged);
HistoryUnion.fromAnalysisCompleted(this.AnalysisCompleted);
```

Defining a one-line constructor for each option... Or maybe we could just use the first constructor, since it has an assert?

jonasfj: Ah, I now see how you're using it,.. hmm... thinking...

app/lib/history/models.dart
```diff
@@ -122,46 +128,72 @@ abstract class HistoryData {
...
-  String getType();
-
-  Map<String, dynamic> toJson();
-
   String formatMarkdown(HistoryData data);
 }
 
-typedef HistoryEvent HistoryEventFromJson(Map<String, dynamic> json);
+@JsonSerializable()
+class HistoryUnion {
```
jonasfj: What if we dropped the abstract class `HistoryEvent` and called this `HistoryEvent`?

app/lib/history/models.dart
jonasfj: Can't @JsonSerializable magically handle subtypes?

isoos: this is not a subtype, this is a composition, and it needs to have fromJson and toJson to work with it

PR #1818, opened 2018-11-27
Merging HistoryData and HistoryEvent.

isoos:
I was pondering on this a bit, and it seemed like a good idea to do. A logical follow-up would be to also include the `source` attribute, and infer the `scope` from the event object. WDYT?


isoos:
> I'm starting to think we should avoid inheritance completely.

I'm not sure what you suggest here. What kind of inheritance would you like to remove?

PR #1821, opened 2018-11-27
Reserved package name google_maps_flutter

jonasfj:


jonasfj:
It'll probably be a while before we redeploy pub anyways...

app/lib/shared/utils.dart
```diff
@@ -173,6 +173,7 @@ final _reservedPackageNames = <String>[
   'in_app_purchase',
   'location_background',
   'webview_flutter',
+  'google_maps_flutter',
```
isoos: maybe in alphabetical order...

jonasfj: I almost want to leave them unordered as motivation for getting rid of this hack :)

Issue #1749, opened 2018-11-06
email uploaders when a package is published


jonasfj:
This can ensure people notice the analysis score, if we can get it to send email after analysis is done.

It's also a good to be aware of when packages your administrate are uploaded, as a security measure.

isoos:
I think this is an easy win, but first we need to decide: which mail API should we use?

- AppEngine standard has an [E-mail API](https://cloud.google.com/appengine/docs/standard/java/mail/), but it is not clear whether we can invoke that from the flex environment.
- Both [AppEngine flex](https://cloud.google.com/appengine/docs/flexible/java/sending-emails-with-mailgun) and the [Compute Engine tutorials](https://cloud.google.com/compute/docs/tutorials/sending-mail/) suggest third-party services.
- We could possibly use a gsuite account's SMTP gateway. I remember that dartlang.org was a gsuite domain at one point.

isoos:
Any update on which API should we use?

jonasfj:
I need to talk to mit about this... I had an item in my calendar yesterday,
but circumstances.. I suspect we'll get an SMTP server we just connect to..

jonasfj:
We can use: https://pub.dartlang.org/packages/mailer

While not pretty, I suggest we store gmail username/password in datastore, that way we can use a different account for the staging site...

jonasfj:
Example of sending email:
```dart
import 'dart:io';
import 'package:mailer/mailer.dart';
import 'package:mailer/smtp_server/gmail.dart';

void main() async {
  final email = 'pub@dartlang.org';
  final pass = Platform.environment['PUB_EMAIL_PASSWORD'];

  final server = gmail(email, pass);

  final message = Message()
    ..from = Address(email, 'Pub Service')
    ..recipients.add('jonasfj@google.com')
    ..subject = 'Pub emails works!'
    ..text =
        'Hi,\nThis message is supposed to demonstrate that **emails work**.'
    ..html =
        'Hi,<br>This message is supposed to demonstrate that <b>emails work</b>.';
  final report = await send(message, server); // this won't throw exceptions
  assert(report.every((r) => r.sent), 'failed to send some message');
}
```

isoos:
Agreed, I use this package in other projects.

> I suggest we store gmail username/password in datastore

We have `PrivateKey` entity for such use-cases.

> that way we can use a different account for the staging site

Not sure about this. I'd rather have the staging site only log the e-mail, but not actually send it.

[...]

PR #1822, opened 2018-11-27
Preparations for sending emails.

isoos:
- Extracted and renamed `Author` to `EmailAddress`.
- Created `EmailMessage` and `createPackageUploadedEmail`
- Long e-mail body lines will be split by 72-character long lines.
- #1749

### DIFF
--- a/app/lib/shared/email.dart
+++ b/app/lib/shared/email.dart
@@ -2,7 +2,10 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:meta/meta.dart';
+
 final _nameEmailRegExp = new RegExp(r'^(.*)<(.+@.+)>$');
+final _defaultFrom = new EmailAddress('Pub Site Admin', 'pub@dartlang.org');
 
 /// Represents a parsed e-mail address.
 class EmailAddress {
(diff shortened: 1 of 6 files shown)


app/lib/shared/email.dart
```diff
@@ -0,0 +1,145 @@
...
+    bodyText.writeln('The listed authors are the following:');
+    authors.forEach((author) {
+      bodyText.writeln('- $author');
+    });
+    bodyText.writeln();
+  }
+
+  bodyText
+    ..writeln(
+        'If you think this is a mistake or fraud, contact us at ${_defaultFrom.email}')
```
jonasfj: don't write that email, file an issue on github:
https://github.com/dart-lang/pub-dartlang-dart/issues

app/lib/shared/email.dart
```diff
@@ -0,0 +1,137 @@
...
+  final String email;
+
+  EmailAddress(this.name, this.email);
+
+  factory EmailAddress.parse(String value) {
+    value = value.trim();
+    String name = value;
+    String email;
+
+    final match = _nameEmailRegExp.matchAsPrefix(value);
```
jonasfj: Couldn't we propose this parse string to email, and reflow body text as static method in the `mailer` package?

Or at-least use the types defined in the mailer package, that way we don't have to wrap back and forth..?

app/lib/shared/email.dart
```diff
@@ -0,0 +1,137 @@
...
+/// Creates the [EmailMessage]s that we be sent on new package upload.
+EmailMessage createPackageUploadedEmail({
+  @required String packageName,
+  @required String packageVersion,
+  @required String uploaderEmail,
+  @required List<EmailAddress> authorizedUploaders,
+  @required List<EmailAddress> authors,
+}) {
+  final subject = 'Package upload on pub: $packageName $packageVersion';
+  final bodyText = new StringBuffer()
```
jonasfj: Couldn't we just use a multi-line string in dart?

app/lib/shared/email.dart
```diff
@@ -0,0 +1,137 @@
...
+    ..writeln('Dear package maintainer,')
+    ..writeln()
+    ..writeln(
+        '$uploaderEmail uploaded a new version of package $packageName: $packageVersion')
+    ..write('\n');
```
jonasfj: `writeln()`

app/lib/shared/email.dart
```diff
@@ -0,0 +1,137 @@
...
+
+  void addRecipient(EmailAddress address) {
+    if (address.email == null || addedEmails.contains(address.email)) {
+      return;
+    }
+    addedEmails.add(address.email);
+    recipients.add(address);
+  }
+
+  authorizedUploaders.forEach(addRecipient);
```
jonasfj: Or we could do this with:
```dart
final emails = Set<String>();  // avoid duplicate recipients
final recipients = authorizedUploaders.where((u) => u.email != null && emails.add(u)).toList();
```

Or perhaps we could move this logic to `mailer` with a PR.

app/lib/shared/email.dart
isoos: `EmailAddress` (previously: `Author`) is something we use in our code outside of the scope of email. I wouldn't rely on a third-party package to provide the data structure for it. Besides: I haven't seen mailer to parse the string, it only has a structure to hold it.

`EmailMessage` on the other hand could be replaced.

app/lib/shared/email.dart
isoos: Yeah, it started out as a bit more complex template, now a multi-line string could be just fine. Changing it...

app/lib/shared/email.dart
isoos: I've filed and issue in the `mailer` package and added a TODO here to later clear this up. With the current logic, we could just use the `authorizedUploaders` list without any further processing. Do you want to keep this filtering here, or shall I simplify that part too?

jonasfj: I'm okay either way :)

isoos: Simplicity won :)

PR #1823, opened 2018-11-27
Tool to remove all *Analysis entities.

isoos:
This should be ran after the next prod release.

### DIFF
--- a/app/bin/tools/remove_all_analysis.dart
+++ b/app/bin/tools/remove_all_analysis.dart
@@ -9,6 +9,9 @@ import 'package:gcloud/db.dart';
 import 'package:pub_dartlang_org/analyzer/models.dart';
 import 'package:pub_dartlang_org/frontend/service_utils.dart';
 
+/// The tool is intended to remove all of the obsolete *Analysis entities from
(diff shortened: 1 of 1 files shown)


app/bin/tools/remove_all_analysis.dart
```diff
@@ -0,0 +1,39 @@
...
+
+import 'dart:async';
+
+import 'package:gcloud/db.dart';
+
+import 'package:pub_dartlang_org/analyzer/models.dart';
+import 'package:pub_dartlang_org/frontend/service_utils.dart';
+
+Future main(List<String> arguments) async {
+  await withProdServices(() async {
```
jonasfj: Maybe add a comment or something on when we should run this and remove the tool itself..

isoos:
> Nice! I assume we're still a few deployments away from triggering this...

Depending on how many instances we keep running, yes, it it is a few deployments in. Alternatively, I could implement this along the lines of the general [`gcBeforeRuntimeVersion`](https://github.com/dart-lang/pub-dartlang-dart/blob/master/app/lib/shared/versions.dart#L35), and that way it is less of a worry.

PR #1825, opened 2018-11-28
Use runtimeVersion as the key in Job.

isoos:
- Instead of rewriting the entire thing for `ScoreCard` I figured it would be the best to just adapt `Job` a bit.
- The GC will be done in a subsequent PR.
- Before the next deploy the indexes should be updated, but not cleared (so that the old releases will continue to work too).


isoos:
> This means the old versions will keep running dartdoc and analysis?

Correct. It also means that on a release rollback it should just work without any issues (minus the dartdoc generated content, but I'm on that too).

PR #1827, opened 2018-11-28
Prefer current runtime over future runtimes in dartdoc serving.

isoos:
This is to address the consistency issue: when there is a new release and the dartdoc content is generated, we shouldn't immediately expose it (before the traffic was migrated). With this change we'll serve only if the new release generated content while the old one did not.

### DIFF
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -187,7 +187,18 @@ class DartdocBackend {
       if (hasServing) {
         completedList.removeWhere((entry) => !entry.isServing);
       }
-      completedList.sort((a, b) => -a.timestamp.compareTo(b.timestamp));
+      completedList.sort((a, b) {
+        // content goes first
+        if (a.hasContent && !b.hasContent) return -1;
+        if (!a.hasContent && b.hasContent) return 1;
+        // next release goes to the back of the list
+        final rva = a.runtimeVersion.compareTo(shared_versions.runtimeVersion);
+        final rvb = b.runtimeVersion.compareTo(shared_versions.runtimeVersion);
+        if (rva > 0 && rvb <= 0) return 1;
+        if (rva <= 0 && rvb > 1) return -1;
+        // otherwise compare by timestamp descending
+        return -a.timestamp.compareTo(b.timestamp);
+      });
       return completedList.firstWhere((entry) => entry.hasContent,
           orElse: () => completedList.first);
     }
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -178,29 +178,20 @@ class DartdocBackend {
     }
 
     Future<DartdocEntry> loadVersion(String v) async {
-      final List<DartdocEntry> completedList =
+      final List<DartdocEntry> entries =
           await _listEntries(storage_path.entryPrefix(package, v));
-      if (completedList.isEmpty) return null;
-      final hasServing =
-          completedList.any((entry) => entry.isServing && entry.hasContent);
-      // don't remove non-serving entries if they are the only ones with content
-      if (hasServing) {
-        completedList.removeWhere((entry) => !entry.isServing);
+      // utility to remove anything not matching pred, unless that is everything
+      void retainIfAny(bool pred(DartdocEntry entry)) {
+        if (entries.any(pred)) {
+          entries.retainWhere(pred);
+        }
       }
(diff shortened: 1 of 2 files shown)


app/lib/dartdoc/backend.dart
```diff
@@ -187,7 +187,18 @@ class DartdocBackend {
...
+        if (a.hasContent && !b.hasContent) return -1;
+        if (!a.hasContent && b.hasContent) return 1;
+        // next release goes to the back of the list
+        final rva = a.runtimeVersion.compareTo(shared_versions.runtimeVersion);
+        final rvb = b.runtimeVersion.compareTo(shared_versions.runtimeVersion);
+        if (rva > 0 && rvb <= 0) return 1;
+        if (rva <= 0 && rvb > 1) return -1;
+        // otherwise compare by timestamp descending
+        return -a.timestamp.compareTo(b.timestamp);
+      });
```
jonasfj: Can I block this PR on: https://github.com/dart-lang/pub-dartlang-dart/issues/1828
I'm not quite sure what this attempts to do...

Also couldn't we do it simpler? Something along the lines of:
```dart
import '../shared/versions.dart' show runtimeVersion, semanticRuntimeVersion;

...

    Future<DartdocEntry> loadVersion(String v) async {
      final List<DartdocEntry> entries =
          await _listEntries(storage_path.entryPrefix(package, v));
      // utility to remove anything not matching pred, unless that is everything
      void prefer(pred) => entries.any(pred) ? entries.retainWhere(pred) : null;
      prefer((e) => e.isServing && e.hasContent);
      prefer((e) => e.hasContent);
      prefer((e) => e.runtimeVersion == runtimeVersion);
      entries.sort((a, b) => -a.timestamp.compareTo(b.timestamp));
      return entries.isEmpty ? null : entries.first;
    }
```

jonasfj: If possible can we at-least avoid hacking with sort... sorting by timestamp is fine... but mixing in `hasContent` and `runtimeVersion` makes this very complicated..

app/lib/dartdoc/models.dart
```diff
@@ -16,23 +16,58 @@ import 'storage_path.dart' as storage_path;
 
 part 'models.g.dart';
 
+/// Describes the details of a dartdoc-generated content.
 @JsonSerializable()
 class DartdocEntry {
+  /// Unique identifier of the entry, generated via a random UUID.
```
jonasfj: Could we instead write:
"Random uuid for lookup in storage bucket, see storage_paths"

I can guess what a uuid is, and that it's random (well, one can hope it's random), but what is it good for :) hehe

app/lib/dartdoc/models.dart
```diff
@@ -16,23 +16,58 @@ import 'storage_path.dart' as storage_path;
...
   final String uuid;
+
+  /// The package name.
```
jonasfj: not needed

app/lib/dartdoc/models.dart
```diff
@@ -16,23 +16,58 @@ import 'storage_path.dart' as storage_path;
...
   final String packageName;
+
+  /// The package version.
```
jonasfj: not needed

app/lib/dartdoc/models.dart
```diff
@@ -16,23 +16,58 @@ import 'storage_path.dart' as storage_path;
...
+
+  /// The version of Flutter that was used to fetch dependencies.
   final String flutterVersion;
+
+  /// The version of pub site customization over the dartdoc-generated HTML.
+  /// See [versions.customizationVersion].
   final String customizationVersion;
+
+  /// When the content was generated.
   final DateTime timestamp;
```
jonasfj: For future reference, maybe this is more relevant in history where we still can make changes (maybe)...

But we should have named this field `created` and then we wouldn't need a documentation comment :)

app/lib/dartdoc/models.dart
isoos: I kind of like the consistent spacing of the fields (e.g. one empty line + comment + field), but sure, removing it.

app/lib/dartdoc/models.dart
isoos: Shall I add a new `created` field and do some backward-compatible migration?

jonasfj: No, it was just a note for the future :)

Issue #1831, opened 2018-11-29
Pub is not feeling well


ThinkDigitalRepair:
Requested URL: https://pub.dartlang.org/packages?q=firebase
Request ID: c67490dba16f7283c41ca128abebd466

isoos:
```
Error:
    Exception: Service returned status code 502
```

Looks like the search service was not reachable for a minute or two, and we've got a few 500 on the the frontend.

jonasfj:
from our alerts:
>alert start: Nov 29, 2018 at 7:13AM UTC (3 minutes ago)
>alert end:  Nov 29, 2018 at 7:13AM UTC (4 minutes ago)
>Alert firing lasted
>1 min 0 secs

isoos:
We've got double-restarts (when two instances were starting within 5 minutes) this week at the following times on `search` service (listing only the first of the pairs):
- `2018-11-26 11:34:07.547 CET`
- `2018-11-26 22:07:23.757 CET`
- `2018-11-28 10:10:53.893 CET`
- `2018-11-29 07:04:38.704 CET`
- `2018-11-29 08:11:01.628 CET`

At the moment the p99 event loop latency is 1176 - 1241 ms, which means that the period sampling timer we occasionally start takes more than 1 second to respond in >1% of the cases.

Contrast that with the `frontend` service:
- `2018-11-26 11:07:46.899 CET`
- `2018-11-26 11:11:31.098 CET`

It also has <2ms p99 event loop latency.

My suspicion: AppEngine's default health check times out and triggers a service restart. I think we can override the default timeouts and reduce the number of restarts (while the long-term goal would be to make search more responsive my calculating less stuff on each search).

isoos:
I suspect these are the settings we need to tune:
https://cloud.google.com/appengine/docs/flexible/custom-runtimes/configuring-your-app-with-app-yaml#health_checks

jonasfj:
> At the moment the p99 event loop latency is 1176 - 1241 ms

Isn't this basically a metric saying that our main thread is super busy?
So what we can do to make it less busy? Do we throw more instances at the problem?
Or is there some optimization we should be looking into?

ThinkDigitalRepair:
Some sites were getting DDoSed yesterday. 3 that I use regularly. BBVA was
also facing issues. It's probably related.

isoos:
> Isn't this basically a metric saying that our main thread is super busy?

[...]

PR #1835, opened 2018-11-30
Use lookup candidates in search.

isoos:
- Reduces the problematic query time (#1831) from 10+ seconds to under 0.5 seconds.
- Refactored ngram and added tests.
- Added documentation to several places.

The core of the change is: for each token, we will store a set of derived tokens, e.g. for `geolocation`, we will store (among others): `geolo`, `ation`, `gelocation`, `geoocation` and many more... At the time of the search we transfer the query input the same way, and we use these lookups to get the original terms that we want to filter on.

The downside is that I've remove the cached pre-calculated ngrams, which makes the query calculate slightly more longer, but keeps the memory in balance (only <10% memory increase locally).

### DIFF
--- a/app/lib/search/text_utils.dart
+++ b/app/lib/search/text_utils.dart
@@ -128,17 +128,17 @@ Set<String> ngrams(String input, int minLength, int maxLength) {
 }
 
 /// Generates lookup candidates that are, either:
-/// - derived by deleting one character from [input],
-/// - are the prefix part of [input] (min 4 characters)
-/// - are the suffix part of [input] (min 4 characters)
-Set<String> deriveLookupCandidates(String input) {
+/// - derived by deleting one character from [token],
+/// - are the prefix part of [token] (min 4 characters)
+/// - are the suffix part of [token] (min 4 characters)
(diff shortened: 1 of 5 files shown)


app/lib/search/text_utils.dart
```diff
@@ -111,3 +111,41 @@ Map<String, double> tokenize(String originalText) {
...
+  }
+  return ngrams;
+}
+
+/// Generates lookup candidates that are, either:
+/// - derived by deleting one character from [input],
+/// - are the prefix part of [input] (min 4 characters)
+/// - are the suffix part of [input] (min 4 characters)
+Set<String> deriveLookupCandidates(String input) {
+  final set = new Set<String>();
```
jonasfj: Should we rename input to `token` as the input is a token, right?

Also should we not include the `input` in `set` ?

isoos: > Also should we not include the input in set ?

That could potentially add an extra, non-required entry in the `_lookupCandidates`, e.g. a token pointing to a Set that contains only that token. For the candidate calculation at lookup time we can cheaply add it to the appropriate set.

PR #1836, opened 2018-12-03
Update Flutter SDK

isoos:


isoos:
I've deployed this to staging over the weekend and it is working ok.

PR #1838, opened 2018-12-03
Sending e-mails with package:mailer.

isoos:
#1749

### DIFF
--- a/app/lib/frontend/email_sender.dart
+++ b/app/lib/frontend/email_sender.dart
@@ -1,0 +1,88 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:gcloud/db.dart';
+import 'package:gcloud/service_scope.dart' as ss;
+import 'package:logging/logging.dart';
+import 'package:mailer/mailer.dart';
+import 'package:mailer/smtp_server.dart';
+
+import '../shared/email.dart';
+import 'models.dart' show PrivateKey;
+
+final _logger = new Logger('pub.email');
+
+/// Sets the active [EmailSender].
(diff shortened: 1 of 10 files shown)


app/lib/frontend/email_sender.dart
```diff
@@ -0,0 +1,88 @@
...
+          });
+        }
+      } catch (e, st) {
+        _logger.severe('Sending e-mail failed: $logText.', e, st);
+      }
+    }
+  }
+
+  Future _update() async {
+    final now = new DateTime.now().toUtc();
```
jonasfj: Add a comment saying that you don't update, if password/username is less than 10 minutes old

app/lib/frontend/email_sender.dart
```diff
@@ -0,0 +1,88 @@
...
+    if (_lastUpdated != null && now.difference(_lastUpdated).inMinutes < 10) {
+      return;
+    }
+    final list = await _db.lookup([
+      _db.emptyKey.append(PrivateKey, id: 'smtp.username'),
```
jonasfj: Please add a few comments documenting this... maybe add a configuration entry in the `README.md` that explains how to set the smtp secrets...

jonasfj: Also either add a new kind for storing this kind of data... or document that `PrivateKey` is used to stored secrets injected into the application.. ie. in `models.dart` we should have a doc comment on `PrivateKey` saying what it is used for..

We could also remove the `PrivateKey` kind, and add a `Secret` kind instead... Just because `PrivateKey` is rather misleading.

isoos: I'll add a `Secret` in this PR, and remove `PrivateKey` in a separate one (I want to make sure that we don't need it anymore).

isoos:
PTAL. I've added `Secret` (and a tool to update it) and documentation in `doc/secrets.md`.

PR #1844, opened 2018-12-04
Added missing import dart:async

jonasfj:
My editor showed warnings because `Future` wasn't defined, how did this work?


isoos:
Which SDK are you using? I think a recent once started to export `Future` and `Stream` in `dart:core`.

Issue #1759, opened 2018-11-08
Click to get added


mit-mit:
Would be good to click to consent

isoos:
Adding an uploader is now done only after an e-mail verification.

PR #1848, opened 2018-12-05
Basic verification model and access methods.

isoos:
I was going back and forth on this one, and the current model seems to be a reasonable start:
- An event can create a Verification entry in Datastore (providing extra parameters), which is then returns an ID.
- The ID can be shared with the recipients via an URL.
- As the recipient clicks on the URL, the ID is checked for confirmation, but only with the action matched. Different action (and their different parameters) can't be mixed. The return value is the parameters (for the first confirmation) or null (if does not exists, expired, or already confirmed).

What I couldn't decide yet: whether we want to delete such entries. I think we can keep them around in case they prove to be useful, and if they are not needed, we can remove them e.g. after a few months. But the same confirm method could delete them too.

#1759

### DIFF
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -161,18 +161,18 @@ class Backend {
     return repository.downloadUrl(package, version);
   }
 
-  /// Stores a verification entry in the Datastore and returns its id that can
+  /// Stores a [UrlSecret] entry in the Datastore and returns its id that can
   /// be used in client-communication, e.g. sending via e-mail.
   ///
   /// Duplicate detection relies on the serialized JSON form of [parameters],
   /// and it is expected that callers will order the keys consistently.
   /// When a duplicate is detected, the return value becomes `null`.
-  Future<String> createVerification(
+  Future<String> createUrlSecret(
       String action, Map<String, dynamic> parameters,
       {Duration expires = const Duration(days: 1)}) async {
     final id = _uuid.v4().toString();
     final now = new DateTime.now().toUtc();
-    final verification = new models.Verification()
+    final verification = new models.UrlSecret()
       ..parentKey = db.emptyKey
       ..id = id
       ..created = now
@@ -181,7 +181,7 @@ class Backend {
       ..parameters = parameters
       ..updateHash();
 
-    final query = db.query<models.Verification>()
+    final query = db.query<models.UrlSecret>()
       ..filter('dedupHash =', verification.dedupHash);
     await for (var v in query.run()) {
       if (!v.isActive()) continue;
@@ -196,13 +196,13 @@ class Backend {
     return id;
   }
 
-  /// Updates the verification entry (identified by [action] and [id]), and
+  /// Updates the [UrlSecret] entry (identified by [action] and [id]), and
   /// return it parameters (or null if it has been expired).
-  Future<Map<String, dynamic>> confirmVerification(
+  Future<Map<String, dynamic>> confirmUrlSecret(
       String action, String id) async {
     return await db.withTransaction((tx) async {
-      final models.Verification v =
-          (await tx.lookup([db.emptyKey.append(models.Verification, id: id)]))
+      final models.UrlSecret v =
+          (await tx.lookup([db.emptyKey.append(models.UrlSecret, id: id)]))
               .single;
       final now = new DateTime.now().toUtc();
       if (v == null || !v.isActive() || v.action != action) {
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -172,7 +172,7 @@ class Backend {
       {Duration expires = const Duration(days: 1)}) async {
     final id = _uuid.v4().toString();
     final now = new DateTime.now().toUtc();
-    final verification = new models.UrlNonce()
+    final urlNonce = new models.UrlNonce()
       ..parentKey = db.emptyKey
       ..id = id
       ..created = now
@@ -182,7 +182,7 @@ class Backend {
       ..updateHash();
 
     final query = db.query<models.UrlNonce>()
-      ..filter('dedupHash =', verification.dedupHash);
+      ..filter('dedupHash =', urlNonce.dedupHash);
     // TODO: match parameters too for 100% certainty
     // The chance of mis-classification is low, but we should eventually address it.
     final count = await query
@@ -193,7 +193,7 @@ class Backend {
       return null;
     }
 
-    await db.commit(inserts: [verification]);
+    await db.commit(inserts: [urlNonce]);
     return id;
   }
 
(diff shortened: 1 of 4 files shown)


app/lib/frontend/models.dart
```diff
@@ -239,6 +241,50 @@ abstract class SecretKey {
   ];
 }
 
+/// A random ID, typically sent via e-mail, to verify and confirm an action.
+///
+/// [action] identifies the trigger and scope of the verification. It ensures
+/// that the given verification instance can't be reused for different purpose
+/// (e.g. filling the [parameters] for one goal and then trigger an action in
+/// another part of the application for another goal).
+@db.Kind(name: 'Verification', idType: db.IdType.String)
```
jonasfj: Can we call this entity a `UrlSecret` instead?

It's a secret that we put into the URL, isn't that right?

Better names are also welcome :)

app/lib/frontend/models.dart
```diff
@@ -239,6 +241,50 @@ abstract class SecretKey {
   ];
 }
 
+/// A random ID, typically sent via e-mail, to verify and confirm an action.
```
jonasfj: Secret random UUID, typically embedded in a URL and sent via e-mail...

app/lib/frontend/backend.dart
```diff
@@ -159,12 +160,66 @@ class Backend {
...
+    final verification = new models.Verification()
+      ..parentKey = db.emptyKey
+      ..id = id
+      ..created = now
+      ..expires = now.add(expires)
+      ..action = action
+      ..parameters = parameters
+      ..updateHash();
+
+    final query = db.query<models.Verification>()
```
jonasfj: This seems racy... can't we do this some other way?

app/lib/frontend/backend.dart
```diff
@@ -159,12 +160,66 @@ class Backend {
     assert(repository.supportsDownloadUrl);
     return repository.downloadUrl(package, version);
   }
+
+  /// Stores a verification entry in the Datastore and returns its id that can
+  /// be used in client-communication, e.g. sending via e-mail.
+  ///
+  /// Duplicate detection relies on the serialized JSON form of [parameters],
+  /// and it is expected that callers will order the keys consistently.
+  /// When a duplicate is detected, the return value becomes `null`.
```
jonasfj: What is the problem with duplicates?

If I try to add someone as uploader twice, I guess they get two emails and so what?

app/lib/frontend/backend.dart
sigurdm: I'm not sure I entirely understand the need for dedupping.

app/lib/frontend/backend.dart
jonasfj: I'm asking because I might be missing somewhere else where duplication is a problem :)

But isn't it better to ensure that the handlers that do something after verifying the secret is idempotent?

app/lib/frontend/backend.dart
```diff
@@ -159,12 +160,66 @@ class Backend {
...
+      }
+    }
+
+    await db.commit(inserts: [verification]);
+    return id;
+  }
+
+  /// Updates the verification entry (identified by [action] and [id]), and
+  /// return it parameters (or null if it has been expired).
+  Future<Map<String, dynamic>> confirmVerification(
```
jonasfj: You should document that any action taken following this verification should be idempotent.

We should probably allow this to go through, if the user tries to verify within 5 minutes.

app/lib/frontend/backend.dart
isoos: The verification can be abused and send 100s of emails to the same address. I guess it can be abused in many other ways, e.g. sending separate emails to 100s of different addresses too, but at least the repeated sending to one email is filtered with this one easily.

E.g. I imagine when this is live, and someone adds a new uploader, and won't see it on the pub site right away, they will repeat the command many times.

app/lib/frontend/models.dart
```diff
@@ -239,6 +241,50 @@ abstract class SecretKey {
...
+  @db.DateTimeProperty()
+  DateTime created;
+
+  @db.DateTimeProperty()
+  DateTime expires;
+
+  @db.StringProperty(required: true)
+  String action;
+
+  @db.StringProperty()
```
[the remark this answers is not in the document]
isoos: I'd expect the `action` (or some derivative of it) to be exposed as part of the URL. I think it adds a nice touch to the verification flow that the URL hints what action to expect when the recipient will click on it.

As it would be exposed and treated separately anyway, why not keep it separate?

app/lib/frontend/backend.dart
isoos: Is this in case the verification process is aborted for some reason? Yeah, an additional 5-10 minutes window would be okay I guess.

app/lib/frontend/models.dart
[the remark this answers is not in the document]
sigurdm: Mainly to simplify this class.
However I can see now how it might be useful.
What about the parameters? They could also be exposed in the URL - then it would be clear what package you are adding yourself to, and what the email is.

No strong opinions - I have not seen the client code yet.

app/lib/frontend/backend.dart
```diff
@@ -159,12 +160,66 @@ class Backend {
...
+  ///
+  /// Duplicate detection relies on the serialized JSON form of [parameters],
+  /// and it is expected that callers will order the keys consistently.
+  ///
+  /// When multiple duplicate is detected, the return value becomes `null`.
+  Future<String> createUrlNonce(String action, Map<String, dynamic> parameters,
+      {Duration expires = const Duration(days: 1)}) async {
+    final id = _uuid.v4().toString();
+    final now = new DateTime.now().toUtc();
+    final verification = new models.UrlNonce()
```
jonasfj: variable name

app/lib/frontend/models.dart
```diff
@@ -239,6 +241,56 @@ abstract class SecretKey {
...
+    dedupHash =
+        sha256.convert(utf8.encode('$action/$parametersJson')).toString();
+  }
+
+  bool isActive() {
+    final now = new DateTime.now().toUtc();
+    if (confirmed != null && now.difference(confirmed).inMinutes < 10) {
+      return true;
+    }
+    return expires.isAfter(now);
```
jonasfj: huh, what about `&& confirmed == null`

jonasfj: please add a doc comment explaining what `isActive` means

app/lib/frontend/backend.dart
```diff
@@ -159,12 +160,66 @@ class Backend {
...
+
+  /// Updates the [models.UrlNonce] entry (identified by [action] and [id]), and
+  /// return it parameters (or null if it has been expired).
+  Future<Map<String, dynamic>> confirmUrlNonce(String action, String id) async {
+    return await db.withTransaction((tx) async {
+      final models.UrlNonce v =
+          (await tx.lookup([db.emptyKey.append(models.UrlNonce, id: id)]))
+              .single;
+      final now = new DateTime.now().toUtc();
+      if (v == null || !v.isActive() || v.action != action) {
```
jonasfj: If it's been confirmed within the past 10 min we should allow it through again...

isoos:
Closing, preferring #1860 to land.

PR #1849, opened 2018-12-05
Removing HistoryScope

isoos:
This is rather a thought-experiment than a solid intent: I'm not really sure of the usefulness of the `scope` field, and removing it seems to simplify both the query patterns and probably the display of the entries too. Let's discuss if you think we should have it for whatever reasons I'm overlooking right now.


isoos:
> We can always derive this from content anyways...

And re-index... true.

PR #1850, opened 2018-12-06
Rediscache

jonasfj:
This includes a few packages for use in pub, I might move them to a separate repository and publish them later... The dependencies on dartis via git will be removed when a new version is published:
https://github.com/jcmellado/dartis/issues/14

At this point I'm asking for a preliminary review... Later I plan to do away with `SimpleMemcache` and only use `Cache<T>` with codecs for custom types.


pkg/simple_cache/pubspec.yaml
```diff
@@ -0,0 +1,22 @@
...
+  sdk: '>=2.0.0 <3.0.0'
+
+dependencies:
+  logging: '^0.11.3'
+  identity_codec:
+    path: ../identity_codec
+  retry:
+    path: ../retry
+  dartis:
+    git:
```
isoos: I think this is a high-risk dependency even if you are in control of that repository, going outside of the usual review cycle. I'd rather have a copy hosted here too.

jonasfj: there is work in progress to have it published. I would want that before we land this PR :)
Otherwise, we should vendor it into `third_party/` per policy.

pkg/identity_codec/lib/identity_codec.dart
```diff
@@ -0,0 +1,29 @@
...
+
+/// A [Codec] that converts from T to T doing absolutely nothing.
+///
+/// This is the identity codec, the one that passes input to output when
+/// decoding and encoding. This codec is mostly useful when implementing generic
+/// code that can be fused with a [Codec] and you need an identity codec for the
+/// base case.
+///
+/// Note, that when fused with another [Codec] the identity codec disppears.
+class IdentityCodec<T> extends Codec<T, T> {
```
isoos: Instead of having a generic identity codec, wouldn't it be better to specify a raw type (e.g. `List<int>` or `Uint8List`) and require a codec for that? Especially if the value type is not immutable, it could cause inconsistency between a serialization-based cache impl and an in-memory identity-based one.
I have a package working with `Uint8List` called `package:buffer`, and it can be used to efficiently work with different custom buffers.

pkg/simple_cache/lib/simple_cache.dart
```diff
@@ -0,0 +1,219 @@
...
+    return _owner._codec.decode(value);
+  }
+
+  @override
+  Future<T> set(T value, [Duration ttl]) async {
+    if (value == null) {
+      await purge();
+      return null;
+    }
+    if (ttl == null) {
```
isoos: `ttl ??= _owner.ttl;`

pkg/simple_cache/lib/src/providers/redis.dart
```diff
@@ -0,0 +1,112 @@
...
+  Future<_RedisContext> _ctx;
+
+  Future<_RedisContext> _connect() => _r.retry(() async {
+        // Attempt to make a connection, timeout after _connectTimeLimit
+        final conn = await Connection.connect(_connectionString)
+            .timeout(_connectTimeLimit);
+        // When connection is closed, error or somehow we reset _ctx.
+        // Any errors have already been returned if any commands was affect
+        // otherwise the error is just a broken TCP connection and we ignore it.
+        conn.done
```
isoos: `conn.done.whenComplete()`? it looks like you won't be using neither the completion value nor the error.

pkg/simple_cache/lib/cache_provider.dart
```diff
@@ -0,0 +1,94 @@
...
+
+  /// Clear value stored at [key].
+  ///
+  /// After this has returned future calls to [get] for the given [key] should
+  /// not return any value, unless a new value have been set.
+  Future purge(String key);
+}
+
+/// Interface for a [CacheProvider] that can be closed.
+abstract class CloseableCacheProvider<T> extends CacheProvider<T> {
```
isoos: I'd probably just add the `close();` method on the `CacheProvider` class, and if the implementation doesn't need it, it can still use `Future close() async => null;`

pkg/simple_cache/lib/simple_cache.dart
```diff
@@ -0,0 +1,219 @@
...
+  Cache<T> withTTL(Duration ttl);
+
+  /// Create a [Cache] wrapping a [CacheProvider].
+  factory Cache(CacheProvider<T> provider) {
+    return _Cache<T, T>(provider, '', _logger, false, IdentityCodec());
+  }
+
+  /// Create an in-memory [CacheProvider] holding a maximum of [maxSize] cache
+  /// entries.
+  static CacheProvider<List<int>> inMemoryCacheProvider(int maxSize) {
```
isoos: This part is a bit confusing: `Cache.something` is used to create a `CacheProvider`, and also has a `Cache(provider)` constructor. I'd probably expect only the later, and maybe another factory constructor for the in-memory cache, and the redis-based implementation should be separate from this one.

pkg/simple_cache/lib/cache_provider.dart
```diff
@@ -0,0 +1,94 @@
...
+/// a [CacheProvider] and provide a more convinient high-level interface.
+///
+/// Implementers of [CacheProvider] can implement something that stores a value
+/// of anytime `T`, but usually implementers should aim to implement
+/// `CacheProvider<List<int>>` which stores binary data.
+///
+/// If interfacing a remote system over network where intermittent errors may
+/// occur, implementers should take care to retry operations as appropripate.
+///
+/// **Cache evition**: implementers must support [purge] to clear cache entries
```
isoos: * eviction

app/lib/shared/memcache.dart
```diff
@@ -30,13 +30,13 @@ const String searchServiceResultPrefix = 'search_service_result_';
 // Keeping it under that limit in order to offset char coding or other payloads.
 const _contentLimit = 1000 * 1000;
 
-class SimpleMemcache {
+class SimpleMemcache2 {
```
sigurdm: ??

pkg/simple_cache/lib/src/providers/inmemory.dart
```diff
@@ -0,0 +1,70 @@
...
+      }
+      // Remove, if expired
+      _old.remove(key);
+    }
+    return null;
+  }
+
+  @override
+  Future set(String key, T value, [Duration ttl]) async {
+    if (ttl == null) {
```
isoos: Why not add a default ttl for the in-memory cache too? Could simplify this code and the above getter too.

pkg/simple_cache/lib/simple_cache.dart
```diff
@@ -0,0 +1,219 @@
...
+///
+/// A [Cache] can be _fused_ with a [Codec] using [withCodec] to get a cache
+/// that stores a different kind of objects. It is also possible to create
+/// a chuild cache using [withPrefix], such that all entries in the child
+/// cache have a given prefix.
+abstract class Cache<T> {
+  /// Get [Entry] wrapping data cached at [key].
+  Entry<T> operator [](String key);
+
+  /// Get a [Cache] wrapping this cache with given [prefix].
```
sigurdm: of

pkg/simple_cache/lib/cache_provider.dart
jonasfj: Yeah, I guess that makes it simpler... I like that idea :)

pkg/simple_cache/lib/simple_cache.dart
jonasfj: If we decide that all `CacheProvider` implementations have a `close()` method, then this becomes rather simple.

app/lib/shared/memcache.dart
jonasfj: oh, left over, I probably meant to delete this file completely :)

pkg/simple_cache/lib/src/providers/inmemory.dart
```diff
@@ -0,0 +1,70 @@
...
+  final T value;
+  final DateTime _expires;
+  _InMemoryEntry(this.value, [this._expires]);
+  bool get isExpired => _expires != null && _expires.isBefore(DateTime.now());
+}
+
+/// Simple two-generational LRU cache inspired by:
+/// https://github.com/sindresorhus/quick-lru
+class InMemoryCacheProvider<T> extends CacheProvider<T> {
+  Map<String, _InMemoryEntry<T>> _new = {}; // new generation
```
sigurdm: Add types to Map literals
Use dartdoc comments

pkg/identity_codec/lib/identity_codec.dart
isoos: If all of them are using `List<int>`, then starting with `List<int>` seems to be better to me. I can barely imagine why anyone would use the `IdentityCodec` otherwise, it doesn't seem to be that useful to publish it as a separate package.

jonasfj: For the record, I'm hoping to land `IdentityCodec` in https://github.com/dart-lang/convert/pull/17, otherwise I'll probably publish elsewhere..

pkg/simple_cache/lib/simple_cache.dart
jonasfj: I correct myself: we need this to expose access to `close()`

pkg/simple_cache/lib/src/providers/inmemory.dart
jonasfj: default is no `ttl`, so no expiration happens until the system runs out of memory... or whatever eviction policy the cache provider has..

pkg/simple_cache/lib/simple_cache.dart
```diff
@@ -0,0 +1,219 @@
...
+  /// If given [ttl] specifies the time-to-live. Notice that this is advisatory,
+  /// the underlying [CacheProvider] may choose to evit cache entries at any
+  /// time. However, you should not expect entries to live far past their [ttl].
+  Future<T> set(T value, [Duration ttl]);
+
+  /// Clear the value stored in this cache entry.
+  Future purge();
+}
+
+class _Cache<T, V> implements Cache<T> {
```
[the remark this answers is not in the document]
jonasfj: I really want to hide the  `<T, V>` type parameters :)

pkg/simple_cache/lib/src/providers/redis.dart
```diff
@@ -0,0 +1,112 @@
...
+        // Create context
+        final client = Client(conn);
+        return _RedisContext(
+          connection: conn,
+          client: client,
+          commands: client.asCommands(),
+        );
+      }, retryIf: (e) => e is TimeoutException || e is SocketException);
+
+  Future<T> _withRedis<T>(_op<T> fn) async => _r.retry(() async {
```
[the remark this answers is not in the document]
jonasfj: maybe another day...

pkg/simple_cache/test/cacheprovider_test.dart
```diff
@@ -0,0 +1,107 @@
...
+import 'package:simple_cache/src/providers/redis.dart';
+import 'package:simple_cache/cache_provider.dart';
+import 'package:identity_codec/identity_codec.dart';
+import 'package:test/test.dart';
+
+/// Simple class to wrap a `CacheProvider<T>` + `Codec<String, T>` to get a
+/// `CacheProvider<String>`.
+///
+/// This is just meant to be useful for testing.
+class StringCacheProvider<T> implements CacheProvider<String> {
```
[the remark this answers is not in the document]
jonasfj: I also need to set stuff..

pkg/simple_cache/lib/simple_cache.dart
```diff
@@ -0,0 +1,219 @@
...
+  @override
+  Cache<S> withCodec<S>(Codec<S, T> codec) => _Cache(
+      _provider, _prefix, _logger, _ignoreErrors, codec.fuse(_codec), _ttl);
+
+  @override
+  Cache<T> withTTL(Duration ttl) =>
+      _Cache(_provider, _prefix, _logger, _ignoreErrors, _codec, ttl);
+}
+
+class _Entry<T, V> implements Entry<T> {
```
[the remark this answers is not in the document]
jonasfj: I want to hide `T` and `V`

pkg/simple_cache/lib/src/providers/inmemory.dart
```diff
@@ -0,0 +1,70 @@
+import 'dart:async';
+import '../../cache_provider.dart';
+
+class _InMemoryEntry<T> {
```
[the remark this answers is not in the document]
jonasfj: Maybe, it's really an entry in the inmemory cache..

app/lib/scorecard/scorecard_memcache.dart
```diff
@@ -7,10 +7,9 @@ import 'dart:convert' as convert;
 
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:logging/logging.dart';
-import 'package:memcache/memcache.dart';
 import 'package:meta/meta.dart';
 
-import '../shared/memcache.dart';
+import '../shared/rediscache.dart';
```
isoos: I think `redis_cache.dart` would be a better name for it.

app/lib/shared/rediscache.dart
```diff
@@ -0,0 +1,93 @@
...
+  ]);
+  // Validate that we got a connection string
+  if (s.isEmpty || !(s[0] is Secret) || (s[0] as Secret).value.isEmpty) {
+    throw Exception('Secret ${SecretKey.redisConnectionString} is missing');
+  }
+  final connectionString = (s[0] as Secret).value;
+
+  // Create and register a cache
+  final cacheProvider = Cache.redisCacheProvider(connectionString);
+  ss.register(#_cache, Cache(cacheProvider));
```
isoos: I hate it for the boilerplate, but I'd prefer to use the `registerCache(Cache cache)` pattern here too, as it is easier to scan for its uses.

app/lib/shared/rediscache.dart
```diff
@@ -0,0 +1,93 @@
...
+    // Call fn
+    return await fn();
+  } finally {
+    await cacheProvider.close();
+  }
+}
+
+/// Run [fn] with an in-memory cache for [cache].
+Future _withInmemoryCache(FutureOr Function() fn) async {
+  ss.register(#_cache, Cache(Cache.inMemoryCacheProvider(4096)));
```
isoos: See above.

app/lib/shared/rediscache.dart
```diff
@@ -0,0 +1,93 @@
...
+}
+
+class SimpleMemcache {
+  final Cache<String> _sCache;
+  final Cache<List<int>> _bCache;
+
+  SimpleMemcache(String prefix, Logger logger, Duration ttl)
+      : _sCache = cache
+            .withPrefix('$runtimeVersion/')
+            .withPrefix('$prefix/')
```
isoos: why not `.withPrefix('$runtimeVersion/$prefix/')`?

pkg/simple_cache/lib/src/providers/redis.dart
```diff
@@ -0,0 +1,101 @@
...
+  final RetryOptions _retryOptions;
+  final Duration _connectTimeLimit;
+  final Duration _commandTimeLimit;
+
+  Future<_RedisContext> _context;
+  bool _isClosed = false;
+
+  RedisCacheProvider(
+    this._connectionString, {
+    RetryOptions retryOptions = const RetryOptions(),
```
isoos: I'm a bit worried about the default settings. Normally we want to have a cache that answers the requests fast, and if I understand it correctly, these settings will (a) retry it many times, and (b) timeout after 5 seconds, which is way above the usual time for our non-cached request serving. I'd probably start with a 1-retry, <200ms setting first.

jonasfj: Good point, this is just caching, we should be relaxed about retrying.. and fail fast instead.

app/lib/shared/rediscache.dart
jonasfj: I could care less, it's just internal boiler plate...

jonasfj: besides the symbol is internal, so it can be set outside this file...

pkg/retry/lib/retry.dart
```diff
@@ -0,0 +1,164 @@
...
+  });
+
+  /// Delay after [attempt] number of attempts.
+  ///
+  /// This is computed as `pow(2, attempt) * delayFactor`, then is multiplied by
+  /// between `-randomizationFactor` and `randomizationFactor` at random.
+  Duration delay(int attempt) {
+    assert(attempt >= 0, 'attempt cannot be negative');
+    if (attempt <= 0) {
+      return Duration();
```
isoos: `Duration.zero`

pkg/retry/lib/retry.dart
```diff
@@ -0,0 +1,164 @@
...
+    }
+    final rf = (randomizationFactor * (_rand.nextDouble() * 2 - 1) + 1);
+    var delay = math.min(
```
isoos: I think something like this would be more compact:
```
final delay = (math.pow(2, attempt) * delayFactor* rf);
return delay > maxDelay ? maxDelay : delay;
```

pkg/retry/lib/retry.dart
```diff
@@ -0,0 +1,164 @@
...
+  ///
+  /// If [randomizationFactor] is `0.25` (default) this indicates 25 % of the
+  /// delay should be increased or decreased by 25 %.
+  final double randomizationFactor;
+
+  /// Maximum delay between retries, defaults to 30 seconds.
+  final Duration maxDelay;
+
+  /// Maximum number of attempts before giving up, defaults to 8.
+  final int attempts;
```
isoos: `maxAttempt` would be a better name

pkg/retry/lib/retry.dart
```diff
@@ -0,0 +1,164 @@
...
+  /// function [fn] will be invoked at-most [this.attempts] times.
+  Future<T> retry<T>(
+    FutureOr<T> Function() fn, {
+    FutureOr<bool> Function(Exception) retryIf,
+    FutureOr<void> Function(Exception) onRetry,
+  }) async {
+    int attempt = 0;
+    // ignore: literal_only_boolean_expressions
+    while (true) {
+      attempt += 1; // first invocation is the first attempt
```
isoos: `++`

pkg/retry/lib/retry.dart
```diff
@@ -0,0 +1,164 @@
...
+///   retryIf: (e) => e is SocketException || e is TimeoutException,
+/// );
+/// print(response.body);
+/// ```
+Future<T> retry<T>(
+  FutureOr<T> Function() fn, {
+  Duration delayFactor = const Duration(milliseconds: 200),
+  double randomizationFactor = 0.25,
+  Duration maxDelay = const Duration(seconds: 30),
+  int attempts = 8,
```
isoos: `maxAttempts`

pkg/simple_cache/lib/simple_cache.dart
```diff
@@ -0,0 +1,220 @@
...
+/// A [Cache] can be _fused_ with a [Codec] using [withCodec] to get a cache
+/// that stores a different kind of objects. It is also possible to create
+/// a chuild cache using [withPrefix], such that all entries in the child
+/// cache have a given prefix.
+abstract class Cache<T> {
+  /// Get [Entry] wrapping data cached at [key].
+  Entry<T> operator [](String key);
+
+  /// Get a [Cache] wrapping of this cache with given [prefix].
+  Cache<T> withPrefix(String prefix);
```
isoos: I'm wondering: it seems that we'll use `withPrefix`, `withTtl`, `withCodec` together most of the time. Maybe a single method of `fork({prefix, ttl, codec, logger})` would make this simpler?

pkg/simple_cache/lib/cache_provider.dart
```diff
@@ -0,0 +1,58 @@
...
+  ///
+  /// When given [ttl] is advisory, however, implementers should avoid returning
+  /// entries that are far past their [ttl].
+  Future<void> set(String key, T value, [Duration ttl]);
+
+  /// Clear value stored at [key].
+  ///
+  /// After this has returned future calls to [get] for the given [key] should
+  /// not return any value, unless a new value have been set.
+  Future<void> purge(String key);
```
isoos: `purge` feels to me like an operation that affects the entirety of the cache. Maybe `delete` could be better name here (it is also more closer to the collections API).

pkg/retry/lib/retry.dart
jonasfj: Oh, I had a hard time figuring out of these operators was implemented...

pkg/simple_cache/lib/simple_cache.dart
jonasfj: Maybe... We could add a `wrap({prefix, ttl, codec, logger})` or `child({...`, but let's see what the usage patterns are like first... we can easily refactor this later...

pkg/simple_cache/lib/cache_provider.dart
jonasfj: `clear` is the word we used before, but really you are purging a cache entry.

pkg/simple_cache/pubspec.yaml
```diff
@@ -0,0 +1,19 @@
...
+environment:
+  sdk: '>=2.0.0 <3.0.0'
+
+dependencies:
+  logging: '^0.11.3'
+  identity_codec:
+    path: ../identity_codec
+  retry:
+    path: ../retry
+  dartis: ^0.3.0
```
isoos: Let's pin this to the specific `0.3.0` version.

isoos:
Could you please also write a line or two about the setup requirements of (a) Redis on AppEngine (e.g. a link to a how-to) (b) configuring in the app (pointer to the `set_secret` tool)?

jonasfj:
I added a tiny note to the `README.md` pointing to `doc/secrets.md`.

We probably should rewrite the `README.md` file and document things like `mono_repo`, etc...

tool/travis.sh
```diff
@@ -32,7 +32,7 @@ while (( "$#" )); do
   test) echo
     echo -e '\033[1mTASK: test\033[22m'
     echo -e 'pub run test --run-skipped'
-    pub run test --run-skipped -j 1 || EXIT_CODE=$?
```
isoos: Can you run this a couple of times on Travis? A few months ago it was still flaky, but I think we may have upgraded the Dart SDK since then.

tool/travis.sh
```diff
@@ -12,7 +12,7 @@ if [ "$#" == "0" ]; then
 fi
 
 pushd $PKG
-pub get || exit $?
+pub upgrade || exit $?
```
isoos: Please revert this, let's just use the versions in `pubspec.lock`.

jonasfj: Can we do this in a way that is maintained by `mono_repo`

I would like to avoid modifying auto-generated files..

jonasfj: but fine... let's just do it, we can cleanup CI later :)

isoos: It is not just the CI vs auto-generated files, but also the baked-in versions we use in prod vs. the versions we are testing against. E.g. in the past we needed to pin `analyzer` to specific versions, and I think it is better to stick with the `pubspec.lock` versions and do a controlled upgrade. Such an issue could be hidden if we blindly do `pub upgrade`.

PR #1854, opened 2018-12-06
Remove json silly

kevmoo:


kevmoo:
I was using that analysis HTTP endpoint for some...analysis. So 😝

isoos:
Ah, sorry for that. Does /api/packages/pana/metrics?reports&pretty help in any way?

kevmoo:
Ah yes it does! Thanks

PR #1855, opened 2018-12-07
Verification e-mail.

isoos:
#1759

### DIFF
--- a/app/lib/shared/email.dart
+++ b/app/lib/shared/email.dart
@@ -118,11 +118,11 @@ Pub Site Admin
 }
 
 /// Creates the [EmailMessage] that will be sent to the new uploader for verification.
-EmailMessage createUploaderVerificationEmail({
+EmailMessage createUploaderConfirmationEmail({
   @required String packageName,
   @required String activeAccountEmail,
   @required String addedUploaderEmail,
-  @required String verificationUrl,
+  @required String confirmationUrl,
 }) {
   final subject = 'Pub invitation to collaborate: $packageName';
   final bodyText = '''Dear package maintainer,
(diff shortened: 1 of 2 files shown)


app/lib/shared/email.dart
```diff
@@ -114,6 +114,31 @@ https://github.com/dart-lang/pub-dartlang-dart/issues
 Pub Site Admin
 ''';
 
-  return new EmailMessage(
-      _defaultFrom, authorizedUploaders, subject, bodyText.toString());
+  return new EmailMessage(_defaultFrom, authorizedUploaders, subject, bodyText);
+}
+
+/// Creates the [EmailMessage] that will be sent to the new uploader for verification.
+EmailMessage createUploaderVerificationEmail({
```
jonasfj: Can we call it a `Confirmation` instead of a `Verification`

app/test/shared/email_test.dart
```diff
@@ -135,4 +135,35 @@ void main() {
...
+        verificationUrl:
+            'https://pub.dartlang.org/verification/add-uploader/abcdef1234567890',
+      );
+      expect(message.from.toString(), 'Pub Site Admin <pub@dartlang.org>');
+      expect(message.recipients.map((e) => e.toString()).toList(),
+          ['uploader@example.com']);
+      expect(message.subject, 'Pub invitation to collaborate: pkg_foo');
+      expect(
+          message.bodyText,
+          'Dear package maintainer,\n'
```
jonasfj: Do we really want to maintain tests like this?

Wouldn't be easier to just verify that `bodyText` and `subject` contains `pkg_foo` and `abcdef1234567890`, as sanity checks. This other things is just going to break whenever someone changes the email.

isoos: The point of such tests (both in pana and on the pub site, e.g. golden files) is that we are aware of the extent of the change. It also help us to double-check the results if the templates go more complex.

sigurdm: I made a small function to maintain goldens in a file: 
https://github.com/dart-lang/dart-protoc-plugin/blob/master/test/golden_file.dart
To update the file, just delete it and run again.

PR #1858, opened 2018-12-10
Link to the uploaded package version in notication e-mail.

isoos:


isoos:
> We could consider also including the new version number in the clear text...

It is in the subject of the email. Shall I just keep it in "new version of package [pkg] [version]:" just before the URL?

PR #1860, opened 2018-12-10
PackageInvite: activation and confirmation.

isoos:
#1759

### DIFF
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -183,7 +183,7 @@ class Backend {
       // Existing and active invite with throttled notification.
       if (invite != null && !invite.isExpired() && !invite.shouldNotify()) {
         await tx.rollback();
-        return new InviteStatus(delayDuration: invite.delayDuration);
+        return new InviteStatus(nextNotification: invite.nextNotification);
       }
 
       // Existing and active invite with notification enabled.
@@ -278,12 +278,12 @@ class Backend {
 /// The status of an invite after being created or updated.
 class InviteStatus {
   final String urlNonce;
-  final Duration delayDuration;
+  final DateTime nextNotification;
 
-  InviteStatus({this.urlNonce, this.delayDuration});
+  InviteStatus({this.urlNonce, this.nextNotification});
 
   bool get isActive => urlNonce != null;
-  bool get isDelayed => delayDuration != null;
+  bool get isDelayed => nextNotification != null;
 }
 
 /// A read-only implementation of [PackageRepository] using the Cloud Datastore
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -6,11 +6,13 @@ library pub_dartlang_org.backend;
 
 import 'dart:async';
 import 'dart:io';
+import 'dart:math';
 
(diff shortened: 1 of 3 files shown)


isoos:
I slightly prefer this one over #1848. I feared that it will be more complex, but it ended up not too bad.

app/lib/frontend/backend.dart
```diff
@@ -159,6 +162,119 @@ class Backend {
     assert(repository.supportsDownloadUrl);
     return repository.downloadUrl(package, version);
   }
+
+  /// Stores a [models.PackageInvite] entry in the Datastore and returns its
+  /// urlNonce can be used in client-communication, e.g. sending via e-mail.
+  ///
+  /// Returns null if invite has been updated recently and at the moment no
```
sigurdm: Maybe null is not enough information? Should we not send a message back to the client to display?

app/lib/frontend/models.dart
```diff
@@ -239,6 +239,54 @@ abstract class SecretKey {
...
+
+  /// Create a composite id.
+  static String createId(String type, String recipientEmail) =>
+      '$type/$recipientEmail';
+
+  bool isExpired() => new DateTime.now().toUtc().isAfter(expires);
+
+  /// Whether a new notification should be sent.
+  bool shouldNotify() {
+    final delay = Duration(minutes: 1 << notificationCount);
```
sigurdm: Can this overflow and wrap around to become small again?

sigurdm: Ah. So we only create a new notification *after* the timeout is over. :)
Otherwise it would be easy.

app/lib/frontend/backend.dart
sigurdm: I do not have enough overview of the program structure to answer that. I just think it might be nice to answer the client with why no new mail has been sent, and when the wait is over.

isoos: PTAL. I've added a return value class that has enough information to communicate both the success or the amount of wait.

app/lib/frontend/backend.dart
```diff
@@ -159,6 +162,128 @@ class Backend {
...
+    await for (var invite in query.run()) {
+      db.commit(deletes: [invite.key]);
+    }
+  }
+}
+
+/// The status of an invite after being created or updated.
+class InviteStatus {
+  final String urlNonce;
+  final Duration delayDuration;
```
sigurdm: Should this be a timestamp of the actual availability (creationTime + delayDuration)?

isoos: I was thinking that the message to the client would be "You need to wait x more minutes before re-sending the invite.", but as they are almost equivalent, we can use timestamp too. Which do you prefer?

sigurdm: Ah - I read the code wrong - it is already calculated relative to now.
Then go with whatever suits you better.

app/lib/frontend/models.dart
isoos: I'm going to change this and the InviteStatus to `DateTime nextNotification`, as it would clear the ambiguity of the piece of code.

PR #1862, opened 2018-12-11
Uploader confirmation

isoos:
#1759

While a few parts missing, I wanted to start a discussion about it:
- There is no invalid code / expired invite error message, we just fall back on the default 404.
- There is no second click to actually confirm and/or check the id of the current user.
- Tests: the current (failing) tests with this change seems to be not useful anymore, I lean on removing them entirely. This is the first time when using `package:pub_server`'s internals seems to be in our way.

I think this is the bare minimum that should be done for the feature, and we can add the missing pieces on top as needed. WDYT?

### DIFF
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -211,7 +211,18 @@ class Backend {
         ..expires = now.add(Duration(days: 1))
         ..notificationCount = 1
         ..lastNotified = now;
-      tx.queueMutations(inserts: [invite]);
+
+      final inserts = <Model>[invite];
+      if (historyBackend.isEnabled) {
+        final history = new History.entry(new UploaderInvited(
+          packageName: packageName,
+          currentUserEmail: fromEmail,
+          uploaderUserEmail: recipientEmail,
+        ));
+        inserts.add(history);
+      }
+
+      tx.queueMutations(inserts: inserts);
       await tx.commit();
       return new InviteStatus(urlNonce: invite.urlNonce);
     }) as InviteStatus;
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -601,7 +601,12 @@ class GCloudPackageRepository extends PackageRepository {
       final packageKey = db.emptyKey.append(models.Package, id: packageName);
       final models.Package package = (await db.lookup([packageKey])).first;
 
-      _validateNewUploader(userEmail, package, uploaderEmail);
+      _validateActiveUser(userEmail, package);
+
+      if (package.hasUploader(uploaderEmail)) {
+        // The requested uploaderEmail is already part of the uploaders.
+        return;
+      }
 
       final status = await backend.updatePackageInvite(
         packageName: packageName,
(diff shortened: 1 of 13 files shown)


app/lib/frontend/backend.dart
```diff
@@ -591,53 +597,103 @@ class GCloudPackageRepository extends PackageRepository {
...
+      final packageKey = db.emptyKey.append(models.Package, id: packageName);
+      final models.Package package = (await db.lookup([packageKey])).first;
 
-        // Fail if package doesn't exist.
-        if (package == null) {
-          await T.rollback();
-          throw new GenericProcessingException(
-              'Package "$package" does not exist');
-        }
+      _validateNewUploader(userEmail, package, uploaderEmail);
```
jonasfj: This call isn't idempotent..., why not?

If the person being added as uploader is already added, why don't we just return 200 ok?

If we're trying to make restful APIs we should aim for idempotency.

----
I know this changes semantics from what they used to be. But is there a reason we didn't have idempotent APIs?

app/lib/frontend/backend.dart
```diff
@@ -591,53 +597,103 @@ class GCloudPackageRepository extends PackageRepository {
...
+        email: uploaderEmail,
+        urlNonce: status.urlNonce,
+      );
+      final message = createUploaderConfirmationEmail(
+        packageName: packageName,
+        activeAccountEmail: userEmail,
+        addedUploaderEmail: uploaderEmail,
+        confirmationUrl: confirmationUrl,
+      );
+      await emailSender.sendMessage(message);
```
jonasfj: Don't we need a history event to register that an invitation was sent out?

app/lib/frontend/backend.dart
isoos: Ah. You are right, we could do a silent success without sending an email.

app/lib/frontend/backend.dart
isoos: Good point. I guess it could be part of the transaction of the `updatePackageInvite`, or would you like to have it after the email is sent?

app/lib/frontend/backend.dart
```diff
@@ -591,53 +597,103 @@ class GCloudPackageRepository extends PackageRepository {
...
-          final history = new History.entry(new UploaderChanged(
-            packageName: packageName,
-            currentUserEmail: userEmail,
-            addedUploaderEmails: [uploaderEmail],
-          ));
-          inserts.add(history);
-        }
+  Future confirmUploader(
+      String userEmail, String packageName, String uploaderEmail) async {
+    return db.withTransaction((Transaction T) async {
```
sigurdm: Nit: why not use tx as above?

isoos: I kept the old code for preserving more parts of the history, but happy to rename it.

app/lib/frontend/backend.dart
jonasfj: Before sending the email... ideally, before creation the invitation entry... since dying after we make the invitation just shows that someone tried to send an invitation...

isoos: Adding it in the transaction where we create/update the invitation entry. With that we will have a separate history entry for each time we wanted to send out the email.

isoos:
In the end I managed to fix/extend the test to work with the new logic.

PR #1886, opened 2018-12-17
Minimal e-mail validation for adding an uploader.

isoos:


app/lib/shared/email.dart
```diff
@@ -59,6 +57,20 @@ class EmailAddress {
   }
 }
 
+/// Minimal accepted format for an email.
+bool isValidEmail(String email) {
```
jonasfj: Maybe add:
```dart
///
/// This function only disallows the most obvious non-email strings.
/// It still allows strings that aren't legal email  addresses.
```

app/lib/shared/email.dart
```diff
@@ -59,6 +57,20 @@ class EmailAddress {
...
+  if (email == null) return false;
+  if (email.length < 5) return false;
+  if (!email.contains('@')) return false;
+  final atParts = email.split('@');
+  if (atParts.length != 2) return false;
+  if (atParts.any((s) => s.trim().isEmpty)) return false;
+  if (!atParts[1].contains('.')) return false;
+  final dotParts = atParts[1].split('.');
+  if (dotParts.any((s) => s.trim().isEmpty)) return false;
+  return true;
```
jonasfj:
```dart
var _emailPattern = RegExp(r'^[^@]+@[^@]+\.[^@]+$');
bool isValidEmail(String email) => _emailPattern.hasMatch(email);
```
Or even simpler, but not exactly the same:
```dart
var _emailPattern = RegExp(r'^.+@.+\..+$');
bool isValidEmail(String email) => _emailPattern.hasMatch(email);
```

isoos: good point, although I'd go with non-whitespace chars instead of the `.`

Issue #1881, opened 2018-12-16
Solution for pub's Health issues and suggestions


hyochan:
![image](https://user-images.githubusercontent.com/27461460/50053604-79592000-017b-11e9-8456-ddf7cf8ca7ab.png)
Recently, I'm seeing the `Document public APIs issue in pub page. How can I solve this -9.86 points?

Our plugin in is [flutter_calendar_carousel](https://pub.dartlang.org/packages/flutter_calendar_carouse).

isoos:
> is it required that all public members are documented?

Right now we show that setting if the documented API members are <20%.

> When looking at the source I can see that there is a lot of trivial members, and documenting them seems plain overkill

One could argue that trivial for one developer may not be trivial to another.

Having said that, in this case it seems that using full English words and CamelCase names should be a good-enough substitute for documentation.

hyochan:
@jonasfj  I've written about `defaultValues` in [Props](https://github.com/dooboolab/flutter_calendar_carousel#props) table which seems not sufficient in `pub`. I am still confused. What is the solution currently to fix `-9.86 points`? What I've got here are 
1. listing `contributors`
2. removing `print` message
3. writing up `defaultValues` separately.
Are these correct? Am I understanding correctly?

isoos:
> What is the solution currently to fix -9.86 points?

You need to put dartdoc on fields (or at least 20% of them), e.g.
````dart
  /// The border color of the previous day of the month.
  final Color prevMonthDayBorderColor;
````

Or, you could ignore it for a while, because (likely in January) we could create a new heuristic, that would accept field names like the above without documentation (because they seem to be descriptive enough).

I would suggest that you just ignore it for a while.

hyochan:
Thanks. I'll come back `2019`. 🥺

isoos:
@hyochan: The bad news is: after a bit of experimentation, we have postponed our plans to automatically detect 'self-describing' API symbols. While many of your getters and methods are descriptive, we will consider only proper `/// comments as valid documentation`.

The good news part is that we are decreasing the penalty, and in your case, it won't be `-9.86`, rather around `-0.7` (in a near-future release). **To eliminate it, please document a few more API members (methods, fields, classes).**

Note: our aim is to encourage people to document their package, and in the future we may slowly increase the thresholds for penalties.

hyochan:
[...]

PR #1887, opened 2018-12-17
Self-documenting API symbols count as documented symbols.

isoos:
This makes @filiph's `english_words` package our dependency, and we will use the top 5000 word as a filter for checking whether an API symbol is made of English words. The words are inlined Dart code (const lists).

https://github.com/dart-lang/pub-dartlang-dart/issues/1881

### DIFF
--- a/app/lib/dartdoc/dartdoc_runner.dart
+++ b/app/lib/dartdoc/dartdoc_runner.dart
@@ -15,6 +15,7 @@ import '../job/backend.dart';
 import '../job/job.dart';
 import '../scorecard/backend.dart';
 import '../scorecard/models.dart';
+import '../search/text_utils.dart' show isSelfDocumenting;
 import '../shared/configuration.dart' show envConfig;
 import '../shared/tool_env.dart';
 import '../shared/urls.dart';
@@ -231,12 +232,26 @@ class DartdocJobProcessor extends JobProcessor {
     double coverageScore = 0.0;
     if (hasContent && dartdocData != null) {
       final total = dartdocData.apiElements.length;
-      final documented = dartdocData.apiElements
-          .where((elem) =>
-              elem.documentation != null &&
-              elem.documentation.isNotEmpty &&
-              elem.documentation.trim().length >= 5)
-          .length;
+
+      bool isDocumented(ApiElement elem) {
+        // Normal documentation.
+        if (elem.documentation != null &&
+            elem.documentation.isNotEmpty &&
+            elem.documentation.trim().length >= 5) {
+          return true;
+        }
+
+        // Whether the symbol name is descriptive enough to substitute documentation.
+        final symbol = elem.name.split('.').last;
+        if (isSelfDocumenting(symbol)) {
+          return true;
+        }
+
+        // Otherwise: no documentation.
+        return false;
+      }
+
+      final documented = dartdocData.apiElements.where(isDocumented).length;
       if (total == documented) {
         // this also handles total == 0
         coverageScore = 1.0;
--- a/app/lib/search/text_utils.dart
+++ b/app/lib/search/text_utils.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:english_words/english_words.dart' as english_words;
 import 'package:html/parser.dart';
 
 import '../shared/markdown.dart';
@@ -87,20 +88,8 @@ Map<String, double> tokenize(String originalText) {
 
     tokens[normalizedWord] = 1.0;
 
-    // Scan for CamelCase phrases and extract Camel and Case separately.
-    final changeIndex = <int>[0];
-    bool prevLower = _isLower(word[0]);
-    for (int i = 1; i < word.length; i++) {
-      final bool lower = _isLower(word[i]);
-      if (!lower && prevLower) {
-        changeIndex.add(i);
-      }
-      prevLower = lower;
-    }
-    changeIndex.add(word.length);
-    for (int i = 1; i < changeIndex.length; i++) {
-      final token = normalizeBeforeIndexing(
-          word.substring(changeIndex[i - 1], changeIndex[i]));
+    for (String extracted in extractCamelCase(word)) {
+      final token = normalizeBeforeIndexing(extracted);
       final double weight = token.length / word.length;
       if ((tokens[token] ?? 0.0) < weight) {
         tokens[token] = weight;
@@ -110,7 +99,42 @@ Map<String, double> tokenize(String originalText) {
   return tokens;
 }
 
+// Scan for CamelCase phrases and extract Camel and Case separately.
+Iterable<String> extractCamelCase(String word) sync* {
+  final changeIndex = <int>[0];
+  bool prevLower = _isLower(word[0]);
+  for (int i = 1; i < word.length; i++) {
+    final bool lower = _isLower(word[i]);
+    if (!lower && prevLower) {
+      changeIndex.add(i);
+    }
+    prevLower = lower;
+  }
+  changeIndex.add(word.length);
+  for (int i = 1; i < changeIndex.length; i++) {
+    final token = word.substring(changeIndex[i - 1], changeIndex[i]);
+    yield token;
+  }
+}
+
 bool _isLower(String c) => c.toLowerCase() == c;
+
+/// Returns whether an API symbol can be considered as self-documenting.
+///
+/// When an API symbol uses common english words for most of its parts, we
+/// assume that it is descriptive enough for most use-case.
+bool isSelfDocumenting(String symbol) {
+  final words = extractCamelCase(symbol).map((s) => s.toLowerCase()).toSet();
+  final englishLength = words
+      .where(
+        (word) =>
+            english_words.all.contains(word) ||
+            english_words.all.contains(word.toLowerCase()),
+      )
+      .map((word) => word.length)
+      .fold<int>(0, (a, b) => a + b);
+  return englishLength > 10 || englishLength > (symbol.length * 0.5).floor();
+}
 
 /// Generates the N-grams of [input], which are continuous character strings of
 /// length between [minLength] and [maxLength] (both inclusive).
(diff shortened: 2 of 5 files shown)


app/lib/search/text_utils.dart
```diff
@@ -110,8 +99,43 @@ Map<String, double> tokenize(String originalText) {
...
+/// Returns whether an API symbol can be considered as self-documenting.
+///
+/// When an API symbol uses common english words for most of its parts, we
+/// assume that it is descriptive enough for most use-case.
+bool isSelfDocumenting(String symbol) {
+  final words = extractCamelCase(symbol).map((s) => s.toLowerCase()).toSet();
+  final englishLength = words
+      .where(
+        (word) =>
+            english_words.all.contains(word) ||
```
sigurdm: I think `english_words.all` is a list. Probably this would be faster if you convert it to a Set<String> first.

app/test/search/text_utils_test.dart
```diff
@@ -105,7 +105,25 @@ Other useful methods will be added soon...
...
+      expect(extractCamelCase('CamelCaseWord').toList(),
+          ['Camel', 'Case', 'Word']);
+      expect(extractCamelCase('firstLowerCase').toList(),
+          ['first', 'Lower', 'Case']);
+      expect(extractCamelCase('snake_case_word').toList(), ['snake_case_word']);
+    });
+
+    test('isSelfDocumenting positive examples', () {
+      expect(isSelfDocumenting('todayBorderColor'), true);
+      expect(isSelfDocumenting('weekendTextStyle'), true);
```
sigurdm: I would like these tests to have more examples to be able to see where the border goes between descriptive and nondescriptive names.

isoos:
> This would force the library writer to consider documenting all members, but avoid having stupid documentation that just repeats the name:

I think once people figure out the minimum length we consider as a documentation, they'd rather write something like this instead of the well-thought-out `ignore: no_doc_comment`:

````
/// TODO: [+optional letters]
method()
````

There are way too many ways to game this (e.g. copy-paste the `ignore:` line all-over the place without thinking, adding a dummy class with 1000 methods that all have good documentation), and I think first we should try to figure out the ease the trivial burden.

> such that the first documentation string would give you a lot of points, and the last point is achieved when all members have been documented.

While this PR touches the overall scoring part, it just refactors it. The focus is whether a field name in itself can be considered descriptive enough. I think we should have a good approximation for it, even if it is imperfect.

File: app/test/search/text_utils_test.dart (hunk 2 above)
Comment (isoos) [reply app/test/search/text_utils_test.dart]: I've added a few more test cases. The boundary also depends on whether the word was popular enough to get into the top 5000, or less so (e.g. if it is a technical term). I've considered to use a broader dictionary (100k+ English words), but I'm not sure if rare words would be as descriptive on their own as frequently used ones.

Comment (sigurdm) [reply app/test/search/text_utils_test.dart]: My feeling is that they are less generic and therefore more descriptive.
theObject says almost nothing, while randomizedDictionary gives some clue of the purpose.

File: app/test/search/text_utils_test.dart [hunk 3]
```diff
@@ -105,7 +105,36 @@ Other useful methods will be added soon...
...
+      expect(isSelfDocumenting('todayBorderColor'), true);
+      expect(isSelfDocumenting('updateReport'), true);
+      expect(isSelfDocumenting('weekendTextStyle'), true);
+    });
+
+    test('isSelfDocumenting negative examples', () {
+      expect(isSelfDocumenting('abcdef'), false);
+      expect(isSelfDocumenting('AbcDef'), false);
+      expect(isSelfDocumenting('deriveLookupCandidates'), false);
+      expect(isSelfDocumenting('isSelfDocumenting'), false);
```
sigurdm: I would expect this to be considered self-documenting

app/test/search/text_utils_test.dart
```diff
@@ -105,7 +105,36 @@ Other useful methods will be added soon...
...
+      expect(isSelfDocumenting('latestSemanticVersion'), false);
```
sigurdm: And this

app/test/search/text_utils_test.dart
```diff
@@ -105,7 +105,36 @@ Other useful methods will be added soon...
...
+      expect(isSelfDocumenting('readmeFilename'), false);
```
sigurdm: And this

app/lib/search/text_utils.dart
```diff
@@ -110,8 +102,39 @@ Map<String, double> tokenize(String originalText) {
...
+    prevLower = lower;
+  }
+  changeIndex.add(word.length);
+  for (int i = 1; i < changeIndex.length; i++) {
+    final token = word.substring(changeIndex[i - 1], changeIndex[i]);
+    yield token;
+  }
+}
+
 bool _isLower(String c) => c.toLowerCase() == c;
```
sigurdm: Not sure if this is a bottle-neck, but considering that dart-identifiers are ascii-only this could done with:
```dart
  const lowerA = 97;
  const lowerZ = 122;
  bool _isLower(int char) => lowerA <= char && char <= lowerZ;
```

app/test/search/text_utils_test.dart
isoos: I think what we don't want to encourage is to use the same words in the documentation that were already used in the API symbol. Such documentation feel trivial and unnecessary.

When an API symbol is already a commonly used word, chances are that its documentation won't be complex either. I understand that there will be false positives and false negatives in either case, and this can't be perfect, however, for the developers, this will boil down to:

1) Does pub nags me to do trivial things? (for authors)
2) Can I understand this package easily? (for library users)

This is more about decreasing the probability of (1), while we don't really compromise (2). I think the linked issue (#1881) clearly falls into the first category, I don't feel that most of the fields need additional documentation there.

app/test/search/text_utils_test.dart
isoos: `documenting` is not a frequently used word

app/test/search/text_utils_test.dart
isoos: neither is `filename` or `readme`

sigurdm:
I still think developers would be confused by this kind of heuristic (I would) unless it really really closely corresponds to an intuitive sense of a descriptive name.

The comment on the scoring was meant to point at an alternative way of relieving the pain for package developers without having them fighting heuristics. (A little documentation brings you a long way on the scoring - but to really be considered fully documented it has to be ... fully documented)

app/test/search/text_utils_test.dart
sigurdm: No - that is my critique of the heuristic. The name 'is' self-documenting (in my opinion), but the predicate says it is not. I don't think matching with frequently used words catches the concept of self-documenting very well.

isoos:
> The comment on the scoring was meant to point at an alternative way of relieving the pain for package developers without having them fighting heuristics. (A little documentation brings you a long way on the scoring - but to really be considered fully documented it has to be ... fully documented)

That's a good point, and totally valid. But even with non-linear scoring, I think we should add some kind of heuristic to eliminate trivialities.

sigurdm:
I don't want to block this if you think it is the right thing to do - jonas is sick today - but let us wait and hear his opinion.

app/test/search/text_utils_test.dart
isoos: The worst case of these (false-negative) miss-classification is that we still consider a symbol non-documented, while in the real world we understand it perfectly. It doesn't effect the score either way, and if we improve on it in the future, it is a non-breaking experience for the site's users.

I would be more worried if you'd disagree with the positive examples.

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -231,12 +232,26 @@ class DartdocJobProcessor extends JobProcessor {
...
     if (hasContent && dartdocData != null) {
       final total = dartdocData.apiElements.length;
-      final documented = dartdocData.apiElements
-          .where((elem) =>
-              elem.documentation != null &&
-              elem.documentation.isNotEmpty &&
-              elem.documentation.trim().length >= 5)
-          .length;
+
+      bool isDocumented(ApiElement elem) {
```
jonasfj: Somewhat unrelated... but since we starting to grow the complexity around detecting documentation level..
How about we move all the logic into a function in a separate file... I mean isn't it enough to pass `apiElements` and have it return `Suggestion` or `null`...

```dart
/// Returns a [Suggestion] or `null`, if the documentation level is sufficient.
Suggestion documentationSuggestion(apiElements) {
  ...
}
```

jonasfj: With the other PR to make non-linear scores this code grows even more.

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -15,6 +15,7 @@ import '../job/backend.dart';
 import '../job/job.dart';
 import '../scorecard/backend.dart';
 import '../scorecard/models.dart';
+import '../search/text_utils.dart' show isSelfDocumenting;
```
jonasfj: Maybe we should move this file to `shared/` or split out the non-search specific parts and put them in `shared/`..

I know it's hard to get perfect, but if we could strive to avoid too many cross-dependencies that might be nice :)

app/test/search/text_utils_test.dart
jonasfj: > When an API symbol is already a commonly used word, chances are that its documentation won't be complex either.

I suppose the counter argument is that the more common a word is the less information in conveys. For example `ImageWidth` of what image? in what unit?
Where as a less common word is more likely to be a domain specific term, which requires less explanation for people who understand it, and more explanantion for people who don't.

I'm not going to propose that we invert the heuristic... just slowly starting to wonder if looking at the frequency of the words is a good heuristic at all.. If it says anything useful.

isoos:
Good points about splitting up this file, I'll close this PR and do the split first. (It also allows better testing).

> I suppose the counter argument is that the more common a word is the less information in conveys. For example ImageWidth of what image? in what unit?
Where as a less common word is more likely to be a domain specific term, which requires less explanation for people who understand it, and more explanantion for people who don't.

I believe we desperately need supporting data for a decision. I'll grab the top-N package's extract, and try to figure out a reasonable statistics about the used words, the overlap of documentation text with their symbol, and their comparison to the common word frequencies. Let's revisit this question after we have some kind of data.

PR #1889, opened 2018-12-18
Dartdoc coverage: separate, non-linear score.

isoos:
- As we want to have a non-linear score, `DartdocReport` should store both the coverage (percent) and the scores separately.
- I've added a non-linear score that should encourage the initial effort.
- We keep apply the `coverage < 0.1` threshold, e.g. a package with 20% dartdoc coverage won't get any penalties yet.

### DIFF
--- a/app/lib/dartdoc/dartdoc_runner.dart
+++ b/app/lib/dartdoc/dartdoc_runner.dart
@@ -5,6 +5,7 @@
 import 'dart:async';
 import 'dart:convert' as convert;
 import 'dart:io';
+import 'dart:math';
 
 import 'package:logging/logging.dart';
 import 'package:pana/pana.dart' hide Pubspec;
@@ -228,6 +229,7 @@ class DartdocJobProcessor extends JobProcessor {
       await toolEnvRef.release();
     }
 
+    double coverage = 0.0;
     double coverageScore = 0.0;
     if (hasContent && dartdocData != null) {
       final total = dartdocData.apiElements.length;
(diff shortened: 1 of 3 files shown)


app/lib/scorecard/models.dart
```diff
@@ -427,6 +429,7 @@ class DartdocReport implements ReportData {
 
   DartdocReport({
     @required this.reportStatus,
+    @required this.coverage,
```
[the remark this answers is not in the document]
isoos: This is one of the metrics which is interesting for the /api/.../metrics endpoint, even though we don't display it yet on other places.

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -239,15 +237,29 @@ class DartdocJobProcessor extends JobProcessor {
...
-            : SuggestionLevel.hint;
+      // 0.01 coverage -> 0.03 score -> 9.7 penalty
+      // 0.02 coverage -> 0.06 score -> 9.4 penalty
+      // 0.03 coverage -> 0.09 score -> 9.1 penalty
+      // 0.05 coverage -> 0.14 score -> 8.6 penalty
+      // 0.10 coverage -> 0.27 score -> 7.3 penalty
+      // 0.20 coverage -> 0.49 score -> 5.1 penalty
+      // 0.30 coverage -> 0.66 score -> 3.4 penalty
+      // 0.50 coverage -> 0.88 score -> 1.2 penalty
+      // 0.75 coverage -> 0.98 score -> 0.2 penalty
```
sigurdm: Add 
```
// 1.00 coverage -> 1.00 score -> 0.0 penality
```
?

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -239,15 +237,29 @@ class DartdocJobProcessor extends JobProcessor {
...
+      coverageScore = 1.0 - pow(1.0 - coverage, 3);
```
sigurdm: Should the `*10` be here to make it a score?

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -257,7 +269,7 @@ class DartdocJobProcessor extends JobProcessor {
               '$undocumented out of $total API elements (library, class, field '
               'or method) have no adequate dartdoc content. Good documentation '
               'improves code readability and discoverability through search.',
-              score: (1.0 - coverageScore) * 10.0),
+              score: reducedCoveragePenalty),
```
sigurdm: Should you not use a score here, not a penality?

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -257,7 +269,7 @@ class DartdocJobProcessor extends JobProcessor {
               '$undocumented out of $total API elements (library, class, field '
               'or method) have no adequate dartdoc content. Good documentation '
               'improves code readability and discoverability through search.',
```
sigurdm: We should perhaps update this comment to explain the non-linear scores

app/lib/dartdoc/dartdoc_runner.dart
isoos: ack

app/lib/dartdoc/dartdoc_runner.dart
isoos: That is only needed for the displayed penalty. I know that semantically both of these are scores, but we only have [0.0-100.0] in the penalty scores, otherwise we always use [0.0-1.0].

app/lib/dartdoc/dartdoc_runner.dart
isoos: ack

PR #1912, opened 2019-01-07
Issue 1900

### LINKED ISSUE #1900: Exclude dart2-incompatible packages from search
We should probably start excluding all dart2-incompatible packages from searches

jonasfj:
Fixes #1900

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -55,11 +55,12 @@ class SearchBackend {
   /// Packages are omitted when:
   ///  * latest version is missing,
   ///  * analysis is missing,
+  ///  * the package is not dart 2.0 compatible.
   Future<List<PackageDocument>> loadDocuments(List<String> packageNames) async {
     final List<Key> packageKeys = packageNames
         .map((String name) => _db.emptyKey.append(Package, id: name))
         .toList();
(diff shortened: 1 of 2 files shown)


app/lib/search/backend.dart
```diff
@@ -51,13 +51,16 @@ class SearchBackend {
 
   /// Loads the list of packages, their latest stable versions and returns a
   /// matching list of [PackageDocument] objects for search.
-  /// When a package, its latest version or its analysis is missing, the method
-  /// returns with null at the given index.
+  ///
```
isoos: I'd rather keep the return of null values. It follows the same consistency model as `DatastoreDB.lookup`: if you pass in 10 items, the result for the 3rd will be always in the given index, and you don't need to check if it is the same position as you've requested.

I'd prefer adding another flag to `PackageDocument` and post-filtering on it when needed.

jonasfj:
That's a good question... last I grepped through the logs unique we had less than 2.5k unique IPs downloading pub packages with a version less than 2.0 in the past 60 days..

@mit-mit, do we want to rid dart 1.x packages from all package listings making them undiscoverable unless you know the link?

We could also score them at the very end, regardless of relevancy..

jonasfj:
Post filtering is fine with me too :)

--
Regards Jonas Finnemann Jensen.

PR #1913, opened 2019-01-07
Mono repo

kevmoo:


app/pubspec.lock
```diff
@@ -674,4 +674,4 @@ packages:
     source: hosted
     version: "2.1.15"
 sdks:
-  dart: ">=2.1.0-dev.5.0 <3.0.0"
+  dart: ">=2.1.0 <3.0.0"
```
isoos: Let's just do `>=2.1.0-dev.9.4` (consistent with Dockerfile)

app/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: pub_dartlang_org
 author: Dart Team <misc@dartlang.org>
 description: The pub.dartlang.org website.
 environment:
-  sdk: ">=2.0.0 <3.0.0"
+  sdk: ">=2.1.0 <3.0.0"
```
isoos: `>=2.1.0-dev.9.4`

pkg/_popularity/mono_pkg.yaml
```diff
@@ -0,0 +1 @@
+# See https://github.com/dart-lang/mono_repo for details
```
isoos: Do we need these empty files around?

kevmoo: They allow `mono_repo --recursive pub get` to work...

kevmoo:
@isoos – PTAL

PR #1918, opened 2019-01-08
Filter legacy

### LINKED ISSUE #1900: Exclude dart2-incompatible packages from search

jonasfj:
Fix #1900.

It's possible we should be setting `includeLegacy: true`  when listing packages.


isoos:
> It's possible we should be setting includeLegacy: true when listing packages.

I'm a bit undecided on that. Maybe let's start with true for now, and do a switch later?

jonasfj:
Let's add a checkbox to search entry... that way there is visibility into what is happening here.

static/css/style.css
```diff
@@ -1194,7 +1194,7 @@ pre > code {
           align-items: center;
   border-radius: 3px;
   background: #f5f5f5;
-  max-width: 550px;
+  max-width: 650px;
```
isoos: why the change?

jonasfj: because there wasn't room for the two check boxes...

isoos: SG.

PR #1924, opened 2019-01-09
Adding minimal tests to make CI happy.

isoos:
packages without valid `test/` directories are failing, adding minimal (but valid) tests.


jonasfj:
Files #1925 for the cleanup... I really don't care when we get around to that...

PR #1929, opened 2019-01-09
Apply suggestions to emails (credits: @kwalrath)

jonasfj:


app/lib/shared/email.dart
```diff
@@ -118,16 +118,19 @@ EmailMessage createPackageUploadedEmail({
 }) {
   final url =
       pkgPageUrl(packageName, version: packageVersion, includeHost: true);
-  final subject = 'Package upload on pub: $packageName $packageVersion';
+  final subject = 'Package uploaded: $packageName $packageVersion';
   final bodyText = '''Dear package maintainer,
+Dear package maintainer,
```
isoos: This is a duplicate with the above line, remove one.

isoos:
Also: `email_test.dart` needs to be updated too.

jonasfj:
@isoos, I made tests a bit saner... keeping the full text only serves as a quick way to approve the contents... I might approve if we had neat framework that left a temp file I just needed to rename or something..

But in this setup the result isn't easy to see anyway.. and there is no easy way to fix this.. So I think a sanity check test is better..

isoos:
You'll need to update the backend_tests too.

PR #1948, opened 2019-01-15
Show hash link on hovering header rows.

### LINKED ISSUE #1947: Show link anchors when hovering over headers in markdown content
Original issue by @srawlins: https://github.com/dart-lang/pub-dartlang-dart/issues/1429

> Link anchors at headers can only be found by looking at source. GitHub, e.g. provides you that great mouse-over icon: https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md

At first I thought that this is not feasible to do in our `markdown.dart`, because the `<h1>` element did not have an `id` attribute, but it looks like the markdown API has a different field for it (`generatedId`).

We could use the `#` character or the Unicode [link symbol](https://unicode-table.com/en/1F517/) as the visible text of the link.

isoos:
Fixes #1947.

Displays the `#` character on hovering, which is clickable and links to the given row:

screen shot 2019-01-15 at 13 29 15

### DIFF
--- a/app/lib/shared/markdown.dart
+++ b/app/lib/shared/markdown.dart
@@ -25,12 +25,40 @@ String markdownToHtml(String text, String baseUrl) {
   final lines = text.replaceAll('\r\n', '\n').split('\n');
   final nodes = document.parseLines(lines);
 
-  final visitor = new _UrlRewriter(sanitizedBaseUrl);
+  final urlRewriter = _UrlRewriter(sanitizedBaseUrl);
+  final hashLink = _HashLink();
   for (final node in nodes) {
-    node.accept(visitor);
+    node.accept(urlRewriter);
+    node.accept(hashLink);
   }
 
   return m.renderToHtml(nodes) + '\n';
+}
+
+/// Adds an extra <a href="#hash">#</a> element to all h1,h2,h3..h6 elements.
+class _HashLink implements m.NodeVisitor {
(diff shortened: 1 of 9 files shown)


isoos:
Oh, forgot: I've deployed this to staging, you can try it out at any package page, which has a header in their readme.

app/lib/shared/markdown.dart
```diff
@@ -25,14 +25,42 @@ String markdownToHtml(String text, String baseUrl) {
...
+class _HashLink implements m.NodeVisitor {
+  @override
+  void visitText(m.Text text) {}
+
+  @override
+  bool visitElementBefore(m.Element element) => true;
+
+  @override
+  void visitElementAfter(m.Element element) {
+    if (!element.tag.startsWith('h') ||
```
sigurdm: Nit: I think this would read easier as:
```
if (!(element.tag.startsWith('h') &&
    element.tag.length == 2 &&
    element.generatedId != null &&
    element.children.length == 1)) {
  return;
}
```

sigurdm: I think the tag `<hr>` could also fall into this? Or it doesn't have children?
Perhaps better to test for membership in a set of 'h1' .. 'h6'.

isoos: Updated, PTAL.

Issue #1954, opened 2019-01-16
Searches includes packages that are marked discontinued


mit-mit:
For example, https://pub.dartlang.org/packages?q=email%3Amisc%40dartlang.org&page=15

This was not the original specified behavoir:
https://github.com/dart-lang/pub-dartlang-dart/issues/289#issuecomment-362585628

mit-mit:
![screenshot 2019-01-16 at 15 54 26](https://user-images.githubusercontent.com/13644170/51257020-0542af80-19a7-11e9-8671-ce393e4760d0.png)

isoos:
It looks like that the search index part of the `isDiscontinued` flag handling works, but there is no actual update of the index after the flag changes.

isoos:
I've debugged the related code with great details, and my current understanding is the following:

- Setting a package discontinued works as expected, with a small caveat: the `bin/tools/flag.dart` tool triggers an update only for the current (source-code) `runtimeVersion`, and if that differs from the deployed one, the update may happen later.

- The need to update the index's `PackageDocument` is picked up, the loading process recognizes it as discontinued, does not load it, marks it to-be-removed, and removes it from both the index and the snapshot.

Prod is now working as expected (at least for `docgen`). I've observed the issue on staging yesterday, the index was indeed not updated, but something changed today, and now it is updated correctly.

It may be related to caching. I've seen a similar glitch in the search result pagination caching: a certain page index was cached way longer than the intended 10 minutes, and it produced a really confusing result, as the index was updated behind it.

isoos:
This is now fixed at staging, and will be part of the next prod release.

PR #1956, opened 2019-01-16
Exclude discontinued packages from the search index

jonasfj:
Fixing #1954.


jonasfj:
hmm, maybe we still need to force rebuild the search index from scratch after this?
Could we do that by bumping the `runtimeVersion`?

isoos:
Actually, I think the following code will remove this from the package list:
https://github.com/dart-lang/pub-dartlang-dart/blob/master/app/lib/search/updater.dart#L121

It doesn't work with the removed packages because we don't have a similar way to know about the package deleted status.

PR #1959, opened 2019-01-17
Cleanup: Analysis removal.

isoos:
- removed `backend.dart` (not in use)
- updated batch size to appengine maximum (validated in staging: faster)
- added the 'deprecated' flag in the processing, because staging had an entry with that

After the merge, you should run the following script to remove the entries:
````
dart bin/tools/remove_all_analysis.dart 
````


isoos:
It took about 10-15 minutes to get the entries removed in staging.

PR #1996, opened 2019-01-28
Harden Content-Security-Policy headers: frame-src, script-src

isoos:
#1691

### DIFF
--- a/app/lib/shared/handlers.dart
+++ b/app/lib/shared/handlers.dart
@@ -59,6 +59,10 @@ shelf.Response htmlResponse(String content,
   headers ??= <String, String>{};
   headers['content-type'] = 'text/html; charset="utf-8"';
   headers['x-content-type-options'] = 'nosniff';
+  headers['content-security-policy'] = [
+    "frame-src 'none'",
+    "script-src 'self' https://www.googletagmanager.com/ https://www.google.com/ https://ajax.googleapis.com/",
+  ].join(';');
   return new shelf.Response(status, body: content, headers: headers);
 }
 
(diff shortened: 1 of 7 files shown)


app/lib/shared/handlers.dart
```diff
@@ -59,6 +59,10 @@ shelf.Response htmlResponse(String content,
   headers ??= <String, String>{};
   headers['content-type'] = 'text/html; charset="utf-8"';
   headers['x-content-type-options'] = 'nosniff';
+  headers['content-security-policy'] = [
+    "frame-src 'none'",
+    "script-src 'self' https://www.googletagmanager.com/ https://www.google.com/ https://ajax.googleapis.com/",
```
jonasfj: Maybe we should put these into a config file somewhere? Do we have a config file where the gtag is given?

Or at-least write them as a list...

isoos: I've updated the source of the policy, and also added a few more items that were otherwise blocked on the staging. Unfortunately we still need the `unsafe-eval` policy, because `dart2js` still uses eval().

app/lib/shared/handlers.dart
```diff
@@ -54,11 +54,30 @@ shelf.Response jsonResponse(
   );
 }
 
+final _contentSecurityPolicyMap = <String, List<String>>{
+  'frame-src': <String>["'self'"],
+  'script-src': <String>[
+    "'self'",
+    "'unsafe-eval'", // dart2js requires it
```
jonasfj: Let's file an issue for this... and put in the backlog, there is no reason to rush to doing it... but it's a nice thing to fix on a rainy day.

whesse: I think that the referenced issue just talks about a test that is written incorrectly, so it shouldn't affect anything except that particular test.

PR #2001, opened 2019-01-31
New account model and handling: User

isoos:
Fist step towards #1642

Couple of changes from the initial design decisions:

1) `User` has a composite `id`, in the format of: `dartlang-pub-1:<oauth2_user_id>`:
- Random UUID could work, but the create-or-update logic becomes complex, and potential race conditions may take longer to resolve.
- Single `<oauth_user_id>` is not recommended, because if we need to change the account of the OAuth2 flow, or if we want to have multiple such flows with different providers/accounts, it would risk key collision.

2) Storing both `email` (the last e-mail address reported for the account), and `emails` (all the email addresses that was reported to the account).

3) `lastAccess` is written outside of the lookup flow, in order to minimize the lookup latency.

I've changed the current token lookup logic to use this new method, and with that we'll start migrating active users automatically.


app/lib/account/backend.dart
```diff
@@ -0,0 +1,180 @@
...
+  AccountBackend(this._db);
+
+  Future close() async {
+    await _defaultAuthProvider.close();
+  }
+
+  Future<User> lookupOrCreateUser(String accessToken) async {
+    final auth = await _defaultAuthProvider.tryAuthenticate(accessToken);
+    if (auth == null) {
+      return null;
```
sigurdm: The docstring should probably explain what it means that `null` is returned.

app/lib/account/backend.dart
```diff
@@ -0,0 +1,180 @@
...
+      User user = (await _db.lookup<User>([userKey])).single;
+      if (user == null) {
+        // create user
+        user = User()
+          ..parentKey = _db.emptyKey
+          ..id = compositeId
+          ..authProvider = auth.provider
+          ..authUserId = auth.userId
+          ..email = auth.email
+          ..emails = <String>[auth.email]
```
sigurdm: Perhaps this field can have a more descriptive name: `allHistoricalEmails` or similar?

app/lib/account/backend.dart
```diff
@@ -0,0 +1,180 @@
...
+          ..created = accessTime
+          ..updated = accessTime;
+        _db.commit(inserts: [user]);
+      } else {
+        // update user if e-mail has been changed
+        if (user.updateEmail(auth.email)) {
```
sigurdm: having the updateEmail both mutate and return `changed` is a bit confusing to me.
Would it not be cleaner to write:
```
  if (user.email != auth.email) {...}
```
You are anyway repeating the updateEmail call below for the looked up user.

app/lib/account/backend.dart
```diff
@@ -0,0 +1,180 @@
...
+        }
+      }
+      return user;
+    });
+
+    if (user == null) {
+      return null;
+    }
+
+    if (!_pool.isClosed) {
```
sigurdm: What does it mean that the pool is closed?
Who would ever close it?
Can we really skip the database update in that case?

app/lib/account/backend.dart
```diff
@@ -0,0 +1,180 @@
...
+              tx.queueMutations(inserts: [user]);
+              await tx.commit();
+            } else {
+              await tx.rollback();
+            }
+          });
+        });
+      });
+    }
+
```
sigurdm: Redundant newline?

app/lib/account/models.dart
```diff
@@ -0,0 +1,69 @@
...
+  bool updateEmail(String email) {
+    bool changed = false;
+
+    if (this.email != email) {
+      this.email = email;
+      changed = true;
+    }
+    if (emails == null || !emails.contains(email)) {
+      this.email = email;
+      emails ??= <String>[];
```
sigurdm: Is this what happens for existing users?
I have not worked with these data classes before. Is there some way to ensure it is not null from a constructor? It seems a bit brittle that we always have to care about the null case.

Alternatively you could reorder the logic:
```dart
emails ??= <String>[];
if (!emails.contains(email)) {\
  ...
}
```
A tiny bit of overhead, but I think easier to read.

app/lib/account/backend.dart
```diff
@@ -0,0 +1,180 @@
...
+          info.verifiedEmail != true ||
+          info.email == null ||
+          info.email.isEmpty) {
+        _logger.warning('OAuth2 token info invalid: ${info.toJson()}');
+        return null;
+      }
+
+      return AuthResult(name, info.userId, info.email);
+    } on oauth2_v2.ApiRequestError catch (e) {
+      _logger.info('Access denied for OAuth2 access token.', e);
```
sigurdm: Should this logging include the `accessToken`?

app/lib/account/models.dart
isoos: It was me being too paranoid. Having checked the `ListProperty` sources, it always returns a non-empty String on decoding the database content.

app/lib/account/backend.dart
isoos: I've left out `_pool.close()` from the `AccountBackend.close()`, adding it.

It is part of the shutdown sequence (which we don't trigger right now), and at that point we shouldn't receive requests. The pool was already marked as closing, which means even if we schedule a new update, the scheduled command will not run.

However, to prevent accidental data inconsistency, we can let the code flow call `_pool.withResource`, because if the pool is already closed, it will throw an exception, making it more consistent: no access to the `User` object, no update to the `lastAccess` field.

app/lib/account/backend.dart
isoos: `historicalEmails`?

sigurdm: Fine with me. You can also just add a docstring explaining what it is.

app/lib/account/backend.dart
isoos: I'd rather leave the actual token info out of the logs as much as possible.

app/lib/account/models.dart
```diff
@@ -33,23 +33,13 @@ class User extends db.ExpandoModel {
   @db.DateTimeProperty()
   DateTime lastAccess;
 
-  bool updateEmail(String email) {
-    bool changed = false;
-
+  void updateEmail(String email) {
     if (this.email != email) {
       this.email = email;
```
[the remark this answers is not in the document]
isoos: it is updated in the caller block, and I think it may be better there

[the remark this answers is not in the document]
sigurdm: Ah - missed that.

app/lib/account/models.dart
```diff
@@ -0,0 +1,59 @@
...
+  bool hideEmail = false;
+
+  @db.DateTimeProperty()
+  DateTime created;
+
+  @db.DateTimeProperty()
+  DateTime updated;
+
+  @db.DateTimeProperty()
+  DateTime lastAccess;
```
jonasfj: Let's rename this to `lastLogin`, as access could imply the last time this record was accessed.

app/lib/account/models.dart
```diff
@@ -0,0 +1,59 @@
...
+  String authProvider;
+
+  @db.StringProperty()
+  String authUserId;
+
+  @db.StringProperty()
+  String email;
+
+  @db.StringListProperty()
+  List<String> historicalEmails = <String>[];
```
jonasfj: I suggest we drop this... What need to do we have? Also what are the privacy implications.

IMO, the only reason we store `email` is because:
 1. We want to be able to contact users
 2. We need a display name that has some trust

app/lib/account/models.dart
```diff
@@ -0,0 +1,59 @@
...
+
+import 'package:gcloud/db.dart' as db;
+
+/// User data model.
+@db.Kind(name: 'User', idType: db.IdType.String)
+class User extends db.ExpandoModel {
+  String get userId => id as String;
+
+  @db.StringProperty()
+  String authProvider;
```
jonasfj: If we want this, we need documentation.
I suggest that we don't do this... we should only have one auth provider... and keep it simple...

If we decide to add a different auth provider in the future, then we'll do a data migration and create an entity of kind `user_1` or something else :) hehe
(note the kind name is just in the annotation)

app/lib/account/backend.dart
```diff
@@ -0,0 +1,195 @@
...
+  final String userId;
+  final String email;
+
+  AuthResult(this.provider, this.userId, this.email);
+
+  String get compositeId => '$provider:$userId';
+}
+
+/// Authenticates access tokens.
+abstract class AuthProvider {
```
jonasfj: This seems like an unnecessary abstraction...

app/lib/account/backend.dart
```diff
@@ -0,0 +1,195 @@
...
+  /// Returns the `User` entry that is associated with the [accessToken].
+  ///
+  /// When no entry exists in Datastore, this method will create a new one.
+  ///
+  /// When the authenticated e-mail of the user changes, the email field will
+  /// be updated to the latest one, and the `historicalEmails` field contains
+  /// all of the earlier and current values.
+  ///
+  /// The method returns null if the access token is invalid.
+  Future<User> lookupOrCreateUser(String accessToken) async {
```
jonasfj: How about we create two methods:
 * `authenticateWithAccessToken(String accessToken)`
 * `authenticateWithIdToken(String idToken)`

Or even better we just a take `authenticateUser(request)`, then we can add all the auth methods we want... using cookies or different bearer tokens, etc...

app/lib/account/models.dart
```diff
@@ -0,0 +1,59 @@
...
+    if (!historicalEmails.contains(email)) {
+      historicalEmails.add(email);
+    }
+  }
+}
+
+/// Derived data for [User] for fast lookup.
+@db.Kind(name: 'UserInfo', idType: db.IdType.String)
+class UserInfo extends db.ExpandoModel {
+  String get userId => id as String;
```
jonasfj: I'm tempted to suggest a static method:
```dart
static db.Key key(String userId) => db.emptyKey.append(User, id: userId);
```

Then in the code we can just write:
```dart
User user = (await db.lookup<User>([User.key(myUserId)])).single;
```

jonasfj: Feel free to ignore this idea... it's purely syntactic nonsense, hehe :)

app/lib/account/models.dart
jonasfj: hmm, thinking...what if we change this to `lastLoginDate`, and then reduce precision to just the date... That way we only need to update it once per day per user...  Multiple requests the same on the same date won't cause updates...

This seems like the kind of thing that could cause write contention. And having the last date is probably fine for most users..

isoos: removed

app/lib/account/models.dart
isoos: removed

app/lib/account/models.dart
isoos: removed

app/lib/account/backend.dart
isoos: I think the abstraction could be still useful for local development and/or tests.

app/lib/account/models.dart
isoos: `User.key` is already taken by the `db.ExpandoModel`. `User.keyOf` could work like this:
````
  static db.Key keyOf(String userId) =>
      db.dbService.emptyKey.append(User, id: userId);

````
However, this will access `db.dbService`, without any contextual separation. I don't think it is a concern right now, just wanted to check with you.

File: app/lib/account/models.dart [hunk 15]
```diff
@@ -0,0 +1,59 @@
...
+  List<String> historicalEmails = <String>[];
+
+  @db.BoolProperty()
+  bool hideEmail = false;
+
+  @db.DateTimeProperty()
+  DateTime created;
+
+  @db.DateTimeProperty()
+  DateTime updated;
```
jonasfj: Do we even need this?

app/lib/account/backend.dart
```diff
@@ -0,0 +1,193 @@
...
+    final mappingKey = _db.emptyKey.append(User, id: auth.userId);
+
+    final user = await _retry.retry(() async {
+      // Check existing mapping.
+      final mapping = (await _db.lookup<OAuthUserID>([mappingKey])).single;
+      if (mapping != null) {
+        final user = (await _db.lookup<User>([mapping.userIdKey])).single;
+        // TODO: we should probably have some kind of consistency mitigation
+        if (user == null) {
+          throw StateError('Incomplete OAuth userId mapping.');
```
jonasfj: Throw an `AssertionError` instead.. and make the message highlight a datastore invariant violation.

app/lib/account/backend.dart
```diff
@@ -0,0 +1,193 @@
...
+
+      // Query the first use of the e-mail address.
+      DateTime created = DateTime.now().toUtc();
+      final uploadedVersions = _db.query<PackageVersion>()
+        ..filter('uploaderEmail =', auth.email);
+      await for (var version in uploadedVersions.run()) {
+        if (created.isAfter(version.created)) {
+          created = version.created;
+        }
+      }
```
jonasfj: I suggest we backfull the `User` entities in a one-off script...
This way, the `User` entity might exist with a `user.oauthUserId == null`, until the user logs in the first time... at which point we create the `OAuthUserID` entity AND update `user.oauthUserId = ...` in the same transaction...

This way, our one-off script can move us completely away from emails... We can replace the `packageVersion.uploaderEmail` with `packageVersion.uploader` and `package.uploaderEmails` with `package.uploaders`.
Then we can make the old fields as deprecated and as soon as we switch over traffic we stop using them.


Note: if we really smart, we design the one-off migration script such that it searches for a `User` entity with the given email before creating one. This way we can ensure that we can re-run the one-off script multiple times... Thus, we'll run it before and after we switch over traffic, how does that sound?

isoos: PTAL: I've updated the lookup to handle a pre-migrated User entity too. (I'd do the migration script in a separate PR)

PR #2004, opened 2019-02-01
Explicit pkg/pub_dartdoc initialization

isoos:


doc/development.md
```diff
@@ -18,6 +22,12 @@ pub global activate mono_repo
 - `mono_repo.yaml` (in the root directory)
 - `mono_pkg.yaml` (in each package directory)
 
+### Update/upgrade dependencies
+
+````bash
+pub global run mono_repo --recursive pub get
```
sigurdm: Should this be ?
```
pub global run mono_repo --recursive pub upgrade
```

isoos: We could display both: `get` for a freshly cloned repository, `upgrade` for a dependency upgrade.

The reason I'd rather default to `get` is `pkg/pub_dartdoc`: it is on a separate dependency tree than the rest of the packages, and while there is very little risk of screwing it up with a upgrade, I'd rather not suggest it in the doc.

sigurdm: No problem then.

PR #2007, opened 2019-02-04
Upgrade pana and Flutter SDK

isoos:


isoos:
(CI will fail until pana is published)

PR #2010, opened 2019-02-04
Additional CSP rules.

### LINKED ISSUE #1986: Forbid embedding of images over HTTP
This is only really possible using markdown.

I propose that we just remove these images.. such that embedding images over HTTP results in a broken image icon (or something like that).

note: embedding images over HTTP is a mixed-content violation and in the future this will likely break as a result CSP rules.

isoos:
- Fixes #1986 (images only over https)
- Addresses most of #1691 (`block-all-mixed-content` is pending)

### DIFF
--- a/app/lib/shared/handlers.dart
+++ b/app/lib/shared/handlers.dart
@@ -54,8 +54,27 @@ shelf.Response jsonResponse(
   );
 }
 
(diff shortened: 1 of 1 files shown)


isoos:
This is currently deployed in staging.

PR #2015, opened 2019-02-05
Return only match ScoreCardData that has pana and dartdoc reports.

### LINKED ISSUE #2011: Weirdly blank analysis data for json_serializable
URL: https://pub.dartlang.org/packages/json_serializable#-analysis-tab-

screen shot 2019-02-04 at 12 40 50 pm

CC @jonasfj @isoos

isoos:
Fixes #2011.

I'll have a follow-up/cleanup PR after this as some of the ScoreCard-handling code could use simplification.

### DIFF
--- a/app/lib/scorecard/backend.dart
+++ b/app/lib/scorecard/backend.dart
@@ -46,9 +46,11 @@ class ScoreCardBackend {
     String packageName,
     String packageVersion, {
     @required bool onlyCurrent,
-    List<String> requiredReportTypes,
   }) async {
-    requiredReportTypes ??= const <String>[ReportType.pana, ReportType.dartdoc];
+    final requiredReportTypes = const <String>[
+      ReportType.pana,
+      ReportType.dartdoc,
+    ];
     if (packageVersion == null || packageVersion == 'latest') {
       final key = _db.emptyKey.append(Package, id: packageName);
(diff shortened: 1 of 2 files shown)


app/lib/scorecard/backend.dart
```diff
@@ -59,16 +61,18 @@ class ScoreCardBackend {
     final cached = await scoreCardMemcache.getScoreCardData(
         packageName, packageVersion, versions.runtimeVersion,
         onlyCurrent: onlyCurrent);
-    if (cached != null) {
+    if (cached != null && cached.hasReports(requiredReportTypes)) {
       return cached;
```
jonasfj: Let's just ignore stuff from the cache, if both reports are not set..

app/lib/scorecard/backend.dart
```diff
@@ -46,7 +46,9 @@ class ScoreCardBackend {
     String packageName,
     String packageVersion, {
     @required bool onlyCurrent,
+    List<String> requiredReportTypes,
```
jonasfj: Do we ever call this with anything ,but `null`?

How about we already reload unless both types are present? It's simpler and we have fewer options to deal with...
The overhead will likely be small?

jonasfj: If adding new options, these should be documented... :)

PR #2017, opened 2019-02-05
Limit readme, changelog and example length.

isoos:
This is not an exact byte-length, but something to start with. I think we'll need to have a pana suggestion and some kind of penalty for large files.

### DIFF
--- a/app/lib/shared/utils.dart
+++ b/app/lib/shared/utils.dart
@@ -68,7 +68,8 @@ Future<List<String>> listTarball(String path) async {
       .toList();
 }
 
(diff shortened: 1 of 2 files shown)


app/lib/shared/utils.dart
```diff
@@ -68,7 +68,8 @@ Future<List<String>> listTarball(String path) async {
       .toList();
 }
 
-Future<String> readTarballFile(String path, String name) async {
+Future<String> readTarballFile(String path, String name,
```
jonasfj: please add a documentation comment...

PR #2022, opened 2019-02-07
Formatted template golden files.

isoos:
- Fixed small template issues that prevented XML-parsing of our HTML output.
- Moved golden files to pretty-printed XML files, that way we can focus on readable templates, while still check the changes against normalized and indented output.


app/test/frontend/templates_test.dart
```diff
@@ -63,12 +64,16 @@ void main() {
         validateHtml(root);
       }
 
+      final xmlDoc = xml
+          .parse(isFragment ? '<fragment>' + content + '</fragment>' : content);
+      final xmlContent = xmlDoc.toXmlString(pretty: true, indent: '  ');
```
sigurdm: Could you insert a trailing newline at the end?

app/test/frontend/templates_test.dart
```diff
@@ -63,12 +64,16 @@ void main() {
         validateHtml(root);
       }
 
+      final xmlDoc = xml
```
jonasfj: please add a comment like: `// pretty printing output using XML parser`

PR #2025, opened 2019-02-08
Tool: backfill user ids.

isoos:


app/lib/frontend/models.dart
```diff
@@ -50,6 +50,9 @@ class Package extends db.ExpandoModel {
   @db.ModelKeyProperty(propertyName: 'latest_dev_version')
   db.Key latestDevVersionKey;
 
+  @CompatibleStringListProperty()
```
jonasfj: Do we have to use`CompatibleStringListProperty` for new properties?

jonasfj: Can we give it a comment `/// List of [User.userId]`

Can we give `uploaderEmails` a comment, that that is deprecated, and should not be modified again?

app/bin/tools/backfill_user_ids.dart
```diff
@@ -0,0 +1,138 @@
...
+    print(
+        '${emailsWithMultipleUuid.length} e-mail address have multiple Users.');
+    print('${emailToUuuid.length} e-mail -> User mapping.');
+
+    counter = 0;
+    print('Scanning Packages...');
+    final pkgQuery = dbService.query<Package>();
+    await for (Package p in pkgQuery.run()) {
+      if (p.uploaders.isEmpty) {
+        packagesToMigrate.add(p.key);
```
jonasfj: Let's always migrate all packages... see details below...

app/bin/tools/backfill_user_ids.dart
```diff
@@ -0,0 +1,138 @@
...
+      await dbService.commit(inserts: [user]);
+
+      emailToUuuid[email] = id;
+
+      incCounter(() => email);
+    }
+
+    counter = 0;
+    print('Migrating Packages... [${packagesToMigrate.length}]');
+    for (Key key in packagesToMigrate) {
```
jonasfj: Let's always migrated all packages...

app/bin/tools/backfill_user_ids.dart
```diff
@@ -0,0 +1,138 @@
...
+      await dbService.withTransaction((tx) async {
+        final p = (await tx.lookup<Package>([key])).single;
+        p.uploaders =
+            p.uploaderEmails.map((email) => emailToUuuid[email]).toList();
```
jonasfj: Let's do a set union:
```dart
final uploaders = Set();
if (p.uploaders != null) {
  uploaders.addAll(p.uploaders);
}
uploaders.allAll(p.uploaderEmails.map((email) => emailToUuuid[email]));
p.uploaders = uploaders.toList();
```

There might be a more elegant way to write this...

But this ensure that we can always run this script... Hence, we can run the script before/after we've switched over traffic...

app/bin/tools/backfill_user_ids.dart
```diff
@@ -0,0 +1,138 @@
...
+import 'package:gcloud/db.dart';
+import 'package:uuid/uuid.dart';
+
+import 'package:pub_dartlang_org/account/models.dart';
+import 'package:pub_dartlang_org/frontend/models.dart';
+import 'package:pub_dartlang_org/frontend/service_utils.dart';
+
+final _uuid = Uuid();
+
+Future main(List<String> args) async {
```
jonasfj: Can we give the script a `--help` argument... and can we write a nice multi-line string that explain what it does.
Something like:
> Usage: dart backfull_user_ids.dart
Ensures a `User` entity exists for each `email` used in `PackageVersion.uploaderEmail` and `Package.uploaderEmails`.
Ensures that `Package.uploaders` is a subset of `userId`s for emails in `Package.uploaderEmails`.
Ensures that `PackageVersion.uploader` is the `userId` of `PackageVersion.uploaderEmail`.

app/lib/frontend/models.dart
isoos: > Do we have to use`CompatibleStringListProperty` for new properties?

Switching to `StringListProperty`.

> Can we give `uploaderEmails` a comment, that that is deprecated, and should not be modified again?

I'll do a follow-up PR that changes both properties in-sync (after this and the PR with the AuthenticatedUser gets merged). After that is deployed, we can remove it altogether.

isoos:
PTAL.

isoos:
I've removed the created-time scan from the `authenticate...` method, because it introduces a lag that is not really required there (e.g. another run of the backfill script should fix it anyway).

app/bin/tools/backfill_user_ids.dart
```diff
@@ -0,0 +1,129 @@
...
+    } else if (pv.uploader != user.userId) {
+      throw AssertionError(
+          'upload missmatch in ${pv.package} ${pv.version}: ${pv.uploaderEmail} maps to ${user.userId} but already contains ${pv.uploader}');
+    }
+  }
+
+  await dbService.withTransaction((tx) async {
+    final p = (await dbService.lookup<Package>([pkgKey])).single;
+    final uploaders = Set<String>();
+    for (String email in p.uploaders) {
```
jonasfj: `p.uploaderEmails`, right?

isoos: /facepalm

PR #2026, opened 2019-02-08
Use full authenticated user info in place of e-mail address.

isoos:
Needed to restructure tests a bit, as previously they were updating/operating a single testPackage entry. (They still do use that, but mutable operations get their own instance.)

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -9,6 +9,7 @@ import 'package:gcloud/service_scope.dart' as ss;
 import 'package:googleapis/oauth2/v2.dart' as oauth2_v2;
 import 'package:http/http.dart' as http;
 import 'package:logging/logging.dart';
+import 'package:pub_server/repository.dart' show UnauthorizedAccessException;
 import 'package:retry/retry.dart';
 import 'package:uuid/uuid.dart';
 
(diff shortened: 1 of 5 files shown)


app/lib/account/backend.dart
```diff
@@ -153,19 +174,26 @@ class AccountBackend {
...
+class AuthenticatedUser {
   final String userId;
   final String email;
 
-  AuthResult(this.userId, this.email);
+  AuthenticatedUser(this.userId, this.email);
+}
+
+class AuthResult {
+  final String oauthUserId;
```
jonasfj: I love the consistent naming  of `oauthUserId` :)

app/lib/account/backend.dart
```diff
@@ -31,6 +32,25 @@ void registerAccountBackend(AccountBackend backend) =>
...
+
+/// Calls [fn] with the currently authenticated user as an argument.
+///
+/// If no user is currently authenticated, this will throw an
+/// `UnauthorizedAccess` exception.
+Future<R> withAuthenticatedUser<R>(Future<R> fn(AuthenticatedUser user)) async {
+  if (authenticatedUser == null) {
+    throw UnauthorizedAccessException('No active user.');
+  }
+  return await fn(authenticatedUser);
```
jonasfj: This only only works because `appengine` creates a _service scope_ per request... We are **not** going to clean this up here.. but I would really love to see this being less fragile. This basically and other reason why I want to get away from the `pub_server` abstractions.

isoos: Yeah, and the tests doesn't help much either, because there is no guaranteed separation between the tests for accessing these fields. We have many reasons to refactor that part of the code :)

PR #2030, opened 2019-02-11
Remove tools: copy_data, datastore.

isoos:
As the backup-restore cycle to update staging seems to be easy, there is no need for tools like the two scripts (which will get outdated soon as we do more model changes).


sigurdm:
Does this give some datastore tables we can remove?

sigurdm:
I meant cloud storage buckets

isoos:
No, these scripts were working with local files.

sigurdm:
Thanks!

PR #2034, opened 2019-02-11
Bug fixes for userId migration.

isoos:
- the type param for `mappingKey` was wrong
- the migration script used wrong key type to access first created timestamp (strange, should have been caught by type checking)
- Changing AssertionError to Exception, because otherwise we won't have a reasonable message in the stacktrace.

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -69,12 +69,12 @@ class AccountBackend {
   /// Returns the `User` entry for the [email] or creates a new one if it does
   /// not exists.
   ///
-  /// Throws AssertionError if more then one `User` entry exists.
+  /// Throws Exception if more then one `User` entry exists.
   Future<User> lookupOrCreateUserByEmail(String email) async {
     final query = _db.query<User>()..filter('email =', email);
     final list = await query.run().toList();
     if (list.length > 1) {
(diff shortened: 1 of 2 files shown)


app/lib/account/backend.dart
```diff
@@ -69,12 +69,12 @@ class AccountBackend {
...
   /// not exists.
   ///
-  /// Throws AssertionError if more then one `User` entry exists.
+  /// Throws Exception if more then one `User` entry exists.
   Future<User> lookupOrCreateUserByEmail(String email) async {
     final query = _db.query<User>()..filter('email =', email);
     final list = await query.run().toList();
     if (list.length > 1) {
-      throw AssertionError('More than one User exists for e-mail: $email');
+      throw Exception('More than one User exists for e-mail: $email');
```
jonasfj: `Exception` can be caught... I asked on gitter, if we're just not supposed to be using `AssertionError`, I'm curious why it has a public constructor if we're not supposed to be using it :)

isoos: My issue is that it ignores the `message` in its `toString()` method:
```
  final Object message;
  AssertionError([this.message]);
  String toString() => "Assertion failed";
```

PR #2038, opened 2019-02-12
Expose dartdoc result in failure message.

isoos:


app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -254,7 +256,7 @@ class DartdocJobProcessor extends JobProcessor {
         score: 10.0,
       ));
     } else {
-      maintenanceSuggestions.add(getDartdocRunFailedSuggestion());
+      maintenanceSuggestions.add(getDartdocRunFailedSuggestion(dartdocResult));
```
jonasfj: Please, limit the error message to say 8kb? (or something reasonable)..
Maybe we take the first 4kb insert "...[omitted for brevity]..." followed by the next 4kb. If there is more than 8kb...

Naturally, we should add this logic in:
https://github.com/dart-lang/pana/blob/2bd013c67c1fbeaae3825d66f4cc185742a50755/lib/src/dartdoc_analyzer.dart#L23

I know the `googleapis` package have very very long log output, mostly just warnings regarding generated code.

isoos: Good point. I've updated the code to use _mergeOutput instead, which caps the output stream to 40-40 lines. The total size should be under 10k. PTAL.

PR #2044, opened 2019-02-13
Remove text placeholders from dartdoc customization.

isoos:
Since we no longer care about these, why keep them...


isoos:
> Because you're normalizing with `xml` now?

Yes, at least normalizing with `xml` in the tests. The .html file won't that nice anymore, but maintaining the whitespaces was never really fun...

PR #2055, opened 2019-02-15
Fix search update: query the right field.

isoos:
https://github.com/dart-lang/pub-dartlang-dart/issues/2053


isoos:
This probably goes back to the `Analysis`->`ScoreCard` switch: `Analysis` had a `timestamp` field, while `ScoreCard` has only `updated`.

PR #2056, opened 2019-02-15
Simplify download to file code.

isoos:
Confirmed that backend tests work as expected.


isoos:
> Wow, this is much nicer... but don't we have to do something to close the incoming stream? it'll probably be GC'ed eventually...
> Or does this actually work?

Oh, actually, you are right, `addStream` does not close the File writes, I've been using it in the other way around in a different project: the stream needs to be piped to the sink to get automatic closing (and with that the file content is also flushed to the disk, while my first version won't do it automatically).

Updating, PTAL.

PR #2059, opened 2019-02-15
cronjob for backup of datastore

jonasfj:
I'm thinking we can deploy this to staging before we configure
a bucket to be used for backup. If we roll to staging sometime
Monday, we can probably search the logs tuesday/wednesday for
a failed cronjob, as it's supposed to shout if there is
no bucket configured.

Trying this in tests makes little sense, as the thing most likely to fail here is the appengine integration.

### DIFF
--- a/app/lib/frontend/cronjobs.dart
+++ b/app/lib/frontend/cronjobs.dart
@@ -1,0 +1,86 @@
+import 'package:shelf/shelf.dart' show Request, Response;
+import 'package:logging/logging.dart' show Logger;
+import 'package:googleapis_auth/auth_io.dart' as auth;
+import 'package:googleapis/datastore/v1.dart' as datastore;
+import 'package:ulid/ulid.dart' show Ulid;
+
+final _log = new Logger('pub.cronjobs');
+
+class CronJobs {
+  final String backupBucketName;
+
+  CronJobs({
+    this.backupBucketName,
+  });
+
+  Future<Response> handler(Request request) async {
+    final job = _resolve(request.url.path);
+    if (job == null) {
+      return Response.notFound('no-such-cron-job');
+    }
+    _log.finest('### Cron: ${request.url.path}');
+    final ok = await job();
+    if (!ok) {
+      _log.shout('### Cron failed: ${request.url.path}');
+      return Response.internalServerError(
+        body: 'cron failed ${request.url.path}',
+      );
+    }
+    _log.finest('### Cron successful: ${request.url.path}');
+    return Response.ok('cron successful: ${request.url.path}');
+  }
+
+  /// Map from path to cronjob.
+  Future<bool> Function() _resolve(String path) {
+    return <String, Future<bool> Function()>{
+      '/cron/datastore-backup': datastoreBackup,
+    }[path];
+  }
+
+  /// Backup datastore
+  Future<bool> datastoreBackup() async {
+    if (backupBucketName == null) {
+      _log.shout('datastore backup not configured!!!');
+      return false;
+    }
+
+    // Get credentials via. metadata service (this only works on GCP/AppEngine)
+    final client = await auth.clientViaMetadataServer();
+    try {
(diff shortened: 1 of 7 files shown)


app/lib/shared/handler_helpers.dart
```diff
@@ -19,22 +19,28 @@ import 'handlers.dart';
...
   if (sanitize) {
     handler = _sanitizeRequestWrapper(handler);
   }
   handler = _redirectToHttpsWrapper(handler);
   handler = _logRequestWrapper(logger, handler);
-  return runAppEngine((HttpRequest request) => _handleRequest(request, handler),
-      shared: true);
+  await runAppEngine((HttpRequest request) {
+    if (cronJobHandler != null && isCronJobRequest(request)) {
+      shelf_io.handleRequest(request, cronJobHandler);
```
isoos: Why not put this inside the `appHandler` in `lib/frontend/handlers.dart`? I know that technically this is not an app-related handler, but I would probably look there first to see where this is going to be routed.

app/lib/frontend/cronjobs.dart
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    // Get credentials via. metadata service (this only works on GCP/AppEngine)
+    final client = await auth.clientViaMetadataServer();
+    try {
+      // Create a unique job id, in case multiple jobs are trigger concurrently
+      final id = Ulid().toString();
+
+      // Get date as: YYYY-MM-DD
+      final now = DateTime.now().toUtc();
```
isoos: DateTime.now().toUtc().toIso8601String().split('T').first

app/lib/frontend/cronjobs.dart
```diff
@@ -0,0 +1,86 @@
...
+    if (backupBucketName == null) {
+      _log.shout('datastore backup not configured!!!');
+      return false;
+    }
+
+    // Get credentials via. metadata service (this only works on GCP/AppEngine)
+    final client = await auth.clientViaMetadataServer();
+    try {
+      // Create a unique job id, in case multiple jobs are trigger concurrently
+      final id = Ulid().toString();
```
isoos: I like the use of ulid :), but we can get away using the uuid which we use elsewhere: as the file is prefixed by the date, and there will be a timestamp on the backup file too.

jonasfj: I don't trust the uuid package it has phrases like:
>  * Cryptographically strong random number generation on all platforms
>    * Defaults to non-crypto generator, see UuidUtil for cryptoRNG

This might be different in 2.0.0, but we haven't upgraded to that yet, and the `README` for the uuid package still looks sketchy... :)

app/lib/shared/handler_helpers.dart
jonasfj: The `isCronJobRequest` operates on `HttpRequest` and not `shelf.Request`...

We could just reimplement it using `shelf.Request`... I'll agree this all feels a little stitched together..

app/lib/shared/handler_helpers.dart
```diff
@@ -19,21 +19,26 @@ import 'handlers.dart';
...
   }
   handler = _redirectToHttpsWrapper(handler);
   handler = _logRequestWrapper(logger, handler);
-  return runAppEngine((HttpRequest request) => _handleRequest(request, handler),
-      shared: true);
-}
-
-Future _handleRequest(HttpRequest request, shelf.Handler handler) =>
+  await runAppEngine((HttpRequest request) {
+    if (request.headers['x-appengine-cron'].contains('true') &&
```
isoos: Wouldn't it be better to have a strict equals check?

jonasfj: This is a list of headers... a header may have more than one value...

isoos: Ah, yeah.

Issue #2060, opened 2019-02-16
Awaiting analysis to complete taking a long time


felangel:
I've submitted updates to several packages over an hour ago and they are all still "Awaiting analysis to complete"

In addition, it's impossible to search for them. Is there some sort of maintenance which is causing all of this? Normally, everything is updated almost immediately after a submission.

See the following for an example:
https://pub.dartlang.org/packages/bloc
https://pub.dartlang.org/packages/flutter_bloc

isoos:
The log entries show that both packages were picked up and the analysis was completed, there were no indication of an error while we stored the analysis results. It looks like we have an issue with the caching, and the new results didn't show up on the page yet. In worst case, you should see the results in ~24 hours.

I'll address the caching issue next week, sorry for the inconvenience.

felangel:
Thanks for the quick reply! I was planning on pushing additional updates later today, do you think that would invalidate the cache?

isoos:
@felangel: it looks like your new version of `flutter_bloc` did get the cache updated for that version, but the `bloc` did not (yet). The search is connected with the analysis, and it is failing for the same reason (and another search-related one, see #2053 and #2057).

In general we don't recommend to publish a new version in the hope that it will "fix the site". It is very likely that you'll hit the same issue, and the new version won't fix everything by itself. A missing analysis is not a show-stopper for neither publishing, nor downloading packages, and we will do our best to fix it on regular work hours. Once the related code is fixed, the package will get re-analyzed and you'll see the scores.

felangel:
@isoos just another update:
Now all packages have a score but when I search for bloc (https://pub.dartlang.org/packages/bloc) it’s nowhere to be found.

When I search for angular bloc, bloc shows up 🤔

isoos:
@felangel: yes, search index missing the new analysis is known, and fix is done for that, but not deployed to production yet.

felangel:
@isoos looks like everything is back to normal 🎉. Would you like me to close this issue?

isoos:
@felangel let's keep this open until we fix the underlying causes, thanks!

fadhly-permata:
> 
> 
> @felangel: yes, search index missing the new analysis is known, and fix is done for that, but not deployed to production yet.

is this problem the same as what happened at #2068 ?

isoos:
Yes, it is the same root cause. The fix is already done, but it hasn't been deployed to prod yet.

fadhly-permata:
[...]

Issue #2057, opened 2019-02-15
No plugins are detected.


JosephNK:
URL: https://pub.dartlang.org/packages?q=flutter_kakao_login => found
URL: https://pub.dartlang.org/flutter/packages?q=flutter_kakao_login => not found

The flutter_kakao_login plug-in will only search if the filter condition is set to ALL at https://pub.dartlang.org/flutter.
If the filter condition is FLUTTER, it is not searched.
Please let me know why.

isoos:
This has the same root cause as #2053, which will be addressed via #2055.

There are two additional info with the symptoms:
- the search index loaded only the version `0.0.6`, and not the `0.0.7`
- the loaded scorecard apparently did not have the `uses-flutter` flag.

isoos:
I'm starting to understand the root cause of not having the `uses-flutter` flag in the search index: at the time when the index tries to load the package document for the freshly uploaded version, there was no analysis yet, neither with that runtimeVersion, nor with any previous one.

There was no fallback to an earlier version, and a placeholder null-analysis was used, which didn't have the platform information. It was made worse by the underlying bug of #2055, which didn't trigger an update even after the analysis was done.

Instead of the placeholder null-analysis we'll need to just keep using the existing entry in the index, which should be the latest correct analysis result for a past version that was latest stable at that time.

JosephNK:
@isoos 

URL: https://pub.dartlang.org/packages?q=flutter_kakao_login => found
URL: https://pub.dartlang.org/flutter/packages?q=flutter_kakao_login => found

It works now.
Did your search issue resolve?
If it is resolved, you may close this issue.

isoos:
@JosephNK it was expected that it will be resolved automatically (after ~24-26 hours at most), but I'd keep this issue open because the underlying cause is not addressed yet.

isoos:
The fix was deployed to prod today, closing the issue.

Issue #2053, opened 2019-02-15
camera plugin not listed


zoechi:
![image](https://user-images.githubusercontent.com/405837/52842820-93be6400-3100-11e9-98fc-8cb60d9ba9af.png)

Might just be an intermittent issue. I thought I report it anyway. 
Feel free to close if you think it only needs waiting for some job to run or similar.

isoos:
Poking the search service, I was able to force the following exception, looking into it:

```
Error:
    NoSuchMethodError: The getter 'pkgScore' was called on null.
    Receiver: null
    Tried calling: pkgScore

Stack:
    dart:core/runtime/libobject_patch.dart 50:5                   Object.noSuchMethod
    package:pub_dartlang_org/search/index_simple.dart 258:47      SimplePackageIndex.search
```

isoos:
The above exception is unrelated, it happened because of a URL query combination that is unreachable from the UI, I've queried the service backend directly.

The `camera` package is listed on page 7, which is clearly not good: https://pub.dartlang.org/packages?q=camera&page=7

isoos:
Looks like the package score is off in the index, because we have 3 packages that were updated in the past 24 hours and they are in the back of the list when ordered by overall score:
https://pub.dartlang.org/packages?q=camera&sort=top&page=8

I'll investigate why Datastore scanning part of the index did not pick up the package scores (e.g. `firebase_ml_vision`, `camera`, `google_maps_flutter`). The daily index update should pick it up, so the issue should be gone by tomorrow at latest.

isoos:
The fix was deployed to prod today, closing the issue.

PR #2061, opened 2019-02-18
Update search index only if there is an analysis available.

isoos:
This addresses the search index part of #2053, #2057 and #2060. With this change, the index updater checks if there is a valid analysis for the loaded version, and if there is nothing yet, it will not load the document into the index. As the analysis completes, the index should pick up the updated scorecard, and will trigger a new update again.

I'm not a fan of using exception for such control flow, but it just seemed the best choice.

### DIFF
--- a/app/lib/frontend/handlers/package.dart
+++ b/app/lib/frontend/handlers/package.dart
@@ -148,10 +148,8 @@ Future<shelf.Response> _packageVersionHandlerHtml(
       }
     }
     final Stopwatch serviceSw = new Stopwatch()..start();
-    final analysisKey =
-        new AnalysisKey(selectedVersion.package, selectedVersion.version);
-    final AnalysisView analysisView =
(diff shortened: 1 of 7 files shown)


jonasfj:
I don't understand why shouldn't a package be indexed before the analysis is completed?
It should be reindexed when the analysis is done, but why not have it in the index while analysis is running?

isoos:
> I don't understand why shouldn't a package be indexed before the analysis is completed? It should be reindexed when the analysis is done, but why not have it in the index while analysis is running?

At that point we don't have e.g. platform information, health and maintenance scores. As a consequence, a search will yield the package in "all dart packages", but won't give it in "Flutter packages" (assuming it is a Flutter-packages. A popular package which would be on the first page usually would experience to be back at the 10th page until the analysis of it completes.

isoos:
PTAL.

PR #2065, opened 2019-02-18
Update uploaderEmail: no more required, using only id to check uploaders.

isoos:


isoos:
This is getting a bit tricky: I've realized that we should lowercase the e-mail address when looking up userId for a given e-mail, however we already have a bunch of e-mails migrated in prod. We'll need to validate whether all of the current e-mail addresses are indeed lowercase.

isoos:
Closing this in favor of #2072.

PR #2069, opened 2019-02-19
Temp tool: checking user e-mails are lowercase.

isoos:
Ideally we should remove this in 1-2 weeks.


app/bin/tools/check_user_emails.dart
```diff
@@ -0,0 +1,38 @@
...
+
+import 'dart:async';
+
+import 'package:gcloud/db.dart';
+
+import 'package:pub_dartlang_org/account/backend.dart';
+import 'package:pub_dartlang_org/account/models.dart';
+import 'package:pub_dartlang_org/frontend/service_utils.dart';
+
+Future main(List<String> args) async {
```
jonasfj: At least give a comment....

PR #2072, opened 2019-02-19
Fix lowercase e-mails in user accounts.

isoos:
- `frontend/backend.dart#975` was removed in #2052, not deployed to production, and I felt it is safer to keep it there for the time being
- migration script was updated to also update email to lowercase if needed.
- migration script was updated to remove `User` entities if the email is not lowercased (the appropriate ids will get updated in prior steps, and package and version will have the correct IDs).

### DIFF
--- a/app/bin/tools/backfill_user_ids.dart
+++ b/app/bin/tools/backfill_user_ids.dart
@@ -63,6 +63,18 @@ Future main(List<String> args) async {
         await _updateUserCreatedTime(user, created);
       }
     }
+
+    await for (User user in dbService.query<User>().run()) {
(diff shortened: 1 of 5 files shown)


app/bin/tools/backfill_user_ids.dart
```diff
@@ -63,6 +63,18 @@ Future main(List<String> args) async {
         await _updateUserCreatedTime(user, created);
       }
     }
+
+    await for (User user in dbService.query<User>().run()) {
```
jonasfj: A single line comment saying, delete users with non-lowercases emails..

PR #2078, opened 2019-02-20
Delete more of old ScoreCards

isoos:
- The current GC threshold is `2018.12.05` (runtimeVersions before this time will be deleted).
- Each analysis should be re-done in ~30 days, it is very unlikely that a report for the current runtime did succeeded at one point, but failed ever since in the past half year, and if it does, we should probably delete that report anyway.

### DIFF
--- a/app/lib/scorecard/backend.dart
+++ b/app/lib/scorecard/backend.dart
@@ -220,43 +220,31 @@ class ScoreCardBackend {
     final deletes = <db.Key>[];
     final now = new DateTime.now();
 
-    // deleting reports
-    final reportQuery = _db.query<ScoreCardReport>()
-      ..filter('runtimeVersion <', versions.gcBeforeRuntimeVersion);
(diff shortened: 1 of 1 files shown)


app/lib/scorecard/backend.dart
```diff
@@ -220,43 +220,31 @@ class ScoreCardBackend {
...
     final now = new DateTime.now();
 
-    // deleting reports
-    final reportQuery = _db.query<ScoreCardReport>()
-      ..filter('runtimeVersion <', versions.gcBeforeRuntimeVersion);
-    await for (ScoreCardReport report in reportQuery.run()) {
-      final age = now.difference(report.updated);
-      if (age <= _deleteThreshold) {
-        continue;
+    Future deleteQuery<T extends db.Model>(db.Query<T> query) async {
```
jonasfj: Maybe give this a short  comment...

PR #2084, opened 2019-02-22
Cache /api/documentation/<package> requests.

isoos:


app/lib/shared/dartdoc_memcache.dart
```diff
@@ -59,14 +66,29 @@ class DartdocMemcache {
...
       _entry.invalidate(_entryKey(package, 'latest')),
+      _apiSummary.invalidate(_apiSummaryKey(package)),
     ]);
   }
 
   String _entryKey(String package, String version) => '/$package/$version';
 
   String _fileInfoKey(String objectName) => objectName;
+
+  String _apiSummaryKey(String package) => package;
```
jonasfj: You don't need this method... it just serves as unnecessary indirection :) hehe

PR #2085, opened 2019-02-22
Convert UnauthorizedAccessException to HTTP 401

isoos:


app/lib/shared/handler_helpers.dart
```diff
@@ -104,10 +105,18 @@ shelf.Handler _sanitizeRequestWrapper(shelf.Handler handler) {
...
   return (shelf.Request request) async {
     await registerLoggedInUserIfPossible(request);
-    return await handler(request);
+    try {
+      return await handler(request);
+    } on UnauthorizedAccessException catch (e) {
+      return jsonResponse(
+        {'message': e.toString()},
+        status: 401,
+        pretty: isPrettyJson(request),
```
jonasfj: If this includes the `access_token` we should not be sending it back over HTTPS, it might end up in logs, and other inappropriate places...

I suspect it's better to just say try `pub logout` to re-initialize your login session..

isoos:
PTAL

Issue #1595, opened 2018-09-10
Use more fake_gcloud and end2end tests


isoos:
- [ ] ToolEnvironment initialization
- [ ] `pub upgrade` calls
- [ ] end2end pana runner
- [ ] end2end dartdoc runner

PR #2090, opened 2019-02-26
pkg/fake_gcloud

isoos:
- a fake DatastoreDB, which stores the objects in memory, and it clones them for both storing and lookup,
- a fake Storage, which stores the file content in memory

#1595


pkg/fake_gcloud/lib/mem_storage.dart
```diff
@@ -0,0 +1,228 @@
...
+    if (info == null) {
+      throw DetailedApiRequestError(404, '$name does not exists');
+    }
+    return info;
+  }
+
+  @override
+  Stream<List<int>> read(String objectName, {int offset, int length}) async* {
+    if (!objectName.startsWith('/')) {
+      objectName = '/$objectName';
```
jonasfj: Does the original API do this... if so leave a comment saying it, otherwise it's probably dangerous to fix the objectName, and better to throw an `ArgumentError`, that way we're forced to fix the code...

pkg/fake_gcloud/lib/src/cloner.dart
```diff
@@ -0,0 +1,42 @@
+import 'dart:mirrors';
+
+/// Configures a clone helper.
```
jonasfj: Needs documentation... If we need this at all, see comment on how we might not need to work at the high-level... then we don't have use mirrors and we can just rely on the mirrors already used in `gcloud`.

pkg/fake_gcloud/lib/mem_db.dart
```diff
@@ -0,0 +1,187 @@
...
+import 'package:gcloud/datastore.dart' as ds;
+import 'package:gcloud/db.dart';
+
+import 'src/cloner.dart';
+
+/// Implementation of [DatastoreDB] interface with in-memory storage.
+///
+/// Entities are cloned, and should be safe to modify them once they are stored
+/// or read.
+class MemDatastoreDB implements DatastoreDB {
```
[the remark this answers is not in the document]
isoos: PTAL: I've implemented `Datastore`.

pkg/fake_gcloud/lib/mem_storage.dart
isoos: PTAL: I've changed to required the '/' prefix. I couldn't find if the API returns or requires it, for now it will be fine to pre-validate it.

pkg/fake_gcloud/lib/mem_datastore.dart
```diff
@@ -0,0 +1,135 @@
...
+      {List<Entity> inserts,
+      List<Entity> autoIdInserts,
+      List<Key> deletes,
+      Transaction transaction}) async {
+    deletes?.forEach((key) => _entities.remove(key));
+    inserts?.forEach((e) {
+      _entities[e.key] = e;
+    });
+    if (autoIdInserts != null && autoIdInserts.isNotEmpty) {
+      throw UnimplementedError();
```
jonasfj: Give it a brief message saying what you haven't implemented :) it helps a lot when we end up debugging this error later :)

pkg/fake_gcloud/lib/mem_storage.dart
```diff
@@ -0,0 +1,223 @@
...
+    for (String name in _files.keys) {
+      if (prefix == null || name.startsWith(prefix)) {
+        yield _BucketEntry(name, true);
+      }
+    }
+  }
+
+  @override
+  Future<Page<BucketEntry>> page({String prefix, int pageSize = 50}) async {
+    throw UnimplementedError();
```
jonasfj: Same for all these.. Just something like: `throw UnimplementedError('fake_gcloud.Bucket.page is not implemented ');`

PR #2114, opened 2019-03-06
Reject uploads on case insensitive filename matches.

### LINKED ISSUE #2104: reject packages with filenames that are case-insensitive equal
Such packages cannot be extracted on windows. as it has a case-insensitive (case-preserving) file system.

Example: https://github.com/dart-lang/pub/issues/2072

We might also have to scan all existing packages for this... I'm not sure we can remove them... that might be something we can editorial decision on... I suspect all cases of this is mistakes :)

isoos:
- Fixes #2104
- Removed directory depth limit from `listTarball`, as we want to check all files.
- I wanted to limit the number of libraries in the same PR, but was undecided on the limit to use (e.g. how should https://pub.dartlang.org/packages/googleapis#-installing-tab- be limited)?

### DIFF
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -860,6 +860,19 @@ Future<_ValidatedUpload> _parseAndValidateUpload(
 
   final files = await listTarball(filename);
 
+  // Check whether the files can be extracted on case-preserving file systems
+  // (e.g. on Windows). We can't allow two files with the same case-insensitive
+  // name.
+  final lowerCaseFiles = <String>{};
+  for (String file in files) {
+    final lower = file.toLowerCase();
+    if (lowerCaseFiles.contains(lower)) {
+      throw GenericProcessingException(
+          'Filename collision on case-preserving file systems: $file.');
+    }
+    lowerCaseFiles.add(lower);
+  }
+
   // Searches in [files] for a file name [name] and compare in a
   // case-insensitive manner.
   //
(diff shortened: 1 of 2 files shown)


app/lib/frontend/backend.dart
```diff
@@ -860,6 +860,19 @@ Future<_ValidatedUpload> _parseAndValidateUpload(
...
+  // (e.g. on Windows). We can't allow two files with the same case-insensitive
+  // name.
+  final lowerCaseFiles = <String>{};
+  for (String file in files) {
+    final lower = file.toLowerCase();
+    if (lowerCaseFiles.contains(lower)) {
+      throw GenericProcessingException(
+          'Filename collision on case-preserving file systems: $file.');
+    }
+    lowerCaseFiles.add(lower);
```
jonasfj: Nit, but `Set.add` does return a boolean :)
```dart
final lowerCaseFiles = <String>{};
for (String f in files) {
  if (!lowerCaseFiles.add(f.toLowerCase())) {
    throw GenericProcessingException(....);
  }
}
```
// https://api.dartlang.org/stable/2.2.0/dart-core/Set/add.html

isoos: I find that negating add is less readable (I usually need to lookup what `add()` returns if item not already present). I find being explicit on the contains, but I don't mind changing it to add either...

jonasfj: ++ , it was just a nit... this is perfectly fine... :)

PR #2127, opened 2019-03-13
enable backup cronjob

jonasfj:


app/lib/frontend/cronjobs.dart
```diff
@@ -1,33 +1,77 @@
...
-      return Response.internalServerError(
-        body: 'cron failed ${request.url.path}',
-      );
-    }
-    _log.finest('### Cron successful: ${request.url.path}');
-    return Response.ok('cron successful: ${request.url.path}');
+
+    _log.info('### Cron starting: $path');
+    final c = Completer<bool>();
+    scheduleMicrotask(() async {
```
isoos: why the need for completer + scheduleMicrotask? Wouldn't it be simpler without it?

isoos: Never mind, I get it.

PR #2088, opened 2019-02-25
Authorize account before added as uploader

isoos:
The current single-page will be split into two:
- First an "uploader approval" view is presented, with a link to Google OAuth and a disclaimer about being added will also display the e-mail address.
- After approval the same page will check the auth credentials and if authorized, the new uploader will be added to the package.

URLs:
- The uploader confirmation URL is in the structure of: `/admin/confirm/new-uploader/<package>/<uploader-email>/<nonce>`.
- The OAuth redirect/callback URL will be `/oauth/callback`
- The OAuth `state` parameter will be the URL of the original confirmation page.
- The OAuth callback handler uses the `state` parameter and redirects the request to the original uploader confirmation page.

The uploader confirmation handler updates/confirms the request only if a valid auth `code` is present AND the verified user e-mail is the same as was on the invite.

PR #2134, opened 2019-03-20
I added the oauth clientId and updated `set_secret` to allow `'oauth-secret-'`-prefixed keys.

jonasfj:


jonasfj:
cla/yes, (I probably did an edit on github by mistake, which set the wrong email).

jonasfj:
Also see #2088 for most of the context on this change.

PR #2143, opened 2019-03-22
Fix: delete obsolete PackageInvite entries.

isoos:
- we didn't call the `deleteObsoleteInvites` method.
- I don't see a reason to keep the entity around for one more day after it expired.

### DIFF
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -265,12 +265,19 @@ class Backend {
     await uiPackageCache.invalidateUIPackagePage(invite.packageName);
   }
 
-  /// Removes obsolete (== expired more than a day ago) invites from Datastore.
+  /// Removes obsolete/expired invites from Datastore.
   Future deleteObsoleteInvites() async {
     final query = db.query<models.PackageInvite>()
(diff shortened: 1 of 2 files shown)


app/lib/frontend/backend.dart
```diff
@@ -265,12 +265,19 @@ class Backend {
     await uiPackageCache.invalidateUIPackagePage(invite.packageName);
   }
 
-  /// Removes obsolete (== expired more than a day ago) invites from Datastore.
+  /// Removes obsolete/expired invites from Datastore.
   Future deleteObsoleteInvites() async {
     final query = db.query<models.PackageInvite>()
-      ..filter('expires <', DateTime.now().subtract(Duration(days: 1)));
+      ..filter('expires <', DateTime.now().toUtc());
```
jonasfj: The reason  to wait a day is to avoid clock drift issues...

But our servers shouldn't drift by hours :)

jonasfj: so I doubt it's important :)

isoos: Let's deal with that once we have complaints about it :)

PR #2146, opened 2019-03-25
Refactoring configuration

isoos:


app/lib/shared/configuration.dart
```diff
@@ -46,63 +57,78 @@ class Configuration {
...
-  bool get hasCredentials => envConfig.hasCredentials;
+  final auth.ServiceAccountCredentials credentials;
 
   /// Create a configuration for production deployment.
   ///
   /// This will use the Datastore from the cloud project and the Cloud Storage
   /// bucket 'pub-packages'. The credentials for accessing the Cloud
   /// Storage is retrieved from the Datastore.
-  Configuration._prod() : this._local('dartlang-pub');
+  factory Configuration.prod() {
```
jonasfj: Is this ever called outside this file? if not, let's make it private :)

app/lib/shared/configuration.dart
```diff
@@ -46,63 +57,78 @@ class Configuration {
...
-            ? 'https://pub.dartlang.org/oauth/callback'
-            : 'https://dartlang-pub-dev.appspot.com/oauth/callback',
-        pubSiteAudience = projectId == 'dartlang-pub'
-            ? '818368855108-e8skaopm5ih5nbb82vhh66k7ft5o7dn3.apps.googleusercontent.com'
-            : '621485135717-idb8t8nnguphtu2drfn2u4ig7r56rm6n.apps.googleusercontent.com';
+  factory Configuration.dev() {
+    final projectId = 'dartlang-pub-dev';
+    return Configuration(
+      projectId: projectId,
+      packageBucketName: '$projectId--pub-packages',
```
jonasfj: **Unrelated:** but analysis and dartdoc still use prod bucket for packages on staging, right?
(we probably should change that, such that analysis and dartdoc uses the staging server by specifying `PUB_HOSTED_URL`, right?)

This just got me thinking that we would be able to test a lot more if we did that :)

PR #2150, opened 2019-03-26
No longer use uploaderEmail(s).

isoos:


app/lib/frontend/models.dart
```diff
@@ -197,13 +191,9 @@ class PackageVersion extends db.ExpandoModel {
   @db.IntProperty(propertyName: 'sort_order')
   int sortOrder;
 
-  // TODO: Set required attribute after [uploaderEmail] is removed.
-  @db.StringProperty()
+  @db.StringProperty(required: true)
   String uploader;
 
-  @db.StringProperty()
-  String uploaderEmail;
```
jonasfj: Is there a way to delete these properties from datastore in the future?

isoos: We will use this script:
https://github.com/dart-lang/pub-dartlang-dart/blob/master/app/bin/tools/clear_package_properties.dart

jonasfj: Sweet, can we put that in the `CHANGELOG` for next next version :)

PR #2151, opened 2019-03-26
Remove namespace fields.

isoos:
I think the idea of `namespace` is going away in this separate form. Maybe the `qualifiedVersion` should be revisited, the `-` may not be the best long-term separator there...


jonasfj:
please update change log...

PR #2155, opened 2019-03-27
Set background color to white

nicoeg:
Unavngivet

Steam's in-game browser uses a black background by default which makes it impossible to read the content on the site. Setting the background color to white should fix it.

### DIFF
--- a/static/css/style.css
+++ b/static/css/style.css
@@ -6,11 +6,8 @@
   box-sizing: border-box;
 }
 
-html {
-  background-color: #fff;
-}
-
 body {
+  background-color: #fff;
   line-height: 1.6;
   margin: 0;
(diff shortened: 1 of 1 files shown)


nicoeg:
I signed it!

static/css/style.css
```diff
@@ -6,6 +6,10 @@
   box-sizing: border-box;
 }
 
+html {
+  background-color: #fff;
```
isoos: Would that work on the `body` element too? if it does, please move it to there.

nicoeg: Yes that should work too, fixed.

isoos:
Not sure why the CI does not complete, but change looks good.

PR #2156, opened 2019-03-28
Use staging PUB_HOSTED_URL for staging services.

### LINKED ISSUE #2148: Use PUB_HOSTED_URL='https://dartlang-pub-dev.appspot.com'  for dartdoc/analysis on staging
Let's use `PUB_HOSTED_URL='https://dartlang-pub-dev.appspot.com' ` for dartdoc/analysis on staging, so we don't fetch from production this will give our API server a bit of load too :)

isoos:
Fixes #2148

### DIFF
--- a/app/lib/shared/configuration.dart
+++ b/app/lib/shared/configuration.dart
@@ -131,6 +131,13 @@ class Configuration {
       throw Exception('Unknown project id: ${env.gcloudProject}');
     }
   }
+
+  /// Whether the current project is the production one.
+  bool get isProdProjectId => projectId == 'dartlang-pub';
+
+  /// The PUB_HOSTED_URL to use in analyzer and dartdoc.
+  String get pubHostedUrl =>
+      isProdProjectId ? null : 'https://dartlang-pub-dev.appspot.com';
 }
 
 /// Configuration from the environment variables.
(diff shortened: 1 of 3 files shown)


app/lib/shared/configuration.dart
```diff
@@ -131,6 +131,13 @@ class Configuration {
       throw Exception('Unknown project id: ${env.gcloudProject}');
     }
   }
+
+  /// Whether the current project is the production one.
+  bool get isProdProjectId => projectId == 'dartlang-pub';
+
+  /// The PUB_HOSTED_URL to use in analyzer and dartdoc.
+  String get pubHostedUrl =>
+      isProdProjectId ? null : 'https://dartlang-pub-dev.appspot.com';
```
jonasfj: Can't we avoid this by always using `'https://${projectId}.appspot.com'` ? I don't see any downside to using that in production...

jonasfj: We could also just make this a regular field... so that it's set depending on wether the `_prod()` or `_dev()`  constructor is used... It's simple and consistent with how all fields are set..

Also no reason we couldn't use `https://pub.dartlang.org` in production...

isoos: Updated, PTAL.

PR #2163, opened 2019-04-01
Enable inline <br> in markdown.

### LINKED ISSUE #2153: Line break html code visible
URL: https://pub.dartlang.org/packages/flutter_map

It looks like there is meant to be a line break in this documentation, but there was a typo.

Image here:

https://user-images.githubusercontent.com/37519914/55016761-a607a980-4fac-11e9-994d-cc59ba5c6063.png

isoos:
Fixes #2153

### DIFF
--- a/app/lib/shared/markdown.dart
+++ b/app/lib/shared/markdown.dart
@@ -8,15 +8,21 @@ import 'package:pana/pana.dart' show getRepositoryUrl;
 
 final Logger _logger = Logger('pub.markdown');
 
-final List<m.InlineSyntax> _inlineSyntaxes = m
-    .ExtensionSet.gitHubWeb.inlineSyntaxes
-    .where((s) => s is! m.InlineHtmlSyntax)
-    .toList();
+List<m.InlineSyntax> _inlineSyntaxes;
+
+void _initInlineSyntaxes() {
+  if (_inlineSyntaxes != null) return;
+  _inlineSyntaxes = m.ExtensionSet.gitHubWeb.inlineSyntaxes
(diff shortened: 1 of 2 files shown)


app/lib/shared/markdown.dart
```diff
@@ -63,6 +69,17 @@ class _HashLink implements m.NodeVisitor {
   }
 }
 
+/// Enables inline <br> to be emitted as <br/> HTML element.
+class _InlineBrHtmlSyntax extends m.TextSyntax {
+  _InlineBrHtmlSyntax() : super(r'(<br>)|(<br\s*/>)|(<BR>)|(<BR\s*/>)');
```
jonasfj: Wouldn't it be better to:
A) move this upstream, and ensure that the `markdown` package has a safe subset of inline-HTML supported?
B) handle more generic cases? ie. a whitelist of permitted tags or something like that?

isoos: It could be better code-wise, but the effort/benefit doesn't yield us that much.

jonasfj: So far there is one package that has this problem, it's not like property HTML rendering is an urgent feature we need to deliver :)

Reading on package `markdown` it refers to: https://github.com/showdownjs/showdown/wiki/Markdown%27s-XSS-Vulnerability-(and-how-to-mitigate-it)

Which leads me to think we need a separate package for sanitizing HTML... hmm..

isoos: The `[some text](javascript:alert('xss'))` gets us, but the rest is a non-issue, as we don't allow inline HTMLs (besides this `<br>`)

PR #2166, opened 2019-04-01
Use mono_repo v2, merge smoke_test jobs

kevmoo:


isoos:
Did it change how we should run the `mono_repo` tool? Can you check it here?
https://github.com/dart-lang/pub-dartlang-dart/blob/master/doc/development.md#working-with-mono_repo

kevmoo:
> Did it change how we should run the `mono_repo` tool? Can you check it here?
> https://github.com/dart-lang/pub-dartlang-dart/blob/master/doc/development.md#working-with-mono_repo

Updated @isoos

PR #2169, opened 2019-04-02
Fix configuration initialization.

isoos:
#2146 changed the way we load the security credentials: previously it was lazily initialized, and only accessed after a `hasCredentials` check. Now we are trying to load it early, and we should allow to return null in that method. I've added an extra logging for it.

### DIFF
--- a/app/lib/shared/configuration.dart
+++ b/app/lib/shared/configuration.dart
@@ -5,7 +5,10 @@
 import 'dart:io';
 
 import 'package:googleapis_auth/auth.dart' as auth;
+import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
+
+final _logger = Logger('configuration');
 
 Configuration _configuration;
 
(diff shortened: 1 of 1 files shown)


app/lib/shared/configuration.dart
```diff
@@ -185,8 +188,14 @@ class EnvConfig {
 /// Configuration from the environment variables.
 final EnvConfig envConfig = EnvConfig._detect();
 
-auth.ServiceAccountCredentials _loadCredentials([String path]) {
-  path ??= envConfig.gcloudKey;
-  final content = File(path).readAsStringSync();
-  return auth.ServiceAccountCredentials.fromJson(content);
+auth.ServiceAccountCredentials _loadCredentials() {
```
jonasfj: How does this work in production? there is no `GCLOUD_KEY` variable...

isoos: It has a different code path:
https://github.com/dart-lang/pub-dartlang-dart/blob/master/app/bin/service/frontend.dart#L166-L175

The issue is that without this change, it will throw an exception regardless of that check.

PR #2170, opened 2019-04-02
Self-contained dependency graph checker.

isoos:
- uses `PackageVersionPubspec` instead of `PackageVersion`, much faster Datastore responses (half the total time on my dev machine)
- triggers the notifications internally via the callback function

The later enables us to decouple it from the package upload callback of the `frontend`, making it possible to move it out of `frontend`, and have it running in `analyzer` and `dartdoc` separately.


isoos:
> Okay, so we won't be triggering jobs from the post-upload handler anymore..., right?

Correct. I don't see it as a big issue, as the analysis may not start right away in either case.

Issue #2089, opened 2019-02-25
Integration tests for "pub get" and "pub publish"


jonasfj:
We should have integration-style tests using `pub get` and `pub publish`. Hopefully, we can write some in `test/frontend/` that runs the frontend and tests if `pub get` works, etc..

It would also be nice to have a tool in `app/bin/tools/` that tests whether `pub get` and `pub publish` work on the production/staging site... publishing a dummy package might be fine in that case..

PR #2180, opened 2019-04-04
Fake pub server MVP

isoos:
- #2089 
- Frontend server on 8080, storage server on 8081.
- It can accept a package upload, and after that it can render the package page, but the download is not working yet (lack of time, will do it in a subsequent PR).
- Search (and anything that relies on it) does not work yet, it will be added later.
- Analysis and dartdoc is not running, not triggered.
- The resolved user id and e-mail is always the same, any token works.
- There is no real upload signing, nor a real verification. That part could be simpler a bit, but once I got it working I kind of gave up on refinements, maybe in a later work.
- It lacks any feedback that would indicate that the app is running and operational. The integration test could just ping certain URLs and decide the internal state on that.

### DIFF
--- a/pkg/fake_pub_server/lib/fake_pub_server.dart
+++ b/pkg/fake_pub_server/lib/fake_pub_server.dart
@@ -1,0 +1,112 @@
+import 'package:fake_gcloud/mem_datastore.dart';
+import 'package:fake_gcloud/mem_storage.dart';
+import 'package:gcloud/db.dart';
+import 'package:logging/logging.dart';
+import 'package:pub_server/repository.dart';
+import 'package:pub_server/shelf_pubserver.dart';
+import 'package:shelf/shelf.dart' as shelf;
+
+import 'package:pub_dartlang_org/account/backend.dart';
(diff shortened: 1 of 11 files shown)


pkg/fake_pub_server/lib/fake_pub_server.dart
```diff
@@ -0,0 +1,112 @@
...
+import 'package:pub_dartlang_org/shared/analyzer_client.dart';
+import 'package:pub_dartlang_org/shared/configuration.dart';
+import 'package:pub_dartlang_org/shared/handler_helpers.dart';
+import 'package:pub_dartlang_org/shared/package_memcache.dart';
+import 'package:pub_dartlang_org/shared/redis_cache.dart';
+import 'package:pub_dartlang_org/shared/storage.dart';
+
+final _logger = Logger('fake_server');
+
+class FakePubServer {
```
jonasfj: Okay, this is a good start... IMO it would great if we had a dependency injection to the point that we didn't need to do this in a separate package... But let's not open that can of worms now.

PR #2182, opened 2019-04-05
Fix NPE in obsolete dartdoc deletion.

### LINKED ISSUE #2175: NPE in job handling
```
pub.job: dartdoc job error angular 5.0.0-alpha+7

Error:
   Invalid argument(s)

Stack:
   #0      _RegExp.firstMatch (dart:core/runtime/libregexp_patch.dart:192:25)
   #1      new Version.parse (package:pub_semver/src/version.dart:114:36)
   #2      DartdocBackend.removeObsolete (package:pub_dartlang_org/dartdoc/backend.dart:300:31)
   <asynchronous suspension>
   #3      DartdocJobProcessor.process (package:pub_dartlang_org/dartdoc/dartdoc_runner.dart:291:26)
   <asynchronous suspension>
   #4      JobProcessor.run (package:pub_dartlang_org/job/job.dart:55:28)
   <asynchronous suspension>
   #5      JobMaintenance.run (package:pub_dartlang_org/job/job.dart:95:18)
   #6      _workerMain.<anonymous closure> (file:///project/app/bin/service/dartdoc.dart:81:26)
   #7      _asyncThenWrapperHelper.<anonymous closure> (dart:async/runtime/libasync_patch.dart:77:64)
   #8      StackZoneSpecification._registerUnaryCallback.<anonymous closure>.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:129:26)
   #9      StackZoneSpecification._run (package:stack_trace/src/stack_zone_specification.dart:209:15)
   #10     StackZoneSpecification._registerUnaryCallback.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:129:14)
   #11     _rootRunUnary (dart:async/zone.dart:1132:38)
   #12     _CustomZone.runUnary (dart:async/zone.dart:1029:19)
   #13     _FutureListener.handleValue (dart:async/future_impl.dart:126:18)
   #14     Future._propagateToListeners.handleValueCallback (dart:async/future_impl.dart:639:45)
   #15     Future._propagateToListeners (dart:async/future_impl.dart:668:32)
   #16     Future._complete (dart:async/future_impl.dart:473:7)
   #17     _SyncCompleter.complete (dart:async/future_impl.dart:51:12)
   #18     _AsyncAwaitCompleter.complete (dart:async/runtime/libasync_patch.dart:28:18)
  ```

### PR DESCRIPTION
**Pull request by isoos**
Fixes #2175

### DIFF
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -296,7 +296,9 @@ class DartdocBackend {
         if (index >= 0) {
           final entry = completedList[index];
           completedList.removeAt(index);
-          versions.remove(Version.parse(entry.runtimeVersion));
+          if (entry.runtimeVersion != null) {
+            versions.remove(Version.parse(entry.runtimeVersion));
+          }
         }
       }
 
@@ -312,7 +314,6 @@ class DartdocBackend {
 
       // Keep the otherwise latest version (may be an ongoing release).
       if (versions.isNotEmpty) {
-        if (versions.isEmpty) return;
         final version = versions.removeLast();
         final index = completedList.lastIndexWhere(
             (entry) => entry.runtimeVersion == version.toString());
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -282,7 +282,6 @@ class DartdocBackend {
     if (completedList.isNotEmpty) {
       final versions = completedList
           .map((entry) => entry.runtimeVersion)
-          .where((s) => s != null) // protect against old entries without it
           .toSet()
           .map((String version) => Version.parse(version))
           .toList();
@@ -296,9 +295,7 @@ class DartdocBackend {
         if (index >= 0) {
           final entry = completedList[index];
           completedList.removeAt(index);
-          if (entry.runtimeVersion != null) {
-            versions.remove(Version.parse(entry.runtimeVersion));
-          }
+          versions.remove(Version.parse(entry.runtimeVersion));
         }
       }
 
--- a/app/lib/dartdoc/models.dart
+++ b/app/lib/dartdoc/models.dart
@@ -75,7 +75,9 @@ class DartdocEntry {
     @required this.hasContent,
     @required this.archiveSize,
     @required this.totalSize,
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: app/lib/dartdoc/backend.dart [hunk 1]
```diff
@@ -296,7 +296,9 @@ class DartdocBackend {
         if (index >= 0) {
           final entry = completedList[index];
           completedList.removeAt(index);
-          versions.remove(Version.parse(entry.runtimeVersion));
+          if (entry.runtimeVersion != null) {
```
jonasfj: Is this a really old entity? If so, how about we write a tiny tool to delete it? and remove associated resources...

Maybe a tool to list number of things associated with this, and then delete it? Isn't that better than making exceptions for every case of old data we have in datastore that we're never going to use again?

I assume we're never going to use this again, is that wrong?

isoos: I think a tool would only duplicate this logic (e.g. we want to preserve the latest serving version, even if it does not have a `runtimeVersion`). While it would be nice to have it as a separate tool, I think the overhead to write and test it does not really provide that much value.

jonasfj:
That makes sense, can we leave a comment why we have this that way... And
can we document in the model that old entries upon which we still rely
might not have this attribute?

Then we also need to search the entire codebase to make sure we never
assume it's not null...

Or we could consider removing the entries and either not have docs for
these packages or regenerate the docs..

My point is, lets aim to have a consistent state on datastore... We can do
this workaround and file an issue to follow-up and clean up the datastore
too.

This must be old versions or dart 1.x packages?

isoos:
Good points! I've updated the code the use the "epoch" `runtimeVersion` if the json file is missing it. That would protect us from NPEs, and we can still decide later if/when we want to remove the documentation for such packages.

app/lib/dartdoc/models.dart
```diff
@@ -75,7 +75,7 @@ class DartdocEntry {
     @required this.hasContent,
     @required this.archiveSize,
     @required this.totalSize,
-  });
+  }) : runtimeVersion = runtimeVersion ?? versions.dartdocRuntimeEpoch;
```
jonasfj: nit, but a comment about old versions not having this would be great.

PR #2188, opened 2019-04-10
Apply CSP to all HTML pages

### LINKED ISSUE #2185: Apply CSP to dartdoc pages

isoos:
- fixes #2185
- Moved CSP header extension to a top-level handler wrapper.
- removed hostname from dartdoc customization (we required it while we were serving out of dartdocs.org)
- CSP needed an extra host, because older dartdoc content refers to the github style with full URL.


app/lib/shared/handler_helpers.dart
```diff
@@ -33,6 +33,7 @@ Future<void> runHandler(
   }
   handler = _redirectToHttpsWrapper(handler);
   handler = _logRequestWrapper(logger, handler);
+  handler = _cspHeaderWrapper(handler);
```
jonasfj: Isn't this the same as self? Well, maybe not when we're on new domain :)

app/lib/shared/handler_helpers.dart
```diff
@@ -49,6 +50,24 @@ Future<void> runHandler(
...
+shelf.Handler _cspHeaderWrapper(shelf.Handler handler) {
+  return (shelf.Request request) async {
+    final rs = await handler(request);
+    final contentType = rs.headers['content-type'];
+    final isHtml = contentType != null && contentType.startsWith('text/html');
+    if (isHtml) {
+      return rs.change(headers: {
+        'x-content-type-options': 'nosniff',
+        'content-security-policy': contentSecurityPolicy,
+      });
```
jonasfj: Don't we also need something about frame options to protect against click jacking in old browser..
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options

isoos: `X-Frame-Options: sameorigin`?

app/lib/shared/handler_helpers.dart
isoos: The issue is with absolute URLs, and yeah, we may need to add an extra rule if we move to a new domain. But this PR already reduces the absolute URLs is dartdoc content, and hopefully the remaining ones are not heavily used.

app/lib/shared/handler_helpers.dart
jonasfj: do we use frames anywhere? If not we can just forbid it, and allow it when we need to...

app/lib/shared/handler_helpers.dart
jonasfj: ++

app/lib/shared/handler_helpers.dart
isoos: ack, `deny` it is.

PR #2191, opened 2019-04-10
Do not create User for hasUploader check.

isoos:
I can't really decide what to make of the `lookupOrCreateUserByEmail` method name. We could use
`lookupUserByEmail` with the same signature, or make `create = false` by default, and change the current callers to follow the new signature. wdyt?

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -121,7 +121,8 @@ class AccountBackend {
   /// not exists.
   ///
   /// Throws Exception if more then one `User` entry exists.
-  Future<User> lookupOrCreateUserByEmail(String email) async {
+  Future<User> lookupOrCreateUserByEmail(String email,
+      {bool create = true}) async {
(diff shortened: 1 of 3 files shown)


app/lib/account/backend.dart
```diff
@@ -121,7 +121,8 @@ class AccountBackend {
   /// not exists.
   ///
   /// Throws Exception if more then one `User` entry exists.
-  Future<User> lookupOrCreateUserByEmail(String email) async {
+  Future<User> lookupOrCreateUserByEmail(String email,
```
jonasfj: How about creating two methods `lookupOrCreateUserByEmail` and `lookupUserByEmail`.

We can use `lookupUserByEmail` inside `lookupOrCreateUserByEmail`, so we don't have to do the lookup logic twice...

isoos: PTAL, I've updated the code to use the two methods.

PR #2192, opened 2019-04-10
Trigger analysis and dartdoc generation on upload.

### LINKED ISSUE #2187: Awaiting analysis to complete.
I upgrade my plugin today.
Twenty minutes later, there is no score yet.

url: https://pub.dartlang.org/packages/flutter_ijkplayer#-analysis-tab-

isoos:
#2170 removed the immediate trigger on upload, I'm re-adding that part, without the dependency graph triggers, I think those can happen later.

Fixes #2187

### DIFF
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -19,6 +19,8 @@ import 'package:uuid/uuid.dart';
 import '../account/backend.dart';
 import '../history/backend.dart';
 import '../history/models.dart';
+import '../shared/analyzer_client.dart';
+import '../shared/dartdoc_client.dart';
 import '../shared/email.dart';
 import '../shared/package_memcache.dart';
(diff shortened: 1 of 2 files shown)


app/lib/frontend/backend.dart
```diff
@@ -534,6 +536,12 @@ class GCloudPackageRepository extends PackageRepository {
       ),
     );
 
+    // Trigger analysis and dartdoc generation
+    await analyzerClient.triggerAnalysis(
+        newVersion.package, newVersion.version, <String>{});
```
jonasfj: Please document that you left out dependent packages as this will be done later by a background process.

isoos:
> If you're sure this won't stop the background process from detecting that a new package version was added and thus running dependencies, I'm all for it :)

The only risk is that it may duplicate the analyzer and the dartdoc jobs for this `PackageVersion`.

PR #2198, opened 2019-04-11
Refactor template cache

isoos:
Enabling `escapeValues: false` was a bit strange, we should rather use the default escaping and mark the non-escaped values in the template (with `{{{` or `{{&`). I've checked the templates and all of the values that are now marked as such are prepared with HTML escaping in the preparation step.

The pre-loading of the template is better for the dual-setup template rendering: we'll have simpler logic when we are about to decide which template to use.

### DIFF
--- a/app/lib/frontend/templates/_cache.dart
+++ b/app/lib/frontend/templates/_cache.dart
@@ -30,10 +30,8 @@ class TemplateCache {
       (file) {
         final t = mustache.Template(file.readAsStringSync(), lenient: true);
         final relativePath = path.relative(file.path, from: templateFolder);
-        final ext = path.extension(relativePath);
-        final simplePath =
-            relativePath.substring(0, relativePath.length - ext.length);
-        _parsedMustacheTemplates[simplePath] = t;
+        final name = path.withoutExtension(relativePath);
+        _parsedMustacheTemplates[name] = t;
       },
     );
   }
--- a/app/lib/frontend/templates/_cache.dart
+++ b/app/lib/frontend/templates/_cache.dart
@@ -13,25 +13,37 @@ final templateCache = TemplateCache();
 
 /// Loads, parses, caches and renders mustache templates.
 class TemplateCache {
-  /// The path to the directory which contains mustache templates.
-  ///
-  /// The path should not contain a trailing slash (e.g. "/tmp/views").
-  final String _templateDirectory;
-
   /// A cache which keeps all used mustache templates parsed in memory.
-  final Map<String, mustache.Template> _cachedMustacheTemplates = {};
+  final Map<String, mustache.Template> _parsedMustacheTemplates = {};
 
-  TemplateCache({String templateDirectory})
-      : _templateDirectory = templateDirectory ??
-            path.join(resolveAppDir(), 'lib/frontend/templates/views');
+  TemplateCache() {
+    _loadDirectory(path.join(resolveAppDir(), 'lib/frontend/templates/views'));
+  }
+
+  void _loadDirectory(String dirPath) {
+    Directory(dirPath)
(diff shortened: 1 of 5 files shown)


app/lib/frontend/templates/_cache.dart
```diff
@@ -13,28 +13,37 @@ final templateCache = TemplateCache();
...
-    final mustache.Template parsedTemplate =
-        _cachedMustacheTemplates.putIfAbsent(template, () {
-      final file = File('$_templateDirectory/$template.mustache');
-      return mustache.Template(file.readAsStringSync(),
-          htmlEscapeValues: escapeValues, lenient: true);
-    });
+  String renderTemplate(String template, values) {
+    final parsedTemplate = _parsedMustacheTemplates[template];
+    if (parsedTemplate == null) {
+      throw Exception('Template $template was not found.');
```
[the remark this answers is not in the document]
jonasfj: Using `ArgumentError` is probably fine :)

app/lib/frontend/templates/views/layout.mustache
```diff
@@ -15,8 +15,8 @@
   <meta name="twitter:card" content="summary" />
   <meta name="twitter:site" content="@dart_lang" />
   <meta property="og:site_name" content="Dart Packages" />
-  <title>{{title}}</title>
-  <meta property="og:title" content="{{title}}" />
+  <title>{{& title}}</title>
+  <meta property="og:title" content="{{& title}}" />
```
jonasfj: Is there any harm in escaping this twice? Or is there any problem using HTML escape provided by mustache instead of he custom one we have?

I suppose it's because it's a page title?

app/lib/frontend/templates/views/pagination.mustache
```diff
@@ -5,8 +5,8 @@
 <ul class="pagination">
   {{#page_links}}
     <li class="{{#active}}-active{{/active}}{{#disabled}}-disabled{{/disabled}}">
-      <a href="{{#render_link}}{{href}}{{/render_link}}"{{#rel_prev}} rel="prev"{{/rel_prev}}{{#rel_next}} rel="next"{{/rel_next}}>
-        <span>{{{text}}}</span>
+      <a href="{{#render_link}}{{& href}}{{/render_link}}"{{#rel_prev}} rel="prev"{{/rel_prev}}{{#rel_next}} rel="next"{{/rel_next}}>
+        <span>{{& text}}</span>
```
jonasfj: Why do we have custom escaping here?

app/lib/frontend/templates/views/layout.mustache
isoos: `{{&` is the non-escaped content. We already escaped the `title` variable in the preparation step.

app/lib/frontend/templates/views/pagination.mustache
isoos: `{{{ }}}` is the same as `{{& }}` == non-escaped content. I've changed the three `{` to the other version because I think it is easier to notice what's happening there.

app/lib/frontend/templates/_cache.dart
```diff
@@ -13,28 +13,37 @@ final templateCache = TemplateCache();
...
+        .listSync(recursive: true)
+        .where((fse) => fse is File)
+        .cast<File>()
+        .where((f) => f.path.endsWith('.mustache'))
+        .forEach(
+      (file) {
+        final t = mustache.Template(file.readAsStringSync(), lenient: true);
+        final relativePath = path.relative(file.path, from: dirPath);
+        final ext = path.extension(relativePath);
+        final simplePath =
```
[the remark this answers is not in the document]
isoos: TIL

PR #2199, opened 2019-04-12
Better correlation of dartdoc runner's log.

isoos:


jonasfj:
This convinces me that we need structured logging at some point :)

isoos:
> This convinces me that we need structured logging at some point :)
++

PR #2201, opened 2019-04-12
Dual-setup for experimental templates and styles.

isoos:
- Templates are now autoresolved to `${dir}/experimental_${file}` if experimental mode is on and the experimental template exists.
- `<body class="experimental">` is applied if experimental mode is on. I'm planning to use it as a differentiator rule as the template structures may co-evolve with the current ones.

### DIFF
--- a/app/lib/frontend/templates/layout.dart
+++ b/app/lib/frontend/templates/layout.dart
@@ -51,9 +51,8 @@ String renderLayoutPage(
       : serializeSearchOrder(searchQuery.order);
   final platformDict = getPlatformDict(platform);
   final isRoot = type == PageType.landing && platform == null;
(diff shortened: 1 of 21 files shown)


app/lib/frontend/templates/layout.dart
```diff
@@ -50,7 +51,9 @@ String renderLayoutPage(
       : serializeSearchOrder(searchQuery.order);
   final platformDict = getPlatformDict(platform);
   final isRoot = type == PageType.landing && platform == null;
+  final bodyClass = requestContext.isExperimental ? 'experimental' : '';
   final values = {
+    'body_class': bodyClass,
```
jonasfj: nit, but why not just `'body_class': requestContext.isExperimental ? 'experimental' : ''`

Issue #2194, opened 2019-04-11
Put a cache-control header on `static/`


jonasfj:
This is a simple win, we don't need a high cache expiration value to get a decent win...

5-10 minutes is better than nothing.

**Problem:** When updating the static content and deploying a new release, you might get new HTML with old static files from cache..

We could solve that by serving `/static/` with a prefix: `/static/<hash>/` where `<hash>` is computed from the contents of the files in `/static/`... We could also attach hash in the query string.. which I think we already do for some files.

isoos:
> We could solve that by serving /static/ with a prefix: /static/<hash>/ where <hash> is computed from the contents of the files in /static/... We could also attach hash in the query string.. which I think we already do for some files.

This will solve it only if we store the old versions alongside the new ones, which would be either manually copying, or storing them as part of the deploy. With the current rate and scope of changes, I'm not sure if it is worth doing that.

jonasfj:
For `https://pub.dartlang.org/static/img/dart-logo.svg` we currently use:
```
cache-control: max-age: 300
```
but we should be using:
```
cache-control: max-age=300
```

jonasfj:
Let's also:
 A) Refuse to add a `cache-control` header to requests that don't have a `?hash=...` value,
 B) Set cache control to a week or something like that.

Note. we don't have to verify the hash value, if we do, we can just shout an error in logs when it's mismatching... we don't have to return 5xx...

isoos:
> Refuse to add a cache-control header to requests that don't have a ?hash=... value,
> B) Set cache control to a week or something like that.
> [..] we don't have to verify the hash value

The current code checks the hash value, and if it matches the file's, it adds a 24-hours long cache age (barring the `:` vs `=`), otherwise 5-minutes of caching. I'll increase the long cache time to a week, and remove it otherwise, and increase the number of hashed assets.

PR #2203, opened 2019-04-15
Fix cache headers

isoos:
- #2194
- increased max-age to a week for hashed assets
- fixed header value

### DIFF
--- a/app/lib/frontend/handlers/misc.dart
+++ b/app/lib/frontend/handlers/misc.dart
@@ -59,19 +59,17 @@ Future<shelf.Response> staticsHandler(shelf.Request request) async {
       return shelf.Response.notModified();
     }
     final String hash = request.requestedUri.queryParameters['hash'];
-    final Duration cacheAge = hash != null && hash == staticFile.etag
-        ? staticLongCache
-        : staticShortCache;
-    return shelf.Response.ok(
-      staticFile.bytes,
-      headers: {
-        HttpHeaders.contentTypeHeader: staticFile.contentType,
-        HttpHeaders.contentLengthHeader: staticFile.bytes.length.toString(),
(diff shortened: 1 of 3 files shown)


app/lib/frontend/handlers/misc.dart
```diff
@@ -59,19 +59,17 @@ Future<shelf.Response> staticsHandler(shelf.Request request) async {
...
-        HttpHeaders.cacheControlHeader: 'max-age: ${cacheAge.inSeconds}',
-      },
-    );
+    final headers = <String, String>{
+      HttpHeaders.contentTypeHeader: staticFile.contentType,
+      HttpHeaders.contentLengthHeader: staticFile.bytes.length.toString(),
+      HttpHeaders.lastModifiedHeader: formatHttpDate(staticFile.lastModified),
+      HttpHeaders.etagHeader: staticFile.etag,
+    };
+    if (hash == staticFile.etag) {
```
jonasfj: Can we do `&& hash.isNotEmpty` please :)

app/lib/frontend/handlers/misc.dart
```diff
@@ -59,19 +59,17 @@ Future<shelf.Response> staticsHandler(shelf.Request request) async {
...
+      headers[HttpHeaders.cacheControlHeader] =
+          'max-age=${staticLongCache.inSeconds}';
```
jonasfj: Let's also make it private, following the rationale that we're over HTTPS, so hopefully there is no proxies.
Either way, we should make it explicit that it's either public or private, and private just seems most safe :)

isoos: Isn't private the default? Anyway, setting it won't hurt.

jonasfj: Uh, maybe, I don't know :)

jonasfj:
r+

Issue #1518, opened 2018-08-10
Publisher profiles


mit-mit:
Potential feature idea: Add the notion of a ‘publisher’ of a package. This is similar to the publisher shown in traditional apps stores. It adds a publisher profile page that the uploader can maintain themselves, and this page is hyperlinked from all packages uploaded by that publisher.

PR #2207, opened 2019-04-16
Publisher Datastore model

isoos:
#1518

- I remember that we've talked about `Publisher` containing the member entries, but I don't think we should do that. Having a separate entity for the membership allows us to track invites and roles separately.
- The invite also differs from the `PackageInvite` because it doesn't need a nonce, we can do it on the pub site, without a shared secret.
- Added `publisherId` to both `Package` and `PackageVersion`, the later to track if a specific version was published under the publisher or earlier (or under a different publisher if it gets moved between them).

### DIFF
--- a/app/lib/frontend/models.dart
+++ b/app/lib/frontend/models.dart
@@ -50,6 +50,10 @@ class Package extends db.ExpandoModel {
   @db.ModelKeyProperty(propertyName: 'latest_dev_version')
   db.Key latestDevVersionKey;
 
+  /// The publisher id (if the package is associated with a publisher).
+  @db.StringProperty()
+  String publisherId;
+
   /// List of User.userId
   @db.StringListProperty()
   List<String> uploaders;
@@ -196,6 +200,10 @@ class PackageVersion extends db.ExpandoModel {
   // TODO: set it to required: true once the integrity check script passes
   @db.StringProperty(required: false)
   String uploader;
+
+  /// The publisher id at the time of the upload.
+  @db.StringProperty()
+  String publisherId;
 
   // Convenience Fields:
 
--- a/app/lib/frontend/models.dart
+++ b/app/lib/frontend/models.dart
@@ -50,7 +50,7 @@ class Package extends db.ExpandoModel {
   @db.ModelKeyProperty(propertyName: 'latest_dev_version')
   db.Key latestDevVersionKey;
 
-  /// The publisher id (if the package is associated with a publisher).
+  /// The publisher id (null, if the package does not have a publisher).
   @db.StringProperty()
   String publisherId;
 
--- a/app/lib/publisher/models.dart
+++ b/app/lib/publisher/models.dart
@@ -19,6 +19,9 @@ class Publisher extends db.ExpandoModel {
   @db.StringProperty()
   String websiteUrl;
 
+  @db.StringProperty()
+  String contactEmail;
+
   @db.DateTimeProperty()
   DateTime created;
 
--- a/app/lib/publisher/models.dart
+++ b/app/lib/publisher/models.dart
@@ -83,7 +83,4 @@ class PublisherMember extends db.ExpandoModel {
   /// One of [PublisherMemberRole].
   @db.StringProperty()
   String role;
-
-  @db.BoolProperty()
-  bool isPublic;
 }
--- a/app/lib/publisher/models.dart
+++ b/app/lib/publisher/models.dart
@@ -8,7 +8,7 @@ import 'package:gcloud/db.dart' as db;
 @db.Kind(name: 'Publisher', idType: db.IdType.String)
 class Publisher extends db.ExpandoModel {
   /// The associated domain name of the publisher.
-  String get domainName => id as String;
+  String get publisherId => id as String;
 
   /// Markdown formatted description of the publisher.
   ///
@@ -33,7 +33,7 @@ class Publisher extends db.ExpandoModel {
 @db.Kind(name: 'PublisherInfo', idType: db.IdType.String)
 class PublisherInfo extends db.ExpandoModel {
   /// The associated domain name of the publisher.
-  String get domainName => id as String;
+  String get publisherId => id as String;
 
   @db.DateTimeProperty()
   DateTime updated;
@@ -72,7 +72,7 @@ class PublisherMember extends db.ExpandoModel {
   db.Key get publisherKey => parentKey;
 
   /// The associated domain name of the publisher.
-  String get domainName => publisherKey.id as String;
+  String get publisherId => publisherKey.id as String;
 
   /// The userId of the member.
   String get userId => id as String;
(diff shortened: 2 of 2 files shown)


isoos:
A structure concern to discuss: should we start a new directory or put it somewhere else? E.g. validating an uploader would tie this to the frontend's `Package` object, but otherwise it feels like it should be separate from it, and `hasUploader` check should call publisher/backend's check if needed.

app/lib/frontend/models.dart
```diff
@@ -197,6 +201,10 @@ class PackageVersion extends db.ExpandoModel {
   @db.StringProperty(required: false)
   String uploader;
 
+  /// The publisher id at the time of the upload.
+  @db.StringProperty()
+  String publisherId;
```
jonasfj: Do we have a use-case for this? If not I suggest we avoid tracking it..

We still plan to store the email of the uploader, right?

app/lib/frontend/models.dart
```diff
@@ -50,6 +50,10 @@ class Package extends db.ExpandoModel {
   @db.ModelKeyProperty(propertyName: 'latest_dev_version')
   db.Key latestDevVersionKey;
 
+  /// The publisher id (if the package is associated with a publisher).
```
jonasfj: Could we make it:
```
/// The publisher id (null, if the package does not have a publisher).
```

app/lib/publisher/models.dart
```diff
@@ -0,0 +1,89 @@
...
+  List<String> admins;
+
+  /// List of userIds that are public members of this publisher.
+  /// (their e-mail address is public information)
+  @db.StringListProperty()
+  List<String> publicMembers;
+}
+
+/// Values for [PublisherMember.role]
+abstract class PublisherMemberRole {
```
jonasfj: didn't we drop roles from the initial design?

jonasfj: I think, just having `member` and `pending` would be good.

app/lib/publisher/models.dart
```diff
@@ -0,0 +1,89 @@
...
+
+  @db.DateTimeProperty()
+  DateTime accepted;
+
+  /// One of [PublisherMemberRole].
+  @db.StringProperty()
+  String role;
+
+  @db.BoolProperty()
+  bool isPublic;
```
jonasfj: Shouldn't we drop this initially too?

app/lib/publisher/models.dart
```diff
@@ -0,0 +1,89 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:gcloud/db.dart' as db;
+
+/// Canonical publisher data.
+@db.Kind(name: 'Publisher', idType: db.IdType.String)
+class Publisher extends db.ExpandoModel {
+  /// The associated domain name of the publisher.
+  String get domainName => id as String;
```
jonasfj: Perhaps it's better to call this `publisherId` everywhere, and then leave in a comment that this is the verified domain?

Who knows, if people don't verify domains and don't want to use this, it could happen that in some future we simply allow creation of arbitrary publisherIds... We can't tell if this will be used or not, maybe users find domain verification too hard.

app/lib/publisher/models.dart
```diff
@@ -0,0 +1,89 @@
...
+
+  /// Accepted membership.
+  static const String member = 'member';
+
+  /// Invited but not accepted.
+  static const String pending = 'pending';
+}
+
+/// Stores the membership information of a single user.
+@db.Kind(name: 'PublisherMember', idType: db.IdType.String)
```
jonasfj: Why don't we need a nonce? Are you assuming people have logged into to pub before they were invited to join a publisher?

Or do we just want to rely on the email verification that Google does in it's oauth?

app/lib/frontend/models.dart
isoos: I think it is important to track if a given PackageVersion was published while under the ownership of a given publisher.

app/lib/publisher/models.dart
isoos: How would you track the administrators of a publisher?

app/lib/publisher/models.dart
isoos: Yes, I assumed that once we have login, we could remove the nonce-based notifications: we can still send an e-mail, but we require a login, and then the user clicks the approve button.

app/lib/frontend/models.dart
jonasfj: But we already track who uploaded it, so what does it matter whether they were given permission to do so through a publisher, or because they were uploader at the time... ?

I think we're best off leaving audit logging as a separate thing..

app/lib/publisher/models.dart
jonasfj: I think we just say that all members of a publisher are equal... all are administrators..

app/lib/frontend/models.dart
isoos: But if the upload happened while under a publisher, we will hide the uploaderEmail, and display only the publisher. Assume we have 1.0.0 under user@example.com, 1.1.0 with user@secret.com under flutter.io, 2.0.0-dev1 under user@example.com (de-transitioned from publisher). When displaying 1.1.0 we'll need to know whether to display the userEmail or not.

app/lib/publisher/models.dart
isoos: But we need to display at least one admin e-mail for the publisher. Some of the motivation behind publishers was that some team members don't want to expose their e-mail address while being able to publish the package.

app/lib/publisher/models.dart
jonasfj: Fair point, we trust that Google accounts have verified emails.

app/lib/publisher/models.dart
jonasfj: Oh, I was thinking we would have a contact email... and have a validation flow for said email address.

I imagine that users would often want to use a mailinglist... which isn't a specific user, but an email that can be validated..

isoos: Does Google's oauth flow work against an e-mail list?

jonasfj: Doubtful, an email list isn't a Google account... you could perhaps create a google account for your email list, but this seems unnatural...

app/lib/publisher/models.dart
isoos: Maybe we should start with an UUID identifier, and store the verified domain as a separate field?

app/lib/publisher/models.dart
isoos: I've added `contactEmail` to the `Publisher` entity to reflect this. However, this also means we should only have `admin` and `pending`, as first everybody that is being added to the publisher will be full-access administrator. Maybe I just leave it there for now, we can easily check if the consts is in use or not.

PR #2208, opened 2019-04-16
Bump dartdoc to 0.28.3+2 and markdown to 2.0.3

srawlins:
This bumps markdown to a version which properly escapes fenced code block "info strings."

### DIFF
--- a/app/lib/shared/versions.dart
+++ b/app/lib/shared/versions.dart
@@ -16,7 +16,7 @@ final RegExp runtimeVersionPattern = RegExp(r'\d{4}\.\d{2}\.\d{2}');
 /// Increment the version when a change is significant enough to trigger
 /// reprocessing, including: version change in pana, dartdoc, or the SDKs,
 /// or when an feature or bugfix should be picked up by the analysis ASAP.
-const String runtimeVersion = '2019.04.02';
+const String runtimeVersion = '2019.04.16';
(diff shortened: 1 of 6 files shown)


isoos:
@srawlins: you'll need to update the `runtimeVersion` too, and the hash in `versions_test.dart`.

srawlins:
Thanks, @isoos ! Done.

isoos:
Looks like we have a couple of markdown output that needs to be updated too. You can update the expected files in `templates_test.dart` with the `_regenerateGolden` flag, but I think the `markdown_test` needs manual updates.

jonasfj:
@isoos, might be easier if we just take it from here... It quickly starts to get very complicated..

srawlins:
Should I merge this as is, or wait a few hours till daytime in Europe? 👅

jonasfj:
Don't worry about we'll merge tomorrow...


Note, there is no rush, this won't get deployed anytime soon anyways. We have mitigation in place and there is very little to exploit at this point. So we do deployments in a calm an orderly fashion avoiding other mistakes :)

isoos:
> might be easier if we just take it from here... 

Ack, I'm doing a new PR with this.

isoos:
Closing, I'll will use this in #2209. There are a few more things to add to this one.

srawlins:
Excellent, thanks much!

PR #2213, opened 2019-04-17
Use dart2js --csp, and remove unsafe-eval CSP rule.

### LINKED ISSUE #2212: Content-Security-Policy block unsafe-eval
Requires us to use `dart2js --csp`.

I haven't tested this, but `dart2js -h -v` says this will work.

isoos:
Fixes #2212. Tested it locally, looks good.


isoos:
> Let's also check that dartdoc search works...

Confirmed.

I've also updated the `script.dart` to make sure we do trigger a new build of it.

PR #2215, opened 2019-04-18
Updated script.dart build trigger

isoos:
- introduced a `build.sh` script that invokes `dart2js`
- running the script from `Dockerfile`
- running the script from the static file initialization code, when any file in the `pkg/web_app` directory changes (this is useful for local development). previously it did check only the `script.dart` file, but (a) I'm planning to refactor that into multiple files, and (b) we also want to trigger the build when the build script changes.

### DIFF
--- a/pkg/web_app/build.sh
+++ b/pkg/web_app/build.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+# Compiles script.dart to JavaScript and places the output in the static/js directory.
 
 set -e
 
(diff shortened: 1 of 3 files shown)


pkg/web_app/build.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
```
jonasfj: Please write two lines at the top of this script that's says what it is used for... Like script to build client side dart to js..

Issue #1750, opened 2018-11-06
Reduce memory of search index


isoos:
The current in-memory index is getting ever-larger, and I think it would be worth to check if something is an easy win or not:
- using bitarray indexes as the inverted index base data structure
- storing weights in reduced resolution (instead of double values, we may be able to store them as bytes?)
- maybe store parts of it on disk and use memory-mapped reads (is this possible in Dart?)

jonasfj:
> maybe store parts of it on disk and use memory-mapped reads (is this possible in Dart?)

I would be surprised if it was...

Maybe we should explore throwing it into elasticsearch or something else.. and make it less of our problem.

isoos:
> Maybe we should explore throwing it into elasticsearch or something else.. and make it less of our problem.

My experience with operating ElasticSearch cluster and indexing was that it is always our problem. I'm referring to mostly the index tuning, but also the various failure cases that such clusters can go out-of-sync.

isoos:
I've did a quick experiment with the tokenization part: changing the max word length from 80 to 20 reduces the memory used by 20-25%, while changing it from 80 to 10 reduced the memory use by ~50%.

sigurdm:
That sounds like a nice easy win.
My guess is that searches for such long words are rare.
Can we test that hypothesis from the logs?

isoos:
Well, `geolocation` and `face recognition` are two recent searches, both have phrases longer than 10 characters, although just by 1. It is a nice plan-B if the number of packages jump a lot in a short time though...

sigurdm:
But not longer than 20...

sigurdm:
ByteData supports storing 32 bit floats in a packed array if you wanted to save memory on the weights.
It would require mapping document ids to integers

isoos:
In practice we don't even need 32 bits for the weights, I think we could get away with 4 or 5 bits and still keep the current ranking.

isoos:
I've done an experiment today using `package:bit_array` to create the inverted index. While it had a speed benefit, the memory requirement was about 3 times than the current index, and it took much more to build the index. For now I'd probably drop that line of exploration.

The next step could be to explore if the current index format could be read efficiently from a binary-file format, and then figuring out how to build it.

PR #2219, opened 2019-04-18
Reduce spelling correction candidates: no suffix list.

isoos:
This is in preparation to introduce trie to store the candidates.

~140 MB less memory (~1.8M word candidates removed). #1750

I was a bit worried on the (geo)location example, but it turns out that it is likely that the API of the package contains the location word separately or as part of a camelcased word, in the end we will find it. I didn't find any query that was doing worse with this change.

### DIFF
--- a/app/lib/search/text_utils.dart
+++ b/app/lib/search/text_utils.dart
@@ -130,7 +130,6 @@ Set<String> ngrams(String input, int minLength, int maxLength) {
 /// Generates lookup candidates that are, either:
 /// - derived by deleting one character from [token],
 /// - are the prefix part of [token] (min 4 characters)
-/// - are the suffix part of [token] (min 4 characters)
 Set<String> deriveLookupCandidates(String token) {
   final set = Set<String>();
   if (token.length <= 3) {
@@ -141,9 +140,6 @@ Set<String> deriveLookupCandidates(String token) {
     final suffix = i == token.length - 1 ? '' : token.substring(i + 1);
     if (prefix.length > 3) {
       set.add(prefix);
-    }
-    if (suffix.length > 3) {
-      set.add(suffix);
     }
     set.add(prefix + suffix);
   }
--- a/app/test/search/index_simple_test.dart
+++ b/app/test/search/index_simple_test.dart
@@ -94,11 +94,11 @@ void main() {
       for (String name in names) {
         index.add(name, name);
       }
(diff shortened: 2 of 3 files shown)


jonasfj:
> This is in preparation to introduce trie to store the candidates.

Why is this relevant for storing in a trie? Can't we still have suffixes? Isn't the trie just a different structure than using a map?

isoos:
> Why is this relevant for storing in a trie? Can't we still have suffixes? Isn't the trie just a different structure than using a map?

A trie is great for prefix-matching of any length. For proper suffix-matching we'll need to reverse both the indexed and the lookup strings, and it just feels much simpler if that subclass of problem is not present.

jonasfj:
> A trie is great for prefix-matching of any length. For proper suffix-matching we'll need to reverse both the indexed and the lookup strings, and it just feels much simpler if that subclass of problem is not present.

I suspect we can just insert every suffix in the trie, why not?

I imagine a trie is something like:
```
class Node {
  Map<String,Node> next = <String,Node>{};
  Set<String> matched = <String>{};
}

Node buildTrie(Set<String> tokens) {
  final root = Node();
  for (final token in tokens) {
    for (final candidate in deriveLookupCandidates(token)) {
      Node n = root;
      for(int i = 0; i < cancidate.length; i++) {
        final c = candicate[i];
        n.next[c] ??= Node();
        n = n.next[c];
      }
      n.matched.add(token);
    }
  }
  return root;
}
```

Maybe that's a bit naive... we can do a lot smarter for prefixes if we sort all the tokens... We can also do smarter for `matched` collecting them as we go when matching later... But we could easily just do the naive thing first, just to see how slow it is and how much memory it takes..

Regardless, we can easily insert the suffixes in the trie too... the construction is just a little less efficient if we don't reverse the and build a second trie.

isoos:
Closing this for now. Let's do the trie first, and see how memory is affected by the suffixes.

isoos:
> I imagine a trie is something like:

I've implemented this simple approach, and it added an extra 5GB memory with the suffixes, and 3.5GB without them. More importantly I've learned that we have ~21.4M nodes in the full, and ~16M nodes in the reduces (without suffixes) case. In both cases ~15% of the nodes were leaf nodes.

I have doubts if either a non-Map-based, or a non-String based version would yield better memory footprint than the current simple Map.

PR #2238, opened 2019-04-30
Extend list of production hosts.

isoos:


app/lib/shared/urls.dart
```diff
@@ -9,11 +9,17 @@ import 'package:pub_server/repository.dart' show GenericProcessingException;
 import 'packages_overrides.dart';
 import 'versions.dart';
 
-const pubHostedDomain = 'pub.dartlang.org';
+const _pubDartlangOrg = 'pub.dartlang.org';
+const _pubDev = 'pub.dev';
+const _apiPubDev = 'api.pub.dev';
```
jonasfj: Couldn't we configure some of this?

isoos: As discussed offline, we'll start with the requestContext-based config.

PR #2240, opened 2019-04-30
Use RequestContext to classify requested hostname.

isoos:


app/lib/frontend/request_context.dart
```diff
@@ -14,7 +14,22 @@ RequestContext get requestContext =>
 
 /// Holds flags for request context.
 class RequestContext {
+  /// Whether the request is on a production host (vs. staging or dev)
+  final bool isProductionHost;
```
jonasfj: What is this used for anymore?

app/lib/shared/handler_helpers.dart
```diff
@@ -58,12 +58,26 @@ Future<void> runHandler(
 /// Populates [requestContext] with the extracted request attributes.
 shelf.Handler _requestContextWrapper(shelf.Handler handler) {
   return (shelf.Request request) async {
-    final isPubDev = request.requestedUri.host == 'pub.dev';
+    final host = request.requestedUri.host;
+    final isProductionHost = host == 'pub.dartlang.org' ||
+        host == 'pub.dev' ||
+        host == 'api.pub.dev';
```
jonasfj: can we move these hostnames into the config file?

app/lib/frontend/request_context.dart
isoos: It is used to guard the cache updates in certain pages. Even if it is not used anymore, I'd rather remove it after next week.

PR #2242, opened 2019-05-01
Updated configuration and request context variables.

isoos:
- split `isProductionHost` into two to better represent the meaning of it:
  - `uiCacheEnabled` to control whether a request should use UI cache to serve the content
  - `searchEnginesEnabled` to control whether the request should serve `robots.txt` and `sitemap.txt` 
    - it also depends on the project-level configuration, only prod is enabled for search engines
    - to make it testable on staging, if the experimental cookie is set, we do serve the above files

- updated config to also contain the list of host names that are considered 'production' (and staging gets its own one)

- removed pkg/doc split, as we are not going to do it right now

### DIFF
--- a/app/lib/shared/configuration.dart
+++ b/app/lib/shared/configuration.dart
@@ -69,8 +69,8 @@ class Configuration {
   /// the Datastore.
   final auth.ServiceAccountCredentials credentials;
 
-  /// Whether indexing of the content by search engines is enabled.
-  final bool searchEnginesEnabled;
+  /// Whether indexing of the content by robots should be blocked.
+  final bool blockRobots;
 
   /// The list of hostnames which are considered production hosts (e.g. which
   /// are not limited in the cache use).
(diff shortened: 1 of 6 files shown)


app/lib/shared/configuration.dart
```diff
@@ -77,18 +84,21 @@ class Configuration {
...
+      popularityDumpBucketName: '$projectId--popularity',
+      searchSnapshotBucketName: '$projectId--search-snapshot',
+      backupSnapshotBucketName: '$projectId--backup-snapshots',
+      searchServicePrefix: 'https://search-dot-$projectId.appspot.com',
+      pubHostedUrl: 'https://pub.dartlang.org',
+      pubClientAudience: _pubClientAudience,
+      pubSiteAudience:
+          '818368855108-e8skaopm5ih5nbb82vhh66k7ft5o7dn3.apps.googleusercontent.com',
+      credentials: _loadCredentials(),
+      searchEnginesEnabled: true,
```
jonasfj: Wouldn't be it better to rename this `blockRobots` and the serve a `robots.txt` that blocks robots from the staging site, I assume that is the purpose here?

jonasfj: Also, I don't know maybe `searchEnginesEnabled` isn't the best name... when this service also features a search engine :)

Could we call it `robotsTxtEnabled`?

jonasfj: Or something like that... the "Txt" part was particularly ugly, hehe :)

isoos: `blockRobots` sounds better, renaming

PR #2246, opened 2019-05-01
pub and dart site urls in a central place

isoos:


app/lib/shared/urls.dart
```diff
@@ -207,9 +208,13 @@ String inferServiceProviderName(String url) {
   return null;
 }
 
+String panaUrl() {
```
kevmoo: `=>` syntax is sexy!

PR #2248, opened 2019-05-02
Update exp. search bar with new styles.

isoos:


pkg/web_css/lib/src/_base.scss
```diff
@@ -43,6 +43,10 @@ main {
   margin: 0 auto;
   padding: 0 20px;
 }
+// TODO: replace searchbar's 'main' element with something non-main
+body > main {
+  min-height: calc(100vh - 360px); // top nav + search banner + footer margin + footer padding
```
jonasfj: side note: there is limited IE 11 support for this... should we be concerned...

PR #2252, opened 2019-05-03
Legacy/primary host split.

isoos:
- most changes are updating the tests
- I've defined a blacklist of non-redirected patterns (or whitelist, depending on from where you are looking at it).


isoos:
> Don't we also want to bump the runtimeVersion?

Good point, adding it.

PR #2253, opened 2019-05-03
Fix dynamic redirect_uri calculation.

isoos:
Verified on staging with the `uploader add` flow.


jonasfj:
Uh, please update `CHANGELOG` to reflect this change...

PR #2257, opened 2019-05-06
Fix package removal: don't use removed qualifiedPackage field

isoos:


isoos:
> This is because we're removing the namespace field we've never used, right?

Yes, correct.

PR #2261, opened 2019-05-06
Increase HSTS duration.

isoos:


isoos:
> Please add a note in calendar that we should bump this to 1 year... at some point... that does not have to be now... maybe in a few months :)

I've added a calendar entry for myself to do it in August.

Issue #2281, opened 2019-05-16
Confusing upload error


nex3:
The only uploader listed for [the `sass_migrator` package](https://pub.dev/packages/sass_migrator) is `sass.bot.beep.boop@gmail.com`, but when that user tries to upload a new version, it gets an error message:

> ```
> Uploading...
> UnauthorizedAccess: Unauthorized user: sass.bot.beep.boop@gmail.com is not allowed to upload versions to package sass_migrator..
> Script failed with status 1
> ```

See [this Travis log](https://travis-ci.com/sass/migrator/jobs/200567426) for an example of this happening (click on "Deploying application" for the pub log).

mit-mit:
cc @jonasfj

isoos:
It looks like that the user was authenticated properly, but the resolved userId didn't match the approved uploader userids on the `Package` object.

[The upload history](https://pub.dartlang.org/api/history?package=sass_migrator&version=1.0.0-alpha.1) says the uploaderId was `bb60b590-e00d-4240-8ffe-cf675913e454`, while the log message says a different userId:
```
2019-05-16 03:06:01.844 CEST
pub.cloud_repository: User 8e4b508f-851f-467d-aa95-e8f94704adb7 (sass.bot.beep.boop@gmail.com) is not an uploader for package sass_migrator, rolling transaction back.
```

Staging has a user `8e4b508f-851f-467d-aa95-e8f94704adb7` with the single consistent `OAuthUserID` entry, I don't see issues there, but `sass_migrator` is a new package and it is not in staging yet.

@nex3:
- Was the upload (or the pub authentication) done with an aliased Google account?
- Was there anything special with the account done in the past few months?
- I see that it is also an uploader for sass, was it working properly?

@jonasfj:
- Could you run the integrity check?
- Could you check the `uploaders` field on the `Package` object for `sass_migrator`?

nex3:
> * Was the upload (or the pub authentication) done with an aliased Google account?

I'm not sure what this means.

> * Was there anything special with the account done in the past few months?

I don't think so...

> * I see that it is also an uploader for sass, was it working properly?

We haven't tried to release a new version of `sass` recently, so I'm not sure.

isoos:
>> Was the upload (or the pub authentication) done with an aliased Google account?
> I'm not sure what this means.

[...]

PR #2297, opened 2019-05-20
Remove support for legacy auth tokens (without user_id).

isoos:
#2281

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ AppEngine version, listed here to ease deployment and troubleshooting.
 ## Next Release (replace with git tag when deployed)
 
  * Removed support for legacy auth tokens (without `user_id`).
+   Users with legacy tokens will be requested to login again.
 
 ## `20190508??`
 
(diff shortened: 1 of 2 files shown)


jonasfj:
This deserves a changelog entry.

app/lib/account/backend.dart
```diff
@@ -398,6 +379,11 @@ class GoogleOauth2AuthProvider extends AuthProvider {
         return null;
       }
 
+      if (info.userId == null) {
```
jonasfj: Just uncomment the lines above which say:
```
          // TODO: Enable when pub client is requesting the correct scopes.
          // info.userId == null ||
          // info.userId.isEmpty ||
```

isoos: good catch!

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ AppEngine version, listed here to ease deployment and troubleshooting.
 
 ## Next Release (replace with git tag when deployed)
 
+ * Removed support for legacy auth tokens (without `user_id`).
```
sigurdm: Can we add something like:
```
Users with legacy tokens will be requested to login again.
```
?

isoos: Sure, added.

PR #2302, opened 2019-05-21
Guard agains post-processing issues around package upload.

isoos:
This is to cover the morning's issue with the pana publishing.

I don't think any of the steps warrant a retry (e.g. Jobs will be created eventually even if they fail here), a bulk try-catch should shield the users to the internals of the post-processing.

### DIFF
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -522,30 +522,35 @@ class GCloudPackageRepository extends PackageRepository {
       }
     });
 
-    final uploaderEmails =
-        await accountBackend.getEmailsOfUserIds(package.uploaders);
+    try {
+      final uploaderEmails =
+          await accountBackend.getEmailsOfUserIds(package.uploaders);
 
-    // Notify uploaders via e-mail that a new version has been published.
-    await emailSender.sendMessage(
-      createPackageUploadedEmail(
-        packageName: newVersion.package,
-        packageVersion: newVersion.version,
-        uploaderEmail: user.email,
-        authorizedUploaders:
-            uploaderEmails.map((email) => EmailAddress(null, email)).toList(),
-      ),
-    );
+      // Notify uploaders via e-mail that a new version has been published.
+      await emailSender.sendMessage(
+        createPackageUploadedEmail(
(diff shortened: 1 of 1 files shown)


app/lib/frontend/backend.dart
```diff
@@ -522,30 +522,35 @@ class GCloudPackageRepository extends PackageRepository {
...
-        .triggerAnalysis(newVersion.package, newVersion.version, <String>{});
-    await dartdocClient
-        .triggerDartdoc(newVersion.package, newVersion.version, <String>{});
+      // Trigger analysis and dartdoc generation. Dependent packages can be left
+      // out here, because the dependency graph's background polling will pick up
+      // the new upload, and will trigger analysis for the dependent packages.
+      await analyzerClient
+          .triggerAnalysis(newVersion.package, newVersion.version, <String>{});
+      await dartdocClient
+          .triggerDartdoc(newVersion.package, newVersion.version, <String>{});
```
jonasfj: Why not trigger these concurrently?

jonasfj: We could even do emailing concurrently with this...

app/lib/frontend/backend.dart
```diff
@@ -522,30 +522,35 @@ class GCloudPackageRepository extends PackageRepository {
...
 
-    if (finishCallback != null) {
-      await finishCallback(newVersion);
+      if (finishCallback != null) {
+        await finishCallback(newVersion);
```
jonasfj: Please document in a single line what this is used for?

isoos: I think we should probably remove this, it is no longer used. I'm not sure why we had it in the first place, but after a while we were using it for triggering the analysis, and then after a refactor we've got a new place for that.
Should I do this in the PR?

app/lib/frontend/backend.dart
isoos: updated to do that.

app/lib/frontend/backend.dart
jonasfj: I filed https://github.com/dart-lang/pub-dev/issues/2304

PR #2307, opened 2019-05-21
New tool: merge users.

isoos:
#2281

### DIFF
--- a/app/bin/tools/merge_users.dart
+++ b/app/bin/tools/merge_users.dart
@@ -1,0 +1,143 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:args/args.dart';
+import 'package:gcloud/db.dart';
+
+import 'package:pub_dartlang_org/account/models.dart';
+import 'package:pub_dartlang_org/frontend/models.dart';
+import 'package:pub_dartlang_org/frontend/service_utils.dart';
+import 'package:pub_dartlang_org/history/models.dart';
+
+final _argParser = ArgParser()
+  ..addOption('source-user-id',
+      help: 'The userId of the User that will be removed.')
+  ..addOption('target-user-id',
+      help:
+          'The userId of the User that will take over the entries of the removed User.')
+  ..addFlag('help', abbr: 'h', defaultsTo: false, help: 'Show help.');
+
+Future main(List<String> args) async {
+  final argv = _argParser.parse(args);
+  if (argv['help'] as bool == true) {
+    print(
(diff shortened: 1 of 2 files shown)


app/bin/tools/merge_users.dart
```diff
@@ -0,0 +1,143 @@
...
+
+    print('Found: ${packages.length} packages and '
+        '${versions.length} versions.');
+
+    final historyPackages = packages.map((p) => p.name).toSet();
+    versions.forEach((pv) => historyPackages.add(pv.package));
+    for (String package in historyPackages) {
+      print('Querying History for package $package...');
+      final query = dbService.query<History>()
+        ..filter('packageName =', package);
```
jonasfj: This won't work for packages that the user used to be an uploader for? and where the user didn't upload any versions, but only added/removed other uploaders.

jonasfj: We probably need to do a full table scan...

app/bin/tools/merge_users.dart
```diff
@@ -0,0 +1,143 @@
...
+        .single;
+    final targetUser = (await dbService
+            .lookup<User>([dbService.emptyKey.append(User, id: targetUserId)]))
+        .single;
+    if (targetUser == null) {
+      print('Target user does not exists.');
+      exit(-1);
+    }
+    if (targetUser.oauthUserId == null) {
+      print('Target user does not have oauth user id.');
```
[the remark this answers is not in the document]
isoos: I just wanted to make sure we are migrating in the good direction: e.g. accidentally not migrating a live user to an "old" User.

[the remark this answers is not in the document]
jonasfj: Print a warning instead, and ask for confirmation...

app/bin/tools/merge_users.dart
isoos: updated

app/bin/tools/merge_users.dart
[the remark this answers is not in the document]
isoos: Let's do that once we have more cases to deal with. I don't think it adds much value for our single current one.

jonasfj:
Can we do some confirm with:
```dart
print('please say yes:');
  if (stdin.readLineSync() != 'y') {
    exit(1);
  }
  print('YES');
```

app/bin/tools/merge_users.dart
```diff
@@ -0,0 +1,147 @@
...
+      print('Please say "yes" if you still want to continue, CTRL+C to abort:');
+      final answer = stdin.readLineSync().toLowerCase();
+      if (answer != 'y' && answer != 'yes') {
+        exit(1);
+      }
+      print('YES');
```
jonasfj: haha, that's some copy/pasta :)

Maybe, just print warning if oauthUserId isn't set, and always ask for confirmation instead of `await Future.delayed(Duration(seconds: 5));`.

This is a fairly destructive act, we should ask for confirmation...

PR #2325, opened 2019-05-27
 end2end pub get and upload verification with fake_pub_server

isoos:
- #2089
- `retry` package is faked if it doesn't exists (in case a local fake_pub_server is running)
- `_dummy_pkg` is always with a random build version (and increments minor version if exists)
- tests `pub get` and `pub publish` (from / to the local in-memory storage server)
- tests the package page for the build version to be present
- code is structured in a way that is should be useful in staging and prod too (but not tested against them yet)


app/lib/frontend/backend.dart
```diff
@@ -1070,6 +1072,9 @@ class TarballStorage {
 ///   gs://<bucket-name>/packages/<package-name>-<version>.tar.gz
 ///   gs://<bucket-name>/ns/<namespace>/packages/<package-name>-<version>.tar.gz
 class TarballStorageNamer {
+  /// The tarball object storage prefix
+  final String storagePrefix;
```
jonasfj: Can we call this `storageBaseUrl`

app/lib/shared/redis_cache.dart
```diff
@@ -30,16 +30,21 @@ void _registerCache(Cache<List<int>> cache) => ss.register(#_cache, cache);
...
+    return await withCache(fn);
   });
 }
 
+// Run [fn] with a redis or an in-memory cache.
+Future withCache(FutureOr Function() fn) async {
+  // Use in-memory cache, if not running on AppEngine
+  if (Platform.environment.containsKey('GAE_VERSION')) {
+    return await _withRedisCache(fn);
+  } else {
```
jonasfj: Drop the `else`-block as you're returning from the `if`-block.

pkg/pub_integration/lib/src/dummy_pkg.dart
```diff
@@ -0,0 +1,104 @@
...
+
+/// The content of example/bin/main.dart
+String exampleMain() => '''// example
+import 'package:_dummy_pkg/_dummy_pkg.dart';
+
+void main() => sayHello();
+''';
+
+/// Creates the files of the dummy package in [dir] with [version].
+Future createDummyPkg(String dir, String version) async {
```
jonasfj: Can we please not do this? Let's store the files in a `test/.../testdata/` folder. Then this just copies the folder recursively with all files... instead.

Issue #1418, opened 2018-06-28
Sanitize output HTML


isoos:
- [x] sanitize user-provided content
  - [x] readme
  - [x] changelog
  - [x] example
  - [x] license
- [x] sanitize pub-generated content
  - [x] dartdoc html
  - [x] pana suggestions (which may contain references to the user-provided content)

isoos:
I'm moving this to "On Deck" for now, as this turns out to be a longer task than it first seemed. The issues I've uncovered so far:

There are two packages that I can start working with: `package:xml` and `package:html`.

`package:xml` has `XmlPrettyWriter`, which should be able to format the XML output nicely, however, it screwed up the formatting around `<pre>` and `<code>` blocks. It would require either hooks or even more post-processing to compensate for the template glitches.

`package:html` has no indent-based output, the only way to extract the result is to call `doc.outerHtml`. While I can remove the whitespaces and add the appropriate indents to the Elements, it also removes closing tag in several cases (e.g. for meta Elements in the `<head>` section. In the past we had issues with those, and switched to stricter templates.

In either case, the indenting requires more work than the straightforward parse-and-replace, and moving this to a later target date.

isoos:
I think `package:sanitize_html` would be good for us after https://github.com/google/dart-neats/pull/3 gets published. As we already have our markdown processing centralized, most of the above items can be handled with a single change.

However, the dartdoc-generated content may be a bit harder: we'll need to load and update the HTML files, or update `package:dartdoc` to also include the `sanitize_html` as a step in their processing. I'll examine the later alongside with the debugging of the library-inclusion-bug.

PR #2328, opened 2019-05-28
Sanitize HTML and enable more GitHub markdown.

isoos:
#1418

### DIFF
--- a/app/lib/shared/markdown.dart
+++ b/app/lib/shared/markdown.dart
@@ -38,7 +38,10 @@ String markdownToHtml(String text, String baseUrl, {String baseDir}) {
   final html = sanitizeHtml(
     m.renderToHtml(nodes),
     allowElementId: (String id) => true, // TODO: blacklist ids used by pub site
-    allowClassName: (String cn) => _whitelistedClassNames.contains(cn),
+    allowClassName: (String cn) {
+      if (cn.startsWith('language-')) return true;
+      return _whitelistedClassNames.contains(cn);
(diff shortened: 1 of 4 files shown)


app/lib/shared/markdown.dart
```diff
@@ -6,28 +6,22 @@ import 'package:logging/logging.dart';
...
-void _initInlineSyntaxes() {
-  if (_inlineSyntaxes != null) return;
-  _inlineSyntaxes = m.ExtensionSet.gitHubWeb.inlineSyntaxes
-      .where((s) => s is! m.InlineHtmlSyntax)
-      .toList();
-  _inlineSyntaxes.add(_InlineBrHtmlSyntax());
-}
+const _whitelistedClassNames = <String>[
+  'hash-header',
+  'hash-link',
```
jonasfj: How about `language-dart` and `language-yaml`.... maybe even `language-<whatever>` to allow for syntax highlighting?

isoos: Good catch! Coincidentally (as we were talking about tests), this was caught by the templates_test.dart too :)

kevmoo: FYI – could be a Set instead of a list 😄

isoos: ah, sorry, missed this comment... not sure though if it will be faster with only two items in it...

kevmoo: 🤷‍♂

PR #2331, opened 2019-05-29
Move some of the redirects to routes.dart

isoos:


isoos:
Re: usage statistics:
- `/search` and `/server` is used occasionally, but I think it is the same use as e.g. `/doc`: people type it by accident. The domain is pub.dev, not pub.dartlang.org, I don't think they arrive via a link.
- `/server/packages` is not used
- `/flutter/plugins` is used via pub.dartlang.org domain

Should I remove everything except for `/flutter/plugins`?

PR #2333, opened 2019-05-30
Code coverage reporting tool

isoos:
- only the unit tests are included at the moment, the integration tests seems to require a mechanism to finish the program without calling `exit` or `SIGKILL` signal.
- I've created a tool that scans the unit tests and creates a single test suite that includes all of the tests. This tool may be published as a separate package later.
- As they are now executed in a single isolate, some of the tests needed adjustments, e.g. static files-related test update the global state.
- I've also create a tool to create a formatted tree report. The current coverage for `app/lib` is about 50%.

### DIFF
--- a/app/test/frontend/static_files_test.dart
+++ b/app/test/frontend/static_files_test.dart
@@ -12,8 +12,9 @@ import 'package:test/test.dart';
 
 import 'package:pub_dartlang_org/frontend/static_files.dart';
 
-Future main() async {
-  await updateLocalBuiltFiles();
+void main() {
+  test('init', () => updateLocalBuiltFiles(),
+      timeout: Timeout(Duration(minutes: 10)));
 
   group('dartdoc assets', () {
     Future checkAsset(String url, String path) async {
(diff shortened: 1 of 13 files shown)


app/test/frontend/static_files_test.dart
```diff
@@ -12,8 +12,9 @@ import 'package:test/test.dart';
 
 import 'package:pub_dartlang_org/frontend/static_files.dart';
 
-Future main() async {
-  await updateLocalBuiltFiles();
+void main() {
+  test('init', () => updateLocalBuiltFiles(),
```
jonasfj: Is this not supposed to be using a `setUp`, see: https://pub.dev/documentation/test_api/latest/test_api/setUp.html

pkg/code_coverage/pubspec.yaml
```diff
@@ -0,0 +1,14 @@
+name: code_coverage
```
jonasfj: Please give this package a `README.md` explaining what this does... and how to invoke it :)

app/test/frontend/static_files_test.dart
isoos: yeah... but there is no timeout for that. Not sure if this needs 10 minutes, I just remembered that in one of the projects I had the init time was more then the setup timeout... maybe I was remembering wrong and this was not that...

PR #2334, opened 2019-05-31
WIP: login

isoos:


pkg/web_app/lib/src/account.dart
```diff
@@ -0,0 +1,77 @@
...
+
+import 'package:googleapis_auth/auth_browser.dart';
+
+AutoRefreshingAuthClient client;
+
+String _clientId;
+Element _loginElem;
+Element _accountInfoElem;
+
+Future setupAccount() async {
```
sigurdm: Fly-by nit: `Future<void>`

pkg/web_app/pubspec.yaml
```diff
@@ -3,3 +3,6 @@ publish_to: none
 
 environment:
   sdk: '>=2.1.0 <3.0.0'
+
+dependencies:
+  googleapis_auth: ^0.2.7
```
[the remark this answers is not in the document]
jonasfj: hmm, reading:
https://developers.google.com/identity/sign-in/web/reference#gapiauth2signinoptions

It says:
> As gapi.auth2.init will automatically sign in a user to the application if previously signed in, calling signIn({prompt: 'none'}) will usually fail.

From:
https://developers.google.com/identity/sign-in/web/reference#gapiauth2initparams

>When you initialize the GoogleAuth object, you configure the object with your OAuth 2.0 client ID and any additional options you want to specify. Then, if the user has already signed in, the GoogleAuth object restores the user's sign-in state from the previous session.

So it seems, if we do it the regular way, we might simply get everything setup for us, and the user session restored when we initialize the gapi auth library..

isoos:
PTAL. At this stage this is a PoC to demonstrate that the login / logout is working. Proper styling and UI widget handling will be updated separately.

isoos:
Deployed to staging.

isoos:
I have no idea why CI fails... should I re-create the PR?

jonasfj:
I restarted the travis build..

pkg/web_app/lib/src/account.dart
```diff
@@ -0,0 +1,132 @@
...
+        AuthenticatedClient(_currentUser.getAuthResponse(true)?.access_token);
+    _updateOnCredChange();
+  }
+  _updateVisibility();
+  print('updated user.');
+}
+
+Future<void> setupAccount() async {
+  context['pubAuthInit'] = () {
+    if (getAuthInstance().isSignedIn.get()) {
```
jonasfj: Don't you still have to do:
```
  gapi.load('auth2', function() { // Ready. });
```
See: https://developers.google.com/identity/sign-in/web/reference

pkg/web_app/lib/src/google_auth_js.dart
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+@JS('gapi.auth2')
```
jonasfj: Please leave a documentation comment at the library level, saying this is partial bindings for:
https://developers.google.com/identity/sign-in/web/reference

app/lib/frontend/handlers/routes.dart
```diff
@@ -228,6 +229,9 @@ class PubSiteService {
   // **** Custom API
   // ****
 
+  @Route.get('/api/account/v1/info')
```
jonasfj: Let's remove the version numbers from the URL...

isoos:
I'm giving up on fixing the CI, will re-create the PR with fixing the current comments.

pkg/web_app/lib/src/account.dart
isoos: This is a callback that will be executed once the script loads and auth is initialized. (it is triggered via the request param in the script)

isoos:
Continued work in #2344

pkg/web_app/lib/src/account.dart
jonasfj: This is equivalent of the `init` function specified in:
```
<script src="https://apis.google.com/js/platform.js?onload=init" async defer></script>
```
From the docs... where do you get that you don't need to call `gapi.load('auth2',` ?

isoos: This is exactly what we are doing, I just named it `pubAuthInit` instead of `init`.

`gapi.load(...)`: not sure, but I think there was a half-sentence about the `onload` callback will be called after the gapi is loaded/initialized.

jonasfj: It's very possible that `gapi` is loaded, but we also have to load the `auth2` library.

I suspect it's working because our browsers have all this stuff cached...
Regardless of why it works without this step, we should include it as to follow documentation so that it keeps working in the future :)

isoos: done (in the other PR)

PR #2335, opened 2019-05-31
Shutdown fake_pub-server on SIGHUP

isoos:
This is a prerequisite for the code coverage running the `pub_integration` tests.

### DIFF
--- a/pkg/fake_pub_server/lib/fake_storage_server.dart
+++ b/pkg/fake_pub_server/lib/fake_storage_server.dart
@@ -1,4 +1,6 @@
+import 'dart:async';
 import 'dart:convert';
+import 'dart:io';
 
 import 'package:buffer/buffer.dart';
 import 'package:logging/logging.dart';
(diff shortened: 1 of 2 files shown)


pkg/fake_pub_server/lib/fake_storage_server.dart
```diff
@@ -18,7 +20,12 @@ class FakeStorageServer {
     final server = await IOServer.bind('localhost', port);
     serveRequests(server.server, _handler);
     _logger.info('Storage server running on port $port');
-    await Future.delayed(Duration(days: 10000));
+    StreamSubscription sigsubs;
+    sigsubs = ProcessSignal.sighup.watch().listen((_) async {
+      await sigsubs.cancel();
+      await server.close();
+    });
+    await sigsubs.asFuture();
```
jonasfj: Why not just:
```dart
await ProcessSignal.sighub.watch().first;
await server.close();
```

Also why not use `sigterm`? (I think it's the idiomatic one to use for terminating a process gracefully).

Issue #2336, opened 2019-06-02
Pub is not feeling well. Fatal package site error.


komapeb:
You asked me for this issue, here are the details:

```
Requested URL: https://pub.dev/packages?q=data+connection&page=3
Request ID: ffb554334f52295caca5b70914b68506
```

jonasfj:
Can you reference where we asked for this... I think I might have forgotten the context..

Sorry, we're are more than one person and we juggle at lot of issues.

What is the problem, are you getting an error message?

isoos:
I believe it has something to do with the google cloud outage that happened roughly 2 hours ago.

komapeb:
This is on the pub.dev site, exactly on the requested URL provided in the details of the first comment.

isoos:
Actually, I think this is not caused by the gcloud network issues, rather our `search` service restarting. The failing request was hitting the `search` service during the bootup process and it responded with status code `600` (a custom code indicating that it was not ready yet).

jonasfj:
@isoos, so that sounds like we've 3 possible problems:
1) the `search` service returns 200 ok to health checks before it's ready.
2) the health check configuration for the `search` service is wrong,
3) the `search` service is crashing and causing restarts.

isoos:
Looking at the logs in depth, and it seems that when the `search` service was starting up, it failed to load the snapshot (and the popularity data as well), and it started to populate search via normal Datastore queries.

- The snapshot failure happened for multiple service instances between `2019-06-02 20:53:36.385 CEST` and `2019-06-02 20:58:23.088 CEST`.
- Based on the logs this self-building process completed at `2019-06-02 22:41:30.015 CEST`, almost 2 hours later.
- In the meantime another instance restarted and was able to merge the snapshot and was up and running at `2019-06-02 22:19:48.489`.

We have a retry in place that would try another request if the search index was not ready, however, in such prolonged case there is a good chance that the second request goes to the same uninitialized index. We'll need to refactor the service startup and use not only the health checks, but also the readiness checks to indicate when the load balancer can route requests towards the service.

jonasfj:
Yes, "readiness checks" (sorry, I confused that with health checks).

isoos:
Yeah, correct, we'll need to implement `readiness_check` (and split the health checks for it).

isoos:
[...]

Issue #2111, opened 2019-03-06
use "updated health" checks


jonasfj:
See: https://cloud.google.com/appengine/docs/flexible/custom-runtimes/configuring-your-app-with-app-yaml#health_checks

I propose we add the following to `app|analyzer|dartdoc|search.yaml`
```yaml
liveness_check:
readiness_check:
  app_start_timeout_sec: 600
```

Maybe we can give search longe startup time than the others...

isoos:
In addition to that, we could use readiness checks to:
- wait for initial popularity data fetch (`app` and `search`)
- wait for search snapshot (`search`)

jonasfj:
Yeah, if they are reasonably fast to do... I suspect/hope we can just avoid listening for TCP connections until we are ready...

PR #2337, opened 2019-06-03
Harden search service.

isoos:
- #2336 
- increased number of instances to reduce the chance of all of them rebooted
- switching to split health checks (instruction in CHANGELOG)
- index updates are tied to the successful initialization
- initialization is done either via snapshot, or a minimal package index

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ AppEngine version, listed here to ease deployment and troubleshooting.
 
 ## Next Release (replace with git tag when deployed)
 
+ * Update project to use split health checks, run:
+   `gcloud app update --split-health-checks --project dartlang-pub`
+ * `search` service is using custom liveness and readiness checks.
 
 ## `20190529t163905-all`
 
--- a/app/bin/service/search.dart
+++ b/app/bin/service/search.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:io';
 import 'dart:isolate';
 
 import 'package:_discoveryapis_commons/_discoveryapis_commons.dart'
@@ -100,6 +101,7 @@ Future _main(FrontendEntryMessage message) async {
         await batchIndexUpdater.init();
       } catch (e, st) {
         _logger.shout('Error initializing search service.', e, st);
+        exit(1);
       }
 
       final scheduler = TaskScheduler(
--- a/app/bin/service/search.dart
+++ b/app/bin/service/search.dart
@@ -94,27 +94,33 @@ Future _main(FrontendEntryMessage message) async {
     _updateDartSdkIndex().whenComplete(() {});
 
     final BatchIndexUpdater batchIndexUpdater = BatchIndexUpdater();
-    await batchIndexUpdater.initSnapshot();
+    // Don't block on init, we need to serve liveliness and readiness checks.
+    batchIndexUpdater.init().then(
+      (_) {
+        final scheduler = TaskScheduler(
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ AppEngine version, listed here to ease deployment and troubleshooting.
 
 ## Next Release (replace with git tag when deployed)
 
+ * Update project to use split health checks, run:
+   `gcloud app update --split-health-checks --project dartlang-pub`
```
jonasfj: Won't we need to also update the other services to use the new health checks?

app/bin/service/search.dart
```diff
@@ -94,27 +94,33 @@ Future _main(FrontendEntryMessage message) async {
...
-        DatastoreHeadTaskSource(
-          db.dbService,
-          TaskSourceModel.scorecard,
-          sleep: const Duration(minutes: 10),
-          skipHistory: true,
-        ),
-        batchIndexUpdater.periodicUpdateTaskSource,
-      ],
+    // Don't block on init, we need to serve liveliness and readiness checks.
+    batchIndexUpdater.init().then(
```
jonasfj: How about we do:
```dart
scheduleMicrotask(() async {
  await batchIndexUpdater.init();

  final scheduler = ...
  ....
});
```

Just to avoid implicit micro tasks with `.then`...

app/bin/service/search.dart
```diff
@@ -94,27 +94,33 @@ Future _main(FrontendEntryMessage message) async {
...
+          ],
+        );
+        scheduler.run();
+
+        Timer.periodic(const Duration(minutes: 5), (_) {
+          updateLatestStats(scheduler.stats());
+        });
+      },
+      onError: (e, StackTrace st) {
+        _logger.shout('Error initializing search service.', e, st);
```
jonasfj: If this fails, then what?

Don't we need to crash the service? Exit non-zero, or something? We could have retries, but assuming all the fragile I/O is already wrapped in retry statements, maybe it's best to just give up at this point.

Otherwise, the service will keep being alive. And have a liveness check that returns OK.

app/lib/search/updater.dart
```diff
@@ -37,6 +37,26 @@ class BatchIndexUpdater implements TaskRunner {
...
+      _logger.info('Loading minimum package index...');
+      int cnt = 0;
+      await for (final pd in searchBackend.loadMinimumPackageIndex()) {
+        await packageIndex.addPackage(pd);
+        cnt++;
+        if (cnt % 500 == 0) {
+          _logger.info('Loaded $cnt minimum package data (${pd.package})');
+        }
+      }
+      await packageIndex.merge();
```
jonasfj: As dig into what `merge()` and `isReady` means... Maybe these are things we should document?

I know that sort of outside the scope of this PR..

CHANGELOG.md
isoos: I've tested that on staging, and they seem to be fine, if the split checks are not defined, appengine  seems to use the default health checks.

app/bin/service/search.dart
isoos: We could probably just call exit...

app/lib/search/updater.dart
isoos: These should be refactored (on my list, but very low priority), as the merge is not really a merge with the current library...

CHANGELOG.md
jonasfj: Okay, that's fine we can finish https://github.com/dart-lang/pub-dev/issues/2111 later :)

isoos: ah, ack

PR #2338, opened 2019-06-05
Code coverage to include pub_integration's fake_pub_server test

isoos:
Combined coverage is about 54% for `app/`.


pkg/pub_integration/lib/src/fake_pub_server_process.dart
```diff
@@ -20,8 +20,13 @@ class FakePubServerProcess {
...
+  static Future<FakePubServerProcess> start({
+    String pkgDir,
+    int port,
+    int storagePort,
+    bool collectCoverage = false,
+    int vmServicePort = 29999,
+  }) async {
     pkgDir ??= p.join(Directory.current.path, '../fake_pub_server');
     // TODO: check for free port
     port ??= 20000 + _random.nextInt(990);
```
jonasfj: Unrelated to this PR, but usually you just use port zero and it picks a random ephemeral port.. Not sure if it always picks an idle port.

pkg/pub_integration/lib/src/fake_pub_server_process.dart
```diff
@@ -34,6 +39,11 @@ class FakePubServerProcess {
     final process = await Process.start(
       'dart',
       [
+        if (collectCoverage) ...[
+          '--pause-isolates-on-exit',
+          '--enable-vm-service=$vmServicePort',
+          '--disable-service-auth-codes',
```
jonasfj: This feels like a hack.. we'll have a hard time ripping this out... maybe we can add a line of documentation :)

Or we could pass an environment variable `PUB_DART_ARGUMENTS`, that way the `pkg/integration_test` don't need to know anything about coverage collection...

pkg/pub_integration/lib/src/fake_pub_server_process.dart
isoos: but then we'd need to parse it to access it from the client. Shall I just use a fixed port?

pkg/pub_integration/lib/src/fake_pub_server_process.dart
isoos: Yeah, that could work out better...

pkg/pub_integration/lib/src/fake_pub_server_process.dart
jonasfj: You'd get it back after listening, it's just a nit...

PR #2342, opened 2019-06-06
Separate pub.dartlang.org -only redirects.

isoos:


app/lib/frontend/handlers.dart
```diff
@@ -34,6 +36,14 @@ Future<shelf.Response> appHandler(
 ) async {
   _logPubHeaders(request);
 
+  // do pub.dartlang.org-only routes
+  if (request.requestedUri.host == legacyHost) {
+    final rs = await PubDartlangOrgService().router.handler(request);
```
jonasfj: We should not be creating a `router` for every request... That's expensive...

jonasfj: I just saw we have similar problem with `PubSiteService(shelfPubApi).router.handler`

app/lib/frontend/handlers/routes.dart
```diff
@@ -283,3 +266,37 @@ class PubSiteService {
   @Route.get('/packages.json')
   Future<Response> packagesJson(Request request) => packagesHandler(request);
 }
+
+/// Routes that are only processed by the old pub domain.
+class PubDartlangOrgService {
+  Router get router => _$PubDartlangOrgServiceRouter(this);
```
jonasfj: I think this can be cheaper if we do:
```dart
final Router router = _$PubDartlangOrgServiceRouter(this);
```
Maybe this works? I'm not sure, otherwise ignore it as we should be caching the entire object regardless...

app/lib/frontend/handlers/routes.dart
```diff
@@ -283,3 +266,37 @@ class PubSiteService {
   @Route.get('/packages.json')
   Future<Response> packagesJson(Request request) => packagesHandler(request);
 }
+
+/// Routes that are only processed by the old pub domain.
```
jonasfj: Can we put this in a separate file...

app/lib/frontend/handlers.dart
jonasfj: How about the first handler in `PubSiteService` be something like:
```dart
class PubSiteService {

  final _pubDartlangOrgHandler = PubDartlangOrgService().router.handler;

  @Route.all('/<path|.*>')
  Future<Response> _handleLegacyHost(Request request) {
    if (request.uri.host != legacyHost) {
      return null;
    }
    return _pubDartlangOrgHandler(request);
  }
}
```

jonasfj: This is just illustrative code... Maybe we don't have host available?

app/lib/frontend/handlers/routes.dart
isoos: I'd like to keep it in the same file, I think it is really nice to have a single file for all the routes we have.

app/lib/frontend/handlers.dart
isoos: I'd do this in a follow-up, to include other host redirects (e.g. dartdocs.org)

app/lib/frontend/handlers/routes.dart
isoos: Updated code to initialize routes only once.

isoos:
PTAL.

app/lib/frontend/handlers.dart
```diff
@@ -28,21 +28,28 @@ void _logPubHeaders(shelf.Request request) {
...
+
+    // do pub.dartlang.org-only routes
+    if (request.requestedUri.host == legacyHost) {
+      final rs = await pubDartlangOrgHandler(request);
+      if (rs != null) {
+        return rs;
+      }
+    }
+
+    final redirected = tryHandleRedirects(request);
```
jonasfj: Why don't we move `PubDartlangOrgService` into `redirects.dart` and handle those redirects in `tryHandleRedirects`?

isoos: We can rearrange that after the other domain-specific redirects are migrated, but most of all I really like the fact that all routes are in a single file. I think it is better than having two or more files based on the domain name...

Issue #2316, opened 2019-05-23
Support sign-in with Google on pub.dev


jonasfj:
We should use a client-side login flow, as this allows caching of the content.

PR #2344, opened 2019-06-07
Login / Logout PoC

isoos:
#2316


pkg/web_app/lib/src/account.dart
```diff
@@ -0,0 +1,106 @@
...
+  return _client ??=
+      getAuthenticatedClient(currentUser.getAuthResponse(true)?.access_token);
+}
+
+Future<void> setupAccount() async {
+  // Initialization hook that will run after the auth library is loaded and
+  // initialized. Method name is passed in as request parameter when loading
+  // the auth library.
+  context['pubAuthInit'] = () {
+    _initialized = true;
```
jonasfj: According to docs: https://developers.google.com/identity/sign-in/web/reference
You need to call `gapi.load('auth2', () {....});`

At-least that's how I read it, see the "Auth setup" section.

app/lib/frontend/templates/views/layout.mustache
```diff
@@ -37,9 +37,18 @@
   {{#include_survey}}
   <script async="async" defer="defer" src="//www.google.com/insights/consumersurveys/async_survey?site=5wdvu4vc5lhop3xbatxof6lzcm"></script>
   {{/include_survey}}
+  {{#oauth_client_id}}
+  <meta name="google-signin-client_id" content="{{oauth_client_id}}" />
+  <script src="https://apis.google.com/js/platform.js?onload=pubAuthInit" defer="defer"></script>
```
jonasfj: Why not `async`?

isoos: `async` makes the execution order unpredictable, and this way the `pubAuthInit` will be called after the callback is populated by our script, and the google auth is loaded and initialized.

pkg/web_app/lib/src/account.dart
isoos: True... I guess we shall call that, although it is working without it :)

jonasfj: Maybe it works because it's cached or something? Or maybe it just always works. I would just prefer to rely on documented behavior :)

app/lib/frontend/templates/views/layout.mustache
jonasfj: Okay, that's fine...

isoos:
I think the experimental flag is good-enough to not expose this to wide audience before it gets ready.

PR #2347, opened 2019-06-07
 Extend integration test with uploader add/remove.

isoos:
#2089

### DIFF
--- a/pkg/pub_integration/test/fake_pub_server_test.dart
+++ b/pkg/pub_integration/test/fake_pub_server_test.dart
@@ -5,6 +5,7 @@
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:html/parser.dart' as html_parser;
 import 'package:http/http.dart' as http;
 import 'package:path/path.dart' as p;
 import 'package:test/test.dart';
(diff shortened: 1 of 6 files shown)


pkg/pub_integration/test/fake_pub_server_test.dart
```diff
@@ -37,12 +39,38 @@ void main() {
...
+        final pageUri = Uri.parse(pageUrl);
+        final confirmUrl = pageUri.replace(
+          scheme: 'http',
+          host: 'localhost',
+          port: fakePubServerProcess.port,
+        );
+        print('Requesting confirm URL $confirmUrl');
+        final pageRs = await httpClient.get(confirmUrl);
+        final acceptLine = pageRs.body.split('\n').firstWhere(
+            (line) => line.contains('Authorize and Accept Invite</a>'));
```
jonasfj: Could we make this less fragile by parsing the HTML? Or maybe just inserting an id... I imagine that this text might change over time.

pkg/pub_integration/test/fake_pub_server_test.dart
```diff
@@ -37,12 +39,38 @@ void main() {
     tearDownAll(() async {
       await tempDir?.delete(recursive: true);
       await fakePubServerProcess?.kill();
+      httpClient.close();
     });
 
     test('standard integration', () async {
+      final inviteUrlFuture = fakePubServerProcess
+          .waitForLine((line) => line.contains('/admin/confirm/new-uploader/'));
```
jonasfj: You're fake email provider writes emails to stdout, right?

isoos: yes, that's correct

Issue #2351, opened 2019-06-11
img/background-pattern-darkblue.jpg is no cached


jonasfj:
We should set a cache policy on this image...

From a quick look I'm not sure we do have one...

Also we should set one for the `favicon.ico`, while we're at it :)

PR #2353, opened 2019-06-12
Using hash in CSS files.

isoos:
- #2351
- the more important part is the test that will make sure we do use hashes in the CSS files

### DIFF
--- a/app/test/frontend/static_files_test.dart
+++ b/app/test/frontend/static_files_test.dart
@@ -8,13 +8,13 @@ import 'dart:async';
 
 import 'package:crypto/crypto.dart' as crypto;
 import 'package:http/http.dart' as http;
+import 'package:path/path.dart' as p;
 import 'package:test/test.dart';
 
 import 'package:pub_dartlang_org/frontend/static_files.dart';
 
 void main() {
-  test('init', () => updateLocalBuiltFiles(),
-      timeout: Timeout(Duration(minutes: 10)));
+  setUpAll(() => updateLocalBuiltFiles());
 
   group('dartdoc assets', () {
     Future checkAsset(String url, String path) async {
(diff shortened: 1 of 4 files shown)


app/test/frontend/static_files_test.dart
```diff
@@ -77,5 +77,35 @@ void main() {
         expect(f.etag.contains('mocked_hash_'), isFalse);
       });
     }
+
+    test('proper hash in css content', () {
+      final cssHashes = <String, String>{};
+      final css = cache.getFile('/static/css/style.css');
+      for (Match m in RegExp('url\\("(.*?)"\\);')
```
jonasfj: Why the question mark? Why not just look for all urls...

```
RegExp('url\\([\'"](.*)[\'"]\\)')
```
Don't we want all URLs?

app/test/frontend/static_files_test.dart
```diff
@@ -77,5 +77,35 @@ void main() {
...
+          .allMatches(css.contentAsString)) {
+        final matched = m.group(1);
+        if (matched.contains('data:image')) continue;
+        final uri = Uri.parse(matched);
+        final absPath = p.normalize(p.join('/static/css', uri.path));
```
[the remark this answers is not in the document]
jonasfj: You'll want to use `Uri.parse('/static/css/style.css').resolve(matched)`

Example: https://dartpad.dartlang.org/d8041639aeb4430737be055fa14063fa

app/test/frontend/static_files_test.dart
```diff
@@ -77,5 +77,35 @@ void main() {
...
+        final hash = uri.queryParameters['hash'] ?? '_no_hash_';
+        if (cssHashes.containsKey(absPath) && cssHashes[absPath] != hash) {
+          throw Exception(
+              'Multiple hash for key: $absPath ($hash vs ${cssHashes[absPath]})');
+        }
+        cssHashes[absPath] = hash;
+      }
+
+      final expectedHashes = <String, String>{};
+      final containedFiles = [
```
jonasfj: This is a really tightly coupled test... instead of collecting `cssHashes` and comparing that...
How about we just do:
```dart
if (hash != cache.getFile(absPath).etag) {
  throw ....
}
```
Inside the loop above... that way we don't have to touch this test every time we tweak the CSS

PR #2356, opened 2019-06-12
Hash for more assets.

isoos:
- #2351
- More `?hash=` urls, e.g. `/favicon.ico?hash=`. As this is outside of /static/, it required to change a few more things, and in effect it also changed the mocked hash codes in tests.
- Removed the list of hashed urls (it was only used in test), it makes the updates of assets simpler.

### DIFF
--- a/app/lib/frontend/static_files.dart
+++ b/app/lib/frontend/static_files.dart
@@ -164,6 +164,12 @@ class StaticUrls {
   }
 
   /// A hashed version of the static assets.
+  ///
+  /// For each file like /static/img/logo.gif we create a key and value of:
+  /// `img__logo_gif => /static/img/logo.gif?hash=etag_hash`
+  ///
+  /// The /static/ prefix is stripped from the start of the key in order to
+  /// reduce the length of the mostly used files in the mustache templates.
   Map<String, String> get assets {
     if (_assets == null) {
       _assets = <String, String>{};
--- a/app/lib/frontend/static_files.dart
+++ b/app/lib/frontend/static_files.dart
@@ -21,7 +21,7 @@ StaticUrls _staticUrls;
 StaticFileCache get staticFileCache =>
     _cache ??= StaticFileCache.withDefaults();
 
-StaticUrls get staticUrls => _staticUrls ??= StaticUrls();
+StaticUrls get staticUrls => _staticUrls ??= StaticUrls._();
 
 /// Register the static file cache.
 void registerStaticFileCacheForTest(StaticFileCache cache) {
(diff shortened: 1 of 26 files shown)


app/lib/frontend/static_files.dart
```diff
@@ -172,30 +158,32 @@ class StaticUrls {
     };
   }
 
-  /// A hashed version of the static assets referenced in [hashedFiles].
+  /// A hashed version of the static assets.
```
jonasfj: Is this a mapping from something to something? Could we document what it maps from and to?

app/lib/frontend/static_files.dart
```diff
@@ -172,30 +158,32 @@ class StaticUrls {
...
   Map<String, String> get assets {
     if (_assets == null) {
       _assets = <String, String>{};
-      for (String hashedFile in hashedFiles) {
+      for (String requestPath in staticFileCache.keys) {
+        final inStatic = requestPath.startsWith('$_defaultStaticPath/');
+        // Removing the /static/ prefix from the keys in order to make them
+        // shorter in the templates.
+        final hashedFile = inStatic
+            ? requestPath.substring(_defaultStaticPath.length + 1)
```
jonasfj: Why can you remove `/static/`? are these always being used as relative?

app/lib/frontend/static_files.dart
```diff
@@ -124,46 +127,29 @@ class StaticFile {
...
   final String smallDartFavicon;
   final String dartLogoSvg;
   final String flutterLogo32x32;
   final String documentationIcon;
   final String downloadIcon;
+  final String pubDevLogo2xPng;
   Map _versionsTableIcons;
   Map<String, String> _assets;
 
   StaticUrls()
```
jonasfj: Why does this have a public constructor?

app/lib/frontend/static_files.dart
```diff
@@ -124,46 +127,29 @@ class StaticFile {
...
-  'img/dart-packages.png',
-  'img/dart-packages-white.png',
-  'img/flutter-packages.png',
-  'img/flutter-packages-white.png',
-  'img/pub-dev-logo-2x.png',
-  'js/gtag.js',
-  'js/script.dart.js',
-];
-
 class StaticUrls {
```
jonasfj: Is this simply exposing the static urls being hosted by a `StaticFileCache`? If so, why doesn't the constructor take a specific `StaticFileCache` instance? I'm guessing their are all singletons... Maybe we could hide the constructors? At mark them visibleForTestsOnly...

app/lib/frontend/static_files.dart
isoos: https://github.com/dart-lang/pub-dev/blob/master/app/lib/frontend/templates/views/layout.mustache#L34-L36

```{{{static_assets.js__script_dart_js}}}``
vs
```{{{static_assets.static__js__script_dart_js}}}```

jonasfj: Please illustrate that in the doc comment... :)

it makes a lot of sense though :)

Issue #2349, opened 2019-06-09
Search result is empty about flame game engine


chengenzhao:
URL: https://pub.dev/flutter/packages?q=flame

Hi:

We all love a minimal game engine called Flame
and when I tried to search pub.dev
sadly the result didn't show up
and I pretty sure I could get the correct search result before
so any idea why it doesn't show up?

search url: https://pub.dev/flutter/packages?q=flame
repo url: https://pub.dev/packages/flame

isoos:
> so any idea why it doesn't show up?

Yeah, the core reason is that the latest version depends on Flutter version that we don't use for analysis yet. (We only support stable releases, [see this link](https://flutter.dev/docs/development/tools/sdk/releases?tab=linux)). Because of that, we were not able to analyze the package, and it didn't received (a) platform classification nor (b) good package scores.

Because there is no platform classification, the package is not listed under /flutter/. You can still find it under "All" packages search: https://pub.dev/packages?q=flame (7th item at the time of writing this).

@jonasfj: maybe we could surface packages with matching names even if it doesn't have platform match?

jonasfj:
If we have an exact name match, then yeah, it seems fair to ignore the filters and scores and show said package at the very top.

PR #2357, opened 2019-06-12
Include exact name match even if it is not in the filtered result set.

isoos:
#2349


app/lib/search/index_simple.dart
```diff
@@ -245,14 +245,12 @@ class SimplePackageIndex implements PackageIndex {
         final matchingPackage = queryText == null
```
jonasfj: you'll need to update the comment above... to reflect the current state..

jonasfj: This comment:
> // If there is an exact match for a package name (in the filtered result
        // set), promote that package to the top position.

PR #2362, opened 2019-06-13
Integration test script running against staging or prod.

isoos:


pkg/pub_integration/bin/pub_integration.dart
```diff
@@ -0,0 +1,105 @@
...
+
+  final invitedEmail = argv['invited-email'] as String;
+  if (invitedEmail == null || invitedEmail.isEmpty) {
+    print('--invited-email must be set');
+    exit(1);
+  }
+
+  String credentialsFile = argv['credentials-json'] as String;
+  if (credentialsFile == null && Platform.environment['HOME'] != null) {
+    credentialsFile =
```
jonasfj: Also search `PUB_CACHE` env var... if that's define we could use that..

pkg/pub_integration/bin/pub_integration.dart
```diff
@@ -0,0 +1,105 @@
...
+import 'package:args/args.dart';
+import 'package:path/path.dart' as p;
+
+import 'package:pub_integration/pub_integration.dart';
+
+/// Runs integration test against prod or staging server.
+Future main(List<String> args) async {
+  final _argParser = ArgParser()
+    ..addOption('pub-hosted-url', help: 'The PUB_HOSTED_URL to use.')
+    ..addOption('verifier-email', help: 'The e-mail of the verifier account.')
```
jonasfj: What is this? Is the verifier the person running the verification...
Can we call this "uploader-email" instead?

Or maybe we drop the check here:
https://github.com/dart-lang/pub-dev/blob/92b878169af907013662562aa59953b5a30eaf43/pkg/pub_integration/lib/pub_integration.dart#L151-L153

Since it offers very little value... presumably the user we're testing with is already an uploader, if not a different part of the test will fail. And we are already testing that uploaders are listed on the page.

jonasfj: Also can the help message say that this must be an "uploader" for the `_dummy_pkg` package.
(I assume that's the package we're always uploading)

isoos: Right. I'll remove it for now. If we want to use it for some reason later, we shall be able to call the new `/api/account/info` endpoint to get the e-mail address, and we won't need the CLI flag anyway.

PR #2366, opened 2019-06-14
Bump dart to 2.3.2

jonasfj:
Let's bump to 2.3.2 there is a few nice fixes in `dart:io`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,58 +8,58 @@ services:
 jobs:
   include:
     - stage: smoke_test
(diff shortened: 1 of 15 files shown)


isoos:
I believe you need to change all the mono-repo config files in the sub-packages and then run the mono-repo travis command to regenerate travis.

jonasfj:
haha, I only updated what the tests checked, so guilty :) hehe

jonasfj:
@isoos, done, PTAL

PR #2369, opened 2019-06-17
Retry most storage Bucket operations.

isoos:
- Implemented with a Bucket wrapper, and exposed through the getOrCreateBucket which most of our code uses. 
- Retry only if the storage service responds with status code 500.


jonasfj:
Should we do this in the `gcloud` package instead, see https://github.com/dart-lang/gcloud/issues/70

That way we won't pollute our code with unnecessary wrappers..

isoos:
If we are sure that the retry rules (and retry options) are universal, we can move this to the `gcloud` package.

PR #2374, opened 2019-06-17
Refactor redirect routes

isoos:
- moved domain selection to `handlers.dart`
- moved most redirect route config to `redirects.dart`

### DIFF
--- a/app/lib/frontend/handlers.dart
+++ b/app/lib/frontend/handlers.dart
@@ -7,6 +7,7 @@ library pub_dartlang_org.handlers;
 import 'package:logging/logging.dart';
 import 'package:shelf/shelf.dart' as shelf;
 
+import '../shared/handlers.dart';
 import '../shared/urls.dart';
 
 import 'handlers/misc.dart';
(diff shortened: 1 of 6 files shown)


app/lib/frontend/handlers.dart
```diff
@@ -29,21 +30,36 @@ void _logPubHeaders(shelf.Request request) {
...
   final pubDartlangOrgHandler = PubDartlangOrgService().router.handler;
   final pubSiteHandler = PubSiteService(shelfPubApi).router.handler;
   return (shelf.Request request) async {
     _logPubHeaders(request);
 
+    // legacy dartdocs.org URLs
+    final host = request.requestedUri.host;
+    if (host == 'www.dartdocs.org' || host == 'dartdocs.org') {
+      final rs = await legacyDartdocHandler(request);
+      return rs ?? shelf.Response.notFound('Not Found.');
```
jonasfj: Let's move `shelf.Response.notFound('Not Found.')` into `LegacyDartdocService`.

Using a catch-all rule at the end of the class..

app/lib/frontend/handlers.dart
```diff
@@ -53,3 +69,11 @@ shelf.Handler createAppHandler(shelf.Handler shelfPubApi) {
     return formattedNotFoundHandler(request);
   };
 }
+
+bool _shouldRedirectToPubDev(String path) {
+  if (path.startsWith('/api/')) return false;
+  if (path.endsWith('.tar.gz')) return false;
+  if (path.endsWith('.json')) return false;
+  if (path == '/debug') return false;
```
jonasfj: Why `/debug` ?

app/lib/frontend/handlers.dart
```diff
@@ -53,3 +69,11 @@ shelf.Handler createAppHandler(shelf.Handler shelfPubApi) {
     return formattedNotFoundHandler(request);
   };
 }
+
+bool _shouldRedirectToPubDev(String path) {
+  if (path.startsWith('/api/')) return false;
+  if (path.endsWith('.tar.gz')) return false;
+  if (path.endsWith('.json')) return false;
```
jonasfj: Why don't we redirect these? Can we maybe leave a comment at the end of each of these lines?

app/lib/frontend/handlers.dart
isoos: good catch, no reason

PR #2375, opened 2019-06-17
Upgrade dependencies in app/

isoos:


jonasfj:
Please add note in the CHANGELOG that we upgraded `appengine` package, a lot of things happened there we should watch for side-effects...

PR #2376, opened 2019-06-17
Output minified styles.css

isoos:


isoos:
> Do we get source maps with this?

The .map file is emitted the same way, and the sources themselves are not served in either way. 

> And if so, don't we need to tweak the static file handling to expose the source maps?

No, you can check it on pub.dev, it identifies the source location: e.g. `_pkg.scss:12`, but clicking on it is not able to load the source. The later was not really a problem in practice.

PR #2377, opened 2019-06-17
Retry storage requests with http_retry

jonasfj:
A quick attempt to setup the storage service with retries.

### DIFF
--- a/app/bin/service/analyzer.dart
+++ b/app/bin/service/analyzer.dart
@@ -25,6 +25,7 @@ import 'package:pub_dartlang_org/shared/scheduler_stats.dart';
 import 'package:pub_dartlang_org/shared/service_utils.dart';
 import 'package:pub_dartlang_org/shared/storage.dart';
 import 'package:pub_dartlang_org/shared/redis_cache.dart';
+import 'package:pub_dartlang_org/shared/storage_retry.dart';
 
 import 'package:pub_dartlang_org/analyzer/handlers.dart';
 import 'package:pub_dartlang_org/analyzer/pana_runner.dart';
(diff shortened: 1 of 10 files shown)


app/bin/service/analyzer.dart
```diff
@@ -55,6 +56,7 @@ Future _frontendMain(FrontendEntryMessage message) async {
   ));
 
   await withAppEngineAndCache(() async {
+    registerStorageWithRetry();
```
isoos: Looks like this is a common pattern. Wouldn't it be better to move the `registerStorageWithRetry();` call inside (and in the beginning of) `withAppEngineAndCache`?

jonasfj: Yes, and no... it would have to be renamed then :)

I'll leave that for later, as we should refactor how all these components are loaded.

app/lib/shared/storage_retry.dart
```diff
@@ -0,0 +1,46 @@
...
+/// Note. this should not be used if uploading files that won't easily fit in
+/// memory! As [RetryClient] will cache the entire request body in memory
+/// in-order to do retries.
+void registerStorageWithRetry() {
+  // Ensure we're we have a client...
+  if (authClientService != null) {
+    throw StateError('gcloud/appengine must be setup');
+  }
+  // Find the projectId.
+  final projectId = Platform.environment['GCLOUD_PROJECT'];
```
isoos: let's use `activeConfiguration.projectId` from `shared/configuration.dart`

jonasfj: ++

isoos:
I think there is a `hint • Unused import: 'dart:io' at lib/shared/storage_retry.dart:1:8 • unused_import`...

Issue #1519, opened 2018-08-10
Support package self-management


mit-mit:
Potential feature idea: Allow uploaders to login to Pub, and from there manage their own packages:

* Mark package as discontinued
* Mark package as 'available for adoption'
* Mark a package version as 'know bad'

mit-mit:
When doing this, consider support for 2FA

isoos:
Admin tab + multiple features have been implemented.

PR #2382, opened 2019-06-19
Self-service: discontinued.

isoos:
- #1519 and #1642 
- `pkg/client_data` to have a single class definition for server and `pkg/web_app`.
  - `PageData` is a generic metadata that the script can read on load. It doesn't have user-specific fields, and can be cached. The extraction is a simple scan of the top-level `script` elements, it follows the ld-json specification (similar to the search api that we already use).
  - `AccountInfo` is a user-specific metadata (optionally requesting package-specific data too). It is not only about the admin/uploader status, but later on it could contain e.g. whether the user had starred the package or not.
- `backend.updateFlag` is something we should call from `bin/tool/flag.dart` but it involves a bit more change about the authenticated user, deferring that into a later PR.
- `/api/packages/<package>/flag/<flag>/<status>` endpoint is a bit awkward, happy to redo it differently :)
- Updating the page rendering to make the `Package.isDiscontinued` flag the priority, in order to immediately render the package as discontinued (maybe we should use only that and not the rest?). 
- The UI change is very simple right now, I'd fine-tune it separately.

Screen Shot 2019-06-19 at 17 20 21


pkg/web_app/lib/src/account.dart
```diff
@@ -147,3 +161,35 @@ class _AccountNavWidget {
     }
   }
 }
+
+class _PkgAdminWidget {
+  Element _root;
+  void init(Element root) {
+    if (!pageData.isPackagePage) return;
+    _root = root;
+    _root.append(elem('h3', classes: ['title'], text: 'Self-admin'));
```
jonasfj: Can we just make this 'Admin'

pkg/web_app/lib/src/account.dart
```diff
@@ -147,3 +161,35 @@ class _AccountNavWidget {
...
+    final discontinueToggle = elem(
+      'a',
```
jonasfj: We need to use a button for actions of this nature. We might even want to do a simple javascript `confirm(....)` for this sort of thing.

isoos: Agreed, but I'd do that in a subsequent PR, big enough already.

app/lib/frontend/backend.dart
```diff
@@ -269,6 +269,35 @@ class Backend {
       }
     }
   }
+
+  /// Updates flags on [package].
```
jonasfj:
```dart
/// Updates flags on [package], as the user of the current request.
```

or is that implicit?

app/lib/frontend/backend.dart
```diff
@@ -269,6 +269,35 @@ class Backend {
...
+  Future updateFlag(String package,
+      {bool isDiscontinued, bool doNotAdvertise}) async {
+    final pkgKey = db.emptyKey.append(models.Package, id: package);
+    await withAuthenticatedUser((user) async {
+      String latestVersion;
+      await db.withTransaction((tx) async {
+        final p = (await tx.lookup<models.Package>([pkgKey])).single;
+        if (p == null) {
+          throw Exception('Package does not exists.');
```
jonasfj: Don't you need to throw an exception that indicates that we can show the message to the user..?

Otherwise, updating this flag for a package that doesn't exist will cause a 5xx, instead of a 4xx.

app/lib/frontend/handlers/account.dart
```diff
@@ -2,13 +2,31 @@
...
 
+import 'package:client_data/account_info.dart';
 import 'package:shelf/shelf.dart' as shelf;
 
 import '../../account/backend.dart';
+import '../../frontend/backend.dart';
 import '../../shared/handlers.dart';
 
 /// Handles /api/account/v1/info
 Future<shelf.Response> accountInfoHandler(shelf.Request request) async {
```
jonasfj: This is not a good REST API...

We don't need to request the email from `pub.dev` we already know that client-side from the login session.

I thought we only had this to prove that we could do authentication... 
but I guess this partially serves to ensure the `User` entity is created on the server?

-----

hmm, IMO the best way to replace this would be to build an end-point that returns the list of packages for which you are an uploader.

I see how that will make publishers harder, and it could grow large for some users...

app/lib/frontend/handlers/routes.dart
```diff
@@ -245,6 +245,11 @@ class PubSiteService {
   Future<Response> apiDocumentation(Request request, String package) =>
       apiDocumentationHandler(request, package);
 
+  @Route.post('/api/packages/<package>/flag/<flag>/<status>')
```
jonasfj: I would suggest we do something like:

```
GET /api/packages/<package>/options
{
  'isDiscontinued': true | false,
}

PUT /api/packages/<package>/options
{
  'isDiscontinued': true | false
}
```

Then we strictly limit accepted input to `isDiscontinued` which must be bool. So that input with any other properties will be rejected. Then we can add more properties in the future.

If we make these nice, we can later declare them stable and perhaps augment the CLI client, etc.

app/lib/frontend/templates/views/layout.mustache
```diff
@@ -148,5 +148,11 @@
 {{& schema_org_searchaction_json}}
 </script>
 {{/schema_org_searchaction_json}}
+
+{{#page_data_json}}
+<script type="application/ld+json">
+{{& page_data_json}}
+</script>
+{{/page_data_json}}
```
jonasfj: Interesting... I'm curious, do you imagine we'll ship a lot of data this way?

How about just using data properties, `<div data-package-name='...' data-package-version='...'>` on some element we already have...

app/lib/frontend/backend.dart
isoos: Would this be a custom exception that we use for emitting 4xx with a custom message?

app/lib/frontend/handlers/account.dart
isoos: I agree about the e-mail part, kept it because it would allow us a bit of debugging in case we needed it, but of course the rest of the payload is unrelated.

I think returning all the packages is problematic on many parts, e.g. it is a larger operation on the Datastore as well, and most of its data gets discarded anyway. What we need is a user-scope view of a package (e.g. it contains its uploader status, later it could contain whether the package is starred or not). Some options:
- We could extend some of the current API endpoints that return e.g. the current version request could include the uploader status, e.g. https://pub.dev/api/packages/pana/versions/0.12.18
- We could augment the `/api/packages/pana/*` path with `/user` to fetch data for the current user (similar to the `/metrics` which returns the current analysis summary).
- We could do a new path `/api/account/packages/<package>`?

app/lib/frontend/templates/views/layout.mustache
isoos: `<div>` could work, but I think it is more future-proof to define a JSON structure on both ends and do the encoding/decoding. For example this way we can have embedded maps and lists, and we don't need to wrangle with attribute names how they are mapped to lists.

app/lib/frontend/handlers/routes.dart
isoos: Why `options`? Although, I think I like it a bit more than `flags` as it would allow non-boolean values later.

app/lib/frontend/backend.dart
isoos: At other methods of the `Backend` class we already use the current user implicitly (e.g. upload, uploader add/remove), I think we could assume that for future methods we'd have the same implicit assumption. Or would you like to have it differently?

app/lib/frontend/handlers/routes.dart
jonasfj: Yeah, allowing for other things in the future was the thinking... I'm not sure what those would be...

But things that you can change about a package without publishing a new version... and things that aren't package-version specific could be called _options_ :)

app/lib/frontend/handlers/account.dart
jonasfj: >  and most of its data gets discarded anyway.

Not really, imagine we'll what to show this in the menu when signed-in... maybe not just now, but someday.

It would also be a generically useful API end-point, for the CLI or other integrators.

-----
But good point about wanting to expose whether it's starred! :)

I do like the idea that we put it under `/api/account/...` then... as it's user-specific information (requires you to be signed-in).

This could also have email notification preferences, should we ever add this. 

So this is perhaps user-package-options or user-package-preferences... Just brainstorming here..

 * `GET|PUT /api/account/preferences/packages/<package>`, or,
 * `GET|PUT /api/packages/<package>/user-options`

app/lib/frontend/backend.dart
jonasfj: nvm :)

app/lib/frontend/handlers/account.dart
isoos: These seems to be good to me:
- `/api/account/preferences` - global user preferences (e.g. show/hide e-mail)
- `/api/account/preferences/packages/<package>` - package-specific info (e.g. isUploader) and preferences (e.g. stars)

I have a slight dislike for communicating a read-only status of `isUploader` through "preferences", but I think it is be a good-enough term for it. `fields` or `attributes` doesn't sound any better, does it?

jonasfj: I always imagined stars would be something like:

 * `GET|PUT /api/user/<userId>/stars/<package>`

but I'm starting to think that `<userId>` in URLs is a bad idea :)

Maybe we call it "options"  instead?

`GET|PUT /api/account/options/package/<package>`

isoos: I've settled with `/api/account/options/packages/<package>`, it is pointing towards more consistency with the `/api/packages/<package>/options` (also note: the plural in packages).

isoos:
PTAL, quite a few things changed.

app/lib/frontend/handlers/custom_api.dart
```diff
@@ -232,3 +239,36 @@ Future<shelf.Response> apiSearchHandler(shelf.Request request) async {
...
+      await backend.updateOptions(package, options);
+      return jsonResponse({'success': true}, pretty: false);
+    } on UnauthorizedAccessException catch (_) {
+      return jsonResponse({'error': 'Not Authorized.'},
+          status: 403, pretty: false);
+    } on ClientInputException catch (e) {
+      return jsonResponse({'error': e.message},
+          status: e.status, pretty: false);
+    } catch (e, st) {
+      _logger.warning('Error processing flag update.', e, st);
```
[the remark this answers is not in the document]
isoos: We have a generic catch-all, but not something like this. It could become generic though (in a follow-up PR), e.g. if a request is going for `/api/` it should have a generic json response like the above (otherwise the generic html error page).

pkg/web_app/lib/src/page_data.dart
```diff
@@ -0,0 +1,34 @@
...
+import 'dart:convert';
+import 'dart:html';
+
+import 'package:client_data/page_data.dart';
+
+PageData _data;
+
+/// The server-provided config/data of the current page.
+PageData get pageData {
+  return _data ??= _extractData() ?? PageData();
```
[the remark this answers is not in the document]
isoos: It would make the `isPackagePage` call more involved, I'd rather keep it like this.

pkg/client_data/test/ensure_build_test.dart
```diff
@@ -0,0 +1,53 @@
...
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:test/test.dart';
+
+final _pub = p.join(p.dirname(Platform.resolvedExecutable), 'pub');
+
+void main() {
+  test('ensure local build succeeds with no changes', () {
```
jonasfj: Why is this not using `build_verify`?

app/lib/frontend/handlers/account.dart
```diff
@@ -2,12 +2,25 @@
...
+import '../../frontend/backend.dart';
 import '../../shared/handlers.dart';
 
-/// Handles /api/account/v1/info
-Future<shelf.Response> accountInfoHandler(shelf.Request request) async {
-  return jsonResponse({'email': authenticatedUser?.email ?? '-'});
+/// Handles /api/account/options/packages/<package>
+Future<shelf.Response> accountPkgOptionsHandler(
+    shelf.Request request, String package) async {
+  if (request.method.toUpperCase() == 'GET') {
```
jonasfj: Don't check the method here?

pkg/web_app/lib/src/account.dart
```diff
@@ -147,3 +156,39 @@ class _AccountNavWidget {
...
   }
 }
+
+class _PkgAdminWidget {
+  Element _root;
+  void init(Element root) {
+    if (!pageData.isPackagePage) return;
+    _root = root;
+    _root.append(elem('h3', classes: ['title'], text: 'Admin'));
+    final discontinueToggle = elem(
```
jonasfj: Aren't we also talking about moving this to a separate tab...

isoos: Yes, but not in this PR.

pkg/client_data/test/ensure_build_test.dart
isoos: This is the only test in `test/` and if we don't have anything else, `pub run test` will complain: `No test files were passed and the default "test/" directory doesn't exist.`.

jonasfj: But isn't exact what `build_verify` does?

https://pub.dev/packages/build_verify

Given that the package contains something that should be build, using this seems to make sense...

isoos: It is, and you are right, but then we'd need to introduce fake tests just to run `pub run test`.

PR #2386, opened 2019-06-20
Full-width tab bar.

isoos:
Screen Shot 2019-06-20 at 12 44 06


isoos:
> Looks good, let's double check how this works on different screensizes... like phones.

On small screen sizes this doesn't change anything (the tab bar is already multiline and the sidebar is already at the bottom).

Issue #2227, opened 2019-04-24
refactor validation of uploaded packages


jonasfj:
We should refactor validation of packages when they are uploaded, such that we don't allow packages that the `pub` CLI client wouldn't allow to be published...

For example the `pub` CLI client requires there to be a LICENSE file... we should also require this.

Ideally, we refactor the code such that the same checks can be used both in the CLI and on the server... These should not include static analysis... this should be light weight package checks.

Note. even if we share this code, not all the of the checks should be shared. Checks that are enforced on the client cannot be loosen again, so we should be careful about apply new checks there.

However, if we add too many checks on the server, we can easily revert those checks.

sigurdm:
Another example:

I tried to upload a package with a dependency_override and got this message:
```
Missing requirements:
* Your pubspec.yaml must not override non-dev dependencies.
  This ensures you test your package against the same versions of its dependencies
  that users will have when they use it.
Sorry, your package is missing a requirement and can't be published yet.
```
I believe we don't enforce this on the server - but we should.

isoos:
On a cursory scan of the problem space, I think we will depend on such shared package from both `pub` client, `pub` server, and possibly `pana` too (e.g. for the full definition of filename patterns, maybe even extracting the content of them)... While the current pub server checks throw an exception right away when they encounter an issue, pub client and pana would continue scanning the archive and work on the entire report.

I think the best way to get this started is to create a new package somewhere with a minimal skeleton, and start porting features to it (e.g. filename candidates and content extraction). Any preference where it should live?

jonasfj:
> Any preference where it should live?

We can just stick it in `pkg/` for now... ideally, we'll adopt it server-side before we adopt it client-side, as the server is more flexible (ie. we can always rollback a mistake). I agree, we should have a separate repo for it in the long run.

sigurdm:
I don't think the server and client checks are the same in general.
Client checks are more general, and can be ignored.
server checks are fundamental.

I don't see that we save much by sharing the code.

PR #2388, opened 2019-06-21
pkg/package_archive

isoos:
- #2227
- I wanted to have a design where we are able to display all of the issues at once for the user before uploader, but also to have an easy way to throw an exception on it (as before) to keep it the same way on the server as before.


pkg/package_archive/pubspec.yaml
```diff
@@ -0,0 +1,13 @@
+name: package_archive
+publish_to: none
+
```
jonasfj: Please give this a description and maybe a `README` that just says what this package does..

pkg/package_archive/lib/src/tar_utils.dart
```diff
@@ -0,0 +1,48 @@
...
+
+/// Reads a text content of [name] from the tar.gz file identified by [path].
+///
+/// When [maxLength] is specified, the text content is cut at [maxLength]
+/// characters (with `[...]\n\n` added to it).
+Future<String> readTarballFile(String path, String name,
+    {int maxLength = 0}) async {
+  final result = await Process.run(
+    'tar',
+    ['-O', '-xzf', path, name],
```
jonasfj: What happens if this a 10GB sparse file?

pkg/package_archive/lib/package_archive.dart
```diff
@@ -0,0 +1,229 @@
...
+
+import 'src/names.dart';
+import 'src/tar_utils.dart';
+
+// The maximum stored length of `README.md` and other user-provided file content
+// that is stored separately in the database.
+final _maxStoredLength = 128 * 1024;
+
+/// A validation error in the package archive.
+class ArchiveError {
```
jonasfj: Maybe call it an `Issue` as `Error` indicates subclassing `Error`.

pkg/package_archive/lib/package_archive.dart
```diff
@@ -0,0 +1,229 @@
...
+  final String message;
+
+  ArchiveError(this.message);
+
+  @override
+  String toString() => message;
+}
+
+/// The observed / extracted information of a package archive.
+class PackageArchive {
```
jonasfj: Is it perhaps a `PackageSummary` instead?

pkg/package_archive/lib/package_archive.dart
```diff
@@ -0,0 +1,229 @@
...
+    this.examplePath,
+    this.exampleContent,
+    this.libraries,
+  });
+
+  bool get hasError => errors != null && errors.isNotEmpty;
+}
+
+/// Observe the .tar.gz archive on [archivePath] and return the results.
+Future<PackageArchive> observePackageArchive(String archivePath) async {
```
jonasfj: Could call it `summarizePackage`?

pkg/package_archive/README.md
```diff
@@ -0,0 +1,5 @@
+# package_archive
```
jonasfj: Couldn't it be called `package_reader` or something else? `pub_package_reader` or `package_validator`?

pkg/package_archive/lib/package_archive.dart
```diff
@@ -0,0 +1,229 @@
...
+  final libraries = files
+      .where((file) => file.startsWith('lib/'))
+      .where((file) => !file.startsWith('lib/src'))
+      .where((file) => file.endsWith('.dart'))
+      .map((file) => file.substring('lib/'.length))
+      .toList();
+
+  errors.addAll(validatePackageName(pubspec.name));
+  errors.addAll(syntaxCheckHomepageUrl(
+      pubspec.homepage ?? pubspec.repository?.toString()));
```
jonasfj: By increasing the level of abstraction I mean... we probably need something like:
https://github.com/dart-lang/pub/tree/master/lib/src/validator

So, maybe we have a `Package` class which has a function to list and read files... and long with parsed pubspec, and README, etc...

We would then have a long list of `Validators` that are passed a `Package` instance, and then return a list of `Issues`.

This way, each `Validator` is probably just a function... we can have them in separate files, and we can easily switch on/off which ones are active by adding removing the functions to/from a list of functions.

Just as importantly, these validators can be tested independently, and each one of them would deal with a separate and independent concern. One for `syntaxCheckHomepageUrl` another for `validatePackageName`.... maybe a different one to `pingHomepageUrl`, etc...

pkg/package_archive/lib/src/tar_utils.dart
isoos: I'd treat this PR as a refactoring, and don't add further checks. We have a check at upload time to block upload if it is larger than a certain size, but it wouldn't hurt to double-check here.

pkg/package_archive/lib/package_archive.dart
isoos: Agreed, but let's do it in a separate PR.

PR #2395, opened 2019-06-25
Updating bin/tools/clear_package_properties.dart

isoos:
- added flag for `dry-run`
- added flag for `verbose` logging
- added stat collection

### DIFF
--- a/app/bin/tools/clear_package_properties.dart
+++ b/app/bin/tools/clear_package_properties.dart
@@ -98,7 +98,7 @@ Future _clearAdditionalProperties<T extends ExpandoModel>(T model) async {
   }
 
   for (String prop in props) {
-    final statKey = '${model.key.type.toString()}-$prop';
+    final statKey = '${model.key.type.toString()}.$prop';
     _removedCount[statKey] = (_removedCount[statKey] ?? 0) + 1;
   }
 
--- a/app/bin/tools/clear_package_properties.dart
+++ b/app/bin/tools/clear_package_properties.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:convert';
 
 import 'package:args/args.dart';
 import 'package:gcloud/db.dart';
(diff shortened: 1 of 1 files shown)


isoos:
Btw. staging has the following properties to be cleared:
-  `Package.uploaderEmails`
- `PackageVersion.uploaderEmail`
-  `PackageVersionPubspec.namespace`
-  `PackageVersionPubspec.qualifiedPackage`
-  `PackageVersionInfo.namespace`
-  `PackageVersionInfo.qualifiedPackage`

app/bin/tools/clear_package_properties.dart
```diff
@@ -27,68 +35,91 @@ Future main(List<String> args) async {
...
-      tx.queueMutations(inserts: [p]);
-      await tx.commit();
-    });
+  if (package.additionalProperties.isNotEmpty) {
+    await _clearAdditionalProperties(package);
+  }
+}
+
+Future _clearAdditionalProperties<T extends ExpandoModel>(T model) async {
+  final props = model.additionalProperties.keys.toList();
```
jonasfj: Can't we move `if (model.additionalProperties.isEmpty) { return; }` in here..?

app/bin/tools/clear_package_properties.dart
```diff
@@ -27,68 +35,91 @@ Future main(List<String> args) async {
...
-  bool packageEntityNeedsUpdate = true,
-}) async {
-  print('Processing package: $package');
-  final pkgKey = dbService.emptyKey.append(Package, id: package);
+Future _processPackage(Package package) async {
+  print('Processing package: ${package.name}');
 
-  final versionQuery = dbService.query<PackageVersion>(ancestorKey: pkgKey);
+  final versionQuery =
+      dbService.query<PackageVersion>(ancestorKey: package.key);
```
jonasfj: This won't work, as you can only do one transaction per second on keys with the same parent... I would assume this causes problems at some point...

Given the size of these, I'm also not sure it's wise to do all of these in one transaction.. but that might be an option.
We would likely need to merge such a transaction with `_clearAdditionalProperties(package)`.

isoos: I've tried it on staging, and it works. It does throttle the operations one second at a time, but it does not fail. We had similar processing previously in other scripts, and it was the same behaviour.

jonasfj: Okay, then let's give it a shot :)

PR #2397, opened 2019-06-25
Log markers for package uploads

jonasfj:
Allowing us to make log metrics for:
 * `[new-package-uploaded]`, and,
 * `[package-uploaded]`.


isoos:
run dartfmt

PR #2402, opened 2019-06-25
Migration atom_feed_test.dart to use fake_gcloud.

isoos:
- #1595
- also fixes a bug in `fake_gcloud` and updates the test package models with the required field.


isoos:
> Wouldn't it be easier to maintain this as an integration test?

We can always improve the integration tests, but I think it is also worth to keep such tests around.

PR #2407, opened 2019-06-26
Throttle TaskScheduler in case there was a network issue.

isoos:


app/lib/shared/task_scheduler.dart
```diff
@@ -96,6 +99,12 @@ class TaskScheduler {
       }
 
       await runTask(task);
+
+      // Throttle processing in case there was a network issue.
+      if (errorCount >= 5) {
+        await Future.delayed(Duration(minutes: 1));
```
jonasfj: Please print a message like: `logger.warning('sleeping 1 minute after 5 errors, before trying more tasks');`

PR #2411, opened 2019-06-26
Updated integration tests.

isoos:
- refactored a few bits, separating `pub` CLI tool from other checks
- I've added checks for a few public pages. It is a bit hard to get a reasonable verification because at the moment the fake server does not have search service connected.
- closing the `http.Client` when the verification ends (no hanging at the end)

### DIFF
--- a/app/lib/frontend/name_tracker.dart
+++ b/app/lib/frontend/name_tracker.dart
@@ -71,11 +71,13 @@ class NameTracker {
 class NameTrackerUpdater {
   final DatastoreDB _db;
   DateTime _lastTs;
+  Completer _sleepCompleter;
+  Timer _sleepTimer;
   bool _stopped = false;
 
   NameTrackerUpdater(this._db);
 
-  // Note, this method never returns.
+  /// The returned future completes after the `stop` method is called.
   Future startNameTrackerUpdates() async {
     final sw = Stopwatch()..start();
     _logger.info('Scanning existing package names');
@@ -93,7 +95,7 @@ class NameTrackerUpdater {
     _logger.info(
         'Scanned initial package names (${nameTracker.length}) in ${sw.elapsed}.');
 
-    await Future.delayed(_pollingInterval);
+    await _sleep();
 
     _logger.info('Monitoring new package creation.');
     for (;;) {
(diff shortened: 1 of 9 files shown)


app/lib/frontend/name_tracker.dart
```diff
@@ -71,14 +71,18 @@ class NameTracker {
 class NameTrackerUpdater {
```
jonasfj: Shameless self-promotion, have you considered: https://pub.dev/packages/neat_periodic_task

If you ignore the status provider part...it has both start and stop.. and it has a timeout :)
 + there are some tests that start/stop works, and of what happens when errors occur... + a bit of logging..

Just saying :)

Note: this is not important.

app/lib/frontend/name_tracker.dart
```diff
@@ -91,27 +95,49 @@ class NameTrackerUpdater {
...
       } catch (e, st) {
         _logger.severe(e, st);
       }
-      await Future.delayed(_pollingInterval);
+      await _sleep();
     }
   }
 
+  Future _sleep() async {
+    _sleepCompleter = Completer();
```
jonasfj: Since you can only ever stop once... `stop = false` is never flipped back, I would suggest creating a single `_sleepCompleter` at the start.. and maybe call it `stopping`...

Note: this is not important.

app/lib/frontend/name_tracker.dart
```diff
@@ -71,14 +71,18 @@ class NameTracker {
...
   DateTime _lastTs;
+  Completer _sleepCompleter;
+  Timer _sleepTimer;
+  bool _stopped = false;
 
   NameTrackerUpdater(this._db);
 
-  // Note, this method never returns.
+  /// The returned future completes after the `stop` method is called.
   Future startNameTrackerUpdates() async {
```
jonasfj: This should throw a `StateError` if it has already been stopped... or at-least that would be good :)

Note: this is not important.

pkg/fake_pub_server/lib/fake_pub_server.dart
```diff
@@ -97,7 +102,10 @@ class FakePubServer {
         _logger.info('fake_pub_server running on port $port');
 
         await ProcessSignal.sigterm.watch().first;
+
+        _logger.info('fake_pub_server shutting down');
         await server.close();
+        nameTrackerUpdater.stop();
```
jonasfj: Why not set this up in the service scope? so that it stops on exit from the service scope?

jonasfj: Crazy idea:

`name_tracker.dart` should export a function called `setupNameTracker()` which will:
 * create a nametracker
 * start the name tracker
 * register the name tracker in the current service_scope
 * register stopping the name tracker in the current service_scope

Then we can call `setupNameTracker()` from `setupServices()` in `services.dart`. We might need a boolean to disable this on processes that doesn't need a name tracker running...

isoos:
I think these are fair points, but I'd defer them to a later PR.

isoos:
Filed #2416 for follow-up

PR #2415, opened 2019-06-27
Use fixed tab labels instead of filenames.

isoos:
- saves some horizontal space
- once `PackageVersionAsset` is ready, it will be easier to render the links without actually loading the asset content (e.g. a rollup-info need to know only if there is a readme file, no need to know also its name).


isoos:
CI is green, but status not updated. Merging anyway.

PR #2419, opened 2019-06-27
Tab change: don't default on readme.

isoos:
- In the future cases where we have tabs without content (only links), we should not default to the readme tab on initialization, rather: use the one activated from the page, or the first one that has content.
This should allow us to have pages with a few intra-page tabs and extra-page tabs mixed, and won't cause weird initialization result where a link-only tab is shown as active. 

- Also updated the style names we are using (removed the weird `js-tabs` and `js-content`), and made the processing methods a bit shorter and more resilient to later changes.

### DIFF
--- a/pkg/web_app/lib/src/tabs.dart
+++ b/pkg/web_app/lib/src/tabs.dart
@@ -44,9 +44,15 @@ void _changeTabOnUrlHash() {
     hash = hash.substring(1);
   }
   // Navigating back to a non-hashed package page will result an empty hash.
-  // Displaying the default tab: readme.
+  // Displaying the first tab (with content) by default.
   if (hash.isEmpty) {
-    changeTab('-readme-tab-');
+    final active = tabContents.firstWhere(
+      (e) => e.classes.contains('-active'),
+      orElse: () => null,
+    );
+    if (active == null) {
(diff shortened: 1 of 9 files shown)


pkg/web_app/lib/src/tabs.dart
```diff
@@ -38,15 +38,21 @@ void _setEventsForTabs() {
...
   String hash = window.location.hash ?? '';
   if (hash.startsWith('#')) {
     hash = hash.substring(1);
   }
   // Navigating back to a non-hashed package page will result an empty hash.
-  // Displaying the default tab: readme.
+  // Displaying the first tab (with content) by default.
   if (hash.isEmpty) {
-    changeTab('-readme-tab-');
+    final active = _contentRoot.children.firstWhere(
```
jonasfj: Have you considered `_contentRoot.querySelector('.-active')`, see:
https://api.dartlang.org/stable/2.4.0/dart-html/Element/querySelector.html

It should return just one element :)

Just a nit..

isoos: In this case it must be one if its direct children. I haven't checked if `>.-active` would work too, but will do (I'll need to update parts of this as tabs will change a bit more).

PR #2420, opened 2019-06-27
Add OpenSearch file

### LINKED ISSUE #2273: Add back link rel="search" – for pub
```
<link href="..." rel="search" title="..." type="application/opensearchdescription+xml">
```

jack-r-warren:
I made a pub.dev OpenSearch file for myself and figured I'd make a PR since it fixes #2273. OpenSearch is what powers Chrome's [Tab to Search](https://www.chromium.org/tab-to-search), Firefox's [one-click search engines](https://developer.mozilla.org/en-US/docs/Web/OpenSearch#OpenSearch_description_file), and similar behavior in other browsers. It boils down to hosting an XML file online and linking to it in the `<head>` tag of any related pages.

Summary of changes: 
- Added a static `osd.xml` file next to the `favicon.ico`
- Updated header template to include the link to the file
- Updated the static file handling to include `osd.xml` the same way it does `favicon.ico` and `robots.txt`
- Updated tests that included the header to include the new element

There's a number of areas in particular where changes might be necessary. 
- Tests: I don't have a dev environment set up for this stuff. I can test the OSD file locally with Firefox but the tests won't run on Windows. Travis should run them but a human glance wouldn't hurt.
- Icons: Firefox needs anything larger than 10KB to be base64 encoded in the OpenSearch file itself, and the spec seems to be a bit specific about sizes. I used the `dart-logo.svg` file to make some custom sizes and put them into the file, though I don't know how best to document that.
- Names: the OpenSearch file's ShortName, LongName, and Description fields maybe need to be changed to align with branding. Anecdotally, most other sites with OpenSearch files have their ShortName more similar to the tab title than the URL (so "Dart packages" rather than "pub.dev").

The one other issue that might come up is that the `osd.xml` file should be served with `Content-Type: application/opensearchdescription+xml`. I don't know the extents to which this is actually necessary, to which the [`mime`](https://pub.dev/packages/mime) package does this for us,or to which we have any control over this outside of the `mime` package. Getting that wrong won't break any existing functionality, however.

Happy to make any changes!


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ AppEngine version, listed here to ease deployment and troubleshooting.
 
 ## Next Release (replace with git tag when deployed)
 
+ * Made search accessible via an OpenSearch description file at `/osd.xml`.
```
isoos: In general we put items here that we need to look out when we are deploying a new version. This is not a risk factor, it can be removed.

app/lib/frontend/static_files.dart
```diff
@@ -13,7 +13,7 @@ import 'package:pana/pana.dart' show runProc;
 import 'package:path/path.dart' as path;
 
 const String _defaultStaticPath = '/static';
-const _staticRootPaths = <String>['favicon.ico', 'robots.txt'];
+const _staticRootPaths = <String>['favicon.ico', 'robots.txt', 'osd.xml'];
```
isoos: `/favicon.ico` and `/robots.txt` are exceptions because they are expected to be there. For `osd.xml` there is no such expectation, let it go under `/static/osd.xml` (and you can remove it from here).

app/lib/frontend/static_files.dart
```diff
@@ -135,6 +135,7 @@ class StaticFile {
 class StaticUrls {
   final String staticPath = _defaultStaticPath;
   final String smallDartFavicon;
+  final String openSearchDescription;
```
isoos: No need for this (see later).

app/lib/frontend/templates/layout.dart
```diff
@@ -31,6 +31,7 @@ String renderLayoutPage(
   @required String title,
   String pageDescription,
   String faviconUrl,
+  String openSearchDescriptionUrl,
```
isoos: No need for this (see later).

app/lib/frontend/templates/layout.dart
```diff
@@ -61,6 +62,8 @@ String renderLayoutPage(
     'no_index': noIndex,
     'static_assets': staticUrls.assets,
     'favicon': faviconUrl ?? staticUrls.smallDartFavicon,
+    'open_search_description':
```
isoos: No need for this (see later).

app/lib/frontend/templates/views/layout.mustache
```diff
@@ -23,6 +23,9 @@
   {{#favicon}}
   <link rel="shortcut icon" href="{{{favicon}}}" />
   {{/favicon}}
+  {{#open_search_description}}
```
isoos: If you are removing it from the root and using `/static/osd.xml`, it can be simply this:

```
<link rel="search" type="application/opensearchdescription+xml" title="Dart packages" href="{{{static_assets.osd_xml}}}"/>
```
(note the closing tag, tests are failing because the lack of it).

static/osd.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
isoos: Copyright (c) 2019

static/osd.xml
```diff
@@ -0,0 +1,27 @@
...
+    </Image>
+    <Url type="text/html" template="https://pub.dev/packages">
+        <Param name="q" value="{searchTerms}"/>
+    </Url>
+    <Url type="application/opensearchdescription+xml" rel="self" template="https://pub.dev/osd.xml"/>
+    <moz:SearchForm>https://pub.dev</moz:SearchForm>
+</OpenSearchDescription>
```
isoos: nit: newline

static/osd.xml
```diff
@@ -0,0 +1,27 @@
...
+    </Image>
+    <Url type="text/html" template="https://pub.dev/packages">
+        <Param name="q" value="{searchTerms}"/>
+    </Url>
+    <Url type="application/opensearchdescription+xml" rel="self" template="https://pub.dev/osd.xml"/>
```
isoos: template="https://pub.dev/static/osd.xml"

jack-r-warren:
Thanks for the pointers! All changes made; tests updated.

jack-r-warren:
Fixed a bug from removing the dedicated variable for the location of the OpenSearch file...

isoos:
The `templates_test` is failing. Please re-run the test with re-generating the golden files:
https://github.com/dart-lang/pub-dev/blob/master/app/test/frontend/templates_test.dart#L37
(and revert it to false before checking in the results)

jack-r-warren:
I'm not able to run any of the tests or regenerate the files since I'm using Windows. I edited the golden files by hand based on the failed tests in Travis--fingers crossed.

app/test/frontend/golden/authorized_page.html
```diff
@@ -15,6 +15,7 @@
     <meta itemprop="image" content="/static/img/dart-logo-400x400.png?hash=mocked_hash_58383222"/>
     <meta property="og:image" content="/static/img/dart-logo-400x400.png?hash=mocked_hash_58383222"/>
     <link rel="shortcut icon" href="/favicon.ico?hash=mocked_hash_985685822"/>
+    <link rel="search" type="application/opensearchdescription+xml" title="Dart packages" href="/static/osd.xml?hash=mocked_hash_869379763">
```
isoos: You'll need to close the element.
Btw. what's the error message you get on Windows running this test?

app/test/frontend/golden/flutter_landing_page.html
```diff
@@ -15,6 +15,7 @@
     <meta itemprop="image" content="/static/img/dart-logo-400x400.png?hash=mocked_hash_58383222"/>
     <meta property="og:image" content="/static/img/dart-logo-400x400.png?hash=mocked_hash_58383222"/>
     <link rel="shortcut icon" href="/favicon.ico?hash=mocked_hash_985685822"/>
+    <link rel="search" type="application/opensearchdescription+xml" title="Dart packages" href="/static/osd.xmlosd.xml?hash=mocked_hash_869379763">
```
isoos: `osd.xmlosd.xml` is wrong here

app/test/frontend/golden/authorized_page.html
jack-r-warren: The `setUpAll` fails because it tries to run `pub get` and can't find the executable on Windows. It is anecdotal but I think I have a fairly typical Windows setup; a workaround I figured out in other projects was to use the `runInShell` flag when running the process. Maybe there's something odd about how Dart looks at the Windows path such that the shell does a better job? I dunno.

Even if that were fixed, though, the next line runs a `build.sh` and that will fail regardless. I tried skipping the setup step of the tests altogether but most things then fail because nothing gets built.

app/test/frontend/golden/pkg_index_page.html
```diff
@@ -16,6 +16,7 @@
     <meta itemprop="image" content="/static/img/dart-logo-400x400.png?hash=mocked_hash_58383222"/>
     <meta property="og:image" content="/static/img/dart-logo-400x400.png?hash=mocked_hash_58383222"/>
     <link rel="shortcut icon" href="/favicon.ico?hash=mocked_hash_985685822"/>
+    <link rel="search" type="application/opensearchdescription+xml" title="Dart packages" href="/static/osd.xml?hash=mocked_hash_869379763">
```
isoos: All of the tests require a closing tag (you've updated like the half of it).
If you want to, I can take over from here, it is not that fun to update these tests by hand.

jack-r-warren: Perhaps, my attempts at using find/replace are being pretty agonizingly slow.

isoos:
@jack-r-warren: Thank you again for your dedicated contribution!

PR #2431, opened 2019-07-02
Fix package_test and add generic fake service initialization.

isoos:
- #2427 disabled the test in `package_test.dart` (did not call `await fn()` inside the callback block).
- configuration uses `services_scope`, allows tests to populate it without interfering with each other
- separated the data and service initialization
- enables easier migration of other tests for #1595


app/lib/shared/configuration.dart
```diff
@@ -12,18 +13,21 @@ import 'urls.dart' as urls;
...
+  Configuration config = ss.lookup(_configurationKey) as Configuration;
+  if (config == null) {
+    config = Configuration.fromEnv(envConfig);
+    ss.register(_configurationKey, config);
+  }
+  return config;
 }
 
+/// Sets the active configuration.
 void registerActiveConfiguration(Configuration configuration) {
```
jonasfj: Why don't we just expose 2 methods:
 * `setupConfigFromEnv`
 * `setupConfigForFakePubServer`

or maybe we could even have just `setupConfig` which reads and env var, determines whether it's prod, staging or fake... (we can modify env vars before calling, in the case of fake stuff).
And then just registers it immediately...

This way only the types, the `activeConfiguration` getter and the `setupConfig` function would need to be public.

app/test/frontend/test_services.dart
```diff
@@ -0,0 +1,54 @@
...
+import 'package:pub_dartlang_org/scorecard/backend.dart';
+import 'package:pub_dartlang_org/scorecard/scorecard_memcache.dart';
+import 'package:pub_dartlang_org/shared/analyzer_client.dart';
+import 'package:pub_dartlang_org/shared/configuration.dart';
+import 'package:pub_dartlang_org/shared/redis_cache.dart';
+
+import '../shared/utils.dart';
+import 'utils.dart';
+
+void testWithServices(String name, Future fn()) {
```
jonasfj: Can we give this a comment so that we know what it does?

app/lib/shared/configuration.dart
isoos: You can't modify the env variable while you are inside a VM (e.g. running `pub run test`). The rest is kind of already in place:
- `factory Configuration.fromEnv`
- `factory Configuration.fakePubServer`

I think the consistency of the `register*` is worth to keep here too, or we should change it at other places too.

PR #2435, opened 2019-07-03
Enable the removal of latest stable version.

isoos:


app/bin/tools/remove_package.dart
```diff
@@ -161,13 +161,28 @@ Future removePackageVersion(String packageName, String version) async {
...
+          'Last version detected. Use full package removal without the version qualifier.');
+    }
+
+    bool updatePackage = false;
+    if (package.latestVersion == version) {
+      package.latestVersionKey = null;
+      updatePackage = true;
+    }
+    if (package.latestDevVersion == version) {
+      package.latestDevVersionKey = null;
```
jonasfj: Don't we need to set this to something?

isoos: Yeah, it looks like we always populate this, even if a package has no dev version. I'd keep it this way now, and we can do a cleanup of that later.

isoos:
Tested it on staging (with the NPE fix).

app/bin/tools/remove_package.dart
```diff
@@ -161,13 +161,30 @@ Future removePackageVersion(String packageName, String version) async {
...
-    if ('${package.latestSemanticVersion}' == version) {
-      throw Exception('Cannot delete the latest version of $packageName.');
+    if (versionNames.length == 1 && versionNames.single == version) {
+      throw Exception(
+          'Last version detected. Use full package removal without the version qualifier.');
+    }
+
+    bool updatePackage = false;
+    if (package.latestVersion == version) {
+      package.latestVersionKey = null;
```
jonasfj: Don't we need to update it to something other than `null`

app/lib/frontend/models.dart
```diff
@@ -101,14 +101,14 @@ class Package extends db.ExpandoModel {
 
   void updateVersion(PackageVersion pv) {
```
jonasfj: Should we document what this does... just a one line comment..

PR #2436, opened 2019-07-03
Separate "Versions" tab from package page.

isoos:
- Introduced `Tab` data structure to have a typed way of describing the different tabs.
- Updated styles: links look and behave the same way as content-tabs (e.g. underline when active or hovering, same clickable content area).
- Package page no longer requires to load all of the versions.
- Package version page has the same header, tab-bar and sidebar as the package page.
- Documentation status request is only sent when loading the versions page.


app/lib/frontend/templates/package.dart
```diff
@@ -292,12 +263,57 @@ String renderPkgShowPage(
...
+    @required this.contentHtml,
+    this.isMarkdown = false,
+  }) : titleHtml = titleHtml ?? htmlEscape.convert(title);
+
+  Tab.withLink({
+    String title,
+    String titleHtml,
+    @required String href,
+  })  : titleHtml =
+            '<a href="$href">${titleHtml ?? htmlEscape.convert(title)}</a>',
```
jonasfj: Are we properly escaping the `href` here?

isoos: This is a url that we generate, we don't escape these in most places.

app/lib/frontend/templates/package.dart
```diff
@@ -292,12 +260,57 @@ String renderPkgShowPage(
...
 }
 
-List<Map<String, String>> _tabsData(
+/// Renders the `views/pkg/tabs.mustache` template.
+String renderPkgTabs(List<Tab> tabs) {
+  final tabsData = tabs.map((t) => t.toMustacheData()).toList();
+  // active: the first one with content
+  for (int i = 0; i < tabs.length; i++) {
+    if (tabs[i].contentHtml != null) {
+      tabsData[i]['active'] = '-active';
```
jonasfj: Couldn't this be called `classes`?

isoos: It is hard, because we use `markdown-body` in the content classes (if `isMarkdown` is set), while `-active` is used in both the headers and the content.
I'll make an effort in a subsequent PR to make them more uniform.

PR #2438, opened 2019-07-03
Use neat_cache instead of pkg/simple_cache

jonasfj:


app/lib/shared/redis_cache.dart
```diff
@@ -81,13 +81,9 @@ class SimpleMemcache {
...
       : _sCache = cache
             .withPrefix('$runtimeVersion/$prefix/')
-            .withLogger(logger, ignoreErrors: true)
             .withTTL(ttl)
             .withCodec(utf8),
-        _bCache = cache
-            .withPrefix('$runtimeVersion/$prefix/')
-            .withLogger(logger, ignoreErrors: true)
-            .withTTL(ttl);
+        _bCache = cache.withPrefix('$runtimeVersion/$prefix/').withTTL(ttl);
```
isoos: `.withLogger`?

isoos:
also need to update .travis

app/lib/shared/redis_cache.dart
jonasfj: It disappeared :)

`neat_cache` always ignores errors now... and just returns `null` if `get()` fails.

PR #2439, opened 2019-07-04
PackageDir abstraction

isoos:
A step towards having a separate verification interface.

### DIFF
--- a/pkg/pub_package_reader/lib/src/package_dir.dart
+++ b/pkg/pub_package_reader/lib/src/package_dir.dart
@@ -1,0 +1,71 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'tar_utils.dart';
+
+/// Helper methods to access package content.
+abstract class PackageDir {
+  /// List all the file names in the directory.
+  Future<List<String>> listFileNames();
+
+  /// Whether the file on the given relative [path] exists.
+  Future<bool> exists(String path);
+
+  /// Read the content of the file on the given relative [path].
+  Stream<List<int>> read(String path);
+
+  /// Read the content of the file on the given relative [path] as String.
+  Future<String> readAsString(
+    String path, {
+    int maxLength,
+    String cutoffPostfix = '[...]',
+  }) async {
+    String content =
+        await Utf8Codec(allowMalformed: true).decodeStream(read(path));
+    if (maxLength != null && maxLength > 0 && content.length > maxLength) {
+      content = content.substring(0, maxLength);
+      if (cutoffPostfix != null) {
+        content += cutoffPostfix;
+      }
+    }
+    return content;
+  }
+
+  /// Free up resources (if it was extracted to a temp directory).
+  Future close();
+}
+
+/// Uses archive.tar.gz to provide access to the package archive.
+///
+/// Caches list of files for faster access.
+class ArchiveTarGzPackageDir extends PackageDir {
+  final String _path;
+  List<String> _fileNames;
+
+  ArchiveTarGzPackageDir(this._path);
+
+  @override
+  Future<bool> exists(String path) async {
+    final names = await listFileNames();
+    return names.contains(path);
+  }
+
+  @override
+  Future<List<String>> listFileNames() async {
+    _fileNames ??= await listTarball(_path);
+    return _fileNames;
+  }
+
+  @override
+  Stream<List<int>> read(String path) {
+    return openTarballFile(_path, path);
+  }
+
+  @override
+  Future close() async {
+    // nothing to do, we don't delete the archive.tar.gz
+  }
+}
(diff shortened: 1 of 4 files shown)


pkg/pub_package_reader/lib/src/package_dir.dart
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  /// Free up resources (if it was extracted to a temp directory).
+  Future close();
+}
+
+/// Uses archive.tar.gz to provide access to the package archive.
+///
+/// Caches list of files for faster access.
+class ArchiveTarGzPackageDir extends PackageDir {
```
jonasfj: Can we make the type private and just expose a function that takes a `tarballPath` and returns `PackageDir`?

pkg/pub_package_reader/lib/src/package_dir.dart
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'tar_utils.dart';
+
+/// Helper methods to access package content.
+abstract class PackageDir {
```
jonasfj: Can we call it `PackageFileSystem` or `PackageContents` or `PackageFolder` ?

jonasfj: What exceptions can this throw, any? It's fine if it only throws `Error` then we don't have document it.

let's also document that the path format, we use slashes, probably without stating slash...

pkg/pub_package_reader/lib/src/package_dir.dart
```diff
@@ -0,0 +1,71 @@
...
+  /// List all the file names in the directory.
+  Future<List<String>> listFileNames();
+
+  /// Whether the file on the given relative [path] exists.
+  Future<bool> exists(String path);
+
+  /// Read the content of the file on the given relative [path].
+  Stream<List<int>> read(String path);
```
jonasfj: What happens if the `path` doesn't exist?

pkg/pub_package_reader/lib/src/package_dir.dart
```diff
@@ -0,0 +1,71 @@
...
+
+import 'tar_utils.dart';
+
+/// Helper methods to access package content.
+abstract class PackageDir {
+  /// List all the file names in the directory.
+  Future<List<String>> listFileNames();
+
+  /// Whether the file on the given relative [path] exists.
+  Future<bool> exists(String path);
```
jonasfj: What happens for folders? symlinks? etc?

pkg/pub_package_reader/lib/src/tar_utils.dart
```diff
@@ -26,21 +26,28 @@ Future<List<String>> listTarball(String path) async {
...
+/// Opens the file [name] to read (as binary).
+Stream<List<int>> openTarballFile(String path, String name) async* {
+  final p = await Process.start(
+    'tar',
+    ['-O', '-xzf', path, name],
+  );
+  yield* p.stdout;
+  final code = await p.exitCode;
+  if (code != 0) {
+    throw Exception('Failed to read tarball contents (code=$code).');
```
jonasfj: Exceptions thrown should be documented

pkg/pub_package_reader/lib/src/package_dir.dart
jonasfj: Can we just make this `Future<List<int>>` for simplicity?

pkg/pub_package_reader/lib/src/package_dir.dart
isoos: I think that's a fair concern, but I'm not sure how should I add to this part. Should it return true only if the file (and parent directories) are not symlinked?

pkg/pub_package_reader/lib/src/package_dir.dart
isoos: Should it also have a `maxLength` property to limit the memory use. Not sure if cleverly compressed archives are still a thing...

pkg/pub_package_reader/lib/src/package_folder.dart
```diff
@@ -0,0 +1,88 @@
...
+
+  /// Whether the file on the given relative [path] exists.
+  ///
+  /// Throws IOException if directory listing or file access fails.
+  Future<bool> exists(String path);
+
+  /// Read the content of the file on the given relative [path].
+  ///
+  /// Emits IOException if directory listing or file access fails.
+  Stream<List<int>> read(String path);
```
jonasfj: May I suggest that we just return `Uint8List`? and if the file doesn't exist we return `null`? It's a lot simpler?

jonasfj: Then none of the methods here can throw any exceptions, we just return `null` if the file doesn't exist..

jonasfj: If we have I/O error, then we throw it as a fatal error and crash..

isoos:
Closing for now, let's do a `package:file` version of the same.

PR #2443, opened 2019-07-04
Add help flag

jonasfj:
Just because it's annoying to read the source ahead of every
invocation..


pkg/pub_integration/bin/pub_integration.dart
```diff
@@ -12,11 +12,16 @@ import 'package:pub_integration/pub_integration.dart';
 /// Runs integration test against prod or staging server.
 Future main(List<String> args) async {
   final _argParser = ArgParser()
+    ..addFlag('help', defaultsTo: false)
```
isoos: `abbr: 'h'`?

PR #2444, opened 2019-07-04
Use neat_cache for FileInfo

jonasfj:
I've been wanting to use the `Cache` abstraction like this for a long time.
Ideally, I want to refactor most of our cache usages to get rid of `SimpleCache`.

Note. This doesn't attempt to cleanup the error handling. We really should handle
relevant exceptions, ignore others, print them at a higher level, monitor logs to find them,
handle them specifically and then eventually switch to crashing when they occur.

For now, I'm just logging them.

### DIFF
--- a/app/lib/shared/dartdoc_memcache.dart
+++ b/app/lib/shared/dartdoc_memcache.dart
@@ -24,7 +24,6 @@ DartdocMemcache get dartdocMemcache =>
 
 class DartdocMemcache {
   final SimpleMemcache _entry;
-  final SimpleMemcache _fileInfo;
   final SimpleMemcache _apiSummary;
 
   DartdocMemcache()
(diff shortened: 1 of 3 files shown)


app/lib/shared/dartdoc_memcache.dart
```diff
@@ -24,7 +24,6 @@ DartdocMemcache get dartdocMemcache =>
 
 class DartdocMemcache {
   final SimpleMemcache _entry;
-  final SimpleMemcache _fileInfo;
```
isoos: There is a benefit of keeping related caches close together, otherwise sooner or later we'll ask: what kind of caches are we using for X? and we'll need to hunt them down with more effort.

If we want to change this specific pattern of grouping, let's do so only with a clear idea where we are heading (consistently with other similar cache groups). If the goal is to replace `SimpleCache` with `cache`, let's do it here, and not move it yet.

jonasfj: Good point, I'll play around with this.

jonasfj:
Closed in favor of alternatives.

Issue #2447, opened 2019-07-05
Package dependency flutter is not hosted on pub.dev.


ledfuse:
I am trying to publish an update for a Flutter plugin on pub.dev. The same operation would work fine just 2 days ago, but now it will refuse to start.

```
$ flutter pub pub publish
Publishing dvote 0.2.2 to https://pub.dartlang.org:
|-- .gitignore
[...]
|   |-- gateway.dart
|   '-- util
|       '-- content-uri.dart
|-- pubspec.yaml
'-- test
    '-- dvote_test.dart

Looks great! Are you ready to upload your package (y/n)? y
Uploading...
Package dependency flutter is not hosted on pub.dev.
pub finished with exit code 1
```

Is there a known cause for `Package dependency flutter is not hosted on pub.dev `?

Thank you!

jonasfj:
Please post your pubspec, probably you have `dependency: flutter: ...` when you actually want an `sdk` constraint instead..
Example:
https://github.com/flutter/plugins/blob/821c04152f211f132496888c14aa53d8de77dae4/packages/camera/pubspec.yaml#L15-L17

https://github.com/flutter/plugins/blob/821c04152f211f132496888c14aa53d8de77dae4/packages/camera/pubspec.yaml#L28-L30

Does this make sense?

ledfuse:
Thanks for your response @jonasfj 

It's weird, because the `pubspec.yaml` file looks pretty much the same as when it was scaffolded by the CLI:

```yaml
name: dvote
description: Flutter plugin to allow native mobile apps to interact with Vocdoni decentralized votes.
version: 0.2.2
author: DVote Labs <root@dvotelabs.com>
homepage: https://gitlab.com/vocdoni/dvote-flutter
repository: https://gitlab.com/vocdoni/dvote-flutter.git
documentation: https://vocdoni.io/docs

environment:
  sdk: ">=2.1.0 <3.0.0"

dependencies:
  flutter:
    sdk: flutter
  http: ^0.12.0+2
  web_socket_channel: ^1.0.13

dev_dependencies:
  flutter_test:
    sdk: flutter

# For information on the generic Dart part of this file, see the
# following page: https://www.dartlang.org/tools/pub/pubspec

# The following section is specific to Flutter.
flutter:
  # This section identifies this Flutter project as a plugin project.
  # The androidPackage and pluginClass identifiers should not ordinarily
  # be modified. They are used by the tooling to maintain consistency when
[...]

PR #2449, opened 2019-07-05
Fix SDK dependency check.

isoos:
#2447


pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -243,6 +243,16 @@ Iterable<ArchiveIssue> forbidGitDependencies(Pubspec pubspec) sync* {
       );
       continue;
     }
+
+    if (entry.value is SdkDependency) {
+      final dep = entry.value as SdkDependency;
+      if (dep.sdk != 'flutter') {
```
jonasfj: There is also `flutter_test`, let's just allow all SDK constraints.

PR #2456, opened 2019-07-08
 simplifying cache setup

jonasfj:
Please have a look, I'm still not entirely convinced we
shouldn't move all the logic into dartdoc backend and
give up on type safety for string patterns...

But at-least now we don't have to do any setup.


app/lib/shared/convert.dart
```diff
@@ -0,0 +1,42 @@
+import 'dart:convert';
```
isoos: I wish we had these in `dart:convert`.

jonasfj: Yes, I considered adding them as a constructor...

app/lib/dartdoc/backend.dart
```diff
@@ -156,12 +156,16 @@ class DartdocBackend {
     // deleted, but the [removeObsolete] should be able to validate it.
     await deleteFromBucket(_storage, entry.inProgressObjectName);
 
-    await dartdocMemcache?.invalidate(entry.packageName, entry.packageVersion);
+    await Future.wait([
+      cache.dartdocEntry(entry.packageName, entry.packageVersion).purge(),
```
isoos: The three related purge may be worth to move to the cache class? Similar to the `invalidatePackageCache` method. (Or maybe, we should have only one `invalidatePackageCache` that also purges the non-versioned ones e.g. cache.dartdocEntry(entry.packageName, 'latest') and the cache.dartdocApiSummary(entry.packageName))

app/lib/frontend/backend.dart
```diff
@@ -55,12 +58,16 @@ Backend get backend => ss.lookup(#_backend) as Backend;
 class Backend {
   final DatastoreDB db;
   final GCloudPackageRepository repository;
-  final UIPackageCache uiPackageCache;
 
-  Backend(DatastoreDB db, TarballStorage storage, {UIPackageCache cache})
```
isoos: this change itself makes it worth it :)

isoos:
Some test mocks need to be updated.

app/lib/dartdoc/backend.dart
jonasfj: I would like the cache class to have no auxiliary methods or magic... just be a mechanical mapping from pattern to `Entry<T>`.

It's only the frontend backend where purging is actually called many places..

PR #2466, opened 2019-07-10
Expand packages handler test with some sanity checks on the package content

isoos:
This is similar to the pub_integration's dev_versions script, but with smaller scope: this doesn't check the alternating display of the versions, only that the page is something like a package page.

### DIFF
--- a/app/test/frontend/handlers/_utils.dart
+++ b/app/test/frontend/handlers/_utils.dart
@@ -34,12 +34,22 @@ Future<shelf.Response> issueGetUri(Uri uri) async {
 }
 
 Future<String> expectHtmlResponse(shelf.Response response,
-    {int status = 200}) async {
+    {int status = 200, List<Pattern> present, List<Pattern> absent}) async {
   expect(response.statusCode, status);
   expect(response.headers['content-type'], 'text/html; charset="utf-8"');
   final content = await response.readAsString();
   expect(content, contains('<!DOCTYPE html>'));
   expect(content, contains('</html>'));
+  for (Pattern p in present ?? []) {
+    if (p.allMatches(content).isEmpty) {
+      throw Exception('$p is missing from the content.');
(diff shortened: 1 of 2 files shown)


app/test/frontend/handlers/_utils.dart
```diff
@@ -34,12 +34,22 @@ Future<shelf.Response> issueGetUri(Uri uri) async {
...
 
 Future<String> expectHtmlResponse(shelf.Response response,
-    {int status = 200}) async {
+    {int status = 200, List<Pattern> present, List<Pattern> absent}) async {
   expect(response.statusCode, status);
   expect(response.headers['content-type'], 'text/html; charset="utf-8"');
   final content = await response.readAsString();
   expect(content, contains('<!DOCTYPE html>'));
   expect(content, contains('</html>'));
+  for (Pattern p in present ?? []) {
```
jonasfj: Use default values in the parameter list instead...
`Future<String> expectHtmlResponse(shelf.Response response, {int status = 200, List<Pattern> present = [], List<Pattern> absent = []})`

app/test/frontend/handlers/_utils.dart
```diff
@@ -34,12 +34,22 @@ Future<shelf.Response> issueGetUri(Uri uri) async {
 }
 
 Future<String> expectHtmlResponse(shelf.Response response,
```
jonasfj: Could we define:
```dart
Future<String> expectHtmlResponse(String path, {
  int status,
  List<Pattern> present = [],
  List<Pattern> absent = [],
}) {
  ...
}
```

Or maybe that's something we do later..

isoos: Do you want to also invoke the handler via path? Then it would be a different for the different services.

I'm also not sure what status would be outside of 200?

PR #2475, opened 2019-07-12
New Admin tab

isoos:
- #1519

- New page on `/packages/<pkg>/admin`.

- The admin tab has new visibility rules:
  - always visible on the admin page
  - becomes visible on other pages if the user has admin rights

- The content of the admin tab has now three blocks:
  - (A): it describes login is needed (default)
  - if the JS code block has authenticated the user, it switches to (B)
  - (B): it describes that the user needs admin rights to access the admin tab
  - if the JS code block has authorized the user, it switches to (C)
  - (C): it has the pre-build HTML blocks of the commands.
- The client-side script will change the visibility of the above blocks, and also hooks-up the event listeners.

- Removed the part which extended the sidebar with the admin command.

- Added a browser-based confirmation that the user really wants to change the status.

- Postponed to a later PR:
  - button-like action button
  - when switch back from discontinued to normal, the analysis is still the old one, we should rather have awaiting status

### DIFF
--- a/app/lib/frontend/handlers/routes.dart
+++ b/app/lib/frontend/handlers/routes.dart
@@ -138,11 +138,6 @@ class PubSiteService {
   // **** Packages
   // ****
 
-  @Route.get('/packages/<package>/versions/<version>/admin')
-  Future<Response> packageVersionAdmin(
-          Request request, String package, String version) =>
-      packageAdminHandler(request, package, versionName: version);
-
   @Route.get('/packages/<package>/versions/<version>')
   Future<Response> packageVersion(
           Request request, String package, String version) =>
--- a/app/lib/frontend/handlers/routes.dart
+++ b/app/lib/frontend/handlers/routes.dart
@@ -138,10 +138,19 @@ class PubSiteService {
   // **** Packages
   // ****
 
+  @Route.get('/packages/<package>/versions/<version>/admin')
+  Future<Response> packageVersionAdmin(
+          Request request, String package, String version) =>
+      packageAdminHandler(request, package, versionName: version);
+
   @Route.get('/packages/<package>/versions/<version>')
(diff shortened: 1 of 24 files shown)


pkg/web_app/lib/src/account.dart
```diff
@@ -159,38 +159,56 @@ class _AccountNavWidget {
...
+  Future _toogleDiscontinued() async {
+    if (!window.confirm(
+        'Are you sure you want change the "discontinued" status of the package?')) {
+      return;
+    }
+    final options =
+        PkgOptions(isDiscontinued: !pageData.pkgData.isDiscontinued);
+    final rs = await client.put(
+      '/api/packages/${pageData.pkgData.package}/options',
+      body: json.encode(options.toJson()),
```
jonasfj: Let's not do this now :)

But would be nice if in `package:client_data` we had something like:
```dart
class PkgOptions {
  ...
  fromJson(...) => ...
  toJson() {...};
}

class PubApiException implements Exception {
  final int status:
  final String mesage;
  ...
}

class PubApi {
  final http.Client _client;
  PubApi(this._client);

  Future<void> putPkgOptions(PkgOptions options) async {
    // HTTP + JSON related boiler plate only.
  }
}
```

app/lib/frontend/handlers/routes.dart
```diff
@@ -138,11 +138,20 @@ class PubSiteService {
   // **** Packages
   // ****
 
+  @Route.get('/packages/<package>/versions/<version>/admin')
```
jonasfj: We don't need this yet...

pkg/web_app/lib/src/account.dart
isoos: Good idea. We could probably reuse that in the pub-integration tests too.

PR #2476, opened 2019-07-12
Fix NPE: modified canonicalizeVersion

### LINKED ISSUE #2474: Visiting package version that doesn't exist causes 500
Requested URL: https://pub.dev/packages/wxx_plugin/versions/0.0.1.
Request ID: e5b7945b2a3f9a090143309f00261cb3

isoos:
- Fixes #2474
- `canonicalizeVersion` now returns null in various failure cases, and the usual lookup afterwards also returns with null. Only the package upload needs to fail hard if the canonicalization fails.
- Also updated redirect url to version page url, removed version tab url.

### DIFF
--- a/app/lib/shared/utils.dart
+++ b/app/lib/shared/utils.dart
@@ -47,10 +47,19 @@ Future<T> withTempDirectory<T>(Future<T> func(Directory dir),
   });
 }
 
+/// Returns null if version is invalid or cannot be normalized.
 String canonicalizeVersion(String version) {
+  if (version == null || version.trim().isEmpty) {
+    return null;
+  }
   // NOTE: This is a hack because [semver.Version.parse] does not remove
   // leading zeros for integer fields.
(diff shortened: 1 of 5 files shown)


app/lib/shared/utils.dart
```diff
@@ -47,10 +47,19 @@ Future<T> withTempDirectory<T>(Future<T> func(Directory dir),
...
+  if (version == null || version.trim().isEmpty) {
+    return null;
+  }
   // NOTE: This is a hack because [semver.Version.parse] does not remove
   // leading zeros for integer fields.
-  final v = semver.Version.parse(version);
+  semver.Version v;
+  try {
+    v = semver.Version.parse(version);
+  } catch (_) {
```
jonasfj: Could we settle with catching `FormatException`, isn't that what is thrown?

It's probably fine to get an error of we're passing null or something else... Or if semver was null...

PR #2478, opened 2019-07-15
Style admin page button.

isoos:
> Screen Shot 2019-07-15 at 12 38 18


app/lib/frontend/templates/views/pkg/admin_page.mustache
```diff
@@ -21,9 +21,9 @@
         If you are no longer developing the package, you can mark it as discontinued.
         Pub site won't run analysis on it, and we won't display it in search results.
```
jonasfj: Maybe add:
> (The package will still remain available)

PR #2484, opened 2019-07-16
Edit publisher description.

isoos:
- #1518
- This is also to create some initial infrastructure (backend registration, test with fake_gcloud).
- TBD:
  - API endpoint
  - `websiteUrl` (should we delete the field or rather update it in the same operation as the description).

### DIFF
--- a/app/lib/publisher/backend.dart
+++ b/app/lib/publisher/backend.dart
@@ -35,6 +35,7 @@ class PublisherBackend {
 
   /// Checks whether the current authenticated user has admin role of the
   /// publisher, and executes [fn] if it does.
+  /// Otherwise, it throws [UnauthorizedAccessException].
   Future<R> _withPublisherAdmin<R>(
       String publisherId, Future<R> fn(Publisher p)) async {
     return await withAuthenticatedUser((user) async {
@@ -58,10 +59,10 @@ class PublisherBackend {
   /// Updates the publisher data.
   Future updatePublisherData(String publisherId, String description) async {
     if (description == null) {
-      throw Exception('Description must not be null.');
+      throw ArgumentError.notNull('description');
     }
     if (description.length > 64 * 1024) {
-      throw Exception('Description too long.');
+      throw ArgumentError('Description too long.');
     }
     await _withPublisherAdmin(publisherId, (_) async {
       await _db.withTransaction((tx) async {
--- a/app/lib/publisher/backend.dart
+++ b/app/lib/publisher/backend.dart
@@ -1,0 +1,76 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:gcloud/db.dart';
+import 'package:gcloud/service_scope.dart' as ss;
+import 'package:logging/logging.dart';
+import 'package:pub_server/repository.dart' show UnauthorizedAccessException;
+
+import '../account/backend.dart';
+
+import 'models.dart';
+
+final _logger = Logger('pub.publisher.backend');
+
+/// Sets the publisher backend service.
+void registerPublisherBackend(PublisherBackend backend) =>
+    ss.register(#_publisherBackend, backend);
+
+/// The active publisher backend service.
+PublisherBackend get publisherBackend =>
+    ss.lookup(#_publisherBackend) as PublisherBackend;
+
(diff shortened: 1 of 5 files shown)


app/test/publisher/backend_test.dart
```diff
@@ -0,0 +1,90 @@
...
+
+void main() {
+  group('PublisherBackend', () {
+    group('Update description', () {
+      testWithServices('No active user', () async {
+        await publisherBackend
+            .updatePublisherData('example.com', 'new description')
+            .catchError(expectAsync1((e) {
+          expect('$e', 'UnauthorizedAccess: No active user.');
+        }));
```
jonasfj:
```dart
await expectLater(
  publisherBackend.updatePublisherData('example.com', 'new description'),
  throwsA(isException),
);
```

This way you assert that the exception actually happens.

app/lib/publisher/backend.dart
```diff
@@ -0,0 +1,76 @@
...
+  PublisherBackend(this._db);
+
+  /// Loads a publisher (or returns null if it does not exists).
+  Future<Publisher> getPublisher(String publisherId) async {
+    final pKey = _db.emptyKey.append(Publisher, id: publisherId);
+    return (await _db.lookup<Publisher>([pKey])).single;
+  }
+
+  /// Checks whether the current authenticated user has admin role of the
+  /// publisher, and executes [fn] if it does.
```
jonasfj: `/// Otherwise, it throws [UnauthorizedAccessException].`

app/lib/publisher/backend.dart
```diff
@@ -0,0 +1,76 @@
...
+        throw UnauthorizedAccessException('User is not an admin.');
+      }
+      return await fn(p);
+    });
+  }
+
+  /// Updates the publisher data.
+  Future updatePublisherData(String publisherId, String description) async {
+    if (description == null) {
+      throw Exception('Description must not be null.');
```
jonasfj: Is this an `Exception` or an `Error`, seems like an `ArgumentError`, consider:
```dart
ArgumentError.checkNotNull(description, 'description');
```
See: https://api.dartlang.org/stable/2.4.0/dart-core/ArgumentError/checkNotNull.html

We should do this whenever it's wrong to call a method with `null` as an argument.

-----
Side note, I suspect it's almost always wrong to throw `Exception`, as users of the function should never be doing a catch-all, thus, we should throw a subclass, either a custom or existing one. If you're intending to communicate this to the request handler and further up to the API caller, then it's probably best to throw a custom `Exception` and catch it later.

We should never catch `Exception` and then print the `e.toString()` in a `400` response to the user. Because we could be exposing exceptions that we didn't intend to catch. We could catch an exception from a grpc/http client that embeds the authentication token used or request signature in the exception message (unlikely, but be defensive). If we get an exception we didn't anticipate it's always an internal error -> 500.

app/test/publisher/backend_test.dart
```diff
@@ -0,0 +1,90 @@
...
+
+import 'package:pub_dartlang_org/account/backend.dart';
+import 'package:pub_dartlang_org/publisher/backend.dart';
+import 'package:pub_dartlang_org/publisher/models.dart';
+
+import '../frontend/utils.dart';
+import '../shared/test_services.dart';
+
+void main() {
+  group('PublisherBackend', () {
```
jonasfj: Do we really want to test this? Can't we cover all this by hitting the API end-points? That way we won't have to refactor tests when/if we refactor the implementation.

I mean the REST API is something we can aim to keep stable, and if we get the same test coverage by hitting it, then why do we need low level tests. Can we get the same test coverage is a fair question though? :)

isoos: Fair point, but at the same time we could just migrate these tests to use API endpoints instead of directly calling the methods here.

jonasfj: Sure, tht works too...

app/lib/publisher/backend.dart
```diff
@@ -0,0 +1,77 @@
...
+        throw UnauthorizedAccessException('User is not an admin.');
+      }
+      return await fn(p);
+    });
+  }
+
+  /// Updates the publisher data.
+  Future updatePublisherData(String publisherId, String description) async {
+    if (description == null) {
+      throw ArgumentError.notNull('description');
```
jonasfj: Check out: https://api.dartlang.org/stable/2.4.0/dart-core/ArgumentError/checkNotNull.html
```dart
ArgumentError.checkNotNull(description, 'description');
```
No need for an if-statement.

PR #2488, opened 2019-07-18
Minimal handler stubs for publishers.

isoos:
#1518


app/lib/frontend/handlers/publisher.dart
```diff
@@ -0,0 +1,71 @@
...
+
+/// Handles requests for GET /api/publisher/<publisherId>
+Future<shelf.Response> getPublisherApiHandler(
+    shelf.Request request, String publisherId) async {
+  // TODO: implement
+  return notFoundHandler(request);
+}
+
+/// Handles requests for PUT /api/publisher/<publisherId>
+Future<shelf.Response> putPublisherApiHandler(
```
jonasfj: Can we call this `updatePublisherApiHandler`

PR #2489, opened 2019-07-18
Added publisher API data structures

jonasfj:


pkg/client_data/lib/publisher_api.dart
```diff
@@ -0,0 +1,150 @@
...
+  /// If left as `null` this field will be ignored, and the description of the
+  /// publisher will not be changed.
+  final String description;
+
+  /// Email to be set as contact email for the publisher.
+  ///
+  /// If left as `null` this field will be ignored.
+  /// If changed, the change will not take effect until the confirmation email
+  /// have been confirmed by the user.
+  final String contact;
```
isoos: should we leave it out for now?

isoos: Or... rather... for now it can be changed only to one of the members...

jonasfj: didn't we resolve that we could use the consent flow for this?

Note. when writing code we can throw a UnimplementedError, until the day we implement it..
It's better to do many small PRs in quick successesion :)

PR #2491, opened 2019-07-18
Update publisher description using the client_data.

isoos:


isoos:
(I didn't migrate all the backend tests yet, first I'll need to refactor some of the test utils, and I'd rather do it in smaller chunks).

app/lib/frontend/handlers/publisher.dart
```diff
@@ -31,8 +33,17 @@ Future<shelf.Response> getPublisherApiHandler(
...
-  return notFoundHandler(request);
+  final body = await request.readAsString();
+  final data = json.decode(body) as Map<String, dynamic>;
+  final rq = UpdatePublisherRequest.fromJson(data);
+  if (rq.contact != null) {
+    // TODO: implement
+    throw UnimplementedError('Contact change is not implemented yet.');
+  }
+  if (rq.description != null) {
+    await publisherBackend.updatePublisherData(publisherId, rq.description);
```
jonasfj: Let's enforce a maximum length of the description... any limit will do... maybe 256kb is sufficient just as a sanity limit.
If the value is higher than 256kb we should return 4xx.

app/lib/frontend/handlers/publisher.dart
```diff
@@ -31,8 +33,17 @@ Future<shelf.Response> getPublisherApiHandler(
 /// Handles requests for PUT /api/publisher/<publisherId>
 Future<shelf.Response> updatePublisherApiHandler(
     shelf.Request request, String publisherId) async {
-  // TODO: implement
-  return notFoundHandler(request);
+  final body = await request.readAsString();
+  final data = json.decode(body) as Map<String, dynamic>;
```
jonasfj: If I send garbage instead of I'll get a 500, If I send `null` as body, I'll get a 500.

Users should never be able to provoke a 5xx error from us. As any 5xx is a bug in our code.

app/lib/shared/handler_helpers.dart
```diff
@@ -142,7 +142,7 @@ $fileAnIssueContent
 Add these details to help us fix the issue:
 ````
 Requested URL: ${request.requestedUri}
-Request ID: ${context.traceId}
+Request ID: ${context?.traceId}
```
jonasfj: Nice catch! :)

app/lib/frontend/handlers/publisher.dart
isoos: I think the backend internal logic should enforce that (which it does and throws ArgumentError). We should have a centralized handler for such exceptions in the handler wrappers though...

app/lib/frontend/handlers/publisher.dart
isoos: ++
should I wait on your code generator on that, or should I introduce something here first?

isoos:
Closing, #2511 will use the api_builder generated code.

PR #2492, opened 2019-07-18
Separated test_models.dart

isoos:


app/test/frontend/backend_test.dart
```diff
@@ -1049,3 +1049,16 @@ void main() {
     }, timeout: Timeout.factor(2));
   });
 }
+
+class TestDelayCompletion {
```
jonasfj: Could we give this some docs, I can't immediately see what it does :)

isoos: I didn't see it either, but I hope that moving tests to fake_gcloud would somehow eliminate the need for it :)

PR #2494, opened 2019-07-18
api_generator

jonasfj:
I just want to know what you think...
I think this could remove a lot of the boiler-plate we're about to write.


jonasfj:
@isoos, please take a look when you wake up.

This is by no means perfect... I'm not sure what we should do for end-points that don't return a JSON body... maybe exposing `http.Response` is better.... But all in all, I think this is reasonably sane.

We can use the clients in the web app too... we'll just have to move the code into `app/lib/client/` and generate dart2js on it... I think this is feasible..

If we combine with `http_retry` we can even get retry for free :)

Note: this design is not suitable for large blobs of data, but afaik we don't serve any binary data, we redirect to GCS for packages.

------
If you think this is reasonably sane, I suggest we make an `api.dart` and put all new API end-points into that... we can evolve this to support more features in the future. Adding querystring and headers to the client as optional arguments probably won't be hard.

app/lib/frontend/handlers/publisher_client.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:client_data/publisher_api.dart' as p;
+import 'package:http/http.dart' as http;
+import 'package:api_builder/_client_utils.dart' as utils;
+
+export 'package:api_builder/_client_utils.dart' show RequestException;
+
+class PublisherApiClient {
+  final utils.Client _client;
+
+  PublisherApiClient(String baseUrl, {http.Client client})
```
isoos: What's the benefit of using this over the generated `PublisherApiClient`?

pkg/api_builder/lib/_client_utils.dart
```diff
@@ -0,0 +1,93 @@
...
+  final String _baseUrl;
+  final http.Client _client;
+
+  Client(String baseUrl, {http.Client client})
+      : _baseUrl = baseUrl,
+        _client = client {
+    ArgumentError.checkNotNull(_baseUrl, 'baseUrl');
+  }
+
+  Future<T> _withClient<T>(FutureOr<T> Function(http.Client) fn) async {
```
isoos: Nit: This won't reuse HTTP connections. I guess it won't be too hard to extend it with "like reuse for 1000 requests".

pkg/api_builder/lib/src/client_library_generator.dart
```diff
@@ -0,0 +1,211 @@
...
+code.Class _buildClientClass(
+  ClassElement cls,
+  List<Handler> handlers,
+) =>
+    code.Class(
+      (b) => b
+        ..docs.addAll([
+          '/// Client for invoking `${cls.name}` through the generated router.',
+          '///',
+          '/// Reponses other than 2xx causes the methods to throw',
```
isoos: Wondering: what would happen if the response is a 300 redirect?

pkg/api_builder/lib/src/client_library_generator.dart
```diff
@@ -0,0 +1,211 @@
...
+      ''') : Code.scope((r) => '''
+        return ${r(returnTypeRef)}.fromJson(await _client.requestJson(
+          verb: '${h.verb.toLowerCase()}',
+          path: '$urlPattern',
+          ${hasPayload ? 'body: payload.toJson(),' : ''}
+        ));
+      '''),
+  );
+}
+/*
```
isoos: Remove?

pkg/client_data/lib/publisher_api.dart
```diff
@@ -144,7 +144,7 @@ class InviteMemberRequest {
 
   // json_serializable boiler-plate
   InviteMemberRequest({@required this.reason, @required this.email});
-  factory InviteMemberRequest.fromJson(Map<String, dynamic> json) =>
+  static InviteMemberRequest fromJson(Map<String, dynamic> json) =>
```
isoos: why?

app/lib/frontend/handlers/publisher_client.dart
jonasfj: oh, my bad this was the hand written code I aimed to create with the generator... this should just be deleted :) hehe

pkg/api_builder/lib/_client_utils.dart
jonasfj: It will if you specify `client` in the constructor.

pkg/api_builder/lib/src/client_library_generator.dart
jonasfj: I think `http.Client` follows redirects.

pkg/api_builder/lib/src/client_library_generator.dart
jonasfj: YES :)

sorry, it got a little late last night, hehe

pkg/client_data/lib/publisher_api.dart
jonasfj: Oh, yeah, nvm..

I suspect `static` is always better since you cannot tear off a constructor, but since I made the thing with code-gen it doesn't matter anymore.

pkg/api_builder/lib/_client_utils.dart
isoos: Ah, true, my mistake.

pkg/api_builder/lib/src/shared.dart
```diff
@@ -0,0 +1,81 @@
...
+                e,
+              )))
+          .expand((i) => i)
+          .toList();
+    }
+    if (classes.isEmpty) {
+      return null; // nothing to do if nothing was annotated
+    }
+
+    // TODO: Type checking at code-gen time would be nice to have.
```
isoos: Nit: I'm not sure, what is the actionable things for us to do here?

pkg/api_builder/lib/src/client_library_generator.dart
```diff
@@ -0,0 +1,166 @@
...
+                  }),
+                )
+                .code,
+          )))
+        ..methods.addAll(
+          _removeDuplicateHandlers(handlers).map(_buildClientMethod),
+        ),
+    );
+
+final _parser = RegExp(r'([^<]*)(?:<([^>|]+)(?:\|([^>]*))?>)?');
```
isoos: Please document this.

pkg/api_builder/lib/api_builder.dart
```diff
@@ -0,0 +1,99 @@
...
+
+  /// Route `TRACE` requests matching [route] to annotated method.
+  const EndPoint.trace(this.route) : verb = 'TRACE';
+}
+
+/// Thrown to cause a different response than what the end-point would normally
+/// return.
+class ResponseException implements Exception {
+  final int status;
+  final List<int> body;
```
isoos: I'm wondering: would it be better if `RequestException` and/or `ResponseException` could accept any body, and then callers of this would not need to manually encode the json.

On the other hand, if it is only this utility that emits them, maybe we could auto-decode them to `Map` or `String` (if we know of the type).

jonasfj: I was thinking we would just make subclasses of `ResponseException`, so we would make the `AuthorizationException extends RequestException` somewhere in our code..

But `RequestException` is only thrown by the client. I don't think we can do anything better there..

jonasfj: Actually, we could have subclasses if `RequestException` in `api_builder`..

Note: `AuthorizationException` is something I would define in `app/shared/exceptions.dart` or something like that...

pkg/api_builder/lib/src/shared.dart
jonasfj: We don't have to do this.. but we could check if the first parameter is a `shelf.Request`, etc... things like this... But we don't have to do this... `shelf_router` will check these things, but it's a lot of coding for little gain.

I left the todo so we can extend if we want to some day.

jonasfj:
> Overall looks good, the exception body (the message payload) could be improved later too.

Yes, we can also improve the JSON parsing and body reading, what if an exception is thrown while trying to read the body... say the client connection breaks? We should handled that and ensure that it doesn't cause nasty log entries in our application..

What if the clients incoming connection break, are we going to return 500... Should we timeout the part where we read the incoming body? Maybe limit the size... things like that.. But we can improve those things over time :)

PR #2495, opened 2019-07-19
Unified issueGet for tests.

isoos:


isoos:
I couldn't move the tests to use the wrappers in one go because wrappers rely on configuration, which are not auto-created for tests.

PR #2496, opened 2019-07-19
Generic exception handler wrappers.

isoos:
Not entirely sure if the current `emitsJsonResponse` flag handles everything, but we can fix it as we go ahead with the handler-level tests.

### DIFF
--- a/app/lib/shared/handler_helpers.dart
+++ b/app/lib/shared/handler_helpers.dart
@@ -61,6 +61,7 @@ shelf.Handler wrapHandler(
     handler = _sanitizeRequestWrapper(handler);
   }
   handler = _httpsWrapper(handler);
+  handler = _exceptionHandlerWrapper(logger, handler);
   handler = _logRequestWrapper(logger, handler);
   handler = _cspHeaderWrapper(handler);
   handler = _requestContextWrapper(handler);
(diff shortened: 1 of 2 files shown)


app/lib/shared/handler_helpers.dart
```diff
@@ -125,15 +131,56 @@ shelf.Handler _logRequestWrapper(Logger logger, shelf.Handler handler) {
...
+  };
+}
 
-      final title = 'Pub is not feeling well';
-      Map<String, String> debugHeaders;
-      if (context.traceId != null) {
-        debugHeaders = {'package-site-request-id': context.traceId};
+shelf.Handler _exceptionHandlerWrapper(Logger logger, shelf.Handler handler) {
+  return (shelf.Request request) async {
+    if (requestContext.emitsJsonResponse) {
```
jonasfj: We could also decide this based on the `accept` header... I'm not suggesting we do :)

isoos: Does `pub` client emits that header? We can move our client-side codes to emit it, and that way it would be much simpler. I'd think that in such case we could omit the `requestContext` part entirely...

isoos: PTAL, I've updated the code to use the `Accept:` header.

isoos:
Closing this in favour of using `ResponseException`.

PR #2501, opened 2019-07-19
Migrate one of the backend test groups.

isoos:


isoos:
I wanted to do more at once, but it seems that to properly test these, we'll need a better initial dataset, e.g. more packages, more package versions on each package. I'll prepare that separately.

PR #2502, opened 2019-07-22
Remove unused SearchClientMock

isoos:


isoos:
I have no idea why CI fails.

PR #2503, opened 2019-07-22
bump package to 2.4.0

jonasfj:


isoos:
Probably need to update travis config too.

PR #2509, opened 2019-07-22
Rename to ApiReponseException

jonasfj:


pkg/api_builder/lib/api_builder.dart
```diff
@@ -76,15 +92,10 @@ abstract class $utilities {
...
-        json.fuse(utf8).encode({
-          'message': 'malformed JSON payload',
-        }),
-        {
-          'content-type': 'application/json; charset="utf-8"',
-          'x-content-type-options': 'nosniff',
-        },
+      throw ApiResponseException(
+        status: 400,
+        code: 'InvalidInput',
```
isoos: invalid-input? not sure if it is important... maybe not at all.
Or `InvalidInput` could be emitted on `InvalidInputException`?

pkg/api_builder/lib/api_builder.dart
```diff
@@ -96,4 +107,17 @@ abstract class $utilities {
...
+  /// Utility method exported for use in generated code.
+  static shelf.Response unhandledError(Object e, StackTrace st) {
+    final incidentId = Uuid().v4();
+    _log.severe(
+        'Unhandled error in API handler (incidentId: $incidentId)', e, st);
+    return ApiResponseException(
+      status: 500,
+      code: 'InternalError',
+      message:
+          'internal server error, ask operator to lookup incidentId: $incidentId',
```
isoos: Capital I in 'internal'

pkg/api_builder/lib/api_builder.dart
jonasfj: I don't care much what the codes are, as long as we try to be consistent :) hehe

There is only a very few of these codes that we'll ever what to handle.

Note: I do prefer `CamelCase` for these error codes as they will be identifiers in clients. We could make an enum of all the error codes, but I think we'll be fine if we just document the ones clients need to handled in the API docs.

Oh, yes, let's say that the comment infront of the `@EndPoint...` annotation is the API documentation.. later we can code-gen some API docs from those :) hehe

PR #2511, opened 2019-07-23
Use client_data for updating publisher description.

isoos:
I'm not entirely sure that it is worth to keep the `frontend/handlers/publisher.dart` part, maybe we should put everything in the `publisher/backend.dart`?


isoos:
I'll update that in a separate PR.

PR #2512, opened 2019-07-23
Move update Publisher API handler to the publisher backend.

isoos:


app/lib/publisher/backend.dart
```diff
@@ -57,21 +58,35 @@ class PublisherBackend {
...
+        publisherId, api.UpdatePublisherRequest(description: description));
+  }
+
+  /// Updates the publisher data.
+  Future<api.PublisherInfo> updatePublisher(
+      String publisherId, api.UpdatePublisherRequest update) async {
+    if (update.description != null) {
+      ArgumentError.checkNotNull(update.description, 'description');
+      if (update.description.length > 64 * 1024) {
+        throw ArgumentError('Description too long.');
```
jonasfj: Awesome, we have to use different exception for this..

PR #2514, opened 2019-07-23
Publisher tests to use API handlers.

isoos:
- updated some of the exceptions.
- fixed the `/api/publishers/` path
- migrated backend-based tests to API handlers


app/test/shared/handlers_test_utils.dart
```diff
@@ -5,9 +5,41 @@
...
 
+import 'package:logging/logging.dart';
 import 'package:shelf/shelf.dart' as shelf;
 import 'package:test/test.dart';
 
+import 'package:pub_dartlang_org/frontend/handlers.dart';
+import 'package:pub_dartlang_org/shared/handler_helpers.dart';
+import 'package:pub_dartlang_org/shared/urls.dart';
+
+Future<shelf.Response> httpRequest(
```
jonasfj: Document what this does and what it's supoosed to be used for :)

app/test/publisher/api_test.dart
```diff
@@ -0,0 +1,96 @@
...
+import '../shared/test_models.dart';
+import '../shared/test_services.dart';
+
+void main() {
+  group('Publisher API', () {
+    group('Update description', () {
+      // invoke the API endpoint
+      Future<shelf.Response> update(String publisherId,
+          {Map<String, dynamic> jsonBody, String authToken}) async {
+        return await httpRequest('PUT', '/api/publishers/$publisherId',
```
jonasfj: No rush, but ideally, we should just use the generated client we made... this is what it's for...

isoos: Good point. I guess I'll do it in a subsequent PR.

PR #2597, opened 2019-07-25
Migrate landing tests.

isoos:
- bridging the search client and the search handlers with mock http client
- fake popularity scores

TBD later: I think we'll need to have a single method that rebuilds the package index and returns. That would be useful both here and later on with the integration tests too.

### DIFF
--- a/app/lib/shared/utils.dart
+++ b/app/lib/shared/utils.dart
@@ -238,8 +238,6 @@ Future<http.Response> getUrlWithRetry(http.Client client, String url,
         return result;
       }
     } catch (e, st) {
-      print(e);
-      print(st);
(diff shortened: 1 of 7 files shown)


app/lib/shared/utils.dart
```diff
@@ -238,6 +238,8 @@ Future<http.Response> getUrlWithRetry(http.Client client, String url,
         return result;
       }
     } catch (e, st) {
+      print(e);
+      print(st);
```
jonasfj: If this is for testing only, let's move to `test/utils.dart`

isoos: leftover debug statement, removing

PR #2598, opened 2019-07-25
Unify and handle exceptions

jonasfj:
This is highly inspired by the PR @isoos filed recently.

### DIFF
--- a/app/lib/shared/exceptions.dart
+++ b/app/lib/shared/exceptions.dart
@@ -175,14 +175,14 @@ class AuthorizationException extends ResponseException
   /// of [package].
   factory AuthorizationException.userCannotUploadNewVersion(String package) =>
       AuthorizationException._(
-        'Insufficient permissions to upload new versions of package `$package`',
+        'Insufficient permissions to upload new versions of package `$package`.',
       );
 
   /// Signaling that the user does not have permissions to change uploaders for
   /// given [package].
   factory AuthorizationException.userCannotChangeUploaders(String package) =>
       AuthorizationException._(
-        'Unsufficient permissions to change uploaders for `$package`',
+        'Unsufficient permissions to change uploaders for `$package`.',
       );
 
   /// Signaling that the user is not an administrator for the given [publisher]
@@ -214,11 +214,11 @@ class ConflictException extends ResponseException {
 /// Thrown when the analysis for a package is not done yet.
 class MissingAnalysisException extends NotFoundException {
   MissingAnalysisException()
-      : super('Analysis is not ready for the given package');
+      : super('Analysis is not ready for the given package.');
 }
 
 /// Thrown when package or versions is missing or has flags indicating that it
 /// should be removed from the search index.
 class RemovedPackageException extends NotFoundException {
-  RemovedPackageException() : super('Package has been removed');
+  RemovedPackageException() : super('Package has been removed.');
 }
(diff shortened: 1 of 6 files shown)


jonasfj:
@isoos, there is other exception types defined in `pub_server` we should make subclasses of those, and move them all into `exceptions.dart`. But let's do that in a later PR..

app/lib/shared/exceptions.dart
```diff
@@ -2,21 +2,223 @@
...
+  /// than [maximum].
+  static void checkRange<T extends num>(
+    T value,
+    String name, {
+    T minimum,
+    T maximum,
+  }) {
+    _check(value != null, () => '"$name" cannot be `null`');
+    assert(name != null, '"name" must not be `null`');
+    _check((minimum == null || value >= minimum),
```
isoos: It is a bit confusing to read this, I was excepting to throw the exception when the condition is true. Maybe `_assertOrThrow`?

app/lib/shared/exceptions.dart
```diff
@@ -2,21 +2,223 @@
...
+      );
 
-/// Exception when a package or a version is missing, or has other flags that
-/// indicate it should be removed from the search index.
-class RemovedPackageException extends NotFoundException {}
+  /// Signaling that the user does not have permissions to upload a new version
+  /// of [package].
+  factory AuthorizationException.userCannotUploadNewVersion(String package) =>
+      AuthorizationException._(
+        'Insufficient permissions to upload new versions of package `$package`',
```
isoos: Nit: ending .

app/lib/shared/exceptions.dart
```diff
@@ -2,21 +2,223 @@
...
+      );
+
+  /// Signaling that the user does not have permissions to change uploaders for
+  /// given [package].
+  factory AuthorizationException.userCannotChangeUploaders(String package) =>
+      AuthorizationException._(
+        'Unsufficient permissions to change uploaders for `$package`',
```
isoos: Nit: ending .

app/lib/shared/exceptions.dart
```diff
@@ -2,21 +2,223 @@
...
+  /// Create a [ConflictException] with a [message] explaining what the conflict
+  /// is and how to resolve it.
+  ConflictException._(String message)
+      : super._(409, 'RequestConflict', message);
+}
+
+/// Thrown when the analysis for a package is not done yet.
+class MissingAnalysisException extends NotFoundException {
+  MissingAnalysisException()
+      : super('Analysis is not ready for the given package');
```
isoos: Nit: ending .

app/lib/shared/exceptions.dart
```diff
@@ -2,21 +2,223 @@
...
+}
+
+/// Thrown when package or versions is missing or has flags indicating that it
+/// should be removed from the search index.
+class RemovedPackageException extends NotFoundException {
+  RemovedPackageException() : super('Package has been removed');
```
isoos: Nit: ending .

app/lib/shared/exceptions.dart
jonasfj: Checks that the condition is true... maybe we could name it `_ensure` but it's really named after the public variant..

isoos:
Landing this with the failing publisher backend tests, as they are (a) not in use yet, and (b) will be refactored in #2514 (and there is already enough merge conflict between the two, not worth to fix the tests.)

PR #2599, opened 2019-07-25
hack to avoid listing packages that are not present

jonasfj:


jonasfj:
It might worth adding tests covering this hack... I just omitted as we had to deploy quickly last night.

PR #2602, opened 2019-07-26
Migrate publisher tests to PubApiClient

isoos:


app/test/shared/handlers_test_utils.dart
```diff
@@ -5,42 +5,18 @@
...
-    body: jsonBody == null ? null : json.encode(jsonBody),
-  );
-  final handler = wrapHandler(
-    Logger.detached('test'),
-    createAppHandler(null),
-    sanitize: true,
-  );
-  return await handler(request);
+Future expectApiException(
+    Future future, int status, String code, String message) async {
```
jonasfj: I think it would be fine if `status`, `code` and `message` was optional.

I don't think we should make tests covering `message`, that's too fragile... If we want we can pass a `Matcher` that message should match... Or accept a message substring..

app/test/shared/test_services.dart
```diff
@@ -95,13 +99,42 @@ void testWithServices(String name, Future fn()) {
...
+  String authToken,
+}) {
+  handler ??= createAppHandler(null);
+  handler = wrapHandler(
+    Logger.detached('test'),
+    handler,
+    sanitize: true,
+  );
+  return http_testing.MockClient(
+      _wrapShelfHandler(handler, authToken: authToken));
```
jonasfj: Probably unrelated to this PR, just curious why don't we listen on a port an make actual http requests, instead of using MockClient, which seems like it's a fake client?

app/test/shared/test_services.dart
```diff
@@ -95,13 +99,42 @@ void testWithServices(String name, Future fn()) {
...
+/// Creates local, non-HTTP-based API client with [authToken].
+PubApiClient createPubApiClient({String authToken}) =>
+    PubApiClient('http://localhost:0/',
+        client: httpClient(authToken: authToken));
+
+/// Returns a HTTP client that bridges HTTP requests and shelf handlers without
+/// the actual HTTP transport layer.
+///
+/// If [handler] is not specified, it will use the default frontend handler.
+http_testing.MockClient httpClient({
```
jonasfj: Consider of this needs to be exported

app/test/shared/test_services.dart
isoos: It is much faster to do this wrapping than waiting for the OS to open an unused port number. I think with the current wrapping we cover a lot in our code, and using the HTTP wrapping wouldn't cover more of it.

app/test/shared/handlers_test_utils.dart
isoos: Updated: the optional message will be matched with `contains(message)`.

PR #2605, opened 2019-07-26
Migrate listing search test.

isoos:


app/test/shared/test_services.dart
```diff
@@ -136,7 +139,10 @@ http_testing.MockClientHandler _wrapShelfHandler(
         ...rq.headers,
       },
     );
-    final rs = await handler(shelfRq);
+    shelf.Response rs;
+    await fork(() async {
```
jonasfj: Maybe add a comment explaining why we need to fork a service scope here :)

PR #2617, opened 2019-08-02
Publisher API

isoos:
- #1518
- I've split the `role` and `isPending` properties in `PublisherMember`, I think it is better choice than the initial `role=pending` (this way we can have non-admin invites too).
- Test coverage for `publisher/backend.dart` is 97%+.
- The consent backend is a bit strange with the action plugging, but we can iterate on it later. (Same for the actual e-mail and consent message.)


app/lib/account/consent_backend.dart
```diff
@@ -0,0 +1,176 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:client_data/account_api.dart' as api;
```
jonasfj: FYI. I LOVE THIS, it makes the code so much more readable.

app/lib/account/consent_backend.dart
```diff
@@ -0,0 +1,176 @@
...
+      String consentId, api.ConsentResult result) async {
+    return await withAuthenticatedUser((user) async {
+      final key = _db.emptyKey
+          .append(User, id: user.userId)
+          .append(Consent, id: consentId);
+      final c = (await _db.lookup<Consent>([key])).single;
+      if (c == null) {
+        throw NotFoundException.resource('consent: $consentId');
+      }
+      if (result.granted ?? false) {
```
jonasfj: The `ConsentResult` does not specify the value of `granted` if left as `null`, this might be an oversight, since the property doesn't have documentation.

But it's probably better to do:
`InvalidInputException.checkNotNull(result.granted, 'granted');`

app/lib/account/models.dart
```diff
@@ -47,3 +49,79 @@ class UserInfo extends db.ExpandoModel {
...
+
+  /// Whether a new notification should be sent.
+  bool shouldNotify() =>
+      notificationCount == 0 ||
+      DateTime.now().toUtc().isAfter(nextNotification);
+}
+
+/// Calculates the dedupId of a consent request.
+String consentDedupId(String type, List<String> args) =>
+    Uri(pathSegments: [type, ...args]).toString();
```
jonasfj: Why is this a URI? how about `[type, ...args].join('/')`, and why not hash it?

app/lib/account/consent_backend.dart
```diff
@@ -0,0 +1,176 @@
...
+class InviteStatus {
+  final bool notificationSent;
+  final DateTime nextNotification;
+
+  InviteStatus(this.notificationSent, this.nextNotification);
+}
+
+/// Callback that will be called on consent actions.
+abstract class ConsentAction {
+  /// Callback on accepting the consent.
```
jonasfj: I suggest we add:
 * `renderEmailSubject(List<String> args)`
 * `renderEmailContents(String activeAccountEmail, String invitedEmail, String consentUrl, List<String> args)`
 * `renderConsentDescription(String activeAccountEmail, String invitedEmail, String consentUrl, List<String> args)`

This way the templates for these can exist in the locations where we define the consent `type`.

app/lib/account/models.dart
```diff
@@ -47,3 +49,79 @@ class UserInfo extends db.ExpandoModel {
...
+  /// The user that this consent is for.
+  String get userId => parentKey.id as String;
+
+  /// A [Uri.path]-like concatenation of identifiers from [type] and [args].
+  /// It should be used to query the Datastore for duplicate detection.
+  @db.StringProperty()
+  String dedupId;
+
+  @db.StringProperty()
+  String type;
```
jonasfj: nit. but `kind` might be a better word, it's less overloaded.

app/lib/account/consent_backend.dart
```diff
@@ -0,0 +1,176 @@
...
+
+  /// Removes obsolete/expired [Consent] entries from Datastore.
+  Future deleteObsoleteConsents() async {
+    final query = _db.query<Consent>()
+      ..filter('expires <', DateTime.now().toUtc());
+    await for (var entry in query.run()) {
+      try {
+        await _delete(entry);
+      } catch (e) {
+        _logger.info(
```
jonasfj: please make it a warning. This is problematic... Also please handle exceptions in the case where the entity doesn't exist, because of racing delete requests.

jonasfj: IMO, we should handle everything that could go wrong, and leave the rest as `_logger.shout`.

app/lib/publisher/backend.dart
```diff
@@ -130,14 +127,19 @@ class PublisherBackend {
     if (userId == authenticatedUser?.userId) {
       throw ConflictException.cantUpdateSelf();
     }
+    if (update.role != null) {
+      InvalidInputException.check(
```
jonasfj: Consider using `InvalidInputException.checkAnyOf(update.role, 'role', PublisherMemberRole.values);`

app/lib/account/consent_backend.dart
```diff
@@ -174,3 +172,16 @@ abstract class ConsentAction {
   /// Callback on rejecting the consent or timeout.
   Future onDelete(String userId, List<String> args);
 }
+
+/// Callbacks for publisher member consents.
+class _PublisherMemberAction implements ConsentAction {
```
jonasfj: What if we give `ConsentAction` a constructor that takes two callbacks... then we won't have to create so many classes...


```dart
class ConsentAction {
  final Future<void> Function(String userId, List<String> args) onAccept;
  ...

  ConsentAction(this.onAccept, ...);
}
```

app/lib/publisher/backend.dart
```diff
@@ -93,6 +96,40 @@ class PublisherBackend {
     return _asPublisherInfo(p);
   }
 
+  /// Invites a user to become a publisher admin.
+  Future<account_api.InviteStatus> invitePublisherMember(
+      String publisherId, api.InviteMemberRequest invite) async {
+    return await _withPublisherAdmin(publisherId, (p) async {
+      InvalidInputException.check(
```
jonasfj: Add some sanity limits..
```dart
InvalidInputException.checkNotNull(invite.email, 'email');
InvalidInputException.checkStringLength(invite.email, 'email', maximum: 4096);
InvalidInputException.checkNotNull(invite.reason, 'reason');
InvalidInputException.checkStringLength(invite.reason, 'reason', maximum: 4096);
```

app/lib/publisher/backend.dart
```diff
@@ -93,6 +96,40 @@ class PublisherBackend {
...
+          isValidEmail(invite.email), 'Invalid e-mail: `${invite.email}`');
+      final user = await accountBackend.lookupOrCreateUserByEmail(invite.email);
+
+      final userId = user.userId;
+      final key = p.key.append(PublisherMember, id: userId);
+      final pm = (await _db.lookup<PublisherMember>([key])).single;
+      if (pm != null) {
+        InvalidInputException.check(pm.isPending, 'User is already a member.');
```
jonasfj: If already a member, then isn't that just fine? If we want to return some error it's a conflict.

IMO, it would be okay to do this for the case where the invite has been accepted, otherwise, keep idempotency in mind.

app/lib/publisher/backend.dart
```diff
@@ -169,6 +206,41 @@ class PublisherBackend {
...
+  /// A callback from consent backend, when a consent is granted.
+  Future inviteConsentGranted(String publisherId, String userId) async {
+    await _db.withTransaction((tx) async {
+      final key = _db.emptyKey
+          .append(Publisher, id: publisherId)
+          .append(PublisherMember, id: userId);
+      final list = await tx.lookup<PublisherMember>([key]);
+      final member = list.single;
+      if (member == null) {
+        throw NotFoundException('Membership invite was deleted.');
```
jonasfj: Create a custom constructor please :)

app/lib/publisher/backend.dart
```diff
@@ -169,6 +206,41 @@ class PublisherBackend {
...
+      }
+      if (member.isPending) {
+        member.isPending = false;
+        tx.queueMutations(inserts: [member]);
+        await tx.commit();
```
jonasfj: Handle exceptions... we need to check if there was a competing request that this exactly this, in which case we can return ok, since the API is then idempotent.

app/lib/publisher/backend.dart
```diff
@@ -169,6 +206,41 @@ class PublisherBackend {
...
+    await _db.withTransaction((tx) async {
+      final key = _db.emptyKey
+          .append(Publisher, id: publisherId)
+          .append(PublisherMember, id: userId);
+      final list = await tx.lookup<PublisherMember>([key]);
+      final member = list.single;
+      if (member == null) return;
+      if (member.isPending) {
+        tx.queueMutations(deletes: [member.key]);
+        await tx.commit();
```
jonasfj: Handle case where there are racing delete requests, if the thing doesn't exist we should just return ok...

app/test/publisher/api_test.dart
```diff
@@ -61,6 +63,241 @@ void main() {
...
+                reason: 'Need more admin.', email: joeUser.email));
+        final consents = await dbService
+            .query<Consent>(ancestorKey: joeUser.key)
+            .run()
+            .toList();
+        final consentId = consents.single.consentId;
+        final client2 = createPubApiClient(authToken: joeUser.userId);
+        final rs2 = await client2.resolveConsent(
+            consentId, account_api.ConsentResult(granted: false));
+        expect(rs2.granted, isFalse);
```
jonasfj: Please test rejecting the invite twice, from where I'm standing that should work, right?

app/lib/publisher/backend.dart
```diff
@@ -169,6 +206,41 @@ class PublisherBackend {
     });
   }
 
+  /// A callback from consent backend, when a consent is granted.
+  Future inviteConsentGranted(String publisherId, String userId) async {
```
jonasfj: Couldn't we wait with creating the `PublisherMember` until we get here? It's up the consent backend to do deduplication and avoid sending emails too frequently... Maybe the consent backend should offer a flow to query for pending consent flows... hmm, maybe... I'm not sure... it just seems we're duplicating data, and we don't really want the `isPending` in `PublisherMember`..

app/lib/publisher/backend.dart
```diff
@@ -91,6 +91,23 @@ class PublisherBackend {
     return _asPublisherInfo(p);
   }
 
+  /// List the members of a publishers
+  Future<api.PublisherMembers> listPublisherMembers(String publisherId) async {
```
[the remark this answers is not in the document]
isoos: The endpoint should not do caching, because we need to authenticate and then authorize the request, and doing that with the code generator would add too much complexity.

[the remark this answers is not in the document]
isoos: (adding TODO)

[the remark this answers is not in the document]
isoos: Ah, this is still ambiguous: the generated HTTP endpoint should not do the caching, as the implementation does the authorization. We could do memcache-based caching in the implementation method, I'm adding a TODO for it.

app/lib/account/models.dart
isoos: This way the `/` in any segment should get escaped properly.

app/lib/publisher/backend.dart
```diff
@@ -117,14 +117,49 @@ class PublisherBackend {
...
+      if (pm == null) {
+        throw NotFoundException.resource('member: $userId');
+      }
+      if (update.role != null && update.role != pm.role) {
+        if (pm.role == PublisherMemberRole.pending) {
+          throw ConflictException.invitePending();
+        }
+        await _db.withTransaction((tx) async {
+          final current = (await tx.lookup<PublisherMember>([key])).single;
+          current.role = update.role;
```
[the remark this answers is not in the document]
isoos: This is checked ~15 lines above.

app/lib/account/models.dart
jonasfj: `[type, ...args].map(Uri.encodeComponent).join('/')`

Rather than abusing `Uri()` for something that's not a URI.

app/lib/publisher/backend.dart
isoos: I'm not using the reason String for anything. Maybe we can remove it? (Why would an admin state a reason to invite somebody else?)

app/lib/publisher/backend.dart
isoos: This check will be removed once we remove `isPending`.

app/lib/publisher/backend.dart
isoos: This check will be removed once we remove isPending.

isoos:
PTAL. I'm deferring the `isPending` removal to a subsequent PR.

app/lib/publisher/backend.dart
jonasfj: fair enough, let's remove it...

I was thinking we would surface it in an email to all the other members.

PR #2623, opened 2019-08-05
Increase HSTS duration to a year.

isoos:


jonasfj:
Maybe leave a comment that this might not be necessary for `.dev` domains...

> Your security is our priority. The .dev top-level domain is included on the HSTS preload list, making HTTPS required on all connections to .dev websites and pages without needing individual HSTS registration or configuration. Security is built in.


See: https://get.dev/#benefits

jonasfj:
Note. setting a HSTS header probably won't hurt though :)

isoos:
> setting a HSTS header probably won't hurt though :)

Especially because we are continue serving `pub.dartlang.org`

PR #2630, opened 2019-08-06
Added createPublisher end-point

jonasfj:


app/lib/publisher/backend.dart
```diff
@@ -60,6 +64,104 @@ class PublisherBackend {
...
+          throw ConflictException.publisherAlreadyExists(publisherId);
+        }
+        // Avoid creating the same publisher again, this end-point is idempotent
+        // if we just do nothing here.
+        return;
+      }
+
+      // Create publisher
+      tx.queueMutations(inserts: [
+        Publisher()
```
isoos: Also create the first member with the current user.

jonasfj: Nice catch! :)

app/lib/publisher/backend.dart
```diff
@@ -60,6 +64,104 @@ class PublisherBackend {
...
 
+  /// Create publisher.
+  Future<api.PublisherInfo> createPublisher(
+    String publisherId,
+    api.CreatePublisherRequest body,
+  ) async {
+    InvalidInputException.checkMatchPattern(
+      publisherId,
+      'publisherId',
+      RegExp(r'^[a-z]+\.[a-z]+$'), // validate we have a top-level domain.
```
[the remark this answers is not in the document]
jonasfj: Good catch..

PR #2632, opened 2019-08-06
Tweak the NGram-similarity function.

### LINKED ISSUE #2626: Search functionality in pub.dev does not take into consideration common typos
URL: https://pub.dev/packages?q=lombock

The above URL returns an empty search result, even though there's a package called `lombok` (very close to `lombock`). I think a Google product should support matching on similar strings :) not only exact search!

isoos:
- Fixes #2626
- The rest of the tests change minimally, and I haven't noticed much side-effect.


isoos:
> So using double made it more precise?

Using double allowed the `math.pow(x, 1.5)` operation.

PR #2636, opened 2019-08-07
Migrate GCloudRepository.addUploader tests.

isoos:
#1595

### DIFF
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -766,8 +766,7 @@ class GCloudPackageRepository extends PackageRepository {
         // Fail if package doesn't exist.
         if (package == null) {
           await T.rollback();
-          throw GenericProcessingException(
-              'Package "$packageName" does not exist');
+          throw NotFoundException.resource(packageName);
         }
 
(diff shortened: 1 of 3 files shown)


app/lib/frontend/backend.dart
```diff
@@ -743,10 +743,10 @@ class GCloudPackageRepository extends PackageRepository {
   }
 
   void _validatePackageUploader(
-      models.Package package, String userId, String userEmail) {
+      String packageName, models.Package package, String userId) {
     // Fail if package doesn't exist.
     if (package == null) {
-      throw GenericProcessingException('Package "$package" does not exist');
+      throw NotFoundException.resource(packageName);
```
jonasfj: Don't we need to make sure that `NotFoundException` implements `GenericProcessingException` such that `package:pub_server` does the right thing? Or is this a custom error we define in pub-dev?

isoos: Good question. `pub_server` will return a 400 with a slightly different JSON structure:
`{'error': {'message': message}}`.

I'd do the `implements GenericProcessingException` for now, and we should consider if/how we want to standardize the message formats.

PR #2638, opened 2019-08-07
Migrate tests in frontend/backend.dart

isoos:
#1595

### DIFF
--- a/app/lib/shared/exceptions.dart
+++ b/app/lib/shared/exceptions.dart
@@ -141,6 +141,14 @@ class InvalidInputException extends ResponseException {
   }
 }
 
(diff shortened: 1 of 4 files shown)


isoos:
> Let's fix the exception, otherwise I think this is good to ship.

I'm not sure what to fix here, maybe a comment did not get through?

app/lib/shared/exceptions.dart
```diff
@@ -141,6 +141,14 @@ class InvalidInputException extends ResponseException {
   }
 }
 
+/// Throws when a resource size exceeded a given limit.
+class LimitExceededException extends ResponseException
```
jonasfj: Rename to `PackageRejectedException`

PR #2642, opened 2019-08-09
Updated uploader checks.

isoos:


app/lib/frontend/backend.dart
```diff
@@ -297,6 +296,38 @@ class Backend {
...
+    }
+    if (p.publisherId == null) {
+      return p.containsUploader(userId);
+    } else {
+      final memberKey = db.emptyKey
+          .append(Publisher, id: p.publisherId)
+          .append(PublisherMember, id: userId);
+      final list = await db.lookup<PublisherMember>([memberKey]);
+      final member = list.single;
+      return member?.role == PublisherMemberRole.admin;
```
jonasfj: Can it ever not have a  `role`?

Why would we do `null` checks, if we don't allow `null`?

app/lib/frontend/backend.dart
```diff
@@ -297,6 +296,38 @@ class Backend {
...
+    }
+  }
+
+  /// Whether the current authenticated user (or [userId] if provided) is a
+  /// package admin (through direct uploaders list or publisher admin).
+  ///
+  /// Throws exception if the user is not an admin.
+  Future checkPackageAdmin(models.Package package, {String userId}) async {
+    // Fail if calling user doesn't have permission to admin package.
```
jonasfj: Also check first that the user is authenticated, if not throw the `AuthenticationMissingException`. Since `isPackageAdmin` won't check if the user is authenticated.

app/lib/frontend/backend.dart
```diff
@@ -297,6 +296,38 @@ class Backend {
       await analyzerClient.triggerAnalysis(package, latestVersion, <String>{});
     });
   }
+
+  /// Whether the current authenticated user (or [userId] if provided) is a
+  /// package admin (through direct uploaders list or publisher admin).
+  ///
+  /// Returns false if the user is not an admin.
+  Future<bool> isPackageAdmin(models.Package p, {String userId}) async {
+    userId ??= authenticatedUser?.userId;
```
jonasfj: Do we use this default value else where but here, if not I suggest we make `userId` a required parameter.

Normally, I would prefer that when we read from `authenticatedUser`, if `authenticatedUser` is `null`, then we throw an `AuthenticationMissingException`.

app/lib/shared/exceptions.dart
```diff
@@ -150,6 +150,20 @@ class PackageRejectedException extends ResponseException
...
 
+/// Thrown when the operation is rejected because of the internal state of a resource.
+class StateException extends ResponseException
+    implements GenericProcessingException {
+  /// The operation tried to update the list of uploaders, but it can't be done
+  /// while the package is owned by a publisher.
+  StateException.publisherOwnedPackageNoUploader(
+      String packageName, String publisherId)
+      : super._(
+            400,
```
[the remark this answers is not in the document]
isoos: I think if the `code` is `OperationForbidden` then the exception class should be `OperationForbiddenException`. Otherwise sounds good to me.

app/lib/frontend/handlers/account.dart
```diff
@@ -30,6 +29,6 @@ Future<shelf.Response> accountPkgOptionsHandler(
     return notFoundHandler(request);
   }
   final options =
-      AccountPkgOptions(isUploader: p.hasUploader(authenticatedUser.userId));
+      AccountPkgOptions(isUploader: await backend.isPackageAdmin(p));
```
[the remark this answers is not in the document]
isoos: `isAdmin` should be enough, given the context is a package.

app/lib/frontend/backend.dart
isoos: `member` can be null (e.g. user is not a member).

app/lib/frontend/backend.dart
isoos: it does check it:
- `authenticatedUser?.userId` is null if the user is not authenticated
- on the next line it returns if value is null

app/lib/frontend/backend.dart
isoos: We do use `userId` unfortunately. I could do a second round of refactoring if I can get that out of the way, but first I'd rather have this merged, as it unblocks other publisher API implementation.

isoos: Hm, actually, I don't think we can get rid of the `userId` parameter: when confirming the addition of an uploader, we do double-check if the originator of the invite is still and admin, and that needs the `userId` parameter.

app/lib/frontend/backend.dart
jonasfj: ah, nevermind :)

app/lib/frontend/backend.dart
jonasfj: Yes, but it doesn't throw `AuthenticationMissingException`

app/lib/frontend/backend.dart
jonasfj: I suggest we make `userId` required, instead of optional. Such that we never reference `authenticatedUser` from within this function.

Normally, if `authenticatedUser` is `null` we should throw `AuthenticationMissingException`. That way we wouldn't make the mistake we made in `checkPackageAdmin` which will throw an authorization error instead of an authentication error if the user is not authenticated.

app/lib/frontend/backend.dart
```diff
@@ -297,6 +296,38 @@ class Backend {
...
+      final list = await db.lookup<PublisherMember>([memberKey]);
+      final member = list.single;
+      return member?.role == PublisherMemberRole.admin;
+    }
+  }
+
+  /// Whether the current authenticated user (or [userId] if provided) is a
+  /// package admin (through direct uploaders list or publisher admin).
+  ///
+  /// Throws exception if the user is not an admin.
```
jonasfj: nit: When we throw exceptions we should document what exceptions we throw.

app/lib/frontend/backend.dart
jonasfj: My point was, we should throw an `AuthenticationMissingException` if th euser is not authenticated.

Throwing `AuthorizationException` when the user is not authenticated is always incorrect. How can we know if the user is authorized, if the user is not authenticated.

isoos:
PTAL. The change in `handlers/account.dart` is a good argument for making the `userId` a required parameter, otherwise it would have missed it...

PR #2645, opened 2019-08-09
Fix escape in downloadUrl

### LINKED ISSUE #2093: When redirecting to blob we're not encoding the URL
This might be a bug we can't fix... Or it might not be a bug at all...

If you list versions available for a package, for example: https://pub.dartlang.org/api/packages/google_sign_in

It contains:
```json
"archive_url":"https://pub.dartlang.org/packages/google_sign_in/versions/4.0.1%2B1.tar.gz"
```
Which is correct...

If you fetch that URL we're redirected to:
```console
curl -v https://pub.dartlang.org/packages/google_sign_in/versions/4.0.1%2B1.tar.gz
...
< HTTP/2 303 
< location: https://storage.googleapis.com/pub-packages/packages/google_sign_in-4.0.1+1.tar.gz
```
Which is not encoded... this **might** be a mistake... I'm not 100% sure...

Note: fixing this might not be possible, as it's probably stored this way in GCS... And if GCS didn't decode the `+` into a space then maybe we can just start to encode it... If GCS does do percent decoding.

Credits @chenglu for sending me down this rabbit hole with https://github.com/dart-lang/pub/issues/2082

isoos:
Fixes #2093

### DIFF
--- a/app/lib/frontend/backend.dart
+++ b/app/lib/frontend/backend.dart
@@ -1052,9 +1052,7 @@ class TarballStorageNamer {
             (namespace == null || namespace.isEmpty) ? '' : 'ns/$namespace/';
 
   /// The GCS object name of a tarball object - excluding leading '/'.
-  String tarballObjectName(String package, String version)
-      // TODO: Do we need some kind of escaping here?
-      =>
+  String tarballObjectName(String package, String version) =>
       '${prefix}packages/$package-$version.tar.gz';
 
   /// The GCS object name of an temporary object [guid] - excluding leading '/'.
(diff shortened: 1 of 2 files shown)


app/lib/frontend/backend.dart
```diff
@@ -1052,17 +1052,15 @@ class TarballStorageNamer {
...
+  String tarballObjectName(String package, String version) =>
       '${prefix}packages/$package-$version.tar.gz';
 
   /// The GCS object name of an temporary object [guid] - excluding leading '/'.
   String tmpObjectName(String guid) => 'tmp/$guid';
 
   /// The http URL of a publicly accessable GCS object.
   String tarballObjectUrl(String package, String version) {
-    final object = tarballObjectName(package, version);
+    final object = tarballObjectName(package, Uri.encodeComponent(version));
```
jonasfj: Are you sure this is a problem? why have anything been working before?

Have you tested this on staging...


Sorry for the fuzz, it just seems like something that we could easily double encode by accident.. and something that could totally break `pub get`.

isoos: I'll test it in staging. Integration test does test the escaped URL and it works.

isoos: I did run the integration test on staging, it looks good.

PR #2650, opened 2019-08-13
Package transfer API (to/from publisher)

isoos:
I was contemplating to restrict transfer from publisher to publisher, but it would be a bit more code and about the same amount of test, so I just went with supporting it.

### DIFF
--- a/app/lib/publisher/backend.dart
+++ b/app/lib/publisher/backend.dart
@@ -327,9 +327,7 @@ api.PublisherInfo _asPublisherInfo(Publisher p) =>
 /// Throws AuthorizationException if the user is not an admin for the publisher.
 Future<R> withPublisherAdmin<R>(
     String publisherId, String userId, FutureOr<R> fn(Publisher p)) async {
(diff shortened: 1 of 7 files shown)


app/lib/publisher/backend.dart
```diff
@@ -329,3 +319,29 @@ class PublisherBackend {
...
     api.PublisherInfo(description: p.description, contactEmail: p.contactEmail);
+
+/// Loads [publisherId] and checks if [userId] is an admin of the publisher, and
+/// runs the callback [fn] with it.
+///
+/// Throws AuthenticationException if the user is provided.
+/// Throws AuthorizationException if the user is not an admin for the publisher.
+Future<R> withPublisherAdmin<R>(
+    String publisherId, String userId, FutureOr<R> fn(Publisher p)) async {
+  if (userId == null) {
```
jonasfj: nit: use `ArgumentError.checkNotNull(userId, 'userId');` since this is just an assertion. Callers are supposed to check authentication first.

PR #2656, opened 2019-08-13
Set publisherId at the time of upload.

isoos:


isoos:
(Submitting to not forget about it, tests will come later)

PR #2674, opened 2019-08-14
Refactor SecretBackend + models.

isoos:


app/lib/service/secret/backend.dart
```diff
@@ -0,0 +1,57 @@
...
+class SecretBackend {
+  final DatastoreDB _db;
+
+  SecretBackend(this._db);
+
+  /// Loads the Secret value from the Datastore.
+  /// Returns null if no Secret entity is found.
+  /// Throws InvalidInputException if the [id] is no valid.
+  Future<String> lookup(String id) async {
+    InvalidInputException.check(SecretKey.isValid(id), '$id is not valid');
```
jonasfj: No, this is not an invalid input exception, this kind of flaw should never cause our app to return 4xx.

Do
```dart
if (!SecretKey.isValid(id)) {
  throw ArgumentError.value(id, 'id', 'invalid secret key identifier');
}
``` 

This should never happen, we should never catch it, therefore it should be an error and not an exception.
If it propagates to our request handler (in some way) it should be a 5xx.

app/lib/service/secret/backend.dart
```diff
@@ -0,0 +1,57 @@
...
+    final key = _db.emptyKey.append(Secret, id: id);
+    final secret = (await _db.lookup<Secret>([key])).single;
```
jonasfj: I love that we abstract away this lookup logic :)

app/lib/service/secret/backend.dart
```diff
@@ -0,0 +1,57 @@
...
+    return secret?.value;
+  }
+
+  /// Updates a Secret value.
+  /// Throws InvalidInputException if the [id] is no valid.
+  Future update(String id, String value) async {
+    InvalidInputException.check(SecretKey.isValid(id), '$id is not valid');
```
jonasfj: Same here.

PR #2675, opened 2019-08-15
Simple tests for uploading with publisher-owned package.

isoos:


app/test/package/package_publisher_test.dart
```diff
@@ -77,6 +81,42 @@ void main() {
...
+  group('Upload with a publisher', () {
+    testWithServices('not an admin', () async {
+      final client = createPubApiClient(authToken: hansUser.userId);
+      await client.setPackagePublisher(
+        hydrogen.package.name,
+        PackagePublisherInfo(publisherId: 'example.com'),
+      );
+      await dbService.commit(inserts: [
+        publisherMember(hansUser.userId, 'non-admin'),
+      ]);
```
jonasfj: Wouldn't it be easier to create a helper function that can do this through the API?

Just saying, if we avoid cheating we get higher test coverage for the interaction of all these things.
(Note, this is not specific to this PR, just a general comment on how we do tests, if it's super slow then maybe we want to cheat)

app/test/package/package_publisher_test.dart
```diff
@@ -77,6 +81,42 @@ void main() {
...
 
+  group('Upload with a publisher', () {
+    testWithServices('not an admin', () async {
+      final client = createPubApiClient(authToken: hansUser.userId);
+      await client.setPackagePublisher(
+        hydrogen.package.name,
+        PackagePublisherInfo(publisherId: 'example.com'),
+      );
+      await dbService.commit(inserts: [
+        publisherMember(hansUser.userId, 'non-admin'),
```
jonasfj: Are you testing a role other than admin here?

isoos: Exactly. With the current API it is not really an approved value, so I couldn't set it that way, just "hacking" the DB for a non-admin value instead, which should be future-proof in the test if we ever introduce more roles.

app/test/package/package_publisher_test.dart
isoos: We want to have various "initial DB state" -> action -> "expected DB state" tests, and while a single case could be extracted via a helper function, each test case is fundamentally either for a different initial state or a different action.

app/test/package/package_publisher_test.dart
jonasfj: @isoos, it might be good to document that you're testing what happens if the datastore is in an invalid state. or not this somehow, that this is not currently possible.

PR #2678, opened 2019-08-16
Admin API

isoos:
- I figured out the `__key__` for the pagination part, but its `fake_gcloud` implementation for it is very limited. However I couldn't figure out how to query `PublisherMember` for `userId` (which is part of its key), so I've added an extra field for it.
- I'm also planning to add more tests, added TODO items for them.

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -7,7 +7,6 @@ import 'dart:convert';
 
 import 'package:gcloud/db.dart';
 import 'package:gcloud/service_scope.dart' as ss;
-import 'package:logging/logging.dart';
 import 'package:neat_cache/neat_cache.dart';
 import 'package:retry/retry.dart';
 import 'package:uuid/uuid.dart';
@@ -19,7 +18,6 @@ import 'auth_provider.dart';
 import 'google_oauth2.dart' show GoogleOauth2AuthProvider;
 import 'models.dart';
 
-final _logger = Logger('pub.account.backend');
 final _uuid = Uuid();
 
 /// Sets the account backend service.
@@ -178,16 +176,7 @@ class AccountBackend {
     if (auth == null) {
       return null;
     }
-    final user = await _lookupOrCreateUserByOauthUserId(auth);
-    if (user.isDeleted) {
-      // We don't expect this to happen, but it might, e.g. when an account was
-      // created with a non-Google e-mail address, and later deleted, then
-      // re-created.
-      _logger.severe(
-          'Authentication attempt with a deleted User: ${user.userId} ${user.email}');
-      return null;
-    }
-    return user;
+    return await _lookupOrCreateUserByOauthUserId(auth);
   }
 
   Future<User> _lookupOrCreateUserByOauthUserId(AuthResult auth) async {
@@ -216,9 +205,10 @@ class AccountBackend {
           .toList();
       // TODO: trigger consistency mitigation if more than one email exists
       if (usersWithEmail.length == 1 &&
-          usersWithEmail.single.oauthUserId == null) {
-        // We've found a single pre-migrated User with empty oauthUserId: need
-        // to create OAuthUserID for it.
+          usersWithEmail.single.oauthUserId == null &&
+          !usersWithEmail.single.isDeleted) {
+        // We've found a single pre-migrated, non-deleted User with empty
+        // `oauthUserId` field: need to create OAuthUserID for it.
         final updatedUser = await _db.withTransaction((tx) async {
           final user =
               (await tx.lookup<User>([usersWithEmail.single.key])).single;
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -7,7 +7,7 @@ import 'dart:convert';
 
 import 'package:gcloud/db.dart';
 import 'package:gcloud/service_scope.dart' as ss;
-
+import 'package:logging/logging.dart';
 import 'package:neat_cache/neat_cache.dart';
 import 'package:retry/retry.dart';
 import 'package:uuid/uuid.dart';
@@ -19,6 +19,7 @@ import 'auth_provider.dart';
 import 'google_oauth2.dart' show GoogleOauth2AuthProvider;
 import 'models.dart';
 
+final _logger = Logger('pub.account.backend');
 final _uuid = Uuid();
 
 /// Sets the account backend service.
@@ -179,6 +180,11 @@ class AccountBackend {
     }
     final user = await _lookupOrCreateUserByOauthUserId(auth);
     if (user.isDeleted) {
+      // We don't expect this to happen, but it might, e.g. when an account was
+      // created with a non-Google e-mail address, and later deleted, then
+      // re-created.
+      _logger.severe(
+          'Authentication attempt with a deleted User: ${user.userId} ${user.email}');
       return null;
     }
     return user;
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -177,7 +177,11 @@ class AccountBackend {
     if (auth == null) {
       return null;
     }
-    return await _lookupOrCreateUserByOauthUserId(auth);
+    final user = await _lookupOrCreateUserByOauthUserId(auth);
+    if (user.isDeleted) {
+      return null;
+    }
+    return user;
   }
 
   Future<User> _lookupOrCreateUserByOauthUserId(AuthResult auth) async {
--- a/app/lib/account/models.dart
+++ b/app/lib/account/models.dart
@@ -22,10 +22,24 @@ class User extends db.ExpandoModel {
   @db.DateTimeProperty()
   DateTime created;
 
-  @db.BoolProperty()
-  bool deletedFlag;
+  /// Set to `true` if user is deleted, may otherwise be `null` or `false`.
+  ///
+  /// Use [isDeleted] to avoid `null` checking.
+  @db.BoolProperty(propertyName: 'isDeleted')
+  bool isDeletedFlag;
 
-  bool get isDeleted => deletedFlag == true;
+  /// [isDeleted] is set when a user account is deleted.
+  /// When this happens user-data such as preferences are purged.
+  ///
+  /// However, we retain the user entity if and only if the user has uploaded
+  /// packages or appears in the history by other means. This is to ensure that
+  /// we can see:
+  /// (A) who uploaded a package, and,
+  /// (B) who granted the permissions that allowed said package to be uploaded.
+  bool get isDeleted => isDeletedFlag == true;
+  set isDeleted(bool value) {
+    isDeletedFlag = value;
+  }
 }
 
 /// Maps Oauth user_id to User.id
--- a/app/lib/account/models.dart
+++ b/app/lib/account/models.dart
@@ -13,12 +13,21 @@ class User extends db.ExpandoModel {
   /// A random UUID id.
   String get userId => id as String;
 
+  /// The Google OAuth2 ID of the [User].
+  ///
+  /// This may be `null` for users that never logged in since we've started
+  /// tracking authentications, or if the user [isDeleted] and the [User] entity
+  /// is retained for audit purposes.
   @db.StringProperty()
   String oauthUserId;
 
   @db.StringProperty()
   String email;
 
+  /// [DateTime] the [User] entity was created.
+  ///
+  /// This may be `null` if the user [isDeleted] and the [User] entity is
+  /// retained for audit purposes.
   @db.DateTimeProperty()
   DateTime created;
 
--- a/app/lib/account/models.dart
+++ b/app/lib/account/models.dart
@@ -21,6 +21,11 @@ class User extends db.ExpandoModel {
 
   @db.DateTimeProperty()
   DateTime created;
+
+  @db.BoolProperty()
+  bool deletedFlag;
+
+  bool get isDeleted => deletedFlag == true;
 }
 
 /// Maps Oauth user_id to User.id
--- a/app/lib/admin/backend.dart
+++ b/app/lib/admin/backend.dart
@@ -43,6 +43,7 @@ class AdminBackend {
   /// List users.
   Future<api.AdminListUsersResponse> listUsers(
       {String continuationToken, int limit = 1000}) async {
+    InvalidInputException.checkRange(limit, 'limit', minimum: 1, maximum: 1000);
     return await _withAdmin((user) async {
       final query = _db.query<User>()
         ..order('__key__')
--- a/app/lib/admin/backend.dart
+++ b/app/lib/admin/backend.dart
@@ -16,6 +16,7 @@ import '../shared/configuration.dart';
 import '../shared/exceptions.dart';
 
 final _logger = Logger('pub.admin.backend');
+final _continuationCodec = stringContinuationCodec();
 
 /// Sets the admin backend service.
 void registerAdminBackend(AdminBackend backend) =>
@@ -52,7 +53,7 @@ class AdminBackend {
       if (continuationToken != null) {
         String lastId;
         try {
-          lastId = _decodeContinuation(user.userId, continuationToken);
+          lastId = _continuationCodec.decode(continuationToken);
         } on FormatException catch (_) {
           throw InvalidInputException.continuationParseError();
         }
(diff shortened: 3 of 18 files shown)


app/lib/admin/backend.dart
```diff
@@ -0,0 +1,206 @@
...
+
+        final id = map['id'] as String;
+        // NOTE: we should fix https://github.com/dart-lang/gcloud/issues/23
+        //       and remove the toDatastoreKey conversion here.
+        final key =
+            _db.modelDB.toDatastoreKey(_db.emptyKey.append(User, id: id));
+        if (id != null) {
+          query.filter('__key__ >', key);
+        }
+      }
```
jonasfj: I think this continuation token logic is too complex. Let's keep it simple:
```dart
import 'package:convert' show hex;

String _decodeContinuationToken(String continuationToken) {
  try {
    return hex.fuse(utf8).decode(continuationToken);
  } on FormatException {
    throw InvalidInputException.invalidContinuationToken();
  }
}

....
```

That'll leave us with:
```dart
// Add key to start from, if continuation token is given.
if (continuationToken != null) {
  query.filter('__key__ >', _db.modelDB.toDatastoreKey(_db.emptyKey.append(
    User,
    id: _decodeContinuationToken(continuationToken),
  ));
}
```

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,206 @@
...
+                ),
+              )
+              .toList(),
+          continuationToken: users.length < limit
+              ? null
+              : _encodeContinuation(user.userId, {'id': users.last.userId}),
+        );
+      } catch (e, st) {
+        print(e);
+        print(st);
```
jonasfj: Cleanup :)

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  /// List users.
+  Future<api.AdminListUsersResponse> listUsers(
+      {String continuationToken, int limit = 1000}) async {
+    return await _withAdmin((user) async {
+      final query = _db.query<User>()
+        ..order('__key__')
+        ..limit(limit);
+
```
jonasfj: `InvalidInputException.checkRange(limit, 'limit', maximum: 1000, minimum: 1);` or something like that..

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,206 @@
...
+        final seniorUser =
+            await accountBackend.lookupUserById(seniorMember.userId);
+        p.contactEmail = seniorUser.email;
+      }
+      tx.queueMutations(inserts: [p], deletes: [member.key]);
+      await tx.commit();
+    });
+  }
+
+  Future<PublisherMember> _remainingSeniorMember(
```
jonasfj: documentation comment, please

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,206 @@
...
+        rethrow;
+      }
+    });
+  }
+
+  /// Removes user from the Datastore and updates the packages and other
+  /// entities they may have controlled.
+  Future removeUser(String userId) async {
+    await _withAdmin((u) async {
+      // TODO: remove once [withAuthenticatedUser] is using [User]
```
jonasfj: I don't follow -- what is the todo?

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,206 @@
...
+      await for (final m in memberQuery.run()) {
+        await _removeMember(user, m);
+      }
+
+      // OAuthUserID
+      if (user.oauthUserId != null) {
+        final key = _db.emptyKey.append(OAuthUserID, id: user.oauthUserId);
+        final mapping = (await _db.lookup<OAuthUserID>([key])).single;
+        if (mapping != null) {
+          await _db.commit(deletes: [mapping.key]);
```
jonasfj: Should we not do all of this in a transaction? We probably can't do all in a transaction since packages have different roots, and there can be many... but we can remove `User` and `OAuthUserID` in a single transaction, right?

And if we remove those last, then any partial removal from packages is fine, as the requester should retry the request later... so long as the user entity still exists..

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,206 @@
...
+      if (!activeConfiguration.admins.contains(user.oauthUserId)) {
+        throw AuthorizationException.userIsNotAdminForPubSite();
+      }
+      return await fn(u);
+    });
+  }
+
+  /// List users.
+  Future<api.AdminListUsersResponse> listUsers(
+      {String continuationToken, int limit = 1000}) async {
```
jonasfj: Can we allow additional query string parameters to query by `email` and `oauthUserId` ?

I know, if either one of those are set, then ideally the database should only have a single matching user entity.

app/lib/admin/backend.dart
isoos: I'm planning to refactor the `withAuthenticatedUser` callback to return with an `User` object instead of the `AuthenticatedUser` object. If that happens, the line that follows which uses the `AuthenticatedUser.userId` to lookup the `User` won't be needed anymore.

app/lib/admin/backend.dart
isoos: I was considering to **not** remove the `User` object, just to clear its fields. That way we won't ever reuse the `userId` not even by accident.

Anyway, we can bundle these together for now.

app/lib/admin/backend.dart
isoos: Sure, but let's do that separately.

app/lib/admin/backend.dart
isoos: I'll rewrite `package:continuation_token` to support the encoding of single Strings. I'd like to keep the base64 encoding as it is shorter than hex.

app/lib/admin/backend.dart
jonasfj: If the uuid is random the risk of reusing one is very small.

app/lib/admin/backend.dart
jonasfj: Separate PR or separate end-point?

isoos: Separate PR.

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,210 @@
...
+
+/// The active admin backend service.
+AdminBackend get adminBackend => ss.lookup(#_adminBackend) as AdminBackend;
+
+/// Represents the backend for the admin handling and authentication.
+class AdminBackend {
+  final DatastoreDB _db;
+  AdminBackend(this._db);
+
+  Future<R> _withAdmin<R>(Future<R> fn(AuthenticatedUser user)) async {
```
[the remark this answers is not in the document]
isoos: Let's do this in a separate PR. PTAL.

app/lib/account/models.dart
```diff
@@ -21,6 +21,11 @@ class User extends db.ExpandoModel {
 
   @db.DateTimeProperty()
   DateTime created;
+
+  @db.BoolProperty()
+  bool deletedFlag;
+
+  bool get isDeleted => deletedFlag == true;
```
isoos: Note: I think this is a much better pattern than having to use `isDeleted == true` in all the other places. If you like it then we should migrate the `Package.isDiscontinued` and similar flags to follow the same pattern.

jonasfj: Regardless of what we do, this is non-trivial to understand. We need to explain why we have this property and what it means.

Maybe, something like:
>   `isDeleted` is set when a user account is deleted. When this happens user-data such as preferences are purged.
However, we retain the user entity if and only if the user has uploaded packages or appears in the history by other means. This is to ensure that we can see:
 (A) who uploaded a package, and,
 (B) who granted the permissions that allowed said package to be uploaded.

If we don't write this down, we have zero chance of remembering exactly why we retained user entities with an `isDeleted` property.

app/lib/admin/backend.dart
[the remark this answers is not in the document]
jonasfj: Agreed, I've started most of this logic in #2684

app/lib/account/backend.dart
```diff
@@ -180,6 +180,9 @@ class AccountBackend {
       return null;
     }
     final user = await _lookupOrCreateUserByOauthUserId(auth);
+    if (user.isDeleted) {
+      return null;
```
jonasfj: This should not be possible, let's write a warning to the logs.
Said log message also serves as documentation of what this is checking for.

app/lib/account/models.dart
```diff
@@ -21,6 +21,11 @@ class User extends db.ExpandoModel {
 
   @db.DateTimeProperty()
   DateTime created;
```
jonasfj: If this and `oauthUserId` can suddenly be `null` we should document this in documentation comments. Otherwise, it's hard to know what the state of these properties is.

Example:
```dart
/// [DateTime] the [User] entity was created.
///
/// This may be `null` if the user [isDeleted] and the [User] entity is retained for audit purposes.
@db.DateTimeProperty()
DateTime created;
```

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,200 @@
...
+
+import '../account/backend.dart';
+import '../account/models.dart';
+import '../package/models.dart';
+import '../publisher/models.dart';
+import '../shared/configuration.dart';
+import '../shared/exceptions.dart';
+
+final _logger = Logger('pub.admin.backend');
+final _continuationCodec = stringContinuationCodec();
```
jonasfj: I would still suggest we keep this as simple as `hex`

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,200 @@
...
+        }
+      }
+
+      final users = await query.run().toList();
+      // We may return a page with users less then a limit, but we always
+      // set the continuation token to the correct value.
+      final newContinuationToken = users.length < limit
+          ? null
+          : _continuationCodec.encode(users.last.userId);
+      users.removeWhere((u) => u.isDeleted);
```
jonasfj: Can't we filter on this in the query? if we add `isDeleted` to the index?

app/lib/admin/backend.dart
jonasfj: You do realize that the user being removed is different from the user making the call, right?

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,200 @@
...
+      if (user == null) return;
+      if (user.isDeleted) return;
+
+      _logger.info(
+          '${u.userId} (${u.email}) initiated the delete of ${user.userId} (${user.email})');
+
+      // Package.uploaders
+      final pkgQuery = _db.query<Package>()..filter('uploaders =', user.userId);
+      await for (final p in pkgQuery.run()) {
+        await _removeUploaderFromPackage(p.key, user.userId);
```
jonasfj: Maybe we should use: https://pub.dev/packages/pool
Such that we can do 10 concurrent calls...

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,200 @@
...
+      if (user.oauthUserId != null) {
+        final key = _db.emptyKey.append(OAuthUserID, id: user.oauthUserId);
+        final mapping = (await _db.lookup<OAuthUserID>([key])).single;
+        if (mapping != null) {
+          await _db.commit(deletes: [mapping.key]);
+        }
+      }
+
+      // User
+      await _markUserDeleted(user);
```
jonasfj: Let's do this in the same transaction where we removed the `OAuthUserID` entity.

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,200 @@
...
+    });
+  }
+
+  Future _removeMember(User user, PublisherMember member) async {
+    final seniorMember =
+        await _remainingSeniorMember(member.publisherKey, member.userId);
+    await _db.withTransaction((tx) async {
+      final p = (await tx.lookup<Publisher>([member.publisherKey])).single;
+      if (seniorMember == null) {
+        p.contactEmail = null;
```
jonasfj: Okay, if we have this (and I think you're right, there is no other option), then we must document in `Publisher.contactEmail` that the property can be `null`. And we'll likely have to handle that case in templates too..

app/lib/admin/backend.dart
```diff
@@ -0,0 +1,200 @@
...
+      if (a.role == PublisherMemberRole.admin) return -1;
+      if (b.role == PublisherMemberRole.admin) return 1;
+      return a.created.compareTo(b.created);
+    });
+
+    return otherMembers.first;
+  }
+
+  Future _markUserDeleted(User user) async {
+    await _db.withTransaction((tx) async {
```
jonasfj: If the user doesn't have any interactions in `History` and doesn't have any interactions in `PackageVersion`, then we could delete the user, right?

But I guess we have no efficient way to query `History` entities to see if the user is present there.. Let's leave a TODO for this. I'm sure we can clean that up when we get around to refactoring the `History` entity.

app/lib/frontend/handlers/pubapi.dart
```diff
@@ -262,11 +263,17 @@ class PubApi {
...
+        continuationToken:
+            request.requestedUri.queryParameters['continuationToken'],
+      );
 
   @EndPoint.delete('/api/admin/users/<userId>')
-  Future<Response> adminRemoveUser(Request request, String userId) async =>
-      Response.notFound('no such thing');
+  Future<Response> adminRemoveUser(Request request, String userId) async {
+    await adminBackend.removeUser(userId);
+    return jsonResponse({'status': 'OK'});
```
jonasfj: nit. `204` with no body is customary..

app/lib/admin/backend.dart
isoos: Oops :)

app/lib/account/models.dart
```diff
@@ -21,6 +21,11 @@ class User extends db.ExpandoModel {
...
+
+  @db.BoolProperty()
+  bool deletedFlag;
```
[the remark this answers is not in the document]
isoos: The `_isDeleted` did not work (I think the `gcloud` mirror resolution tries to resolve the `isDeleted` field for some reason), but `isDeletedFlag` works.

app/lib/admin/backend.dart
isoos: I'm not sure if we can filter on values that are `null`. I thought those entities are not indexed at all.

app/lib/frontend/handlers/pubapi.dart
isoos: I prefer to also have a string output, it is better for debugging (no contemplating whether the body got lost somewhere, is empty on purpose, or had other issues).

app/lib/admin/backend.dart
jonasfj: Should we just make a utility that adds `isDeleted` to all User entities and then have the integrity_check verify this as well?

Essentially, I'm suggesting that we make it an invariant that `isDeleted` is set to either true or false.

isoos: Yeah, let's do this in a separate PR.

app/lib/account/backend.dart
```diff
@@ -177,7 +178,16 @@ class AccountBackend {
     if (auth == null) {
       return null;
     }
-    return await _lookupOrCreateUserByOauthUserId(auth);
+    final user = await _lookupOrCreateUserByOauthUserId(auth);
+    if (user.isDeleted) {
+      // We don't expect this to happen, but it might, e.g. when an account was
+      // created with a non-Google e-mail address, and later deleted, then
+      // re-created.
```
jonasfj: I think this is the wrong place to do this check...

Inside: `_lookupOrCreateUserByOauthUserId` we do a lookup for `usersWithEmail`, this is probably the place where if the user we found an email for `isDeleted` then we should not associate the oauthId with said user.

This means that we can end up with the case where more than one user entity has the same email. But that should be fine.

app/lib/account/backend.dart
```diff
@@ -177,7 +178,16 @@ class AccountBackend {
     if (auth == null) {
       return null;
     }
-    return await _lookupOrCreateUserByOauthUserId(auth);
+    final user = await _lookupOrCreateUserByOauthUserId(auth);
+    if (user.isDeleted) {
```
jonasfj: If this happens, it's likely because the account was deleted and then recovered. In this case, I think we should change `isDeleted` to true. Hmm... I'm not sure... we should think this through, if we get it wrong we could introduce a security vulnerability.

isoos: I wouldn't set `isDeleted` to `false` automatically (if that's what you've meant with true :) ), rather let the user escalate the issue. This should be rare, it should be a very exceptional case, we should rather spend a bit of time with it instead automatically give it a pass...

app/lib/account/backend.dart
isoos: Good point, moved.

PR #2684, opened 2019-08-20
Support for authentication with id_tokens

jonasfj:


app/lib/account/auth_provider.dart
```diff
@@ -0,0 +1,27 @@
+class AuthResult {
```
isoos: Nit: license header

app/lib/account/google_oauth2.dart
```diff
@@ -0,0 +1,248 @@
+import 'dart:convert' show json;
```
isoos: Nit: license header

app/lib/account/auth_provider.dart
jonasfj: Good catch! :)

PR #2689, opened 2019-08-21
Upload with publisher: e-mail notifications to admins + test

isoos:


app/lib/publisher/backend.dart
```diff
@@ -217,6 +217,20 @@ class PublisherBackend {
     });
   }
 
+  /// The list of e-mail addresses of the members with admin roles. The list
+  /// should be used to notify admins on upload events.
+  Future<List<String>> getAdminMemberEmails(String publisherId) async {
+    final result = <String>[];
+    final key = _db.emptyKey.append(Publisher, id: publisherId);
+    final query = _db.query<PublisherMember>(ancestorKey: key)
+      ..filter('role =', PublisherMemberRole.admin);
```
jonasfj: Why not just email all members of the publisher... I don't think we need to filter on email...

app/lib/publisher/backend.dart
```diff
@@ -217,6 +217,20 @@ class PublisherBackend {
...
+    await for (final m in query.run()) {
```
jonasfj: Should we tweak `getEmailOfUserId` to accept a list of userIds and return a list of emails..
Wouldn't that be more efficient in terms of datastore operations?

app/test/package/backend_test.dart
```diff
@@ -480,6 +482,56 @@ void main() {
           // TODO: check history
           // TODO: check assets
         });
+
+        testWithServices('package under publisher', () async {
```
jonasfj: Couldn't we cover this by hitting the API? thus, creating a test that we don't need to update when we migrate away from `pub_server`?

jonasfj:
All comments, are nit, we can do them later --- or not at all, just thoughts :)

app/test/package/backend_test.dart
isoos: We could, but it is much more work to get all the required boilerplate setup and going. `pkg/pub_integration` tests do that, not sure if we really need to have that kind of setup for all the tests.

app/lib/publisher/backend.dart
isoos: Good point, we already have a batching for it, switched to `getEmailsOfUserIds`. Under the hood it uses a cache too.

app/lib/publisher/backend.dart
isoos: Sure, let's start with that.

PR #2690, opened 2019-08-21
Simple upload flow: reuse async upload.

isoos:
I know that we don't use the simple upload flow from the client, but we do have tests for both. If we have a unified code like this one, we could remove the test duplicates and only test for the simple upload flow.


jonasfj:
> ... and only test for the simple upload flow.

I think only testing the simple upload flow is undesirable..

isoos:
> I think only testing the simple upload flow is undesirable..

With this change there will be no difference between the async test and the simple test, except the simple will be shorter. We already have end2end test with the pub client, which covers the async flow.

PR #2691, opened 2019-08-21
HTML-safe shared page data.

### LINKED ISSUE #2663: remove custom ld+json we mis-use for meta-data
See: https://search.google.com/structured-data/testing-tool/u/0/#url=https%3A%2F%2Fpub.dev%2Fpackages%2Fretry

We use custom ld+j́son for meta-data we wish to feed our scripts:
```html
<script type="application/ld+json">
{"@context":"https://pub.dev","data":{"pkgData":{"package":"retry","version":"3.0.0","isDiscontinued":false}}}
</script>
```

We could probably just use a property on an HTML element instead, like:
`<body data-package-info="<json-blob>">`

Nobody outside our javascript will attempt to interpret that :)

----
Other ideas are welcome, but it might be simpler and better to do something custom that nobody else will attempt to interpret.

isoos:
- Fixes #2663.
- The `meta` header felt a little but better than our previous discussion with the `body` attribute, but it can be easily changed to that.

### DIFF
--- a/pkg/client_data/lib/page_data.dart
+++ b/pkg/client_data/lib/page_data.dart
@@ -2,10 +2,15 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:convert';
+
 import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart';
 
 part 'page_data.g.dart';
+
+/// Codec to serialize the [PageData] JSON content on the HTML page.
+final pageDataJsonCodec = json.fuse(utf8).fuse(base64);
 
 /// The server-provided config/data for the current page.
 ///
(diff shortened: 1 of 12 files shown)


pkg/client_data/lib/page_data.dart
```diff
@@ -2,11 +2,16 @@
...
 
+import 'dart:convert';
+
 import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart';
 
 part 'page_data.g.dart';
 
+/// Codec to serialize the [PageData] JSON content on the HTML page.
+final pageDataJsonCodec = json.fuse(utf8).fuse(base64);
```
jonasfj: There is strictly no need for base64 encoding, but I'm okay with this :)

isoos: The non-base64 JSON string was longer, as all the `"` got escaped to `&quot;`, and it was ugly to look at it. This way, it is just a bunch of bytes :)

PR #2693, opened 2019-08-22
Refactoring admin authorization.

isoos:
- Hardened authentication flow.
- Added logging.
- Added admin audience.
- Admin authorization uses `userId` instead of `oauthUserId`

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -7,6 +7,7 @@ import 'dart:convert';
 
 import 'package:gcloud/db.dart';
 import 'package:gcloud/service_scope.dart' as ss;
+import 'package:logging/logging.dart';
 import 'package:neat_cache/neat_cache.dart';
 import 'package:retry/retry.dart';
 import 'package:uuid/uuid.dart';
@@ -18,6 +19,7 @@ import 'auth_provider.dart';
 import 'google_oauth2.dart' show GoogleOauth2AuthProvider;
 import 'models.dart';
 
+final _logger = Logger('account.backend');
 final _uuid = Uuid();
 
 /// Sets the account backend service.
@@ -61,6 +63,7 @@ class AccountBackend {
               <String>[
                 activeConfiguration.pubClientAudience,
                 activeConfiguration.pubSiteAudience,
+                activeConfiguration.adminAudience,
               ],
             );
 
@@ -177,7 +180,17 @@ class AccountBackend {
     if (auth == null) {
       return null;
     }
-    return await _lookupOrCreateUserByOauthUserId(auth);
+    final user = await _lookupOrCreateUserByOauthUserId(auth);
+    if (user.isDeleted) {
+      // This can only happen if the Google account had been deleted and then
+      // later re-enabled. We don't automatically re-enable such User accounts,
+      // as the new account might be a new owner on the same e-mail address
+      // (e.g. non-Google e-mails on GSuite).
+      _logger
+          .severe('Login on deleted account: ${user.userId} / ${user.email}');
+      throw StateError('Account had been deleted, login is not allowed.');
+    }
+    return user;
   }
 
   Future<User> _lookupOrCreateUserByOauthUserId(AuthResult auth) async {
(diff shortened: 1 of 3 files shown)


app/lib/shared/configuration.dart
```diff
@@ -138,14 +142,15 @@ class Configuration {
...
       credentials: _loadCredentials(),
       blockEmails: true,
       blockRobots: true,
       productionHosts: const ['dartlang-pub-dev.appspot.com'],
       primaryApiUri: Uri.parse('https://dartlang-pub-dev.appspot.com'),
       primarySiteUri: Uri.parse('https://dartlang-pub-dev.appspot.com'),
-      admins: [
-        '111042304059633250784', // isoos
+      adminUserIds: [
+        '6be21685-e61d-465b-ba60-083fb8eb99ad', // isoos
```
jonasfj: This is not an oauth2UserId... I think we should make this `adminOauthUserId`, such that it's hardcoded against an external identifier that's easy to lookup.

jonasfj: We could also use `email`, I'm a bit torn on that... `email` is easier to read, but `oauthUserId` feel stronger.

Hmm, I think the email for a service account can be reused... so I think `oauthUserId` is better.

app/lib/admin/backend.dart
```diff
@@ -35,7 +35,9 @@ class AdminBackend {
 
   Future<R> _withAdmin<R>(Future<R> fn(User user)) async {
     return await withAuthenticatedUser((user) async {
-      if (!activeConfiguration.admins.contains(user.oauthUserId)) {
+      if (!activeConfiguration.adminUserIds.contains(user.userId)) {
```
jonasfj: I think we use `oauthUserId`..

app/lib/account/backend.dart
```diff
@@ -192,7 +205,7 @@ class AccountBackend {
         final user = (await _db.lookup<User>([mapping.userIdKey])).single;
         // TODO: we should probably have some kind of consistency mitigation
         if (user == null) {
-          throw Exception('Incomplete OAuth userId mapping: '
+          throw StateError('Incomplete OAuth userId mapping: '
               'missing User (`${mapping.userId}`) referenced by `${mapping.id}`.');
```
jonasfj: Nice! :) We should be catching this.

app/lib/account/backend.dart
```diff
@@ -224,6 +236,16 @@ class AccountBackend {
         return updatedUser;
       }
 
+      // We've probably found a non-valid User state, it is better to fail the
+      // authentication. Let's investigate it before allowing the creation of
+      // the new User.
+      if (usersWithEmail.isNotEmpty) {
+        final userIds = usersWithEmail.map((u) => u.userId).join(', ');
+        _logger.severe(
+            'Probably User state violation while looking for ${auth.email}: $userIds');
```
jonasfj: I think it's okay to have two users with the same email, as long as they have different `ouathUserId` or one of them is deleted.

jonasfj: I think we just continue if this is the case. What are the cases where this can happen?

User account is deleted and new account with the same email is created. Isn't that the only case where this can happen?

If that's not the case, it's likely datastore inconsistency?

jonasfj: I am not sure here... On the one hand I'm tempted to throw an error and give up.

app/lib/account/backend.dart
```diff
@@ -176,7 +179,17 @@ class AccountBackend {
     if (auth == null) {
       return null;
     }
-    return await _lookupOrCreateUserByOauthUserId(auth);
+    final user = await _lookupOrCreateUserByOauthUserId(auth);
+    if (user.isDeleted) {
+      // This can only happen if the Google account had been deleted and then
+      // later re-enabled. We don't automatically re-enable such User accounts,
+      // as the new account might be a new owner on the same e-mail address
+      // (e.g. non-Google e-mails on GSuite).
```
jonasfj: No, this should only happen if there is a consistency issue, right?

Otherwise, the `oauthUserId` is stripped when `isDeleted` is set, and we never allow upgrading a `User` entity with `isDeleted == true`... right?

app/lib/account/backend.dart
isoos: I'd rather err on the side where we don't allow it first, and maybe allow it on a case-by-case basis.

app/lib/account/backend.dart
isoos: Ah, yeah, true.

app/lib/account/backend.dart
jonasfj: Is there any risk to creating a new `User` entity?

Clearly if emails can be reused and transferred between google accounts (as they can if they are not @gmail.com address), then this can happen, and is perfectly acceptable.

I think creating a new user is fine, that won't give you access to anything, right?

isoos: Removed.

PR #2695, opened 2019-08-22
Backfill User.isDeleted

isoos:


jonasfj:
Maybe add a note in the `CHANGELOG`, that we need to run this...

Issue #2618, opened 2019-08-02
Security page


mit-mit:
dart.dev now has a security page: https://dart.dev/security

I think we should add a similar page on pub.dev (linked from the pub.dev footer).

isoos:
Should we use the same e-mail address for reporting?

I guess the best practices section should be a bit more customized to the site. Do you have any text that you want to display there?

PR #2696, opened 2019-08-22
Security page

isoos:
- #2618 
- not much content yet, but easier to extend once we have a start on it

### DIFF
--- a/app/lib/frontend/templates/views/security.mustache
+++ b/app/lib/frontend/templates/views/security.mustache
@@ -4,13 +4,13 @@
 <div style="max-width: 600px">
   <h2>Security</h2>
   <p>
-      The Dart team takes the security of Pub and its published packages seriously.
+      The Dart team takes the security of pub.dev seriously.
       This page describes how to report any vulnerabilities you may find.
   </p>
 
   <h3>Reporting vulnerabilities</h3>
   <p>
-      In the rare event that you find a vulnerability in Pub site itself, email us at
+      In the rare event that you find a vulnerability in pub.dev, email us at
       <a href="mailto:product-security@google.com">product-security@google.com</a>.
       For more information about how Google handles security issues, see
       <a href="https://www.google.com/about/appsecurity/">Google’s security philosophy</a>.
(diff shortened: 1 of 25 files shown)


app/lib/frontend/templates/views/security.mustache
```diff
@@ -0,0 +1,18 @@
...
+      The Dart team takes the security of Pub and its published packages seriously.
+      This page describes how to report any vulnerabilities you may find.
+  </p>
+
+  <h3>Reporting vulnerabilities</h3>
+  <p>
+      In the rare event that you find a vulnerability in Pub site itself, email us at
+      <a href="mailto:product-security@google.com">product-security@google.com</a>.
+      For more information about how Google handles security issues, see
+      <a href="https://www.google.com/about/appsecurity/">Google’s security philosophy</a>.
```
jonasfj: @mit-mit, do you have any comments on the contents here?

If we want `product-security@google.com` handling, we probably have to do some setup.

app/lib/frontend/templates/views/security.mustache
```diff
@@ -0,0 +1,18 @@
+{{! Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+    for details. All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file. }}
+<div style="max-width: 600px">
+  <h2>Security</h2>
+  <p>
+      The Dart team takes the security of Pub and its published packages seriously.
```
mit-mit: I think we should be more general here: s/of Pub and its published packages/of Pub.dev/

app/lib/frontend/templates/views/security.mustache
mit-mit: I already did that (see cl/261665758)

app/lib/frontend/templates/views/security.mustache
```diff
@@ -0,0 +1,18 @@
...
+      This page describes how to report any vulnerabilities you may find.
+  </p>
+
+  <h3>Reporting vulnerabilities</h3>
+  <p>
+      In the rare event that you find a vulnerability in Pub site itself, email us at
```
mit-mit: s/Pub site itself/Pub.dev/

PR #2702, opened 2019-08-26
Fix cache initialization.

isoos:


app/lib/service/services.dart
```diff
@@ -100,6 +101,6 @@ Future<void> withPubServices(FutureOr<void> Function() fn) async {
     registerScopeExitCallback(searchClient.close);
     registerScopeExitCallback(searchService.close);
 
-    return await fn();
+    return await withCache(fn);
```
jonasfj: This is very likely to be too late... We might use cache in a constructor somewhere... Or when we start something...

Let's set it up right after storage and secrets?

isoos: We don't use cache in constructors, only when processing requests, that comes later. Nevertheless, I wanted to have a second refactor of the cache, I may just do it with this PR...

isoos: It will be a bit more involved, I'll do it separately.

PR #2703, opened 2019-08-26
Integration test for publisher upload.

isoos:


pkg/pub_integration/lib/script/publisher.dart
```diff
@@ -0,0 +1,95 @@
...
+      await Future.delayed(Duration(seconds: 1));
+      await _verifyDummyPkg(version: '2.0.0');
+    } finally {
+      await _temp.delete(recursive: true);
+      _pubHttpClient.close();
+      await _pubToolClient?.close();
+    }
+  }
+
+  Future _publishVersion(String version) async {
```
jonasfj: rename `_publishDummyPkgVersion` or `_publishDummyPkg` or give a documentation comment.

pkg/pub_integration/lib/script/publisher.dart
```diff
@@ -0,0 +1,95 @@
...
+    assert(_pubToolClient == null);
+    _pubHttpClient = PubHttpClient(pubHostedUrl);
+    _temp = await Directory.systemTemp.createTemp('pub-integration');
+    try {
+      _pubToolClient = await PubToolClient.create(
+          pubHostedUrl: pubHostedUrl,
+          credentialsFileContent: credentialsFileContent);
+
+      await _publishVersion('1.0.0');
+      await Future.delayed(Duration(seconds: 1));
```
jonasfj: If this is waiting for the server to be ready.. caches to be cleared, etc... why not do polling instead.

So instead of sleep and calling `_verifyDummyPkg`, we call `await _verifyDummyPkg(version: '1.0.0');` until it successful..
example:
```dart
await poll(
  () => _verifyDummyPkg(version: '1.0.0'),
  timeout: Duration(seconds: 10),
);
```
then in `utils.dart` or somewhere else we define:
```dart
Future<T> poll<T>(T Function() fn, {@required Duration timeout, Duration interval = const Duration(seconds: 2)}) async {
  // something about a timer..
  while(timer.elapsed > timeout) {
    try {
      return await fn();
    } on Exception {
      await Future.delayed(interval);
    }
  }
}
```

or something like that... The idea being that this is more robust...

pkg/pub_integration/lib/script/publisher.dart
```diff
@@ -0,0 +1,95 @@
...
+      );
+      await _pubHttpClient.setPackagePublisher(
+        authToken: 'user-at-example-dot-com',
+        package: '_dummy_pkg',
+        publisherId: 'example.com',
+      );
+
+      final publisherInfo =
+          await _pubHttpClient.getContent('/api/packages/_dummy_pkg/publisher');
+      if (publisherInfo != '{"publisherId":"example.com"}') {
```
jonasfj: Never assume how JSON is rendered... I don't think our API attempts to make any such promises.

pkg/pub_integration/lib/script/publisher.dart
isoos: It is not required at the moment, but I had plans to also include search in the integration tests, and that would take some time to happen. I think polling makes this a bit more complicated than needed, if it doesn't happen in 1 second, it should be considered not happening :)

PR #2707, opened 2019-08-26
Rename querystring parameters

jonasfj:


app/lib/admin/backend.dart
```diff
@@ -61,20 +61,19 @@ class AdminBackend {
       final query = _db.query<User>()..limit(limit);
 
       if (email != null) {
-        InvalidInputException.checkNull(oauthUserId, 'oauthUserId');
-        InvalidInputException.checkNull(continuationToken, 'continuationToken');
         query.filter('email =', email);
-      } else if (oauthUserId != null) {
-        InvalidInputException.checkNull(continuationToken, 'continuationToken');
```
isoos: if continuation token filter is set on email or on oauthUserId filter, it will fail, as we don't have the required index for it (because at the same time we order by on __key__ with continuation token). Let's keep these around.

jonasfj: Okay, that's fine...

Maybe we should document the reasoning behind this in a comment... well, maybe later..

PR #2708, opened 2019-08-26
Added support for querystring to api_builder..

jonasfj:
Based on the previous PR to refactor existing querystring parameters.


pkg/api_builder/lib/src/client_library_generator.dart
```diff
@@ -131,6 +131,9 @@ code.Method _buildClientMethod(
     returnTypeRef = code.refer('List<int>');
   }
 
+  final queryMap =
+      h.queryParameters.map((p) => '\'${p.name}\': ${p.name}').join(', ');
```
isoos: please add a comment that this is the serialized format of the query parameters map.

PR #2709, opened 2019-08-27
Fix 500s related to URL parsing.

isoos:
- `canonicalizeVersion` returns `null` when the version is not a valid versions string (e.g. `README` or `null`). In such cases the backend should return `null`, otherwise the Datastore lookup fails with an exception. (also added change to our fake datastore to match this behaviour).
- percent-decoding the Uri.path at the time of url sanitization.
- I've changed tests to use the wrappers and other sanitization in order to test the full processing chain at all times (at least for most frontend handler tests).

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -132,6 +132,9 @@ class PackageBackend {
   }
 
   /// Looks up a specific package version.
+  ///
+  /// Returns null if the version is not a semantic version or if the version
+  /// entity does not exists in the datastore.
   Future<models.PackageVersion> lookupPackageVersion(
       String package, String version) async {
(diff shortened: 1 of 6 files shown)


app/lib/package/backend.dart
```diff
@@ -135,6 +135,7 @@ class PackageBackend {
   Future<models.PackageVersion> lookupPackageVersion(
```
jonasfj: Please update documentation comment with why and when it returns `null`.

Whenever, I'm calling something that gets a value (but the value might not be there), it's awesome to know if it'll throw or return `null`.

jonasfj: (just a nit)

PR #2712, opened 2019-08-28
API changes in NameTracker

isoos:
I wanted to start on the caching logic, but it seems to be not that small change, so I'd just do this API change that reduces the surface area of `NameTracker`.


jonasfj:
Ship it :)

PR #2716, opened 2019-08-28
Create publisher page.

isoos:
Screen Shot 2019-08-28 at 18 06 25

### DIFF
--- a/app/lib/frontend/templates/views/publisher/create.mustache
+++ b/app/lib/frontend/templates/views/publisher/create.mustache
@@ -1,0 +1,28 @@
+{{! Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+    for details. All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file. }}
+
+<div style="max-width: 600px">
+  <h1>Create a publisher</h1>
+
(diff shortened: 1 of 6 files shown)


app/lib/frontend/templates/views/publisher/create.mustache
```diff
@@ -0,0 +1,28 @@
...
+    <p>
+      Only signed-in users can create a publisher. Please sign in to continue.
+    </p>
+  </div>
+
+  <div id="-admin-authenticated">
+    <div>
+      <p>
+        Publishers are identified by a domain name. To create a publisher,
+        you need to be a site administrator in
```
jonasfj:
```html
you need to be a <i>verified domain property</i> owner in
<a href="https://search.google.com/search-console/welcome" target="_blank" rel="noreferrer">Google Search Console</a>.
For help see <a href="https://support.google.com/webmasters/answer/34592?hl=en" target="_blank" rel="noreferrer">Add a website property</a>.

<p/><p>
Note: It is not sufficient to verify a <i>URL prefix property</i>, or to be added as a collaborator on a property somebody else verified. The user account used to create a publisher must have verified the <i>domain property</i>.
Once verified and a publisher has been created on pub.dev continued domain verification is not a requirement.  
```

PR #2720, opened 2019-08-30
Retry JobBackend operations.

isoos:


app/lib/job/backend.dart
```diff
@@ -86,57 +87,59 @@ class JobBackend {
...
-        if (hasNotChanged) {
-          if (!shouldProcess) {
-            // no reason to re-schedule the job
-            return;
-          }
-          if (current.state == JobState.available &&
-              current.lockedUntil == null) {
-            // already scheduled for processing
-            return;
+    await retry(() async {
```
jonasfj: Let's create a method `Future<T> retryWithTransaction<T>(Future<T> function(...)) async {....};` method...

I think both of these goes well together in the `JobBackend`, am I wrong?

Also, let's reduce the maximum delay to `2s` and limit what exceptions we retry on. We should not retry on `FormatExcpetion`, etc.. only network related ones, or transaction related ones..

isoos: Good idea and keeps the change surface smaller.
However, I think both the network and the transaction-related problems throw `Error` and not `Exception` :(

isoos: Nevertheless, we could catch that error and wrap it into an exception... it is a not very nice hack. How would you feel to update the `retry` interface instead?

jonasfj: > How would you feel to update the retry interface instead?

Package `retry` is used elsewhere too.. And it does the right thing. `Error` subclasses are not something you should catch, it's for programmer mistakes.

IMO, it's already sketchy to be retrying on all exceptions, but it's certainly better than retrying on both error and exception. We should never retry on error, if it's an Error that should be fixed upstream instead.

jonasfj: `class DatastoreError implements Exception {`
From `package:gcloud`, so I think these are `Exception` even if the type name says `XxxError`.

isoos: `DetailedApiRequestError` is a real `Error` though... `DatastoreError` seems to be good-enough to start with, updating...

jonasfj: Isn't it also often that `DetailedApiRequestError` is a 4xx error...

PR #2724, opened 2019-09-02
Cache publisher page.

isoos:
- #1518
- The auto-clearing cache may be a bit more than we actually need, but it has the benefit that the  potentially-mutating operations will clear the cache by default.

### DIFF
--- a/app/lib/publisher/backend.dart
+++ b/app/lib/publisher/backend.dart
@@ -14,6 +14,7 @@ import '../account/backend.dart';
 import '../account/consent_backend.dart';
 import '../shared/email.dart';
 import '../shared/exceptions.dart';
+import '../shared/redis_cache.dart' show cache;
 import 'domain_verifier.dart' show domainVerifier;
 
 import 'models.dart';
@@ -169,6 +170,7 @@ class PublisherBackend {
       return p;
     }) as Publisher;
 
+    await purgePublisherCache(publisherId);
     return _asPublisherInfo(p);
   }
 
@@ -262,6 +264,7 @@ class PublisherBackend {
       });
     }
     final updated = (await _db.lookup<PublisherMember>([key])).single;
+    await purgePublisherCache(publisherId);
     return await _asPublisherMember(updated);
   }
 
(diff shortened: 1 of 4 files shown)


app/lib/publisher/backend.dart
```diff
@@ -40,17 +41,6 @@ class PublisherBackend {
     return (await _db.lookup<Publisher>([pKey])).single;
   }
 
-  /// Checks whether the current authenticated user has admin role of the
-  /// publisher, and executes [fn] if it does.
-  /// Otherwise, it throws [AuthorizationException].
-  /// TODO: remove this and migrate to use withPublisherAdmin
-  Future<R> _withPublisherAdmin<R>(
```
jonasfj: Hmm, can we have `withPublisherAdmin` call `withAuthenticatedUser`? Or does it need to be separate, it seems like two wrappers always used together...

We could also change these to: `final u = await ensureAuthenticatedUser();` and `await ensurePublisherAdmin(u);` just to reduce indentation... I guess the with-pattern is a bit safer though...

app/lib/publisher/backend.dart
```diff
@@ -335,10 +354,14 @@ api.PublisherInfo _asPublisherInfo(Publisher p) =>
...
 ///
-/// Throws AuthenticationException if the user is provided.
+/// Throws AuthenticationException if the user is not provided.
 /// Throws AuthorizationException if the user is not an admin for the publisher.
 Future<R> withPublisherAdmin<R>(
-    String publisherId, String userId, FutureOr<R> fn(Publisher p)) async {
+  String publisherId,
+  String userId,
+  FutureOr<R> fn(Publisher p), {
+  bool clearCache = true,
```
jonasfj: Let's not mix concerns... Can't we make a `_purgeCache(String publisherId)` on the backend instead?

isoos: I've removed this from the PR, keeping only a separate `purgePublisherCache` method.

app/lib/publisher/backend.dart
isoos: Removed this from the PR, will do the other refactor separately.

PR #2725, opened 2019-09-02
Publisher admin page.

isoos:
- #1518
- only the barebones for the admin page (description field-only, lame layout)
- we'll need a similar endpoint that we have for packages (which tells us that the current user is an admin for the publisher), using the `.../members` listing endpoint for now.
- refactored UI widgets a bit: the decision to display the authorized/unathorized content seems to be the same across the widgets, let's not repeat that

### DIFF
--- a/pkg/web_app/lib/src/account.dart
+++ b/pkg/web_app/lib/src/account.dart
@@ -21,7 +21,7 @@ import 'tabs.dart';
 
 bool _initialized = false;
 GoogleUser _currentUser;
-bool _isPkgUploader = false;
+bool _isAdmin = false;
 
 /// Returns whether the user is currently signed-in.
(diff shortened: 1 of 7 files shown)


pkg/web_app/lib/src/account.dart
```diff
@@ -280,9 +308,43 @@ class _CreatePublisherWidget {
...
 
-  bool get isActive =>
-      _unauthenticatedRoot != null &&
-      _authenticatedRoot != null &&
-      _publisherIdInput != null &&
-      _createButton != null;
+  bool get isActive => _publisherIdInput != null && _createButton != null;
+}
+
+class _PublisherAdminWidget {
```
jonasfj: Could we write a doc-comment saying where this is active, this would make it slightly easier to track..

PR #2728, opened 2019-09-03
Entrypoints with command line arguments: --local-dev flag for the default frontend.

isoos:
- The `GAE_SERVICE` initialization is a bit of a hack, but it works. :)
- I want to expand the `local-dev` flag to also include monitoring and refreshing of static files, which would reduce some of the UI update pain.

### DIFF
--- a/app/lib/service/entrypoint/frontend.dart
+++ b/app/lib/service/entrypoint/frontend.dart
@@ -5,6 +5,7 @@
 import 'dart:async';
 import 'dart:math';
 
+import 'package:args/command_runner.dart';
 import 'package:http/http.dart' as http;
 import 'package:gcloud/db.dart' as db;
 import 'package:gcloud/service_scope.dart';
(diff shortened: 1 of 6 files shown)


app/lib/service/entrypoint/frontend.dart
```diff
@@ -35,12 +36,30 @@ import '_isolate.dart';
...
+  String get name => 'default';
+
+  @override
+  String get description => 'The default frontend service entrypoint.';
+
+  DefaultCommand() {
+    argParser.addFlag(
+      'local-dev',
+      defaultsTo: false,
+      help: 'Reduce the number of background services that are started.',
```
jonasfj: Add: `(Useful for local development)`

jonasfj:
uh, maybe update: https://github.com/dart-lang/pub-dev/blob/master/doc/development.md

Feel free to simplify it  :)

isoos:
Yes, `-h` is there with `CommandRunner`.

PR #2731, opened 2019-09-03
Less polling for name_tracker.

### LINKED ISSUE #2706: Add caching in the name-tracker
The name tracker queries for all names periodically, and this is expensive, furthermore it would probably be faster if we cached the list of all package names, and then purged it when a new package is published.

isoos:
Fixes #2706

### DIFF
--- a/app/lib/package/name_tracker.dart
+++ b/app/lib/package/name_tracker.dart
@@ -169,7 +169,7 @@ class _NameTrackerUpdater {
       if (_stopped) return;
       nameTracker.add(p.name);
     }
-    _lastTs = now.subtract(const Duration(minutes: 5));
+    _lastTs = now.subtract(const Duration(hours: 1));
   }
 
   void stop() {
--- a/app/lib/package/name_tracker.dart
+++ b/app/lib/package/name_tracker.dart
@@ -51,6 +51,10 @@ class NameTracker {
       _reducedNames.contains(reducePackageName(name));
 
   /// Whether to accept the upload attempt of a given package [name].
+  ///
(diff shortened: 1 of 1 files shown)


app/lib/package/name_tracker.dart
```diff
@@ -153,15 +160,16 @@ class _NameTrackerUpdater {
...
     final query = _db.query<Package>()..order('created');
     if (_lastTs != null) {
       query.filter('created >', _lastTs);
     }
     await for (Package p in query.run()) {
       if (_stopped) return;
       nameTracker.add(p.name);
-      _lastTs = p.created;
     }
+    _lastTs = now.subtract(const Duration(minutes: 5));
```
jonasfj: Nice touch... 5 minutes should be safe... maybe add a comment that this is to account for cases where the indexing in datastore is lagging.

Actually, maybe it's safer to make this 15 min... or even 1 hour... As indexing could be behind a bit... also machines could have clock drift... and the number of packages uploaded in an hour is going to be small regardless.

app/lib/package/name_tracker.dart
```diff
@@ -51,8 +51,15 @@ class NameTracker {
       _reducedNames.contains(reducePackageName(name));
 
   /// Whether to accept the upload attempt of a given package [name].
-  Future<bool> accept(String name) async =>
-      _hasPackage(name) || !_hasConflict(name);
+  Future<bool> accept(String name) async {
+    // fast track:
+    if (_hasPackage(name)) return true;
```
jonasfj: I don't understand this...

isoos: (updated doc)

PR #2733, opened 2019-09-03
Transfer to/from publisher on package admin page.

isoos:
#1518

### DIFF
--- a/app/lib/frontend/templates/views/pkg/admin_page.mustache
+++ b/app/lib/frontend/templates/views/pkg/admin_page.mustache
@@ -15,6 +15,34 @@
 <div id="-admin-authorized">
   <h2>Admin package</h2>
 
+    <h3>Publisher</h3>
+    {{^has_publisher}}
+    <div>
+      <p>
+        You can transfer your package to a publisher (if you are also an admin of the given publisher).
+        The current uploaders will lose access over the package, and only the publisher's administrators
+        will be able to upload new versions.
+      </p>
+      <p>
+        <input id="-admin-set-publisher-input" class="pub-input" placeholder="publisher e.g. example.com"/>
+        <button id="-admin-set-publisher-button" class="pub-button">Transfer to publisher</button>
+      </p>
+    </div>
+    {{/has_publisher}}
+    {{#has_publisher}}
+    <div>
+      <p>
+        You can transfer your package away from the current publisher (<code>{{publisher_id}}</code>).
+        The publisher's current administrators will lose access over the package, and you will become
+        the single authorized uploader.
+      </p>
+      <p>
+        <button id="-admin-remove-publisher-button" class="pub-button">Remove from publisher</button>
+      </p>
+    </div>
+    {{/has_publisher}}
+
+    <h3>Discontinued</h3>
     {{^is_discontinued}}
     <div>
       <p>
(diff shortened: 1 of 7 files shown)


app/lib/frontend/templates/views/pkg/admin_page.mustache
```diff
@@ -15,6 +15,34 @@
...
+      <p>
+        You can transfer your package away from the current publisher (<code>{{publisher_id}}</code>).
+        The publisher's current administrators will lose access over the package, and you will become
+        the single authorized uploader.
+      </p>
+      <p>
+        <button id="-admin-remove-publisher-button" class="pub-button">Remove from publisher</button>
+      </p>
+    </div>
+    {{/has_publisher}}
```
jonasfj: Maybe we should have both div's and switch between them in javascript instead? Just to keep our template logic simple.

I guess whether it has a publisher or not is public information... I'm just thinking it would be easier to debug the JS if the admin page always had the same HTML structure..

This is a nit, I think this is also fine.. Just a few thougths..

app/lib/frontend/templates/views/pkg/admin_page.mustache
```diff
@@ -15,6 +15,34 @@
...
+    {{#has_publisher}}
+    <div>
+      <p>
+        You can transfer your package away from the current publisher (<code>{{publisher_id}}</code>).
+        The publisher's current administrators will lose access over the package, and you will become
+        the single authorized uploader.
+      </p>
+      <p>
+        <button id="-admin-remove-publisher-button" class="pub-button">Remove from publisher</button>
```
jonasfj: I think we should allow transfering to a different publisher. But never transfer back to using `uploaders`.

app/lib/frontend/templates/views/pkg/admin_page.mustache
```diff
@@ -15,6 +15,34 @@
 <div id="-admin-authorized">
   <h2>Admin package</h2>
 
+    <h3>Publisher</h3>
+    {{^has_publisher}}
+    <div>
+      <p>
+        You can transfer your package to a publisher (if you are also an admin of the given publisher).
+        The current uploaders will lose access over the package, and only the publisher's administrators
```
jonasfj: s/`publisher's administrator`/`publisher's members`

pkg/client_data/lib/page_data.dart
```diff
@@ -33,15 +33,17 @@ class PageData {
 }
 
 /// The server-provided data about the current package.
-@JsonSerializable()
+@JsonSerializable(includeIfNull: false)
 class PkgData {
   final String package;
   final String version;
+  final String publisherId;
```
jonasfj: How about giving this a documentation comment that says:
```dart
/// PublisherId of publisher that owns this package, `null` if the package isn't owned by a publisher.
```

If something can be `null`, it's always nice to explain when this is the case.

isoos:
PTAL about the message on the UI.

app/lib/frontend/templates/views/pkg/admin_page.mustache
```diff
@@ -15,6 +15,35 @@
...
+        You can transfer your package to a publisher
+        (if you are also an admin of the given publisher).
+        The current uploaders will lose access over the package,
+        and only the publisher's members  will be able to upload
+        new versions.
+      </p>
+      <p>
+        <strong>This is an irreversible operation.</strong>
+        Packages can be transferred between publishers, but can't
+        be used as standalone packages in the future.
```
jonasfj: > be converted back to legacy-ownership per-package _uploaders_.

PR #2737, opened 2019-09-04
Publisher admin UI: invite new member.

isoos:


pkg/web_app/lib/src/account.dart
```diff
@@ -379,9 +386,36 @@ class _PublisherAdminWidget {
...
+      window.alert('Please specify a valid e-mail.');
+      return;
+    }
+    if (!window.confirm('Are you sure you want to invite "$email" as an '
+        'administrator member to this publisher?')) {
+      return;
+    }
+    final rq = InviteMemberRequest(email: email);
+    final rs = await client.post(
+      '/api/publishers/${pageData.publisher.publisherId}/invite-member',
```
jonasfj: It doesn't have to be in this PR.. but why don't we steal the generated client code? Either make code-gen write it somewhere else... or just import it from the server package... I know it's a hack, but it works and we'll eventually want to have this figured out anyways.

Notably, this gives us better exceptions, and it allows us to unify retry handling, exception handling... then we can catch unhandled exceptions in our js client and possibly convert those into messages we can flash on the screen...that's all further out..

jonasfj: This is not really a comment on this PR :)

isoos: ++, will take a look tomorrow

PR #2740, opened 2019-09-05
Caching PackageView

isoos:
PackageView is the compact snippet, with displayable information for the search result (listing) pages, as we'll be using it on the publisher pages too, it makes sense to cache it.

### DIFF
--- a/app/lib/package/search_service.dart
+++ b/app/lib/package/search_service.dart
@@ -93,16 +93,19 @@ class SearchService {
 
   /// Returns the [PackageView] instance for each package in [packages], using
   /// the latest stable version.
-  Future<List<PackageView>> getPackageViews(Iterable<String> packages) async {
+  Future<List<PackageView>> getPackageViews(List<String> packages) async {
     // TODO: consider a cache-check first and batch-load the rest of the packages
     return await Future.wait(packages.map((p) => getPackageView(p)));
   }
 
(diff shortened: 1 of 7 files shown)


app/lib/scorecard/backend.dart
```diff
@@ -219,7 +219,12 @@ class ScoreCardBackend {
       await tx.commit();
     });
 
-    await cache.scoreCardData(packageName, packageVersion).purge();
+    final isLatest = package.latestVersion == version.version;
+    await Future.wait([
+      cache.scoreCardData(packageName, packageVersion).purge(),
+      if (isLatest) cache.uiPackagePage(packageName, null).purge(),
```
jonasfj: What about `cache.uiPackagePage(packageName, version.version).purge()` ? (just wondering)

app/lib/package/search_service.dart
```diff
@@ -75,48 +74,46 @@ class SearchService {
 
   Future close() async {}
 
+  /// Returns the [PackageView] instance for [package] on its latest stable version.
+  Future<PackageView> getPackageView(String package) async {
```
jonasfj: Should we consider making this `getPackageViews(List<String> packages)` to ensure that can still do batch loading?

isoos: We do have and user `getPackageViews`, but not in a batching mode (yet, see TODO comment below). I'm making this private so that we don't encourage its use.

PR #2743, opened 2019-09-06
Refactor tabs rendering.

isoos:
This will be reused on the publisher page.


app/lib/frontend/templates/tabs.dart
```diff
@@ -0,0 +1,73 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'package:meta/meta.dart';
+
+import '_cache.dart';
+
+/// Renders the `views/pkg/tabs.mustache` template.
+String renderPkgTabs(List<Tab> tabs) {
```
jonasfj: Maybe rename it `renderTabs`, or even:`Tab.render(List<Tab> tabs)` as a static method.

app/lib/frontend/templates/tabs.dart
```diff
@@ -0,0 +1,73 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'package:meta/meta.dart';
+
+import '_cache.dart';
+
+/// Renders the `views/pkg/tabs.mustache` template.
```
jonasfj: Update the comment to have the correct path.

app/lib/frontend/templates/tabs.dart
```diff
@@ -0,0 +1,73 @@
...
+    String title,
+    String titleHtml,
+    @required String href,
+    this.isHidden = false,
+  })  : titleHtml =
+            '<a href="$href">${titleHtml ?? htmlEscape.convert(title)}</a>',
+        contentHtml = null,
+        isMarkdown = false;
+
+  Map toMustacheData() {
```
jonasfj: Why is this method exported? It could even be inline in `renderPkgTabs`...

PR #2752, opened 2019-09-09
Publisher admin: websiteUrl and contactEmail

isoos:
- added new fields on the UI
- added `websiteUrl` to API objects
- handling `websiteUrl` in creation and update. (requiring `https://[domain]/` prefix for changes)

### DIFF
--- a/app/lib/frontend/templates/views/publisher/admin_page.mustache
+++ b/app/lib/frontend/templates/views/publisher/admin_page.mustache
@@ -15,9 +15,17 @@
 <div id="-admin-authorized">
   <h2>Admin publisher: {{publisher_id}}</h2>
   <h4>Description</h4>
-  <p>
+  <div>
     <textarea id="-publisher-description" class="pub-input" rows="16">{{description}}</textarea>
-  </p>
+  </div>
+  <h4>Website URL</h4>
+  <div>
+    <input id="-publisher-website-url" class="pub-input" value="{{website_url}}" />
+  </div>
+  <h4>Contact e-mail</h4>
+  <div>
+    <input id="-publisher-contact-email" class="pub-input" value="{{contact_email}}" />
+  </div>
   <p>
       <button id="-publisher-update-button" class="pub-button">Update</button>
   </p>
--- a/app/lib/publisher/backend.dart
+++ b/app/lib/publisher/backend.dart
@@ -160,18 +160,23 @@ class PublisherBackend {
       final key = _db.emptyKey.append(Publisher, id: publisherId);
       final p = (await tx.lookup<Publisher>([key])).single;
 
+      // If websiteUrl has changed, check that it's under the [publisherId] domain.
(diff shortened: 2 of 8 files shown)


app/lib/frontend/templates/views/publisher/admin_page.mustache
```diff
@@ -15,9 +15,17 @@
...
     <textarea id="-publisher-description" class="pub-input" rows="16">{{description}}</textarea>
-  </p>
+  </div>
+  <h4>Website URL</h4>
+  <div>
+    <input id="-publisher-website-url" class="pub-input" value="{{website_url}}" />
+  </div>
+  <h4>Contact e-mail</h4>
+  <div>
+    <input id="-publisher-contact-email" class="pub-input" value="{{contact_email}}" />
```
jonasfj: Shouldn't we be using `<form>` and `<label for='-publisher-contact-email'>Contact e-mail<label/>`, instead of a `<h4>`.

app/lib/publisher/backend.dart
```diff
@@ -160,6 +160,18 @@ class PublisherBackend {
       final key = _db.emptyKey.append(Publisher, id: publisherId);
       final p = (await tx.lookup<Publisher>([key])).single;
 
+      if (update.websiteUrl != null && p.websiteUrl != update.websiteUrl) {
+        final parsedUrl = Uri.tryParse(update.websiteUrl);
+        final isValid = parsedUrl != null &&
+            parsedUrl.isAbsolute &&
+            parsedUrl.toString() == update.websiteUrl;
```
jonasfj: Are you sure this is a good idea? the ordering of query params might change? I haven't tested if this is an identity function for all valid URLs.

You could also test the `scheme == https` and the `host == publisherId`... and `port` not specified.

app/lib/publisher/backend.dart
```diff
@@ -160,6 +160,18 @@ class PublisherBackend {
       final key = _db.emptyKey.append(Publisher, id: publisherId);
       final p = (await tx.lookup<Publisher>([key])).single;
 
+      if (update.websiteUrl != null && p.websiteUrl != update.websiteUrl) {
```
jonasfj: Please give this section a comment:
```dart
// If websiteUrl has changed, check that it's under the [publisherId] domain.
if (update.websiteUrl != null && p.websiteUrl != update.websiteUrl) {
```

app/lib/frontend/templates/views/publisher/admin_page.mustache
isoos: Maybe, but form also implies ability to send it as a separate request, and we are not doing that, as we need to augment the request with the auth header, so it won't work without JS anyway.

app/lib/publisher/backend.dart
isoos: I'll add those checks too. The equality check filters out input urls that are translated to escaped characters, and I thought that requiring immutability is not too bad at this point. We can relax it if the need arises.

PR #2766, opened 2019-09-12
Publisher: packages list on main page.

isoos:
- `/publishers/<id>` - "Top Packages"
- `/publishers/<id>/about` - "About" - long description.
- `/publishers/<id>/admin` - "Admin"

### DIFF
--- a/app/lib/frontend/templates/publisher.dart
+++ b/app/lib/frontend/templates/publisher.dart
@@ -4,6 +4,7 @@
 
 import 'package:client_data/page_data.dart';
 
+import '../../package/models.dart' show PackageView;
 import '../../publisher/models.dart' show Publisher;
 import '../../shared/markdown.dart';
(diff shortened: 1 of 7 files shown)


app/lib/frontend/templates/publisher.dart
```diff
@@ -132,3 +161,22 @@ String renderPublisherAdminPage(Publisher publisher) {
...
 }
+
+Tab _packagesLinkTab(String publisherId) => Tab.withLink(
+      id: 'packages',
+      title: 'Top Packages',
+      href: urls.publisherUrl(publisherId),
+    );
+
+Tab _aboutLinkTab(String publisherId) => Tab.withLink(
+      id: 'about',
```
jonasfj: I'm not sure we need this about page... but maybe that's something we discuss with mit when we get further..

PR #2772, opened 2019-09-13
Publisher metadata (created time) on detail header.

isoos:
I've tried a few things, and I think instead of putting 'Publisher:' in the title, putting it in the metadata is better. Also the 'created on' felt like that the domain was registered on that date, not sure if the current registered on feels better, but maybe a little bit.

### DIFF
--- a/app/lib/frontend/templates/publisher.dart
+++ b/app/lib/frontend/templates/publisher.dart
@@ -2,12 +2,15 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:convert';
+
 import 'package:client_data/page_data.dart';
 
 import '../../package/models.dart' show PackageView;
 import '../../publisher/models.dart' show Publisher;
 import '../../shared/markdown.dart';
 import '../../shared/urls.dart' as urls;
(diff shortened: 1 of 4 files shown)


app/lib/frontend/templates/publisher.dart
```diff
@@ -162,6 +165,14 @@ String renderPublisherAdminPage(Publisher publisher) {
   );
 }
 
+String _renderDetailHeader(Publisher publisher) {
+  final shortCreated = shortDateFormat.format(publisher.created);
+  return renderDetailHeader(
+    title: publisher.publisherId,
+    metadataHtml: htmlEscape.convert('Publisher registered on $shortCreated.'),
```
jonasfj: It could also say: `Created: $shortCreated`? But I'm not sure it's a better option...

Maybe `Publisher created: $shortCreated`?

isoos: My issue with the "created" is the following: when I visit the publisher page, and I see that a publisher was created a few months ago, I may infer that all of their packages are younger than that. Using "registered" may shift that perception a bit, because it implies (at least for me) that the packages are not necessarily younger than that date.

jonasfj: ah, over time I don't think that'll matter much... but good point.

PR #2774, opened 2019-09-16
RPC-wrapper for the web-app.

isoos:
- handles button and input disable/enable around RPC calls
- handles error messages consistently
- updated CSS style for the disabled button and input

### DIFF
--- a/pkg/web_app/lib/src/account.dart
+++ b/pkg/web_app/lib/src/account.dart
@@ -232,13 +232,10 @@ class _PkgAdminWidget {
     }
     final options =
         PkgOptions(isDiscontinued: !pageData.pkgData.isDiscontinued);
-    try {
+    await _rpc(() async {
       await client.setPackageOptions(pageData.pkgData.package, options);
       window.location.reload();
-    } on RequestException catch (e) {
-      final map = e.bodyAsJson();
-      window.alert(map['message'] as String);
-    }
+    });
   }
 
   Future _setPublisher() async {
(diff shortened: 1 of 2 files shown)


pkg/web_app/lib/src/account.dart
```diff
@@ -516,29 +497,62 @@ class _ConsentWidget {
...
+  inputs.forEach((e) => e.disabled = true);
+  try {
+    return await fn();
+  } on RequestException catch (e) {
+    final message = _requestExceptionMessage(e) ?? 'Unexpected error: $e';
+    // TODO: render as a modal alert window
+    // TODO: render as HTML
+    window.alert(message);
+    rethrow;
+  } finally {
```
jonasfj: Let's at-least add a TODO to handle network errors, or other errors the browser may throw... We could certainly display a generic error that we had connectivity issues :)

pkg/web_app/lib/src/account.dart
```diff
@@ -516,29 +497,62 @@ class _ConsentWidget {
...
-      window.alert(map['message'] as String);
-    }
+    });
   }
 
   bool get isActive => _content != null && _buttons != null;
 }
+
+Future<R> _rpc<R>(Future<R> fn()) async {
+  // Disable only the selected Elements that are not disabled at the moment.
```
jonasfj: This doesn't only disable the _selected elements_ :)

This disables all input input elements that are not already disabled..

So you can't edit the description while the "send invite" operation is running...
I think this is fine... I'm not 100% sure.. but let's take it for a spin.

Maybe we want any operation that submits data, like "send invite" or "update description" to remove the entire tab... and run with a spinner being displayed... When completed then either an error is showed on the current page, or the user is redirect to a different page where the change is reflected... Or in the case of "send invite" a text says "invite sent" and a button says "continue", which then reloads the current page.

isoos: I'm planning to do a modal dialog for error messages that would be in the front of the page. We could use the same thing for communicating that "something is ongoing, please wait".

PR #2775, opened 2019-09-16
Link to the search page from the publisher infobox.

isoos:


app/lib/frontend/templates/publisher.dart
```diff
@@ -98,6 +98,7 @@ String _renderPublisherInfoBox(Publisher publisher) {
     'publisher_id': publisher.publisherId,
     'website_url': publisher.websiteUrl,
     'contact_email': publisher.contactEmail,
+    'search_link': urls.searchUrl(q: 'publisher:${publisher.publisherId}'),
```
jonasfj: Let's call this `list_packages_search_link` instead of `search_link`.

PR #2779, opened 2019-09-16
UI updates: markdown, modal popups.

isoos:


isoos:
(deploying to staging)

pkg/web_css/lib/src/_base.scss
```diff
@@ -485,3 +490,39 @@ pre > code {
...
   text-align: center;
 }
+
+.pub-modal {
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  top: 0;
+  background: rgba(0, 0, 0, 0.7);
```
jonasfj: Lovely! :)

nit: what about `background: white`, and then `opacity: 0.7` and `transition: opacity 80ms`, just to give a smooth feel :)

pkg/web_app/lib/src/_dom_helper.dart
```diff
@@ -2,9 +2,65 @@
...
+            ]
+        ],
+    ];
+  elem.onClick.listen((_) {
+    if (!completer.isCompleted) {
+      completer.complete(false);
+    }
+  });
+  document.body.append(elem);
+  completer.future.then((_) {
```
jonasfj: Can we set focus on the button here?
And maybe... handle onKey escape on the too?

pkg/web_css/lib/src/_base.scss
isoos: Maybe... but if we also want to use this or a similar thing for the "please wait" indication, then it may not be good to have the transition there. Let's postpone this a bit.

PR #2782, opened 2019-09-17
Link to /create-publisher in the top nav.

isoos:


app/lib/frontend/templates/views/layout.mustache
```diff
@@ -76,6 +76,14 @@
             <a class="link" target="_blank" rel="noopener" href="{{& dart_site_root}}/tools/pub">Pub tool</a>
           </div>
         </div>
+        {{#show_publisher_nav_menu}}
+        <div class="sub-wrap hoverable">
+          <button class="button">Publisher</button>
```
jonasfj: I think we should move it into the menu when logged in..

PR #2785, opened 2019-09-17
use 800px page width for standalone pages

jonasfj:
I'm not sure why we wouldn't do this...


isoos:
I've found it weird to look at the full width text of the help page, and I think it is easer to read it if you have not too many characters in a single line.

PR #2787, opened 2019-09-17
Do not display login is required until, we have checked if login is done

jonasfj:


isoos:
This may be a bit better if we had also a visual indication that something is happening, because with this change it will be an empty page first. E.g. instead of the "You need to be logged in" we should display: "Please wait until we decide if you are logged in."

Note: I think the above "Please wait until we decide if you are logged in." is lame, and we should do it :). It is just feels like we should display something, and in the lack of other indication, the "You need to be logged in" seems to be better.

isoos:
Approved, assuming it helps you to fix the UX flow we've discussed.

jonasfj:
I agree, this is not ideal... Next step is to cache the state and image in `sessionStorage` or `localStorage`, and then use cached state until we've verified login.

Note: We should not cache the `id_token` or `access_token`.

PR #2789, opened 2019-09-17
Publisher list page.

isoos:
- reusing the package list styles for now (lot of room for improvement)
- added the link to the top nav dropdown menu
- added simple verification in `pub_integration`


app/lib/frontend/templates/views/publisher/publisher_list.mustache
```diff
@@ -0,0 +1,28 @@
...
+<h2>Publishers</h2>
+
+{{#has_publishers}}
+<ul class="package-list">
+  {{#publishers}}
+  <li class="list-item">
+    <h3 class="title"><a href="{{& url}}">{{publisher_id}}</a></h3>
+    <p class="metadata">
+      Registered on {{short_created}}.
+    </p>
```
jonasfj: Showing number of members and number packages would be too expensive, right?

isoos: We may do it, depending on where we put the aggregates.

PR #2792, opened 2019-09-18
Note that the description is in markdown

jonasfj:


app/lib/frontend/templates/views/publisher/admin_page.mustache
```diff
@@ -13,7 +13,7 @@
 </div>
 
 <div id="-admin-authorized">
-  <h4>Description</h4>
+  <h4>Description (markdown)</h4>
```
sigurdm: Do you want to link to a description of markdown?

jonasfj: Not here... maybe when we do this materialized...

PR #2793, opened 2019-09-18
Remove merge_users.dart tool

isoos:


app/lib/history/models.dart
```diff
@@ -290,18 +258,6 @@ class UploaderInvited implements HistoryEvent {
   }
 
   Map<String, dynamic> toJson() => _$UploaderInvitedToJson(this);
-
-  @override
-  HistoryEvent migrateUser(
```
jonasfj: Let's retain this constructor and document it deprecated, leave a comment that we're only keeping for documenting the existing datastore..

PR #2794, opened 2019-09-18
New History events.

isoos:


app/lib/history/models.dart
```diff
@@ -36,16 +37,23 @@ class History extends db.ExpandoModel {
...
       event: event,
     );
   }
 
-  @db.StringProperty(required: true)
+  @db.StringProperty()
+  String publisherId;
+
+  @db.StringProperty()
   String packageName;
```
jonasfj: Doc string that says this is `null` if it's not package related (ie. publisher related). Maybe the same for `publisherId`.

PR #2796, opened 2019-09-19
Material dialogs, textbox and button for create-publisher

jonasfj:
I'm sure this also makes the website heavier.. But that just means we can optimize later :)


app/lib/frontend/templates/views/layout.mustache
```diff
@@ -31,9 +31,11 @@
   <meta name="description" content="{{& pageDescription}}" />
   <meta property="og:description" content="{{& pageDescription}}" />
   <link rel="alternate" type="application/atom+xml" title="Updated Packages Feed for Pub" href="/feed.atom" />
+  <link href="https://unpkg.com/material-components-web@latest/dist/material-components-web.min.css" rel="stylesheet" type="text/css" />
```
isoos: While the unpkg.com seems to be okay, for versioning and also to reduce third-party risk, we should download these files and have them referenced locally.

pkg/web_css/lib/src/_account.scss
```diff
@@ -25,3 +25,7 @@
 #-account-profile {
   display: none;
 }
+
+#-admin-create-publisher {
+  margin-left: 20px;
```
isoos: I think if we want to have that left-margin, we should have it for all of the admin buttons, don't we?

jonasfj: Not clear, it depends on the layout...

isoos:
You'll need to regenerate tempaltes_test's golden files.

isoos:
I think you'll need to rebase first and then regenerate the files.

PR #2798, opened 2019-09-20
New sitemap for publishers.

isoos:
- Future plans:
  - one-hour caching for both sitemaps
  - moving packages to `/sitemap/packages.txt` (may keep only static pages in the `/sitemap.txt`, maybe moving that too inside `/sitemap/` directory).
  - maybe sharding (with some kind of `packages-[postfix].txt`

### DIFF
--- a/static/robots.txt
+++ b/static/robots.txt
@@ -1,5 +1,3 @@
 User-agent: *
 Disallow: /experimental/
 Sitemap: https://pub.dev/sitemap.txt
-# TODO: enable this after publishers are launched
-# Sitemap: https://pub.dev/sitemap/publishers.txt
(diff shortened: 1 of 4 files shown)


static/robots.txt
```diff
@@ -1,3 +1,5 @@
 User-agent: *
 Disallow: /experimental/
 Sitemap: https://pub.dev/sitemap.txt
+# TODO: enable this after publishers are launched
+# Sitemap: https://pub.dev/sitemap/publishers.txt
```
jonasfj: Can we maybe just make this `https://pub.dev/sitemap-1.txt`  and `https://pub.dev/sitemap-2.txt`.

I mean, this is not an API anyone should be using. It's only for URL discovery. We don't want people using this to guess what publishers/packages exist.

PR #2799, opened 2019-09-20
SessionUserId

isoos:
Just a quick proposal, similar to what we have for `OAuthUserId`.

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -265,4 +265,42 @@ class AccountBackend {
 
     return user;
   }
+
+  /// Creates a new session for the current authenticated user.
+  Future<String> createNewSession() async {
+    final user = await requireAuthenticatedUser();
+    final now = DateTime.now().toUtc();
+    final sessionId = _uuid.v4().toString();
+    await _db.commit(inserts: [
+      SessionUserId()
+        ..id = sessionId
+        ..userIdKey = user.key
+        ..created = now
+        ..expires = now.add(Duration(days: 7)),
+    ]);
+  }
+
+  /// Returns the userId associated with the [sessionId] or null if it does not
+  /// exists.
+  Future<String> lookupSessionOwner(String sessionId) async {
+    /// TODO: use local or redis cache
+    final key = _db.emptyKey.append(SessionUserId, id: sessionId);
+    final list = await _db.lookup<SessionUserId>([key]);
+    final session = list.single;
+    if (session == null) {
+      return null;
+    }
+
+    if (session.isExpired()) {
+      await _db.commit(deletes: [key]);
+      // TODO: clear cache (if used)
+      return null;
+    }
+
+    // TODO: decide about extending the expiration time (maybe asynchronously)
+    return session.userId;
+  }
+
+  // TODO: periodically remove expired sessions from datastore and cache
+  // TODO: expire all sessions of a given user from datastore and cache
 }
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -8,6 +8,7 @@ import 'dart:convert';
 import 'package:gcloud/db.dart';
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
 import 'package:neat_cache/neat_cache.dart';
 import 'package:retry/retry.dart';
 import 'package:uuid/uuid.dart';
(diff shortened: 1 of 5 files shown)


app/lib/account/models.dart
```diff
@@ -63,6 +63,25 @@ class OAuthUserID extends db.ExpandoModel {
   String get userId => userIdKey.id as String;
 }
 
+/// Maps the session id (from cookie) to User.id
+class SessionUserId extends db.ExpandoModel {
+  /// Same as [id].
+  String get sessionId => id as String;
```
jonasfj: State that this is a uuid v4

app/lib/account/models.dart
```diff
@@ -63,6 +63,25 @@ class OAuthUserID extends db.ExpandoModel {
...
+  db.Key userIdKey;
+
+  String get userId => userIdKey.id as String;
+
+  @db.DateTimeProperty(required: true)
+  DateTime created;
+
+  @db.DateTimeProperty(required: true)
+  DateTime expires;
+
```
jonasfj: How about image and email...

app/lib/account/backend.dart
```diff
@@ -265,4 +265,42 @@ class AccountBackend {
...
+        ..userIdKey = user.key
+        ..created = now
+        ..expires = now.add(Duration(days: 7)),
+    ]);
+  }
+
+  /// Returns the userId associated with the [sessionId] or null if it does not
+  /// exists.
+  Future<String> lookupSessionOwner(String sessionId) async {
+    /// TODO: use local or redis cache
```
jonasfj: Redis cache is important...

app/lib/account/models.dart
isoos: My first thinking was that we can cache those on the client side instead, but you are right, we should do the caching here.

app/lib/account/backend.dart
```diff
@@ -265,4 +267,51 @@ class AccountBackend {
...
+    final now = DateTime.now().toUtc();
+    final sessionId = _uuid.v4().toString();
+    await _db.commit(inserts: [
+      UserSession()
+        ..id = sessionId
+        ..userIdKey = user.key
+        ..email = user.email
+        ..imageUrl = imageUrl
+        ..created = now
+        ..expires = now.add(Duration(days: 7)),
```
jonasfj: I'm okay with 14 days.

app/lib/account/backend.dart
```diff
@@ -265,4 +267,51 @@ class AccountBackend {
 
     return user;
   }
+
+  /// Creates a new session for the current authenticated user.
```
jonasfj:
```dart
/// And returns the `sessionId`.
/// 
/// The `sessionId` is a secret that will be stored in a secure cookie.
/// Presence of this sessionId in a cookie, can only be used to authorize user specific content to be embedded
/// in HTML pages (such pages must have `Cache-Control: private`, and may not be cached in server-side). 
/// JSON APIs whether fetching data or updating data cannot be authorized with a cookie carrying the `sessionId`.
```

Maybe this should be written somewhere else... Ideally in the place where HTTP request handlers lookup the current session. Like `requiredAuthenticatedUser` we'll probably need something similar for sessions (though that probably just returns `null` instead of throwing an exception.

app/lib/account/backend.dart
```diff
@@ -265,4 +267,51 @@ class AccountBackend {
...
+        ..created = now
+        ..expires = now.add(Duration(days: 7)),
+    ]);
+    return sessionId;
+  }
+
+  /// Returns the user session associated with the [sessionId] or null if it
+  /// does not exists.
+  Future<UserSessionData> lookupSession(String sessionId) async {
+    final cacheEntry = cache.userSessionData(sessionId);
```
jonasfj: nit: `ArgumentError.checkNotNull(sessionId, 'sessionId');`

app/lib/shared/redis_cache.dart
```diff
@@ -31,6 +32,17 @@ class CachePatterns {
   // NOTE: This class should only contain methods that return Entry<T>, as well
   //       configuration options like prefix and TTL.
 
+  /// Cache for [UserSessionData].
+  Entry<UserSessionData> userSessionData(String sessionId) => _cache
+      .withPrefix('account-usersession')
+      .withTTL(Duration(hours: 1))
```
jonasfj: Why not just give it 24 hours... I doubt it'll take up much space :)

app/lib/account/models.dart
```diff
@@ -63,6 +66,68 @@ class OAuthUserID extends db.ExpandoModel {
...
+  bool isExpired() => DateTime.now().isAfter(expires);
+}
+
+/// The cacheable version of [UserSession].
+@JsonSerializable()
+class UserSessionData {
+  final String sessionId;
+  final String userId;
```
jonasfj: What is this? Can we document some of these... just a single comment for each.. 

SessionId is a random UUIDv4, and stuff like that... In particular highlight if they can be `null` or not..

app/lib/account/models.dart
```diff
@@ -63,6 +66,68 @@ class OAuthUserID extends db.ExpandoModel {
...
+  final String email;
+  final String imageUrl;
+  final DateTime created;
+  final DateTime expires;
```
jonasfj: Let's state what this means, as it's rather important :)

PR #2810, opened 2019-09-24
Login menu: styles and menu items.

isoos:
- removed top nav menu
- added "My *" links to account menu
- updated styles

### DIFF
--- a/app/lib/frontend/templates/views/layout.mustache
+++ b/app/lib/frontend/templates/views/layout.mustache
@@ -76,15 +76,6 @@
             <a class="link" target="_blank" rel="noopener" href="{{& dart_site_root}}/tools/pub">Pub tool</a>
           </div>
         </div>
-        {{#show_publisher_nav_menu}}
-        <div class="sub-wrap hoverable">
-          <button class="button">Publishers</button>
-          <div class="sub-nav">
-              <a class="link" href="/publishers">Browse publishers</a>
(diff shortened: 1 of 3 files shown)


app/lib/frontend/templates/views/layout.mustache
```diff
@@ -76,24 +76,20 @@
...
         <a id="-account-login" class="link">Login</a>
         <div id="-account-profile" class="sub-wrap hoverable">
           <img class="profile-img" id="-account-profile-img" />
           <div class="sub-nav sub-nav-right">
-            <div id="-account-profile-email"></div>
-            <a id="-account-logout" class="link">Logout</a>
+            <div class="account-nav-info"><span id="-account-profile-email"></span></div>
+            <div class="account-nav-divider"></div>
+            <div class="account-nav-link"><a href="/account/packages">My packages</a></div>
+            <div class="account-nav-link"><a href="/account/publishers">My publishers</a></div>
```
jonasfj: We haven't introduced these yet, right?
Can't we do `/user/packages`, `/my-packages´, `/~packages`, `/owned-packages`, I'm not sure here, @sigurdm?

Maybe `/user-packages`?

isoos: The account-related API endpoints follow the `/api/account/...` pattern.

PR #2815, opened 2019-09-25
Restrict visibility of authenticatedUser in account backend.

isoos:
No pressing reason for it, just looks like a good idea.

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -37,17 +37,17 @@ void registerAuthenticatedUser(User user) =>
     ss.register(#_authenticated_user, user);
 
 /// The active authenticated user.
-User get authenticatedUser => ss.lookup(#_authenticated_user) as User;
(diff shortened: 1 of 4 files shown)


app/lib/account/backend.dart
```diff
@@ -37,17 +37,17 @@ void registerAuthenticatedUser(User user) =>
     ss.register(#_authenticated_user, user);
 
 /// The active authenticated user.
-User get authenticatedUser => ss.lookup(#_authenticated_user) as User;
+User get _authenticatedUser => ss.lookup(#_authenticated_user) as User;
```
jonasfj: Fantastic initiative..

Maybe we should consider moving `_userAuthWrapper` or part of it into `backend.dart`, that way `registerAuthenticatedUser` can also be hidden...
I know we might need `registerAuthenticatedUser` in tests, but then we could annotated it `@visibleForTesting`

jonasfj: Note: not in this PR :)

PR #2816, opened 2019-09-25
Parse HTTP cookie for user session + minimal usage in layout.

isoos:
This is still only a fraction of the story, but wanted to submit for review while I'm working on the rest, as the shape of these parts seems to be not changing too much.

### DIFF
--- a/app/lib/shared/handler_helpers.dart
+++ b/app/lib/shared/handler_helpers.dart
@@ -13,7 +13,6 @@ import 'package:shelf/shelf_io.dart' as shelf_io;
 import 'package:stack_trace/stack_trace.dart';
 
 import '../account/backend.dart';
-import '../account/models.dart' show User, UserSessionData;
 import '../frontend/request_context.dart';
 import '../frontend/templates/layout.dart';
 
@@ -30,8 +29,6 @@ import 'utils.dart' show fileAnIssueContent, parseCookieHeader;
 // https://get.dev/#benefits
 // However, we should still emit it on other domains e.g. on pub.dartlang.org.
 const _hstsDuration = Duration(days: 365);
-
-final _logger = Logger('handlers');
 
 Future<void> runHandler(
   Logger logger,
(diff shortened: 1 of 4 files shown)


app/lib/shared/handler_helpers.dart
```diff
@@ -200,23 +203,42 @@ shelf.Handler _sanitizeRequestWrapper(shelf.Handler handler) {
...
+        parseCookieHeader(request.headers[HttpHeaders.cookieHeader]);
+    final sessionId = cookies['pub_sid'];
+    if (sessionId != null) {
+      userSessionData = await accountBackend.lookupSession(sessionId);
+    }
+
+    // verify if the ids match
+    if (user != null &&
+        userSessionData != null &&
+        user.userId != userSessionData.userId) {
```
jonasfj: Check if session has expired :)

app/lib/shared/handler_helpers.dart
```diff
@@ -200,23 +203,42 @@ shelf.Handler _sanitizeRequestWrapper(shelf.Handler handler) {
...
+    UserSessionData userSessionData;
+    final cookies =
+        parseCookieHeader(request.headers[HttpHeaders.cookieHeader]);
+    final sessionId = cookies['pub_sid'];
+    if (sessionId != null) {
+      userSessionData = await accountBackend.lookupSession(sessionId);
+    }
+
+    // verify if the ids match
+    if (user != null &&
```
jonasfj: I think it would be fine to omit this check.

In fact I think the two middlewares that does session lookup and authorization header check could be independent.

But I'm also not entirely sure, I mean this is fine, it just seems overkill.. Why would the API care about the session.
Let's talk tomorrow..

app/lib/shared/handler_helpers.dart
isoos: The accountBackend already checks for it, and I think that's a good place for such check (at least for the non-cached version, I'll create a PR for checking the cache too.)

app/lib/shared/handler_helpers.dart
isoos: Okay, if you are okay with splitting, I'll move these to separate wrappers and omit the check.

app/lib/shared/handler_helpers.dart
isoos: Doing this in the same PR.

PR #2817, opened 2019-09-26
Set and clear session cookie on login/logout.

isoos:
- The set and clear action is on an API endpoint, which makes it simpler (single place, one operation to test for), but it requires a page reload if the cookie has changed (otherwise the background action is a no-op). The reload should happen only on explicit login and logout actions.
- The javascript code does not have access to the cookie (it is http-only).
- The endpoint result contains the flag whether the method call changed the cookie state, and reload happens based on that. It also contains the expiry, which we can use later for displaying a warning before automatic logout or something like that.
- The auth initialization flow is not yet complete, (you can see it on staging, there is still a flicker of "Login" displayed), I'll fix that as a subsequent PR.

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -21,6 +21,8 @@ import 'auth_provider.dart';
 import 'google_oauth2.dart' show GoogleOauth2AuthProvider;
 import 'models.dart';
 
+/// The name of the session cookie.
+const pubSessionCookieName = 'pub_sid';
 final _logger = Logger('account.backend');
 final _uuid = Uuid();
 
@@ -285,20 +287,19 @@ class AccountBackend {
   /// `Cache-Control: private`, and may not be cached in server-side).
   /// JSON APIs whether fetching data or updating data cannot be authorized with
   /// a cookie carrying the `sessionId`.
-  Future<String> createNewSession({@required String imageUrl}) async {
+  Future<UserSessionData> createNewSession({@required String imageUrl}) async {
     final user = await requireAuthenticatedUser();
     final now = DateTime.now().toUtc();
     final sessionId = _uuid.v4().toString();
-    await _db.commit(inserts: [
-      UserSession()
-        ..id = sessionId
-        ..userIdKey = user.key
-        ..email = user.email
-        ..imageUrl = imageUrl
-        ..created = now
-        ..expires = now.add(Duration(days: 14)),
-    ]);
-    return sessionId;
+    final session = UserSession()
+      ..id = sessionId
+      ..userIdKey = user.key
+      ..email = user.email
+      ..imageUrl = imageUrl
+      ..created = now
+      ..expires = now.add(Duration(days: 14));
+    await _db.commit(inserts: [session]);
+    return UserSessionData.fromModel(session);
   }
 
   /// Returns the user session associated with the [sessionId] or null if it
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -279,12 +279,12 @@ class AccountBackend {
   }
 
   /// Creates a new session for the current authenticated user and returns the
-  /// new `sessionId`.
+  /// new session data.
   ///
-  /// The `sessionId` is a secret that will be stored in a secure cookie.
-  /// Presence of this `sessionId` in a cookie, can only be used to authorize
-  /// user specific content to be embedded in HTML pages (such pages must have
-  /// `Cache-Control: private`, and may not be cached in server-side).
+  /// The [UserSessionData.sessionId] is a secret that will be stored in a
+  /// secure cookie. Presence of this `sessionId` in a cookie, can only be used
+  /// to authorize user specific content to be embedded in HTML pages (such pages
+  /// must have `Cache-Control: private`, and may not be cached in server-side).
   /// JSON APIs whether fetching data or updating data cannot be authorized with
   /// a cookie carrying the `sessionId`.
   Future<UserSessionData> createNewSession({@required String imageUrl}) async {
--- a/app/lib/frontend/handlers/account.dart
+++ b/app/lib/frontend/handlers/account.dart
@@ -2,6 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:convert';
+import 'dart:io' show Cookie, HttpHeaders;
+
 import 'package:client_data/account_api.dart';
 import 'package:shelf/shelf.dart' as shelf;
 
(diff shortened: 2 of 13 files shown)


app/lib/account/backend.dart
```diff
@@ -285,20 +287,19 @@ class AccountBackend {
   /// `Cache-Control: private`, and may not be cached in server-side).
   /// JSON APIs whether fetching data or updating data cannot be authorized with
   /// a cookie carrying the `sessionId`.
-  Future<String> createNewSession({@required String imageUrl}) async {
+  Future<UserSessionData> createNewSession({@required String imageUrl}) async {
```
jonasfj: Update the documentation string to match new return type.

app/lib/frontend/handlers/pubapi.dart
```diff
@@ -183,6 +183,30 @@ class PubApi {
...
+  /// (automatic updates).
+  ///
+  /// The response header will contain the updated session cookie, and the body
+  /// is a [ClientSessionStatus].
+  @EndPoint.post('/api/account/session')
+  Future<Response> updateSession(
+          Request request, ClientSessionData sessionData) =>
+      updateSessionHandler(request, clientSessionData: sessionData);
+
+  /// Registers (or extends) a user session.
```
jonasfj: Removes the user session.

app/lib/frontend/handlers/pubapi.dart
```diff
@@ -183,6 +183,30 @@ class PubApi {
...
+  ///
+  /// This endpoint should be called after an explicit logout action or when the
+  /// client-side auth library logs out the current session.
+  ///
+  /// The response header will contain the session cookie with an immediate
+  /// expiration, and the body is a [ClientSessionStatus].
```
jonasfj: And clears the cookie value.

app/lib/shared/handler_helpers.dart
```diff
@@ -228,7 +228,7 @@ shelf.Handler _userSessionWrapper(shelf.Handler handler) {
   return (shelf.Request request) async {
     final cookies =
         parseCookieHeader(request.headers[HttpHeaders.cookieHeader]);
-    final sessionId = cookies['pub_sid'];
+    final sessionId = cookies[pubSessionCookieName];
     if (sessionId != null) {
```
jonasfj: Also require that `sessionId != ''`

app/lib/shared/handlers.dart
```diff
@@ -35,19 +35,19 @@ shelf.Response redirectToSearch(String query) {
...
-    body: body,
-    headers: {
-      'content-type': 'application/json; charset="utf-8"',
-      'x-content-type-options': 'nosniff',
-    },
-  );
+  headers ??= <String, String>{};
+  headers['content-type'] = 'application/json; charset="utf-8"';
+  headers['x-content-type-options'] = 'nosniff';
+  return shelf.Response(status, body: body, headers: headers);
```
jonasfj: Can't we also do
```dart
return shelf.Response(..., headers: {
  ...headers,
  'content-type': 'application/json; charset="utf-8"',
  'x-content-type-options': 'nosniff',
});
```

Not that it's important, hehe :)

pkg/client_data/lib/account_api.dart
```diff
@@ -2,11 +2,51 @@
...
 
 part 'account_api.g.dart';
 
+/// Session data to be cached from the authenticated client.
+@JsonSerializable()
+class ClientSessionData {
+  final String imageUrl;
+
+  ClientSessionData({
+    @required this.imageUrl,
```
jonasfj: We could also upload the `access_token` and then use it to call the userinfo end-point on the remote side..

We really ought to verify that:
 A) the `access_token` works with the `tokeninfo` end-point,
 B) It has one our allowed `audience`s,
 C) It has the right `user_id`.
We should do (A), (B) and (C) before calling the userinfo end-point.

Finally, we should verify (A), (B) and (C) when we do `createPublisher`.

I think `AccountBackend` has a method to get the user from an `access_token`, we just need to use that..

-------
Sorry for the scope creep, we can do `createPublisher` changes in a different PR.

pkg/client_data/lib/account_api.dart
```diff
@@ -2,11 +2,51 @@
...
+
+/// The server-provided status of the current session.
+@JsonSerializable()
+class ClientSessionStatus {
+  final bool changed;
+  final DateTime expires;
+
+  ClientSessionStatus({
+    @required this.changed,
+    @required this.expires,
```
jonasfj: Can we just expose `sessionChanged: bool` and `sessionIsValid: bool` to keep to reduce information leak...

(Feel free to suggest better names)

pkg/client_data/lib/account_api.dart
```diff
@@ -2,11 +2,51 @@
...
+  factory ClientSessionData.fromJson(Map<String, dynamic> json) =>
+      _$ClientSessionDataFromJson(json);
+
+  Map<String, dynamic> toJson() => _$ClientSessionDataToJson(this);
+}
+
+/// The server-provided status of the current session.
+@JsonSerializable()
+class ClientSessionStatus {
+  final bool changed;
```
jonasfj: A small doc-comment please... `/// True, if the user session has been updated and the current page should be reloaded`

app/lib/frontend/handlers/account.dart
```diff
@@ -14,6 +17,62 @@ import '../../shared/handlers.dart';
 
 import '../templates/admin.dart';
 
+/// Handles POST /api/account/session
+Future<shelf.Response> updateSessionHandler(shelf.Request request,
+    {ClientSessionData clientSessionData}) async {
+  final user = await requireAuthenticatedUser();
+
```
jonasfj: Let's put an `access_token` in the request body, validate it and do a lookup the userinfo service.

pkg/client_data/lib/account_api.dart
isoos: This seems to be a larger change, let's do it separately.

isoos: (especially as changing the access_token to id_token in the auth headers seems to need additional work)

pkg/client_data/lib/account_api.dart
isoos: I don't mind `sessionChanged` (vs the current `changed`), although as the object is `*SessionStatus`, it feels like we are better off with the shorter name.

`[session]IsValid` is a good question: there is not much value keeping it there, because I don't see a scenario where the `changed` doesn't trigger a page reload and the flag is useful (== we can meaningfully act on it, because it provides more information than we have already known). On the other hand, `expires` does have a useful scenario: we can trigger a popup before the session expires, preventing e.g. sessions where people would otherwise lose their status that they were about to send to the server.

jonasfj: > On the other hand, expires does have a useful scenario: we can trigger a popup before the session expires, preventing e.g. sessions where people would otherwise lose their status that they were about to send to the server.

Good point, it means we can re-create the session if the client-side is still signed in...

pkg/client_data/lib/account_api.dart
jonasfj: I'm okay doing this separately, can we file an issue and the land this.

isoos:
> File an issue for the access_token and id_token work..

https://github.com/dart-lang/pub-dev/issues/2830

PR #2819, opened 2019-09-26
'My publishers' page

isoos:
- the unauthenticated page seems to be useful for all of the similar cases
- reusing the publisher list template, only changing a few labels based on global or user-specific list

### DIFF
--- a/app/lib/frontend/handlers/routes.dart
+++ b/app/lib/frontend/handlers/routes.dart
@@ -201,6 +201,11 @@ class PubSiteService {
   Future<Response> oauthCallback(Request request) async =>
       oauthCallbackHandler(request);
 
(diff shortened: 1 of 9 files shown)


app/lib/frontend/handlers/routes.dart
```diff
@@ -201,6 +201,11 @@ class PubSiteService {
   Future<Response> oauthCallback(Request request) async =>
       oauthCallbackHandler(request);
 
+  /// List of the current user's publishers.
+  @Route.get('/account/publishers')
```
jonasfj: Maybe we should re-think this pattern.

isoos: ++, luckily it will be an easy thing to change

PR #2822, opened 2019-09-27
Cloud Build setup for deploying to staging

jonasfj:
On push to the `staging` branch, we now have that `tool/.deploy-staging.yaml` gets triggered on cloud-build in the staging project.

It can only do appengine deploys and cloud builds, but we could give it more permissions if needed.
However, I think this is fine for now.

Maybe next time we augment the datastore we give this script permissions deploy the `index.yaml` file... and add this to the deployment script.

### DIFF
--- a/tool/.deploy-staging.sh
+++ b/tool/.deploy-staging.sh
@@ -1,0 +1,53 @@
+#!/bin/bash -e
+
+# This script is used by .deploy-staging.yaml
+# Do NOT use this script for any other purposes.
+
+# This only works on cloud-build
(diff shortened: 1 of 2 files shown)


tool/.deploy-staging.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash -e
+
+# This script is used by .deploy-staging.yaml
+# Do NOT use this script for any other purposes.
```
isoos: Please add a note about the gcloud credential setup, how this script gets those variables.

PR #2824, opened 2019-09-27
Fix NPE when package was deleted but search still has it.

### LINKED ISSUE #2823: NPE while searching
```
Error:
    NoSuchMethodError: The getter 'name' was called on null.
    Receiver: null
    Tried calling: name

Stack:
    dart:core-patch/object_patch.dart 50:5                         Object.noSuchMethod
    package:pub_dartlang_org/package/search_service.dart 109:24    SearchService._loadResultForPackages
    package:pub_dartlang_org/package/search_service.dart 49:12     SearchService.search
    package:pub_dartlang_org/frontend/handlers/listing.dart 51:44  _packagesHandlerHtmlCore
    package:pub_dartlang_org/frontend/handlers/listing.dart 30:5   packagesHandlerHtml
    package:pub_dartlang_org/frontend/handlers/routes.dart 66:49   PubSiteService.packages
```

isoos:
Looks like the issue is less critical than I've first thought.
Fixes #2823.

### DIFF
--- a/app/lib/package/search_service.dart
+++ b/app/lib/package/search_service.dart
@@ -78,7 +78,10 @@ class SearchService {
   Future<PackageView> _getPackageView(String package) async {
     return await cache.packageView(package).get(() async {
       final p = await packageBackend.lookupPackage(package);
-      if (p == null) return null;
+      if (p == null) {
+        _logger.warning('Package lookup failed for "$package".');
+        return null;
+      }
 
       final version = p.latestVersion;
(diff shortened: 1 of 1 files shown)


app/lib/package/search_service.dart
```diff
@@ -78,7 +78,10 @@ class SearchService {
   Future<PackageView> _getPackageView(String package) async {
```
jonasfj: Update doc-comment to say it can return `null`? :)

PR #2826, opened 2019-09-27
Reduce the extracted dartdoc data for large packages.

isoos:
By removing the documentation text (if the package has more than 1000 documented API entries),  this reduces the file size (parsing, update time) and the index memory allocation.  The file size of `package:googleapis` changes from 21MB to 9.6MB.

The coverage data is now included in the `pub-data.json` file, because with the removal we are losing the data source for the coverage percent.


pkg/pub_dartdoc_data/lib/pub_dartdoc_data.dart
```diff
@@ -2,16 +2,20 @@
...
+import 'dart:math';
+
 import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart';
 
 part 'pub_dartdoc_data.g.dart';
 
 @JsonSerializable()
 class PubDartdocData {
+  final Coverage coverage;
```
jonasfj: If this can be `null` we should document the conditions.

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -244,7 +243,7 @@ class DartdocJobProcessor extends JobProcessor {
       await toolEnvRef.release();
     }
 
-    final coverage = calculateCoverage(dartdocData);
+    final coverage = dartdocData.coverage ?? Coverage(total: 0, documented: 0);
```
jonasfj: When can `dartdocData.coverage` be `null`? When loading old data?

Can we bump the runtime version or somehow avoid this? I'm not a fan of sprinkling obscure conditions all over the code.

isoos: Good point, here this can't be `null`. But reading extracted data from old reports may make it null. I'll update the comment.

PR #2834, opened 2019-09-30
Hardening the session cookie

isoos:


jonasfj:
See https://github.com/dart-lang/pub-dev/pull/2835

app/lib/frontend/handlers/account.dart
```diff
@@ -45,9 +45,11 @@ Future<shelf.Response> updateSessionHandler(shelf.Request request,
   final cookie = Cookie(pubSessionCookieName, newSession.sessionId)
     ..expires = newSession.expires
     ..httpOnly = true
+    ..secure = request.requestedUri.scheme != 'http'
+    ..domain = request.requestedUri.host
     ..path = '/';
   final headers = <String, String>{
-    HttpHeaders.setCookieHeader: cookie.toString(),
+    HttpHeaders.setCookieHeader: '$cookie; SameSite=strict',
```
jonasfj: Oh, retrack approve... `strict` won't work for us.

jonasfj: See: https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02#section-8.8.2

This is also why we need to be very explicit that session is only for read operations... and not for any JSON APIs, not even JSON read operations.

PR #2835, opened 2019-09-30
Cookie hardening

jonasfj:


app/lib/frontend/handlers/account.dart
```diff
@@ -42,12 +42,25 @@ Future<shelf.Response> updateSessionHandler(shelf.Request request,
...
+      // Hence, such a cookie cannot have been set by another website or an
+      // HTTP proxy for this website.
+      '$pubSessionCookieName=${newSession.sessionId}',
+      // Send cookie to anything under '/' required by '__Host-' prefix.
+      'Path=/',
+      // Do not include the cookie in CORS requests, unless the request is a
+      // top-level navigation to the site, as recommended in:
+      // https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-02#section-8.8.2
+      'SameSite=Lax',
+      'Secure', // Only allow this cookie to be sent when making HTTPS requests.
```
isoos: can we have a conditional for localhost?
e.g. 
```
if (request.requestedUri.scheme != 'http') 'Secure',
```
(or even `if (!request.requestedUri.toString().startsWith('http://localhost:8080/') 'Secure',`

jonasfj: I think we should somehow put into the `configuration` whether the process was started with `--local-dev` and only in that case should we do something special.

People might be hitting the service with `Host: localhost` in production, you never know. Certainly there is no good reason to allow things like that in production. Let's be defensive :)

isoos: ++

jonasfj:
@isoos, please take a look, I did a lot of changes since last review.. Now it should work on staging and on localhost (provided you're using the right host).

I also added more hosts to the oauth2 flow for our staging site, so we can use the version specific host name.

jonasfj:
This currently deployed to `https://jonasfj-dot-dartlang-pub-dev.appspot.com`

app/lib/account/session_cookie.dart
```diff
@@ -0,0 +1,91 @@
...
+
+/// Parse [cookieString] and return `sessionId` or `null`.
+///
+/// The [cookieString] is the value of the `cookie:` request header.
+String parseSessionCookie(String cookieString) {
+  ArgumentError.checkNotNull(cookieString, 'cookieString');
+
+  // The cookieString is separated by '; ', and contains 'name=value'
+  // See: https://tools.ietf.org/html/rfc6265#section-5.4
+  final cookie = cookieString.split('; ').firstWhere(
```
isoos: Nit: We have `parseCookieHeader` in `shared/utils.dart`.

jonasfj: Okay, did a few tweaks and adopted this..

PR #2839, opened 2019-10-01
My packages page + search index with owners.

isoos:
- This is similar (both in behaviour and in layout) to `/packages` and `/publishers/<id>/packages`, only with the narrow scope for the current account's email + its publishers
- I think the layout could be improved if we had tabs similar to publisher (just with "My packages" and "My publishers"), but deferring to a subsequent PR.

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -114,7 +114,7 @@ class SearchBackend {
       dependencies: _buildDependencies(analysisView),
       publisherId: p.publisherId,
       emails: await _buildEmails(p, pv),
-      owners: await _buildOwners(p),
+      uploaderEmails: await _buildUploaderEmails(p),
       apiDocPages: apiDocPages,
       timestamp: DateTime.now().toUtc(),
     );
@@ -140,13 +140,13 @@ class SearchBackend {
     return emails.toList()..sort();
   }
 
-  Future<List<String>> _buildOwners(Package p) async {
-    if (p.publisherId != null) {
-      return [p.publisherId];
-    } else {
+  Future<List<String>> _buildUploaderEmails(Package p) async {
+    if (p.publisherId == null) {
       final owners = await accountBackend.getEmailsOfUserIds(p.uploaders);
       owners.sort();
       return owners;
+    } else {
+      return <String>[];
     }
   }
 
@@ -174,7 +174,7 @@ class SearchBackend {
         maintenance: 0.0,
         publisherId: p.publisherId,
         emails: await _buildEmails(p, null),
-        owners: await _buildOwners(p),
+        uploaderEmails: await _buildUploaderEmails(p),
         timestamp: DateTime.now().toUtc(),
       );
     }
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -114,6 +114,7 @@ class SearchBackend {
       dependencies: _buildDependencies(analysisView),
       publisherId: p.publisherId,
       emails: await _buildEmails(p, pv),
+      owners: await _buildOwners(p),
       apiDocPages: apiDocPages,
       timestamp: DateTime.now().toUtc(),
     );
@@ -137,6 +138,16 @@ class SearchBackend {
       emails.add(author.email);
     }
     return emails.toList()..sort();
+  }
+
+  Future<List<String>> _buildOwners(Package p) async {
+    if (p.publisherId != null) {
+      return [p.publisherId];
+    } else {
+      final owners = await accountBackend.getEmailsOfUserIds(p.uploaders);
+      owners.sort();
+      return owners;
+    }
   }
 
   List<ApiDocPage> _apiDocPagesFromPubDataText(String text) {
@@ -163,6 +174,7 @@ class SearchBackend {
         maintenance: 0.0,
         publisherId: p.publisherId,
         emails: await _buildEmails(p, null),
+        owners: await _buildOwners(p),
         timestamp: DateTime.now().toUtc(),
       );
     }
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -141,13 +141,12 @@ class SearchBackend {
   }
 
   Future<List<String>> _buildUploaderEmails(Package p) async {
-    if (p.publisherId == null) {
-      final owners = await accountBackend.getEmailsOfUserIds(p.uploaders);
-      owners.sort();
-      return owners;
-    } else {
+    if (p.publisherId != null) {
       return <String>[];
     }
+    final uploaders = await accountBackend.getEmailsOfUserIds(p.uploaders);
+    uploaders.sort();
+    return uploaders;
   }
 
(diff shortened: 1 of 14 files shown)


app/lib/search/search_service.dart
```diff
@@ -273,6 +281,7 @@ class SearchQuery {
     final String q = uri.queryParameters['q'];
     final String platform =
         uri.queryParameters['platform'] ?? uri.queryParameters['platforms'];
+    final owners = uri.queryParametersAll['owners'];
```
jonasfj: The semantics is that you want results filtered by "is owned by one of these", right?

Hence, I suggest the query parameter name `owned_by_one_of`, so url would be `../?q=...&owned_by_one_of=jonasfj@google.com,dart.dev`.

Then we just do `(uri.queryParameter['owned_by_one_of'] ?? '').split(',')` or something like that.

app/lib/search/search_service.dart
```diff
@@ -350,18 +363,25 @@ class SearchQuery {
...
     if (publisherId != null && publisherId.isNotEmpty) {
       path = '/publishers/$publisherId/packages';
     }
+    if (owners != null && owners.isNotEmpty) {
+      path = '/account/packages';
+    }
+    return path;
+  }
+
+  String toSearchLink({int page}) {
```
jonasfj: What does this method do?

app/lib/search/search_service.dart
```diff
@@ -218,6 +221,7 @@ class SearchQuery {
   final String query;
   final ParsedQuery parsedQuery;
   final String platform;
+  final List<String> owners;
```
jonasfj: Can we document this? also wouldn't it be better to call this `ownersIsOneOf` or something more descriptive?

app/lib/search/search_service.dart
```diff
@@ -60,6 +60,7 @@ class PackageDocument {
   final Map<String, String> dependencies;
   final String publisherId;
   final List<String> emails;
```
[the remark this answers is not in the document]
isoos: uploaders + specific uploader of the latest stable version + pubspec.authors of the latest stable version

app/lib/search/search_service.dart
```diff
@@ -60,6 +60,7 @@ class PackageDocument {
...
+  final List<String> owners;
```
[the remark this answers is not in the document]
isoos: That would soft-break the current behaviour, I wanted to keep the option to have them separately. As long as we expose the e-mail search on the package's infobox, I'd rather keep it separately.

app/lib/search/index_simple.dart
```diff
@@ -168,6 +168,17 @@ class SimplePackageIndex implements PackageIndex {
       });
     }
 
+    // filter on owners
+    if (query.owners != null && query.owners.isNotEmpty) {
+      packages.removeWhere((package) {
+        final doc = _packages[package];
+        for (String owner in query.owners) {
+          if (doc.owners.contains(owner)) return false;
```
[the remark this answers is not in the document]
isoos: In other words: keep only docs where the two has intersection.

app/lib/search/search_service.dart
isoos: We could do that too, but this is a backend URL, which we don't expose to the users, so it could also use the already built-in multi-value URL parsing. At some point we probably need to have a proper JSON API for this too.

app/lib/search/index_simple.dart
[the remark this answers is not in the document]
jonasfj: ah, never mind I read it wrong... Now I see the `return true` :)

app/lib/search/search_service.dart
isoos: Do we expect to have queries where we have owners with the semantics of all-of-them? I think there may be a query where we want to know which packages we have a shared ownership of, but the use-case is probably not that common. If that is not something we want to have here, let's keep it short.

app/lib/search/search_service.dart
```diff
@@ -58,9 +58,17 @@ class PackageDocument {
   final double maintenance;
 
   final Map<String, String> dependencies;
+
+  /// The publisher id of the package
```
jonasfj:
```suggestion
  /// The publisher id of the package, `null` if not owned by a publisher.
```

app/lib/search/search_service.dart
jonasfj: Since this is an internal thing, maybe it's better to keep it long and descriptive... I do like short names, but this is also sort of a hack because we don't have OR support in queries... so we make this work as a hack.

We can make it clear that this is a workaround for not having or... by writing `ownerIsOneOf` or `ownedByOneOf`... as this won't conflict with a future filter we might add..

app/lib/search/backend.dart
```diff
@@ -139,6 +140,16 @@ class SearchBackend {
     return emails.toList()..sort();
   }
 
+  Future<List<String>> _buildUploaderEmails(Package p) async {
+    if (p.publisherId == null) {
+      final owners = await accountBackend.getEmailsOfUserIds(p.uploaders);
+      owners.sort();
+      return owners;
```
jonasfj:
```suggestion
      final uploaders = await accountBackend.getEmailsOfUserIds(p.uploaders);
      uploaders.sort();
      return uploaders;
```

app/lib/search/backend.dart
```diff
@@ -139,6 +140,16 @@ class SearchBackend {
     return emails.toList()..sort();
   }
 
+  Future<List<String>> _buildUploaderEmails(Package p) async {
+    if (p.publisherId == null) {
```
jonasfj:
```suggestion
    if (p.publisherId != null) {
      return <String>[];
    }
```
nit: prefer to return early...

app/lib/search/index_simple.dart
```diff
@@ -168,6 +168,24 @@ class SimplePackageIndex implements PackageIndex {
       });
     }
 
+    // filter on owners
+    if (query.uploaderOrPublishers != null &&
+        query.uploaderOrPublishers.isNotEmpty) {
```
jonasfj: nit: Can it be empty? If not let's add an assert instead...

app/lib/search/index_simple.dart
```diff
@@ -168,6 +168,24 @@ class SimplePackageIndex implements PackageIndex {
...
+          if (doc.publisherId == owner) {
+            return false;
+          }
+          if (doc.uploaderEmails != null &&
+              doc.uploaderEmails.contains(owner)) {
+            return false;
+          }
+        }
+        return true;
+      });
```
jonasfj: nit:
```dart
    if (query.uploaderOrPublishers != null &&
        query.uploaderOrPublishers.isNotEmpty) {
      packages.removeWhere((package) {
        final doc = _packages[package];
        if (doc.publisherId != null) {
          return !query.uploaderOrPublishers.contains(doc.publisherId);
        }
        if (doc.uploaderEmails == null) {
          return true; // turn this into an error in the future.
        }
        return !query.uploaderOrPublishers.any(doc.uploaderEmails.contains);
      });
```

PR #2843, opened 2019-10-02
Remove obsolete user sessions.

isoos:
I've also moved the periodic timer to the worker isolate, that way these won't block the frontend serving.

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -340,6 +340,14 @@ class AccountBackend {
     await cache.userSessionData(sessionId).purge();
   }
 
-  // TODO: periodically remove expired sessions from datastore and cache
+  /// Removes the expired sessions from Datastore and Redis cache.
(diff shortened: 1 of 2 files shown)


app/lib/account/backend.dart
```diff
@@ -340,6 +340,14 @@ class AccountBackend {
     await cache.userSessionData(sessionId).purge();
   }
 
-  // TODO: periodically remove expired sessions from datastore and cache
+  /// Removes the expired sessions from Datastore and Redis cache.
+  Future deleteObsoleteSessions() async {
+    final query = _db.query<UserSession>()
+      ..filter('expires <', DateTime.now().toUtc());
```
jonasfj: nit: Maybe wait for 15 min... just to account for possible clock skew...

(I just generally do this whenever I do anything with timestamps and expiration to reduce risk of bugs)

PR #2846, opened 2019-10-03
Updating account-related pages

isoos:
- `/my-packages` and `/my-publishers`
- using detail-page template
- added missing test for my-publishers rendering


app/lib/frontend/handlers/account.dart
```diff
@@ -136,10 +136,27 @@ Future<shelf.Response> accountPackagesPageHandler(shelf.Request request) async {
...
   );
   return htmlResponse(html);
 }
+
+/// Handles requests for GET /my-publishers
+Future<shelf.Response> accountPublishersPageHandler(
+    shelf.Request request) async {
+  if (userSessionData == null) {
+    return htmlResponse(renderUnauthenticatedPage());
+  } else {
```
jonasfj: nit: Return early, rather than having an `else` branch.

isoos:
> I'm not sure we need an info-box on the /my-packages page..

Not sure either, but easy to remove.

PR #2849, opened 2019-10-03
Fixed authentication race condition, and support sign-in on-demand

jonasfj:


pkg/web_app/lib/src/account.dart
```diff
@@ -18,31 +19,53 @@ import 'google_js.dart';
...
 /// The pub API client to use with account credentials.
 PubApiClient get client {
-  if (_client == null) {
-    _httpClient ??=
-        getAuthenticatedClient(currentUser?.getAuthResponse(true)?.id_token);
-    final uri = Uri.parse(window.location.href);
-    _client = PubApiClient(uri.resolve('/').toString(), client: _httpClient);
-  }
-  return _client;
+  return PubApiClient(_baseUrl, client: createAuthenticatedClient(() async {
```
isoos: Shouldn't we cache the instance? Or the BrowserClient is a throw-away instance anyway, and we don't gain anything from it?

jonasfj: I haven't looked at `BrowserClient`, but if it's implemented in a browser I doubt it can cache anything... As connections are managed by the browser.

If any event, if we profile performance issues then I suggest we cache, until then head -> sand :see_no_evil: 

(seriously, though, I/O is the expensive part here, I doubt the rest has much cost).

PR #2851, opened 2019-10-03
Materialize admin page.

jonasfj:
* Still missing the members list.


isoos:
After this is merged, we can remove the async info fetch for the description and other fields from the web app, am I right?

jonasfj:
Yes, we can remove all async fetching... I'm perfectly fine with that :)

jonasfj:
In fact it would be easier to do the current members in material if rendered it serverside...

PR #2852, opened 2019-10-04
Consent page with session data.

isoos:
- In `consent_backend.dart`, mixing the API serving with regular backend seems to be non-ideal in this case, however, the separate API method would be almost identical with the non-API one.

### DIFF
--- a/app/lib/account/consent_backend.dart
+++ b/app/lib/account/consent_backend.dart
@@ -38,8 +38,8 @@ class ConsentBackend {
   ConsentBackend(this._db);
 
   /// Returns the consent details.
-  Future<api.Consent> getConsent(String consentId) async {
-    final user = await requireAuthenticatedUser();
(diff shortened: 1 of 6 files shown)


app/lib/account/consent_backend.dart
```diff
@@ -38,8 +38,8 @@ class ConsentBackend {
   ConsentBackend(this._db);
 
   /// Returns the consent details.
-  Future<api.Consent> getConsent(String consentId) async {
-    final user = await requireAuthenticatedUser();
+  Future<api.Consent> getConsent(String consentId, {User user}) async {
```
jonasfj: I'd rather we duplicate.. or that we have `getConsentApi` which calls `requireAuthenticatedUser` and then calls `getConsent(user)`.

Notably, I don't like the implication that you can avoid authorization by passing a user object.

PR #2855, opened 2019-10-04
Spinner + result handling for RPCs in web app.

isoos:


pkg/web_app/lib/src/_dom_helper.dart
```diff
@@ -4,11 +4,85 @@
...
 import 'dart:html';
+
+import 'package:markdown/markdown.dart' as markdown;
 import 'package:mdc_web/mdc_web.dart' show MDCDialog;
+import 'package:meta/meta.dart';
+
+import 'pubapi.client.dart';
+
+/// Wraps asynchronous server calls with an optional confirm window, a spinner,
+/// error- and success handling.
```
jonasfj: Would be nice to document parameters...

pkg/web_app/lib/src/account.dart
```diff
@@ -182,20 +179,22 @@ class _PkgAdminWidget {
...
-      await client.setPackageOptions(pageData.pkgData.package, options);
-      window.location.reload();
-    });
+    await rpc(
+      confirmQuestion:
+          'Are you sure you want change the "discontinued" status of the package?',
+      fn: () async {
+        final options =
+            PkgOptions(isDiscontinued: !pageData.pkgData.isDiscontinued);
+        await client.setPackageOptions(pageData.pkgData.package, options);
```
jonasfj: nit,
```
        await client.setPackageOptions(pageData.pkgData.package, PkgOptions(
          isDiscontinued: !pageData.pkgData.isDiscontinued,
        ));
```

Maybe it's just me... but this is a case where I love trailing comma, and avoid introducing variable I don't need :)

pkg/web_app/lib/src/account.dart
```diff
@@ -182,20 +179,22 @@ class _PkgAdminWidget {
...
+      },
+      successMessage: '"discontinued" status changed. The page will reload.',
+      onSuccess: (_) {
```
jonasfj: nit: `(_) => window.location.reload();`

PR #2860, opened 2019-10-04
Include the list of members in the publisher admin page, and materialize the look-and-feel

jonasfj:


pkg/web_app/lib/src/account.dart
```diff
@@ -303,8 +301,13 @@ class _PublisherAdminWidget {
     if (isActive) {
       _updateButton.onClick.listen((_) => _updatePublisher());
       _inviteMemberButton.onClick.listen((_) => _inviteMember());
+      for (final btn in document.querySelectorAll('.-pub-remove-user-button')) {
+        btn.onClick.listen((_) => _removeMember(
+              btn.dataset['user-id'],
```
isoos: Nice!

pkg/web_css/lib/src/_account.scss
```diff
@@ -48,6 +48,14 @@
   }
 }
 
+.mdc-data-table__table {
+  margin-bottom: 0px !important;
+}
+
+.-pub-remove-user-button {
+  background-color: #ff4242 !important;
```
isoos: Please add a comment that this is a destructive operation and styled differently.

jonasfj: Good point...

PR #2875, opened 2019-10-07
Fixed broken JS on site, because mdc was not available

jonasfj:
I know we have to update the golden templates.
But not in this PR, as it would conflict with the other pending PR.
So let's just do that when we merge.


isoos:
> it would conflict with the other pending PR

I think this can be updated concurrently, and won't conflict.

PR #2877, opened 2019-10-08
Init mdc after load

jonasfj:


pkg/web_app/lib/script.dart
```diff
@@ -14,7 +14,7 @@ import 'src/search.dart';
 import 'src/tabs.dart';
 
 void main() {
-  window.onContentLoaded.listen((_) => mdc.autoInit());
+  window.onLoad.listen((_) => mdc.autoInit());
```
jonasfj: Not ideal, but it works reliably.

Ideal would be to `layout()` all components after `onload`, but that's complicated maybe we do this later..

PR #2879, opened 2019-10-08
Enable publishers.

isoos:
Small functional changes:
- sitemap does not contain `/publishers` (we do allow it, but not promote it)
- sitemap does not contain `/create-publisher` (as you'd need to be logged in for it)
- scoped tests auto-register the configuration object in the scope (because layout rendering needs to access it)

I've also checked and there is not `experimental` style used in our CSS or web_app


jonasfj:
Can we deploy this on staging?

(note to self: review this by following references)

PR #2880, opened 2019-10-08
Sort control on publisher- and my-packages.

isoos:
I've removed the `max-width` from the CSS style, as it was breaking it into two lines on the "My packages" page. In practice this does not affect the layout much, even with long search queries.


jonasfj:
Should we consider shipping this later?

PR #2888, opened 2019-10-09
Fix publisher listing page caching.

isoos:


app/lib/frontend/handlers/publisher.dart
```diff
@@ -30,10 +30,17 @@ Future<shelf.Response> createPublisherPageHandler(shelf.Request request) async {
...
+  if (requestContext.uiCacheEnabled) {
+    final content = await cache.uiPublisherListPage().get(() async {
+      final publishers = await publisherBackend.listPublishers(limit: 1000);
+      return renderPublisherListPage(publishers);
+    });
+    return htmlResponse(content);
+  }
+
+  // no caching for logged-in user
+  final publishers = await publisherBackend.listPublishers(limit: 1000);
```
jonasfj: We need a TODO here for supporting paging... And we should cache the list of publishers, not in this PR, but just in general :)

app/lib/shared/handler_helpers.dart
```diff
@@ -67,10 +67,11 @@ shelf.Handler wrapHandler(
   }
   handler = _httpsWrapper(handler);
   handler = _logRequestWrapper(logger, handler);
+  handler =
+      _requestContextWrapper(handler); // need to run after session wrapper
```
jonasfj: Maybe not that handlers wrap other handlers, so they are listed in reverse order...

PR #2900, opened 2019-10-09
Update dialog to explain that we only support second-level domains

jonasfj:


pkg/web_app/lib/src/account.dart
```diff
@@ -242,9 +242,13 @@ class _CreatePublisherWidget {
   }
 
   void _triggerCreate(String publisherId) async {
-    if (publisherId.isEmpty || !publisherId.contains('.')) {
+    if (publisherId.isEmpty || publisherId.split('.').length != 2) {
       await modalMessage(
-          'Input validation', 'Please use a domain name as publisher id.');
+        'Input validation',
+        'Please use a second-level domain name as publisher identifier.<br />'
+            'Third-level domains names like <code>example.co.uk</code> is '
```
isoos: s/is/are

isoos: domains names -> domain names

Issue #2909, opened 2019-10-10
pub.dev login flow/text issues


kwalrath:
If I log in to pub.dev and then log out and then click **Login** again, I don't get a choice of identities to sign in as, and I still see **Login**.

However, the login did apparently work: If I reload the page, I see my face (as the original identity I signed in as).

Also, to conform to Google standards (e.g. see google.dev):
* **Login** -> **Sign in**
* **Log out** -> **Sign out**

jonasfj:
Hmm, interesting, if signed in to multiple accounts in the same browser, you do get the account selector.

We probably have to force account selection whenever you sign-in:
https://developers.google.com/identity/sign-in/web/reference#gapiauth2signinoptions

PR #2919, opened 2019-10-11
Always select the google account on login

isoos:
#2909


jonasfj:
Maybe update changelog, as login stuff is useful to track..

PR #2922, opened 2019-10-15
Add API endpoints and backend implementation for package likes.

szakarias:


app/lib/account/backend.dart
```diff
@@ -180,6 +180,46 @@ class AccountBackend {
...
+  }
+
+  /// Return an iterable with the names of all the packages that the given [user] likes.
+  Future<List<String>> listPackageLikes(User user) async {
+    final likes = await _db.query<Like>(ancestorKey: user.key).run().toList();
+    return likes.map((Like l) => l.package).toList();
+  }
+
+  /// Creates an a package like entry for the given [user] and [package].
+  Future<void> likePackage(User user, String package) async {
```
jonasfj: nit, returning the `Like` entity would be nice :)

That way we expose the `created` timestamp to the caller.

app/lib/frontend/handlers/account.dart
```diff
@@ -117,6 +117,58 @@ Future<AccountPkgOptions> accountPkgOptionsHandler(
       isAdmin: await packageBackend.isPackageAdmin(p, user.userId));
 }
 
+/// Handles GET /api/account/likes
+Future<LikedPackagesRepsonse> packageLikesHandler(shelf.Request request) async {
+  final user = await requireAuthenticatedUser();
+  final packages = await accountBackend.listPackageLikes(user);
+  final List<PackageLikeResponse> packageLikes = List.from(
+      packages.map((String p) => PackageLikeResponse(liked: true, package: p)));
```
jonasfj: You're missing `created`

app/lib/account/backend.dart
```diff
@@ -180,6 +180,46 @@ class AccountBackend {
...
+    try {
+      like = await _db.lookupValue(key);
+    } on KeyNotFoundException {
+      return null;
+    }
+    return like;
+  }
+
+  /// Return an iterable with the names of all the packages that the given [user] likes.
+  Future<List<String>> listPackageLikes(User user) async {
```
jonasfj: We might want to return the `Like` entitites...

PR #2928, opened 2019-10-18
Add 'Package.likes' property.

szakarias:


app/lib/account/backend.dart
```diff
@@ -199,26 +200,62 @@ class AccountBackend {
...
+  /// Creates and returns a package like entry for the given [user] and
+  /// [package], and increments the 'likes' property on [package].
   Future<Like> likePackage(User user, String package) async {
-    final newLike = Like()
-      ..parentKey = user.key
-      ..id = package
-      ..created = DateTime.now().toUtc();
-    await _db.commit(inserts: [newLike]);
-    return newLike;
+    final like = await _db.withTransaction<Like>((tx) async {
```
jonasfj: nit: `return await _db,withTransaction<Like>(....`

app/lib/account/backend.dart
```diff
@@ -199,26 +200,62 @@ class AccountBackend {
...
+      final packageKey = _db.emptyKey.append(Package, id: package);
+      final p = await tx.lookupValue<Package>(packageKey);
```
jonasfj:
```suggestion
      final p = await tx.lookupValue<Package>(packageKey, orElse: () => null);
```

Hmm, I think we should add test cases for "package does not exist" and things like that.

app/lib/account/backend.dart
```diff
@@ -199,26 +200,62 @@ class AccountBackend {
...
-        _db.emptyKey.append(User, id: user.id).append(Like, id: package);
-    try {
-      await _db.lookupValue(key);
-    } on KeyNotFoundException {
-      return;
-    }
-    await _db.commit(deletes: [key]);
+    await _db.withTransaction<void>((tx) async {
+      final packageKey = _db.emptyKey.append(Package, id: package);
+      final p = await tx.lookupValue<Package>(packageKey);
```
jonasfj:
```suggestion
      final p = await tx.lookupValue<Package>(packageKey, orElse: () => null);
```
Just for sanity maybe do a `grep lookupValue` and see if you missed `orElse` somewhere :)

app/test/account/like_test.dart
```diff
@@ -33,12 +33,12 @@ void main() {
...
+      expect(rs2.package, hydrogen.package.name);
       expect(rs2.liked, true);
 
-      final rs3 = await client.likePackage('helium');
-      expect(rs3.package, 'helium');
+      final rs3 = await client.likePackage(helium.package.name);
+      expect(rs3.package, helium.package.name);
       expect(rs3.liked, true);
 
       final rs4 = await client.listPackageLikes();
```
jonasfj: nit. can we add `expect(rs.likedPackages.map((e) => e.package), contains(hydrogen.package.name));`...

app/test/account/like_test.dart
```diff
@@ -87,12 +87,51 @@ void main() {
...
+      final rs8 = await client2.getPackageLikes(hydrogen.package.name);
+      expect(rs8.likes, 1);
+
+      await client2.unlikePackage(hydrogen.package.name);
+
+      final rs9 = await client.getPackageLikes(hydrogen.package.name);
+      expect(rs9.likes, 0);
+      final rs10 = await client2.getPackageLikes(hydrogen.package.name);
+      expect(rs10.likes, 0);
+    });
```
jonasfj: What if:
 * User likes a package that doesn't exist,
 * User unlikes a package that doesn't exist
 * User unlikes a package the user didn't liked (does this decrement, does it cause an error -- it shouldn't)
 * A client requests the list of likes for a package that doesn't exist
 * A client requests the list of likes for a package without attaching an `authorization` header? (You can call `createPubApiClient()` without an `authToken` to get a client doesn't send an `authorization` header), this is just to check that the API end-point doesn't require a user, since we're just requesting public information.

Note: You can use `expectApiException` to test exceptions that makes it pretty easy :)

PR #2937, opened 2019-10-22
Explain that missing consents might be expired.

sigurdm:
Also validate consent Ids before showing them.

### DIFF
--- a/app/lib/account/consent_backend.dart
+++ b/app/lib/account/consent_backend.dart
@@ -37,15 +37,9 @@ class ConsentBackend {
 
   ConsentBackend(this._db);
 
-  /// Throws if [consentId] is malformed.
-  void validateConsent(String consentId) {
-    InvalidInputException.checkMatchPattern(
-        consentId, 'consentId', RegExp(r'^[a-zA-Z0-9]*$'));
-  }
-
   /// Returns the consent details.
   Future<api.Consent> getConsent(String userId, String consentId) async {
-    validateConsent(consentId);
+    InvalidInputException.checkUlid(consentId, 'consentId');
     final key =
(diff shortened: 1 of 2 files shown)


app/lib/account/consent_backend.dart
```diff
@@ -37,8 +37,15 @@ class ConsentBackend {
 
   ConsentBackend(this._db);
 
+  /// Throws if [consentId] is malformed.
+  void validateConsent(String consentId) {
+    InvalidInputException.checkMatchPattern(
+        consentId, 'consentId', RegExp(r'^[a-zA-Z0-9]*$'));
```
jonasfj: or add `InvalidInputException.checkUlid(value, name)` I imagine we'll use ulid's else where too.

app/lib/account/consent_backend.dart
```diff
@@ -37,8 +37,15 @@ class ConsentBackend {
 
   ConsentBackend(this._db);
 
+  /// Throws if [consentId] is malformed.
+  void validateConsent(String consentId) {
```
isoos: make this private

app/lib/shared/exceptions.dart
```diff
@@ -78,11 +78,10 @@ class InvalidInputException extends ResponseException {
   }
 
   /// Throw [InvalidInputException] if [value] doesn't match [pattern].
-  static void checkMatchPattern(String value, String name, Pattern pattern) {
```
isoos: Please match the method name and param (e.g. `checkMatchRegExp` or use `Pattern` instead of `RegExp`).

sigurdm: Done

PR #2943, opened 2019-10-24
Fix banner bg reference in CSS

### LINKED ISSUE #2939: Background image 404 in js console
When loading `pub.dev` we have:
```
GET https://pub.dev/img/background-pattern-darkblue.jpg?hash=tqigd6sddnrgq97c7crv8d2l6agjtrob 404
```
In the console... I think our CSS is loading the image incorrectly somewhere. It's missing `/static/`..

Note, that it's loaded correctly somewhere else, so I don't see any impact from this other than ugly logs, 404s in our server logs and minor performance degradation... But it's also possible I'm missing some implication :)

isoos:
Fixes #2939

### DIFF
--- a/pkg/web_css/lib/src/_base.scss
+++ b/pkg/web_css/lib/src/_base.scss
@@ -168,7 +168,7 @@ pre > code {
 
 ._banner-bg {
   background: $color-searchbar-dark-bg;
-  background-image: url('../img/background-pattern-darkblue.jpg?hash=tqigd6sddnrgq97c7crv8d2l6agjtrob');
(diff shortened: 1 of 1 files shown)


pkg/web_css/lib/src/_base.scss
```diff
@@ -168,7 +168,7 @@ pre > code {
 
 ._banner-bg {
   background: $color-searchbar-dark-bg;
-  background-image: url('../img/background-pattern-darkblue.jpg?hash=tqigd6sddnrgq97c7crv8d2l6agjtrob');
+  background-image: url('/static/img/background-pattern-darkblue.jpg?hash=tqigd6sddnrgq97c7crv8d2l6agjtrob');
```
jonasfj: Just curious, how did this work before?

PR #2949, opened 2019-10-24
Ensure atomicity around user Entity creation.

jonasfj:
From reading about datastore transactions I understand that a transaction
will see a consistent snapshot of the datastore.

Hence, if we try to read an entity and see that it doesn't exist before writing it,
then another transaction can't also do the same thing. As each transaction should see
serializable snapshots of the datastore. I think the problem before was that we
weren't reading the entity in the same transaction that we created it in.

We still have an issue with the email lookups, this only works if user entity was
created a while ago. Hence, we should work to remove all the cases where we create
a user entity from an email without an oauth user_id.

https://cloud.google.com/datastore/docs/concepts/transactions#isolation_and_consistency

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -15,6 +15,7 @@ import 'package:retry/retry.dart';
 import 'package:uuid/uuid.dart';
 
 import '../shared/configuration.dart';
+import '../shared/datastore_helper.dart';
 import '../shared/exceptions.dart';
 import '../shared/redis_cache.dart' show cache;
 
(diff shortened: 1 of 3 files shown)


app/lib/account/backend.dart
```diff
@@ -290,79 +291,111 @@ class AccountBackend {
...
-          return user;
-        });
-        return updatedUser;
+    final emptyKey = _db.emptyKey;
+
+    // Attempt to lookup the user, the common case is that the user exists.
+    // If the user exists, it's always cheaper to lookup the user outside a
+    // transaction.
+    // TODO: This should be cached.
+    {
```
isoos: this code block / indent is really weird, remove?

app/lib/shared/datastore_helper.dart
```diff
@@ -0,0 +1,62 @@
+import 'package:gcloud/db.dart';
```
isoos: nit: license

app/lib/shared/datastore_helper.dart
```diff
@@ -0,0 +1,62 @@
...
+import 'package:retry/retry.dart';
+
+class TransactionWrapper {
```
isoos: Add comment what this does.

app/lib/shared/datastore_helper.dart
```diff
@@ -0,0 +1,62 @@
...
+  final Transaction _tx;
+  bool _mutated = false;
+
+  TransactionWrapper._(this._tx);
+
+  Future<List<T>> lookup<T extends Model>(List<Key> keys) =>
```
isoos: let's add `@override` annotations here

app/lib/shared/datastore_helper.dart
```diff
@@ -0,0 +1,62 @@
...
+    _mutated = true;
+    _tx.queueMutations(inserts: inserts, deletes: deletes);
+  }
+}
+
+/// Call [fn] with a [TransactionWrapper] that is either committed or
+/// rolled back when [fn] returns.
+Future<T> withTransaction<T>(
+  DatastoreDB db,
+  Future<T> Function(TransactionWrapper) fn,
```
isoos: This could be `Future<T> Function(Transaction) fn`

app/lib/shared/datastore_helper.dart
jonasfj: It's intentionally not a subclass..  I created the wrapper to hide `commit` and `rollback` since I don't think we need those.

app/lib/shared/datastore_helper.dart
jonasfj: Yes, but `Transaction` exposes `commit()` and `rollback()` which this function calls on your behalf.

app/lib/shared/datastore_helper.dart
isoos: Ah, my bad.

app/lib/account/backend.dart
jonasfj: Moved the first part to separate helper method.

app/lib/account/backend.dart
```diff
@@ -289,80 +289,118 @@ class AccountBackend {
...
+        OAuthUserID,
+        id: auth.oauthUserId,
+      );
+
+      // Check that the user doesn't exist in this transaction.
+      final oauthUserMapping =
+          await tx.lookupOrNull<OAuthUserID>(oauthUserIdKey);
+      if (oauthUserMapping != null) {
+        // If the user does exist we can just return it.
+        return await _db.lookupValue<User>(
```
sigurdm: Consider putting this inside the transaction

app/lib/account/backend.dart
```diff
@@ -289,80 +289,118 @@ class AccountBackend {
...
+            OAuthUserID()
+              ..parentKey = emptyKey
+              ..id = auth.oauthUserId
+              ..userIdKey = user.key,
+          );
           return user;
-        });
-        return updatedUser;
+        }
+        // We have a user with the same userId, this is not ideal.
```
sigurdm: Consider removing this comment. It is redundant given the error message

app/lib/shared/datastore_helper.dart
```diff
@@ -0,0 +1,75 @@
...
+      }
+    }
+  });
+}
+
+/// Call [fn] with a [TransactionWrapper] that is either committed or
+/// rolled back when [fn] returns, and retried if [fn] fails.
+Future<T> withRetryTransaction<T>(
+  DatastoreDB db,
+  Future<T> Function(TransactionWrapper) fn,
```
sigurdm: I think the argument needs a name

sigurdm: Nevermind

jonasfj: no, but giving a name won't harm :)

PR #2951, opened 2019-10-25
Fix flow of getting pub API client

### LINKED ISSUE #2929: Can't sign-in to pub.dev
It seems I can't sign-in to pub.dev, when I click the button to do so, the Google login popup pops up, I then choose the account I want to sign-in with, the popup closes, and nothing at all happens.

### What I tried so far : 

- Refreshing the page
- Disabling my AdBlocker

**EDIT** : This bug is most likely because Google Chrome Dev breaks something in the login flow

szakarias:
In the current setup we could end up with a GoogleUser that wasn't signed in. 

This might fix #2929. We are not able to reproduce.

### DIFF
--- a/pkg/web_app/lib/src/account.dart
+++ b/pkg/web_app/lib/src/account.dart
@@ -52,18 +52,32 @@ PubApiClient get client {
     await _initialized.future;
 
     var user = getAuthInstance().currentUser.get();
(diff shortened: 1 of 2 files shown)


szakarias:
Thanks @isoos @jonasfj :-)

PR #2952, opened 2019-10-25
Merging user data (part 1)

isoos:


app/lib/shared/user_merger.dart
```diff
@@ -0,0 +1,191 @@
...
+    final query = _db.query<User>();
+    final counts = <String, int>{};
+    await for (final user in query.run()) {
+      if (user.oauthUserId == null) continue;
+      counts[user.oauthUserId] ??= 0;
+      counts[user.oauthUserId]++;
+    }
+    final result = counts.keys.where((k) => counts[k] > 1).toList();
+    print('$result OAuthUserID with more than one User.');
+    return result;
```
jonasfj: I think a better remedy is to iterate through all users.. and for each user:
 1. If `user.oauthUserId == null` then continue,
 2. Load the `OAuthUserID` entity and check if it points back to `user.user_id`, if not, then we have a problem.

Then we can make this a stream of userIds that has issues. Hence, we don't need to load all users at once.

app/lib/shared/user_merger.dart
```diff
@@ -0,0 +1,191 @@
...
+  /// Migrates data for User merge.
+  Future mergeUser(String fromUserId, String toUserId) async {
+    print('Merging User: $fromUserId -> $toUserId');
+
+    // Package
+    await _processConcurrently(
+      _db.query<Package>()..filter('uploaders =', fromUserId),
+      (Package m) async {
+        await _db.withTransaction((tx) async {
+          final p = (await tx.lookup<Package>([m.key])).single;
```
jonasfj: Checkout `lookupValue` :)

app/lib/shared/user_merger.dart
```diff
@@ -0,0 +1,191 @@
...
+    );
+
+    // PublisherMember
+    await _processConcurrently(
+      _db.query<PublisherMember>()..filter('userId =', fromUserId),
+      (PublisherMember m) async {
+        await _db.withTransaction((tx) async {
+          final pm = (await tx.lookup<PublisherMember>([m.key])).single;
+          if (pm.userId == fromUserId) {
+            pm.userId = toUserId;
```
jonasfj: This won't work, the entity also has `id = userId`, thus, we have to delete and create the entity.

app/lib/shared/user_merger.dart
isoos: We will need to iterate through all users anyway. We either lookup and check the mapping or keep a count. Once we found a problem, we'll need to load all the related User entities to check the e-mail match. I don't see much limit in the current approach, let's keep it for now.

app/lib/shared/user_merger.dart
```diff
@@ -0,0 +1,225 @@
...
+
+    final query = _db.query<User>()..filter('oauthUserId =', oauthUserId);
+    final users = await query.run().toList();
+    print('Users: ${users.map((u) => u.userId).join(', ')}');
+
+    final mapping = await _db.lookupValue<OAuthUserID>(
+        _db.emptyKey.append(OAuthUserID, id: oauthUserId));
+    print('Primary User: ${mapping.userId}');
+    if (!users.any((u) => u.userId == mapping.userId)) {
+      throw Exception('Primary User is missing!');
```
jonasfj: nit. `Error` not exception, since this is not something anyone should catch.

app/lib/account/models.dart
```diff
@@ -234,6 +234,22 @@ class Consent extends db.Model {
   bool shouldNotify() =>
       notificationCount == 0 ||
       DateTime.now().toUtc().isAfter(nextNotification);
+
+  /// Returns a new [Consent] object with a new parent.
+  /// Should be used only for merging users.
+  Consent changeParentUserId(String userId) {
```
jonasfj: Can't we make this a utility method in `user_merger.dart`. Once we're done with `user_merger.dart` it might be better to delete it, rather than maintain it?

app/lib/publisher/models.dart
```diff
@@ -99,4 +99,16 @@ class PublisherMember extends db.ExpandoModel {
   /// One of [PublisherMemberRole].
   @db.StringProperty(required: true)
   String role;
+
+  /// Returns a new [PublisherMember] object with a new parent.
+  /// Should be used only for merging users.
+  PublisherMember changeParentUserId(String userId) {
```
jonasfj: Can't we make this a utility method in `user_merger.dart`. Once we're done with `user_merger.dart` it might be better to delete it, rather than maintain it?

app/lib/shared/user_merger.dart
```diff
@@ -0,0 +1,225 @@
...
+          tx.queueMutations(inserts: [h]);
+          await tx.commit();
+          print('Updated History(${h.id})');
+          print(fromJson);
+          print(h.eventJson);
+        });
+      },
+    );
+
+    if (_deleteUsers) {
```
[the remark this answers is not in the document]
jonasfj: let me clarify... Let's avoid the `_deleteUsers` option and just always delete the user..

Creating a dry-run option is a much harder problem :)

app/lib/shared/user_merger.dart
jonasfj: Okay, I don't see a huge problem... It just seems like we could make a more robust solution the other way, but I'm fine this is hopefully a one-off :)

app/lib/account/models.dart
isoos: What happens if we add a new field to `Consent`? Chances are that we forget to update it in `user_merger`. I'd rather keep it with the class, and if we are done with `user_merger`, we shall (eventually) delete these too.

app/lib/shared/user_merger.dart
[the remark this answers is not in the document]
isoos: Removing the `_deleteUsers` flag. I've put it in when the `History` update was not implemented yet, no longer need.

PR #2953, opened 2019-10-25
Fix 2905 by allowing arbitrary level domains

jonasfj:


jonasfj:
I tried to create a publisher which works, but I don't have any subdomains in search console.

Issue #2945, opened 2019-10-24
Migrate uploader invites to use consent-flow


jonasfj:
Once this is deployed, we should also ensure that oauth-flow doesn't have options configured that can be deactivated.

PR #2962, opened 2019-10-28
Migrate uploader invites to Consent: sending the e-mail.

isoos:
- #2945
- I think it would be not too hard to modify `Consent` to use non-parented `Key`, so that we can delay `User` creation.


jonasfj:
> I think it would be not too hard to modify Consent to use non-parented Key, so that we can delay User creation.

Go for it, but if we want to mess around with the consent flow, we should probably extend out integration tests (and deployment tests) to cover invite/remove publisher member.

isoos:
> Go for it, but if we want to mess around with the consent flow, we should probably extend out integration tests (and deployment tests) to cover invite/remove publisher member.

++

Issue #2957, opened 2019-10-26
Example tab should link to source


filiph:
URL: https://pub.dev/packages/path_provider#-example-tab-

I like how, when you have an `example/example.dart` file, its contents are immediately rendered in the Example tab on pub.dev. But many packages and plugins nowadays use the approach of having an app, not a single file, in `example/`.

This means that the example tab will often show the contents of the README in `example`, which is not helpful at all.

I think the logic should be the following:

* Do we have a single `.dart` file to show in `example`? Show it.
* In all other cases, try to link to the source. I.e. for most packages, that would be the github link to `example`.
* If there happens to be a README in `example`, show the contents below the link to the source.

sigurdm:
This would be nice. Only slight issue is that we cannot link to the exact commit.

isoos:
> Only slight issue is that we cannot link to the exact commit.

There are a few more failure cases there: git history can get rewritten (the specific commit will be gone), repository can be deleted... I think for now it is already an improvement to link to the file in the repo, and if we want to be pedantic about it, we might just serve the uploaded file in an archive browser.

isoos:
#2963 is not the final solution yet, as in case of [`path_provider`](https://github.com/flutter/plugins/blob/master/packages/path_provider/pubspec.yaml), the `homepage` is the specific URL, the `repository` is derived, and we are using the later as the `baseUrl` in the code. I'll need to review a few examples if it is better to ignore the repository if it was derived.

PR #2963, opened 2019-10-29
Link to example file

isoos:
#2957

### DIFF
--- a/app/lib/frontend/templates/package.dart
+++ b/app/lib/frontend/templates/package.dart
@@ -6,6 +6,7 @@ import 'dart:convert';
 
 import 'package:client_data/page_data.dart';
 import 'package:meta/meta.dart';
+import 'package:pana/src/download_utils.dart' show getRepositoryUrl;
 
 import '../../analyzer/analyzer_client.dart';
 import '../../package/models.dart';
@@ -309,9 +310,12 @@ List<Tab> _pkgTabs(
     final exampleFilename = selectedVersion.example.filename;
     renderedExample = renderFile(selectedVersion.example, baseUrl);
     if (renderedExample != null) {
-      renderedExample = '<p style="font-family: monospace">'
-          '<b>${htmlEscape.convert(exampleFilename)}</b>'
-          '</p>\n'
+      final url = getRepositoryUrl(baseUrl, exampleFilename);
+      final escapedName = htmlEscape.convert(exampleFilename);
+      final link = url == null
+          ? escapedName
+          : '<a href="$url" target="_blank" rel="noopener noreferrer nofollow">$escapedName</a>';
+      renderedExample = '<p style="font-family: monospace"><b>$link</b></p>\n'
           '$renderedExample';
     }
   }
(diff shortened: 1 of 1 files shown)


app/lib/frontend/templates/package.dart
```diff
@@ -309,9 +309,12 @@ List<Tab> _pkgTabs(
     final exampleFilename = selectedVersion.example.filename;
     renderedExample = renderFile(selectedVersion.example, baseUrl);
     if (renderedExample != null) {
-      renderedExample = '<p style="font-family: monospace">'
-          '<b>${htmlEscape.convert(exampleFilename)}</b>'
-          '</p>\n'
+      final url = getRepositoryUrl(baseUrl, exampleFilename);
```
sigurdm: Handle getRepositoryUrl returning null?

sigurdm: Sorry you do that below

app/lib/package/models.dart
```diff
@@ -550,4 +550,15 @@ class PackageLinks {
       issueTrackerUrl: issueTrackerUrl,
     );
   }
+
+  /// URL to use as for linking the referenced resource relative to.
```
sigurdm: Broken sentence

app/lib/frontend/templates/package.dart
```diff
@@ -309,9 +309,12 @@ List<Tab> _pkgTabs(
...
+      final escapedName = htmlEscape.convert(exampleFilename);
+      final link = url == null
+          ? escapedName
+          : '<a href="$url" target="_blank" rel="noopener noreferrer nofollow">$escapedName</a>';
+      renderedExample = '<p style="font-family: monospace"><b>$link</b></p>\n'
```
sigurdm: Side question: why don't we style the link with css?

app/lib/package/models.dart
```diff
@@ -550,4 +550,15 @@ class PackageLinks {
...
+  String get baseUrl {
+    String url = repositoryUrl ?? homepageUrl;
+    if (repositoryUrl != null &&
+        homepageUrl != null &&
+        homepageUrl.startsWith(repositoryUrl)) {
+      url = homepageUrl;
```
jonasfj: Hmm... I'm not sure this is a good idea.

We should always use the `repositoryUrl` to identify the repository, **only** if there is no `repositoryUrl`, then we should try to derive one from the the `homepage`.

Adding a heuristic that tries to derive sub-folders from the `homepage` is just more confiusing. If we want that, we should define how to point to sub-folders in the `repositoryUrl` (I'm not sure this is easy to do, we would need to update docs for `pubspec.yaml`).

I think better linking to repository source is a project we could solve, but it should be a project that involves known git hash of the version too.

jonasfj: IMO; if we want to do a heuristtic for this, we should do it the same way we do for `LICENSE`. Though ideally, we should implement code listing.

isoos: I'll just remove the commit in question from this PR, and address that part later.

app/lib/frontend/templates/package.dart
isoos: It is a good PR to do :)

PR #2966, opened 2019-10-30
Remove warning about unsupported qualifier

### LINKED ISSUE #2845: `dart:` is not a qualifier
When I search for `dart:js` on pub.dev, it says `Your query contains an unsupported qualifier. Check the supported search qualifiers.` although `dart:js` is a valid library name. 

It thinks that `dart:` is a qualifier just like `dependency:`.

isoos:
Fixes #2845, and we won't need such warning when we'll have almost-freeform tags in search.


jonasfj:
Quick note. tags won't be free-form... we won't create new tag-patterns without a discussion around the naming. But this can be re-introduced if necessary.

Issue #2971, opened 2019-10-31
Consent flow without registering User entity upfront


isoos:
- [ ] extending `Consent` entity with `userId` and `email` fields, checking if it is for the current active user (maybe more in the future)
- [ ] new `Consent` entities are created without the parent `User`
- [ ] remove all references to parent `User`

PR #2974, opened 2019-10-31
Headless Consent entities.

isoos:
- #2971
- dedup id includes further attributes (if they are present)
- `User` entities are now created only at the time of accepting the consent, and not upfront
- ignoring the case of user merges with old pending Consent requests (throws an exception during merge)

### DIFF
--- a/app/lib/account/consent_backend.dart
+++ b/app/lib/account/consent_backend.dart
@@ -89,16 +89,21 @@ class ConsentBackend {
   /// - if it was sent recently, do nothing.
   Future<api.InviteStatus> invite({
     @required String userId,
+    @required String email,
     @required String kind,
     @required List<String> args,
   }) async {
     return retry(() async {
       final activeUser = await requireAuthenticatedUser();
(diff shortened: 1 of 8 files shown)


app/lib/account/consent_backend.dart
```diff
@@ -182,17 +190,20 @@ class ConsentBackend {
...
     return c;
   }
 
   Future _accept(Consent consent) async {
     final action = _actions[consent.kind];
     await retry(
       () async {
-        await action?.onAccept(
-            consent.fromUserId, consent.userId, consent.args);
+        await action?.onAccept(consent);
```
jonasfj: Unrelated, but wouldn't be healthy to handle the case where `action==null` explicitly, note: I think this is best in a follow up PR... or maybe I'm missing something. Can't we do the delete in a separate thing, so the intentions are clear?

isoos: > Can't we do the delete in a separate thing, so the intentions are clear?

what do you mean by that?

PR #2983, opened 2019-11-01
Clean up package flag's null-value handling

isoos:


app/lib/package/models.dart
```diff
@@ -63,11 +63,31 @@ class Package extends db.ExpandoModel {
   @db.StringListProperty()
   List<String> uploaders;
 
-  @db.BoolProperty()
-  bool isDiscontinued;
+  /// Set to `true` if package is discontinued, may otherwise be `null` or `false`.
+  ///
+  /// Use [isDiscontinued] to avoid `null` checking.
+  @db.BoolProperty(propertyName: 'isDiscontinued')
```
jonasfj: Le'ts check this is a bool in the integrity check, add a backfill utility and update changelog to say we run this next time... so we can eventually get rid of this.

isoos: ++

PR #2986, opened 2019-11-04
tags support in search index and query

isoos:


app/lib/search/search_service.dart
```diff
@@ -53,6 +53,8 @@ class PackageDocument {
 
   final List<String> platforms;
 
+  final List<String> tags;
```
jonasfj: If this can be `null` during transition, I propose that we document that this might be `null` as we are upgrading the index.
Leaving a TODO that all `null` handling should be removed in the future.

app/lib/search/search_service.dart
```diff
@@ -108,6 +111,7 @@ class PackageDocument {
       doNotAdvertise: doNotAdvertise,
       supportsOnlyLegacySdk: supportsOnlyLegacySdk,
       platforms: platforms?.map(internFn)?.toList(),
+      tags: tags?.map(internFn)?.toList(),
```
jonasfj: If `tags` can be `null` we should documented as noted above, and leave a TODO to remove the `null` handling when `tags` cannot be null anymore.

app/lib/search/search_service.dart
```diff
@@ -418,6 +434,56 @@ class SearchQuery {
...
+class TagsPredicate {
+  /// tag -> required (true) | forbidden (false) map
+  final _values = <String, bool>{};
+
+  TagsPredicate([Map<String, bool> values]) {
+    if (values != null) _values.addAll(values);
+  }
+
+  /// Parses [values] passed via Uri.queryParameters
+  TagsPredicate.parseQueryValues(List<String> values) {
```
jonasfj: Can we annotate this `@visiblelForTesting` and consider it for a few other methods too?

app/lib/search/search_service.dart
```diff
@@ -418,6 +434,56 @@ class SearchQuery {
...
+    for (String tag in values ?? const <String>[]) {
+      bool required = true;
+      if (tag.startsWith('-')) {
+        tag = tag.substring(1);
+        required = false;
+      } else if (tag.startsWith('+')) {
```
jonasfj: But this only supports tags for the purpose of query string filtering?

And we allow `+` but don't require it. And having `+tag` has no special meaning. Maybe we just don't support that syntax yet. And stick to `-tag` and `tag`.

app/lib/search/search_service.dart
```diff
@@ -418,6 +434,56 @@ class SearchQuery {
   }
 }
 
+/// Filter conditions on tags.
+class TagsPredicate {
+  /// tag -> required (true) | forbidden (false) map
+  final _values = <String, bool>{};
```
jonasfj: nits, but two sets or arrays called `withTags` and `withoutTags`... or `requiredTags` and `negatedTags` might be easier...

If the same tags is in both we don't have to have special handling we just make it filter to nothing :)

app/lib/search/search_service.dart
isoos: This is used in `SearchQuery.fromServiceUrl`.

app/lib/search/search_service.dart
isoos: I think people will use `+`, and it is cheap to support it.

app/lib/search/search_service.dart
```diff
@@ -383,7 +384,7 @@ class SearchQuery {
         order == null &&
         parsedQuery.packagePrefix == null &&
         (platform == null || platform.isEmpty) &&
-        (tagsPredicate == null || tagsPredicate._values.isEmpty);
+        (tagsPredicate == null || tagsPredicate.isEmpty);
     if (isEmpty) return false;
```
jonasfj: Why is an empty search query not valid?

app/lib/search/search_service.dart
```diff
@@ -453,34 +458,34 @@ class TagsPredicate {
...
-      if (entry.value && !tags.contains(entry.key)) {
-        return false;
-      }
-      if (!entry.value && tags.contains(entry.key)) {
-        return false;
-      }
+    for (final tag in requiredTags ?? const <String>[]) {
+      if (!tags.contains(tag)) return false;
+    }
+    for (final tag in negatedTags ?? const <String>[]) {
```
jonasfj: nit: couldn't we just give a default value in the constructor?

app/lib/search/search_service.dart
isoos: I remember that at one point it was used to decide between regular pagination and search requests, but apparently that code is gone now, and this method is only used in tests. I'll do a follow-up cleanup once this PR lands.

PR #2987, opened 2019-11-04
Cleanup API implementation

jonasfj:


app/lib/package/backend.dart
```diff
@@ -242,13 +246,10 @@ class PackageBackend {
...
       );
 
-      tx.queueMutations(
-        inserts: [
-          if (hasFlagChange) history,
-          p,
-        ],
-      );
-      await tx.commit();
+      if (hasFlagChange) {
```
isoos: this part needs to be re-based on master :(

isoos:
you'll need to run `pub run build_runner build`

isoos:
> you'll need to run pub run build_runner build

 and probably copy the generated file to the pkg/web_app/lib/src too

PR #2990, opened 2019-11-05
Tags support in models.

isoos:


app/lib/package/models.dart
```diff
@@ -148,6 +148,23 @@ class Package extends db.ExpandoModel {
...
+  List<String> getTags() {
+    return <String>[
+      if (isDiscontinued) PackageTags.isDiscontinued,
+      if (isNewPackage()) PackageTags.isNew,
+    ];
+  }
+}
+
+/// Collection of package-related tags.
+abstract class PackageTags {
```
jonasfj: Let's put this in a separate file like `package_tags.dart`.

app/lib/package/models.dart
```diff
@@ -148,6 +148,23 @@ class Package extends db.ExpandoModel {
...
+  /// Package is marked discontinued.
+  static const String isDiscontinued = 'is:discontinued';
+
+  /// Package is less than 30 days old
+  static const String isNew = 'is:new';
```
jonasfj: Should we consider calling this `is:recent`. Also we should document if this is the package that's less than 30 days or the latest version of the package.

app/lib/package/models.dart
```diff
@@ -148,6 +148,23 @@ class Package extends db.ExpandoModel {
   }
 
   bool isNewPackage() => created.difference(DateTime.now()).abs().inDays <= 30;
+
+  /// List of tags from the flags on the current [Package] entity.
+  List<String> getTags() {
+    return <String>[
+      if (isDiscontinued) PackageTags.isDiscontinued,
+      if (isNewPackage()) PackageTags.isNew,
```
jonasfj: Let's add a TODO: for `is:not-advertized` and `is:dart-1`, `publisher:<publisherId>` and `uploader:<...>`.

With those in-place (later) we can remove a lot of the stuff we have in search.

app/lib/package/package_tags.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Collection of package-related tags.
+abstract class PackageTags {
+  /// Package is marked discontinued.
+  static const String isDiscontinued = 'is:discontinued';
+
+  /// Package is less than 30 days old.
```
jonasfj: > The first version of the package was published less than 30 days ago.

Is that what this means?

PR #2992, opened 2019-11-05
Deep repository URLs.

### LINKED ISSUE #2957: Example tab should link to source

isoos:
- When inferring the repository URLs, we stopped at `/user/repo` level, however in many cases, especially in Flutter plugins, we have deep URLs set as `homepage`. We should keep these deep URLs and not shorten them. (Maybe we should always-keep for GitHub and GitLab URLs, not yet sure).
- Also fixes #2957.

### DIFF
--- a/app/lib/shared/urls.dart
+++ b/app/lib/shared/urls.dart
@@ -102,12 +102,17 @@ String inferRepositoryUrl(String baseUrl) {
     return null;
   }
   if (uri.host == 'github.com' || uri.host == 'gitlab.com') {
-    final segments = uri.pathSegments.take(2).toList();
-    if (segments.length != 2) {
+    if (uri.pathSegments.length < 2) {
       return null;
     }
-    return Uri(scheme: uri.scheme, host: uri.host, pathSegments: segments)
-        .toString();
+    if (uri.pathSegments.length >= 4 &&
+        uri.pathSegments[2] == 'tree' &&
+        uri.pathSegments[3] == 'master') {
+      return uri.toString();
+    } else {
+      final segments = uri.pathSegments.take(2).toList();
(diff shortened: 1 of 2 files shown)


jonasfj:
I'm in favor of making this inference more powerful, but if people correctly specify `repository` they can't make deep references in the repository.. So it seems weird that we can do this with the `homepage`, but not for `repository`.

I'm not sure how to specify a git repository URL to a subfolder in a repository.. also we should document this. This will force mono-repo package to use `homepage` for the repository link as using `repository` won't work as well.

jonasfj:
Maybe we need to make it:
```yaml
repository:
  url: https://myhost.com/repo.git/
  path: /my/folder/in/repo
``` 

So that we have the same support in both `repository` and `homepage`..

Notably, I think we need to also keep `repository` as a key that can be cloned -- in case we want `pana` to analyze repositories and tags to find matching packages.

isoos:
>  but if people correctly specify repository they can't make deep references in the repository.. 

Why is that? You could use `repository: https://github.com/[deep-url]` without any issue. This inferring happens only if you have set `homepage` but no `repository` key in the `pubspec.yaml`.

PR #2993, opened 2019-11-05
Add thumb-up button and likes functionality in the UI.

szakarias:
The feature is added behind the experimental flag.


app/lib/frontend/handlers/package.dart
```diff
@@ -95,8 +95,16 @@ Future<shelf.Response> packageVersionsListHandler(
   final isAdmin =
       await packageBackend.isPackageAdmin(package, userSessionData?.userId);
 
+  final bool isLiked = (userSessionData == null)
+      ? false
+      : await accountBackend.getPackageLikeStatus(
```
isoos: It seems to be unnecessary to lookup `User` just to use its already known id to lookup the like status. Let's refactor `getPackageLikeStatus` to use only the id?

app/lib/frontend/templates/views/shared/detail/header.mustache
```diff
@@ -11,5 +11,35 @@
...
+            class="mdc-icon-button__icon"/>
+          <img
+            height="18"
+            width="18"
+            src="/static/img/thumb_up-filled-24px.svg"
+            class="mdc-icon-button__icon mdc-icon-button__icon--on"/>
+        </button>
+       <span id="likes">{{like_count}}</span></div></div>
+    {{/has_likes}}
+</div>  </div>
```
isoos: nit: indent

app/lib/frontend/templates/views/shared/detail/header.mustache
```diff
@@ -11,5 +11,35 @@
...
-</div>
+    {{#has_likes}}
+    <div class="-likes" >
+      <div>
+        {{#is_liked}}
+        <button
+          id="thumb-up"
+          class="mdc-icon-button mdc-icon-button--on"
+          aria-label="Unstar this item"
+          aria-hidden="true" aria-pressed="true">
```
isoos: nit: the other button has this on separate lines

pkg/web_app/lib/likes.dart
```diff
@@ -0,0 +1,68 @@
...
+
+  if (toggleButton == null) return;
+
+  final iconButtonToggle = MDCIconButtonToggle(toggleButton);
+  int likesDelta = 0;
+
+  String likesString() {
+    final likesCount = pageData.pkgData.likes + likesDelta;
+    return '${NumberFormat.compact().format(likesCount)}'
+        ' ${likesCount == 1 ? 'Like' : 'Likes'}';
```
isoos: Wondering: why not lowercase 'like' / 'likes'?

pkg/web_css/pubspec.lock
```diff
@@ -1,5 +1,5 @@
 # Generated by pub
-# See https://dart.dev/tools/pub/glossary#lockfile
+# See https://www.dartlang.org/tools/pub/glossary#lockfile
```
isoos: When did this change in `pub` client? The URL seems to be off, as we will redirect it first to pub.dev and then away from the site...

pkg/web_app/lib/likes.dart
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  iconButtonToggle.listen(MDCIconButtonToggle.changeEvent, (Event e) {
+    toggleButton.blur();
+    if (iconButtonToggle.on) {
+      // The button has shifted to on.
+      likesDelta++;
+      likes.innerHtml = likesString();
```
jonasfj: Can we use `textContent` instead? If we're only updating text it's much better to use `textContent` as `innerHtml` could be leverage to do HTML injection (assuming we have another bug somewhere else)..

app/lib/frontend/templates/views/shared/detail/header.mustache
```diff
@@ -11,5 +11,35 @@
...
+        {{/is_liked}}
+          <img
+            height="18"
+            width="18"
+            src="/static/img/thumb_up-24px.svg"
+            class="mdc-icon-button__icon"/>
+          <img
+            height="18"
+            width="18"
+            src="/static/img/thumb_up-filled-24px.svg"
```
jonasfj: This is an ugly hack, but if we rewrite it as we did with
`{{{static_assets.img__verified-publisher-blue_svg}}}` then we get better caching :)

Hmm, look up the code around `static_assets`, and maybe rename the image file to something that doesn't have underscore just to make your life easier :)

app/lib/frontend/templates/package_versions.dart
```diff
@@ -20,6 +20,7 @@ import 'package.dart';
 /// Renders the `views/pkg/versions/index` template.
 String renderPkgVersionsPage(
     Package package,
+    bool isLiked,
```
jonasfj: Maybe document that `isLiked == null` implies that the page does not have a like button at all.

isoos:
I think this looks good now, but you'll need to (a) regenerate the templates test's golden files (run the tests with https://github.com/dart-lang/pub-dev/blob/master/app/test/frontend/templates_test.dart#L39 set to true), and (b) check why other tests are timing out..

szakarias:
This is on hold until https://github.com/dart-lang/pub-dev/pull/2998 has landed. (The reason for the test that are timing out.)

jonasfj:
https://github.com/dart-lang/pub-dev/pull/2998 has landed.

Issue #2944, opened 2019-10-24
Automated website testing


jonasfj:
It would be nice with some automated website testing, probably with puppeteer.

Test UI elements like:
 * search
 * tabs
 * ...

Login and other interactive features would also be neat to test.

PR #2994, opened 2019-11-06
Minimal browser test with puppeteer.

isoos:
#2944

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,12 @@ language: dart
 services:
   - "redis-server"
 
+addons:
+  chrome: stable
+env:
+  global:
+    - CHROME_BIN=/usr/bin/google-chrome
+
 jobs:
   include:
     - stage: smoke_test
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,8 @@ language: dart
 # Custom configuration
 services:
   - "redis-server"
-
 addons:
   chrome: stable
-env:
-  global:
-    - CHROME_BIN=/usr/bin/google-chrome
 
 jobs:
   include:
--- a/pkg/pub_integration/lib/src/headless_env.dart
+++ b/pkg/pub_integration/lib/src/headless_env.dart
@@ -8,6 +8,8 @@ import 'package:puppeteer/puppeteer.dart';
 // ignore: implementation_imports
 import 'package:puppeteer/src/page/page.dart' show ClientError;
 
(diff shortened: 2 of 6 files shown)


pkg/pub_integration/lib/src/headless_env.dart
```diff
@@ -0,0 +1,86 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:puppeteer/puppeteer.dart';
+// ignore: implementation_imports
+import 'package:puppeteer/src/page/page.dart' show ClientError;
+
+class HeadlessEnv {
```
jonasfj: Can we give this a online description?

.travis.yml
```diff
@@ -5,6 +5,12 @@ language: dart
 services:
   - "redis-server"
 
+addons:
+  chrome: stable
+env:
+  global:
+    - CHROME_BIN=/usr/bin/google-chrome
```
jonasfj: Let's detect this in our test framework...

pkg/pub_integration/lib/src/headless_env.dart
```diff
@@ -0,0 +1,86 @@
...
+  final clientErrors = <ClientError>[];
+
+  HeadlessEnv({Directory tempDir})
+      : tempDir =
+            tempDir ?? Directory.systemTemp.createTempSync('pub-headless');
+
+  Future _startBrowser() async {
+    if (_browser != null) return;
+    final chromeBin = Platform.environment['CHROME_BIN'];
+    if (chromeBin != null && !File(chromeBin).existsSync()) {
```
jonasfj: Can we please do some bare minimum for automatic resolution, I think we can just search `PATH` or use `google-chrome` if we don't need a absolute path.

pkg/pub_integration/lib/src/headless_env.dart
```diff
@@ -0,0 +1,86 @@
...
+      throw StateError(
+          'Chrome executable `$chromeBin` (specified by CHROME_BIN) does not exists.');
+    }
+    final userDataDir = tempDir.createTempSync('user');
```
jonasfj: let's be `async` when it's easy and cheap :)

.travis.yml
isoos: Keeping it simple for now: I've hard-coded the path in the code, and added a TODO to make this better.

PR #2998, opened 2019-11-07
Fixed timeout and retry logic for transactions

jonasfj:


app/lib/shared/datastore_helper.dart
```diff
@@ -66,10 +67,39 @@ Future<T> withTransaction<T>(
...
+/// This does not retry [ResponseException].
 Future<T> withRetryTransaction<T>(
   DatastoreDB db,
   Future<T> Function(TransactionWrapper tx) fn,
 ) =>
-    retry<T>(() => withTransaction<T>(db, fn));
+    _transactionRetrier.retry<T>(
+      () => withTransaction<T>(db, fn),
+      // Never retry a ResponseException
+      retryIf: (e) => e is! ResponseException,
```
jonasfj: @isoos, do you agree it's sane to retry all exceptions other than `ResponseException`. Or would it be safer to enumerate the specific gcloud, datastore, appengine, grpc, and IO exceptions we want to retry.

I have mixed feelings because the set of exceptions we want to retry seems to change over time as we move to a different grpc library. Maybe we should consider doing this after we've upgraded `package:appengine`

isoos: Yes, I think it is sane. As a safeguard, let's log the exceptions that we retry on. Then after a time let's check the logs and see what we have there, and if they match our expectations...

app/lib/shared/datastore_helper.dart
```diff
@@ -66,10 +67,39 @@ Future<T> withTransaction<T>(
   });
 }
 
+/// Transaction retry options.
+///
+/// These should be retried within the 30s timeout for sending an inital header.
+/// We suspect that AppEngine Flexible has such an timeout, because it uses GCP
```
szakarias: an -> a

PR #2999, opened 2019-11-07
PackageLinks.baseUrl

### LINKED ISSUE #2957: Example tab should link to source

### LINKED ISSUE #3003: Links generated in the "Example" tab are incorrect
URL: https://pub.dev/packages/android_alarm_manager#-example-tab-

This is pulling the `packages/android_alarm_manager/example/README.md` and attempting to link back to it, but the link is broken. So users who see this page are stuck with a mostly empty README and no clear way to find the actual example app.

- Expected link: https://github.com/flutter/plugins/blob/master/packages/android_alarm_manager/example/README.md
- Actual link: https://github.com/flutter/plugins/blob/master/example/README.md

I expect that the problem has something to do with `android_alarm_manager` being in a subdirectory of the `flutter/plugins` repo, instead of a top level folder in its own repo.

This appears to be impacting all of the `flutter/plugins` packages.

/cc @csells

isoos:
- Fixes #2957 (and other places where we are using the wrong URL for deep repositories)
- Fixes #3003
- keeps the `repositoryUrl` separate, so that we treat it as the url of the git repo

### DIFF
--- a/app/lib/shared/urls.dart
+++ b/app/lib/shared/urls.dart
@@ -143,6 +143,19 @@ String inferIssueTrackerUrl(String baseUrl) {
   return null;
 }
 
+/// Infer base URL that can be used to link files from.
+String inferBaseUrl({String homepageUrl, String repositoryUrl}) {
+  String baseUrl = repositoryUrl ?? homepageUrl;
+  // In a few cases people specify only a deep repository URL for their
+  // package (e.g. a monorepo for multiple packages). While we default the
+  // base URL to be the repository URL, this check allows us to use the deep
+  // URL for linking.
+  if (homepageUrl != null && homepageUrl.startsWith(baseUrl)) {
+    baseUrl = homepageUrl;
+  }
+  return baseUrl;
+}
+
 /// Infer the hosting/service provider for a given URL.
 String inferServiceProviderName(String url) {
   if (url == null) {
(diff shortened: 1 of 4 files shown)


jonasfj:
Just to understand it right, this propose an heuristic that:
 * If `repository` is a substring of `homepage`, then the reminder of `homepage` is assumed to be the sub-folder in the repository.

jonasfj:
Hmm, maybe it's fine... this becomes irrelevant if one day we implement source listings.

app/lib/shared/urls.dart
```diff
@@ -143,6 +143,19 @@ String inferIssueTrackerUrl(String baseUrl) {
   return null;
 }
 
+/// Infer base URL that can be used to link files from.
+String inferBaseUrl({String homepageUrl, String repositoryUrl}) {
```
jonasfj: If `repositoryUrl` starts with `https://github.com/` or `https://gitlab.com/` and ends with `.git` then we remove `.git`.

jonasfj: Also remove `www` from github and gitlab.

PR #3002, opened 2019-11-07
Granular admin API

jonasfj:
API for managing ff and data model update, along with backfill, etc.

Also introduce a concept of admin permissions, because we really should segregate what
service accounts can access what resources.

In practice I don't think we'll be adding individual adminstrators in our configuration.
It takes a long time for us to update this configration as it's effectively hardcoded into the application.
Thus, we need to do a full deployment to add/remove admin permissions.

Instead we'll grant admin permissions to a service account, and use IAM in GCP Console
grant individual users permission to impersonate the service account. This makes calling with
admin credentials slightly complicated as an administrator will have to call the IAM API to
obtain a token impersonating the service account listed as administrator.
On the other hand this is very flexible. And once we have the setup for specifying which
service we'll be impersonating, it's pretty easy to call this way.

### DIFF
--- a/app/lib/admin/backend.dart
+++ b/app/lib/admin/backend.dart
@@ -13,10 +13,12 @@ import 'package:pool/pool.dart';
 
 import '../account/backend.dart';
 import '../account/models.dart';
+import '../package/backend.dart' show packageBackend;
 import '../package/models.dart';
 import '../publisher/models.dart';
 import '../shared/configuration.dart';
 import '../shared/exceptions.dart';
+import '../shared/datastore_helper.dart';
 
 final _logger = Logger('pub.admin.backend');
 final _continuationCodec = utf8.fuse(hex);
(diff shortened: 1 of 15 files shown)


app/lib/frontend/handlers/pubapi.dart
```diff
@@ -336,6 +336,19 @@ class PubApi {
     await adminBackend.removeUser(userId);
     return jsonResponse({'status': 'OK'});
   }
+
+  @EndPoint.get('/api/admin/packages/<package>/flutter-favorite')
```
isoos: Note for future API discussions: We have `/api/packages/<package>/options` for updating the `isDiscontinued` flag. I'd expect the admin API to also use `/options` but at the same time I understand the separation in this case. Once we have more APIs around, and before we mark them stable, we should do a bit of renaming to make them consistent.

app/lib/admin/backend.dart
```diff
@@ -247,4 +253,53 @@ class AdminBackend {
...
+    String packageName,
+  ) async {
+    await _requireAdminPermission(AdminPermission.manageFlutterFavorite);
+    final package = await packageBackend.lookupPackage(packageName);
+    if (package == null) {
+      throw NotFoundException.resource(packageName);
+    }
+
+    return api.FlutterFavoriteStatus(
+      isFlutterFavorite: package.isFlutterFavorite,
```
isoos: I will assume that we are not concerned about the existing entities returning null (before the backfill happens).

app/lib/package/models.dart
```diff
@@ -91,6 +91,9 @@ class Package extends db.ExpandoModel {
     doNotAdvertiseFlag = value;
   }
 
+  @db.BoolProperty()
+  bool isFlutterFavorite;
```
isoos: Add a TODO to make this required after the backfill updated the entities in Datastore.

PR #3005, opened 2019-11-08
Require dart 2.6

jonasfj:
This is probably why we had the downgrade to `0.0.19` because it was possible to run with an old SDK. Which we shouldn't allow.


isoos:
> This is probably why we had the downgrade to 0.0.19 because it was possible to run with an old SDK. Which we shouldn't allow.

0.0.20 has the SDK constraint:
https://pub.dev/packages/stream_transform#-analysis-tab-

Maybe this is a bug somewhere in the dependency resolution?

PR #3016, opened 2019-11-11
Better messages for Consent lookup and verification.

### LINKED ISSUE #3014: Loading consent for the wrong user shows 404
Loading a consent while logged in with the wrong user shows 404.

We should show that this is an invitation for a different user account, click "sign-out / sign-in" and use a user account that has the email this invitation was sent to.

Probably we should avoid leaking the email the invite was sent to.

isoos:
Fixes #3014.

### DIFF
--- a/app/lib/account/consent_backend.dart
+++ b/app/lib/account/consent_backend.dart
@@ -183,11 +183,11 @@ class ConsentBackend {
     // Checking that consent is for the current user.
     if (c.userIdOfConsent != null) {
       InvalidInputException.check(c.userIdOfConsent == user.userId,
-          'Consent.userId does not match current User.');
+          'This invitation is not for the user account currently logged in.');
     }
     if (c.email != null) {
-      InvalidInputException.check(
-          c.email == user.email, 'Consent.email does not match current User.');
+      InvalidInputException.check(c.email == user.email,
+          'This invitation is not for the user account currently logged in.');
     }
     return c;
   }
--- a/app/lib/account/consent_backend.dart
+++ b/app/lib/account/consent_backend.dart
@@ -160,8 +160,6 @@ class ConsentBackend {
   }
 
   /// Returns the [Consent] for [consentId] and checks if it is for [user].
-  ///
-  /// Returns null if the consent cannot be found.
   Future<Consent> _lookupAndCheck(String consentId, User user) async {
     // legacy Consent store: under the User entity
     final legacyKey = _db.emptyKey
--- a/app/lib/account/consent_backend.dart
+++ b/app/lib/account/consent_backend.dart
@@ -43,9 +43,6 @@ class ConsentBackend {
   Future<api.Consent> getConsent(String consentId, User user) async {
     InvalidInputException.checkUlid(consentId, 'consentId');
     final c = await _lookupAndCheck(consentId, user);
-    if (c == null) {
-      throw NotFoundException.resource('consent: $consentId');
-    }
     final action = _actions[c.kind];
     final activeAccountEmail =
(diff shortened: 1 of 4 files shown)


pkg/pub_integration/test/publisher_test.dart
```diff
@@ -43,8 +43,8 @@ void main() {
           },
           body: json.encode({'granted': true}),
         );
-        if (rs1.statusCode != 404) {
-          throw Exception('Expected status code 404, got: ${rs1.statusCode}');
+        if (rs1.statusCode != 400) {
```
sigurdm: Why not
```
expect(rs1.statusCode, 400);
```
?

pkg/pub_integration/test/pub_integration_test.dart
```diff
@@ -42,8 +42,8 @@ void main() {
           },
           body: json.encode({'granted': true}),
         );
-        if (rs1.statusCode != 404) {
-          throw Exception('Expected status code 404, got: ${rs1.statusCode}');
+        if (rs1.statusCode != 400) {
```
sigurdm: Why not

expect(rs1.statusCode, 400);
?

app/lib/account/consent_backend.dart
```diff
@@ -187,14 +178,18 @@ class ConsentBackend {
     final c = await _db.lookupValue<Consent>(
         _db.emptyKey.append(Consent, id: consentId),
         orElse: () => null);
-    if (c == null) return null;
+    if (c == null) {
+      throw NotFoundException.resource('consent: $consentId');
```
sigurdm: Update the dartdoc of this function. It can no longer return null.

app/lib/account/consent_backend.dart
```diff
@@ -187,14 +178,18 @@ class ConsentBackend {
...
+    }
 
     // Checking that consent is for the current user.
-    if (c.userIdOfConsent != null && c.userIdOfConsent != user.userId) {
-      return null;
+    if (c.userIdOfConsent != null) {
+      InvalidInputException.check(c.userIdOfConsent == user.userId,
+          'Consent.userId does not match current User.');
```
sigurdm: Are these strings user visible? I think they should not refer to fields in our internal representation.
What about:
`'This invitation is not for the user account currently logged in'`

app/lib/account/consent_backend.dart
```diff
@@ -187,14 +178,18 @@ class ConsentBackend {
...
     }
-    if (c.email != null && c.email != user.email) {
-      return null;
+    if (c.email != null) {
+      InvalidInputException.check(
+          c.email == user.email, 'Consent.email does not match current User.');
```
sigurdm: What about:
`'This invitation is not for the user account currently logged in'`

pkg/pub_integration/test/publisher_test.dart
isoos: The scripts are designed for dual nature: to run as test (against the fake server), and to run as a program (against the staging or production deployment). Although only a fraction is run against the real deployment, the pattern was to use throw-like checks instead of using `package:test`. We could probably add a helper method to make that simpler.

sigurdm: Ok - I suspected there would be a reason.
But you still import package test?

isoos: In this case it is part of a (test-specific) callback that gets evaluated as part of the script, and should throw an exception if the consent was not valid or could not be accepted. I think we could use a `package:test` here, but I'd rather keep it consistent with the script itself.

Issue #2905, opened 2019-10-10
Support third-level domains in verified publishers


mit-mit:
The new verified publishers feature only supports second-level domains (e.g. foo.org). This tracks a request to support third-level domains (e.g. foo.co.uk).

mit-mit:
One concrete suggestion was to use the ICANN domains list in:
https://github.com/publicsuffix/list/blob/master/public_suffix_list.dat#L10 (from https://publicsuffix.org/)

isoos:
I think we may want to have some protective design around the wrapping and displaying the long domain names, but other than that, we could enable more levels.

jonasfj:
Okay, talking about this we reach consensus that we allow arbitrary sub-domains, and limit the publisherId to 64 characters.

mit-mit:
Please make sure the id length limitation and our templates is aligned. 64 chars looks very odd right now:

![Screenshot 2019-10-23 at 13 52 33](https://user-images.githubusercontent.com/13644170/67390207-7b122200-f59c-11e9-902d-7311d219bdbd.png)

jonasfj:
note. we already support 53 + extension length, so that's already a problem -- a different issue though :)

isoos:
Displaying long publisher names seems to be about the following choices:
- we either wrap the long name into multiple lines, or
- we hide the overflow part of the name and keep it in a single line

As the name is a sign of trust, I think it would be better to wrap the name into multiple lines. wdyt?

jonasfj:
Wrap would be ideal... If we can do `max-width: 32ex` so there can be at-least 32 chars on the first line.

isoos:
I think this is done.

PR #3019, opened 2019-11-11
Break long publisher names on landing page.

isoos:
#2905

### DIFF
--- a/pkg/web_css/lib/src/_list.scss
+++ b/pkg/web_css/lib/src/_list.scss
@@ -56,7 +56,8 @@
   font-size: 12px;
   float: right;
   display: block;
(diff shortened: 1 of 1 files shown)


pkg/web_css/lib/src/_list.scss
```diff
@@ -56,7 +56,8 @@
   font-size: 12px;
   float: right;
   display: block;
-  margin-top: 4px;
+  margin: 4px 0px 4px 4px;
+  max-width: 32ex;
```
jonasfj: This works out to around 32 characters right?

isoos: yes

PR #3022, opened 2019-11-11
Style consent reject button differently.

### LINKED ISSUE #3013: Consent accept and reject buttons should have different colors
* "Reject" should be red.
 * "Accept" could be blue, or maybe green? (I'm color blind maybe don't ask me, hehe).

isoos:
- This makes the reject button gray, which is I think better than the green/blue vs red distinction, as described in the original issue. I've also tried the style we are using on the dialog buttons, but that was too light for consent.
- The `-cancel` suffix is probably not the best name, any suggestions?
- Fixes #3013.


isoos:
Let's try this for now. Maybe the new design can also address it.

PR #3025, opened 2019-11-12
Reserve package names from remove fluttery packages

jonasfj:


app/lib/shared/utils.dart
```diff
@@ -119,6 +119,10 @@ final _reservedPackageNames = <String>[
   'math',
   'versions',
   'webview_flutter',
+  // removed in #2853:
```
sigurdm: Please write the full github issue link

PR #3027, opened 2019-11-13
Parse and use tags in user-provided query string.

isoos:
- non-null `tagsPredicate` in both `SearchQuery` (which we control and configure) and `ParsedQuery` (which the user sends as the query string and we parse).
- The combiner function is simple: the user-provided query overrides the search query's settings. (we may change it later if needed)
- I kept `TagsPredicate` construction with `requiredTags` and (renamed) `prohibitedTags`, but the internal data is using a map, as it is much easier to implement `append` with it.

### DIFF
--- a/app/lib/package/package_tags.dart
+++ b/app/lib/package/package_tags.dart
@@ -12,10 +12,20 @@ abstract class PackageTags {
 
   /// The first version of the package was published less than 30 days ago.
   static const String isRecent = 'is:recent';
+
+  static const $all = <String>[
+    isDiscontinued,
+    isNotAdvertized,
+    isRecent,
+  ];
 }
 
 /// Collection of version-related tags.
 abstract class PackageVersionTags {
   /// PackageVersion supports only legacy (Dart 1) SDK.
   static const String isLegacy = 'is:legacy';
+
+  static const $all = <String>[
+    isLegacy,
+  ];
 }
--- a/app/lib/package/package_tags.dart
+++ b/app/lib/package/package_tags.dart
@@ -12,20 +12,10 @@ abstract class PackageTags {
 
   /// The first version of the package was published less than 30 days ago.
   static const String isRecent = 'is:recent';
-
-  static const $all = <String>[
-    isDiscontinued,
-    isNotAdvertized,
-    isRecent,
-  ];
 }
 
 /// Collection of version-related tags.
 abstract class PackageVersionTags {
   /// PackageVersion supports only legacy (Dart 1) SDK.
   static const String isLegacy = 'is:legacy';
-
-  static const $all = <String>[
-    isLegacy,
-  ];
 }
--- a/app/lib/search/search_service.dart
+++ b/app/lib/search/search_service.dart
@@ -466,7 +466,7 @@ class TagsPredicate {
 
   /// Evaluate this predicate against the list of supplied [tags].
   /// Returns true if the predicate matches the [tags], false otherwise.
-  bool evaluate(List<String> tags) {
+  bool matches(List<String> tags) {
     tags ??= const <String>[];
     for (String tag in _values.keys) {
       final present = tags.contains(tag);
--- a/app/lib/search/search_service.dart
+++ b/app/lib/search/search_service.dart
@@ -7,6 +7,8 @@ import 'dart:math' show max;
 
 import 'package:json_annotation/json_annotation.dart';
 
+import '../package/package_tags.dart';
+
 part 'search_service.g.dart';
 
 const int _minSearchLimit = 10;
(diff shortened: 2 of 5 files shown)


app/test/search/tags_test.dart
```diff
@@ -139,5 +139,44 @@ void main() {
...
+    test('User-supplied queried tags #2', () async {
+      final index = SimplePackageIndex();
+      await index.addPackage(
+          PackageDocument(package: 'pkg1', tags: ['is:recent', 'is:legacy']));
+      await index
+          .addPackage(PackageDocument(package: 'pkg2', tags: ['is:recent']));
+      await index.merge();
+
+      final rs = await index.search(SearchQuery.parse(
+          tagsPredicate: TagsPredicate(prohibitedTags: ['is:legacy']),
```
jonasfj: Should the `tagsPredicate` argument to `SearchQuery.parse` perhaps be named `defaultTagsPredicate`?

app/lib/search/search_service.dart
```diff
@@ -20,6 +22,12 @@ const int resultsPerPage = 10;
 /// links from page 5 to page 15.
 const int maxPages = 10;
 
+/// The tags that we can detect in the user-provided search query.
+final _detectedTags = <String>{
+  ...PackageTags.$all.expand((s) => [s, '-$s', '+$s']),
+  ...PackageVersionTags.$all.expand((s) => [s, '-$s', '+$s']),
```
jonasfj: I don't know that we need to support the `+is:legacy` syntax, given that we haven't assigned it any semantics beyond `is:legacy`, maybe it's best we don't?
(not sure where we landed on this last time, either way also okay following that and keeping it).

app/lib/search/search_service.dart
```diff
@@ -444,33 +455,52 @@ class TagsPredicate {
...
       }
     }
-    return TagsPredicate(requiredTags: requiredTags, negatedTags: negatedTags);
+    return p;
+  }
+
+  /// Appends [requiredTags] and [prohibitedTags] to the current set of tags,
+  /// and returns a new [TagsPredicate] instance.
+  /// The appended flags take precedence over the current ones.
+  TagsPredicate append(
```
jonasfj: Do we need this method? it does the same as `appendPredicate`, right?

app/lib/search/search_service.dart
```diff
@@ -444,33 +455,52 @@ class TagsPredicate {
...
+    final p = TagsPredicate();
+    p._values.addAll(_values);
+    requiredTags?.forEach((tag) => _values[tag] = true);
+    prohibitedTags?.forEach((tag) => _values[tag] = false);
+    return p;
+  }
+
+  /// Appends [other] predicate to the current set of tags, and returns a new
+  /// [TagsPredicate] instance.
+  /// The appended flags take precedence over the current ones.
```
jonasfj:
```
//// If there are conflicting tag predicates, the [other] takes precedence over this [TagsPredicate].
```

We could also consider calling it `merge` or `combine`...

app/lib/search/search_service.dart
```diff
@@ -444,33 +455,52 @@ class TagsPredicate {
...
+  TagsPredicate appendPredicate(TagsPredicate other) {
+    final p = TagsPredicate();
+    p._values.addAll(_values);
+    p._values.addAll(other._values);
+    return p;
   }
 
   /// Evaluate this predicate against the list of supplied [tags].
   /// Returns true if the predicate matches the [tags], false otherwise.
   bool evaluate(List<String> tags) {
```
jonasfj: nit: (out of scope) Maybe we should rename this `match`, `isMatch`, `hasMatch` or `satisfied`..

app/lib/package/package_tags.dart
```diff
@@ -12,10 +12,20 @@ abstract class PackageTags {
...
 }
 
 /// Collection of version-related tags.
 abstract class PackageVersionTags {
   /// PackageVersion supports only legacy (Dart 1) SDK.
   static const String isLegacy = 'is:legacy';
+
+  static const $all = <String>[
+    isLegacy,
+  ];
```
jonasfj: This is for parsing tags from the search query later.. hence I think it would be simpler if we expressed tags as a list of prefixes..

I think it would even be safe for us to express everything up-to the the colon... Hence, our tags prefix list might just be `['is:']`.

As we add more tags in `pana` we just put the prefixes into `pub-dev`, or expose them as a library from `pana`.

app/lib/search/search_service.dart
isoos: `merge` implies in-place update to me, I'm undecided about `combine`.

app/lib/search/search_service.dart
isoos: IIRC we have agreed to keep it (as power users will likely expect it to work that way).

PR #3028, opened 2019-11-13
Collect coverage for puppeteer tests.

isoos:
Chrome (and with that `puppeteer`) does not provide coverage ranges for the original source maps, only for the actual content loaded into the browser. We should be able to do it for ourselves with `package:source_maps` eventually, but until that happens, this provides us a rough number of script coverage (which should increase when we update the test).

Current coverage numbers:
- `script.dart.js`: `40.8%`
- `style.css`: `31.62%`

### DIFF
--- a/pkg/pub_integration/lib/src/headless_env.dart
+++ b/pkg/pub_integration/lib/src/headless_env.dart
@@ -48,7 +48,7 @@ class HeadlessEnv {
     return null;
   }
 
-  Future<void> _startBrowser() async {
+  Future<void> startBrowser() async {
     if (_browser != null) return;
     final chromeBin = await _detectChromeBinary();
     final userDataDir = await tempDir.createTemp('user');
@@ -68,10 +68,9 @@ class HeadlessEnv {
     );
   }
 
-  /// Creates a new page and setup overrides and tracking. [Page] must be closed
-  /// with the [closePage] method.
-  Future<Page> newPage({FakeGoogleUser user}) async {
-    await _startBrowser();
+  /// Creates a new page and setup overrides and tracking.
+  Future<R> withPage<R>({FakeGoogleUser user, Future<R> fn(Page page)}) async {
+    await startBrowser();
     final page = await _browser.newPage();
     await page.setRequestInterception(true);
     if (trackCoverage) {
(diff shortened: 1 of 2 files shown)


pkg/pub_integration/lib/src/headless_env.dart
```diff
@@ -95,10 +106,81 @@ class HeadlessEnv {
...
+    for (final c in _cssCoverages.values) {
+      print('${c.url}: ${c.percent.toStringAsFixed(2)}%');
+    }
+  }
+}
+
+class _Coverage {
+  final String url;
+  int textLength;
+  List<Range> _ranges = <Range>[];
```
jonasfj: This might be worth documenting a bit..

pkg/pub_integration/lib/src/headless_env.dart
```diff
@@ -95,10 +106,81 @@ class HeadlessEnv {
       print('Error: $e');
       clientErrors.add(e);
     });
+
+    _trackedPages.add(page);
     return page;
   }
 
+  /// Gets tracking results of [page] and closes it.
+  Future<void> closePage(Page page) async {
```
jonasfj: What if we did:
```dart
class TrackedPage {
  final Page page;
  final HeadlessEnv _env;
  TrackedPage._(this.page, this._env);
  Future<void> close() {
     // Collect coverage and close the page...
  }
}
```
Then we could just return `TrackedPage` from `HeadlessEnv.newPage()`.

We could also subclass `Page` and overwrite `close()`, then we wouldn't have to expose the `TrackedPage` type, as it can just be a private class.

Or possibly, better would be to make a `withPage(FutureOr<T> Function<T>() fn)` function which creates the page, calls `fn`, collects coverage numbers and closes the page. I would suspect we end up writing `testWithPage`, or something like that anyways :)

pkg/pub_integration/lib/src/headless_env.dart
```diff
@@ -14,8 +14,12 @@ class HeadlessEnv {
   final Directory tempDir;
   Browser _browser;
   final clientErrors = <ClientError>[];
+  final bool trackCoverage;
+  final _trackedPages = <Page>[];
+  final _jsCoverages = <String, _Coverage>{};
```
jonasfj: Documentation?

Issue #2830, opened 2019-09-30
Client app should send id_token and not access_token.


isoos:
[x] Send `id_token` as bearer header
[ ] Send `access_token` as payload when needed by the service endpoint (create publisher, session login)
[ ] Verify `access_token` payload against the usual checks (e.g. trusted audience)

PR #3033, opened 2019-11-14
Use access_token during login.

isoos:
- Strengthen the sign-in process with extra verification of the `access_token`.
- Same verification is applied when creating new publisher.
- The `imageUrl` remains client-provided, as I couldn't figure out which API to use to access it. I suspect that we may extend `package:googleapis` to also include userinfo with the `picture` field, but I'm not sure if that is the best way to do it.
- #2830

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -258,16 +258,17 @@ class AccountBackend {
   Future<String> siteAuthCodeToAccessToken(String redirectUrl, String code) =>
       _authProvider.authCodeToAccessToken(redirectUrl, code);
 
-  /// Verifies that the access token is matching the current active `User`.
+  /// Verifies that the access token belongs to the [owner].
   ///
   /// Throws [AuthenticationException] if token cannot be authenticated or the
-  /// OAuth userId differs from the current active user.
-  Future<void> verifyAccessToken(String accessToken) async {
+  /// OAuth userId differs from [owner].
+  Future<void> verifyAccessTokenOwnership(
+      String accessToken, User owner) async {
     final auth = await _authProvider.tryAuthenticate(accessToken);
     if (auth == null) {
       throw AuthenticationException.accessTokenInvalid();
     }
-    if (_authenticatedUser?.oauthUserId != auth.oauthUserId) {
+    if (owner.oauthUserId != auth.oauthUserId) {
       throw AuthenticationException.accessTokenMissmatch();
     }
   }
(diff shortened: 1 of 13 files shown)


jonasfj:
Doesn't the "tokeninfo?access_token=..." end-point return the profile image? I might be wrong..

Otherwise, you're looking for userinfo end-point in googleapis somewhere..

app/lib/account/backend.dart
```diff
@@ -258,8 +258,21 @@ class AccountBackend {
   Future<String> siteAuthCodeToAccessToken(String redirectUrl, String code) =>
       _authProvider.authCodeToAccessToken(redirectUrl, code);
 
-  /// Authenticates with bearer [token] and returns an `AuthenticatedUser`
-  /// object.
+  /// Verifies that the access token is matching the current active `User`.
+  ///
+  /// Throws [AuthenticationException] if token cannot be authenticated or the
+  /// OAuth userId differs from the current active user.
+  Future<void> verifyAccessToken(String accessToken) async {
```
jonasfj: Can't we do a better name... `verifyAccessTokenOwnership(String, User owner)`?

Using `_authenticatedUser` feels a bit implicit. Or maybe it's just me, but this method doesn't verify an access token, it does something very different.

It sort of verifies ownership of the access token.

side note: The caller could just have authenticated with the access token in the first place.

app/lib/frontend/handlers/account.dart
```diff
@@ -45,16 +46,16 @@ Future<shelf.Response> updateSessionHandler(
   if (userSessionData != null &&
       userSessionData.userId == user.userId &&
       userSessionData.email == user.email &&
-      userSessionData.imageUrl == clientSessionData.imageUrl) {
+      userSessionData.imageUrl == body.imageUrl) {
```
jonasfj: For the record, if we want to get the `imageUrl` from oauth2, we can do that using:
https://pub.dev/documentation/googleapis/latest/googleapis.oauth2.v2/UserinfoResourceApi/get.html

Example (not tested):
```dart
import 'package:googleapis/oauth2/v2.dart' as googleapis;
import 'package:googleapis_auth/auth_io.dart' as auth;

final client = http.Client();
final authClient = auth.authenticatedClient(client, AccessCredentials(accessToken: '<accessToken>', '', []));

final oauth2 = googleapis.Oauth2Api(authClient);
final userinfo await = oauth2.userinfo.get();
await client.close();

print(userinfo.id); // oauth2 ID
print(userinfo.name);
print(userinfo.email);
print(userinfo.picture);
```

We don't have to do this now, feel free to file this as a follow up item.

isoos: I'll do this as a follow-up.

PR #3036, opened 2019-11-15
Fake auth JS library for puppeteer tests (ongoing effort).

isoos:
- It is safe to delete the `gcloudProject == null` check, as the non-dev run of the project checks for it separately.
- The JS library does not work yet, but it is close to it. The current problem seems to be that authorization header is not sent for some reason, I will get back to debugging it later.

### DIFF
--- a/pkg/pub_integration/lib/src/fake_platform.js
+++ b/pkg/pub_integration/lib/src/fake_platform.js
@@ -1,0 +1,120 @@
+window.gapi = {
+  load: function (name, onReady) {
+    console.log('gapi.load');
+    onReady();
+  },
+  auth2: {
+    init: function (params) {return googleAuth;},
+    getAuthInstance: function () {return googleAuth;},
+  },
+};
+
+function runPubInit() {
+  setTimeout(function () {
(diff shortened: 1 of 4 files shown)


pkg/pub_integration/lib/src/fake_platform.js
```diff
@@ -0,0 +1,120 @@
...
+    getAuthInstance: function () {return googleAuth;},
+  },
+};
+
+function runPubInit() {
+  setTimeout(function () {
+    if ('pubAuthInit' in window) {
+      pubAuthInit();
+    } else {
+      runPubInit();
```
jonasfj: Is this hack really necessary? Can't we use `DOMContentLoaded` or something?

Actually, given that it's the last script to be loaded (even with defer), it should also execute after the other scripts..

https://javascript.info/script-async-defer

----
Note: I think we also rely on this behavior in production, so it might be best to test it too :)

isoos: Added a TODO for it.

PR #3040, opened 2019-11-18
Removed support for old Consent format.

### LINKED ISSUE #2971: Consent flow without registering User entity upfront

isoos:
Fixes #2971

### DIFF
--- a/app/lib/account/models.dart
+++ b/app/lib/account/models.dart
@@ -172,7 +172,10 @@ class UserInfo extends db.ExpandoModel {
   DateTime updated;
 }
 
-/// An active consent request sent to the recipient [User] (the parent entity).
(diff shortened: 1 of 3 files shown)


app/lib/account/models.dart
```diff
@@ -178,10 +178,6 @@ class Consent extends db.Model {
   /// The consent id.
   String get consentId => id as String;
 
-  /// The user that this consent is for.
```
jonasfj: Also update the documentation comment for the `Content` class, it currently says `[User] (the parent entity).`

PR #3042, opened 2019-11-19
Arbitrary contact e-mail for publishers.

### LINKED ISSUE #2914: Cannot add custom email as a member of the a verified package publisher
**UPDATE**
Apparently, pub.dev identifies my custom email, which is added as a secondary email on my gmail account, same with email I used for the invitation. Is allowing custom emails on top of Gmail accounts, part of the roadmap for the authentication?

**Problem**
I'm trying to add an email, take note this email is not registered in a GSuite subscription. Is this some sort of limitation when we try to add members with custom emails?

**Attachment**
Error
Screen Shot 2019-10-11 at 11 01 52 AM

Publisher Page
https://pub.dev/publishers/flutter.ph/packages

isoos:
Fixes #2914.

### DIFF
--- a/app/lib/account/consent_backend.dart
+++ b/app/lib/account/consent_backend.dart
@@ -29,12 +29,25 @@ void registerConsentBackend(ConsentBackend backend) =>
 ConsentBackend get consentBackend =>
     ss.lookup(#_consentBackend) as ConsentBackend;
 
+/// List of consent kinds.
+abstract class ConsentKind {
+  /// Invite a new uploader to a package.
+  static const packageUploader = 'PackageUploader';
+
+  /// Request permission to use e-mail as publisher contact.
+  static const publisherContact = 'PublisherContact';
+
+  /// Invite a new member to a publisher.
+  static const publisherMember = 'PublisherMember';
+}
+
 /// Represents the backend for the consent handling and authentication.
 class ConsentBackend {
   final DatastoreDB _db;
   final _actions = <String, ConsentAction>{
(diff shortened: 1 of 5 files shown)


app/lib/account/consent_backend.dart
```diff
@@ -29,12 +29,25 @@ void registerConsentBackend(ConsentBackend backend) =>
 ConsentBackend get consentBackend =>
     ss.lookup(#_consentBackend) as ConsentBackend;
 
+/// List of consent kinds.
+abstract class ConsentKind {
```
jonasfj: nit: If we exposed `ConsentBackend.invitePackageUploader`, etc... instead of exposing `ConsentBackend.invite`, then we wouldn't have to export this. And we could ensure that the `args` is type-safe..

I don't think this is important, we can always cleanup this later. It's not like we have a many implicit invariants here.

app/lib/account/consent_backend.dart
```diff
@@ -185,7 +198,8 @@ class ConsentBackend {
       InvalidInputException.check(c.userIdOfConsent == user.userId,
           'This invitation is not for the user account currently logged in.');
     }
-    if (c.email != null) {
+    final permitsDifferentUser = c.kind == ConsentKind.publisherContact;
```
jonasfj: `_actions[c.kind].permitConfirmationByUserWithOtherEmail` and then add `ConsentAction.permitConfirmationByUserWithOtherEmail` (and document what this property means).

I think this is much safer than special casing on `c.kind`.

PR #3048, opened 2019-11-21
Control some search flags through tags.

isoos:
- Improves publisher / my-packages listings: includes discontinued and legacy packages in the list.
- Proper legacy handling will be updated in a subsequent PR (as it touches on the more controversial changes).

### DIFF
--- a/app/lib/search/search_service.dart
+++ b/app/lib/search/search_service.dart
@@ -707,6 +707,7 @@ SearchQuery parseFrontendSearchQuery(
   String platform,
   List<String> uploaderOrPublishers,
   String publisherId,
+  bool includeLegacy = false,
   @required TagsPredicate tagsPredicate,
 }) {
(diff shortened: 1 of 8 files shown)


app/lib/search/search_service.dart
```diff
@@ -708,6 +725,7 @@ SearchQuery parseFrontendSearchQuery(
     offset: offset,
     limit: resultsPerPage,
     apiEnabled: isApiEnabled,
-    includeLegacy: queryParameters['legacy'] == '1',
+    includeLegacy: includeLegacy || queryParameters['legacy'] == '1',
```
jonasfj: So legacy isn't fully moved into tags just yet, right?

jonasfj: > Proper legacy handling will be updated in a subsequent PR (as it touches on the more controversial changes).

Ah, good :)

PR #3056, opened 2019-11-22
Add 'my liked packages' page.

szakarias:


app/lib/frontend/templates/views/pkg/liked_package_list.mustache
```diff
@@ -0,0 +1,25 @@
...
+
+<ul class="package-list">
+  {{#packages}}
+    <li class="list-item -compact">
+      <button
+        class="mdc-button mdc-button--unelevated -pub-like-button"
+        data-mdc-auto-init="MDCRipple"
+        data-package="{{name}}"
+        data-thumb_up_outlined="{{thumb_up_outlined}}"
+        data-thumb_up_filled="{{thumb_up_filled}}">
```
jonasfj: Why not just use `{{{static_assets.img__thumb-up-filled-24px_svg}}}`

app/lib/frontend/templates/listing.dart
```diff
@@ -83,6 +87,23 @@ String renderPackageList(List<PackageView> packages) {
...
+/// Renders the `views/pkg/liked_package_list.mustache` template.
+String renderMyLikedPackagesList(List<LikeData> likes) {
+  final packagesJson = [];
+  for (final like in likes) {
+    final package = like.package;
+    packagesJson.add({
+      'url': urls.pkgPageUrl(package),
+      'name': package,
+      'liked_date': shortDateFormat.format(like.created),
+      'thumb_up_outlined': staticUrls.assets['img__thumb-up-24px_svg'],
```
jonasfj: You don't need these, we always have access to `{{{static_assets.img__thumb-up-filled-24px_svg}}}` in templates..

app/lib/frontend/templates/views/pkg/liked_package_list.mustache
szakarias: Done

app/lib/frontend/templates/views/pkg/liked_package_list.mustache
```diff
@@ -0,0 +1,25 @@
+{{! Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+    for details. All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file. }}
+
+<ul class="package-list">
+  {{#packages}}
+    <li class="list-item -compact">
+      <button
+        class="mdc-button mdc-button--unelevated -pub-like-button"
```
isoos: Is there a screenshot how this looks like?

app/lib/account/backend.dart
```diff
@@ -235,6 +241,7 @@ class AccountBackend {
       p.likes--;
       tx.queueMutations(inserts: [p], deletes: [likeKey]);
     });
+    await cache.userPackageLikes(user.userId).purge();
```
isoos: Nit: Ideally we also would like to remove the like and purge the related cache if a package gets deleted. However, if we want to retain the deleted `Package` object, that may not be necessary and we can keep the entity and also the cache. Not sure if this needs any action now.

app/lib/account/backend.dart
```diff
@@ -179,16 +179,20 @@ class AccountBackend {
...
 
-  /// Return an iterable with the names of all the packages that the given [user] likes.
-  Future<List<String>> listPackageLikes(User user) async {
-    final likes = await _db.query<Like>(ancestorKey: user.key).run().toList();
-    return likes.map((Like l) => l.package).toList();
+  /// Returns a list with [LikeData] of all the packages that the given
+  ///  [user] likes.
+  Future<List<LikeData>> listPackageLikes(User user) async {
+    return await cache.userPackageLikes(user.userId).get(() async {
+      final query = _db.query<Like>(ancestorKey: user.key)..limit(1000);
```
isoos: Please leave a TODO to introduce pagination and/or migrate this to regular search.

app/lib/frontend/handlers/routes.dart
```diff
@@ -200,6 +200,10 @@ class PubSiteService {
   Future<Response> accountPackagesPage(Request request) async =>
       accountPackagesPageHandler(request);
 
+  @Route.get('/my-liked-packages')
```
isoos: Maybe this could become a sub-page under `/my-packages`, e.g. `/my-packages/liked`?

app/lib/frontend/templates/admin.dart
```diff
@@ -128,6 +165,11 @@ String renderAccountPublishersPage({
 Tab _myPackagesLink() =>
     Tab.withLink(id: 'packages', title: 'My packages', href: '/my-packages');
 
+Tab _myLikedPackagesLink() => Tab.withLink(
+    id: 'liked-packages',
+    title: 'My liked packages',
+    href: '/my-liked-packages');
```
isoos: Please add a helper method for this in `shared/urls.dart`.

app/lib/frontend/templates/views/layout.mustache
```diff
@@ -90,6 +90,7 @@
           <div class="account-nav-info"><span id="-account-profile-email">{{user_session.email}}</span></div>
           <div class="account-nav-divider"></div>
           <div class="account-nav-link"><a href="/my-packages">My packages</a></div>
```
isoos: This reminds me that we'd need to update this template to use the urls from the `shared/urls.dart`, could you please add a `TODO` for it?

app/lib/frontend/templates/admin.dart
szakarias: Done.

app/lib/frontend/templates/views/layout.mustache
szakarias: Done

app/lib/account/backend.dart
jonasfj: A TODO about pagination would be fine.

app/lib/frontend/handlers/routes.dart
jonasfj: I think that's a mis-use... `/my-packages` is packages you own, `/liked` is not a subset of that.

app/lib/account/backend.dart
szakarias: Done

app/lib/account/backend.dart
szakarias: It's on my todo implement removal of likes when we delete a package or a user. I'll create some issues.

app/lib/frontend/handlers/routes.dart
szakarias: I agree if the menu becomes much bigger, but I think it is okay as long as we don't have more entries.

isoos: fair point!

PR #3067, opened 2019-11-25
Added assigned-tags

jonasfj:


app/bin/tools/backfill_package_fields.dart
```diff
@@ -21,6 +21,7 @@ Future main(List<String> args) async {
     print('Ensures Package.likes is set to an integer.');
     print('Ensures Package.doNotAdvertise is set to a bool.');
     print('Ensures Package.isDiscontinued is set to a bool.');
+    print('Ensures Package.isFlutterFavorite is set to a bool.');
```
isoos: update?

app/lib/admin/backend.dart
```diff
@@ -247,4 +253,58 @@ class AdminBackend {
...
+    if (package == null) {
+      throw NotFoundException.resource(packageName);
+    }
+
+    return api.AssignedTags(
+      assignedTags: package.assignedTags,
+    );
+  }
+
+  /// Handles PUT '/api/admin/packages/<package>/assigned-tags'
```
isoos: I'd find it better to have this as a `POST` API, with the request body
````
{
  "addedTags": [],
  "removedTags": [],
}
````

It would reduce the chance of accidentally removing (or adding) tags that the caller is removing (or adding) via bad pattern matching.

Comment (jonasfj) [reply app/lib/admin/backend.dart]: I was very close to doing this... I'm onboard with this change if you also think it's sane :)

File: app/lib/admin/backend.dart [hunk 3]
```diff
@@ -247,4 +253,58 @@ class AdminBackend {
...
+        Package,
+        id: packageName,
+      ));
+
+      if (package == null) {
+        throw NotFoundException.resource(packageName);
+      }
+
+      // Only update the package if it is necessary
+      if (package.assignedTags.toSet() != body.assignedTags.toSet()) {
```
sigurdm: You don't do set equality like this

app/lib/admin/backend.dart
```diff
@@ -247,4 +253,58 @@ class AdminBackend {
...
+  /// Handles PUT '/api/admin/packages/<package>/assigned-tags'
+  Future<api.AssignedTags> handlePutAssignedTags(
+    String packageName,
+    api.AssignedTags body,
+  ) async {
+    await _requireAdminPermission(AdminPermission.manageAssignedTags);
+
+    InvalidInputException.check(
+      body.assignedTags.every((tag) => allowedTagPrefixes.any(tag.startsWith)),
+      'Only following tag-prefixes is allowed "${allowedTagPrefixes.join("\", ")}"',
```
sigurdm: Only the following tag-prefixes are allowed

app/lib/package/package_tags.dart
```diff
@@ -2,6 +2,15 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+/// Tag prefixes that is allowed.
```
szakarias: is -> are

app/bin/tools/backfill_package_fields.dart
```diff
@@ -21,6 +21,7 @@ Future main(List<String> args) async {
     print('Ensures Package.likes is set to an integer.');
     print('Ensures Package.doNotAdvertise is set to a bool.');
     print('Ensures Package.isDiscontinued is set to a bool.');
+    print('Ensures Package.assignedTags is set to a bool.');
```
sigurdm: an empty list

app/lib/package/backend.dart
```diff
@@ -116,6 +116,8 @@ class PackageBackend {
   }
 
   /// Looks up a package by name.
+  ///
+  /// Returns `null` if the package doesn't exist.
```
sigurdm: Doesn't seem to be true.

.first throws if there is no element

PR #3071, opened 2019-11-26
Reduce profile image size

jonasfj:


jonasfj:
Pushed to `staging-jonasfj` will be on https://jonasfj-dot-dartlang-pub-dev.appspot.com/ soon (I hope).

PR #3075, opened 2019-11-28
Fix NPE in layout.dart

isoos:


app/lib/frontend/templates/layout.dart
```diff
@@ -57,9 +57,11 @@ String renderLayoutPage(
...
       : {
           'email': userSessionData.email,
-          // Set image size to 30x30 pixels for faster loading, see:
-          // https://developers.google.com/people/image-sizing
-          'image_url': userSessionData.imageUrl + '=s30',
+          'image_url': userSessionData.imageUrl == null
+              ? null
+              // Set image size to 30x30 pixels for faster loading, see:
+              // https://developers.google.com/people/image-sizing
+              : '${userSessionData.imageUrl}=s30',
```
sigurdm: This would render to: 'null=s30'.
I would prefer userSessionData.imageUrl == null ? null : '${userSessionData.imageUrl}=s30'

isoos: There is a null check 3 lines up

isoos: actually the code is exactly like yours :)

sigurdm: Ah sorry - i was only reading this line :)

PR #3079, opened 2019-11-29
Preparing new URLs

isoos:


app/lib/shared/tags.dart
```diff
@@ -28,3 +28,15 @@ abstract class PackageVersionTags {
...
 }
+
+/// Collection of SDK tag values.
+abstract class SdkTagValue {
+  static const String dart = 'dart';
+  static const String flutter = 'flutter';
+}
+
+/// Collection of Dart SDK runtime values.
+abstract class DartSdkRuntimeValue {
```
jonasfj: Also update `allowedTagPrefixes` to include `runtime` and `platform`.

PR #3080, opened 2019-11-29
Non-complete derived tags.

isoos:
I'd use this in staging, so that we are not blocked on `pana` and have some data to work with. Once `pana` is released, we could remove the `_deriveTags` method and use `summary.tags`.

Bumped the `runtimeVersion` in order to separate this data from the current stack.

### DIFF
--- a/app/lib/analyzer/pana_runner.dart
+++ b/app/lib/analyzer/pana_runner.dart
@@ -254,7 +254,7 @@ List<String> _deriveTags(List<String> platformTags) {
       ];
     }
   }
-  return null;
+  return <String>[];
 }
 
 Summary createPanaSummaryForLegacy(String packageName, String packageVersion) {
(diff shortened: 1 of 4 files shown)


app/lib/analyzer/pana_runner.dart
```diff
@@ -209,6 +212,51 @@ PanaReport panaReportFromSummary(Summary summary, {List<String> flags}) {
...
+      return <String>[
+        'sdk:dart',
+        'sdk:flutter',
+        'platform:android',
+        'platform:ios',
+        'runtime:native',
+      ];
+    }
+  }
+  return null;
```
jonasfj: Let's default to an empty list why not?

isoos: It'll be converted to empty list at the next step, but yeah, can do it here too.

PR #3087, opened 2019-12-02
SDK links in search tabs.

isoos:
The links are similar to the platform-based links, only it is with SDK-based links. The landing page doesn't have the non-link switch yet, will be done in a follow-up PR.


jonasfj:
side note: I don't think we have to deploy behind experimental flag, ideally we switch this on soon.. and leave overnight before we switch traffic.

isoos:
> I don't think we have to deploy behind experimental flag, ideally we switch this on soon.. and leave overnight before we switch traffic.

I'd rather keep this behind flag, until we have all the pieces together.

PR #3090, opened 2019-12-02
Upgrade dartdoc

jonasfj:


app/test/frontend/golden/pkg_show_version_page.html
```diff
@@ -66,7 +66,7 @@ <h1 class="_visuallyhidden">Dart pub</h1>
               </div>
             </div>
           </nav>
-          <div></div>
+          <div>&amp;nbsp;</div>
```
isoos: this seems weird

app/test/frontend/golden/pkg_show_version_page.html
```diff
@@ -92,7 +92,7 @@ <h2 class="title">foobar_pkg 0.2.0-dev</h2>
 Published 
             <span>Jan 1, 2014</span>
             <!-- &bull; Downloads: X -->
-  • Updated:
+  &amp;bull; Updated:
```
isoos: also this.

app/test/frontend/golden/pkg_show_version_page.html
jonasfj: Yes, probably because we've upgraded some package.. this one I can't even explain.

jonasfj: Ah, https://pub.dev/packages/xml#360

Minor change in `package:xml`, fixed it...

isoos:
Nice catch on `package:xml`!

PR #3097, opened 2019-12-03
Display SDK and secondary tags.

isoos:
- sdk:X tags are displayed with a link (to `/X/packages`)
- platform:Y and runtime:Y tags are displayed only as text (Y)
- all of these will have a title="prefix:value"

### DIFF
--- a/app/lib/frontend/templates/misc.dart
+++ b/app/lib/frontend/templates/misc.dart
@@ -5,8 +5,11 @@
 import 'package:meta/meta.dart';
 
 import '../../package/models.dart';
+import '../../search/search_service.dart' show SearchQuery;
 import '../../shared/markdown.dart';
+import '../../shared/tags.dart';
 import '../../shared/urls.dart' as urls;
+import '../request_context.dart';
 
 import '_cache.dart';
 import '_consts.dart';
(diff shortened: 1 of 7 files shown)


app/lib/frontend/templates/misc.dart
```diff
@@ -107,39 +112,77 @@ String renderMiniList(List<PackageView> packages) {
...
+    } else {
+      String prefix;
+      if (searchQuery.sdk == SdkTagValue.dart) {
+        prefix = 'runtime:';
+      } else if (searchQuery.sdk == SdkTagValue.flutter) {
+        prefix = 'platform:';
+      }
+      if (prefix != null) {
+        tabValues.addAll(
+          tags.where((s) => s.startsWith(prefix)).map(
```
jonasfj: Some helper function like:
```dart
List<String> tagSuffixes(List<String> tags, String prefix) {
  ... // also sort tags (maybe)
} 
```

might be nice...

isoos: I'll add a TODO for sorting, right now it will use the pana-provided order. I'd defer the method for a later point, I'd want to refactor this a bit more...

PR #3099, opened 2019-12-03
PackageView.tags is nullable, it is not set on everything in cache, and we create instances where it is null

jonasfj:


isoos:
Let's do #3101 instead.

PR #3109, opened 2019-12-04
Calculate and set the SearchPreference cookie

isoos:


app/lib/search/search_service.dart
```diff
@@ -550,15 +550,15 @@ class TagsPredicate {
   /// user-facing search query.
   Map<String, String> asSearchLinkParams() {
     final params = <String, String>{
-      'runtime': _tagPartsWithPrefix('runtime', value: true).join(' '),
-      'platform': _tagPartsWithPrefix('platform', value: true).join(' '),
+      'runtime': tagPartsWithPrefix('runtime', value: true).join(' '),
+      'platform': tagPartsWithPrefix('platform', value: true).join(' '),
```
jonasfj: Something missing here?

app/lib/frontend/handlers/listing.dart
```diff
@@ -58,31 +60,31 @@ Future<shelf.Response> webPackagesHandlerHtml(shelf.Request request) async {
...
-  ));
+  final result = htmlResponse(
+    renderPkgIndexPage(
+      searchResult.packages,
+      links,
+      sdk: sdk,
+      searchQuery: searchQuery,
+      totalCount: totalCount,
+    ),
+    headers: createSearchPreferenceCookie(searchPreferences),
```
jonasfj: Can this conflict with other cookies? Just thinking...

app/lib/search/search_service.dart
isoos: Removed the leading `_` in the name, as I'm using outside of the class.

app/lib/frontend/handlers/listing.dart
isoos: We only set a single cookie at any response, and it does not causes conflicts with already set cookies.

jonasfj: I see no reason to fear conflicts :)

app/lib/search/search_service.dart
jonasfj: nvm

PR #3111, opened 2019-12-04
Added small badge in search results

jonasfj:


app/lib/frontend/templates/listing.dart
```diff
@@ -61,6 +61,7 @@ String renderPackageList(
       'dev_version_url': urls.pkgPageUrl(view.name, version: view.devVersion),
       'last_uploaded': view.shortUpdated,
       'desc': view.ellipsizedDescription,
+      'is_flutter_favorite': (view.tags ?? []).contains('is:flutter-favorite'),
```
isoos: Use `view.tags.contains('...')` as it is non-null on initialization.

jonasfj:
Moved to #3111

PR #3112, opened 2019-12-04
FF search page

jonasfj:
Not sure I got this right..


app/lib/frontend/templates/listing.dart
```diff
@@ -61,6 +61,7 @@ String renderPackageList(
       'dev_version_url': urls.pkgPageUrl(view.name, version: view.devVersion),
       'last_uploaded': view.shortUpdated,
       'desc': view.ellipsizedDescription,
+      'is_flutter_favorite': (view.tags ?? []).contains('is:flutter-favorite'),
```
isoos: Same here as in the other PR: `view.tags.contains()`, as it cannot be null.

jonasfj: Are you sure?

PR #3115, opened 2019-12-05
Use SearchPreference cookie to pre-populate search fields.

isoos:


app/test/frontend/golden/my_publishers.html
```diff
@@ -74,7 +74,7 @@ <h1 class="_visuallyhidden">Dart pub</h1>
     <div class="_banner-bg">
       <div class="container">
         <div class="small-banner">
-          <form class="search-bar" action="/packages">
+          <form class="search-bar" action="/my-packages">
```
jonasfj: Hmm, is this right?

app/test/frontend/golden/my_liked_packages.html
```diff
@@ -74,7 +74,7 @@ <h1 class="_visuallyhidden">Dart pub</h1>
     <div class="_banner-bg">
       <div class="container">
         <div class="small-banner">
-          <form class="search-bar" action="/packages">
+          <form class="search-bar" action="/my-packages">
```
jonasfj: Same question here

app/lib/frontend/templates/layout.dart
```diff
@@ -103,40 +101,53 @@ String renderLayoutPage(
...
 String _renderSearchBanner({
   @required PageType type,
-  @required String platform,
-  @required String sdk,
   @required String publisherId,
   @required SearchQuery searchQuery,
 }) {
+  final sp = searchQuery != null
+      ? SearchPreference.fromSearchQuery(searchQuery)
+      : searchPreference ?? SearchPreference();
```
sigurdm: Please put parens:
```
final sp = searchQuery != null
      ? SearchPreference.fromSearchQuery(searchQuery)
      : (searchPreference ?? SearchPreference());
```
Or however it is meant to be.
Otherwise it is hard to read the meaning.

app/lib/frontend/templates/layout.dart
```diff
@@ -103,40 +101,53 @@ String renderLayoutPage(
 
 String _renderSearchBanner({
   @required PageType type,
-  @required String platform,
-  @required String sdk,
   @required String publisherId,
   @required SearchQuery searchQuery,
 }) {
+  final sp = searchQuery != null
+      ? SearchPreference.fromSearchQuery(searchQuery)
```
jonasfj: Don't this override my `runtime` preferences, once I search just once for anything flutter related?

app/lib/frontend/templates/layout.dart
```diff
@@ -103,40 +101,53 @@ String renderLayoutPage(
...
+      : searchPreference ?? SearchPreference();
   final queryText = searchQuery?.query;
   final escapedSearchQuery =
       queryText == null ? null : htmlAttrEscape.convert(queryText);
+  bool includePreferencesAsHiddenFields = true;
```
sigurdm: Could we change the logic, so this is off by default, and only included where needed. To avoid surprises of cases we did not consider.

app/lib/frontend/templates/layout.dart
sigurdm: I think that would be fine. The alternative seems confusing

app/test/frontend/golden/my_liked_packages.html
isoos: Yes, we are on the "My liked packages" page, and when I search for something, I'd expect it to be inside my liked package, or as a fallback, inside my packages.

app/test/frontend/golden/my_publishers.html
isoos: Yes, we are on the "My publisher" page, and when I search for something, it was surprising to me that it went to the /dart/packages?runtime=web page, which I had in my preferences. Instead, I think it should do a search in my packages... at least that is least surprising on that location.

app/lib/frontend/templates/layout.dart
isoos: No, this does not override it as a cookie, but the search query (== the configuration of /flutter/packages) takes precedence over the cookie value. Now, I think we agree that this needs to be structured better for longer-term maintenance.

PR #3117, opened 2019-12-05
Added titles to filter tabs

jonasfj:


app/lib/frontend/templates/layout.dart
```diff
@@ -222,56 +241,68 @@ String renderSearchTabs({
...
 
-        final url = searchQuery.change(tagsPredicate: tp).toSearchLink();
-        return {
-          'text': v,
-          'href': htmlAttrEscape.convert(url),
-          'active': isActive,
-        };
-      },
-    ).toList(),
+  return templateCache.renderTemplate('shared/search_tabs', {
```
sigurdm: Perhaps call it something different

app/lib/frontend/templates/layout.dart
```diff
@@ -150,22 +150,41 @@ String _renderSearchBanner({
   }
   String secondaryTabsHtml;
   if (searchQuery?.sdk == SdkTagValue.dart) {
-    secondaryTabsHtml = _renderSecondaryTabs(
+    secondaryTabsHtml = _renderFilterTabs(
```
isoos: Technically, SDK is a filter too, but we can get to a consistent naming after we have landed this.

PR #3122, opened 2019-12-06
Infer tags if derivedTags is not ready.

isoos:


app/lib/scorecard/models.dart
```diff
@@ -350,6 +352,53 @@ class ScoreCardData extends Object with FlagMixin {
...
+    } else if (platformTags.length == 1 &&
+        platformTags.single == KnownPlatforms.flutter) {
+      return <String>[
+        'sdk:${SdkTagValue.flutter}',
+      ];
+    } else if (platformTags.length == 1 &&
+        platformTags.single == KnownPlatforms.web) {
+      return <String>[
+        'sdk:${SdkTagValue.dart}',
+        'runtime:${DartSdkRuntimeValue.web}',
```
jonasfj: This also implies `sdk:flutter` and `platform:web`.

isoos: IIRC it is platformtags = ['web'] if there was use of `dart:html` or `dart:js`, that doesn't imply usability in flutter.

app/lib/search/updater.dart
```diff
@@ -229,6 +229,19 @@ class _PeriodicUpdateTaskSource implements TaskSource {
...
   @override
   Stream<Task> startStreaming() async* {
+    // Scan for potential derived tag updates at startup.
+    // If the package has known platform tags (previous analysis was successful),
+    // but has no sdk: tags, we'll schedule it to be updated.
+    // TODO: remove this code after the current release becomes stable
+    final missingTagTasks = _snapshot.documents.values
+        .where((pd) =>
+            (pd.platforms != null && pd.platforms.isNotEmpty) &&
+            (pd.tags == null || pd.tags.every((t) => !t.startsWith('sdk:'))))
```
jonasfj: Packages that fail analysis will get updated all the time..

isoos: Conditioning on  `(pd.platforms != null && pd.platforms.isNotEmpty)` means we only try this with packages that were not failing the analysis.

PR #3123, opened 2019-12-06
Require flutter SDK constraint when using new plugin format

jonasfj:


pkg/pub_package_reader/test/package_archive_test.dart
```diff
@@ -286,5 +286,30 @@ void main() {
...
+          sdk: flutter
+      
+      flutter:
+        plugin:
+          platforms:
+            ios:
+              classPrefix: 'FLT'
+              pluginClass: 'SamplePlugin'
+      ''');
+      expect(forbidConflictingFlutterPluginSchemes(pubspec), isNotEmpty);
```
sigurdm: Can we test for the specific error message?

PR #3157, opened 2019-12-10
Fix alignment of publisher shield and like button on small screens

szakarias:
hUVZ9y94hOZ

AkfO3H9kkwV

CJPKbBFV8sm

pMx0CwwgAGM

UXgG3P8WoVv

### DIFF
--- a/app/lib/frontend/templates/views/pkg/header.mustache
+++ b/app/lib/frontend/templates/views/pkg/header.mustache
@@ -4,7 +4,7 @@
 
 Published <span>{{short_created}}</span>
 {{#publisher_id}}
-&bull; <a href="{{& publisher_url}}"><img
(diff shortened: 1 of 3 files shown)


szakarias:
cc @mit-mit

app/lib/frontend/templates/views/pkg/header.mustache
```diff
@@ -4,7 +4,7 @@
 
 Published <span>{{short_created}}</span>
 {{#publisher_id}}
-&bull; <a href="{{& publisher_url}}"><img
+&bull; <a class="-pub-publisher" href="{{& publisher_url}}"><img
```
isoos: I think you'll need to update the `templates_test.dart`'s golden files...

PR #3166, opened 2019-12-12
Add admin api for removing packages

szakarias:


app/lib/admin/backend.dart
```diff
@@ -271,6 +277,117 @@ class AdminBackend {
     });
   }
 
+  Future<void> removePackage(String packageName) async {
```
isoos: Would it be possible to update the `remove_package.dart` script to use this method?

isoos: Please add a brief documentation, especially concerning the immutability of the call.

app/lib/package/models.dart
```diff
@@ -353,6 +353,28 @@ class PackageVersionInfo extends db.ExpandoModel {
...
+/// Entity representing a package that has been removed.
+@db.Kind(name: 'ModeratedPackage', idType: db.IdType.String)
+class ModeratedPackage extends db.ExpandoModel {
+  @db.StringProperty(required: true)
+  String name;
+
+  @db.DateTimeProperty()
+  DateTime moderated;
+
+  /// The previous publisher id (null, if the package does not have a publisher).
```
isoos: note: does not -> did not

app/lib/package/models.dart
```diff
@@ -353,6 +353,28 @@ class PackageVersionInfo extends db.ExpandoModel {
...
+  @db.StringProperty()
+  String previousPublisherId;
```
isoos: Not sure if we need to use the `previous` prefix on these names. It probably helps to distinguish it in a code search, but otherwise it could be just `publisherId` and `uploaders`, as by definition it stores the values that were set when the delete happened.

app/lib/admin/backend.dart
```diff
@@ -271,6 +277,117 @@ class AdminBackend {
...
+          .lookupValue<ModeratedPackage>(moderatedPkgKey, orElse: () => null);
+      if (moderatedPkg == null) {
+        moderatedPkg = ModeratedPackage()
+          ..parentKey = _db.emptyKey
+          ..id = packageName
+          ..name = packageName
+          ..moderated = DateTime.now().toUtc()
+          ..versions = versions.map((v) => v.version).toList();
+
+        if (package != null) {
```
isoos: why not:

```dart
  .. previousPublisherId = package?.publisherId
  .. previousUploaders = package?.uploaders
```

isoos: If the `package == null`, should we even create the `ModeratedPackage` instance? I guess it has a use for backfilling previously deleted packages, but otherwise it feels like we shouldn't create one.

app/lib/admin/backend.dart
jonasfj: Once we have this, let's remove the `remove_package.dart` script.

app/lib/package/models.dart
szakarias: I've had the same thought, so I'll remove the prefix.

app/lib/admin/backend.dart
szakarias: I'm not entirely sure. I did consider it and I was thinking, if we initiate a package removal for whatever reason, we would probably want the name to be reserved, no? And in that case we need to create a `ModeratedPackage` instance.

isoos: sgtm.

szakarias:
PTAL

PR #3167, opened 2019-12-12
Remove the use of platform/platformTags.

isoos:
- Removed `platformTags` override, as the two packages involved in it seem to be no longer maintained. 
- Removed `tags` override based on `platformTags`.
- Removed `platform` use in queries and search index.
- Replaced platformTag-based specificity score with sdk-based score (which could be extended to also include runtimes/platforms). Kept the same logic, but simplified the code and renamed it to scope-specificity.


app/lib/search/scope_specificity.dart
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import '../shared/tags.dart';
+
+double scoreScopeSpecificity(String sdk, List<String> tags) {
```
jonasfj: Document this method... what does it attempt to do?

isoos:
Good points, this better has a bump in `runtimeVersion`.

PR #3169, opened 2019-12-12
Fix HTML validator errors on the landing page.

### LINKED ISSUE #2234: Fix HTML validation
See:
 * https://validator.w3.org/nu/?doc=https%3A%2F%2Fpub.dartlang.org%2F

This is just nits...

isoos:
Fixes #2234. I've checked a few other pages, the current tab layout seems to be another issue, but I expect to rewrite that for the new design, so let's check validation errors after the rewrite.

### DIFF
--- a/app/lib/frontend/templates/views/layout.mustache
+++ b/app/lib/frontend/templates/views/layout.mustache
@@ -2,7 +2,7 @@
     for details. All rights reserved. Use of this source code is governed by a
     BSD-style license that can be found in the LICENSE file. }}
 <!DOCTYPE html>
-<html lang="en-us">
+<html lang="en-us" itemscope="itemscope" itemtype="http://schema.org/WebPage">
 <head>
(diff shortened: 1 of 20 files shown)


app/lib/frontend/templates/views/layout.mustache
```diff
@@ -2,7 +2,7 @@
     for details. All rights reserved. Use of this source code is governed by a
     BSD-style license that can be found in the LICENSE file. }}
 <!DOCTYPE html>
-<html lang="en-us">
+<html lang="en-us" itemscope="itemscope" itemtype="http://schema.org/WebPage">
```
sigurdm: Do we really need this?
I don't know much about itemscopes or itemprops, but it seems to be useful for search engine optimization. 
And the validation error indicates we describe the itemprop for the dart-logo.
I don't think we need to optimize search for the dart logo.

isoos: Yes, good point, I'll revert this part, and also remove the image's itemprop.

PR #3170, opened 2019-12-12
Upgrade package dependencies.

isoos:
Most notably: `package:appengine` and `package:googleapis` are upgraded.


sigurdm:
> @sigurdm have worked on grpc as used in the new appengine package.

I have done some work on it yes, however I'm not sure if all problems are alleviated.
But if we deploy this without switching traffic, it might give us more insight into the status.

isoos:
I'll revert the `package:appengine` part, let's not do that in the next release yet (which will need to have a pana update).

Issue #3168, opened 2019-12-12
package:browser does not have pana report


isoos:
Even if a package is old, we should have created an empty report.

sigurdm:
Do we know why? Is it pana crashing? Or are we not calling pana?

isoos:
`package:browser` is marked discontinued, that's the core reason `pana` is not called on it. I'll need to dig in a little bit why it was not updated correctly on the UI.

PR #3172, opened 2019-12-13
Updated Job scheduling and dartdoc processing.

isoos:
This should address the underlying problem of #3168.

- Updated `ScoreCardBackend.shouldUpdateReport` to be based only on the report's existence or updated time, package status flags are handled in the runners.
- `pana_runner` was already checking for the legacy and obsolete status and created an empty report on it, doing the same for `dartdoc_runner`.
- `pana_runner` also checks for discontinued status, which we don't do for `dartdoc_runner`.
- The legacy block ordering in `pana_runner` is purely aesthetics, I wanted to start the rejection logic with that.
- Lowered `successThreshold` for `dartdoc_runner`, because I think we may schedule old dartdoc results for deletion anyway.

### DIFF
--- a/app/lib/dartdoc/dartdoc_runner.dart
+++ b/app/lib/dartdoc/dartdoc_runner.dart
@@ -119,14 +119,35 @@ class DartdocJobProcessor extends JobProcessor {
       version,
       ReportType.dartdoc,
       updatedAfter: updated,
-      includeDiscontinued: true,
-      includeObsolete: true,
-      successThreshold: const Duration(days: 90),
     );
   }
 
   @override
   Future<JobStatus> process(Job job) async {
+    final packageStatus = await scoreCardBackend.getPackageStatus(
+        job.packageName, job.packageVersion);
+    if (!packageStatus.exists) {
+      _logger.info('Package does not exist: $job.');
+      return JobStatus.skipped;
+    }
+
+    // We know that dartdoc will fail on this package, no reason to run it.
+    if (packageStatus.isLegacy) {
+      _logger.info('Package is on legacy SDK: $job.');
+      await _storeScoreCard(job, _emptyReport());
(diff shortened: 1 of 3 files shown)


app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -119,14 +119,35 @@ class DartdocJobProcessor extends JobProcessor {
...
-      includeObsolete: true,
-      successThreshold: const Duration(days: 90),
     );
   }
 
   @override
   Future<JobStatus> process(Job job) async {
+    final packageStatus = await scoreCardBackend.getPackageStatus(
+        job.packageName, job.packageVersion);
+    if (!packageStatus.exists) {
```
sigurdm: Perhaps add a comment about when we expect this to happen

app/lib/scorecard/backend.dart
```diff
@@ -275,12 +275,19 @@ class ScoreCardBackend {
...
+  /// Returns whether we should update the [reportType] report for the given
+  /// [packageName] and [packageVersion].
+  ///
+  /// The method will return false, if the package or version does not exists.
+  /// The method will return true, if either of the following is true:
+  /// - it does not have a report yet,
+  /// - the report was updated before [updatedAfter],
+  /// - the report is older than [successThreshold] if it was a success,
+  /// - the report is older than [failureThreshold] if it was a failure.
   Future<bool> shouldUpdateReport(
```
sigurdm: Would it be worth it adding tests for some of this logic?

isoos: I'd rather have a larger integration test on the scheduling and runners with the 4-5 different cases we have. My next planned testing improvement is parsing the source map for browser coverage data, and then I'll take a look on testing these utils too.

Issue #3173, opened 2019-12-13
Use flutter favorite svg instead of pngs


jonasfj:

jonasfj:
There was some suggestions that the png wasn't ideal on high resolution displays, phone, retina, etc.

PR #3174, opened 2019-12-13
Add ff svg resource

jonasfj:
Filed issue #3173  for us to use this image...


sigurdm:
Can we serve this as svgz?

jonasfj:
> Can we serve this as svgz?

@sigurdm, how about we file an issue to optimize all image used on pub.dev, but maybe we should hold off till after the new design?

PR #3175, opened 2019-12-13
Upgrade Flutter and pana

isoos:


app/lib/shared/versions.dart
```diff
@@ -39,10 +39,10 @@ final String runtimeSdkVersion = '2.7.0';
 final String toolEnvSdkVersion = '2.7.0';
 
 // keep in-sync with app/pubspec.yaml
-final String panaVersion = '0.13.1+4';
+final String panaVersion = '0.13.2';
```
sigurdm: I think pana knows its own version. Could we import this from pana?

isoos: Hm, sure, we could do that, I'll do a follow-up.

PR #3178, opened 2019-12-16
Add script for backfilling Like.packageName

szakarias:


app/lib/shared/integrity.dart
```diff
@@ -352,8 +352,18 @@ class IntegrityChecker {
     _logger.info('Scanning Likes...');
 
     await for (Like like in _db.query<Like>().run()) {
-      // TODO(zarah): Add integrity check for like.packageName once backfilling
-      // of this property has been done.
+      // TODO(zarah): Comment in the following integrity check for
+      // like.packageName once backfilling of this property has been done.
```
isoos: I think we can have this code here, without the comment blocks. It will be nice to see the checks working before and after...

szakarias: Okay :-)

PR #3181, opened 2019-12-17
Use search service in integration tests.

isoos:
To make the test reasonable (*), I've added a callback mechanism that triggers a search index update after a non-read HTTP request to the frontend server.

(*) Otherwise we'll need to trigger it in some other way, e.g. HTTP request, which would then be filtered out from tests against staging or prod.

### DIFF
--- a/pkg/fake_pub_server/bin/fake_pub_server.dart
+++ b/pkg/fake_pub_server/bin/fake_pub_server.dart
@@ -5,6 +5,7 @@
 import 'package:args/args.dart';
 import 'package:http/http.dart';
 import 'package:logging/logging.dart';
+import 'package:shelf/shelf.dart' as shelf;
 
 import 'package:fake_gcloud/mem_datastore.dart';
 import 'package:fake_gcloud/mem_storage.dart';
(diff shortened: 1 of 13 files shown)


pkg/fake_pub_server/bin/fake_pub_server.dart
```diff
@@ -25,16 +36,38 @@ Future main(List<String> args) async {
...
+  final searchService = FakeSearchService(datastore, storage);
+
+  final configuration = Configuration.fakePubServer(
+    frontendPort: port,
+    storageBaseUrl: 'http://localhost:$storagePort',
+    searchPort: searchPort,
+  );
+
+  // On each non-read request, we trigger an update of the search index in
+  // `fake_search_service`.
```
jonasfj: Why? Presumably, because you don't want the tests to sleep 30s or so... Then maybe the comment should say so.

But instead of making this so implicit, can't we make an explicit function `await waitForSearchUpdate()` which either:
 * Sleeps 45s when testing against production, or,
 * Calls the `fake-update-all` end-point if testing against the fake setup.

I assume 45s is enough for search to be updated... otherwise make it sleeps longer or shorter time when testing against production deployment.

isoos: In production the search index update may happen in 10-30 minutes, and we'd need a lot more active index updater process to make it down to 60 seconds or less. I'm not sure that we want to have that before re-thinking how search should work, and I wouldn't aim for such tests to be run against prod.

jonasfj: Then maybe the function to update search throws a `UnsupportedError` when running against production..

That way it's more explicit that a test doesn't run against production if it uses this function...

isoos: Updated. PTAL.

pkg/pub_integration/lib/src/pub_http_client.dart
```diff
@@ -19,6 +19,17 @@ class PubHttpClient {
     }
   }
 
+  /// Forces the search index to update.
```
jonasfj: Maybe leave in comment that this intentionally only works locally..

PR #3184, opened 2019-12-18
Track names of moderated packages to disallow as new package names

szakarias:


app/lib/package/backend.dart
```diff
@@ -938,7 +938,7 @@ Future<_ValidatedUpload> _parseAndValidateUpload(
   final pubspec = Pubspec.fromYaml(archive.pubspecContent);
   if (!await nameTracker.accept(pubspec.name)) {
     throw GenericProcessingException(
-        'Package name is too similar to another package.');
+        'Package name is too similar to another active or moderated package.');
```
isoos: not sure: moderated -> reserved ?

app/lib/package/name_tracker.dart
```diff
@@ -43,18 +44,26 @@ class NameTracker {
...
+      _reservedNames.contains(name);
 
   /// Whether to accept the upload attempt of a given package [name].
   ///
   /// Either the package [name] should exists, or it should be different enough
-  /// from already existing package names. An example for the rejection:
-  /// `long_name` will be rejected, if package `longname` or `lon_gname` exists.
+  /// from already existing active or moderated package names. An example for
+  /// the rejection: `long_name` will be rejected, if package `longname` or
+  /// `lon_gname` exists.
```
isoos: nit: `lon_gname` -> `long_name`

szakarias: Isn't it supposed to be `lon_gname`, or am I misunderstanding the comment?

isoos: My bad, automatic spell check error on my part...

app/lib/package/backend.dart
szakarias: I want to keep `moderated` as this is the term we use when we remove the package. But maybe the internal use of `reserved` is a bit confusing. I'll add a bit of documentation.

sigurdm: This is an error that will show up on `pub publish` right?

As an end-user I might get confused by the term 'moderated' without further context.

PR #3187, opened 2019-12-18
Parse source maps + JS/CSS code coverage.

isoos:
- It seems that the JS code coverage is does not have all the original source lines, probably because some inlining happens.
- The mapping and coverage calculation is inefficient, but it is good enough that I don't think it would be worth to optimize it.
- `web_app` - 34.4%
-  `web_css` - 28.5%


pkg/code_coverage/lib/source_map_coverage.dart
```diff
@@ -0,0 +1,127 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:meta/meta.dart';
+import 'package:path/path.dart';
+import 'package:source_maps/parser.dart';
+
+Future<void> main() async {
```
jonasfj: A two line comment about what this script does wouldn't hurt :)

PR #3198, opened 2019-12-20
Delayed GC of dartdoc files.

isoos:
- The delayed start of the GC helps new deployments to produce the required files earlier.
- The lower concurrency reduces the rate limit pressure on the storage bucket.
- We'll have an increased risk that the cleanup takes much longer, and may never complete for some unlucky package versions, but as we are planning to restructure the dartdoc buckets and their storage anyway, this may be acceptable now. (We can also increase the concurrency if we wanted to.)

### DIFF
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -47,6 +47,11 @@ class DartdocBackend {
   final DatastoreDB _db;
   final Bucket _storage;
   final VersionedJsonStorage _sdkStorage;
+
+  /// If the server crashed, the pending GC tasks will disappear. This is
+  /// acceptable, as - eventually - new dartdoc will be generated in the future,
+  /// which will trigger another GC for the package/version, and hopefully
+  /// we'll catch up on these files.
   final _gcTasks = <_GCTask>{};
 
   DartdocBackend(this._db, this._storage)
@@ -267,7 +272,10 @@ class DartdocBackend {
   }
 
   /// Schedules the garbage collection of the [package] and [version].
-  /// The wait queue is in-memory, it is not persisted.
+  ///
+  /// The wait queue is in-memory, it is not persisted, and only only works as a
+  /// best-effort method to clean up obsolete files.
+  /// TODO: implement weekly cleanup process outside of the job processing
   void scheduleGC(String package, String version) {
     _gcTasks.add(_GCTask(package, version));
   }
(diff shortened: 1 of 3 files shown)


app/lib/dartdoc/backend.dart
```diff
@@ -265,8 +266,35 @@ class DartdocBackend {
...
+  ///
+  /// The function never returns.
+  Future<void> processScheduledGCTasks() async {
+    for (;;) {
+      if (_gcTasks.isEmpty) {
+        await Future.delayed(Duration(seconds: 30));
+        continue;
+      }
+      final task = _gcTasks.first;
+      _gcTasks.remove(task);
```
jonasfj: Nit, maybe `final task _gcTasks.removeAt(0);`

app/lib/dartdoc/backend.dart
```diff
@@ -265,8 +266,35 @@ class DartdocBackend {
     await _deleteAllWithPrefix(prefix, concurrency: concurrency);
   }
 
+  /// Schedules the garbage collection of the [package] and [version].
+  /// The wait queue is in-memory, it is not persisted.
+  void scheduleGC(String package, String version) {
```
jonasfj: Let's clearly document that this only does a `best-effort` attempt at cleanup. Since `_gcTasks` is not persistent.

app/lib/dartdoc/backend.dart
```diff
@@ -47,6 +47,7 @@ class DartdocBackend {
   final DatastoreDB _db;
   final Bucket _storage;
   final VersionedJsonStorage _sdkStorage;
+  final _gcTasks = <_GCTask>{};
```
jonasfj: Document that we are aware that if the server crashed pending GC tasks will disappear, and argue why this acceptable.

Presumably, new dartdoc will be generated in the future, which will trigger another GC for the package/version, then hopefully we'll catch up on the files.

isoos:
> My point being, there is never any reason to do this when new results arrive.. And maybe by trying to do so we're being too smart. In this case we're clearly building a queue that is not backed by persistent storage.

+1. We could also gain a lot by restructuring the paths and prefix them with the `runtimeVersion`, so cleanup is only deleting a single directory. If we had a catalog of valid `runtimeVersion` values, the scan/lookup times could also be much better.

app/lib/dartdoc/backend.dart
isoos: `_gcTasks` is a set, it has no `removeAt(int)`.

PR #3229, opened 2020-01-09
Move top-level mustache templates to subdirs.

isoos:
Also deleting `uploader_approval.mustache`, as it is no longer used.


app/lib/frontend/templates/views/uploader_approval.mustache
```diff
@@ -1,23 +0,0 @@
-{{! Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
-    for details. All rights reserved. Use of this source code is governed by a
-    BSD-style license that can be found in the LICENSE file. }}
-
-<h1>Uploader invitation</h1>
```
jonasfj: Did you intend to delete this file?

isoos: Yes, it is no longer used. We have migrated upload approvals to use the consent page, removed the related handler and rendering, but the mustache file was not deleted at that time.

PR #3233, opened 2020-01-09
Optimize+limit text search.

### LINKED ISSUE #3231: Limit long-running search queries.
The common root cause of the recent long-running search queries is the absurdly long query text, most likely via accidental copy-pasting. The record holder both in terms of query time and query text has over 8k characters and 1000 words.

We need to:
- limit the length of the query that we will process
- better filter the results, and stop processing if there are no packages left in the result set
- (maybe) set a timer and return the results after a fixed time

isoos:
Fixes #3231 

Optimizations:
- start word matching from the longest words
- rolling restriction of the result packages, early cut-off when there is no result so far

Limits:
- only the longest 20 words are checked
- after 500 ms the rest of the words (and also the exact phrase matching) is not checked

I've benchmarked long queries, and I've measured 30% drop of processing time from the first two optimization, and further 40% from the 20-words limit.

### DIFF
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -400,7 +400,7 @@ class SimplePackageIndex implements PackageIndex {
 
       for (String word in words) {
         if (packages.isEmpty) break;
-        if (pkgScores.length > 5 && sw.elapsedMilliseconds > 500) {
+        if (sw.elapsedMilliseconds > 500) {
           aborted = true;
           _logger.info(
               'Aborted word lookup after ${pkgScores.length} words and ${sw.elapsedMilliseconds} ms.');
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -403,7 +403,7 @@ class SimplePackageIndex implements PackageIndex {
         if (sw.elapsedMilliseconds > 500) {
           aborted = true;
           _logger.info(
-              'Aborted word lookup after ${pkgScores.length} words and ${sw.elapsedMilliseconds} ms.');
+              '[pub-aborted-search-query] Aborted word lookup after ${pkgScores.length} words and ${sw.elapsedMilliseconds} ms.');
           break;
         }
 
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -7,6 +7,7 @@ import 'dart:math' as math;
 
 import 'package:meta/meta.dart';
 import 'package:gcloud/service_scope.dart' as ss;
+import 'package:logging/logging.dart';
 import 'package:pana/pana.dart' show DependencyTypes;
 import 'package:path/path.dart' as p;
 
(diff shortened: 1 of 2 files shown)


app/lib/search/index_simple.dart
```diff
@@ -378,22 +381,42 @@ class SimplePackageIndex implements PackageIndex {
...
-      final int wordCount = words.length;
+      final words = splitForIndexing(text).toSet().toList();
+      final wordCount = words.length;
+
+      // lookup longer words first, as it may restrict the result set better
+      words.sort((a, b) => -a.length.compareTo(b.length));
+
+      // limit the number of words to lookup
+      if (words.length > 20) {
+        words.removeRange(20, words.length);
```
jonasfj: Should we update `wordCount` too?

app/lib/search/index_simple.dart
```diff
@@ -378,22 +381,42 @@ class SimplePackageIndex implements PackageIndex {
...
+      }
+
       final pkgScores = <Score>[];
       final apiPagesScores = <Score>[];
+      bool aborted = false;
+
       for (String word in words) {
+        if (packages.isEmpty) break;
+        if (pkgScores.length > 5 && sw.elapsedMilliseconds > 500) {
```
jonasfj: Why not always abort after `500ms` ?

app/lib/search/index_simple.dart
```diff
@@ -378,22 +381,42 @@ class SimplePackageIndex implements PackageIndex {
...
+          aborted = true;
+          _logger.info(
+              'Aborted word lookup after ${pkgScores.length} words and ${sw.elapsedMilliseconds} ms.');
```
jonasfj: Can we include some string we can use the make custom metrics for counting number of slow queryies, like `[pub-slow-search-query]` if we include we can easily make a custom metric...

app/lib/search/index_simple.dart
```diff
@@ -770,14 +796,16 @@ class TokenIndex {
 
   /// Returns an {id: score} map of the documents stored in the [TokenIndex].
   /// The tokens in [tokenMatch] will be used to calculate a weighted sum of scores.
+  /// When [ids] is specified, the result will contain only the set of ids in it.
```
jonasfj: Wouldn't be better to name this `packages` or `packageIds` or `onlyPackageIds` or `limitToPackageId`? Since `ids` is really a list of package names, right? or is it something else?

app/lib/search/index_simple.dart
```diff
@@ -378,22 +381,42 @@ class SimplePackageIndex implements PackageIndex {
   }
 
   _TextResults _searchText(Set<String> packages, String text) {
+    final sw = Stopwatch()..start();
```
jonasfj: Can't we print something like `_logger.info('[pub-search-query-duration]: Search query took ${sw.elapsedMilliseconds} ms');`

That would be pretty cool to use for a custom metrics :)

isoos: We could have that in `search/handlers.dart`, however, we already log that as part of the standard URL logging, e.g. every `/search?q=` request on the `search` service is just that.

isoos: (However, if it is easier for the metrics processing, I'll add it in `search/handlers.dart`...)

PR #3234, opened 2020-01-10
Upgrade dartdoc: 0.30.0+1

isoos:
- #3232
- updated pkg/pub_dartdoc (small API and return value change)
- fixing canonical URL calculation (bug in dartdoc)
- updating reference content for customization golden test


isoos:
I've started a staging deploy to test the customization there too.

isoos:
Confirmed on staging, customization and links are good.

Issue #3205, opened 2019-12-27
backfill_likes.dart failing


jonasfj:
When backfilling likes we got a lot of:
```
Failed to update packageName on ike entity for user XXXX and package YYY, error The transaction was aborted.
```
I suspect it's because we are not calling `tx.rollback()` before returning.. Maybe we should try to use `datastore_helper`.

PR #3239, opened 2020-01-13
Fixing #3205

jonasfj:
Not entirely sure if this will fix: https://github.com/dart-lang/pub-dev/issues/3205

But it does some cleanup, it doesn't abandon transactions without removing them... It was never really clear how this failed.

### DIFF
--- a/app/bin/tools/backfill_likes.dart
+++ b/app/bin/tools/backfill_likes.dart
@@ -2,12 +2,15 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
+
 import 'package:appengine/appengine.dart';
 import 'package:args/args.dart';
 import 'package:gcloud/db.dart';
 import 'package:pool/pool.dart';
 import 'package:pub_dev/account/models.dart';
 import 'package:pub_dev/service/entrypoint/tools.dart';
+import 'package:pub_dev/shared/datastore_helper.dart';
 
 final _argParser = ArgParser()
   ..addOption('concurrency',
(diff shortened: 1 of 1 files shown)


app/bin/tools/backfill_likes.dart
```diff
@@ -31,8 +34,14 @@ Future main(List<String> args) async {
 
     useLoggingPackageAdaptor();
     await for (Like l in dbService.query<Like>().run()) {
-      final f = pool.withResource(() => _backfillLikes(l));
-      futures.add(f);
+      final r = await pool.request();
+      futures.add(() async {
```
isoos: I find this really confusing, and it took my a while to interpret what this does and to confirm that it does the same (with more lines). What was your issue with the previous approach?

jonasfj: Well, I still have an unbounded list of futures... but I don't schedule an unbounded list of micro-tasks..

In general, we have a problem in these backfill utilities.. we should always do operations streaming. And block reading the query over all `Like` entities.

isoos: Please add a comment that describes this. `package:executor` has a nice construct to check on the scheduled but not yet running items, and we can selectively wait on the currently running and all of the scheduled ones. In the future we may use that instead of `package:pool`.

jonasfj: The `Executor.onChange` could probably be used to do this.. but it also seems a bit complex...

Ideally, we need a function to concurrently map a stream with a concurrency limit.

isoos: Yeah, a stream transform/processing utility with a concurrency would be nice...

Issue #3246, opened 2020-01-15
Tracking the status of the new UI


isoos:
- [x] Header
  - [ ] mobile view's profile needs different format
  - [ ] mobile view's logout needs to be at the end of the "my ..." section
  - [ ] mobile view has no folding for "Flutter" and "Web & Server"
- [x] Search banner
  - [x] listing page on mobile requires a drawer icon
- [x] Landing page
  - [ ] square background does not have "semi-random" empty spaces like 
  - [ ] images do not go behind package cards
  - [x] weird page layout at some resolutions (some elements may be too large, some wrap weirdly)
  - [ ] featured packages not done yet (but should be easy to do if needed)
- [x] Listing page
  - [x] missing: mobile view hides platforms/runtimes by default
  - [x] missing: mobile view trigger
  - [x] missing: advanced controls (e.g. legacy)
  - [ ] mobile view trigger's inactive state is not from the design (needs white internal bars)
  - [ ] sort control has a simple down arrow instead of the styled one
  - [ ] publisher icon is shield instead of the geared checkmark
  - [x] tags are not styled yet
  - [ ] there is no badge for 'new' (< 30 days old) packages
  - [ ] API results are always shown, not hidden/toggled
- [x] Generic detail pages
  - [x] admin tabs are not styled with padlocks
  - [x] missing wrapper styles for mobile-view info-box
  - [x] icons inside the info-box should be lighter
  - [x] like button is not the icon from the design
  - [x] like block is misaligned
- [x] Package page
  - [ ] icon for experimental platforms #3384
  - [x] Versions tab
    - [ ] documentation icon is still the old one
    - [ ] download icon is still the old one
    - [ ] need to update JS code to switch to a different icon when documentation has failed
    - [x] tab content header style was not updated yet
- [x] Publisher page
  - [x] Admin tab
    - [ ] Input fields have their label inside / on the border, and not outside. Should we use that? (should be consistent with other places, e.g. create publisher)
    - [ ] "Invite member" flow has many extra text currently, new design does not have it. Should we display that as part of the confirm?
    - [x] "Remove member" button is unstyled and non-material (no ripple).
- [x] Help page
- [x] Footer

isoos:
Pending issues were discussed in the last design review meeting, closing this.

PR #3254, opened 2020-01-21
New landing page.

isoos:
Missing pieces tracked in #3246.

Screen Shot 2020-01-21 at 11 48 50


app/lib/frontend/handlers/landing.dart
```diff
@@ -46,13 +47,34 @@ Future<shelf.Response> indexLandingHandler(shelf.Request request) async {
...
-    final topPackages = await topFeaturedPackages();
+    final topPackages =
+        requestContext.isExperimental ? null : await topFeaturedPackages();
+
+    // TODO: implement top rated
+    final mostPopularPackages = requestContext.isExperimental
+        ? await topFeaturedPackages(order: SearchOrder.popularity)
+        : null;
+
+    // TODO: implement top trending
```
jonasfj: No need to track this is TODOs, file an issue... if I recall correctly we've agreed to not implement that as part of the design.. maybe not even this Q.

app/lib/frontend/templates/landing.dart
```diff
@@ -3,26 +3,60 @@
...
+  List<PackageView> mostPopularPackages, // new only
+  List<PackageView> topFlutterPackages, // new only
+  List<PackageView> topDartPackages, // new only
 }) {
-  final hasTagged = taggedPackages != null && taggedPackages.isNotEmpty;
-  final hasTop = topPackages != null && topPackages.isNotEmpty;
+  final isExperimental = requestContext.isExperimental;
+  final hasTagged =
+      !isExperimental && ffPackages != null && ffPackages.isNotEmpty;
+  final hasTop =
```
jonasfj: Maybe we should make a local utility function:

`bool nullOrEmpty(List l) => !isExperimental && l == null && l.isNotEmpty`

app/lib/frontend/templates/views/page/landing_experimental.mustache
```diff
@@ -0,0 +1,60 @@
+{{! Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+    for details. All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE file. }}
+
+{{#has_ff}}
```
jonasfj: Can't we just make a condition on `ff_mini_list_html`? and avoid the `has_...` variables?

app/lib/frontend/templates/landing.dart
```diff
@@ -3,26 +3,60 @@
...
     'has_tagged': hasTagged,
-    'tagged_minilist_html': hasTagged ? renderMiniList(taggedPackages) : null,
+    'tagged_minilist_html': hasTagged ? renderMiniList(ffPackages) : null,
     'tagged_more_url': '/flutter/favorites',
     'has_top': hasTop,
     'top_minilist_html': hasTop ? renderMiniList(topPackages) : null,
     'top_more_url': urls.searchUrl(),
+    // new design's variables
+    'has_ff': hasFF,
+    'ff_mini_list_html': hasFF ? renderMiniList(ffPackages) : null,
```
jonasfj: Maybe make a local utility function:

 `final renderMiniListOrNull = (x) => (x == null || x.isEmpty) ? null : renderMiniList(x);`

jonasfj:
I was a little worried when I saw all the `has...` variables, that seemed a bit messy -- but the logic is simple to follow. Cleaning up with a helper function or two might be nice, but it's a nit :)

app/lib/frontend/templates/views/page/landing_experimental.mustache
isoos: We could, but in the future non-nullability may break such templates unexpectedly, I think it is better to be explicit if the overhead is small.

PR #3257, opened 2020-01-22
New sort control + result counter

isoos:
Instead of the style down-arrow, I'm using a simple one, tracking it in #3246.

Screen Shot 2020-01-22 at 16 25 16

### DIFF
--- a/app/lib/frontend/templates/listing.dart
+++ b/app/lib/frontend/templates/listing.dart
@@ -126,7 +126,7 @@ String renderPkgIndexPage(
     'has_packages': packages.isNotEmpty,
     'pagination': renderPagination(links),
     'total_count': totalCount,
-    // TODO: remove the keys below after we have migrated to the new design
+    // TODO(3246): remove the keys below after we have migrated to the new design
     'title': title ?? topPackages,
     'search_query': searchQuery?.query,
   };
@@ -171,7 +171,7 @@ String renderSortControl(SearchQuery query) {
             })
         .toList(),
     'selected_label': selectedOption.label,
-    // TODO: remove after we have migrated to the new design
+    // TODO(3246): remove after we have migrated to the new design
     'ranking_tooltip_html': sortDict.tooltip,
   });
 }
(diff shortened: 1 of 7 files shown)


app/lib/frontend/templates/listing.dart
```diff
@@ -122,12 +122,13 @@ String renderPkgIndexPage(
   final values = {
     'sort_control_html': renderSortControl(searchQuery),
     'is_search': isSearch,
-    'title': title ?? topPackages,
     'package_list_html': renderPackageList(packages, searchQuery: searchQuery),
     'has_packages': packages.isNotEmpty,
     'pagination': renderPagination(links),
-    'search_query': searchQuery?.query,
     'total_count': totalCount,
+    // TODO: remove the keys below after we have migrated to the new design
```
sigurdm: Would it make sense to mark these with the issue number to make these TODO's easier to find?
```
TODO(3246): remove the keys below after we have migrated to the new design
```

pkg/web_css/lib/src/_list_experimental.scss
```diff
@@ -0,0 +1,73 @@
+/* non-indented rule to restrict the content of this block to the experimental pages */
```
sigurdm: I think this needs a copyright header

pkg/web_css/lib/src/_list.scss
```diff
@@ -1,3 +1,26 @@
+/* non-indented rule to restrict the content of this block to the non-experimental pages */
```
sigurdm: I think this needs a copyright header

app/lib/frontend/templates/listing.dart
isoos: Done. I'll update the earlier similar TODOs if I touch the code nearby.

pkg/web_css/lib/src/_list_experimental.scss
isoos: Done. I'll update the other SCSS files as I touch the styles in them.

PR #3269, opened 2020-01-27
Don't require having a homepage or repository

sigurdm:
Also now validate the repo url when there is a homepage.
Fixes: https://github.com/dart-lang/pub/issues/2312 together with https://github.com/dart-lang/pub/pull/2313


isoos:
Needs to update tests, otherwise LGTM.

PR #3273, opened 2020-01-28
Update search index in smaller batches.

isoos:
On my local tests this increased the wall-clock latency by about ~15-20% (indexing the snapshot at startup increased from 10.4s to 12.1s), it introduces a few ms window of slightly inconsistent index results, but I think it is worth it up.

### DIFF
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -81,25 +81,37 @@ class SimplePackageIndex implements PackageIndex {
 
   @override
   Future<void> addPackage(PackageDocument document) async {
-    final PackageDocument doc = document.intern(_internPool.intern);
+    final doc = document.intern(_internPool.intern);
 
     _packages[doc.package] = doc;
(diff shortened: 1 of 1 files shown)


app/lib/search/index_simple.dart
```diff
@@ -81,25 +81,37 @@ class SimplePackageIndex implements PackageIndex {
...
   @override
   Future<void> addPackage(PackageDocument document) async {
-    final PackageDocument doc = document.intern(_internPool.intern);
+    final doc = document.intern(_internPool.intern);
 
     _packages[doc.package] = doc;
-    _nameIndex.add(doc.package, doc.package);
-    _descrIndex.add(doc.package, doc.description);
-    _readmeIndex.add(doc.package, doc.readme);
+    await Future.delayed(Duration.zero, () {
```
jonasfj: Maybe add a comment why...

jonasfj: Also you can just do:
```dart
await Future.delayed(Duration.zero);
```

PR #3290, opened 2020-01-31
Remove package name from the beginning of the readme.

### LINKED ISSUE #3238: Remove title from README.md if it matches the package name
Example:
dpeLaWLhGEk

Obviously, we would have to parse and modify the markdown.

isoos:
Fixes #3238. 

I think we shouldn't be too smart about this: earlier stagehand templates (before https://github.com/dart-lang/stagehand/pull/578) had the project name in the readme, but it has been dropped since then. I've seen some packages using `##` instead of `#`, but otherwise it is usually either this simple pattern, or a different text. Let's not try to detect more than that, e.g. when we'll have proper markdown parsing, ToC extraction etc...

### DIFF
--- a/app/lib/frontend/templates/_utils.dart
+++ b/app/lib/frontend/templates/_utils.dart
@@ -19,7 +19,7 @@ String renderFile(FileObject file, String baseUrl, {String packageName}) {
     if (_isMarkdownFile(filename)) {
       // Remove leading package name from the beginning of the markdown file.
       if (packageName != null) {
-        final regexp = RegExp('#+ $packageName\n');
+        final regexp = RegExp('#+ ${RegExp.escape(packageName)}\n');
         if (regexp.matchAsPrefix(content) != null) {
           content = content.replaceFirst(regexp, '');
         }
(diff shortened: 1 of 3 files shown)


jonasfj:
I see packages using:
```md
<package name>
=============
```

Hence, I would prefer to use markdown types to do this.. instead of a text search / replace.

app/lib/frontend/templates/_utils.dart
```diff
@@ -12,11 +12,18 @@ import '../../shared/markdown.dart';
...
-String renderFile(FileObject file, String baseUrl) {
+String renderFile(FileObject file, String baseUrl, {String packageName}) {
   final filename = file.filename;
-  final content = file.text;
+  String content = file.text;
   if (content != null) {
     if (_isMarkdownFile(filename)) {
+      // Remove leading package name from the beginning of the markdown file.
+      if (packageName != null) {
+        final regexp = RegExp('#+ $packageName\n');
```
jonasfj: It feels more right to parse the markdown... but if we really wanted to do this, then we should at-least use:
https://api.dart.dev/stable/2.7.1/dart-core/RegExp/escape.html

even if package names can only be safe strings...

isoos:
> I see packages using:

I suspect that this is a much smaller subset of packages. The `# {{name}}` pattern originates with stagehand, while the `=====` pattern is a personal choice of the package authors. I don't think we need to put in effort to solve all the edge cases, just the majority of them.

PR #3293, opened 2020-02-04
Button widget

isoos:
I'm not sure if this is something we want to do while using mustache as templates, but I've started the PR to have a discussion around it. The current template covers the basic button, and it shouldn't be too hard to also add the version with [changing icons](https://github.com/dart-lang/pub-dev/blob/master/app/lib/frontend/templates/views/pkg/liked_package_list.mustache#L14-L19). wdyt?

### DIFF
--- a/app/lib/frontend/templates/admin.dart
+++ b/app/lib/frontend/templates/admin.dart
@@ -19,6 +19,7 @@ import 'detail_page.dart';
 import 'layout.dart';
 import 'listing.dart';
 import 'publisher.dart' show renderPublisherList;
+import 'widget.dart';
 
 /// Renders the `views/account/authorized.mustache` template.
 String renderAuthorizedPage() {
@@ -36,6 +37,15 @@ String renderConsentPage({
   final content = templateCache.renderTemplate('account/consent', {
     'title': title,
     'description_html': descriptionHtml,
+    'reject_button_html': renderButton(
+      label: 'Reject',
+      id: '-admin-consent-reject-button',
+      classes: ['pub-button-cancel'],
+    ),
+    'accept_button_html': renderButton(
+      label: 'Accept',
+      id: '-admin-consent-accept-button',
+    ),
   });
   return renderLayoutPage(
     PageType.standalone,
--- a/app/lib/frontend/templates/views/account/consent.mustache
+++ b/app/lib/frontend/templates/views/account/consent.mustache
@@ -8,12 +8,6 @@
 {{& description_html}}
 </div>
 <p id="-admin-consent-buttons">
-  <button
-    id="-admin-consent-reject-button"
-    class="mdc-button mdc-button--raised pub-button-cancel"
-    data-mdc-auto-init="MDCRipple">Reject</button>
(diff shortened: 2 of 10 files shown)


app/lib/frontend/templates/admin.dart
```diff
@@ -36,6 +37,15 @@ String renderConsentPage({
   final content = templateCache.renderTemplate('account/consent', {
     'title': title,
     'description_html': descriptionHtml,
+    'reject_button_html': renderButton(
```
jonasfj: Imagine we had written:
```
'reject_button': {
  'label': 'Reject'
  'id': '-admin-consent-reject-button',
  'classes': ['pub-button-cancel'],
}
```

app/lib/frontend/templates/views/account/consent.mustache
```diff
@@ -8,12 +8,6 @@
...
 <p id="-admin-consent-buttons">
-  <button
-    id="-admin-consent-reject-button"
-    class="mdc-button mdc-button--raised pub-button-cancel"
-    data-mdc-auto-init="MDCRipple">Reject</button>
-  <button
-    id="-admin-consent-accept-button"
-    class="mdc-button mdc-button--raised"
-    data-mdc-auto-init="MDCRipple">Accept</button>
+  {{& reject_button_html}}
```
jonasfj: If in the input we had written:
```
'reject_button': {
  'label': 'Reject'
  'id': '-admin-consent-reject-button',
  'classes': ['pub-button-cancel'],
},
'accept_button': { 
 ...
```
(instead of `'reject_button_html': renderButton(...`)

Then we could do use partial templates as follows:
```
{{#reject_button}}
  {{> button-template }}
{{/reject_button}}
{{#accept_button}}
  {{> button-template }}
{{/accept_button}}
```
Now in the `button-template.mustache ` file we have access to `{{label}}`, `{{id}}`, and `{{classes}}`.
I think that's quite simple and every powerful, if we don't want the reject button rendered we can just insert `null` and it doesn't get rendered at all..

----
We could also do something like:
```
{{#lambdas.render_button}}reject_button{{/lambdas.render_button}}
```
And then have the global render function inject a set of `lambdas` into all contexts..The lambda can just be a function that reads source and looks up `reject_button` reject button in source.

-----

I think partials is the better approach, especially if we can five the partials a name that indicates that they are files.. such that they are easy to find. The only danger with partials is that we don't get good errors if a variable is missing, but this is probably something we could fix in mustache (something we want to fix long term anyways).

jonasfj: Correction, we use `lenient` mode right now, if we fixed that there would be a lot fewer concerns about missing the wrong variables..

isoos:
I think these are valid points, and I agree, partials seems to be a better approach. (I'll close the PR).

The only thing I'm a bit worried about partials is the data preparation: we may need methods that prepare the data for a complex template (e.g. the like button with icons). Maybe we could just use a Button class with different constructors and a `toJson()`?

PR #3299, opened 2020-02-04
Extend template variables for strict mustache rendering.

isoos:


isoos:
> if we're scared we could try/catch wrap the template exception and re-render in lenient mode..

I wanted to do some manual testing before enabling it, but this seems to be a better approach.

PR #3301, opened 2020-02-05
Strict-mode for mustache templates.

isoos:


app/lib/frontend/templates/_cache.dart
```diff
@@ -38,31 +44,45 @@ class TemplateCache {
...
+    final data = {
       'static_assets': staticUrls.assets,
       ...values,
-    });
+    };
+    // try strict rendering first
+    try {
+      return _getTemplate(template, true).renderString(data);
+    } on mustache.TemplateException catch (e, st) {
+      _logger.warning('Strict template rendering failed for $template', e, st);
```
jonasfj: Include a special token we can search logs for, like `[strict-template-failed]`

Issue #3253, opened 2020-01-17
Create Analytics events for activities within package pages


jayoung-lee:
![pub_package_event](https://user-images.githubusercontent.com/11873270/72643137-d5fcb080-3922-11ea-8fd4-f67a91f9c739.png)

1. Log "thumb-up" button click events
2. Log tab open events for the five tabs (i.e., readme, changelog, example, installing, versions). Also please consider the followings, if possible:
    * Log a tab open event as appropriate when the page initially is opened with a tab name fragment in the URL (e.g, `https://pub.dev/packages/http#-example-tab-`)
    * Log "readme" tab open event when the page is opened with no tab name specified in the URL

Please feel free to use any string for eventCategory, eventAction, or eventLabel. There's no other custom events set for pub.dev.

PR #3306, opened 2020-02-07
Added GA events for tracking a few clicks

jonasfj:


jonasfj:
Should fix #3253.

static/js/gtag.js
```diff
@@ -1,4 +1,39 @@
...
+        return null;
+      }
+      return 'package:' + pkg;
+    }
+    return null;
+  }
+  var label = findPackageName() || '-';
+
+  function sendEvent(e) {
+    gtag('event', e.currentTarget.dataset.gaClickEvent, {
```
isoos: Please add a comment here that currentTarget is the DOM Element where the event listener was added to, and not the Element that the click happened.

static/js/gtag.js
```diff
@@ -1,4 +1,39 @@
...
+      'event_category': 'click',
+      'event_label': label,
```
isoos: Can we have a 'package' parameter instead? Or let's add `package:` prefix to the value, so publishers can also be tracked similarly.

jonasfj: We already have a package prefix added in `findPackageName`.

As I understood events: category is for sorting... action is the thing we count, and label (optional) is a bit of extra context..

On a given session an action can ONLY be counted once.. so label shouldn't necessarily be thought of as an extra level...

Okay, maybe I only somewhat understand the docs... :)

jonasfj: Maybe I should move the prefixing, so that it's less obscure... Or what are you suggesting..

isoos: Maybe we should emit the value that we want to have here in the main HTML, so no lookup is needed?

jonasfj: My thinking was to use the page `<title>` at one point... or maybe the path from the URL, but stripping away query string and hashes...

PR #3315, opened 2020-02-11
Detail page's infobox in mobile view.

isoos:
I've tried to implement it in different ways, but looks like we need to duplicate the infobox content HTML.

Metadata block to trigger the infobox view:

Screen Shot 2020-02-11 at 17 05 48

The infobox view:

Screen Shot 2020-02-11 at 17 06 00


isoos:
Note: most of the test template changes come from the extra top-level DIV in `detail/page.mustache`. I could implement it differently, but it would complicate the UI logic of triggering the view on mobile.

Issue #3314, opened 2020-02-11
Delete v2 of package_config


kevmoo:
Since we don't have https://github.com/dart-lang/pub-dev/issues/378 yet...

![Screen Shot 2020-02-11 at 8 06 55 AM](https://user-images.githubusercontent.com/17034/74254606-87ce9900-4ca5-11ea-9019-c763b8cdd476.png)

We don't like the API shipped in v2. We'd like to make sure no one gets it.

So far, there has only been **1** download of it. I think it's safe to eliminate.

CC @jakemac53 @lrhn

jakemac53:
Lets do this quickly before it does actually pick up some usages please! It should greatly simplify things.

jonasfj:
It's been a bit long after publishing, so need sign-off from mit-mit@ :)

jonasfj:
Resolved.

PR #3316, opened 2020-02-12
Use gtag beacon with additional data parameter.

isoos:


static/js/gtag.js
```diff
@@ -7,9 +7,18 @@ gtag('config', 'UA-26406144-13');
 // with a 'data-ga-click-event' attribute is clicked.
 window.addEventListener('DOMContentLoaded', function () {
   function sendEvent(e) {
-    gtag('event', e.currentTarget.dataset.gaClickEvent, {
+    var elem = e.currentTarget;
+    var callback;
+    if (elem.dataset.gaClickHref) {
```
jonasfj: Or ask if `e.target.attributes[href] != null` or something like that... Then we don't need additional data properties...

Otherwise, I suggest we make so you only have one data property, but we have two different ones depending on how send the metric..

isoos: > Or ask if e.target.attributes[href] != null or something like that... Then we don't need additional data properties...

I've started with this, but the template logic can be complex, e.g. in case of the analysis/score tab, and neither the `.target` nor the `.currentTarget` has the `<a href=""` reference. I think there is a benefit of adding an explicit attribute for it.

jonasfj: We could also put the `data-ga-click-event` property on the `<a>` tag...

Then refactor the template to include the `<a>` tag... I know it means more code, but injecting HTML into our template as we currently do is undesirable..

isoos: We have four kinds of tabs now:
- `li` without a link, text content (on readme tab, linking to changelog)
- `li` without a link, complex content (on readme tab, linking to score)
- `li` with a link, text content (on versions tab, linking to changelog)
- `li` with a link, complex content (on versions tab, linking to score)

We could add `data-ga-click-event` to `a` links inside the tabs, but then they the `li` element should not have it, otherwise it will be counted twice. The tab template will be very similar, different attribute will be masked conditionally. I'm fine either way, just wanted to make it clear before I do the change.

isoos:
Re-implemented in #3314 3318

PR #3322, opened 2020-02-13
Base typography for new design

isoos:
In hindsight we should have started with this, but I've tried to remove the now-obsolete rules that will be inherited from the base style.

The font sizes are coming from various UIs, I think we have only one or two screens where `h1` was `48px` instead of `36px`, those will be handled separately.

### DIFF
--- a/pkg/web_css/lib/src/_base.scss
+++ b/pkg/web_css/lib/src/_base.scss
@@ -22,6 +22,23 @@ body, input, button {
   -webkit-font-smoothing: antialiased;
 }
 
+body.experimental {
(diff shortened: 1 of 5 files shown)


jonasfj:
quick note; we could consider stealing some of this from material design..

isoos:
> quick note; we could consider stealing some of this from material design..

This was the best reference to the material header types and their sizes:
https://material.io/design/typography/the-type-system.html#type-scale

There are too many differences in the side design and that default style, so I'd rather stick with this PR for a while. We can revisit that later.

PR #3325, opened 2020-02-13
Log uploaded file size to help debugging.

isoos:


app/lib/package/backend.dart
```diff
@@ -881,12 +883,13 @@ Future _saveTarballToFS(Stream<List<int>> data, String filename) async {
...
     );
-    await stream.pipe(File(filename).openWrite());
+    await stream.pipe(targetFile.openWrite());
+    fileSize = await targetFile.length();
   } catch (e, st) {
     _logger.warning('An error occured while streaming tarball to FS.', e, st);
     rethrow;
   }
-  _logger.info('Finished streaming tarball to FS.');
+  _logger.info('Finished streaming tarball to FS (fileSize=$fileSize).');
```
sigurdm: Would it make sense to also include the filename in the log?

isoos: Not much. The log search is based on the UUID of the upload, which is already part of the request log.

PR #3327, opened 2020-02-14
New changelog page.

isoos:
I've started to implement this with CSS-only, but it only worked for a limited case. Rewriting the markdown tree is a bit more work, but creates a consistent experience.

Desktop view:

Screen Shot 2020-02-14 at 14 32 41

### DIFF
--- a/app/lib/shared/markdown.dart
+++ b/app/lib/shared/markdown.dart
@@ -6,16 +6,26 @@ import 'package:logging/logging.dart';
 import 'package:markdown/markdown.dart' as m;
 import 'package:pana/pana.dart' show getRepositoryUrl;
 import 'package:path/path.dart' as p;
+import 'package:pub_semver/pub_semver.dart';
 import 'package:sanitize_html/sanitize_html.dart';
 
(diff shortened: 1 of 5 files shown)


app/lib/shared/markdown.dart
```diff
@@ -212,3 +231,66 @@ String _pruneBaseUrl(String url) {
...
+  text = text.trim();
+  if (text.startsWith('v')) {
+    text = text.substring(1).trim();
+  }
+  if (text.isEmpty) return null;
+  try {
+    final v = Version.parse(text);
+    if (v.isEmpty || v.isAny) return null;
+    return v.toString();
+  } catch (_) {
```
sigurdm: Probably should avoid catchall
```
} on FormatException {
  return null;
}
```

isoos:
I agree with all of the suggestions, but the feature space exploded a bit, and it is slightly unrelated to the UI changes. I'd rather land a PR  with the UI structure, and iterate on the other features separately. I've collected the requests that were not addressed yet in the PR here: #3331, let me know which of those do you want to land with this PR.

Issue #3331, opened 2020-02-17
Improve markdown parsing


isoos:
- [ ] Refactor markdown parsing ([discussion](https://github.com/dart-lang/pub-dev/pull/3327#discussion_r380079118))
- [ ] Remove package name from the beginning of the `README.md` (#3238)
- [ ] Remove "Changelog" text from the beginning of the `CHANGELOG.md` ([discussion](https://github.com/dart-lang/pub-dev/pull/3327#discussion_r380083672))
- [ ] Document supported changelog formats
- [ ] Improve structured changelog parser
  - [ ] Accept styled elements inside a changelog header ([discussion](https://github.com/dart-lang/pub-dev/pull/3327#discussion_r380073941))
  - [ ] Support more version formats ([discussion](https://github.com/dart-lang/pub-dev/pull/3327#discussion_r380081403))
  - [ ] Consider limiting changelog structures to h1 and h2 levels. ([discussion](https://github.com/dart-lang/pub-dev/pull/3327#discussion_r380083672))

kevmoo:
Re Remove package name from the beginning of the README.md (#3238)

Instead/also – consider it a style lint (small point drop) to include the title in the readme. It's gratuitous in all contexts (source repo, pub site, etc)

kevmoo:
Ditto for "changelog" at the top of the changelog – do a small point drop for this! Encourage standard format!

PR #3337, opened 2020-02-18
Updated markdown processing.

isoos:
- refactored methods (although not public yet): #3331
- limiting changelog structures to h1, h2, and h3 #3331 (almost, it was h1, h2 there)
- limiting extra hash link to the same h1, h2, h3 headers, but allowing complex/styled headers
- cleanup of changelog's Node: no need to share code between that and `_HashLink`

### DIFF
--- a/app/lib/shared/markdown.dart
+++ b/app/lib/shared/markdown.dart
@@ -11,6 +11,10 @@ import 'package:sanitize_html/sanitize_html.dart';
 
 final Logger _logger = Logger('pub.markdown');
 
+/// Element tags that are treated as structural headers.
+/// h4, h5 and h6 are considered formatting-only headers
+const _structuralHeaderTags = <String>{'h1', 'h2', 'h3'};
+
 const _whitelistedClassNames = <String>[
   'changelog-entry',
   'changelog-version',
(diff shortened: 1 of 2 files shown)


app/lib/shared/markdown.dart
```diff
@@ -250,18 +270,15 @@ Iterable<m.Node> _groupChangelogNodes(List<m.Node> nodes) sync* {
   m.Element lastContentDiv;
   for (final node in nodes) {
     final version = (node is m.Element &&
+            _structuralHeaderTags.contains(node.tag) &&
```
jonasfj: I think we should also require internal consistency in the changelog file.. ie. it uses only h1 or h2 or h3 for all versions... We should avoid mixing them...

isoos: I think that's a good check for scoring the package, but I wouldn't block the display of the structured changelog on it.

jonasfj: Imagine a change logs as follows:
```md
# 2.1.0
 * Zebras can now encode bar-codes
## Upgrading from
### 1.0.0
 * Elephants are now scared of mice. 
### 0.9.0
 * Take care to feed lion before releasing mice.

-------

# 2.0.0
 * Zebras have been added.
 * Elephants are now scared of mice. 

-------

# 1.0.0
 * Take care to feed lion before releasing mice.
```

In this changelog we have headings made correctly, it just so happens that I use sub-headlines within paragraphs to indicate how to upgrade from previous versions... That's very nice of me to provide such elaborate instructions, but also breaks our parser.

side note: I had lots of fun writing this changelog... I almost want to write the package for it :)

isoos: The example was way too good, I needed to put that into a test :). I've updated the code to match the first tag that matches a version pattern, subsequent matches are only used as a separate version if the tag level matches. PTAL.

Issue #3235, opened 2020-01-10
Searching for 'map' doesn't return 'maps' amongst first results


jonasfj:
Not sure if we can fix this...

isoos:
There is no `package:maps`, is there? What would you expect to be returned?

jonasfj:
`google_maps_flutter`, `maps_launcher`, `clustering_google_maps`, `google_maps_webservice` and maybe `flutter_maps`.. These feel map related..

isoos:
Yeah... this is related to #3236 in the prefix part, but a slightly better approach would be to remove the plural from the words both in the index and in the query.

PR #3341, opened 2020-02-18
Include more word in token lookup.

isoos:
- improves #3235 (see `maps_test.dart` changes)
- slightly more candidates means a bit more CPU time, but they are relevant to the query
- the ngram calculation is now done lazily
- the required memory does not change

### DIFF
--- a/app/lib/search/index_simple.dart
+++ b/app/lib/search/index_simple.dart
@@ -702,8 +702,10 @@ class TokenIndex {
       final Map<String, double> weights =
           _inverseIds.putIfAbsent(token, () => <String, double>{});
       // on the first insert of an entry, we populate the similarity candidates
+      print('$token ${weights.isEmpty}');
       if (weights.isEmpty) {
         for (String reduced in deriveLookupCandidates(token)) {
+          print('r: $reduced');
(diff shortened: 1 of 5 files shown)


app/lib/search/index_simple.dart
```diff
@@ -702,8 +702,10 @@ class TokenIndex {
       final Map<String, double> weights =
           _inverseIds.putIfAbsent(token, () => <String, double>{});
       // on the first insert of an entry, we populate the similarity candidates
+      print('$token ${weights.isEmpty}');
```
jonasfj: did you intent to leave print statements?

app/lib/search/index_simple.dart
```diff
@@ -702,8 +702,10 @@ class TokenIndex {
...
       if (weights.isEmpty) {
         for (String reduced in deriveLookupCandidates(token)) {
+          print('r: $reduced');
```
jonasfj: ALso here?

app/lib/search/index_simple.dart
isoos: Ah, sorry, leftover debug statements, removing them...

PR #3345, opened 2020-02-19
Upgrade: Flutter SDK and dartdoc.

isoos:


isoos:
Oh, apparently dartdoc had a breaking API change, we'll need to change our code...

isoos:
Deployed fix to staging, it seems to work, I'll monitor it...

Issue #3343, opened 2020-02-18
Migrate away from using package:pub_server


jonasfj:
We should stop using `package:pub_server` it's being discontinued, and we have a decent http_router in this project now.

Note, when doing this migration I suggest we test with:
 * Latest dart 1.x
 * Dart 2.0.0
 * Dart 2.7.1
 * Latest Dart dev release.

To ensure that this is compatible with existing Dart releases.

Generally, we don't do much to support Dart 1.x users, as there is few left, but this is one of those cases where we're likely to break existing users and the work required to test it is fairly limited -- at-least compared to the significant risk that would break Dart 1.x users.

isoos:
An alternative to consider: release a `pub_server` with a new API that follows our current http router model?

jonasfj:
> An alternative to consider: release a pub_server with a new API that follows our current http router model?

I think we should focus on [documenting the API](https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md) because this is useful for people using the API and people aiming to host a compatible API.

Building a reusable pub-server is a great project of the community to own. The problem with `pub_server` is that it's hard to change as fast as the community wants to move, when it's used in `pub.dev`. And it's hard to add features / tweaks that isn't relevant to `pub.dev`, because they could have side effects on pub.dev and we don't want to risk that.

I think we end up blocking the community if we try to reuse code :)

isoos:
> Latest dart 1.x
> Dart 2.0.0
> Dart 2.7.1
> Latest Dart dev release.

We could extend the pub integration tests to auto-download the referenced SDKs... However, this seems to be costly to do on every PR. Maybe a one-off local integration test script that we run before a release?

PR #3350, opened 2020-02-20
Copy of package:pub_server in app/lib

isoos:
- #3343
- as a side-benefit, we no longer depend on `package:dart2_constant`


jonasfj:
Please test this on staging :)

Anything that touches the `pub get` flow is worth testing an extra time.

isoos:
> Please test this on staging :)

I will test it. But at this point this is a 1:1 copy + lint fixing, low risk... Wait until we start to refactor :)

PR #3355, opened 2020-02-20
Hack: cache build artifacts between travis stages.

isoos:


isoos:
This doesn't seem to work :(

tool/restore-cache-artifacts.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# Restores compiled artifacts from $PUB_CACHE/artifacts
+# Should be used only in travis.
```
jonasfj: Is this wise?
https://docs.travis-ci.com/user/caching/#fetching-and-storing-caches

Surely we would use a new cache folder... see `.travis.yaml`.

I don'ẗ think this will work:
https://docs.travis-ci.com/user/build-stages/#data-persistence-between-stages-and-jobs

If we were to use the cache we should name the file by hashes... but this gets hard to know what hash to read out... also the size of the cache directory grows too large.. and we can never be sure if the cache is populated...

From what I read cache is not a transmission mechanism...

isoos: Not wise at all, but I wanted to see if it works at all.. As it is not working, I've dropped the PR.

PR #3359, opened 2020-02-21
Local integration script to test with multiple SDKs.

isoos:
The intended scope is to test old (and new) SDKs (their `pub` client tool) with the current `pkg/fake_pub_server` (and check the #3343 refactoring to not break anything). It is a local testing only script (download part is for Mac).


jonasfj:
Not saying, we should spend time porting to linux.. It's fine if you run this for pre-deployment validation :)

isoos:
It should be easy to port it :)

Issue #3363, opened 2020-02-24
Fix timeout when search is down.


sigurdm:
We should put a bound on the time we are waiting, and just return 500.

PR #3367, opened 2020-02-24
5-seconds timeout for search calls.

isoos:
#3363

### DIFF
--- a/app/lib/search/search_client.dart
+++ b/app/lib/search/search_client.dart
@@ -37,7 +37,8 @@ class SearchClient {
     final String serviceUrl = '$httpHostPort/search$serviceUrlParams';
 
     Future<PackageSearchResult> searchFn() async {
-      final response = await getUrlWithRetry(_httpClient, serviceUrl);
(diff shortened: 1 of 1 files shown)


app/lib/search/search_client.dart
```diff
@@ -37,7 +37,8 @@ class SearchClient {
     final String serviceUrl = '$httpHostPort/search$serviceUrlParams';
 
     Future<PackageSearchResult> searchFn() async {
-      final response = await getUrlWithRetry(_httpClient, serviceUrl);
+      final response = await getUrlWithRetry(_httpClient, serviceUrl,
+          timeout: Duration(seconds: 5));
```
jonasfj: `getUrlWithRetry` doesn't seem like the best logic... but probably better than nothing.. ie. missing exponential backoff..

isoos: yeah, agreed, but it doesn't immediately fit the `package:retry` model of retries... maybe a thing to refactor later...

PR #3386, opened 2020-02-28
New publisher admin styles.

isoos:
The following items were added as pending to #3246:
- Input fields have their label inside / on the border, and not outside. Should we use that? (should be consistent with other places, e.g. create publisher)
- "Invite member" flow has many extra text currently, new design does not have it. Should we display that as part of the confirm?
- "Remove member" button is unstyled and non-material (no ripple).

Screenshot:
Screen Shot 2020-02-28 at 14 01 29

`_admin.scss` was a bad name for such styles, I've started `_form.scss` instead.

### DIFF
--- a/app/lib/frontend/templates/publisher.dart
+++ b/app/lib/frontend/templates/publisher.dart
@@ -161,11 +161,13 @@ String renderPublisherAdminPage({
     'description': publisher.description,
     'website_url': publisher.websiteUrl,
     'contact_email': publisher.contactEmail,
-    'member_list': members.map((m) => {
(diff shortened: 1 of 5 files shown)


jonasfj:
> Input fields have their label inside / on the border, and not outside. Should we use that? (should be consistent with other places, e.g. create publisher)

Can we make a note in #3246 to discuss this with the designer when we do the design review. I don't think it's big change.

On the other hand, I'm tempted to avoid the fancy border label, it behaves poorly when the page is reloaded. It flashes a labels on top of text as this:

BKM9VX13XPS

Issue #2125, opened 2019-03-12
The pub installing tab doesn't show dev_dependencies


feinstein:
The installing instructions [here](https://pub.dartlang.org/packages/test#-installing-tab-) show:

```
dependencies:
  test: ^1.5.3
```

while it should be:

```
dev_dependencies:
  test: ^1.5.3
```

natebosch:
It would be handy to be able to indicate somewhere in the pubspec when a package is _usually_ a dev dependency.

isoos:
@jonasfj we could infer this from the pubspec.yaml files: if a package is 90%+ referenced only as part of the `dev_dependencies:`, we could display the different label. However, it should be probably part of the package dependency graph or search.

jonasfj:
We could also make a convention `install.md` is render there if present..

isoos:
`install.md` sounds good, and/or we could also extract it from readme (if we had the parsing and table-of-content extraction for it).

devoncarew:
Can we bump the priority of this issue? We're likely going to reference the installing page for package:test from tooling and docs, and it currently provides incorrect guidence.

https://pub.dev/packages/test#-installing-tab-

Even one-off handling pf package:test would be a useful workaround (if a general purpose solution isn't feasible in the near term).

isoos:
@devoncarew: I've submitted #3397 to simply change the `dependencies` to `dev_dependencies` in the case of `package:test`. Is there any other urgent part to be fixed?

devoncarew:
Nope, that's it; thanks!

jonbhanson:
Users of my package, [flutter_native_splash](https://pub.dev/packages/flutter_native_splash), have contacted me with confusion because my readme tells them to add a in `dev_dependencies` (with is correct), but the installing tab tells them to add in `dependencies` (which is incorrect).  Would be nice to get this fixed!

sigurdm:
Closing as duplicate of https://github.com/dart-lang/pub-dev/issues/6400

PR #3397, opened 2020-03-02
Special handling of package:test on install tab

isoos:
#2125

### DIFF
--- a/app/lib/package/overrides.dart
+++ b/app/lib/package/overrides.dart
@@ -19,6 +19,17 @@ const redirectDartdocPages = <String, String>{
   'flutter': 'https://docs.flutter.io/',
 };
 
+/// Known packages that should be put in `dev_dependencies`
+const devDependencyPackages = <String>{
+  'build_runner',
+  'build_test',
+  'build_verify',
+  'build_web_compilers',
+  'test',
+  'test_descriptor',
+  'test_process',
+};
+
 // TODO: remove this after all of the flutter plugins have a proper issue tracker entry in their pubspec.yaml
 const _issueTrackerUrlOverrides = <String, String>{
   'https://github.com/flutter/plugins/issues':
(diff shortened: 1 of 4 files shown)


kevmoo:
This seems like something that should be generalized.

pkg: build_runner, build_web_compilers, build_test, build_verify, test_descriptor, test_process

In general, all of these should be shown as dev dependencies.

We could even use our own stats to drive this, right? Either from pubspecs of published packages or telemetry on the pub site

isoos:
> pkg: build_runner, build_web_compilers, build_test, build_verify, test_descriptor, test_process

Added these. Anything else to immediately add?

> This seems like something that should be generalized. We could even use our own stats to drive this, right?

Agreed. However, we have higher priority items, so let's start with the custom whitelist.

jonasfj:
> This seems like something that should be generalized.

I agree, as I suggested in #2125 I would much prefer if we just rendered `INSTALL.md` on the install tab, if the package contains an `INSTALL.md` file (exact file name match).

I doubt this would be hard to do, we just need to patch pana, update the model, and fork `backfill_example.dart` to backfull `INSTALL.md`. This way it's also useful for packages that are globally activated, like `mono_repo` or packages with external dependencies like things using `dart:ffi`.

app/lib/package/overrides.dart
```diff
@@ -19,6 +19,17 @@ const redirectDartdocPages = <String, String>{
   'flutter': 'https://docs.flutter.io/',
 };
 
+/// Known packages that should be put in `dev_dependencies`
+const devDependencyPackages = <String>{
```
jonasfj: Can we leave a comment here stating that this is a temporary hack that will be removed when:
https://github.com/dart-lang/pub-dev/issues/3403 is probably implemented?

PR #3417, opened 2020-03-09
Migrate pub_server API documentation.

isoos:
#3343


jonasfj:
The plan with the spec in: https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md
was to make it the source of truth.

I guess we still need the docs for publishing, etc. which is not documented in the spec.

isoos:
I've updated the direct link where there was content, otherwise kept the old docs. PTAL.

PR #3425, opened 2020-03-11
Special package name index.

isoos:
Improves search relevancy scores when the search expression is a multi-part substring of the package name.


app/lib/search/index_simple.dart
```diff
@@ -664,6 +660,50 @@ class TokenMatch {
   Map<String, double> get tokenWeights => Map.unmodifiable(_tokenWeights);
 }
 
+/// A simple (non-inverted) index designed for package name lookup.
+class PackageNameIndex {
```
jonasfj, line 664: Why export this?

Issue #3427, opened 2020-03-13
Fix /api/packages/<package>/versions/<version>.tar.gz


isoos:
This seems to be an unused API call, because it is not working for a while now. The api generator parses the version pattern to include the `.tar.gz` postfix in the `version`, making version pattern check to fail.

We should:
- fix the API generator
- maybe remove this API if we are not going to use it?

jonasfj:
Ah, we never hit:
https://github.com/dart-lang/pub-dev/blob/master/app/lib/frontend/handlers/pubapi.dart#L50-L52

because we always match this first:
https://github.com/dart-lang/pub-dev/blob/master/app/lib/frontend/handlers/pubapi.dart#L40-L46

In practice it doesn't matter... but, well, maybe it matters a little :) hehe

I think we should write this as:
```dart
@EndPoint.get('/api/packages/<package>/versions/<version|[^ ]+(?:\.tar\.gz)>')
```

Checkout: https://pub.dev/documentation/shelf_router/latest/shelf_router/Router-class.html
The syntax is `<param|regex>` for giving custom regular expressions.

But this raises a different problem. `1.0.0-dev.tar.gz` is a legitimate semver version number.
You could totally publish a package that has `version: "1.0.0-dev.tar.gz"`, but we would have problems serving it -- because we can't tell if you're requesting the tarball for version `1.0.0-dev` or inspecting the pubspec for version `1.0.0-dev.tar.gz` :see_no_evil: 

Good news is that we've deprecated the last option:
https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md#deprecated-inspect-a-specific-version-of-a-package
(though we still have to support existing clients, the fix is not in Dart `2.7.1`)

Am I missing something?

PR #3428, opened 2020-03-13
Refactor pub_server handlers

isoos:
- #3343
- Tested with:
  - manual API calls
  - integration tests on fake_pub_server with `1.24.3`, `2.7.0` and `2.8.0-dev.10.0` SDKs

Notes for later follow-ups:
- `/api/packages/<package>/versions/<version>.tar.gz` is not working currently, maybe it is a pattern matching issue (#3427).
-  `/api/packages/versions/newUpload` has no known use by us (keeping it in the `shelf_pubserver.dart` for now)

### DIFF
--- a/app/lib/frontend/handlers/pubapi.dart
+++ b/app/lib/frontend/handlers/pubapi.dart
@@ -19,10 +19,6 @@ import 'package.dart';
 part 'pubapi.g.dart';
 
 class PubApi {
-  final Handler _pubServerHandler;
-
-  PubApi() : _pubServerHandler = packageBackend.pubServer.requestHandler;
-
   Router get router => _$PubApiRouter(this);
 
   // ****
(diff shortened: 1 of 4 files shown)


app/lib/frontend/handlers/pubapi.dart
```diff
@@ -33,7 +29,8 @@ class PubApi {
   /// https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md#list-all-versions-of-a-package
   @EndPoint.get('/api/packages/<package>')
   Future<Response> listPackageVersions(Request request, String package) async =>
-      _pubServerHandler(_normalizeHost(request));
+      await packageBackend.pubServer
+          .listVersions(_replaceHost(request.requestedUri), package);
```
jonasfj: I don't this will work see diagnosis in https://github.com/dart-lang/pub-dev/issues/3427

isoos: Not sure: why this won't work?

jonasfj: nvm, I'm wrong..

jonasfj: hmm, maybe it will...

jonasfj:
ship it

PR #3432, opened 2020-03-16
Do not allow filename-like version pattern.

isoos:
Not a full fix yet, but mitigates one vector of #3427.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -925,6 +925,10 @@ Future<_ValidatedUpload> _parseAndValidateUpload(
     throw GenericProcessingException(
         'Unable to canonicalize the version: ${pubspec.version}');
   }
+  if (versionString.endsWith('tar.gz')) {
+    throw GenericProcessingException(
+        'The version string must not contain archive filename pattern: ${pubspec.version}');
+  }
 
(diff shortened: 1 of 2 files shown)


app/lib/package/backend.dart
```diff
@@ -925,6 +925,10 @@ Future<_ValidatedUpload> _parseAndValidateUpload(
     throw GenericProcessingException(
         'Unable to canonicalize the version: ${pubspec.version}');
   }
+  if (versionString.endsWith('tar.gz')) {
```
jonasfj: I'm not convinced this is a good idea.. Also you can settle with `.tar.gz`

jonasfj: I guess there is little harm in this. but it just seems like an arbitrary restriction...

jonasfj: If we want to go down this route a means of hardening against various hacks, I suggest we forbid all common file extensions, but really don't we have to trust the users of versions numbers read the spec for the format that they can have.

isoos: I came up with this while trying to fix the version regexp. It turned out to be simpler than I've originally thought, so this may not be needed at all.

PR #3433, opened 2020-03-16
Reorder API endpoints.

### LINKED ISSUE #3427: Fix /api/packages/<package>/versions/<version>.tar.gz

isoos:
I think this fixes #3427.

### DIFF
--- a/app/lib/frontend/handlers/pubapi.dart
+++ b/app/lib/frontend/handlers/pubapi.dart
@@ -32,17 +32,6 @@ class PubApi {
       await packageBackend.pubServer
           .listVersions(_replaceHost(request.requestedUri), package);
 
-  /// Getting information about a specific (package, version) pair.
-  /// https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md#deprecated-inspect-a-specific-version-of-a-package
-  @EndPoint.get('/api/packages/<package>/versions/<version>')
-  Future<Response> packageVersionInfo(
-    Request request,
-    String package,
-    String version,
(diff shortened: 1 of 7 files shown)


app/lib/frontend/handlers/pubapi.dart
```diff
@@ -55,6 +44,17 @@ class PubApi {
       await packageBackend.pubServer
           .download(_replaceHost(request.requestedUri), package, version);
 
+  /// Getting information about a specific (package, version) pair.
+  /// https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md#deprecated-inspect-a-specific-version-of-a-package
+  @EndPoint.get('/api/packages/<package>/versions/<version>')
```
jonasfj: We need a comment explaining why this ordering is important..

jonasfj: IMO, this can be a `// HACK: ....` explanation, also please reference the issue.

jonasfj: IMO, this hack is acceptable because it only affects package versions that are suffixed `-XXX.tar,gz`

And the API end-point affected is deprecated.

isoos:
Discussed offline: as we haven't used this endpoint in the client yet, we'll remove it and use an unambiguous pattern instead.

PR #3434, opened 2020-03-16
Change (future) archive download endpoint.

### LINKED ISSUE #3427: Fix /api/packages/<package>/versions/<version>.tar.gz

isoos:
- Fixes #3427
- I think `[..]/archive.tar.gz` is better than `[..]/archive`, as it indicates that it will be a file in the given format.

### DIFF
--- a/app/lib/frontend/handlers/pubapi.dart
+++ b/app/lib/frontend/handlers/pubapi.dart
@@ -45,7 +45,7 @@ class PubApi {
 
   /// Downloading package.
   /// https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md#download-a-specific-version-of-a-package
-  @EndPoint.get('/api/packages/<package>/versions/<version>.tar.gz')
+  @EndPoint.get('/api/packages/<package>/versions/<version>/archive.tar.gz')
   @EndPoint.get('/packages/<package>/versions/<version>.tar.gz')
   Future<Response> fetchPackage(
     Request request,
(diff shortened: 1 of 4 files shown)


app/lib/frontend/handlers/pubapi.dart
```diff
@@ -45,7 +45,7 @@ class PubApi {
 
   /// Downloading package.
   /// https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md#download-a-specific-version-of-a-package
-  @EndPoint.get('/api/packages/<package>/versions/<version>.tar.gz')
+  @EndPoint.get('/api/packages/<package>/versions/<version>/archive.tar.gz')
   @EndPoint.get('/packages/<package>/versions/<version>.tar.gz')
```
jonasfj: Can we add a comment here documenting that `/packages/<package>/versions/<version>.tar.gz` is hardcoded into the clients, so while this is deprecated we need to support it indefinitely?

jonasfj:
Also we need to update the URLs in the JSON we return.

jonasfj:
And write unit tests that the JSON we return contains URLs from which the archive can be fetcted..

isoos:
> Also we need to update the URLs in the JSON we return.
> And write unit tests that the JSON we return contains URLs from which the archive can be fetcted..

Doing those in a subsequent PR.

Issue #3390, opened 2020-02-29
Make package tabs links cleaner


devoncarew:
Currently the package sub-tabs (readme, changelog, installing, versions) have links like:
- https://pub.dev/packages/test#-readme-tab-
- https://pub.dev/packages/test#-changelog-tab-
- https://pub.dev/packages/test#-installing-tab-
- https://pub.dev/packages/test/versions#-versions-tab-

It would be nice to make the urls for direct links to these tabs slightly cleaner (this is nice when embedding direct links in documentation for example). So, the above links might then be:

- https://pub.dev/packages/test#readme
- https://pub.dev/packages/test#changelog
- https://pub.dev/packages/test#installing
- https://pub.dev/packages/test/versions#versions

or even:

- https://pub.dev/packages/test
- https://pub.dev/packages/test#changelog
- https://pub.dev/packages/test#installing
- https://pub.dev/packages/test/versions

isoos:
@devoncarew: our long-term (low-priority) plan is to separate the package tabs into their own pages: `[...]/changelog`, `[...]/example`, `[...]/installing` should be their own page, and then referencing to a specific in-page anchor would be much better.

Using the suggested simpler names would conflict with the typical anchors in the `README.md`. E.g. if you have a section `## Installing` in the readme, now the `#installing` anchor would switch to the readme tab and scroll to that given section. Similar thing happens if you do e.g. `#100` will likely go to the changelog entry of `## 1.0.0` version. The current `-` and `-tab-` is intentional: we can make it clear what the bookmarked state of the page was.

devoncarew:
Gotcha; the overall feature request is for cleaner urls - having pages for each (`.../installing`) would do that.

If that refactor is going to be a while, you might consider handling the newer form and redirecting to the anchor form.

isoos:
This was implemented and deployed in Q1/Q2.

PR #3438, opened 2020-03-17
Shared package page data and handler logic.

isoos:
- In preparation for #3390.
- Fixes one bug: package admin page always displayed likes as not-liked.
- Otherwise no functional changes.
- `PackagePageData` is shared across many parts of the handler logic, e.g.:
  - redirecting when package (or version) is missing
  - rendering header, infobox
  - rendering page metadata


app/lib/package/models.dart
```diff
@@ -567,3 +568,37 @@ class PackageLinks {
     );
   }
 }
+
+/// Common data structure shared between package pages.
+class PackagePageData {
```
jonasfj: Nice...

PR #3442, opened 2020-03-19
New package pages

isoos:
- #3390
- new pages are behind experimental flag
- known issue: non-experimental fetch of them is a bit broken as the in-page tab links are rendered but they are not working
- reusing most of the render logic of the current package page (in a follow-up PR we'll need to decide which pages will be indexed, what the canonical page URL should be, and which pages should expose other metadata, e.g. for sharing).
- added templates tests for new pages (integration tests in a follow-up PR)


isoos:
I've also started a staging deployment...

PR #3447, opened 2020-03-20
Do not create User when accepting Consent.

isoos:
I think the `AuthenticationException.userNotFound` should never be thrown, but I left it there to make sure we have a user-facing error message in case it happens (vs. generic server error).

### DIFF
--- a/app/lib/account/consent_backend.dart
+++ b/app/lib/account/consent_backend.dart
@@ -232,10 +232,8 @@ class ConsentBackend {
       }
 
       // Checking that consent is for the current user.
-      if (c.userId != null) {
-        InvalidInputException.check(c.userId == user.userId,
-            'This invitation is not for the user account currently logged in.');
-      }
+      InvalidInputException.check(c.userId == null || c.userId == user.userId,
+          'This invitation is not for the user account currently logged in.');
       final action = _actions[c.kind];
       if (!action.permitConfirmationWithOtherEmail && c.email != null) {
         InvalidInputException.check(c.email == user.email,
--- a/app/lib/account/consent_backend.dart
+++ b/app/lib/account/consent_backend.dart
@@ -221,7 +221,7 @@ class ConsentBackend {
   /// Returns the [Consent] for [consentId] and checks if it is for [user].
   Future<Consent> _lookupAndCheck(String consentId, User user) async {
     final key = _db.emptyKey.append(Consent, id: consentId);
-    final c = await withRetryTransaction(_db, (tx) async {
+    return await withRetryTransaction(_db, (tx) async {
       final c = await tx.lookupOrNull<Consent>(key);
       if (c == null) {
         throw NotFoundException.resource('consent: $consentId');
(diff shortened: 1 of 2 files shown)


app/lib/account/consent_backend.dart
```diff
@@ -411,7 +422,10 @@ class _PublisherMemberAction extends ConsentAction {
   Future<void> onAccept(Consent consent) async {
     final member = consent.userId != null
         ? await accountBackend.lookupUserById(consent.userId)
-        : await accountBackend.lookupOrCreateUserByEmail(consent.email);
+        : await accountBackend.lookupUserByEmail(consent.email);
+    if (member == null) {
+      throw AuthenticationException.userNotFound();
```
jonasfj: Maybe add a command saying: this should never happen because userId of the consent entity will be set when it is loaded.

app/lib/account/consent_backend.dart
```diff
@@ -219,9 +220,16 @@ class ConsentBackend {
 
   /// Returns the [Consent] for [consentId] and checks if it is for [user].
   Future<Consent> _lookupAndCheck(String consentId, User user) async {
-    final c = await _db.lookupValue<Consent>(
-        _db.emptyKey.append(Consent, id: consentId),
-        orElse: () => null);
+    final key = _db.emptyKey.append(Consent, id: consentId);
+    final c = await withRetryTransaction(_db, (tx) async {
+      final c = await tx.lookupValue<Consent>(key, orElse: () => null);
```
jonasfj: Use `tx.lookupOrNull(key)` it does exactly this :D

app/lib/account/consent_backend.dart
```diff
@@ -219,9 +220,16 @@ class ConsentBackend {
...
+      if (c == null) return null;
```
jonasfj: Just throw `NotFoundException` as you would have done below... I'm pretty sure `withRetryTransaction` won't retry a `ResponseException` (but check the code).

IMO, we want `withRetryTransaction` to only retry datastore related exceptions, but we haven't exactly nailed down what these are yet.

app/lib/account/consent_backend.dart
```diff
@@ -219,9 +220,16 @@ class ConsentBackend {
 
   /// Returns the [Consent] for [consentId] and checks if it is for [user].
   Future<Consent> _lookupAndCheck(String consentId, User user) async {
-    final c = await _db.lookupValue<Consent>(
-        _db.emptyKey.append(Consent, id: consentId),
-        orElse: () => null);
+    final key = _db.emptyKey.append(Consent, id: consentId);
+    final c = await withRetryTransaction(_db, (tx) async {
```
jonasfj: Just make this a `return await withRetryTransaction`, since we can throw `ResponseException` from within it.

app/lib/account/consent_backend.dart
```diff
@@ -219,24 +220,29 @@ class ConsentBackend {
...
-    }
-    final action = _actions[c.kind];
-    if (!action.permitConfirmationWithOtherEmail && c.email != null) {
-      InvalidInputException.check(c.email == user.email,
-          'This invitation is not for the user account currently logged in.');
-    }
-    return c;
+      // Checking that consent is for the current user.
+      if (c.userId != null) {
+        InvalidInputException.check(c.userId == user.userId,
```
jonasfj: nit: `InvalidInputException.check(c.userId == null || c.userId == user.userId, ....);`

app/lib/account/consent_backend.dart
```diff
@@ -219,24 +220,29 @@ class ConsentBackend {
...
-    }
+    final key = _db.emptyKey.append(Consent, id: consentId);
+    return await withRetryTransaction(_db, (tx) async {
+      final c = await tx.lookupOrNull<Consent>(key);
+      if (c == null) {
+        throw NotFoundException.resource('consent: $consentId');
+      }
+      if (c.userId == null && c.email == user.email) {
+        c.userId = user.userId;
+        tx.queueMutations(inserts: [c]);
```
jonasfj: nit: `tx.insert(c);`.

Issue #3440, opened 2020-03-18
Migrate away from package:uuid


jonasfj:
The readme says:

> Defaults to non-crypto generator, see UuidUtil for cryptoRNG

See: https://pub.dev/packages/uuid

I suggest we don't use it, it's just too risky -- let's move to `ulid` whenever possible. And when doing so verify if we rely on randomness. I was looking at using `package:uuid` for locking, but don't trust it.

kevmoo:
https://github.com/dart-lang/pub/blob/8f21b1f56dd879c85f05df1125ec9bbe7c82cc9e/lib/src/utils.dart#L524-L545

There are even tests!

https://github.com/dart-lang/pub/blob/429a06039d185149f387a65e3503b0693ce6d24e/test/utils_test.dart#L106-L126

isoos:
Our atom feed uses a v5 (namespace + seed) UUID to preserve a consistent ID for the feed readers. I think instead of that we could use the SHA hash (of the `'$package/$version'`) as the ID going forward.

The rest of our uses are just "random strings". I don't particularly care about the version-identifier bits properly set for a random UUID, so `ulid` or any random string would be fine.

PR #3448, opened 2020-03-20
Change the atom feed ID generator after a certain date.

isoos:
- I am not sure if/when we can move off of the hostname-base seed generation, instead, a date-based change should be better. It makes sure that we have a clear cutoff for the new UUIDs.
- It also removes the `package:uuid` dependency (#3440).

### DIFF
--- a/app/lib/frontend/handlers/atom_feed.dart
+++ b/app/lib/frontend/handlers/atom_feed.dart
@@ -138,7 +138,7 @@ Feed feedFromPackageVersions(Uri requestedUri, List<PackageVersion> versions) {
     if (version.created.isAfter(DateTime(2020, 04, 04))) {
       final hash =
           sha512.convert(utf8.encode('${version.package}/${version.version}'));
-      id = formatUuid(hash.bytes.sublist(0, 16));
+      id = createUuid(hash.bytes.sublist(0, 16));
     } else {
       // TODO: remove the old ID after the above date passed in production use
       final seed = (requestedUri.host == 'pub.dartlang.org')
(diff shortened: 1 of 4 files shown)


app/lib/shared/utils.dart
```diff
@@ -390,3 +390,18 @@ Map<String, String> parseCookieHeader(String cookieHeader) {
     return <String, String>{};
   }
 }
+
+/// Returns the UUID-formatted string of [bytes].
+String formatUuid(List<int> bytes) {
```
jonasfj: This only works if `bytes` is a proper UUID.

app/lib/frontend/handlers/atom_feed.dart
```diff
@@ -132,13 +134,20 @@ Feed feedFromPackageVersions(Uri requestedUri, List<PackageVersion> versions) {
...
-        ? requestedUri
-            .resolve('/packages/${version.package}#${version.version}')
-            .toString()
-        : version.qualifiedVersionKey.toString();
-    final id = uuid.v5(Uuid.NAMESPACE_URL, seed);
+    String id;
+    if (version.created.isAfter(DateTime(2020, 04, 04))) {
+      final hash =
+          sha512.convert(utf8.encode('${version.package}/${version.version}'));
+      id = formatUuid(hash.bytes.sublist(0, 16));
```
jonasfj: `hash` is not a `UUID` and formatting as such is not correct.

jonasfj: We need to encode UUID version in there, right? It's a hack... IMO we should move `formatUuid` into this file as a private method.. write a comment about how this is not real UUIDs.. because we're abusing notation -- but we don't really care for `feed.atom` right?

isoos:
I've adapted the random uuid formatter (updating the correct bits) from pub client. PTAL.

PR #3456, opened 2020-03-24
Remove unused pub_server exception.

isoos:
#3343


isoos:
Closing as #3458 already has it.

PR #3458, opened 2020-03-25
Migrate pub_server exceptions.

isoos:
- #3343
- tested with 1.24.3 and 2.8.0-dev.17.0 too

### DIFF
--- a/app/test/package/backend_test.dart
+++ b/app/test/package/backend_test.dart
@@ -217,7 +217,7 @@ void main() {
             throwsA(isException.having(
                 (e) => '$e',
                 'text',
-                'We have sent an invitation to $newUploader, '
+                'OperationForbidden(403): We have sent an invitation to $newUploader, '
                     'they will be added as uploader after they confirm it.')));
 
         // uploaders do not change yet
@@ -274,7 +274,7 @@ void main() {
         await expectLater(
             rs,
             throwsA(isException.having((e) => '$e', 'toString',
-                'LastUploaderRemoved: Cannot remove last uploader of a package.')));
+                'OperationForbidden(403): Cannot remove last uploader of a package.')));
       });
 
       testWithServices('cannot remove non-existent uploader', () async {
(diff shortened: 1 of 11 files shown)


app/test/package/backend_test.dart
```diff
@@ -572,11 +572,11 @@ void main() {
           registerAuthenticatedUser(hansUser);
 
           expect(await fn('with'),
-              'Package name must not be a reserved word in Dart.');
+              'PackageRejected(400): Package name must not be a reserved word in Dart.');
```
jonasfj: IMO; it would also be fine to refactor this such that `fn` returns the `code` from the response... that way we assert on something less fragile.

app/test/shared/handlers_test_utils.dart
```diff
@@ -19,6 +19,11 @@ Future expectApiException(Future future,
         (e) => e.bodyAsJson(),
         'bodyAsJson',
         {
+          'error': {
+            'code': code ?? isNotNull,
+            'message': message == null ? isNotNull : contains(message),
+          },
+          // TODO: remove after the above gets deployed live
```
jonasfj: Curious, why?

isoos: Why keep it duplicated?

app/test/package/backend_test.dart
isoos: Good point, I'll take a look in a follow-up PR.

app/test/shared/handlers_test_utils.dart
jonasfj: Oh, curious why you keep it duplciated for now, instead of just removing it? (I'm probably missing some reason).

isoos: There is a chance that somebody loaded the page, we switch traffic to new version, they submit the form, it returns an error, and our frontend is unable to process it because the JSON content is not where we are expecting... Which reminds me: we'll need to change the frontend to read the `{'error': {'message': 'XX'}}` too.

isoos: PTAL on the last commit

PR #3459, opened 2020-03-25
Do not autofocus search bar when query text was specified.

### LINKED ISSUE #3452: Search field steals focus too aggressively
When I visit pub.dev, the "Search packages" field takes focus, preventing me from scrolling the page with PgDn or Space on Firefox.

This is particularly annoying when I'm browsing through search results on pub.dev since every time I navigate to the next page of search results, the Search field steals focus, and I must remember to switch to a mouse or to tab away from the Search box to scroll through the search results.

I'm using Firefox 74.0 (64-bit) on Linux.  This is less of an issue with Chrome since it seems that Chrome allows PgDn/PgUp to scroll the page even if a field has a focus.  Space/Shift+Space obviously won't work though.

In contrast, the search field on google.com does not steal focus when showing search results.

isoos:
Fixes #3452. We could probably remove autofocus in other cases, but for now I'd rather keep it to the really justified one: when the user entered some text in search.

### DIFF
--- a/app/lib/frontend/templates/layout.dart
+++ b/app/lib/frontend/templates/layout.dart
@@ -178,7 +178,10 @@ String _renderSearchBanner({
   }
   final isExperimental = requestContext.isExperimental;
   return templateCache.renderTemplate('shared/search_banner', {
-    'show_autofocus': queryText == null,
+    // When search is active (query text has a non-empty value) users may expect
+    // to scroll through the results via keyboard. We should only autofocus the
+    // search field when there is no active search.
+    'autofocus': queryText == null,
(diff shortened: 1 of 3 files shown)


app/lib/frontend/templates/layout.dart
```diff
@@ -178,6 +178,7 @@ String _renderSearchBanner({
   }
   final isExperimental = requestContext.isExperimental;
   return templateCache.renderTemplate('shared/search_banner', {
+    'show_autofocus': queryText == null,
```
jonasfj: Why? does this deserve a comment?

PR #3465, opened 2020-03-26
Updated like: style + label

isoos:
- Server-rendered page didn't use the compact number format. Updated to use it.
- Added weight to new design:
  Screen Shot 2020-03-26 at 16 43 55


isoos:
I've also considered using "like" instead of "0 likes", but it feels more like a button, but clicking on it does nothing.

PR #3475, opened 2020-03-31
Migrate pub_server's startUpload.

isoos:
- #3343
- tested with 1.24.3 and 2.8.0-dev.17.0 too

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -433,7 +433,7 @@ class GCloudPackageRepository extends pub_server.PackageRepository {
   }
 
   @override
-  Future<pub_server.AsyncUploadInfo> startAsyncUpload(Uri redirectUrl) async {
+  Future<api.AsyncUploadInfo> startAsyncUpload(Uri redirectUrl) async {
     _logger.info('Starting async upload.');
     // NOTE: We use a authenticated user scope here to ensure the uploading
     // user is authenticated. But we're not validating anything at this point
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -433,7 +433,7 @@ class GCloudPackageRepository extends pub_server.PackageRepository {
   }
 
   @override
-  Future<api.AsyncUploadInfo> startAsyncUpload(Uri redirectUrl) async {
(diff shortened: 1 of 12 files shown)


pkg/client_data/lib/package_api.dart
```diff
@@ -3,9 +3,33 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:json_annotation/json_annotation.dart';
+import 'package:meta/meta.dart';
 
 part 'package_api.g.dart';
 
+/// Information obtained when starting an asynchronous upload.
+@JsonSerializable()
+class AsyncUploadInfo {
```
jonasfj, line 12: I think we can just call it `UploadInfo` we have no sync-flow.

pkg/client_data/lib/package_api.dart
```diff
@@ -3,9 +3,33 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:json_annotation/json_annotation.dart';
+import 'package:meta/meta.dart';
 
 part 'package_api.g.dart';
 
+/// Information obtained when starting an asynchronous upload.
```
jonasfj, line 10: Information for uploading a package.

pkg/client_data/lib/package_api.dart
jonasfj: Or we can call it `UploadUrl` or `UploadUrlInfo`

app/lib/package/backend.dart
```diff
@@ -433,7 +433,7 @@ class GCloudPackageRepository extends pub_server.PackageRepository {
   }
 
   @override
-  Future<pub_server.AsyncUploadInfo> startAsyncUpload(Uri redirectUrl) async {
+  Future<api.AsyncUploadInfo> startAsyncUpload(Uri redirectUrl) async {
```
jonasfj, line 436: Let's rename this method and call it something else.

isoos: Maybe `startUpload`?

isoos: (merging this, we can do the rename later if `startUpload` is not the best one...)

jonasfj: Or `getUploadUrlInfo` or similar...

PR #3477, opened 2020-03-31
Store the user's name in Session (but not in User).

isoos:
Similar to the profile's image URL, we store this only in `UserSession`.

### DIFF
--- a/app/lib/account/models.dart
+++ b/app/lib/account/models.dart
@@ -130,6 +130,9 @@ class UserSession extends db.ExpandoModel {
   String email;
 
   @db.StringProperty()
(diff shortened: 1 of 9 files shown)


app/lib/account/models.dart
```diff
@@ -129,6 +129,9 @@ class UserSession extends db.ExpandoModel {
   @db.StringProperty(required: true)
   String email;
 
+  @db.StringProperty()
+  String name;
```
jonasfj, line 133: Can we give this a comment, saying what it is? It's not a unique name, right? Can it be null?

PR #3478, opened 2020-03-31
More whitespace in package page detail header.

isoos:


isoos:
> Is isLoose only relevant some places?

Yes, publisher page uses the same detail tab model, and in the new design it is even more complex (different background and layout). Maybe once we migrated, a code cleanup could separate these templates better.

PR #3493, opened 2020-04-03
no-referrer for consent page

### LINKED ISSUE #3490: Consent flow related packages should have `Referrer-Policy: no-referrer` #2430
`Referrer-Policy: no-referrer` ensures that the token is never sent in `Referer` header, when navigating away.. This is a good idea when the querystring contains a secret token.

Docs: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy

### LINKED ISSUE #3491: Set referrer-policy: no-referrer-when-downgrade
No reason not to use `referrer-policy: no-referrer-when-downgrade` for all pages other than consent flow: https://github.com/dart-lang/pub-dev/issues/3490

See also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy

IMO, it's reasonable to external pages to get information about which page on `pub.dev` is linking to them.

isoos:
- Fixes #3491.
- Fixes #3490.

### DIFF
--- a/app/lib/frontend/handlers/account.dart
+++ b/app/lib/frontend/handlers/account.dart
@@ -111,6 +111,8 @@ Future<shelf.Response> consentPageHandler(
       title: consent.titleText,
       descriptionHtml: consent.descriptionHtml,
     ),
+    // Consent pages have the consent ID in the URL. Browsers should not pass on
+    // this ID to the pages that are linked from the consent page.
     noReferrer: true,
   );
 }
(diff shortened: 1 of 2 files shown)


app/lib/frontend/handlers/account.dart
```diff
@@ -105,11 +105,14 @@ Future<shelf.Response> consentPageHandler(
...
-    title: consent.titleText,
-    descriptionHtml: consent.descriptionHtml,
-  ));
+  return htmlResponse(
+    renderConsentPage(
+      consentId: consentId,
+      title: consent.titleText,
+      descriptionHtml: consent.descriptionHtml,
+    ),
+    noReferrer: true,
```
jonasfj, line 114: Please add a comment why this is important.

PR #3497, opened 2020-04-06
Migrate pub_server's download method.

isoos:
- #3343
- Tested with the latest 1.24 and the latest 2.8 dev SDKs.

### DIFF
--- a/app/lib/shared/utils.dart
+++ b/app/lib/shared/utils.dart
@@ -417,3 +417,12 @@ String createUuid([List<int> bytes]) {
     bytes.sublist(10).map(formatByte).join(),
   ].join('-');
 }
+
+bool isSemanticVersion(String version) {
+  try {
+    semver.Version.parse(version);
+    return true;
+  } catch (_) {
+    return false;
+  }
+}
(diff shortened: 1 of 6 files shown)


app/lib/shared/utils.dart
```diff
@@ -417,3 +417,12 @@ String createUuid([List<int> bytes]) {
     bytes.sublist(10).map(formatByte).join(),
   ].join('-');
 }
+
+bool isSemanticVersion(String version) {
+  try {
+    semver.Version.parse(version);
+    return true;
+  } catch (_) {
```
jonasfj, line 425: Can this be anything other than `FormatException`?

app/lib/shared/utils.dart
```diff
@@ -417,3 +417,12 @@ String createUuid([List<int> bytes]) {
     bytes.sublist(10).map(formatByte).join(),
   ].join('-');
 }
+
+bool isSemanticVersion(String version) {
```
jonasfj, line 421: We could also make this a static method on `InvalidInputException`

app/lib/shared/utils.dart
jonasfj: Just saying, if we want to be perfectly safe, it's probably better to ignore `FormatException`, and if throws anything log it with `shout()` (and ignore it).. Or just only catch `FormatException` I'm pretty sure that's the only thing this will throw.

PR #3499, opened 2020-04-06
Migrate pub_server's uploader handlers

isoos:
- #3343
- Tested with the latest 1.24 and the latest 2.8 dev SDKs.

### DIFF
--- a/app/lib/frontend/handlers/pubapi.dart
+++ b/app/lib/frontend/handlers/pubapi.dart
@@ -120,14 +120,15 @@ class PubApi {
   ///     [400 Client Error]
   @EndPoint.post('/api/packages/<package>/uploaders')
   Future<SuccessMessage> addUploader(Request request, String package) async {
-    final body = await request.readAsString();
-    final parts = body.split('=');
-    InvalidInputException.check(
-        parts.length == 2 && parts[0] == 'email' && parts[1].isNotEmpty,
-        'Invalid request.');
-
-    final user = Uri.decodeQueryComponent(parts[1]);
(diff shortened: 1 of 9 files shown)


app/lib/frontend/handlers/pubapi.dart
```diff
@@ -119,9 +120,15 @@ class PubApi {
...
-  Future<Response> addUploader(Request request, String package) async {
+  Future<SuccessMessage> addUploader(Request request, String package) async {
     final body = await request.readAsString();
-    return await packageBackend.pubServer.addUploader(package, body);
+    final parts = body.split('=');
+    InvalidInputException.check(
+        parts.length == 2 && parts[0] == 'email' && parts[1].isNotEmpty,
+        'Invalid request.');
+
+    final user = Uri.decodeQueryComponent(parts[1]);
```
jonasfj, line 130: How about just try parsing it with [`Uri.splitQueryString`](https://api.dart.dev/stable/2.7.2/dart-core/Uri/splitQueryString.html), catch any `FormatException` and throw if there is no `email` key.

This seems to be `application/x-www-form-urlencoded` which is what [`Uri.splitQueryString`](https://api.dart.dev/stable/2.7.2/dart-core/Uri/splitQueryString.html) is designed to parse.

isoos:
> We had changes in 2.8, so I would suggest also testing with 2.7.

That's a given (both on CI and locally).

> Maybe we can make a list of versions we want to test compatibility with somewhere?

That makes sense only if we automate it. E.g. these tools download and run an explicit version of the SDK.

PR #3504, opened 2020-04-07
Migrate pub_server's version API

isoos:
- #3343
- Tested with the latest 1.24 and the latest stable and the latest 2.8 dev SDKs.

### DIFF
--- a/app/lib/frontend/handlers/pubapi.client.dart
+++ b/app/lib/frontend/handlers/pubapi.client.dart
@@ -33,11 +33,12 @@ class PubApiClient {
     );
   }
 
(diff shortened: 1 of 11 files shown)


jonasfj:
You might want to test with `2.0.0` and `2.6.0` as we have deprecated this end-point as of `2.8.0`.

isoos:
> You might want to test with 2.0.0 and 2.6.0 as we have deprecated this end-point as of 2.8.0.

IIRC, as 2.0 -> 2.7 is unchanged (?), I think we need to keep testing with `2.7` after `2.8` is released, but I don't think adding `2.0` or `2.6` will add significant value to it. Was there any significant change that warrants the extra testing?

PR #3505, opened 2020-04-08
Refactor modal dialog, explicitly safe content.

isoos:
The goal of this change is to prepare for modal forms (like adding a new member to a publisher):
- Enables override of 'Ok' and 'Cancel' button labels.
- Customization messages are provided `Element` instances, `text()` and `markdown()` methods can be used to easily create the required content types.

### DIFF
--- a/pkg/web_app/lib/src/_dom_helper.dart
+++ b/pkg/web_app/lib/src/_dom_helper.dart
@@ -5,7 +5,7 @@
 import 'dart:async';
 import 'dart:html';
 
-import 'package:markdown/markdown.dart' as markdown;
+import 'package:markdown/markdown.dart' as md;
 import 'package:mdc_web/mdc_web.dart' show MDCDialog;
 import 'package:meta/meta.dart';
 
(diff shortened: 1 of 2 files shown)


pkg/web_app/lib/src/_dom_helper.dart
```diff
@@ -209,14 +221,24 @@ Element _buildDialog(
...
         Element.div()..classes.add('mdc-dialog__scrim'),
       ];
 
+/// Creates an [Element] with unformatted [text] content.
+Element text(String text) => Element.div()..text = text;
+
+/// Creates an [Element] with Markdown-formatted content.
+Element markdown(String text) => Element.div()
+  ..setInnerHtml(
+    md.markdownToHtml(text),
```
jonasfj, line 238: This uses `sanitize_html` right?

isoos: This is just `dart:html`'s `innerHtml` setter's validator. However the markdown source is either known or coming from our backend. We could add extra checks but probably non-blocking to this PR.

PR #3508, opened 2020-04-09
Migrating pub_server: versions list API.

isoos:
- #3343
- Tested with the latest 1.24 and the latest stable and the latest 2.8 dev SDKs.

- Caching is less efficient, but I'm not sure what the real impact would be.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -377,14 +377,27 @@ class GCloudPackageRepository extends pub_server.PackageRepository {
 
   GCloudPackageRepository(this.db, this.storage);
 
-  // Metadata support.
-
-  @override
-  Stream<pub_server.PackageVersion> versions(String package) {
-    final packageKey = db.emptyKey.append(Package, id: package);
-    final query = db.query<PackageVersion>(ancestorKey: packageKey);
-    return query.run().map((model) => pub_server.PackageVersion(
-        package, model.version, model.pubspec.jsonString));
+  /// Returns the known versions of [package].
+  Future<api.PackageData> packageData(Uri baseUri, String package) async {
+    return await cache.packageData(package).get(() async {
+      final packageVersions = await packageBackend.versionsOfPackage(package);
+      if (packageVersions.isEmpty) {
+        throw NotFoundException.resource('package "$package"');
+      }
+      packageVersions
+          .sort((a, b) => a.semanticVersion.compareTo(b.semanticVersion));
+      final latest = packageVersions.lastWhere(
+        (pv) => !pv.semanticVersion.isPreRelease,
+        orElse: () => packageVersions.last,
+      );
+      return api.PackageData(
+        name: package,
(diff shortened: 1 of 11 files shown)


app/lib/package/backend.dart
```diff
@@ -377,14 +377,27 @@ class GCloudPackageRepository extends pub_server.PackageRepository {
...
-  // Metadata support.
-
-  @override
-  Stream<pub_server.PackageVersion> versions(String package) {
-    final packageKey = db.emptyKey.append(Package, id: package);
-    final query = db.query<PackageVersion>(ancestorKey: packageKey);
-    return query.run().map((model) => pub_server.PackageVersion(
-        package, model.version, model.pubspec.jsonString));
+  /// Returns the known versions of [package].
+  Future<api.PackageData> packageData(Uri baseUri, String package) async {
```
jonasfj, line 381: Can we call this `listVersions`?

pkg/client_data/lib/package_api.dart
```diff
@@ -88,6 +88,24 @@ class Message {
   Map<String, dynamic> toJson() => _$MessageToJson(this);
 }
 
+@JsonSerializable()
+class PackageData {
```
jonasfj, line 92: Let's give this a documentation comment that this is used in `listVersions` API.

app/lib/package/backend.dart
jonasfj: Ideally, also a small doc comment that this is what we use in `pub` for finding which versions exist.

pkg/client_data/lib/package_api.dart
```diff
@@ -88,6 +88,24 @@ class Message {
...
+  final String name;
+  final VersionInfo latest;
```
jonasfj, line 94: Can we add a note that this is merely a convenience property, because the `VersionInfo` for the latest version also exists in the `versions` list.

pkg/client_data/lib/package_api.dart
```diff
@@ -88,6 +88,24 @@ class Message {
   Map<String, dynamic> toJson() => _$MessageToJson(this);
 }
 
+@JsonSerializable()
+class PackageData {
+  final String name;
```
jonasfj, line 93: Documentation comment saying `/// Package name`.

pkg/client_data/lib/package_api.dart
```diff
@@ -88,6 +88,24 @@ class Message {
...
+  final VersionInfo latest;
+  final List<VersionInfo> versions;
```
jonasfj, line 95: Can we add a note in `VersionInfo` to document that as of Dart 2.8 `pub` client uses `archive_url` to find the archive.

Also we can add a documentation comment that these are sorted by version number... how they are sorted.

pkg/client_data/lib/package_api.dart
jonasfj: Can we document that this is the latest that is NOT a pre-release, unless there only is pre-releases.

PR #3509, opened 2020-04-09
Kill (and restart) worker isolate after 24-48 hours.

isoos:


app/lib/service/entrypoint/_isolate.dart
```diff
@@ -138,9 +141,23 @@ Future startIsolates({
...
+    //
+    // This clears us from temporary processing hangups, at most within 2 days,
+    // probably in a few hours (as we have many workers and probably a few will
+    // be triggered earlier).
+    final ttl =
+        Duration(hours: 24 + _random.nextInt(24), minutes: _random.nextInt(60));
+    logger.info('Worker isolate #$workerIndex TTL set to $ttl.');
+    final autoKillTimer = Timer(ttl, () {
+      logger.info('Killing worker isolate #$workerIndex...');
+      isolate.kill();
```
jonasfj, line 154: Won't this cause some analysis tasks to fail? How long before they are retried?

That's what I meant by terminating gracefully.

isoos: > Won't this cause some analysis tasks to fail? How long before they are retried?

Good point. I've updated the code the include an alive-callback mechanism, where the worker will send regular pings to the controller, and that resets the timer. While working, it won't forcefully restart the isolate.

app/lib/service/entrypoint/dartdoc.dart
```diff
@@ -53,6 +53,7 @@ class DartdocCommand extends Command {
       frontendEntryPoint: _frontendMain,
       workerSetup: workerSetup,
       workerEntryPoint: _workerMain,
+      deadWorkerTimeout: Duration(hours: 4),
```
jonasfj, line 56: so if a worker doesn't ping us back in 4-8 hours it gets killed? Seems like 1 hour would be enough? Should any task ever take 1 hour?

jonasfj: or maybe 2 or 4 :D hehe, just thinking...

isoos: 1 hour works too. I wanted to be cautious, but even 1 hour is much more than our standard limit (e.g. I think we kill processes after 10 minutes or so).

PR #3520, opened 2020-04-14
First part of mirror configuration.

isoos:
I wanted to keep the flexibility at bay, and I think this has nice structure. This does not yet define how the resolved lockfile should look like, that but I imagine it could be similarly modelled, while the tool outputs it to YAML.

### DIFF
--- a/app/lib/tool/mirror/mirror_config.dart
+++ b/app/lib/tool/mirror/mirror_config.dart
@@ -1,0 +1,181 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+import 'package:json_annotation/json_annotation.dart';
+import 'package:meta/meta.dart';
+import 'package:yaml/yaml.dart' as yaml;
+
+part 'mirror_config.g.dart';
+
+/// The configuration to use when creating a local (partial) mirror of pub.dev
+@JsonSerializable(explicitToJson: true, includeIfNull: false)
+class MirrorConfig {
+  List<Package> packages;
+  List<Publisher> publishers;
+  List<User> users;
+  String defaultUser;
+
+  MirrorConfig({
+    @required this.packages,
+    @required this.publishers,
+    @required this.users,
+    this.defaultUser,
+  });
+
+  factory MirrorConfig.fromJson(Map<String, dynamic> json,
+      {bool normalize = false}) {
+    final mc = _$MirrorConfigFromJson(json);
+    if (normalize) mc.normalize();
+    return mc;
+  }
+
+  factory MirrorConfig.fromYaml(String source, {bool normalize = false}) {
+    final map = json.decode(json.encode(yaml.loadYaml(source)));
+    final mc = MirrorConfig.fromJson(map as Map<String, dynamic>);
+    if (normalize) mc.normalize();
+    return mc;
+  }
+
+  Map<String, dynamic> toJson() => _$MirrorConfigToJson(this);
+
+  void normalize() {
+    packages ??= <Package>[];
+    publishers ??= <Publisher>[];
+    users ??= <User>[];
+
+    for (final publisher in publishers) {
+      publisher.members ??= <Member>[];
+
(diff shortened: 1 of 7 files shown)


app/lib/tool/mirror/mirror_config.dart
```diff
@@ -0,0 +1,181 @@
...
+      _$PackageFromJson(json);
+
+  Map<String, dynamic> toJson() => _$PackageToJson(this);
+}
+
+@JsonSerializable(explicitToJson: true, includeIfNull: false)
+class Publisher {
+  final String name;
+  List<Member> members;
+  List<Package> packages;
```
jonasfj, line 137: Maybe we should remove this, because packages can be inserted at the top-level and assigned a `publisher` property, instead of `uploaders`.

I get why we allow for `Package` to have either `uploaders` or `publisher`... In this case, I then propose that we avoid nesting... and only have one way to define it.

app/lib/tool/mirror/mirror_config.dart
```diff
@@ -0,0 +1,181 @@
...
+          package.uploaders.add(defaultUser);
+        }
+      }
+    }
+  }
+
+  void _createUserIfNeeded(String email) {
+    final user = users.firstWhere((u) => u.email == email, orElse: () => null);
+    if (user == null) {
+      users.add(User(email: email));
```
jonasfj, line 108: This will have `User.likes = null`, let's just give `likes` a default value.

app/lib/tool/mirror/mirror_config.dart
```diff
@@ -0,0 +1,181 @@
...
+      publisher.members ??= <Member>[];
+
+      // copy package definitions to [MirrorConfig.packages]
+      if (publisher.packages != null) {
+        publisher.packages.forEach((p) {
+          p.publisher = publisher.name;
+          packages.add(p);
+        });
+        publisher.packages = null;
+      }
```
jonasfj, line 59: I propose we omit this... I see that with Dart typing we don't want to have two kinds of `Package` objects, and then this structure is not ideal.

app/lib/tool/mirror/mirror_config.dart
```diff
@@ -0,0 +1,181 @@
...
+
+      // convert admins to members
+      if (publisher.admins != null) {
+        publisher.members.addAll(
+            publisher.admins.map((e) => Member(email: e, role: 'admin')));
+        publisher.admins = null;
+      }
```
jonasfj, line 66: I suggest we omit this

app/lib/tool/mirror/mirror_config.dart
```diff
@@ -0,0 +1,181 @@
...
+
+      // add missing users
+      publisher.members?.forEach((m) => _createUserIfNeeded(m.email));
```
jonasfj, line 69: This shorthand makes sense :D

app/lib/tool/mirror/mirror_config.dart
```diff
@@ -0,0 +1,181 @@
...
+    for (final package in packages) {
+      // add missing users
+      package.uploaders?.forEach(_createUserIfNeeded);
+
+      // add missing publishers
+      if (package.publisher != null) {
+        final publisher = publishers
+            .firstWhere((p) => p.name == package.publisher, orElse: () => null);
+        if (publisher == null) {
+          publishers.add(Publisher(name: package.publisher, members: []));
```
jonasfj, line 81: Hmm, how about we add the defaultUser as member in these cases?

An abandon publisher should still be possible to make, by explicitly stating `members: []´ in `publishers:`... But for the automatically created publishers I suggest default user,

app/lib/tool/mirror/mirror_config.dart
```diff
@@ -0,0 +1,181 @@
...
+        }
+      }
+    }
+
+    defaultUser ??= users.first.email;
+    _createUserIfNeeded(defaultUser);
+
+    for (final publisher in publishers) {
+      if (publisher.members.isEmpty) {
+        publisher.members.add(Member(email: defaultUser, role: 'admin'));
```
jonasfj, line 91: May I propose we don't do this for publishers in `publishers:`, but only for the ones created from `package.publisher`.

I'd argue that creating a publisher without members might be interesting (for testing). See also comment above.

PR #3523, opened 2020-04-15
Remove archive only on successful upload.

isoos:


isoos:
> Do we have any periodic cleanup?

No. We did a manual cleanup ~ a year ago.

isoos:
To clarify: we do have leftovers regardless: either they don't invoke the callback url or they are failing on the prerequisite checks.

PR #3525, opened 2020-04-15
New UI: add uploader dialog.

isoos:
- The new dialog window enables us to stack them on top of each other (like on the screenshot below): a dialog with the main form, then a second dialog to communicate errors, while the core dialog is not immediately hidden.

- The form content is generated the same way on the server side mustache, but it is removed from the DOM in the browser.

Screen Shot 2020-04-15 at 12 10 16

### DIFF
--- a/pkg/web_app/lib/src/_dom_helper.dart
+++ b/pkg/web_app/lib/src/_dom_helper.dart
@@ -114,7 +114,7 @@ String _requestExceptionMessage(RequestException e) {
 
 /// Displays a message via the modal window.
 Future modalMessage(String title, Element content) async {
-  await _modalWindow(
+  await modalWindow(
     titleText: title,
     content: content,
     isQuestion: false,
@@ -124,7 +124,7 @@ Future modalMessage(String title, Element content) async {
 /// Display [content] to user (assumed it has a question format), and return
 /// true if 'OK' was selected.
 Future<bool> modalConfirm(Element content) async {
-  return await _modalWindow(
+  return await modalWindow(
     titleText: 'Confirm',
     content: content,
     isQuestion: true,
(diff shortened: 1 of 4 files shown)


pkg/web_app/lib/src/account.dart
```diff
@@ -345,17 +355,29 @@ class _PublisherAdminWidget {
...
+    await modalWindow(
+      titleText: 'Invite new member',
+      isQuestion: true,
+      okButtonText: 'Add',
+      content: _addMemberContent,
+      onOkButtonClicked: () => _inviteMember(false),
+    );
+  }
+
+  Future<bool> _inviteMember(bool requestConfirm) async {
```
jonasfj, line 368: instead of introducing `requestConfirm` what about making two methods `_inviteMember` and `_inviteMemberExperimental`, then it's easy to remove `_inviteMember` and rename `_inviteMemberExperimental` when the experimental part goes away.

pkg/web_app/lib/src/_dom_helper.dart
```diff
@@ -135,19 +135,31 @@ Future<bool> modalConfirm(Element content) async {
...
   @required Element content,
   @required bool isQuestion,
   String cancelButtonText,
   String okButtonText,
+
+  /// If specified, this callback function will be called on "OK" button clicks,
+  /// and the dialog window will be kept open. Returns with the success status
+  /// of the operation, and if `true`, the dialog will be closed, otherwise the
+  /// dialog is kept open (e.g. the user may update the form).
+  FutureOr<bool> Function() onOkButtonClicked,
```
jonasfj, line 149: Hmm... so RPC is running while the dialog is open... that means that `await modalMessage('Success', successMessage);` is also displayed on top of the creation dialog.

Hmm, could we call this `execute` or some other name?

pkg/web_app/lib/src/account.dart
isoos: Added a `TODO` instead. I think duplicating the method for the 3-4 lines of difference is not worth the risk of fixing something in one and missing it in the other. When `requestConfirm` is removed, it will be a clean update, it is used only in one conditional expression.

PR #3528, opened 2020-04-16
Migrate pub_server's finishUploadAsync.

### LINKED ISSUE #3343: Migrate away from using package:pub_server

isoos:
- Tested with 1.X and 2.8-dev SDKs.
- Fixes #3343. (Technically we have a few refactor left, but those don't change behaviour.)


app/lib/package/backend.dart
```diff
@@ -480,31 +472,28 @@ class GCloudPackageRepository extends pub_server.PackageRepository {
   }
 
   /// Finishes the upload of a package.
-  @override
-  Future<pub_server.PackageVersion> finishAsyncUpload(Uri uri) async {
+  Future<PackageVersion> finishAsyncUpload(Uri uri) async {
```
jonasfj, line 475: We could give this a better name... maybe `processUploadedBlob` or `publishUploadedBlob`

PR #3530, opened 2020-04-16
Redirect from #-<tab-name>- to /tab URLs.

isoos:
- This makes the transition to the new URLs seamless.
- #3390

### DIFF
--- a/pkg/web_app/lib/src/tabs.dart
+++ b/pkg/web_app/lib/src/tabs.dart
@@ -79,6 +79,10 @@ String getTabName(Element elem) {
   return getTabName(elem.parent);
 }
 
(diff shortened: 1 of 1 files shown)


pkg/web_app/lib/src/tabs.dart
```diff
@@ -81,10 +81,14 @@ String getTabName(Element elem) {
 
 void changeTab(String name) {
   final tabOrContentElem = getTabElement(name);
-  if (tabOrContentElem != null &&
-      tabOrContentElem.classes.contains('tab-button')) {
+  print(tabOrContentElem);
```
jonasfj, line 84: debug statement?

Issue #3500, opened 2020-04-07
Sort by likes


jonasfj:
When searching / listing packages it should be possible to sort package by likes.

![DWOBw6Gczvn](https://user-images.githubusercontent.com/149732/78653032-d6be3680-78c2-11ea-9eb8-fe22d2964600.png)

PR #3535, opened 2020-04-17
Include like count in package index.

isoos:
#3500


app/lib/search/index_simple.dart
```diff
@@ -353,30 +356,38 @@ class SimplePackageIndex implements PackageIndex {
     );
   }
 
-  // visible for testing only
+  @visibleForTesting
```
jonasfj, line 359: Lovely

PR #3536, opened 2020-04-17
Resolve test-profile packages.

isoos:


app/lib/tool/test_profile/resolver.dart
```diff
@@ -0,0 +1,73 @@
...
+        if (pr.exitCode != 0) {
+          throw Exception(
+              'pub get on `${package.name} $version` exited with ${pr.exitCode}.\n${pr.stderr}');
+        }
+
+        await dummyDir.delete(recursive: true);
+      }
+    }
+    final pubHostedDir =
+        Directory(p.join(pubCacheDir.path, 'hosted', 'pub.dartlang.org'));
```
jonasfj, line 51: I think it's more stable to read the `pubspec.lock`.

app/lib/tool/test_profile/resolver.dart
```diff
@@ -0,0 +1,73 @@
...
+        .whereType<Directory>()
+        .map((d) => p.basename(d.path))
+        .where((v) => v.contains('-'))
+        .map((v) => v.replaceFirst('-', ':'))
+        .toList()
+          ..sort();
+  });
+}
+
+String _generateDummyPubspec(String package, String version) {
```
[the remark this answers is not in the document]
isoos: Good point, I keep forgetting that option... fixed.

app/lib/tool/test_profile/resolver.dart
isoos: It is also more complex, as we'd need to read it after each `pub get`, and parse it, merge it.. I think for this purpose we can keep just reading the directories, and when the SDK changes this, we adopt.

Issue #3519, opened 2020-04-14
tags in package version listing API


jonasfj:
Let's add _package level tags_ and _package-version level tags_ to version listing API.
```js
// GET <PUB_HOSTED_URL>/api/packages/<PACKAGE>
{
  "name": "<PACKAGE>",
  "tags": [ // <--- NEW
    "is:discontinued",
    "is:flutter-favorite",
    "publisher:<publisher>", // if package has a publisher
    /* other package level tags*/,
  ]
  "latest": {
    "version": "<VERSION>",
    "archive_url": "https://.../archive.tar.gz",
    "tags": [ // <--- NEW
      "sdk:flutter",
      "platform:web",
      /* other package-version level tags */
    ],
    "pubspec": {
      /* pubspec contents as JSON object */
    }
  },
  "versions": [
    {
      "version": "<VERSION>",
      "archive_url": "https://.../archive.tar.gz",
      "tags": [ // <--- NEW
        "sdk:flutter",
        "platform:web",
        /* other package-version level tags */
      ],
      "pubspec": {
        /* pubspec contents as JSON object */
      }
    },
    /* additional versions */
  ]
}
```

isoos:
@jonasfj: I think "other package level tags" only applies to the first block (under `name`), while the rest should be "other version-level tags". Is this assumption correct?

jonasfj:
Correct, I updated it. I obviously meant `package-version level tags`.

isoos:
#3540 exposes the readily available tags, but some (e.g. the ones from analysis) are not cheaply available, as we'd need to load the related `ScoreCard` instance for getting access to those.

Some possible alternatives:
- we only load it for the `latest` version
- we only load it for the version-specific endpoint (so the package aggregate version list does not have it in full detail)
- we update the `PackageVersion[Info]` object with the analysis result of the latest runtime (has some inconsistency when rolling back a version)
- we create a package-level aggregate object with the versions and their analysis results
- .... (your idea comes here)

wdyt?

jonasfj:
> we only load it for the version-specific endpoint (so the package aggregate version list does not have it in full detail)

This end-point has been deprecated.

> we update the `PackageVersion[Info]` object with the analysis result of the latest runtime (has some inconsistency when rolling back a version)

I think this is the best option. I'm even open to making `PackageVersionInfo` runtime versioned.
Since it only reflects derived values we can make it runtimeVersioned and regenerate everything when we make a new `runtimeVersion`.

[...]

PR #3540, opened 2020-04-20
Some tags in /api/packages/<pkg>

isoos:
- #3519
- not all tags are exposed, I'll update the issue with the related notes

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -324,6 +324,7 @@ class PackageBackend {
   /// Used in `pub` client for finding which versions exist.
   Future<api.PackageData> listVersions(Uri baseUri, String package) async {
     return await cache.packageData(package).get(() async {
+      final pkg = await packageBackend.lookupPackage(package);
       final packageVersions = await packageBackend.versionsOfPackage(package);
       if (packageVersions.isEmpty) {
         throw NotFoundException.resource('package "$package"');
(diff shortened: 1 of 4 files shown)


pkg/client_data/lib/package_api.dart
```diff
@@ -95,6 +95,9 @@ class PackageData {
   /// Package name.
   final String name;
 
+  /// Package-level tags
```
jonasfj, line 98: How about: `/// Tags applicable to all versions of this package`

pkg/client_data/lib/package_api.dart
```diff
@@ -121,6 +125,9 @@ class PackageData {
 class VersionInfo {
   final String version;
 
+  /// Version-level tags.
```
jonasfj, line 128: `/// Tags applicable to a specific version of a package`

app/lib/package/backend.dart
```diff
@@ -324,6 +324,7 @@ class PackageBackend {
   /// Used in `pub` client for finding which versions exist.
   Future<api.PackageData> listVersions(Uri baseUri, String package) async {
     return await cache.packageData(package).get(() async {
```
jonasfj, line 326: We're changing the format of the data in the cache, let's bump the `runtimeVersion` for good measure.

sigurdm:
Should we document these in https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md ?

app/lib/package/backend.dart
isoos: That's already happening due to the dartdoc version upgrade.

PR #3545, opened 2020-04-21
Profile image on new my-profile pages.

isoos:
- added a parent wrapper (with display: flex)
- image is scaled down on mobile view

Screenshot 2020-04-21 at 18 01 06


app/lib/frontend/templates/detail_page.dart
```diff
@@ -36,6 +37,8 @@ String renderDetailHeader({
     'like_count': _formatPackageLikes(packageLikes),
     'is_liked': isLiked,
     'has_likes': isLiked != null,
+    'image_url': imageUrl,
```
jonasfj, line 40: When using this we must add a size argument, see the other place we use it... Also let's add this comment to `SessionUserData`
```
// Set image size to 30x30 pixels for faster loading, see:
// https://developers.google.com/people/image-sizing
```

Issue #3553, opened 2020-04-22
Admin UI for add/remove uploaders


jonasfj:
We very rarely need admin API for adding/removing package uploaders and publisher members.
We currently do this using CLI tools, however, these have no testing and is broken from time to time.

isoos:
Duplicate of #3553, closing this one.

PR #3555, opened 2020-04-23
Commit constraint checks in fake_gcloud.

isoos:
In preparation for #3553.

### DIFF
--- a/pkg/fake_gcloud/lib/mem_datastore.dart
+++ b/pkg/fake_gcloud/lib/mem_datastore.dart
@@ -56,6 +56,11 @@ class MemDatastore implements Datastore {
       }
     }
 
+    // TODO: check serializability.
+    // We need to track the keys that have been mutated since the Transaction
+    // was created to ensure that there are no conflicts.
+    // Alternatively: block overlapping transactions.
+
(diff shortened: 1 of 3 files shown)


pkg/fake_gcloud/lib/mem_datastore.dart
```diff
@@ -34,14 +34,34 @@ class MemDatastore implements Datastore {
...
+      if (newKeys.contains(deletedKey)) {
+        throw DatastoreError('Conflicting update and delete on $deletedKey');
+      }
+    }
+
+    // execute commit
+    deletes.forEach((key) => _entities.remove(key));
+    inserts.forEach((e) {
+      _entities[e.key] = e;
+    });
```
jonasfj, line 63: Can we documentation the limitation of these checks... From what I can see this doesn't ensure serializability, which is the important part of here.

We could track the keys that have been mutated since the `Transaction` was created to ensure that there is no conflicts. Or we could simply reject any transaction that overlaps with another transaction :D

isoos: good points! I'll add TODOs for that.

PR #3556, opened 2020-04-23
Switch from tags to explicit fields in package data API

isoos:


pkg/client_data/lib/package_api.dart
```diff
@@ -108,11 +105,14 @@ class PackageData {
   /// The available versions, sorted by their semantic version number (ascending).
   final List<VersionInfo> versions;
 
+  /// `true` if package is discontinued.
+  final bool isDiscontinued;
```
jonasfj, line 109: Can we update the comment to say that if it is omitted, `null` or `false` the package is _not discontinued_.

PR #3558, opened 2020-04-23
Split Travis test execution in app/test

isoos:


isoos:
It seems that the static_files test is breaking in some configuration, not really sure why. I'll try to refactor that test (and the static file handling) first to enable a consistent caching of the built files.

isoos:
I'm not sure why the test was failing with the split and passing without it, but now it seems to be fixed and working. The total wall-clock time is reduced by about 4 minutes.

PR #3562, opened 2020-04-24
Importer for test profiles.

isoos:


app/lib/tool/test_profile/importer.dart
```diff
@@ -0,0 +1,145 @@
...
+  for (final u in profile.users) {
+    final oauthUserId = _baseIdFromEmail(u.email);
+    final userId = _userIdFromEmail(u.email);
+    await dbService.commit(inserts: [
+      User()
+        ..id = userId
+        ..oauthUserId = oauthUserId
+        ..email = u.email
+        ..created = DateTime.now().toUtc()
+        ..isDeleted = false,
```
jonasfj, line 46: Supporting these in the `TestProfile` would be nice..  and having a `defaultUserCreated` would be neat.

app/lib/tool/test_profile/importer.dart
```diff
@@ -0,0 +1,145 @@
...
+          ..created = DateTime.now().toUtc()
+          ..packageName = p,
+      )
+    ]);
+  }
+}
+
+String _baseIdFromEmail(String email) =>
+    email.replaceAll('.', '-dot-').replaceAll('@', '-at-');
+String _userIdFromEmail(String email) => '${_baseIdFromEmail(email)}-uuid';
```
jonasfj, line 145: What if we hash the email and derive a uuid from the hash?

app/lib/tool/test_profile/importer.dart
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  // create users
+  for (final u in profile.users) {
+    final oauthUserId = _baseIdFromEmail(u.email);
+    final userId = _userIdFromEmail(u.email);
+    await dbService.commit(inserts: [
+      User()
+        ..id = userId
+        ..oauthUserId = oauthUserId
```
jonasfj, line 43: Hmm, maybe we should support this in the `TestProfile` too.. and just have be optional... if not given let's do a fallback to a hash of email.

If we ever want to use to this populate a staging site and use our real accounts to sign in, then this would be nice.

app/lib/tool/test_profile/importer.dart
```diff
@@ -0,0 +1,145 @@
...
+
+  // create publishers
+  for (final p in profile.publishers) {
+    await dbService.commit(inserts: [
+      Publisher()
+        ..id = p.name
+        ..contactEmail = p.members.isEmpty ? null : p.members.first.email
+        ..websiteUrl = 'https://${p.name}/'
+        ..created = DateTime.now().toUtc()
+        ..updated = DateTime.now().toUtc()
```
jonasfj, line 61: Again it would probably be a good idea to have optional support for this in `TestProfile` with a fallback to a `defaultPublishedCreated` date-time.

app/lib/tool/test_profile/importer.dart
```diff
@@ -0,0 +1,145 @@
...
+        ..isAbandoned = false,
+      ...p.members.map(
+        (m) => PublisherMember()
+          ..parentKey = dbService.emptyKey.append(Publisher, id: p.name)
+          ..id = _userIdFromEmail(m.email)
+          ..userId = _userIdFromEmail(m.email)
+          ..created = DateTime.now().toUtc()
+          ..updated = DateTime.now().toUtc()
```
jonasfj, line 69: Let's use the minimum of _publishers created_ and _user created_.

app/lib/tool/test_profile/importer.dart
```diff
@@ -0,0 +1,145 @@
...
+
+import 'models.dart';
+import 'resolver.dart';
+
+/// Imports [profile] data into the Datastore.
+@visibleForTesting
+Future<void> importProfile({
+  @required TestProfile profile,
+  List<String> resolvedVersions,
+  @required String archiveCachePath,
```
jonasfj, line 29: Can we use `.dart_tool/pub-dev/test_package_cache/` we don't have to hardcode a default here. This is just a suggestion for later :D

app/lib/tool/test_profile/importer.dart
```diff
@@ -0,0 +1,145 @@
...
+      )
+    ]);
+  }
+
+  // create packages
+  Client client;
+  final archiveCacheDir = Directory(archiveCachePath);
+  await archiveCacheDir.create(recursive: true);
+  for (final rv in resolvedVersions) {
+    final parts = rv.split(':');
```
jonasfj, line 80: You might want to use: https://pub.dev/documentation/basics/latest/lib_string_basics/StringBasics/partition.html

So that we only split in two...

app/lib/tool/test_profile/importer.dart
```diff
@@ -0,0 +1,145 @@
...
+  // create likes
+  for (final u in profile.users) {
+    if (u.likes == null || u.likes.isEmpty) continue;
+    final userId = _userIdFromEmail(u.email);
+    await dbService.commit(inserts: [
+      ...u.likes.map(
+        (p) => Like()
+          ..parentKey = dbService.emptyKey.append(User, id: userId)
+          ..id = p
+          ..created = DateTime.now().toUtc()
```
jonasfj, line 136: Can we make this the same time the user was created?

app/lib/tool/test_profile/importer.dart
isoos: I'm adding a TODO for that.

app/lib/tool/test_profile/importer.dart
isoos: Adding a TODO...

app/lib/tool/test_profile/importer.dart
isoos: Yeah, we can do that, but we also want to have a configuration for CI to cache it there too.

app/lib/tool/test_profile/importer.dart
isoos: I've added a small hash prefix, it will be `'8157080d-dev-at-example-dot-com'`. We can tweak this later, but I wanted to keep it readable for now, for easier verification.

jonasfj: The reason I propose hashing is that I think it's best to test with UUIDs when this is also what we will be using in production.. and we can we derive deterministic UUIDs using a hash of the email.

PR #3602, opened 2020-05-07
Display macOS filter when it is part of the search query.

isoos:


app/lib/frontend/templates/layout.dart
```diff
@@ -294,6 +295,12 @@ String renderSubSdkTabsHtml({@required SearchQuery searchQuery}) {
           tag: FlutterSdkTag.platformWeb,
           title: 'Packages compatible with Flutter on the Web platform',
         ),
+        if (pred != null && pred.isRequiredTag(FlutterSdkTag.platformMacos))
```
jonasfj, line 298: Maybe a comment about why this is conditional for now would be nice :D

PR #3603, opened 2020-05-07
Back to search on package pages.

isoos:
This will be a browser-side-only feature, otherwise it would complicate the caching logic.

Screenshot 2020-05-07 at 13 15 32

### DIFF
--- a/pkg/web_app/lib/src/search.dart
+++ b/pkg/web_app/lib/src/search.dart
@@ -14,6 +14,13 @@ void setupSearch() {
   _setEventForCheckboxChanges();
 }
 
+final _acceptedSearchPaths = <Pattern>[
+  '/packages',
+  '/dart/packages',
+  '/flutter/packages',
+  RegExp(r'^/publishers/([a-z0-9-]{1,63}\.)+[a-z0-9-]{1,63}/packages$'),
+];
+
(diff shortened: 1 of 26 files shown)


pkg/web_app/lib/src/search.dart
```diff
@@ -4,13 +4,50 @@
...
+
+  // abort if the scheme://hostname[:port]/ is different
+  // - location.protocol already contains the colon
+  // - location.host already contains the :port
+  final loc = window.location;
+  final prefix = '${loc.protocol}//${loc.host}/';
+  if (!ref.startsWith(prefix)) return;
+
+  // abort if the referrer was not search
+  if (!ref.contains('/packages')) return;
```
jonasfj, line 40: Couldn't we check that there is a `q` and that it's one of the search urls we think it's supposed to be? Not a dartdoc page or other things?

It seems we should prefer to be overly cautious here.  I'd rather we don't support a particular kind of special search page, than that we have a security vulnerability.

isoos: I've updated this part a bit with proper `Uri` parsing, however, I wouldn't include the `q` parameter: it could be that the user is filtering for platform, it is on page 10, and it would be still helpful to go back to that page.

PR #3605, opened 2020-05-08
Fixed list of accepted runtimeVersions.

isoos:
- Makes the change of `gcBeforeRuntimeVersion` automatic.
- We don't need to handle blacklisted versions.
- Slightly better `ScoreCard` access while the analysis is not ready yet (using lookup instead of query).

### DIFF
--- a/app/lib/scorecard/backend.dart
+++ b/app/lib/scorecard/backend.dart
@@ -56,9 +56,7 @@ class ScoreCardBackend {
     final cached = onlyCurrent
         ? null
         : await cache.scoreCardData(packageName, packageVersion).get();
-    if (cached != null &&
-        cached.hasReports(requiredReportTypes) &&
-        !versions.blacklistedRuntimeVersions.contains(cached.runtimeVersion)) {
+    if (cached != null && cached.hasReports(requiredReportTypes)) {
       return cached;
     }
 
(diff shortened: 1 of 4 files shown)


app/lib/shared/versions.dart
```diff
@@ -17,6 +17,18 @@ import 'utils.dart' show isNewer;
 /// a future dates are used.
 final RegExp runtimeVersionPattern = RegExp(r'^\d{4}\.\d{2}\.\d{2}$');
 
+/// The list (and priority order) of runtimeVersions where version-specific data
+/// is accepted from.
+///
+/// Make sure that at least two versions are kept here as the next candidates
+/// when the version switch happens.
+const acceptedRuntimeVersions = <String>[
+  '2020.05.06',
```
jonasfj, line 26: Can we give this a comment saying `// Current runtimeVersion`

app/lib/shared/versions.dart
```diff
@@ -17,6 +17,18 @@ import 'utils.dart' show isNewer;
...
+  '2020.05.03',
+  '2020.04.22',
+  '2020.04.07',
```
jonasfj, line 29: Maybe recommend that we don't let's this grow larger than 7 entries? or what do you think..

app/lib/scorecard/backend.dart
```diff
@@ -74,33 +72,26 @@ class ScoreCardBackend {
...
-    }
-
-    final latest = cards.fold<ScoreCard>(
-        cards.first,
-        (latest, current) => isNewer(
-                latest.semanticRuntimeVersion, current.semanticRuntimeVersion)
-            ? current
-            : latest);
-    return latest.toData();
+    final fallbackCards = await _db.lookup<ScoreCard>(fallbackKeys);
```
jonasfj, line 82: You so now you lookup all the fallback keys concurrently, I would suggest only looking up 5 keys at the time.. or maybe even 3..

If `fallbackRuntimeVersions` one day contains 50 entries, because we just didn't cleanup that seems problematic.. We can also fix this with a test that complains if `fallbackRuntimeVersions` has more than `N` entries :D

app/lib/shared/versions.dart
isoos: I'll write a test for it, limiting it to 5.

app/lib/scorecard/backend.dart
isoos: Why would have 50? Those versions are unlikely to be running, and we don't want their data to be around...

jonasfj: good point,

PR #3607, opened 2020-05-08
Block upload of non-pre-release package depending on pre-release SDK

### LINKED ISSUE #3557: Block upload of packages with stable versions and pre-release lower SDK constraint
In we https://github.com/dart-lang/pub-dev/issues/3375 discussed a number of [options](https://github.com/dart-lang/pub-dev/issues/3375#issuecomment-591326508) and decided to add warnings when uploading a stable version package with a pre-release SDK or package dependency. I'd like propose that we add a second check, and that we make this second check a hard error, not just warning:

Upload check: packages depending on a non-stable Dart SDK must themselves be a pre-release version. Not doing this is an error, and blocks the upload.

Specifically, if the lower SDK constraint references a non-stable (i.e. dev or beta) SDK, then package itself must use a [pre-release version](https://dart.dev/tools/pub/publishing#publishing-prereleases).

This ensures that you can clearly tell from a published package if it depends on a stable SDK or not, and that we encourage any feature use of non-stable SDK features to only end up as pre-release.

This *by-design* doesn't handle the case where a package in a stable version depends on a package which is in a pre-release version. It just ensures that any package with a non-stable SDK dep itself has proper versioning. Also, with this, a developer can run `pub deps -s compact` and see if they depend on any pre-release packages; if not, they know they don't transitively depend on a non-stable SDK.

isoos:
Fixes #3557.

### DIFF
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -157,6 +157,7 @@ Future<PackageSummary> summarizePackageArchive(String archivePath) async {
   issues.addAll(syntaxCheckUrl(pubspec.homepage, 'homepage'));
   issues.addAll(syntaxCheckUrl(pubspec.repository?.toString(), 'repository'));
   issues.addAll(forbidGitDependencies(pubspec));
+  issues.addAll(forbidPreReleaseSdk(pubspec));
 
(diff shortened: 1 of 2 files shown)


jonasfj:
Yeah, I don't see any objections, so let's ship it :D

PR #3609, opened 2020-05-11
Better error reporting on pubspec.yaml issues.

isoos:
- moved upload tests to a separate file (separate commit)
- added checks for bad yaml, and type problems in the pubspec
- fixed issue when lenient pubspec parsing failed


pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -88,9 +89,19 @@ Future<PackageSummary> summarizePackageArchive(String archivePath) async {
   Pubspec pubspec;
   try {
     pubspec = Pubspec.parse(pubspecContent);
+  } on YamlException catch (e) {
+    issues.add(ArchiveIssue('Error parsing pubspec.yaml: $e'));
+    return PackageSummary(issues: issues);
+  } catch (e) {
```
sigurdm, line 95: Let's not use catch-alls, they can easily cover up bugs.
If we expect other kinds either enumerate them, or at worst catch `Exception`.

isoos: There is no single exception in `pubspec_parse`, it is just a pass-through for others (either `yaml`, `checked_yaml` or `json_annotation`). The later has 2 top-level exceptions, but it is not clear what will happen if they add another one:
https://pub.dev/documentation/json_annotation/latest/json_annotation/json_annotation-library.html#exceptions

I'd stick with the generic `Exception` for now.

PR #3611, opened 2020-05-11
Expose e-mail sending issues to the user.

### LINKED ISSUE #3610: Don't return 200 OK if we fail to send invitation emails
Ideally SMTP `421` could be retried once. But if we give up we should say that we failed to send the email.

It might be fine to create the consent entity, etc. as we currently do, but we should surface a message to the user that sending the email failed.

isoos:
- Fixes #3610
- Added `SmtpClientCommunicationException` as a retryable exception.
- `SmtpMessageValidationException` indicates a possible user input issue.
- everything else is a generic "we have failed to send the mail"

### DIFF
--- a/app/lib/frontend/email_sender.dart
+++ b/app/lib/frontend/email_sender.dart
@@ -13,6 +13,7 @@ import 'package:retry/retry.dart';
 
 import '../service/secret/backend.dart';
 import '../shared/email.dart';
+import '../shared/exceptions.dart';
 
 final _logger = Logger('pub.email');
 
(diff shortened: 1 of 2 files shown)


app/lib/frontend/email_sender.dart
```diff
@@ -49,12 +50,18 @@ class EmailSender {
...
+          retryIf: (e) =>
+              e is TimeoutException ||
+              e is IOException ||
+              e is SmtpClientCommunicationException,
           delayFactor: Duration(seconds: 2),
           maxAttempts: 2,
         );
-      } catch (e, st) {
+      } on SmtpMessageValidationException catch (e) {
+        throw EmailSenderException.invalid(e.message);
```
jonasfj, line 61: Let's not expose raw messages from the SMTP server. I'm not convinced these messages can't contain something sensitive.

I suggest we give a generic reply like "Failed to send email, check that the email address is valid".

Also I think we want to log this.. it could be because our email template contains something that wasn't properly escaped, etc..

app/lib/frontend/email_sender.dart
```diff
@@ -49,12 +50,18 @@ class EmailSender {
...
+      } on MailerException catch (e, st) {
         _logger.severe('Sending email failed: $debugHeader.', e, st);
+        throw EmailSenderException.failed(e.message);
```
jonasfj, line 64: Again, let's be generic and not expose undelying technical issues (which may be internal errors) to the user.

PR #3617, opened 2020-05-12
Pre-release instead of dev on /versions page.

### LINKED ISSUE #3615: Don't show old prereleases at the top of /versions page
Once a version `>= 1.0.0` has been published, there is no reason to show this at the top:
dhBrPu5msCJ

We always show them at the bottom anyways.
Yq1QwTKMhYP

### LINKED ISSUE #3616: Use the terminology "prereleases" instead of "dev releases"
See also: https://dart.dev/tools/pub/publishing#publishing-prereleases

The documentation talks about prereleases, so does semantic versioning: https://semver.org/spec/v2.0.0-rc.1.html

Examples from `https://pub.dev/packages/path/versions`
Rz84CaQ9Pay
JqvnSByRuwD

isoos:
- Fixes #3615 (uses `Package.showDevVersion`)
- Fixes #3616 (uses `pre-release` and not `prerelease`).

### DIFF
--- a/app/lib/frontend/templates/package_versions.dart
+++ b/app/lib/frontend/templates/package_versions.dart
@@ -24,13 +24,15 @@ String renderPkgVersionsPage(
 
   final stableVersionRows = [];
   final devVersionRows = [];
-  PackageVersion latestDevVersion;
+  final latestDevVersion = versions.firstWhere(
+    (v) => v.version == data.package.latestDevVersion,
+    orElse: () => null,
+  );
   for (int i = 0; i < versions.length; i++) {
     final PackageVersion version = versions[i];
     final String url = versionDownloadUrls[i].toString();
     final rowHtml = renderVersionTableRow(version, url);
     if (version.semanticVersion.isPreRelease) {
-      latestDevVersion ??= version;
       devVersionRows.add(rowHtml);
     } else {
       stableVersionRows.add(rowHtml);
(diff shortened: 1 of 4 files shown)


app/lib/frontend/templates/package_versions.dart
```diff
@@ -24,23 +24,27 @@ String renderPkgVersionsPage(
...
   }
 
   final htmlBlocks = <String>[];
-  if (stableVersionRows.isNotEmpty && devVersionRows.isNotEmpty) {
+  if (stableVersionRows.isNotEmpty &&
+      devVersionRows.isNotEmpty &&
+      data.package.showDevVersion) {
     htmlBlocks.add(
-        '<p>The latest dev release was <a href="#dev">${latestDevVersion.version}</a> '
+        '<p>The latest pre-release was <a href="#dev">${latestDevVersion.version}</a> '
```
jonasfj, line 47: @kwalrath Do you have a position on "prereleases" vs. "pre-releases". I recall you pushing us to use `--prereleases` instead of `--pre-releases` for `pub outdated`. And your docs use "prereleases" everywhere.

But [semantic versioning](https://semver.org/) where the term comes from uses "pre-releases". I'm happy either way, but it's probably best we're internally consistent in Dart :)

app/lib/package/models.dart
```diff
@@ -94,6 +94,11 @@ class Package extends db.ExpandoModel {
   Version get latestDevSemanticVersion =>
       latestDevVersionKey == null ? null : Version.parse(latestDevVersion);
 
+  bool get showDevVersion {
```
jonasfj, line 97: It would also be nice to call this `showPrereleaseVersion` just to keep terminology consistent.

app/lib/frontend/templates/package_versions.dart
kwalrath: Yep, we use _prerelease_, no hyphen.

Here's how I decide on things like this. If it isn't in the [Google dev word list](https://developers.google.com/style/wordlist), I follow the advice of the [Google dev spelling guide](https://developers.google.com/style/spelling) and look it up on https://www.merriam-webster.com. https://www.merriam-webster.com/dictionary/prerelease lists the no-hyphen term first, so I'm going with that. Also, Google has a [preference not to hyphenate](https://developers.google.com/style/hyphens#when-to-hyphenate).

app/lib/package/models.dart
isoos: This opens up a few more questions:
- `showPrereleaseVersion` or `showPreReleaseVersion` (`package:semver` has `Version.isPreRelease`)
- Are we going to rename `latestDevSemanticVersion` and similar fields too?
- What about the Datastore field?

jonasfj: Fair enough, we can just leave this for a rainy day.

PR #3618, opened 2020-05-12
Add simple confirmation before deleting entries from Datastore.

isoos:


app/bin/tools/remove_package.dart
```diff
@@ -79,13 +80,13 @@ Future _deleteWithQuery<T>(Query query, {bool Function(T item) where}) async {
     if (shouldDelete) {
       deletes.add(m.key);
       if (deletes.length >= 500) {
-        await dbService.commit(deletes: deletes);
+        await _confirm(action: 'Deleting $T', deletes: deletes);
```
jonasfj, line 83: I think this is too many times we have to type `y`.. I was just thinking a single `y/n` before it started all the deletes.

```
$ dart remove_package.dart remove mypkg
Deleting package: mypkg
Versions: ....
Are you sure you want to do this (y/n).
```

Just to catch cases where you type something wrong on the command line... Similar to how you have to confirm when publishing a package. Reviewing every single transaction is probably too much, especially since we can't do it in a single transaction.

app/bin/tools/remove_package.dart
```diff
@@ -223,3 +244,35 @@ Future removePackageVersion(String packageName, String version) async {
...
+  }
+  print('');
+  print('**** About to commit the following action: $action');
+  inserts?.forEach((m) {
+    print('- insert/update: ${_debugKey(m.key)}');
+  });
+  deletes?.forEach((k) {
+    print('- delete: ${_debugKey(k)}');
+  });
+  print('Use CTRL+C to abort in 5 seconds...');
```
jonasfj, line 266: We have 7 calls to `_deleteWithQuery` so the script will now take `35s` just for waiting.

I'm actually okay trusting that it does the right thing. And not showing anything for each commit.. it was just the initial confirm I thought important :)

Issue #3503, opened 2020-04-07
Tagging of null-safety compliant packages on pub.dev


jonasfj:
A package is said to null-safety compliant if:
 * The package has opted-in by specifying a lower dart sdk bound `>= 2.10`
 * No libraries in the package opts out of the null-safety enabled language version.
 * All (non-dev) dependencies in the latest version resolvable by `pub` are fully null-safety-compliant by this definition.

This needs to be implemented in `pana` and surfaced on `pub.dev`. I propose we tag such packages `is:null-safe` (which we can then use as a search filter).

On the package page null-safety will be displayed with an icon or other cool gadget to be determined.

sigurdm:
Pana now does the analysis for tagging packages: https://github.com/dart-lang/pana/commit/e38880387f4a9e56e72d7cd1d27b3b13d2116107

isoos:
Duplicate of #3354

Issue #3354, opened 2020-02-20
Tag null-safe packages and highlight this on package page


jonasfj:
Exact Dart version that will be the lower bound is still TBD.

jonasfj:
Tag will probably be `is:null-safe` shown in UI as `Supports Null-Safety`.

mit-mit:
We should mark a package as null safe if, and only if, it satisfies all these properties (these assume we use 2.10 as the Dart versioning in which null safety is launched to stable):

  1. The SDK lower constraint is at least 2.10
  1. All direct dependencies of the package have a SDK lower constraint that is at least 2.10
  1. All indirect dependencies of the package have a SDK lower constraint that is at least 2.10

mit-mit:
Punting to beta 2.

isoos:
Duplicate of #4163. Keeping the newer one as it has a screenshot :)

PR #3626, opened 2020-05-14
Null-safety badge on package listing

isoos:
- #3503, #3519, #3354 
- this is only on listing page, not yet on package page
- no icon yet

Screenshot 2020-05-18 at 10 13 06

### DIFF
--- a/app/lib/frontend/templates/views/pkg/package_list_experimental.mustache
+++ b/app/lib/frontend/templates/views/pkg/package_list_experimental.mustache
@@ -33,21 +33,28 @@
     {{/is_external}}
     {{^is_external}}
     <p class="packages-metadata">
-      v <a href="{{& url}}">{{version}}</a>
-      {{#show_prerelease_version}} / <a href="{{& prerelease_version_url}}">{{prerelease_version}}</a>{{/show_prerelease_version}}
-      • Updated: <span>{{last_uploaded}}</span>
+      <span class="packages-metadata-block">
+        v <a href="{{& url}}">{{version}}</a>
+        {{#show_prerelease_version}} / <a href="{{& prerelease_version_url}}">{{prerelease_version}}</a>{{/show_prerelease_version}}
+        • Updated: <span>{{last_uploaded}}</span>
+      </span>
       {{#publisher_id}}
-      <img class="packages-vp-icon"
+      <span class="packages-metadata-block">
+        <img class="packages-vp-icon"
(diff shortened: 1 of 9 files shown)


app/lib/frontend/templates/views/pkg/package_list_experimental.mustache
```diff
@@ -33,21 +33,28 @@
...
-          <img class="package-badge-icon" src="{{static_assets.img__flutter-logo-32x32_png}}" />
-          Flutter Favorite
-        </span>
+      <span class="package-badge">
+        <img class="package-badge-icon" src="{{static_assets.img__flutter-logo-32x32_png}}" />
+        Flutter Favorite
+      </span>
       {{/is_flutter_favorite}}
+      {{#is_null_safe}}
+      <span class="package-badge">null-safe</span>
```
jonasfj, line 56: In the user interface the branding is "null-safety".

So we might have to say "null-safety compatible".

@mit-mit, do we have better terminology for use on the website?

(note. this is still behind experimental, we should keep it so, even when we ship the new design, as this will launch independently).

isoos: > (note. this is still behind experimental, we should keep it so, even when we ship the new design, as this will launch independently).

I guess we'll need another layer of experimental, otherwise this will slip.

mit-mit: What is the string used for? Is this the tool-tip?

isoos: @mit-mit: see the screenshot at the bottom of the description

mit-mit: I suggest the text of the badge is 'null safety' and that the tooltip is 'Supports the null safety language feature'

mit-mit: @isoos WDYT?

isoos: I've just now updated the PR with the text and an extra flag so that we don't necessarily launch it with the rest of the new UI. I'll update the screenshot soon.

isoos: Screenshot updated above

PR #3635, opened 2020-05-18
tools/remove_package use adminBackend

isoos:
- the backend code is almost the same as the one in tools
- updated small inconsistencies (and added TODO for fetching the full list of versions in a rare case)

### DIFF
--- a/app/bin/tools/remove_package.dart
+++ b/app/bin/tools/remove_package.dart
@@ -8,15 +8,13 @@ import 'dart:io';
 import 'package:gcloud/db.dart';
 import 'package:gcloud/storage.dart';
 import 'package:meta/meta.dart';
-import 'package:pub_dev/account/models.dart';
-import 'package:pub_semver/pub_semver.dart';
 
+import 'package:pub_dev/admin/backend.dart';
 import 'package:pub_dev/dartdoc/backend.dart';
 import 'package:pub_dev/package/backend.dart';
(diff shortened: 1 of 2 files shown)


app/lib/admin/backend.dart
```diff
@@ -293,8 +293,9 @@ class AdminBackend {
     List<Version> versionsNames;
     await withRetryTransaction(_db, (tx) async {
       final deletes = <Key>[];
-      final Key packageKey = _db.emptyKey.append(Package, id: packageName);
-      final package = (await tx.lookup([packageKey])).first as Package;
+      final packageKey = _db.emptyKey.append(Package, id: packageName);
+      final package =
+          await tx.lookupValue<Package>(packageKey, orElse: () => null);
```
jonasfj, line 298: nit. but I added `lookupOrNull` specifically for this scenario :D

(In particular because breaking into multiple lines like this annoyed me.. hehe)

app/bin/tools/remove_package.dart
```diff
@@ -43,7 +41,7 @@ Future main(List<String> arguments) async {
       await listPackage(package);
     } else if (command == 'remove') {
       if (version == null) {
-        await removePackage(package);
+        await _removePackage(package);
```
jonasfj, line 44: This changes semantics by creating a `ModeratedPackage` entity... which is probably fine :D

PR #3636, opened 2020-05-18
Refactoring search index code

isoos:
Mostly in preparation for reusing the search index in the default service, no functional code change.


app/lib/search/backend.dart
```diff
@@ -45,6 +45,21 @@ void registerSnapshotStorage(SnapshotStorage storage) =>
...
 
+/// The [PackageIndex] for Dart SDK API.
+PackageIndex get dartSdkIndex => ss.lookup(#_dartSdkIndex) as PackageIndex;
+
+/// Register a new [PackageIndex] for Dart SDK API.
+void registerDartSdkIndex(PackageIndex index) =>
+    ss.register(#_dartSdkIndex, index);
+
+/// The [PackageIndex] registered in the current service scope.
+PackageIndex get packageIndex =>
```
jonasfj, line 56: Can we explain that this is available everywhere?

jonasfj: I assume it is the `packageIndex` for use in case search service is failing, we can fallback to this. So it's available on the frontend service. (Am I correct?)

Issue #2936, opened 2019-10-22
Add support for banner ads/announcements to pub.dev


athomas:
dart.dev currently advertises the new publishers feature:
![image](https://user-images.githubusercontent.com/19993735/67286634-08407280-f4da-11e9-9768-1336d288d867.png)

It would be nice if we can advertise new pub features, or have other announcements (e.g. something that points you at an NNBD package migration guide) on pub.dev as well.

jonasfj:
Only on the frontpage, but yes it would be nice.

isoos:
> Only on the frontpage, but yes it would be nice.

dart.dev puts it on all (most) of the pages, and I think it makes sense if people are opening bookmarked pages. A few parameters that further influence the design:

- Would the user dismiss it and we'll never show it again? (would it be stored in a cookie or in the user session)

- Should it be part of the page (scroll out) or should it be sticky (on scroll out stays on top) or always-floating (displayed separately from the row)?

- Would it be always bound to a new release version (e.g. hard-coded in the app), or configured externally (e.g. HTML comes from Datastore)?

mit-mit:
For simplicity, I would make it on all pages, non-dismissable, but have it scroll out of view. And would be nice if changing it was a configuration option somehow and unrelated to deployments.

PR #3637, opened 2020-05-18
New announcement banner.

isoos:
- #2936
- follows the very simple design of dart.dev (+ thinner fonts from the new design + different link color as default)
- right now only per-deployment announcements are supported
- screenshots:
  
  
Screenshot 2020-05-18 at 17 29 26
Screenshot 2020-05-18 at 17 29 41

Screenshot 2020-05-18 at 17 28 47

Screenshot 2020-05-18 at 17 29 14

### DIFF
--- a/app/lib/service/announcement/backend.dart
+++ b/app/lib/service/announcement/backend.dart
@@ -1,0 +1,52 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:gcloud/service_scope.dart' as ss;
+
+/// Sets the announcement backend service.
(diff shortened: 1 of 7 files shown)


app/lib/service/announcement/backend.dart
```diff
@@ -0,0 +1,52 @@
...
+    ss.lookup(#_announcementBackend) as AnnouncementBackend;
+
+/// Represents the backend for the announcement handling and related utilities.
+class AnnouncementBackend {
+  String _announcementHtml;
+  Timer _updateTimer;
+
+  /// Loads the active announcement from Datastore.
+  Future<void> update() async {
+    // TODO: implement loading announcements from Datastore.
```
jonasfj, line 24: I like this... then we can put a warning on our staging site :D

mit-mit:
Overall looks great. But I'm not sure I like the background color. Let's ask Jeremy for design input.

PR #3643, opened 2020-05-19
Fix back-to-search on package page.

isoos:


pkg/web_app/lib/src/search.dart
```diff
@@ -15,7 +15,8 @@ void setupSearch() {
 }
 
 final _acceptedSearchPaths = <Pattern>[
-  '/packages',
+  RegExp(r'^/packages$'),
+  RegExp(r'^/packages\?'),
   '/dart/packages',
   '/flutter/packages',
```
jonasfj, line 21:
```dart
RegExp(r'^/(:?dart/|flutter/)?packages(?:\?|$)')
```
Or
```dart
RegExp(r'^/packages(?:\?|$)')
RegExp(r'^/dart/packages(?:\?|$)')
RegExp(r'^/flutter/packages(?:\?|$)')
```

So we're being as narrow as reasonably possible with the `/dart/packages` and `/flutter/packages` urls too.. I assume these have to be the first part of the thing..

jonasfj: nit: also `https://pub.dev/flutter/favorites` is a list page.

PR #3647, opened 2020-05-22
Updating announcement via Secret entities.

### LINKED ISSUE #2936: Add support for banner ads/announcements to pub.dev

isoos:
Fixes #2936.

### DIFF
--- a/app/lib/service/announcement/backend.dart
+++ b/app/lib/service/announcement/backend.dart
@@ -6,6 +6,8 @@ import 'dart:async';
 
 import 'package:gcloud/service_scope.dart' as ss;
 
+import '../secret/backend.dart';
+
 /// Sets the announcement backend service.
 void registerAnnouncementBackend(AnnouncementBackend backend) =>
     ss.register(#_announcementBackend, backend);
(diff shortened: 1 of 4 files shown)


app/lib/service/announcement/backend.dart
```diff
@@ -21,7 +23,10 @@ class AnnouncementBackend {
 
   /// Loads the active announcement from Datastore.
   Future<void> update() async {
-    // TODO: implement loading announcements from Datastore.
+    final value = await secretBackend.lookup(SecretKey.announcement);
+    if (value != null && value.trim().isNotEmpty) {
+      _announcementHtml = value.trim();
```
jonasfj, line 28: Can we run this through sanitize_html? I assume it's going to use `<b>` and `</a>` and maybe `<br>`... Just thinking it's better to have some defense in depth, whenever we inject HTML, if we can sanitize it we should.

PR #3649, opened 2020-05-22
Include 'created' timestamp in versions list API.

### LINKED ISSUE #3565: [Feature Request] Add upload date info to public API
I am making something similar to #2339, but it's a CLI instead of a plugin. One of the functionalities will be to state what is the latest version and when it has been released.

I have the option to scrap the page `https://pub.dev/packages/{package}/versions`, but I think it would be nice to have this info on the API instead.

isoos:
Fixes #3565.

### DIFF
--- a/pkg/client_data/lib/package_api.dart
+++ b/pkg/client_data/lib/package_api.dart
@@ -133,13 +133,13 @@ class VersionInfo {
   final String archiveUrl;
 
   /// This is an optional field of the API response, it may be `null` or omitted.
-  final DateTime created;
+  final DateTime published;
 
   VersionInfo({
     @required this.version,
     @required this.pubspec,
     @required this.archiveUrl,
-    @required this.created,
+    @required this.published,
   });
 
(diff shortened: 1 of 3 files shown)


pkg/client_data/lib/package_api.dart
```diff
@@ -132,10 +132,14 @@ class VersionInfo {
   @JsonKey(name: 'archive_url')
   final String archiveUrl;
 
+  /// This is an optional field of the API response, it may be `null` or omitted.
+  final DateTime created;
```
jonasfj, line 136: Should we consider using the word `published` instead?

@sigurdm, thoughts?

(I think API changes are worth bikeshedding a little, because they are super hard to reverse).

sigurdm: I like `published`

jonasfj: or "uploaded"...

jonasfj: Okay, let's go with `published`.

Issue #3450, opened 2020-03-23
Investigate memory usage of front-end


sigurdm:

isoos:
I've done a load test with a local instance connecting to staging. After ~2 hours and 12k+ requests served (random API calls, package pages and searches), I haven't observed an increase in the memory consumption of the VM.

Maybe we should try to close and re-connect the RPC connections (to redis, search service and Datastore)?

isoos:
I've done further investigation into memory use, focusing on `StringInternPool`, which is used both in `search` service and in the frontend's dependency tracker isolate. In both cases I've simulated the long-term use by running the initialization phase multiple times (5-10x), and measured the difference in both currently used and maximum heap memory.

- `search` used it with a limit on how many items are in the pool. There was little to no benefit of using the `StringInternPool`, maybe 1-3 MB difference while the app+index overall took ~900MB.

- Dependency tracker used it without a limit on the maximum numbers, however, it only stored the name of the packages (+ a variant of that). Using the intern pool showed benefits (~40 MB), but to my surprise, rewriting small bits in the tracker (and replacing `StringInternPool`) reduced further 25 MB to the memory used.

Because of the above, I'm going to remove the `StringInternPool`, and use the updated code for dependency tracker.

I have seen no signs of further memory buildup with or without the interning when I've switched from 5 updates to 10.

PR #3652, opened 2020-05-25
Preemptively reopen redis connections every hour.

isoos:
- Maybe related to the frontend memory issue, maybe not. #3450
- Implemented it with a timer, as it won't be blocking on the running services.

### DIFF
--- a/app/lib/shared/redis_cache.dart
+++ b/app/lib/shared/redis_cache.dart
@@ -9,6 +9,7 @@ import 'dart:io';
 import 'package:appengine/appengine.dart';
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:logging/logging.dart';
+import 'package:neat_cache/cache_provider.dart';
 import 'package:neat_cache/neat_cache.dart';
 
 import '../account/models.dart' show LikeData, UserSessionData;
@@ -207,7 +208,8 @@ Future _withRedisCache(FutureOr Function() fn) async {
   }
 
   // Create and register a cache
-  final cacheProvider = Cache.redisCacheProvider(connectionString);
+  final cacheProvider = await _UpdatingCacheProvider.connect(
+      () async => Cache.redisCacheProvider(connectionString));
   _registerCache(CachePatterns._(Cache(cacheProvider)));
 
   try {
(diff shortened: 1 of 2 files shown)


app/lib/shared/redis_cache.dart
```diff
@@ -223,3 +225,57 @@ Future _withInmemoryCache(FutureOr Function() fn) async {
...
+          // discard if `close` was called during the Timer callback
+          await newProvider.close();
+        } else {
+          // otherwise update provider and close old one
+          final d = _delegate;
+          _delegate = newProvider;
+          await d.close();
+        }
+      } catch (e, st) {
+        _log.warning('Failed to update redis connection.', e, st);
```
jonasfj, line 255: Make this a `shout` so it goes into our error reporting? Unless we know it's harmless we can later reduce the noise level.

app/lib/shared/redis_cache.dart
```diff
@@ -223,3 +225,57 @@ Future _withInmemoryCache(FutureOr Function() fn) async {
   _registerCache(CachePatterns._(Cache(Cache.inMemoryCacheProvider(4096))));
   return await fn();
 }
+
+typedef CacheProviderFn<T> = Future<CacheProvider<T>> Function();
+
+class _UpdatingCacheProvider<T> implements CacheProvider<T> {
```
jonasfj, line 231: Can we name this better? `_PeriodicallyReconnectingCacheProvider` or `ConnectionRefreshingCacheProvider`

app/lib/shared/redis_cache.dart
```diff
@@ -223,3 +225,57 @@ Future _withInmemoryCache(FutureOr Function() fn) async {
...
+  final CacheProviderFn<T> _fn;
+  CacheProvider<T> _delegate;
+  Timer _timer;
+  bool _isClosed = false;
+
+  _UpdatingCacheProvider._(this._fn, this._delegate) {
+    _timer = Timer.periodic(Duration(hours: 1), (_) async {
```
jonasfj, line 238: nit, maybe make `Duration` a parameters... and then add a little jitter, like:
`Duration(minutes: 55 + Random.secure().nextInt(10));`

app/lib/shared/redis_cache.dart
jonasfj: Also a one or two line documentation comment wouldn't hurt...

PR #3656, opened 2020-05-26
Add test case that we are using the latest flutter version

jonasfj:


app/test/shared/versions_test.dart
```diff
@@ -90,6 +90,26 @@ void main() {
         isTrue);
   });
 
+  test(
+    'Flutter is using the latest stable',
+    () async {
+      final flutterArchive = await fetchFlutterArchive();
+      final currentStable = flutterArchive.releases.firstWhere(
+        (r) => r.hash == flutterArchive.currentRelease.stable,
```
isoos, line 98: I think this makes the `'Flutter is using a version from the stable channel.'` test obsolete. Could you remove that?

jonasfj: I'm not sure... this is a test case we could easily choose to skip... if we somehow fall behind.

PR #3663, opened 2020-05-27
Fix Bucket.list(prefix:) calls with ending slashes.

isoos:
- hard argument checking to use an ending slash in all of our prefixes
- updated `fake_gcloud` to use the hopefully correct behavior
- added a documentation page check that would have caught the issue at post-deployment checks (but it is not yet automated)


isoos:
Staging looks good so far with the fix, merging.

PR #3674, opened 2020-05-29
Fix Bucket.list + adjusting recursive deleting the folder.

isoos:


app/lib/shared/storage.dart
```diff
@@ -31,13 +32,70 @@ Future<Bucket> getOrCreateBucket(Storage storage, String name) async {
...
+  int concurrency,
+}) async {
+  if (!folder.endsWith('/')) {
+    throw ArgumentError('Folder path must end with `/`: "$folder"');
+  }
+  final deleter = _ObjectDeleter(bucket, concurrency);
+  final folders = <String>[folder];
+  while (folders.isNotEmpty) {
+    final currentFolder = folders.removeLast();
+    await for (final obj in bucket.list(prefix: currentFolder)) {
```
jonasfj, line 66: Hmm, wouldn't it be better to implement something on `Bucket` that supports us in doing this?
Maybe `Bucket.list(prefix: ..., delimiter: '')`

Basically just implement a feature that let's users pass in a delimiter, and then make the empty delimiter special (so that it means "no delimiter")

isoos:
> In fact, we should probably revert the fix in gcloud, add the delimiter feature and make it a breaking change.

I think the fix in `gcloud` was correct but breaking. Adding the delimiter at this point is an incremental improvement, and unlikely to change how we will be using the API.

app/lib/shared/storage.dart
isoos: Maybe a `Bucket.list` with `recursive: true` would help. I'd find it confusing to set the delimiter to empty in order to get recursive results.

PR #3694, opened 2020-06-08
Limit direct dependencies.

### LINKED ISSUE #3693: Limit number of direct package dependencies to 100
Limit number of direct package dependencies to 100, to to have some sanity limit.

Based on current packages I suggest we limit number of direct-dependencies a package can have to 100. If people really need they can:
*  Work around the problem with transitive dependencies, or,
*  File a bug to have the limit increased.
(I suggest we let this be reflected in the error reported to the user.)

There is no particular reason we need a hard limit at 100. But having _some_ limit is always nice.
We do analyze the package dependency relationships and if we let packages have an unbounded number of dependencies this may become painful.

Statistics from current packages: (using data from public API)
```
No package/version has more than mgpcli 0.1.74 with 57 dependencies

## 1123 has dependencies >= 5 (14645 versions)
## 245 has dependencies >= 10 (4397 versions)
## 80 has dependencies >= 15 (1844 versions)
## 11 has dependencies >= 25 (293 versions)
## 5 has dependencies >= 30 (159 versions)
 - build_runner has 44 versions with >= 30 dependencies
 - magpie_cli has 6 versions with >= 30 dependencies
 - mgpcli has 17 versions with >= 30 dependencies
 - test has 32 versions with >= 30 dependencies
 - z_components has 60 versions with >= 30 dependencies
## 3 has dependencies >= 35 (65 versions)
 - magpie_cli has 6 versions with >= 35 dependencies
 - mgpcli has 17 versions with >= 35 dependencies
 - z_components has 42 versions with >= 35 dependencies
## 2 has dependencies >= 40 (23 versions)
 - magpie_cli has 6 versions with >= 40 dependencies
 - mgpcli has 17 versions with >= 40 dependencies
## 1 has dependencies >= 55 (17 versions)
 - mgpcli has 17 versions with >= 55 dependencies
```

isoos:
Fixes #3693.

### DIFF
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -262,7 +262,8 @@ Iterable<ArchiveIssue> syntaxCheckUrl(String url, String name) sync* {
 ///
 /// It ignores `dev_dependencies` as these are for development only.
 Iterable<ArchiveIssue> validateDependencies(Pubspec pubspec) sync* {
-  if (pubspec.dependencies.length > 128) {
+  // This is not an inherently hard limit, it's merely a sanity limitation.
+  if (pubspec.dependencies.length > 100) {
     yield ArchiveIssue('Package must not exceed 128 direct dependencies. '
         '(Please file an issue if you think you have a good reason for more dependencies.)');
   }
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -169,6 +169,7 @@ Future<PackageSummary> summarizePackageArchive(String archivePath) async {
   issues.addAll(validatePackageVersion(pubspec.version));
   issues.addAll(syntaxCheckUrl(pubspec.homepage, 'homepage'));
   issues.addAll(syntaxCheckUrl(pubspec.repository?.toString(), 'repository'));
+  issues.addAll(validateDependencies(pubspec));
   issues.addAll(forbidGitDependencies(pubspec));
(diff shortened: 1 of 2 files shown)


pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -257,6 +258,16 @@ Iterable<ArchiveIssue> syntaxCheckUrl(String url, String name) sync* {
   }
 }
 
+/// Validate that the package does not have too many dependencies.
+///
+/// It ignores `dev_dependencies` as these are for development only.
+Iterable<ArchiveIssue> validateDependencies(Pubspec pubspec) sync* {
+  if (pubspec.dependencies.length > 128) {
+    yield ArchiveIssue('Package must not exceed 128 direct dependencies. '
+        '(Please file an issue if you think you have a good reason for more dependencies.)');
```
jonasfj, line 267: I just updated https://github.com/dart-lang/pub-dev/issues/3693 to say 100, can we do that?
Otherwise, this seems solid :D

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -257,6 +258,16 @@ Iterable<ArchiveIssue> syntaxCheckUrl(String url, String name) sync* {
   }
 }
 
+/// Validate that the package does not have too many dependencies.
+///
+/// It ignores `dev_dependencies` as these are for development only.
+Iterable<ArchiveIssue> validateDependencies(Pubspec pubspec) sync* {
+  if (pubspec.dependencies.length > 128) {
```
jonasfj, line 265: Can we add a comment:
```
// This is not an inherently hard limit, it's merely a sanity limitation.
```

PR #3713, opened 2020-06-12
Skeleton rendering of the new pana Report.

isoos:
Screenshot 2020-06-15 at 14 00 29


sigurdm:
Looks weird that the biggest headlines are indented more than the smaller text.

Maybe we need some kind of indication that sections can be "unfolded"

Besides that: LGTM!

isoos:
> Looks weird that the biggest headlines are indented more than the smaller text.
> Maybe we need some kind of indication that sections can be "unfolded"

Yes, and yes. It is not production-ready yet, I just wanted to push this part out, as the iteration on the icons, indent and colors can happen separately....

isoos:
PTAL. This now has the icons and the color model of the design (black label if OK, red if not max score).

sigurdm:
LGTM!

PR #3719, opened 2020-06-17
Add admin users we can impersonate for doing moderation operations

jonasfj:


jonasfj:
@sigurdm, the plan is to do moderation through the admin API, and authenticate by impersonating this service account.

Hence, we can manage permissions using this service account.

PR #3720, opened 2020-06-17
Use likeCount as part of the default search rank.

isoos:
- tweaked the score calculation: same like counts should get the same score (otherwise it would be somewhat random in tests and unfair in prod)
- the reordering of `http` and `async` in one of the tests validates that the like count is part of the results

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -621,6 +621,8 @@ class _LikeTracker {
     _changed |= removed != null;
   }
 
+  /// Updates `_LikeScore.score` values, setting them between 0.0 (no likes) to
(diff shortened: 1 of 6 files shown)


app/lib/search/mem_index.dart
```diff
@@ -581,3 +588,48 @@ class _PackageNameIndex {
...
+      v.likeCount = likeCount;
+    }
+  }
+
+  void removePackage(String package) {
+    final removed = _values.remove(package);
+    _changed |= removed != null;
+  }
+
+  void _updateScores() {
```
jonasfj, line 624: One line doc-comment explaining what this does wouldn't work... it updates `score` to be a ranking from 1 to 0.

PR #3722, opened 2020-06-18
Bump runtimeVersion

isoos:
- Storing the `dartdocEntryUuid` in the scorecard report takes effect only after the content is generated, forcing the `runtimeVersion` change will force the field to be updated.


isoos:
> if we get a new pana release today

If we don't do the prod release, we can just keep the runtimeVersion if there is also a `pana` release today.

Issue #3726, opened 2020-06-15
Require flutter ^1.20.0 if ios/ is not present for a plugin


jonasfj:
We should require in `pub publish` that, if:
 * `flutter.plugin.platforms != null` (ie. it's a plugin), and,
 * the `ios/` folder is not present.

Then the latest allowed Flutter SDK version must be 1.20.0 in `environment.flutter`.

/cc @cyanglaz,

PR #3736, opened 2020-06-22
Require ios/ folder or flutter 1.20.0 for plugins

sigurdm:
Fixes: https://github.com/dart-lang/pub-dev/issues/3726


pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -327,6 +328,9 @@ Iterable<ArchiveIssue> forbidPreReleaseSdk(Pubspec pubspec) sync* {
   }
 }
 
+const _pluginDocsUrl =
```
isoos, line 331: Let's move this into `shared/urls.dart`.

sigurdm: Would that not make pub_package_reader depend on app?

isoos: Ah, true, never mind.

PR #3737, opened 2020-06-22
Skip tasks when the search snapshot has a newer document.

isoos:


app/lib/search/updater.dart
```diff
@@ -148,6 +148,14 @@ class IndexUpdater implements TaskRunner {
       final requireAnalysis =
           sd != null && now.difference(sd.timestamp).inDays < 7;
 
+      // Skip tasks that originate before the current document in the snapshot
```
sigurdm, line 151: Tiny nit: this could be calculated before `requireAnalysis`

PR #3739, opened 2020-06-23
Reduce concurrent snapshot writes.

isoos:
Writing a snapshot to the storage bucket may take about 10-15 seconds (serialization + upload latency). The current code does not have any randomization or existing file check, and as a result, search instances trying to write the snapshot will try to write it around the same time (initially at least, later it diverges), causing unresponsive windows around the same times.

This PR uses a randomized timer to pace out the events, and also has a check for recency: only one snapshot will be written in every 24 hours, making it unlikely that two running instances will be occupied with the snapshot write at the same time.

### DIFF
--- a/app/lib/search/updater.dart
+++ b/app/lib/search/updater.dart
@@ -178,6 +178,7 @@ class IndexUpdater implements TaskRunner {
   }
 
   Future<void> _updateSnapshotIfNeeded() async {
+    // TODO: make the catch-all block narrower
     try {
       if (await snapshotStorage.wasUpdatedRecently()) {
         _logger.info('Snapshot update skipped (found recent snapshot).');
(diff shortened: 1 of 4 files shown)


app/lib/search/updater.dart
```diff
@@ -159,20 +166,19 @@ class IndexUpdater implements TaskRunner {
...
+      if (await snapshotStorage.wasUpdatedRecently()) {
+        _logger.info('Snapshot update skipped (found recent snapshot).');
+      } else {
         _logger.info('Updating search snapshot...');
         await snapshotStorage.store(_snapshot);
         _logger.info('Search snapshot update completed.');
-      } catch (e, st) {
-        _logger.warning('Unable to update search snapshot.', e, st);
       }
+    } catch (e, st) {
```
sigurdm, line 180: Nit: Let's put the catch-all inside the else-branch to make it as narrow as possible.

isoos: I'd like to keep it there for now, otherwise there is a small chance that the storage bucket returns 5xx on the recency check, and it crashes the search instance. I'm adding a TODO to make it narrower.

PR #3743, opened 2020-06-23
Async update of like scores

isoos:


app/lib/search/mem_index.dart
```diff
@@ -621,17 +629,42 @@ class _LikeTracker {
...
   /// 1.0 (most likes).
-  void _updateScores() {
+  Future<void> _updateScores() async {
+    final sw = Stopwatch()..start();
     final entries = _values.values.toList();
+
+    // The method could be a single sync block, however, while the index update
+    // happens, we are not serving queries. With the forced async segments,
+    // the waiting queries will be served earlier.
+    await Future.delayed(Duration.zero);
```
sigurdm, line 655: I think `await null` should do the trick...

isoos: I always hoped that `await null` will get eventually optimized by the VM. (`Duration.zero` may have the same future...)

isoos: I'll leave it this way, as other parts of `mem_index.dart` are using the same pattern. We can re-visit in a cleanup.

PR #3746, opened 2020-06-25
pub.dev section in Help menu

isoos:
Screenshot 2020-06-25 at 15 39 02
Screenshot 2020-06-25 at 15 39 21


mit-mit:
Love it!

Speaking of Help, I think we can remove the 'Pub tool' entry from the two other lists. That is already covered in the 'Using packages' ones.

mit-mit:
I also wonder if 'Web and Server' should be 'Dart'? I mean, there are also CLI apps.

mit-mit:
lgtm

PR #3749, opened 2020-06-26
Enable Datastore-based dartdoc entry uuid only for the non-concurrent cases.

isoos:


app/lib/frontend/handlers/custom_api.dart
```diff
@@ -48,6 +48,9 @@ Future<shelf.Response> apiDocumentationHandler(
     }
   }
 
+  // TODO: to use `useLastReportedEntry`, the concurrent calls here need
+  //       client-side rate-limit or batched lookups, otherwise they will get
+  //       rate-limited/rejected by Datastore.
```
jonasfj, line 53: So `useLastReportedEntry` only exists for this purpose... and really, using GCS listing instead of datastore lookups isn't any better... I suspect it'll fail just as much -- only maybe it'll fail in a different way.. maybe it's easier for this to regress into something slow, rather than complain about rate-limits.

IMO it would be better to disable the feature of showing docs status on the versions page. Then we can rebuild the feature once we have a non-degenerate source of data for it. IMO, building it in a manner that requires client side requests is a clear indication that we do not have an efficient way to query, hence, queries will be de-generate (all the client side JS request does is mask the issue and change the failure scenario).

In the sandboxing world I'm aiming for a single entity that knows all job information about a single package. Ideally, that'll do we when we get there. We could also in the meantime:
 * Go with this hack,
 * Disable the docs-status feature on versions page,
 * Limit the docs-status feature on the versions page to the 5 most recent versions (implementing the lookups at rendering time, instead of using JS).

PR #3769, opened 2020-07-03
Fix foldable: enable content growth after activated.

isoos:


jonasfj:
This is a pretty neat fix :D

PR #3783, opened 2020-07-07
Fix code blocks horizontal scrolling

isoos:
- Moving padding from outer `pre` to inside `code` block.
- Folding block needed a maximum width setting, otherwise flex-grow took priority.

### DIFF
--- a/pkg/web_css/lib/src/_base.scss
+++ b/pkg/web_css/lib/src/_base.scss
@@ -158,13 +158,15 @@ code {
 .markdown-body pre,
 pre {
   background: #f5f5f7;
-  padding: 30px;
+  line-height: 100%;
+  padding: 0px;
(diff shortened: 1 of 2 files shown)


pkg/web_css/lib/src/_base.scss
```diff
@@ -158,13 +158,15 @@ code {
...
 pre {
   background: #f5f5f7;
-  padding: 30px;
+  line-height: 100%;
+  padding: 0px;
   overflow-x: auto;
 
   > code {
-    padding: 0 !important;
+    padding: 30px !important;
```
jonasfj, line 166: isn't this a bit much?

isoos: ¯\_(ツ)_/¯

This was the padding from the design.

PR #3790, opened 2020-07-07
Use only Datastore-based DartdocEntry for all user-facing queries.

isoos:
(Note: this is planned to be merged only after `2020.06.10` is no longer an accepted `runtimeVersion`)


app/lib/dartdoc/backend.dart
```diff
@@ -168,44 +168,55 @@ class DartdocBackend {
...
+  Future<List<DartdocEntry>> getEntriesForVersions(
+      String package, List<String> versions) async {
+    final entries = <String, DartdocEntry>{};
+
+    for (final v in versions) {
+      final cachedEntry = await cache.dartdocEntry(package, v).get();
+      if (cachedEntry != null) {
+        entries[v] = cachedEntry;
+      }
+    }
```
jonasfj, line 184: Why not do this concurrently?

```dart
final pool = new Pool(10);
await Future.wait(pool.withResource(() => versions.map((v) async {
  final cachedEntry = await cache.dartdocEntry(package, v).get();
  if (cachedEntry != null) {
    entries[v] = cachedEntry;
  }
}));
```

app/lib/dartdoc/backend.dart
```diff
@@ -168,44 +168,55 @@ class DartdocBackend {
...
+        reportType: ReportType.dartdoc,
+        runtimeVersion: rv,
+      );
+      for (var i = 0; i < queryVersions.length; i++) {
+        final r = reports[i];
+        if (r != null) {
+          final version = queryVersions[i];
+          final entry = (r as DartdocReport).dartdocEntry;
+          entries[version] = entry;
+          await cache.dartdocEntry(package, version).set(entry);
```
jonasfj, line 203: Technically, there is no need to await this here... I would suggest using the same pool as before.
```dart
scheduleMicrotask(() async {
  await pool.withResource(() async {
    await cache.dartdocEntry(package, version).set(entry);
  });
});
```

app/lib/scorecard/backend.dart
```diff
@@ -150,6 +151,21 @@ class ScoreCardBackend {
...
+    String packageName,
+    Iterable<String> versions, {
+    @required String reportType,
+    String runtimeVersion,
+  }) async {
+    final keys = versions
+        .map((v) => scoreCardKey(packageName, v, runtimeVersion: runtimeVersion)
+            .append(ScoreCardReport, id: reportType))
+        .toList();
+    final items = await _db.lookup<ScoreCardReport>(keys);
```
jonasfj, line 165: Throw `ArgumentError` if `versions.length > 1000`, see: https://cloud.google.com/datastore/docs/concepts/limits

Or handle by making two requests..

isoos:
Found the culprit: https://github.com/dart-lang/pub-dev/pull/3826

I'd postpone merging this until that fix gets into production.

isoos:
I've verified this on staging (both correctness and performance).

PR #3792, opened 2020-07-08
Slightly better like button

isoos:
Screenshot 2020-07-08 at 8 52 20


isoos:
> I still think we should make it work like:

Agreed, filed issue to track it:
https://github.com/dart-lang/pub-dev/issues/3794

PR #3821, opened 2020-07-13
Use yellow/orange on non-max|non-min scores.

isoos:
Screenshot 2020-07-13 at 15 33 36


isoos:
double-checking with @mit-mit

isoos:
For the record: this restores the colored labels for non-max scores that was changed in https://github.com/dart-lang/pub-dev/pull/3772

PR #3888, opened 2020-07-27
Handle long query texts better.

isoos:
Screenshot 2020-07-27 at 15 19 08

### DIFF
--- a/app/lib/frontend/templates/views/shared/listing_info.mustache
+++ b/app/lib/frontend/templates/views/shared/listing_info.mustache
@@ -15,6 +15,10 @@
     {{#has_owned_by}}
     owned by <code>{{ owned_by }}</code>
     {{/has_owned_by}}
+
+    {{#has_message_from_backend}}
+      {{& message_from_backend_html }}
+    {{/has_message_from_backend}}
   </div>
   {{& sort_control_html }}
 </div>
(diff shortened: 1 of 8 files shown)


app/lib/search/search_client.dart
```diff
@@ -62,6 +62,13 @@ class SearchClient {
       return result;
     }
 
+    // Block search on unreasonably long search queries (when the free-form
+    // text part is longer than one would enter via the search input field).
+    final queryLength = query?.parsedQuery?.text?.length ?? 0;
+    if (queryLength > 256) {
```
sigurdm, line 68: Can we extract and name this magic constant?

app/test/frontend/golden/pkg_index_page.html
```diff
@@ -179,6 +179,7 @@ <h3 class="foldable-button">Dart
     packages
 
 
+
```
sigurdm, line 182: Is this extra newline intentional?

isoos: That is coming from the the mustache template. We could optimize for nicer outputs but then the source templates would be less readable.

isoos: And yes, it is expected from the template change (but not intentional).

app/lib/frontend/templates/views/shared/listing_info.mustache
```diff
@@ -15,6 +15,10 @@
     {{#has_owned_by}}
     owned by <code>{{ owned_by }}</code>
     {{/has_owned_by}}
+
```
sigurdm, line 18: Could you not move the newline into the conditional>?

Issue #3897, opened 2020-07-13
Pre-release SDK constraints should not require a pre-release version


grouma:
The following error should only be a warning:

`Packages with an SDK constraint on a pre-release of the Dart SDK should themselves be published as a pre-release version. `

See this specific example: https://github.com/dart-lang/build/pull/2753

athomas:
@sigurdm and @jonasfj are OOO.

Given that these issues seem connected to the stable release that is slated to go out on Monday, I have asked @isoos and @sortie to prepare for deploying a revert.

leafpetersen:
Thanks @athomas !  @jakemac53 are you unblocked from landing these fixes now?

sortie:
I am currently deploying them to production. It should be fully up in about an hour.

sortie:
The traffic has been switched on pub.

sigurdm:
I guess this means we have to implement the warning in the client?

isoos:
> I guess this means we have to implement the warning in the client?

We could also keep the rule and have a case-by-case opt out.

mit-mit:
I browsed though this thread, and the linked bugs, but I still don't understand why we need a change. 

I still believe this needs to be *an error*: A package which depends on a pre-release package *is* itself pre-release and this should be versioned as such.

What is the canonical use case where this is intractable?

mit-mit:
> Consider also the following scenario for any package depended on by flutter:

> 1. The SDK releases some new feature that a package wants to use.

This is in a non-stable SDK, right?

> 2. The package starts using it, and has to stay at a pre-release version.

Yes, it definitely should. It's using a feature which has no stability promise and could be removed at any point in time.

> 3. We start preparing the stable flutter/dart sdks, and want to update to the latest version of said package in flutter.

> We cannot publish a stable version of the package because it depends on a dev SDK, and the stable SDK is not yet released

Yup, sounds correct.

> At the same time flutter stable should really not be pinning dev releases of packages (null safety being the exception right now, but not likely a strategy we want to repeat)
> How do we resolve this scenario? Do you have to wait two stable releases before using any new sdk feature in a package depended on by flutter?

Why can't we publish the stable version of the package that depends on the about-to-be-stable SDK? It would not be resolvable until the stable SDK ships, but that's the behavoir we want, isn't it?

jakemac53:
[...]

PR #3898, opened 2020-07-31
Temporary rollback of prerelease SDK check on upload

isoos:


athomas:
Note: this is related to #3897.

Issue #2226, opened 2019-04-24
display license file for packages where we can't link


jonasfj:
For packages where we can't detect the license, we should consider showing something scary like.
```
License:
**Custom** (LICENSE)
```
And in cases where we can't link to the license file on github, it's probably better to simply have a "license" tab for hosting the license...

In fact, we could display the license from the package on `https://pub.dartlang.org/packages/<package>/license` and just link to that in the `(LICENSE)` link..

IMO, it's important that we highlight bad licenses, and that we refer to the exact license file the package was published with. Just rendering as a `<pre>` sounds reasonable to me..

-----
This was previously discussed in #201, but I just noticed some packages with very obscure licenses and this was not obvious from looking at the package page.

jonasfj:
@mit-mit, any concerns?

mit-mit:
What's an example of a package with an obscure license? Just want to take a look at how we currently show it.

jonasfj:
I'll forward an email with an example to you..

isoos:
This has been implemented.

PR #3908, opened 2020-08-04
Update LICENSE file handling

### LINKED ISSUE #3449: Pub allows empty license file
If LICENSE doesn't exist pub shows
```
Missing requirements:
* You must have a COPYING, LICENSE or UNLICENSE file in the root directory.
```
and it treats empty LICENSE file as a valid license
If the intent is to prevent unlicensed packages to be published – this should be also an error

Validating license is something hard to implement, b.c. there are no strict rules, so maybe it makes sense to drop LICENSE missing requirement and suggest to generate MIT/BSD license instead?

isoos:
- Empty license file is no longer accepted. fixes #3449 - although there would be more patterns that we can filter on, let's start with the trivial one.
- Storing license file as an asset, to eventually display it as a tab (#2226).

### DIFF
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -442,7 +442,7 @@ Iterable<ArchiveIssue> requireNonEmptyLicense(
     yield ArchiveIssue('LICENSE file `$path` has no content.');
     return;
   }
-  if (content.contains('TODO: Add your license here.')) {
+  if (content.toLowerCase().contains('todo: add your license here.')) {
     yield ArchiveIssue('LICENSE file `$path` contains generic TODO.');
     return;
   }
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -37,6 +37,8 @@ class PackageSummary {
   final String changelogContent;
   final String examplePath;
   final String exampleContent;
+  final String licensePath;
+  final String licenseContent;
   final List<String> libraries;
 
   PackageSummary({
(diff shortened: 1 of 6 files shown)


pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -418,3 +431,19 @@ Iterable<ArchiveIssue> requireIosFolderOrFlutter2_20(
...
+    String path, String content) sync* {
+  if (path == null) {
+    yield ArchiveIssue('LICENSE file not found.');
+    return;
+  }
+  if (content == null || content.trim().isEmpty) {
+    yield ArchiveIssue('LICENSE file `$path` has no content.');
+    return;
+  }
+  if (content.contains('TODO: Add your license here.')) {
```
jonasfj, line 445: Can we make this: `content.trim().toLowerCase() == "TODO: Add your license here.".toLowerCase())`

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -163,6 +167,12 @@ Future<PackageSummary> summarizePackageArchive(String archivePath) async {
     examplePath = null;
   }
 
+  String licensePath = searchForFile(licenseFileNames);
+  final licenseContent = await extractContent(licensePath);
+  if (licenseContent == null) {
+    licensePath = null;
+  }
```
jonasfj, line 174: I think we should be okay to require that new packages have a `LICENSE` file and that the name of the file is `LICENSE`.

Ideally, it would be nice if when we read old packages we maintain the more relaxed rules we've previously had.

isoos: This code is also used by the package asset backfill, and will be applied on old versions.

jonasfj: okay, I guess making what we accept more strict is something we can do later :D

Issue #3905, opened 2020-08-03
Weird version numbers are problematic


jonasfj:
Example: `version: 1.0.04` in https://pub.dev/packages/flutter_walle/versions/1.0.4

jonasfj:
note. we also need a tool to cleanup any side-effects from this in datastore.

isoos:
I think this was cleaned up.

PR #3911, opened 2020-08-05
Better handling of version canonicalization

isoos:
- explicit `Pubspec.canonicalVersion`
- updated end2end test to use non-canonical input
- #3905

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -898,10 +898,7 @@ Future<_ValidatedUpload> _parseAndValidateUpload(
 
   final packageKey = db.emptyKey.append(Package, id: pubspec.name);
 
-  final versionString = canonicalizeVersion(pubspec.version);
-  if (versionString == null) {
-    throw InvalidInputException.canonicalizeVersionError(pubspec.version);
-  }
+  final versionString = pubspec.canonicalVersion;
 
   final version = PackageVersion()
     ..id = versionString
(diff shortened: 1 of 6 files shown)


app/lib/package/model_properties.dart
```diff
@@ -41,7 +45,17 @@ class Pubspec {
...
   String get name => _inner.name;
 
-  String get version => _inner.version.toString();
+  String get nonCanonicalVersion => _inner.version.toString();
+  String get canonicalVersion {
+    if (_canonicalVersion == null) {
+      _canonicalVersion = canonicalizeVersion(nonCanonicalVersion);
+      if (_canonicalVersion == null) {
+        throw InvalidInputException.canonicalizeVersionError(
+            nonCanonicalVersion);
```
jonasfj, line 54: This is not a good choice... If we load something from data-store it should never be in a state where we can't canonicalize it.

jonasfj: Can we throw an `AssertionError` instead, and then have a check for this at upload-time?

isoos:
> We need to prevent storing things that cannot be canonicalized.

We already prevent that: https://github.com/dart-lang/pub-dev/blob/master/app/lib/package/backend.dart#L903

app/lib/package/model_properties.dart
isoos: I think at that point it doesn't really matter which exception we throw (because it is some internal processing), but I can change this to throw `AssertionError`.

app/lib/package/backend.dart
```diff
@@ -898,10 +898,7 @@ Future<_ValidatedUpload> _parseAndValidateUpload(
 
   final packageKey = db.emptyKey.append(Package, id: pubspec.name);
 
-  final versionString = canonicalizeVersion(pubspec.version);
-  if (versionString == null) {
-    throw InvalidInputException.canonicalizeVersionError(pubspec.version);
-  }
+  final versionString = pubspec.canonicalVersion;
```
jonasfj, line 901: Are you sure we want to drop this? Isn't it better to reject than to have an internal error later?

isoos: Good point, fixed.

PR #3918, opened 2020-08-07
Package.isUnlisted flag

isoos:


isoos:
Updated code to backfill `isUnlisted` from `doNotAdvertize`. PTAL.

app/bin/tools/backfill_package_fields.dart
```diff
@@ -60,6 +62,7 @@ Future<void> _backfillPackageFields(Package p) async {
       package.likes ??= 0;
       package.doNotAdvertise ??= false;
       package.isDiscontinued ??= false;
+      package.isUnlisted ??= package.isDiscontinued;
       package.assignedTags ??= [];
```
jonasfj, line 66: Wouldn't it be best if we only made an insert, when we changed one of these properties?

jonasfj (approved):
Update changelog with note about running backfill, otherwise we forget to run backfill

app/bin/tools/backfill_package_fields.dart
isoos: Do you mean setting the `Package.updated` property?

jonasfj: No, I means, if we are changing no properties on the `package` entity, because there is nothing to backfill, we still call `tx.insert(package)`, isn't that slow?

isoos: Oh, that is handled in lines 48-54. If all the potential changes have non-zero value, we just return before doing the transaction.

PR #3919, opened 2020-08-07
Allow un-setting the announcement value in the running instance.

isoos:


app/lib/service/announcement/backend.dart
```diff
@@ -27,6 +27,8 @@ class AnnouncementBackend {
     final value = await secretBackend.lookup(SecretKey.announcement);
     if (value != null && value.trim().isNotEmpty) {
```
sigurdm, line 28: Nit: Perhaps switch the branches for slightly simpler control flow?

You could maybe also skip the `value.trim().isNotEmpty` check. Why can't we set the announcement  to an empty string?

isoos: We have a CLI utility that sets the Datastore value, but nothing that deletes it. If it is set to `''` (empty string), it is assumed the same as being deleted.

PR #3925, opened 2020-08-10
Prepare search result listing + cache to handle non-latest-stable versions

isoos:
- `PackageView` is used (among other things) to display the search results (ideally from cache).
  - Updated cache handling to use versions (if specified).
  - Storing latest stable and prerelease versions on it separately.
- `search` returns `version` if it is explicitly not the latest stable (will be used later when the index stores prerelease versions too)
- SDK index's `isExternal` flag is updated to care only about whether the `url` and the `description` is provided.

### DIFF
--- a/app/bin/tools/flag.dart
+++ b/app/bin/tools/flag.dart
@@ -42,12 +42,13 @@ Future main(List<String> arguments) async {
     if (isRead) {
       await _read(package);
     } else {
-      await _set(
+      final p = await _set(
         package,
         discontinued: discontinued,
         doNotAdvertise: doNotAdvertise,
       );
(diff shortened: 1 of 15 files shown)


jonasfj:
I think this is in support of listing prereleases in search results, let's first decide if we really want that.

If we want this PR anyways, then I'm happy to finish review and land it. But I suspect it's uninteresting unless we decide to have an option for search prereleases.

jonasfj:
This is refactorings to support searching for prereleases, if we want to add this option.

isoos:
I see no reason to keep this PR open, too many changes in parallel in the code, if we need this, it will need to be re-done.

PR #3932, opened 2020-08-12
Update User.email when Google Account email has changed.

### LINKED ISSUE #3921: Update User entity email when Google Account email is changed
Upon successful login we should check if the users email address has changed, and if so, we should update the `User` entity.

isoos:
Fixes #3921.

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -305,6 +305,17 @@ class AccountBackend {
     );
   }
 
+  Future<User> _updateUserEmail(User user, String email) async {
+    return await withRetryTransaction(_db, (tx) async {
+      final u = await tx.lookupValue<User>(user.key);
+      if (u.email != email) {
+        u.email = email;
+        tx.queueMutations(inserts: [u]);
+      }
+      return u;
+    });
+  }
+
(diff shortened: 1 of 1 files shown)


app/lib/account/backend.dart
```diff
@@ -317,6 +328,9 @@ class AccountBackend {
     // If the user exists, it's always cheaper to lookup the user outside a
     // transaction.
     final user = await _lookupUserByOauthUserId(auth.oauthUserId);
+    if (user != null && user.email != auth.email && auth.email != null) {
```
jonasfj, line 331: I assume `auth.email != null` is a sanity check, may I suggest we add `auth.email.isNotEmpty` too? Just as a sanity check :D

app/lib/account/backend.dart
```diff
@@ -305,6 +305,17 @@ class AccountBackend {
     );
   }
 
+  Future<User> _updateUserEmail(User user, String email) async {
+    return await withRetryTransaction(_db, (tx) async {
+      final u = await tx.lookupValue<User>(user.key);
+      if (u.email != email) {
+        u.email = email;
+        tx.queueMutations(inserts: [u]);
```
jonasfj, line 313: nit: `tx.insert(u)`

PR #3934, opened 2020-08-13
Block package upload if archive has symlinks

isoos:


app/lib/package/backend.dart
```diff
@@ -824,7 +825,23 @@ Future<void> _verifyTarball(String filename) async {
...
+  for (final line in lines) {
+    if (line.startsWith('l')) {
+      // report only the source path
+      // if output is non-standard for any reason, this reports the full line
+      final parts = line.split(' -> ');
+      final source = parts.first.split(' ').last;
+      throw PackageRejectedException.containsSymlink(source);
+    }
+  }
 }
```
jonasfj, line 845: Can we put this in a small function somewhere and test it... just for sanity?

isoos: The happy path is tested by integration tests, so I assume you are looking for a test where we create an archive with symlinks? or use one that is known to have it?

sigurdm: Yes - testing that this fails with a symlink would be nice!

PR #3938, opened 2020-08-14
Stronger landing page to survive search outages.

isoos:
- no more hardcoded fallback lists - if both search and cache is unavailable, the sections are not displayed
- 6 hours of caching for each of the top sections should be enough - based on the search outages we've had in the past
- local random selection will pick the first package from the top 20


app/lib/shared/versions.dart
```diff
@@ -21,6 +21,7 @@ final RegExp runtimeVersionPattern = RegExp(r'^\d{4}\.\d{2}\.\d{2}$');
 /// Make sure that at least two versions are kept here as the next candidates
 /// when the version switch happens.
 const acceptedRuntimeVersions = <String>[
+  // TODO: when removing, also remove SearchQuery.randomize
```
jonasfj, line 24: What run runtimeVersion does that speak about... can we say when removing '2020.08.14'...

jonasfj (approved):

PR #3947, opened 2020-08-18
Search input in the top-nav for package pages.

isoos:


app/lib/frontend/templates/layout.dart
```diff
@@ -115,6 +115,7 @@ String _renderSiteHeader(PageType pageType) {
 
   return templateCache.renderTemplate('shared/site_header', {
     'show_site_logo': pageType != PageType.landing,
+    'show_header_search': pageType == PageType.package,
```
jonasfj, line 118: I agree we shouldn't show it the landing page or search result page, but we could show this everywhere else as part of the default design, could we not? publishers, standalone markdown pages (like policy and help).

jonasfj (approved):

app/lib/frontend/templates/layout.dart
isoos: Publishers have their own search, added standalone and error pages to the list.

PR #3960, opened 2020-08-25
Read/write state of pkg/fake_gcloud backends.

isoos:
This will be used to store the state for local developments.

### DIFF
--- a/pkg/fake_gcloud/lib/mem_datastore.dart
+++ b/pkg/fake_gcloud/lib/mem_datastore.dart
@@ -1,4 +1,5 @@
 import 'dart:convert';
+import 'dart:math' as math;
 import 'dart:typed_data';
 
 import 'package:gcloud/common.dart';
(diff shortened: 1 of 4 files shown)


pkg/fake_gcloud/lib/mem_datastore.dart
```diff
@@ -218,6 +219,101 @@ class MemDatastore implements Datastore {
...
+  /// JSON-encoded data format.
+  void writeTo(StringSink sink) async {
+    _entities.forEach((_, entity) {
+      sink.writeln(json.encode({'entity': _encodeEntity(entity)}));
+    });
+    sink.writeln(json.encode({'_unusedId': _unusedId}));
+  }
+
+  /// Reads content as a line-by-line JSON-encoded data format.
+  void readFrom(Iterable<String> lines) {
```
jonasfj, line 233: Wouldn't it be better to make this an async static function that returns an instance of `MemDatastore`?
Same with `MemStorage` ?

jonasfj: I would also suggest taking `Stream<List<int>>` as input, or `List<int>`

jonasfj (approved):

pkg/fake_gcloud/lib/mem_datastore.dart
jonasfj: Just nits...

jonasfj: Note, one could even argue that we should expose `MemDatastore` as a type and just export a function that returns an instance of `Datastore` and have `_MemDatastore` be an internal type..

For saving we could just have a function which takes a `Datastore` instances, casts to `_MemDatastore` or throws an `ArgumentError`.

isoos: I think hiding the type doesn't provide much value, because in the test we should know about it at the time of setup, and also when (re)storing the state. As long as the package is imported as dev_dependency only, we are in the clear. Maybe we could detect in their constructor if we are running inside appengine, and if we are, throw an error - making sure we never use it in prod?

isoos: Re: `List<int>` vs. Strings... I think the current API is useful for in-memory tests and to plug into using the `File` methods. I'll reconsider when implementing the actual file handling for local development.

Issue #3904, opened 2020-08-03
Handle "The transaction was aborted" issues


jonasfj:
Migrate all transactions to use:
https://github.com/dart-lang/pub-dev/blob/13d94ca4adc70027c941149ac8d633e2ffea8e52/app/lib/shared/datastore_helper.dart#L102-L115

PR #3967, opened 2020-08-26
Retry transaction on package upload.

isoos:
- #3904
- moved `try-catch` logging from the upload block to the helper method

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -513,17 +513,14 @@ class PackageBackend {
         validatedUpload.packageVersionPubspec,
         validatedUpload.packageVersionInfo,
         ...validatedUpload.assets,
-      ];
-      if (historyBackend.isEnabled) {
-        final history = History.entry(PackageUploaded(
+        History.entry(PackageUploaded(
           packageName: newVersion.package,
           packageVersion: newVersion.version,
           uploaderId: user.userId,
           uploaderEmail: user.email,
           timestamp: newVersion.created,
-        ));
-        inserts.add(history);
-      }
+        )),
+      ];
 
       _logger.info('Trying to commit datastore changes.');
       tx.queueMutations(inserts: inserts);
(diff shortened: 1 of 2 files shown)


app/lib/shared/datastore_helper.dart
```diff
@@ -52,24 +52,29 @@ Future<T> withTransaction<T>(
...
-      if (!commitAttempted) {
-        await tx.rollback();
-      }
-    }
-  });
+    });
+  } catch (e, st) {
+    _logger.info('Transaction failed.', e, st);
+    rethrow;
+  }
```
jonasfj, line 77: Why do we need this? seems like noise...`withTransaction` is used by `withRetryTransaction` which has an `onRetry`.

We could question if we should even export `withTransaction` I'm not sure there is many places where you  shouldn't retry transactions. Maybe with likes it better to give up than to retry.. but probably not :D

app/lib/package/backend.dart
```diff
@@ -504,52 +501,35 @@ class PackageBackend {
...
+          packageVersion: newVersion.version,
+          uploaderId: user.userId,
+          uploaderEmail: user.email,
+          timestamp: newVersion.created,
+        ));
+        inserts.add(history);
       }
+
+      _logger.info('Trying to commit datastore changes.');
+      tx.queueMutations(inserts: inserts);
```
jonasfj, line 529: nit... just use `tx.insert` then we don't need to make a list of `inserts` at all :)

jonasfj (changes requested):

app/lib/shared/datastore_helper.dart
isoos: I though that we should have a generic logger around transactions and see a trace of what failed and where. (nothing else has changed in this method).

The visibility can be changed after we've migrated its uses to use `withRetryTransaction`.

app/lib/package/backend.dart
isoos: I've moved the history entry inside the `inserts` list. It doesn't seem to worth switching to individual `tx.insert` calls with that...

app/lib/shared/datastore_helper.dart
jonasfj: Can we make a version of `withTransaction` which doesn't do the logging, so that `withRetryTransaction` doesn't log entries it's retrying twice?

jonasfj (approved):

PR #3981, opened 2020-09-01
Move removePackageVersion to admin backend + API + test.

isoos:


app/lib/admin/backend.dart
```diff
@@ -382,13 +382,97 @@ class AdminBackend {
...
 
-  Future _deleteWithQuery<T>(Query query) async {
+  /// Removes the specific package version from the Datastore and updates other
+  /// related entities. It is safe to call [removePackageVersion] on an already
+  /// removed version, as the call is idempotent.
+  Future<void> removePackageVersion(String packageName, String version) async {
+    final caller = await _requireAdminPermission(AdminPermission.removePackage);
+
+    _logger.info('${caller.userId} (${caller.email}) initiated the delete '
+        'of package $packageName $version');
```
jonasfj, line 392: This won't work when called from `app/bin/tools/remove_package.dart`

isoos: Ack. But we want to move that tool to call the API (in a subsequent PR).

jonasfj: Okay...

jonasfj (approved):

PR #3982, opened 2020-09-02
Optionally store fake pub server's internal state.

isoos:


pkg/fake_pub_server/bin/fake_pub_server.dart
```diff
@@ -115,4 +130,58 @@ Future main(List<String> args) async {
     ],
     eagerError: true,
   );
+
+  await state.store();
+  await sigintSubscription.cancel();
+}
+
+class _LocalServerState {
```
jonasfj, line 138:
```suggestion
/// Owns server state, optionally loading / saving state to/from [dataDir]
class _LocalServerState {
```

pkg/fake_pub_server/bin/fake_pub_server.dart
```diff
@@ -115,4 +130,58 @@ Future main(List<String> args) async {
...
+  Future<void> init() async {
+    if (_datastoreFile != null && await _datastoreFile.exists()) {
+      datastore.readFrom(await _datastoreFile.readAsLines());
+    }
+    if (_storageFile != null && await _storageFile.exists()) {
+      storage.readFrom(await _storageFile.readAsLines());
+    }
+  }
+
+  Future<void> store() async {
```
jonasfj, line 163: Perhaps call it `save()`

pkg/fake_pub_server/bin/fake_pub_server.dart
```diff
@@ -115,4 +130,58 @@ Future main(List<String> args) async {
...
+
+  _LocalServerState({String dataDir}) {
+    if (dataDir != null) {
+      _dataDir = Directory(dataDir);
+      _datastoreFile = File(p.join(dataDir, 'datastore.jsonl'));
+      _storageFile = File(p.join(dataDir, 'storage.jsonl'));
+    }
+  }
+
+  Future<void> init() async {
```
jonasfj, line 154: Why not call this method `load()`

pkg/fake_pub_server/bin/fake_pub_server.dart
```diff
@@ -115,4 +130,58 @@ Future main(List<String> args) async {
...
+  Directory _dataDir;
+  File _datastoreFile;
+  File _storageFile;
+  Completer _storingCompleter;
+
+  _LocalServerState({String dataDir}) {
+    if (dataDir != null) {
+      _dataDir = Directory(dataDir);
+      _datastoreFile = File(p.join(dataDir, 'datastore.jsonl'));
+      _storageFile = File(p.join(dataDir, 'storage.jsonl'));
```
jonasfj, line 150: Storing all buckets in a single file seems a bit extreme... but it's probably fine, since we won't store that much :D

wouldn't it be attractive to store it all in a single file, now that we are doing this anyways? That also makes it easier to send snapshots around.

We could just make the file:
```
"datastore-entries"
{...}
{...}
...
{...}
"storage-entries"
{...}
{...}
...
{...}
```

And use `takeWhile()` when reading it in `init()`.

pkg/fake_pub_server/bin/fake_pub_server.dart
```diff
@@ -91,6 +99,13 @@ Future main(List<String> args) async {
     return null;
   }
 
+  // Store the state (and then exit) on CTRL+C.
+  final sigintSubscription = ProcessSignal.sigint.watch().listen((e) async {
+    print('Storing state...');
```
jonasfj, line 104:
```suggestion
    print('Storing state in ${argv['data-dir']}');
```

jonasfj (approved):
Just nits, storing as multiple files is fine :D

But adding a few comments would be great.

pkg/fake_pub_server/bin/fake_pub_server.dart
isoos: Good point, implementing it...

PR #4004, opened 2020-09-08
Expose like and popularity on the /score API endpoint.

### LINKED ISSUE #3996: Expose all three score components
In our new design, the package score has three components:

1. Points (aka Pub Points)
1. Popularity
1. Likes

We currently expose these via several APIs:

For likes: https://pub.dev/api/packages/pana/likes
For popularity: https://pub.dev/api/packages/pana/metrics?pretty
For points https://pub.dev/api/packages/pana/metrics?pretty & https://pub.dev/api/packages/pana/score

I suggest we expand the `metrics` and `score` APIs to have all three score components:

```
{
  "grantedPoints":60,
  "maxPoints":110,
  "popularityScore": 0.6633297696569151,
  "likes": 7,
  "lastUpdated":"2020-09-03T22:11:22.940284Z"
}
```

isoos:
Fixes #3996.

### DIFF
--- a/pkg/client_data/lib/package_api.dart
+++ b/pkg/client_data/lib/package_api.dart
@@ -154,11 +154,15 @@ class VersionInfo {
 class VersionScore {
   final int grantedPoints;
   final int maxPoints;
+  final int likeCount;
+  final double popularityScore;
   final DateTime lastUpdated;
 
   VersionScore({
     @required this.grantedPoints,
     @required this.maxPoints,
+    @required this.likeCount,
+    @required this.popularityScore,
     @required this.lastUpdated,
   });
 
--- a/app/lib/frontend/handlers/custom_api.dart
+++ b/app/lib/frontend/handlers/custom_api.dart
@@ -203,11 +203,17 @@ Future<shelf.Response> apiPackageMetricsHandler(
 Future<VersionScore> packageVersionScoreHandler(
     shelf.Request request, String package,
     {String version}) async {
-  final data = await scoreCardBackend.getScoreCardData(package, version);
+  final pkg = await packageBackend.lookupPackage(package);
+  if (pkg == null) {
+    throw NotFoundException.resource('package "$package"');
+  }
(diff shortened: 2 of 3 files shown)


pkg/client_data/lib/package_api.dart
```diff
@@ -154,11 +154,15 @@ class VersionInfo {
 class VersionScore {
   final int grantedPoints;
   final int maxPoints;
+  final int likeCount;
```
jonasfj, line 157: Can we rename this to `likes`

pkg/client_data/lib/package_api.dart
```diff
@@ -154,11 +154,15 @@ class VersionInfo {
...
+  final double popularityScore;
```
jonasfj, line 158: Can we rename this `popularity`

jonasfj (approved):

pkg/client_data/lib/package_api.dart
isoos: I'd argue that we should use `likeCount` as opposed to `likes` not only here, but other places where we've had `likes` before. `List<Like>` should be `likes` (and `likes` should be `List<Like>`), while the cardinality of the same list is the `likeCount`.

pkg/client_data/lib/package_api.dart
isoos: I was pondering on this, and thought that if we'd call it `popularity`, then we should also transform it into the `0-100` range, and switch the type to `int`. wdyt?

jonasfj: I'm also okay keeping it this way. let's not over think it, long term I think we want absolute numbers here.

pkg/client_data/lib/package_api.dart
jonasfj: Seems fine either way, I just wanted to try to keep the external API neat.

jonasfj (approved):

PR #4007, opened 2020-09-08
Use gmail SMTP relay with SASL XOAUTH2

jonasfj:
I don't think we want to roll this out just yet..

### DIFF
--- a/app/lib/frontend/email_sender.dart
+++ b/app/lib/frontend/email_sender.dart
@@ -4,14 +4,17 @@
 
 import 'dart:async';
 import 'dart:io';
+import 'dart:convert' show json, base64, ascii;
 
 import 'package:gcloud/service_scope.dart' as ss;
+import 'package:gcloud/http.dart' show authClientService;
+import 'package:googleapis/iamcredentials/v1.dart' as iam_credentials;
 import 'package:logging/logging.dart';
 import 'package:mailer/mailer.dart';
 import 'package:mailer/smtp_server.dart';
 import 'package:retry/retry.dart';
+import 'package:http/http.dart' as http;
 
-import '../service/secret/backend.dart';
 import '../shared/email.dart';
 import '../shared/exceptions.dart';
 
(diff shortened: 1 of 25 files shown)


app/lib/frontend/email_sender.dart
```diff
@@ -24,77 +27,201 @@ void registerEmailSender(EmailSender value) =>
...
-    final now = DateTime.now().toUtc();
-    if (_lastUpdated != null && now.difference(_lastUpdated).inMinutes < 10) {
-      return;
+  Future<SmtpServer> _getSmtpServer() async {
+    final maxAge = DateTime.now().subtract(Duration(minutes: 15));
+    if (_accessToken == null && _accessTokenRefreshed.isBefore(maxAge)) {
+      _accessToken = _createAccessToken();
+      _accessTokenRefreshed = DateTime.now();
     }
-    final username = await secretBackend.lookup(SecretKey.smtpUsername);
```
isoos, line 77: Please add a TODO. If these are not used anymore, we'll need to clean that up in the code and in Datastore.

jonasfj: Good point!

app/test/shared/test_services.dart
```diff
@@ -59,7 +59,7 @@ void testWithServices(
       await fork(() async {
         registerAuthProvider(FakeAuthProvider());
         registerDomainVerifier(FakeDomainVerifier());
-        registerEmailSender(FakeEmailSender());
+        registerEmailSender(fakeEmailSender.sendMessage);
```
isoos, line 62: This should be a new instance each time this block is running (and the reason of test failures), otherwise we'd need to clear the buffered list of emails before checking them.

jonasfj: Ah, then sticking to a class with a single method is perhaps better..

isoos (approved):
Need to update one more test expect, but otherwise OK.

PR #4011, opened 2020-09-09
Finalizing new flags and added integrity checks.

isoos:


isoos:
Note: the `required: true` properties on these flags may be premature (and we could have only the integrity check first), but if the backfill did run without issues, it should be fine.

jonasfj (approved):
Can we put in `CHANGELOG.md` to run integrity check before deploying?

I just ran backfills, there was a few new users that needed backfilling, but it looks like they were from yesterday.

PR #4013, opened 2020-09-09
Upload restrictions.

isoos:


app/lib/service/secret/models.dart
```diff
@@ -27,13 +27,19 @@ abstract class SecretKey {
   /// JSON-encoded list of Strings that will be considered as spam.
   static const String spamWords = 'spam-words';
 
+  /// The restriction applied on uploads.
+  ///
+  /// Valid values are `no-uploads`, `only-updates` and `no-restriction` (or ``, `-`).
```
jonasfj, line 32:
```suggestion
  /// Valid values for `upload-restriction` are:
  ///  * `no-uploads`, no package publications will be accepted by the server,
  ///  * `only-updates`, publication of new packages will not be accepted, but new versions of existing packages will be accepted, and,
  ///  * `no-restriction`, (default) publication of new packages and new versions is allowed.
```

app/lib/package/backend.dart
```diff
@@ -763,6 +780,30 @@ class PackageBackend {
...
+    // safe fallback on enabling uploads
+    _logger.warning('Unknown upload restriction status: $value');
+    return UploadRestrictionStatus.noRestriction;
+  }
+}
+
+enum UploadRestrictionStatus {
+  noRestriction,
+  onlyUpdates,
+  noUploads,
```
jonasfj, line 806:
```suggestion
  /// Publication of new packages and new versions of existing packages is allowed.
  noRestriction,
  /// Publication of new packages is **not** allowed, new versions of existing packages is allowed.
  onlyUpdates,
  /// Publication of packages is **not** allowed.
  noUploads,
```

app/lib/service/secret/models.dart
```diff
@@ -27,13 +27,19 @@ abstract class SecretKey {
   /// JSON-encoded list of Strings that will be considered as spam.
   static const String spamWords = 'spam-words';
 
+  /// The restriction applied on uploads.
+  ///
```
jonasfj, line 31:
```suggestion
  ///
  /// This feature is intended as an emergency break.
  ///
```

jonasfj (approved):
I think we should document this in pub-ops, such that incident response instructions highlights the availability of this emergency break.

PR #4020, opened 2020-09-10
Fix session purge issue by using a dedicated String property.

isoos:


app/lib/account/models.dart
```diff
@@ -120,7 +120,12 @@ class UserSession extends db.ExpandoModel<String> {
   @db.ModelKeyProperty(required: true)
   db.Key userIdKey;
```
jonasfj, line 121:
```suggestion
  @Deprecated('Prefer userId')
  @db.ModelKeyProperty(required: true)
  db.Key userIdKey;
```

jonasfj (approved):
Okay, we have no migration script because these entities timeout after 14 days, right?

isoos:
> Okay, we have no migration script because these entities timeout after 14 days, right?

Yes, correct.

PR #4032, opened 2020-09-17
Checkbox-based admin UI for isUnlisted + isDiscontinued.

isoos:


app/lib/frontend/templates/views/pkg/admin_page.mustache
```diff
@@ -48,32 +48,52 @@
...
-  </p>
-  <p>
-    <button
-      id="-admin-is-discontinued-toggle"
-      class="mdc-button mdc-button--raised"
-      data-mdc-auto-init="MDCRipple">Mark as "discontinued"</button>
-  </p>
+  If no one is maintaining this package, you can mark it as discontinued.
+  Discontinued packages remain available, but pub.dev doesn’t display them
+  in search results and won't analyze it for scoring.
```
jonasfj, line 57: Maybe something like:
```html
A package can be marked as <a href="https://dart.dev/tools/pub/publishing#discontinue">discontinued</a>
to inform users that the package is no longer maintained.
<i>Discontinued packages</i> remain available to package users, but they will not show up
in search results on pub.dev unless explicitly included in the advanced search options.
```

@kwalrath if you have any quick suggestions for explanatory text feel free to leave us some wisdom :D

This is the "bla bla bla" section from the mockup in go/is:unlisted-on-pub

app/lib/frontend/templates/views/pkg/admin_page.mustache
```diff
@@ -48,32 +48,52 @@
...
-  </p>
-  <p>
-    <button
-      id="-admin-is-discontinued-toggle"
-      class="mdc-button mdc-button--raised"
-      data-mdc-auto-init="MDCRipple">Remove "discontinued"</button>
-  </p>
+  If you don't want this package to be discovered, you can mark it unlisted.
+  Unlisted packages remain available, they are getting analyzed for scoring,
+  but pub.dev doesn’t display them in search results.
```
jonasfj, line 81: Maybe something like:
```html
A package can be marked as <i>unlisted</i> to avoid crowding search results.
Unlisted packages remain publicly available, but they will not show up in search results on pub.dev
with default search filters (<i>unlisted packages</i> can still be found using advanced search options).
```

IMO, the use-case is: to avoid crowding search results. But we could give examples of when this might be useful. Typically, it's when publishing a package that only used as transitive dependency, for example a package that contains libraries used internally by other packages, but that you don't think people should be using directly.

Perhaps it's better to document this and give examples on a page like: `https://dart.dev/tools/pub/publishing#unlisted`
Filed https://github.com/dart-lang/site-www/issues/2647 to track this.

@kwalrath if you have any quick suggestions for explanatory text feel free to leave us some wisdom :D

This is the "bla bla bla" section from the mockup in go/is:unlisted-on-pub

app/lib/frontend/templates/views/pkg/admin_page.mustache
```diff
@@ -52,9 +52,10 @@
 
 <h3>Discontinued</h3>
 <div>
-  If no one is maintaining this package, you can mark it as discontinued.
-  Discontinued packages remain available, but pub.dev doesn’t display them
-  in search results and won't analyze it for scoring.
+  A package can be marked as <a href="https://dart.dev/tools/pub/publishing#discontinue">discontinued</a>
+  to inform users that the package is no longer maintained.
+  <i>Discontinued packages</i> remain available to package users, but they will not show up
+  in search results on pub.dev unless explicitly included in the advanced search options.
```
kwalrath, line 58:
```suggestion
  in search results on pub.dev unless the user specifies advanced search options.
```

kwalrath (commented):
I'd like to avoid the word "crowding" (seems unclear) and make this a little more active voice and present tense. I don't have any brilliant suggestions, though.

jonasfj (approved):

PR #4039, opened 2020-09-18
Reject queries that override required tags.

isoos:


app/lib/search/search_service.dart
```diff
@@ -371,6 +377,26 @@ class SearchQuery {
...
+      return QueryValidity.reject(rejectReason: 'Query too long.');
+    }
+
+    // Do not allow override of required tags. (E.g. search scope would require
+    // sdk:flutter, do not allow -sdk:flutter to override it).
+    final invalidOverride =
+        tagsPredicate._firstInvalidOverride(parsedQuery.tagsPredicate);
+    if (invalidOverride != null) {
+      return QueryValidity.reject(
+          rejectReason: 'Invalid override: `$invalidOverride`.');
```
jonasfj, line 395: Can't we change this?

Either say `invalid tag`? or something else...

Ideally, we could explain that: You need to search from another context or change "advanced options" to avoid this conflict.

Maybe:
> The tag `sdk:flutter` is given from search filters and cannot be combined with `-sdk:flutter`. 

Or:
> The tag `-sdk:flutter` conflicts with search filters.

app/lib/search/search_service.dart
```diff
@@ -473,12 +508,31 @@ class TagsPredicate {
     return p;
   }
 
+  String _firstInvalidOverride(TagsPredicate other) {
```
jonasfj, line 511: Can we give this a documentation string, and maybe use the terminology "conflictingTags" or something like that.

We don't allow overrides, so any override is an _invalid override_. Hence, the word "override" might not work so well.

isoos: I was thinking that we allow overrides of negative tags, but you are right, we don't need those (if we have tri-state checkboxes for those we allow multiple states).

jonasfj (approved):

Issue #4044, opened 2020-09-21
Add is:unlisted option for packages


jonasfj:
A package marked unlisted:
 * [x] will be shown with an [unlisted] tag on its package details page
     * [x] This tag has a tooltip that explains what this means.
 * [x] is not displayed in default search results
* [x] will not show up on the pub.dev landing page
* [x] can be included in search results using advanced search options.
* [ ] can be included in search results using the tag is:unlisted
* [x] will have a package page with a noindex meta-tag
   * [x] Causing such package pages to not be indexed by Google and other search engines.
 * [x] will still show up on the public list-of-packages for a given publisher (e.g. dart.dev).

---------------
We will need:
  * [x] UI for setting `is:unlisted` (https://github.com/dart-lang/pub-dev/pull/4032)
  * [x] Option for including unlisted (and discontinued) packages in search, through _advanced search options_.

PR #4046, opened 2020-09-22
Display [unlisted] label

isoos:
- https://github.com/dart-lang/pub-dev/issues/4044
- We no longer have a "single" label (e.g. either discontinued or having an SDK), rather we have multiple labels, and adjusted the label selection logic accordingly.
- Unlisted package with SDK information:
  Screenshot 2020-09-22 at 13 36 03
  same on the listing page:
  Screenshot 2020-09-22 at 13 36 18

- Unlisted + discontinued package:
  Screenshot 2020-09-22 at 13 36 25

### DIFF
--- a/app/lib/frontend/templates/misc.dart
+++ b/app/lib/frontend/templates/misc.dart
@@ -198,7 +198,11 @@ String renderTags({@required PackageView package}) {
       'status': 'unlisted',
       'text': 'unlisted',
       'has_href': false,
-      'title': 'Package was unlisted.',
+      'title': 'Package is unlisted, this means that while the package is still '
+          'publicly available the author has decided that it should not appear '
+          'in search results with default search filters. This is typically '
+          'done because this package is meant to support another package, '
+          'rather than being consumed directly.',
     });
   }
   if (package.isObsolete) {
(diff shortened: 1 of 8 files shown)


jonasfj (approved):
@mit-mit should we perhaps pick a different color for the `unlisted` badge?

IMO. _unlisted_ doesn't have to look as scary as _discontinued_.

isoos:
> unlisted doesn't have to look as scary as discontinued.

We have a lighter color for unidentified:

Screenshot 2020-09-22 at 15 41 38

mit-mit:
Same color as unindentified looks good.

Why do we have those `[ ]` brackets in the tags? Consider removing them, I think.

mit-mit:
Also, does UNLISTED have a tooltip explaining what it means?

isoos:
> Also, does UNLISTED have a tooltip explaining what it means?

Only a very simple "Package was unlisted."

app/lib/frontend/templates/misc.dart
```diff
@@ -181,97 +180,102 @@ String renderMiniList(List<PackageView> packages) {
...
+  }
+  if (package.tags.contains(PackageTags.isUnlisted)) {
     tagValues.add({
-      'status': 'discontinued',
-      'text': '[discontinued]',
+      'status': 'unlisted',
+      'text': 'unlisted',
       'has_href': false,
-      'title': 'Package was discontinued.',
+      'title': 'Package was unlisted.',
```
jonasfj, line 201: Let's make this more elaborate, something like:
> Package is unlisted, this means that while the package is still publicly available the author has decided that it
> should not appear in search results with default search filters. This is typically done because this package is meant
> to support another package, rather than being consumed directly.

PR #4047, opened 2020-09-23
Include unlisted: query params + search UI

isoos:
#4044

Screenshot 2020-09-23 at 11 02 34

### DIFF
--- a/app/lib/frontend/templates/views/pkg/index.mustache
+++ b/app/lib/frontend/templates/views/pkg/index.mustache
@@ -31,9 +31,17 @@
         {{& subsdk_tabs_advanced_html }}
       </div>
       {{/has_subsdk_tabs_advanced_html}}
-      <div class="search-controls-checkbox">
-        <input id="search-legacy-checkbox" type="checkbox" name="legacy" valye="1" {{#legacy_search_enabled}} checked="checked"{{/legacy_search_enabled}} />
-        <label for="search-legacy-checkbox">Include Dart 1.x results</label>
+      <div>
+        <div class="search-controls-checkbox">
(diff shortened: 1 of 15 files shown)


jonasfj (approved):

app/lib/frontend/templates/views/pkg/index.mustache
```diff
@@ -31,9 +31,17 @@
         {{& subsdk_tabs_advanced_html }}
       </div>
       {{/has_subsdk_tabs_advanced_html}}
-      <div class="search-controls-checkbox">
-        <input id="search-legacy-checkbox" type="checkbox" name="legacy" valye="1" {{#legacy_search_enabled}} checked="checked"{{/legacy_search_enabled}} />
-        <label for="search-legacy-checkbox">Include Dart 1.x results</label>
+      <div>
+        <div class="search-controls-checkbox">
+          <input id="-search-unlisted-checkbox" type="checkbox" name="unlisted" {{#include_unlisted}} checked="checked"{{/include_unlisted}} />
+          <label for="-search-unlisted-checkbox">Include unlisted packages</label>
```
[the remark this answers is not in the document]
isoos: Without dash, there is a possibility that a readme produces the same id in a title, so we shall use only the dash-prefixed ones.

app/lib/frontend/templates/views/pkg/index.mustache
```diff
@@ -31,9 +31,17 @@
...
+        </div>
+      </div>
+      <div>
+        <div class="search-controls-checkbox">
+          <input id="search-legacy-checkbox" type="checkbox" name="legacy" {{#legacy_search_enabled}} checked="checked"{{/legacy_search_enabled}} />
```
jonasfj, line 42: but no dashes here?

isoos: not yet :)

PR #4074, opened 2020-10-01
Proper API for inviting a new uploader.

isoos:
- The old API was designed to instantly add an uploader, and it is not useful for admin UI.
- #3553

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -7,7 +7,7 @@ library pub_dartlang_org.backend;
 import 'dart:async';
 import 'dart:io';
 
-import 'package:client_data/account_api.dart';
+import 'package:client_data/account_api.dart' as account_api;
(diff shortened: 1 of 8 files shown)


app/lib/package/backend.dart
```diff
@@ -641,12 +643,14 @@ class PackageBackend {
 
   // Uploaders support.
 
-  Future<api.SuccessMessage> addUploader(
-      String packageName, String uploaderEmail) async {
-    uploaderEmail = uploaderEmail.toLowerCase();
+  Future<account_api.InviteStatus> inviteUploader(
+      String packageName, api.InviteUploaderRequest invite) async {
+    InvalidInputException.checkNotNull(invite?.email, 'email');
```
jonasfj, line 648: Calling without `invite` is that not an `ArgumentError.checkNotNull(invite, 'invite');`

jonasfj (approved):

PR #4075, opened 2020-10-01
Fix PackageVersionAsset condition in cleanup script

isoos:


app/bin/tools/remove_noncanonical_entities.dart
```diff
@@ -35,7 +35,7 @@ Future main(List<String> args) async {
   await withToolRuntime(() async {
     await _deleteWithQuery<PackageVersionAsset>(
       dbService.query<PackageVersionAsset>(),
-      where: (a) => canonicalizeVersion(a.packageVersion) != a.packageVersion,
+      where: (a) => canonicalizeVersion(a.version) != a.version,
```
jonasfj, line 38: How did this not delete a lot more?

Also what about adding documentation to `PackageVersionAsset.version` and `PackageVersionAsset.packageVersion` a single doc-comment clarifying the format would have made this easier to spot :D

jonasfj (approved):

PR #4081, opened 2020-10-02
More flexible consent backend

isoos:
This is in preparation the backend for unauthenticated access.

### DIFF
--- a/app/lib/account/consent_backend.dart
+++ b/app/lib/account/consent_backend.dart
@@ -247,10 +247,8 @@ class ConsentBackend {
       () async {
         await action?.onAccept(consent);
         await withRetryTransaction(_db, (tx) async {
-          final c =
-              await tx.lookupValue<Consent>(consent.key, orElse: () => null);
-          if (c == null) return;
-          tx.delete(c.key);
+          final c = await tx.lookupOrNull<Consent>(consent.key);
+          if (c != null) tx.delete(c.key);
         });
       },
       maxAttempts: 3,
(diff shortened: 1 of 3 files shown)


app/lib/account/consent_backend.dart
```diff
@@ -247,10 +247,8 @@ class ConsentBackend {
       () async {
         await action?.onAccept(consent);
         await withRetryTransaction(_db, (tx) async {
-          final c =
-              await tx.lookupValue<Consent>(consent.key, orElse: () => null);
-          if (c == null) return;
-          tx.delete(c.key);
+          final c = await tx.lookupOrNull<Consent>(consent.key);
+          if (c != null) tx.delete(c.key);
```
jonasfj, line 251: Why not simply `tx.delete(consent.key)` -- also should we not being doing this in the same transaction as everything else done when accepting the invite?

Like maybe the `tx` should be passed to `onAccept`

app/lib/account/consent_backend.dart
```diff
@@ -263,10 +261,8 @@ class ConsentBackend {
       () async {
         await action?.onDelete(consent);
         await withRetryTransaction(_db, (tx) async {
-          final c =
-              await tx.lookupValue<Consent>(consent.key, orElse: () => null);
-          if (c == null) return;
-          tx.delete(c.key);
+          final c = await tx.lookupOrNull<Consent>(consent.key);
+          if (c != null) tx.delete(c.key);
```
jonasfj, line 265: Why not simply `tx.delete(consent.key)`? I assume we would have to handle the case where it's already deleted.. but that's about it?

jonasfj (approved):
just things we should consider that doesn't block this PR :D

app/lib/account/consent_backend.dart
isoos: This code predates the retry-datastore-transactions code (hence the outer retry), and it was not merged since. I'll do a follow-up with that.

app/lib/account/consent_backend.dart
isoos: This is the consequence of the same outer retry block. This should be removed once the onaccept and the transaction is merged.

PR #4088, opened 2020-10-05
Split SearchQuery: Frontend* and Service*

isoos:
- Removed unused methods and fields from each.
- Removed the use of `includeLegacy` from search index.
- Other migrations postponed to follow-up PRs (filtering publisher with tag, handle frontend query without tags predicate).


isoos:
Tested on staging, seems to work as expected.

jonasfj (approved):

PR #4093, opened 2020-10-06
Package.replacedBy option.

isoos:


pkg/client_data/lib/package_api.dart
```diff
@@ -97,6 +99,14 @@ class PackageData {
   /// Package name.
   final String name;
 
+  /// `true` if package is discontinued.
+  /// If it is omitted, `null` or `false` the package is *not discontinued*.
+  final bool isDiscontinued;
+
+  /// If [isDiscontinued] is set, this points to the package that should be used
+  /// instead (set by the package admin).
```
jonasfj, line 107:
```suggestion
  /// If [isDiscontinued] is set, this _may_ point to a package that can be used
  /// instead (set by the package admin).
```

app/lib/package/models.dart
```diff
@@ -81,6 +81,11 @@ class Package extends db.ExpandoModel<String> {
   @db.BoolProperty(required: true)
   bool isDiscontinued;
 
+  /// The package that should be used instead of the current package.
+  /// May have a value only if [isDiscontinued] is set.
+  @db.StringProperty()
+  String replacedBy;
```
jonasfj, line 87: Can we augment integrity check to ensure that the package name mentioned here exists?

And that it is `null` whenever `discontinued = false`.

jonasfj (approved):

PR #4098, opened 2020-10-07
Added CSP script urls from secrets

jonasfj:


app/lib/shared/utils.dart
```diff
@@ -413,3 +413,29 @@ String createUuid([List<int> bytes]) {
...
+  /// In other words, if this [DateTime] is more than [duration] time ago, since
+  /// [relativeTo] (or now, if not given). This will return true.
+  ///
+  /// **Example**
+  /// ```
+  /// final start = DateTime.now();
+  ///
+  /// // do some work...
+  ///
+  /// if (start.isTimeAgo(Duration(minutes: 15))) {
```
isoos, line 430: `Stopwatch` would seem to be a more appropriate thing for that (explicit elapsed time measurement). Maybe a different example would be more fit?

app/lib/shared/utils.dart
```diff
@@ -413,3 +413,29 @@ String createUuid([List<int> bytes]) {
...
+  ///   // Start is more than 15 minutes ago.
+  /// }
+  /// ```
+  ///
+  bool isTimeAgo(Duration duration, [DateTime relativeTo]) {
```
isoos, line 435: I have never felt the need for such utility, as it seems to be about the same length-wise compared to the direct diff() + comparison. Are you planning more like that / into this method?

app/lib/shared/handlers.dart
```diff
@@ -99,11 +101,43 @@ final _contentSecurityPolicyMap = <String, List<String>>{
...
+          ..._contentSecurityPolicyMap['script-src'],
+          ...cspScriptUrls,
+        ],
+      };
+
+      return cspMap.keys.map<String>((key) {
+        final list = cspMap[key];
+        return '$key ${list.join(' ')}';
+      }).join(';');
+    }();
```
isoos, line 137: Let's separate the function from the value:
```
Future<String> lookupCsp() async {...}
_cspCache = lookupCsp();
```

app/lib/shared/utils.dart
jonasfj: I felt like I needed it because I'm always doing this wrong... remember last time with the email logic.

jonasfj:
Abandoning this idea, and just harding the few extra things we need instead.

Long-term I think we should use `strict-dynamic nonce-...`, but we need to tweak our templates to do this.

PR #4101, opened 2020-10-07
Use Google Tag Manager for analytics

jonasfj:
We probably still need to migrate the generated dartdoc, but
I don't think we can make that migration experimental.

### DIFF
--- a/app/lib/frontend/templates/views/shared/layout.mustache
+++ b/app/lib/frontend/templates/views/shared/layout.mustache
@@ -4,8 +4,15 @@
 <!DOCTYPE html>
 <html lang="en-us">
 <head>
+  {{#is_experimental}}
+  <script async="async" src="https://www.googletagmanager.com/gtm.js?id=GTM-MX6DBN9"></script>
+  <script src="{{{static_assets.js__gtm_js}}}"></script>
+  {{/is_experimental}}
+  {{^is_experimental}}
   <script async="async" src="https://www.googletagmanager.com/gtag/js?id=UA-26406144-13"></script>
   <script src="{{{static_assets.js__gtag_js}}}"></script>
+  {{/is_experimental}}
+
   <meta charset="utf-8" />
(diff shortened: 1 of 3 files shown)


app/lib/frontend/templates/views/shared/layout.mustache
```diff
@@ -4,8 +4,15 @@
 <!DOCTYPE html>
 <html lang="en-us">
 <head>
+  {{#is_experimental}}
+  <script async="async" src="https://www.googletagmanager.com/gtm.js?id=GTM-MX6DBN9"></script>
```
isoos, line 8: Nit: maybe put only the id behind the {{#is_experimental}} condition?

static/js/gtm.js
```diff
@@ -0,0 +1,47 @@
+window.dataLayer = window.dataLayer || [];
```
isoos, line 1: We should probably put a license header here (+ on gtag.js)

static/js/gtm.js
```diff
@@ -0,0 +1,47 @@
...
+window.dataLayer.push({
```
isoos, line 2: switch the order?

```
function gtag() { dataLayer.push(arguments); }
gtag({
  'gtm.start': new Date().getTime(),
  event: 'gtm.js'
});
```

isoos (approved):
lgtm, only a few nits.

static/js/gtm.js
jonasfj: I don't think that is the same.. `arguments` behaves as an array..

isoos: /facepalm

app/lib/frontend/templates/views/shared/layout.mustache
jonasfj: It's more than the ID, it's a different script... I don't know if the contents is different, but it is a different script :D

static/js/gtm.js
jonasfj: Good idea :D

PR #4112, opened 2020-10-08
Try-catch session cookie processing.

### LINKED ISSUE #4111: Log and ignore errors when reading cookies and loading UserSession
If loading `UserSession` fails or reading cookie throws an error, we should log it and process the request as if it's not authenticated.

isoos:
Fixes #4111.

### DIFF
--- a/app/lib/shared/handler_helpers.dart
+++ b/app/lib/shared/handler_helpers.dart
@@ -68,7 +68,7 @@ shelf.Handler wrapHandler(
   handler = _userAuthWrapper(handler);
   handler =
       _requestContextWrapper(handler); // need to run after session wrapper
-  handler = _userSessionWrapper(handler);
+  handler = _userSessionWrapper(logger, handler);
   handler = _httpsWrapper(handler);
   if (sanitize) {
     handler = _sanitizeRequestWrapper(handler);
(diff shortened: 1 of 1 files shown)


app/lib/shared/handler_helpers.dart
```diff
@@ -230,14 +230,18 @@ shelf.Handler _userSessionWrapper(shelf.Handler handler) {
...
+          request.headers[HttpHeaders.cookieHeader],
+        );
+        if (sessionId != null && sessionId.isNotEmpty) {
+          final sessionData = await accountBackend.lookupSession(sessionId);
+          if (sessionData != null) {
+            registerUserSessionData(sessionData);
+          }
         }
+      } catch (e, st) {
+        logger.warning('Unable to process session cookie.', e, st);
```
jonasfj, line 244: I think this needs to be `logger.severe` or something more than `warning`.

Afaik, warning will not trigger reporting to stack driver error reporting, only `severe` and higher, see:
https://pub.dev/documentation/appengine/latest/appengine/useLoggingPackageAdaptor.html

jonasfj (changes requested):

jonasfj (approved):

PR #4125, opened 2020-10-12
pkg/web_css test to double-check if variables are used + cleanup

isoos:


jonasfj (approved):
A few more comments in `pkg/web_css/test/variables_test.dart ` wouldn't hurt... like a top-level one just explaining what it does :D

isoos:
> A few more comments in pkg/web_css/test/variables_test.dart wouldn't hurt

ack :)

PR #4135, opened 2020-10-14
Package.replacedBy on admin UI + infobox

isoos:
- Added further checks + tests: do not specify self, and do not specify another discontinued package.
- Minimal admin UI:
  Screenshot 2020-10-14 at 12 59 53

- Minimal exposure on InfoBox:
  Screenshot 2020-10-14 at 12 59 45

### DIFF
--- a/app/lib/frontend/templates/views/pkg/admin_page.mustache
+++ b/app/lib/frontend/templates/views/pkg/admin_page.mustache
@@ -51,12 +51,12 @@
 <h2>Package Options</h2>
 
 <h3>Discontinued</h3>
-<div>
+<p>
   A package can be marked as <a href="https://dart.dev/tools/pub/publishing#discontinue">discontinued</a>
   to inform users that the package is no longer maintained.
   <i>Discontinued packages</i> remain available to package users, but they don't appear
   in search results on pub.dev unless the user specifies advanced search options.
-</div>
+</p>
 <div class="mdc-form-field">
   <div class="mdc-checkbox">
     <input type="checkbox"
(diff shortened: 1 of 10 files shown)


app/lib/frontend/templates/views/pkg/admin_page.mustache
```diff
@@ -75,19 +75,36 @@
   <label for="-admin-is-discontinued-checkbox">Mark "discontinued"</label>
 </div>
 
+{{#show_replaced_by}}
+<h3>Replaced by</h3>
```
jonasfj, line 79: Can we call this `suggested replacement`

app/lib/frontend/templates/views/pkg/info_box.mustache
```diff
@@ -4,6 +4,13 @@
 
 {{& labeled_scores_html }}
 
+{{#replaced_by}}
+<h3 class="title">Replaced by</h3>
```
jonasfj, line 8: "suggested replacement" is probably a better word.

app/lib/frontend/templates/views/pkg/admin_page.mustache
```diff
@@ -75,19 +75,36 @@
...
+<div class="-pub-form-row">
```
jonasfj, line 80: We need a bit of text here along the lines of:

```markdown
When a package is <i>discontinued</i> the author may designate a <i>suggested replacement package</i>.
Package users will be suggested to consider using the designated replacement package.

Designating a replacement package is optional, and only serves to guide existing package users.
```

jonasfj (commented):
@mit-mit, any thoughts?

isoos:
Updated texts and screenshots, PTAL.

app/lib/frontend/templates/views/pkg/info_box.mustache
```diff
@@ -4,6 +4,13 @@
 
 {{& labeled_scores_html }}
 
+{{#replaced_by}}
+<h3 class="title">Suggested replacement</h3>
+<p>
+  <a href="{{& replaced_by_link }}">{{ replaced_by }}</a>
```
jonasfj, line 10: Can we give this a link-title? (tooltip).

```suggestion
  <a href="{{& replaced_by_link }}" title="This package is discontinued, but author has suggested package:{{ replaced_by }} as a replacement">{{ replaced_by }}</a>
```

jonasfj (commented):
LGTM,

mit-mit:
LGTM

jonasfj (approved):

PR #4138, opened 2020-10-14
Prepare removal of Package.doNotAdvertise flag.

isoos:


app/lib/package/models.dart
```diff
@@ -102,7 +102,9 @@ class Package extends db.ExpandoModel<String> {
 
   /// Set to `true` if package should not be advertised on the front page,
   /// may otherwise be `false`.
-  @db.BoolProperty(required: true)
+  ///
+  /// TODO: remove after `2020.10.09` is no longer an accepted runtimeVersion.
+  @db.BoolProperty()
```
jonasfj, line 107: Nice transition to `required: false` :D

Can we add `@deprecated` ?

app/lib/shared/versions.dart
```diff
@@ -21,6 +21,7 @@ final RegExp runtimeVersionPattern = RegExp(r'^\d{4}\.\d{2}\.\d{2}$');
 /// Make sure that at least two versions are kept here as the next candidates
 /// when the version switch happens.
 const acceptedRuntimeVersions = <String>[
+  // TODO: Once this version is obsolete, remove Package.doNotAdvertise
```
jonasfj, line 24: Why is this pending on runtimeVersion, isn't it more to do with deployment version? Hence, this deserves an entry in `CHANGELOG.md` that when versions older than next version doesn't exist then we can remove `doNotAdvertize`

jonasfj (approved):

app/lib/package/models.dart
isoos: `@deprecated` doesn't seem to be necessary for now, I'd rather aim for removing it soon :).

app/lib/shared/versions.dart
isoos: Updating changelog...

PR #4141, opened 2020-10-14
Fix profile image for recent sessions

jonasfj:
Currently, if I do a sign-out / sign-in, the profile image is broken.
Probably something changed in auth APIs that started inserting default parameters.
That makes sense, but it breaks when we then add additional parameters.

### DIFF
--- a/app/lib/account/models.dart
+++ b/app/lib/account/models.dart
@@ -199,7 +199,13 @@ class UserSessionData {
     }
     // Double the layout size, for better quality on higher dpi displays.
     final imageSize = layoutSize * 2;
-    return '$imageUrl=s$imageSize';
+
+    // Strip existing options from the imageUrl if there is any
+    var u = imageUrl;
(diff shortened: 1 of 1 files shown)


app/lib/account/models.dart
```diff
@@ -199,7 +199,13 @@ class UserSessionData {
     }
     // Double the layout size, for better quality on higher dpi displays.
     final imageSize = layoutSize * 2;
-    return '$imageUrl=s$imageSize';
+
+    // Strip existing options from the imageUrl if there is any
+    var u = imageUrl;
+    if (RegExp(r'=(?:[swh]\d+)|[cp](?:-(?:[swh]\d+)|[cp])*$').hasMatch(u)) {
```
isoos, line 205: Where is this regexp coming from?

jonasfj: I wrote it to match the format in https://developers.google.com/people/image-sizing

isoos: Please note the url in a comment.

jonasfj: the URL is higher up, but yes, moving it out I'll duplicate.

isoos (approved):

PR #4142, opened 2020-10-14
Add another dartdoc output pattern that doesn't need noisy logging.

isoos:


app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -508,6 +508,9 @@ bool _isKnownFailurePattern(String output) {
       output.contains('Dartdoc.logAnalysisErrors')) {
     return true;
   }
+  if (output.contains('fatal error: unable to locate the input directory at')) {
```
jonasfj, line 511: Add a comment for each of these explaining why we care about this string, and what kind of error we think it is.

jonasfj (approved):

Issue #4136, opened 2020-10-14
Add canonical-link to all pages


jonasfj:
Add `<link rel="canonical" href="https://pub.dev/path/page..." />` to all pages.

See: https://support.google.com/webmasters/answer/139066

Ankitkj1999:
I am new here. Can you please tell me where on what pages canonical URL needs to be added?

isoos:
@Ankitkj1999: Welcome :)

On the question of where: the `canonicalUrl` parameter needs to be set on the `renderLayoutPage` method: https://github.com/dart-lang/pub-dev/blob/master/app/lib/frontend/templates/layout.dart#L47

This is called from a few places, where it is not yet set, and we should add it. To have a full coverage, I'm working to extend our HTML validation to always check for the presence of it. However, as you can see on the earlier related PRs, the challenge is not adding the value itself, rather adding it in a consistent and trackable way.

PR #4144, opened 2020-10-15
Updated canonical URLs for package pages.

isoos:
- #4136
- All package pages should have canonical url.
- The share url should be the same as the canonical url.
- I've create`PkgSubPage` because I think it will be useful in other parts of the package page rendering.

### DIFF
--- a/app/lib/shared/urls.dart
+++ b/app/lib/shared/urls.dart
@@ -32,7 +32,8 @@ const _trustedTargetHost = [
 final _siteRootUri = Uri.parse('$siteRoot/');
 final _pathRootUri = Uri(path: '/');
 
-enum PkgSubPage {
+/// The list of available tabs on the package page.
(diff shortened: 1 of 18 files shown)


app/lib/shared/urls.dart
```diff
@@ -32,42 +32,63 @@ const _trustedTargetHost = [
...
 
+enum PkgSubPage {
+  readme,
+  changelog,
+  example,
+  install,
+  versions,
+  score,
+  admin,
+}
```
jonasfj, line 43: A single documentation comments would be nice :D

jonasfj (approved):

app/lib/shared/urls.dart
isoos: On top of that, I think `PkgPageTab` may be a better name, renaming...

PR #4156, opened 2020-10-19
Upgrade package:sanitize_html and use addLinkRel for <a> tags.

isoos:
- `` elements do not have `rel` attribute, removing that from the customization.
- this improves our sanitization coverage by handling the whitelisted inline `<a>` elements.

### DIFF
--- a/app/lib/shared/markdown.dart
+++ b/app/lib/shared/markdown.dart
@@ -201,12 +201,6 @@ class _UnsafeUrlFilter implements m.NodeVisitor {
       element.attributes.remove(attr);
       return true;
     }
-
-    /// <img src=""> elements are not covered by package:sanitize_html.
-    /// TODO: decide to remove from here or add it to sanitize_html.
-    if (tag == 'img' && uri.shouldIndicateUgc) {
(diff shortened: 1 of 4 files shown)


app/lib/shared/markdown.dart
```diff
@@ -194,7 +201,10 @@ class _UnsafeUrlFilter implements m.NodeVisitor {
...
       return true;
     }
-    if (uri.shouldIndicateUgc) {
+
+    /// <img src=""> elements are not covered by package:sanitize_html.
+    /// TODO: decide to remove from here or add it to sanitize_html.
+    if (tag == 'img' && uri.shouldIndicateUgc) {
       element.attributes['rel'] = 'ugc';
     }
     return false;
```
jonasfj, line 210: Do we need `rel="ugc"` on `<img` tags?

jonasfj (approved):

app/lib/shared/markdown.dart
isoos: Hm, actually, we probably shouldn't have `rel` attribute on `img` tags, as it is not a valid attribute on them. I'll update the PR.

Issue #4058, opened 2020-09-28
Login doesn't work when 3rd party cookies are disabled


jonasfj:
It seems like this is a [known issue for web application sign-in](https://developers.google.com/identity/sign-in/web/troubleshooting#known_issues).

We should consider migrating to a server-side flow.

----------

Example: When third-party cookies are disabled, login on pub.dev does not work:

![8t7BHNimNU9igwi (1)](https://user-images.githubusercontent.com/149732/94421220-88a93200-0185-11eb-9f56-cb2b7b6ff757.png)

jonasfj:
Consider:
 * Using the server-side oauth2 flow
 * Adding an endpoint that returns an XSRF token, that can be used for authentication in combination with the cookie.

PR #4160, opened 2020-10-19
User-friendly error message when sign in is not available.

isoos:
- #4058
- Detects incognito-mode third-party cookie blocking (but misses on other content blocking).
- Screenshot of the message:
  Screenshot 2020-10-19 at 15 44 25
- /cc @mit-mit for message text review

### DIFF
--- a/pkg/web_app/lib/src/account.dart
+++ b/pkg/web_app/lib/src/account.dart
@@ -94,10 +94,28 @@ void setupAccount() {
         .attributes['content'];
     if (clientId == null) return;
     load('auth2', allowInterop(() {
-      init(JsObject.jsify({'client_id': clientId}))
-          .then(allowInterop((_) => _init()));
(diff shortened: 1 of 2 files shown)


mit-mit:
Text lgtm

Do we know if there is a fix coming to the login library so that we wouldn't need these cookies?

isoos:
> Do we know if there is a fix coming to the login library so that we wouldn't need these cookies?

I don't think there is a fix for that - as long as we want to use the JS-based auth (which last time we still wanted - for being safer to use than a self-implemented auth cookie + CSRF token handling).

jonasfj (approved):

jonasfj:
> Do we know if there is a fix coming to the login library so that we wouldn't need these cookies?

It's a known issue for the library we're using. I suspect we could use a different flow, and that is something we should consider.

PR #4169, opened 2020-10-21
Fix the remaining parts of code coverage.

isoos:
- `package:coverage`'s collect now needs to be run from the directory where the `.packages` files points to, otherwise it misses the packages' own `lib/` directory. (may be a bug, will report)
- added a few formatted text in the report for easier copy-paste
- kept our all-tests-in-one-go approach, `package:test`'s `--coverage` required `-j 1` execution and took almost 25 minutes on my machine to complete.

### DIFF
--- a/app/pubspec.lock
+++ b/app/pubspec.lock
@@ -191,7 +191,7 @@ packages:
     source: hosted
     version: "2.1.1"
   coverage:
-    dependency: transitive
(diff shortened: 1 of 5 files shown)


isoos:
btw. we are at 71.8% in app/lib (7907/11015)

jonasfj (approved):

PR #4174, opened 2020-10-21
remove history end-point

jonasfj:


app/lib/frontend/handlers/pubapi.dart
```diff
@@ -304,8 +304,6 @@ class PubApi {
   /// Exposes History entities.
```
isoos, line 304: Remove the comments too.

isoos (approved):

PR #4175, opened 2020-10-21
Locally cache test-profile archives in pkg/fake_pub_server

isoos:
This is useful for local development, and it also should work with CI tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -161,4 +161,4 @@ branches:
 cache:
   directories:
     - "$HOME/.pub-cache"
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -161,3 +161,4 @@ branches:
 cache:
   directories:
     - "$HOME/.pub-cache"
+    - "pkg/fake_pub_server/.test-profile"
```
jonasfj, line 164: Can we use `.dart_tool/pub-test-profile`

jonasfj (approved):

PR #4176, opened 2020-10-21
WIP appengine upgrade

isoos:


app/lib/shared/configuration.dart
```diff
@@ -117,6 +117,9 @@ class Configuration {
   /// The identifier of admins.
   final List<AdminId> admins;
 
+  /// Whether to use metadata server to initialize auth client.
+  final bool useMetadataServer;
```
jonasfj, line 121: Why? The `clientViaApplicationDefaultCredentials()` function should work locally and remotely.

I agree, we need to configure a service account email that the signs the upload requests..
I created `package-uploader-signer@dartlang-pub-dev.iam.gserviceaccount.com` which can create GCS objects on the staging site. And I gave `dartlang-pub-dev@appspot.gserviceaccount.com` permission to impersonate `package-uploader-signer@dartlang-pub-dev.iam.gserviceaccount.com`.

We probably need to document this in comments, as it's getting complicated.

jonasfj: If running locally, humans need to be added to the list of accounts that can impersonate this service account..

We could also grant humans permission to impersonate `dartlang-pub-dev@appspot.gserviceaccount.com`, then wrap that with an impersonation that before calling `registerAuthClientService`.

isoos: > Why? The clientViaApplicationDefaultCredentials() function should work locally and remotely.

Not in tests which shouldn't try to connect to outside services, or in fake pub server, which should probably have these abstracted out.

isoos:
The PR is way out of sync now, and 80% of it has been already merged with related PRs. Closing.

PR #4179, opened 2020-10-23
HTML validation in fake pub server's responses.

isoos:
- This automatically includes the validation in integration tests.
- Added logging for missing content-type header (running locally and in integration tests).


pkg/fake_pub_server/lib/fake_pub_server.dart
```diff
@@ -84,3 +86,21 @@ class FakePubServer {
...
+    shelf.Request rq, shelf.Response rs) async {
+  if (rs == null) return null;
+
+  final ct = rs.headers[HttpHeaders.contentTypeHeader];
+  if (ct == null || ct.isEmpty) {
+    // TODO: force headers everywhere and throw exception instead of logging
+    _logger.warning('Content type header is missing for ${rq.requestedUri}.');
+  } else if (ct.contains('text/html')) {
+    final body = await rs.readAsString();
+    parseAndValidateHtml(body);
```
jonasfj, line 102: Okay, I see that doing it here ensures that it always happens, because we're doing it on the return-path from the server.

But it seems more natural to put it in the test code that is actually calling it... that way it also gets run when we do integration tests against production? Another upside with moving this check into the integration test, is that we have less customization of the fake server, which means it's closer to the real thing, and more likely to produce the same bugs.

Another option is to put it on the server, whenever we create an HTML response.. and then have it be present as an assertion... It's probably too expensive to do that :D   Then one could do an assertion that only runs when we're configured in test-mode, but that's probably how we got to this..

I could be sold on either solution, but keeping the fake server as simple as possible might be worth considering.

jonasfj (approved):

pkg/fake_pub_server/lib/fake_pub_server.dart
isoos: Hm, we could move these to the integration test's client side, if we (a) check the HTML structure as the http client returns it, and (b) intercept it in puppeteer and do the same check there too.

The tricky part is where to share the code. `pkg/pub_integration` does not depend on `app/` yet, and I think it has benefits to keep it that way (it handles it as black box, while fake_pub_server actively uses pub code). Maybe `pkg/shared_validation`?

app/lib/tool/html/html_validation.dart
```diff
@@ -0,0 +1,89 @@
...
+  validateHtml(parser.parse(html));
+}
+
+/// Validates the parsed HTML content and throws ArgumentError if any of the
+/// following issues are present:
+/// - Inline JavaScript actions are present (e.g. `onclick`).
+/// - Links with `<a target="_blank">` do not have `rel="repooner"`.
+/// - `<script> tags have no `src` attribute or have content (except `ld+json`
+///   meta content).
+void validateHtml(Node root) {
```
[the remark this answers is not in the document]
isoos: Yes, this is used in e.g. templates test, where both fragment and full-page HTML is parsed and tested. We could use on the `parseAndValidateHtml` with a bit performance impact, but I don't see why we can't have both (except for the duplication of the documentation, which is yeah, not nice...). Alternatively I could make the parameter `dynamic`, parse it if it is `String`, use it if it is `Node`. wdyt?

pkg/fake_pub_server/lib/fake_pub_server.dart
isoos: Discussed offline. I'll merge this and will move it to a new package.

Issue #4166, opened 2020-10-20
pub.dev fails to run dartdoc: "NoSuchMethodError: The getter 'documentationComment' was called on null."


sebthom:
When we try to publish our Dart library [hotreloader](https://pub.dev/packages/hotreloader), the dartdoc generation task on pub.dev fails with a dartdoc internal error message.

```dart
Unhandled exception:
NoSuchMethodError: The getter 'documentationComment' was called on null.
Receiver: null
Tried calling: documentationComment
#0      Object.noSuchMethod (dart:core-patch/object_patch.dart:51:5)
#1      Accessor.computeDocumentationComment (package:dartdoc/src/model/accessor.dart:67:44)
#2      DocumentationComment.documentationComment (package:dartdoc/src/model/documentation_comment.dart:39:31)
#3      ModelElement.computeDocumentationFrom (package:dartdoc/src/model/model_element.dart:589:9)
#4      ModelElement.documentationFrom (package:dartdoc/src/model/model_element.dart:569:28)
#5      PackageGraph.precacheLocalDocs.<anonymous closure>.precacheOneElement.<anonymous closure> (package:dartdoc/src/model/package_graph.dart:109:16)
#6      _SyncIterator.moveNext (dart:core-patch/core_patch.dart:165:25)
#7      Future.wait (dart:async/future.dart:406:26)
#8      PackageGraph.initializePackageGraph (package:dartdoc/src/model/package_graph.dart:74:18)
#9      PubPackageBuilder.buildPackageGraph (package:dartdoc/src/model/package_builder.dart:77:20)
<asynchronous suspension>
#10     Dartdoc.generateDocsBase (package:dartdoc/dartdoc.dart:182:41)
#11     Dartdoc.generateDocs (package:dartdoc/dartdoc.dart:223:34)
#12     main (file:///project/pkg/pub_dartdoc/bin/pub_dartdoc.dart:32:33)
<asynchronous suspension>
#13     _startIsolate.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:299:32)
#14     _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:168:12)
```

See https://pub.dev/packages/hotreloader/score

![image](https://user-images.githubusercontent.com/426959/96635296-f7f4ea80-131b-11eb-9d8c-8bd5bc7fb560.png)

When running dartdoc locally the documentation is generated successfully. So we cannot reproduce the issue and don't know how to prevent it.

[...]

PR #4181, opened 2020-10-23
Upgrade dartdoc

isoos:
The new version is confirmed to have the fix for #4166.


jonasfj:
I guess we could wait for https://github.com/dart-lang/dartdoc/issues/2403, since it's clear that this is a bug, and it's not a lot of packages that are affected by #4166.

jonasfj (approved):

isoos:
> I guess we could wait for dart-lang/dartdoc#2403, since it's clear that this is a bug, and it's not a lot of packages that are affected by #4166.

Ah, true, I forgot that.

Issue #4182, opened 2020-10-23
Ability to find packages with null-safety in prerelease


jonasfj:
It would be nice if during the beta-phase it was possible to find packages that has migrated to null-safety in a prerelease.
Currently, search does not index prereleases, it only indexes the latest stable release (falling back to a prerelease, if there are no stable releases).

Adding multiple versions of a package to the search index will probably increase the complexity of the search UX/UI and causing significantly more overhead.

However, the documents indexed in search already track what the latest stable and latest prerelease versions are.
Hence, it seems feasible to index the tags from the latest prerelease version with `is:<tag>-in-prerelease`.

There are minor concerns regarding the overhead of loading this data, but that's something we can investigate when implementing it.

isoos:
Launched.

PR #4184, opened 2020-10-23
Extract -in-prerelease tags for search.

isoos:
#4182

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -89,7 +89,9 @@ class SearchBackend {
     final analysisView =
         await analyzerClient.getAnalysisView(packageName, pv.version);
 
-    // Extend tags with `-in-prerelease` for prerelease version.
+    // Find tags from latest prerelease (if there one)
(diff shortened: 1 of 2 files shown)


app/lib/search/backend.dart
```diff
@@ -79,18 +80,33 @@ class SearchBackend {
...
+    final pv = await _db.lookupValue<PackageVersion>(p.latestVersionKey,
+        orElse: () => null);
     if (pv == null) {
       throw RemovedPackageException();
     }
 
     final analysisView =
         await analyzerClient.getAnalysisView(packageName, pv.version);
 
+    // Extend tags with `-in-prerelease` for prerelease version.
```
jonasfj, line 92:
```suggestion
    // Find tags from latest prerelease (if there one)
    // This allows searching for tags with `<tag>-in-prerelease`.
    // Example: `is:null-safe-in-prerelease`, or `platform:android-in-prerelease`
```
This way I think we can better remember what this bit of code does.

jonasfj (approved):

Issue #3745, opened 2020-06-24
Search UI for packages with null safety


mit-mit:
https://github.com/dart-lang/pub-dev/issues/3503 added tagging for packages with null safety. This adds a search UI option for searching only packages that have null safety.

mit-mit:
Note: We'll likely need support for searching in pre-release packages as there is how we currently think null safety versions are likely to first be published.

leafpetersen:
Is this still planned?  If so, for when?

isoos:
Launched.

PR #4190, opened 2020-10-26
Advanced search option for prerelease null-safe packages.

isoos:
- #4182
- #3745 (the same option can be used to search for latest stable with a small update)
- It is behind the experimental flag.

### DIFF
--- a/app/lib/frontend/templates/layout.dart
+++ b/app/lib/frontend/templates/layout.dart
@@ -6,6 +6,7 @@ import 'dart:convert';
 
 import 'package:client_data/page_data.dart';
 import 'package:meta/meta.dart';
+import 'package:pub_dev/frontend/request_context.dart';
 
 import '../../account/backend.dart';
 import '../../search/search_form.dart';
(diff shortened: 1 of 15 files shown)


app/lib/frontend/templates/layout.dart
```diff
@@ -187,6 +188,8 @@ String _renderSearchBanner({
     'include_discontinued': searchForm?.includeDiscontinued ?? false,
     'include_unlisted': searchForm?.includeUnlisted ?? false,
     'legacy_search_enabled': searchForm?.includeLegacy ?? false,
+    'prerelease_null_safe': requestContext.isNullSafetyDisplayed &&
```
jonasfj, line 191: `requestContext.isNullSafetyDisplayed` is still behind experimental flag, right?

jonasfj (approved):

app/lib/frontend/templates/layout.dart
isoos: yes, it is

PR #4192, opened 2020-10-26
Fix paging 404 in /api/packages

isoos:
The root cause was: we were updating the cache based on the list of packages and not on the `isLast` flag.


jonasfj (approved):
Can we add an entry in the CHANGELOG that listing API no longer returns incorrect 400s?

PR #4195, opened 2020-10-28
Manage uploaders on package admin page.

### LINKED ISSUE #3381: Manage invite/remove uploader from package admin tab
This is only relevant for packages that doesn't have a publisher.

If we implement this we can remove the `pub uploader` CLI command. 

This also reduces the API surface between `pub` and `pub.dev`, this is as it makes it more feasible to implement 3rd party pub servers. The simpler the API the better. One could easily imagine a different pub server having a vastly different way to manager user permissions.

@mit-mit, any concerns?

isoos:
- Fixes #3381.
- Screenshot 2020-10-28 at 13 32 28
- Screenshot 2020-10-28 at 13 32 36
- Screenshot 2020-10-28 at 13 32 44

### DIFF
--- a/app/lib/frontend/templates/views/pkg/admin_page.mustache
+++ b/app/lib/frontend/templates/views/pkg/admin_page.mustache
@@ -46,6 +46,61 @@
     You have no verified publisher. Use the <a href="{{& create_publisher_url}}">create publisher</a> page to create one.
   </p>
   {{/user_has_publisher}}
+  {{#show_uploader_admin}}
+  <h3>Uploaders</h3>
+  <div class="mdc-data-table">
+    <table class="mdc-data-table__table" aria-label="Uploaders of package">
+      <thead>
+        <tr class="mdc-data-table__header-row">
+          <th class="mdc-data-table__header-cell email-header" role="columnheader" scope="col">Email</th>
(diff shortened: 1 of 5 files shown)


app/lib/frontend/templates/views/pkg/admin_page.mustache
```diff
@@ -46,6 +46,61 @@
...
+        {{#uploader_emails}}
+        <tr class="mdc-data-table__row">
+          <td class="mdc-data-table__cell">{{ email }}</td>
+          <td class="mdc-data-table__cell">
+            <a class="-pub-remove-uploader-button"
+               data-email="{{ email }}"
+               title="Remove uploader">&times;</a>
+          </td>
+        </tr>
+        {{/uploader_emails}}
```
sigurdm, line 69: I wonder if we could show the current invitations?

isoos: We could probably extend `Consent` to include it as a queryable key for both this and publisher member invites, but that can come later.

jonasfj: hmm, part of this could also come should we build audit logs..

Note. we can't expose invitations sent without considering privacy implication.

jonasfj (approved):

Issue #3336, opened 2020-02-17
Make it easier for me to copy-paste the package name + version for use in pubspec.yaml


bean5:
**Expected**
When copying the package name _and_ the version, a visible/hidden `:` comes with it.

![after](https://user-images.githubusercontent.com/2052646/74676257-d632da80-5172-11ea-8f49-54f70b424676.png)

**Actual**
No helpful `:`, which means I paste it into pubspec.yaml and have to edit it.

![before](https://user-images.githubusercontent.com/2052646/74676253-d3d08080-5172-11ea-80aa-f861fc64d207.png)

isoos:
Maybe we could add `^` too, e.g. `uuid: ^2.0.4`?

bean5:
Well, I don't think we want the page to include that by default, which means we'd want a dropdown with various options for easy copy-paste. Something like Github:

![dropdown options with copy to clipboard](https://user-images.githubusercontent.com/2052646/74688645-61bd6300-5195-11ea-8607-565ca711280f.png)

Perhaps something like this, but less ugly:
![dropdown options with copy to clipboard](https://user-images.githubusercontent.com/2052646/74688900-4bfc6d80-5196-11ea-9236-7d38fae75cd1.png)

**Aside:** I think it is a bit overkill. I just wanted to be a little lazier. Not maximum lazy.

jonasfj:
I was secretly hoping we would one day find time to do a `pub add <package>` command, and insert a thing like: `$ pub add uuid` in the top right corner...

craigomac:
As requested by @bean5, I enclose the suggestion I made for this in #3564: 
![image](https://user-images.githubusercontent.com/6972172/81469521-8f8ad480-91e5-11ea-81dd-3e5bd00304e4.png)

lukas-h:
<img width="257" alt="Screenshot 2020-08-03 at 00 38 02" src="https://user-images.githubusercontent.com/10634693/89380768-613f6700-d6f8-11ea-9ead-c6af119bed38.png">

reuploading image from #3902

pedromassango:
Better to implement the suggestion on https://github.com/dart-lang/pub-dev/issues/3336#issuecomment-626133978 it is less confusing.

miquelbeltran:
I understand why there's a need from more experienced devs to have something like this, but I think that this will let to beginners to make two mistakes:

-  to use "exact" versions rather than "within" versions 
[...]

PR #4198, opened 2020-10-29
'package: ^version' copyable title pattern on package page.

isoos:
- The most requested part of #3336.

- with selection:
  Screenshot 2020-10-29 at 10 11 31

- only hover:
  Screenshot 2020-10-29 at 10 10 59

- without hover:
  Screenshot 2020-10-29 at 10 11 06

### DIFF
--- a/app/lib/frontend/templates/detail_page.dart
+++ b/app/lib/frontend/templates/detail_page.dart
@@ -27,12 +27,9 @@ String renderDetailHeader({
   /// Set true for more whitespace in the header.
   bool isLoose = false,
 }) {
-  if (title == null && titleHtml == null) {
-    throw ArgumentError('One of `title` and `titleHtml` must be specified.');
-  }
-  if (title != null && titleHtml != null) {
+  if ((title == null && titleHtml == null) || (title != null && titleHtml != null)) {
     throw ArgumentError(
-        'Only one of `title` and `titleHtml` must be specified.');
+        'Exactly one of `title` and `titleHtml` must be specified.');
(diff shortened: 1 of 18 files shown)


isoos:
The light-gray `: ^` part is displayed only on hovering, with a short transition.
/cc @mit-mit

isoos:
Alternatively we could go with a copy "clipboard" dropdown, after the title?

sigurdm:
> Alternatively we could go with a "copy to clipboard" icon + dropdown, after the title?

I think that would have much better discoverability - perhaps we could have both, so when you hover the icon, it highlights what you are going to copy.

app/lib/frontend/templates/detail_page.dart
```diff
@@ -26,9 +27,16 @@ String renderDetailHeader({
   /// Set true for more whitespace in the header.
   bool isLoose = false,
 }) {
+  if (title == null && titleHtml == null) {
+    throw ArgumentError('One of `title` and `titleHtml` must be specified.');
+  }
+  if (title != null && titleHtml != null) {
+    throw ArgumentError(
+        'Only one of `title` and `titleHtml` must be specified.');
+  }
```
sigurdm, line 36: nit
```suggestion
  if ((title == null && titleHtml == null) || (title != null && titleHtml != null)) {
    throw ArgumentError(
        'Exactly one of `title` and `titleHtml` must be specified.');
  }
```

isoos:
Closing in favour of #4201

PR #4201, opened 2020-10-30
Copy icon for "package: ^version" pattern in package title.

isoos:
- Fixes #4198.
- Extensible for future multi-value dropdown.

Screenshot 2020-10-30 at 10 57 46

### DIFF
--- a/app/lib/frontend/templates/detail_page.dart
+++ b/app/lib/frontend/templates/detail_page.dart
@@ -27,7 +27,8 @@ String renderDetailHeader({
   /// Set true for more whitespace in the header.
   bool isLoose = false,
 }) {
-  if ((title == null && titleHtml == null) || (title != null && titleHtml != null)) {
+  if ((title == null && titleHtml == null) ||
+      (title != null && titleHtml != null)) {
     throw ArgumentError(
         'Exactly one of `title` and `titleHtml` must be specified.');
   }
--- a/app/lib/frontend/templates/detail_page.dart
+++ b/app/lib/frontend/templates/detail_page.dart
@@ -15,7 +15,8 @@ final wideHeaderDetailPageClassName = '-wide-header-detail-page';
 ///
 /// The like button in the header will not be displayed when [isLiked] is null.
 String renderDetailHeader({
-  @required String title,
+  String title,
+  String titleHtml,
   String imageUrl,
   int packageLikes,
   bool isLiked,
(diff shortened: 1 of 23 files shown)


jonasfj (approved):
So the icon is always visible...

We could consider something like:
```
[data-clipboard-copy] {
    cursor: pointer;
}

[data-clipboard-copy]:hover:after {
    display: inline;
    font-size: initial;
    padding: 2px;
    content: url("https://raw.githubusercontent.com/encharm/Font-Awesome-SVG-PNG/master/black/png/16/clipboard.png") attr(data-clipboard-copy);
    border: 1px solid black;
}


<h1 class="title" data-clipboard-copy="connectivity: ^2.0.1">connectivity 2.0.1</h1>


void _setEventForPackageTitleCopyToClipboard() {
  document.querySelectorAll('data-clipboard-copy')?.forEach((elem) {
    elem.onClick.listen((e) async {
      _copyToClipboard(elem.data['clipboard-copy'].trim());

      // prevent re-adding hover style on parent
      e.stopPropagation();
    });
  });
}

void _copyToClipboard(String text) {
  final ta = TextAreaElement();
  ta.value = text;
  document.body.append(ta);
  ta.select();
  document.execCommand('copy');
  ta.remove();
}
```
And the only have a single thing that can be copied.

nitishk72:
@isoos Thanks for adding this awesome feature.
I would like to suggest some improvements

currently when I copy then `package_name ^1.0.0` is copied to the clipboard. 
I want to suggest if we add a colon after then package name `package_name: ^1.0.0` then we ca simply paste it to the `pubspec.yaml`

nitishk72:
Also when we hover we get something else and when we click on a copy it copies something else.

Text is not being copied when we click on the copy image/icon, it gets copied when we click on a different element. Different element means the element which gets shown on hovering on copy image/icon.

isoos:
> currently when I copy then package_name ^1.0.0 is copied to the clipboard.

What browsers and OS are you using? Chrome on Mac copies the `:`.

> Different element means the element which gets shown on hovering on copy image/icon.

That's intentional, as we will be displaying different options.

nitishk72:
>What browsers and OS are you using? Chrome on Mac copies the 

 `:` is not being copied now but this should be copied because in this way we can directly paste it to pubspec.yaml file. This is a feature request.

>That's intentional, as we will be displaying different options.

 oh, I thought it's a mistake but don't you think it's against human logic. We expect the copy to be work when we click on the copy button.

jonasfj:
> We expect the copy to be work when we click on the copy button.

This might be a very good observation...

@isoos could we not make clicking the button also copy first option?

isoos:
> could we not make clicking the button also copy first option?

We could, but I'd argue that it is an unexpected event. Usually copy-to-clipboard buttons make it evident what will be copied to clipboard, and don't override it unexpectedly (e.g. GitHub won't do it either). Furthermore, on a tablet, where there is no hover event, the clicking on the icon will trigger the hovering effect. Combining that with copy-to-clipboard seems to be too much to me.

nitishk72:
Every website has this as default feature, click on copy button to copy. So, now it's become a standard. We should not make new standard 😔. 

We should give chance to Google for this. Also we should not discuss more about this. 
Thanks

Issue #3711, opened 2020-06-11
Migrate to appengine 0.12.0 using ADC


jonasfj:
Moving to [appengine version 0.12.0](https://pub.dev/packages/appengine/versions/0.12.0), and update README (and other development documentation) to reflect how development should now be done using:
 * `export GOOGLE_CLOUD_PROJECT=project_id`
 * `gcloud auth application-default login`

sigurdm:
We are currently at appengine v 0.13.9

PR #4204, opened 2020-10-30
Configuration-specific initialization of storage with retrying HTTP c…

isoos:
- In preparation of #3711, taking the least controversial part out of #4176.
- I think this `if (.usesRealAppEngine)` block could also host the upload signer initialization, but I left it out for a subsequent PR, to reduce the number of moving parts that can go wrong...


app/lib/service/services.dart
```diff
@@ -44,16 +45,23 @@ import 'spam/backend.dart';
...
 Future<void> withPubServices(FutureOr<void> Function() fn) async {
   return fork(() async {
+    if (activeConfiguration.usesRealAppEngine) {
+      final authClient = await auth
+          .clientViaApplicationDefaultCredentials(scopes: [...Storage.SCOPES]);
+      final storageClient = httpRetryClient(innerClient: authClient);
+      registerStorageService(
+          Storage(storageClient, activeConfiguration.projectId));
+      registerScopeExitCallback(() async => storageClient.close());
+    }
```
jonasfj, line 63: Doesn't this need an `else` branch? to register some other `storageClient`?

Should we consider calling the configuration option something like `Configuration.useFakeServices` or? Don't we need a similar way to configure that we want to use the fake datastore, the fake email sender, the fake cloud storage, the fake signature signer?

isoos: > Doesn't this need an else branch? to register some other storageClient

In the cases where `pkg/fake_gcloud` we register a custom `storageService` and `dbService`. In such case there is no reason to call  `auth. clientViaApplicationDefaultCredentials`, and we don't have HTTP client either to wrap in retrying HTTP client.

jonasfj (approved):

PR #4212, opened 2020-11-03
Detect simple redirect loops.

isoos:


app/lib/shared/handler_helpers.dart
```diff
@@ -76,6 +77,50 @@ shelf.Handler wrapHandler(
...
+    final rs = await handler(request);
+    if (rs.statusCode >= 300 && rs.statusCode < 400) {
+      final location = rs.headers[HttpHeaders.locationHeader];
+
+      // sanity check for the header being present
+      if (location == null) {
+        logger.shout('Redirect response without location header.',
+            Exception('Redirect response without location header.'));
+        return rs;
+      }
```
jonasfj, line 93: Throw a 500 here, there is no reason to 303 when it clearly won't work.

app/lib/shared/handler_helpers.dart
```diff
@@ -76,6 +77,50 @@ shelf.Handler wrapHandler(
...
+
+      // sanity check for the header being valid
+      final uri = Uri.tryParse(location);
+      if (uri == null) {
+        logger.shout(
+            'Redirect response with invalid location header: "$location".',
+            Exception(
+                'Redirect response with invalid location header: "$location".'));
+        return rs;
```
jonasfj, line 102: Same here, let's make this a 500

app/lib/shared/handler_helpers.dart
```diff
@@ -76,6 +77,50 @@ shelf.Handler wrapHandler(
...
+      }
+
+      // path + querystring match
+      if (uri.toString().startsWith(uri.path) &&
+          request.requestedUri.toString().endsWith(uri.toString())) {
+        logger.shout(
+            'Redirect loop detected.', Exception('Redirect loop detected.'));
+        return rs;
+      }
```
jonasfj, line 118: This is if you do `pub.dev/path -> /path` ?

isoos: Yes, also matching it with `?queryString`.

jonasfj (approved):

app/lib/shared/handler_helpers.dart
```diff
@@ -63,6 +63,7 @@ shelf.Handler wrapHandler(
   // Handlers wrap other handlers, and they are called in the revers order of
   // their wrapping. Read this list from the bottom and go up to get the real
   // execution order.
+  handler = _redirectLoopDetectorWrapper(logger, handler);
   handler = _cspHeaderWrapper(handler);
   handler = _userAuthWrapper(handler);
   handler =
```
jonasfj, line 69: @isoos 
unrelated nit: But should we perhaps use [shelf.Pipeline](https://pub.dev/documentation/shelf/latest/shelf/Pipeline-class.html) here?

isoos: Yeah, we are at the point where it would make sense :)

PR #4234, opened 2020-11-10
Fix SDK links on Flutter Favorites listing page.

### LINKED ISSUE #4226: Flutter favorit page navigation bug
URL: https://pub.dev/flutter/favorites?page=3

after going to this link when we try to navigate to dart or flutter specific page in navigation of dart | flutter | any in all ancher link it is staying the same.

isoos:
- Fixes #4226.
- On a Flutter favorite page `Any` doesn't become active.
- With the SDK change there is also a reset to the page number.
- Non-active SDK tabs do have a transparent bottom border.

### DIFF
--- a/app/lib/search/search_form.dart
+++ b/app/lib/search/search_form.dart
@@ -126,7 +126,9 @@ class SearchForm {
   }) {
     if (sdk != null) {
       tagsPredicate ??= this.tagsPredicate ?? TagsPredicate();
-      tagsPredicate = tagsPredicate.removePrefix('sdk:');
+      tagsPredicate = tagsPredicate
+          .removePrefix('sdk:')
+          .withoutTag(PackageTags.isFlutterFavorite);
       if (SdkTagValue.isNotAny(sdk)) {
         tagsPredicate = tagsPredicate
             .appendPredicate(TagsPredicate(requiredTags: ['sdk:$sdk']));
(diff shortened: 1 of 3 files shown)


app/lib/search/search_form.dart
```diff
@@ -188,14 +190,18 @@ class SearchForm {
     return values.isEmpty ? null : values.first;
   }
 
+  /// Returns whether the search is on Flutter Favorites.
+  bool get isFlutterFavorite =>
+      tagsPredicate.isRequiredTag(PackageTags.isFlutterFavorite);
```
jonasfj, line 195: This doesn't affect when users type-in `is:flutter-favorite` in the search query, right?

`TagPredicate` is more of a `SearchContext` really, right? It's the context within which search is presented, is it not?

Just trying to understand, not saying we have to refactor now, just saying wow... there is a lot of implicit knowledge here that is hard to follow (was also hard to follow when I did FF listing and made it worse).

jonasfj (commented):
Will the tab bar not be rendered anymore, or what happens when `any` isn't selected?

app/lib/search/search_form.dart
isoos: > This doesn't affect when users type-in is:flutter-favorite in the search query, right?

Correct.

> TagPredicate is more of a SearchContext really, right? It's the context within which search is presented, is it not?

Yes, we should probably move this toward the `Place` / `Page` concept that I've talk about recently.

isoos:
> Will the tab bar not be rendered anymore, or what happens when any isn't selected?

It is rendered, but no longer active by default. If you click on it, it will go to the `/packages` page.

jonasfj (approved):

PR #4242, opened 2020-11-11
Migrate to assets: render atom feed without PackageVersion.readme.

isoos:
Also added caching to the feed content.

### DIFF
--- a/app/lib/shared/redis_cache.dart
+++ b/app/lib/shared/redis_cache.dart
@@ -232,6 +232,11 @@ class CachePatterns {
         encode: (PublisherPage list) => list.toJson(),
         decode: (data) => PublisherPage.fromJson(data as Map<String, dynamic>),
       ))['$userId'];
+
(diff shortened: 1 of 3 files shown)


app/lib/shared/redis_cache.dart
```diff
@@ -232,6 +232,11 @@ class CachePatterns {
         encode: (PublisherPage list) => list.toJson(),
         decode: (data) => PublisherPage.fromJson(data as Map<String, dynamic>),
       ))['$userId'];
+
+  Entry<String> atomFeedXml() => _cache
```
jonasfj, line 236: I'm guessing that caching is also a new thing :D

jonasfj (approved):

app/lib/shared/redis_cache.dart
isoos: Ah, yeah, slightly related :)

PR #4243, opened 2020-11-12
Use pana's ToolEnvironment in test-profile resolution.

isoos:


app/lib/tool/test_profile/resolver.dart
```diff
@@ -31,17 +39,13 @@ Future<List<ResolvedVersion>> resolveVersions(TestProfile profile) async {
...
 
-        final pr = await Process.run(
-          'pub',
-          ['get'],
-          environment: {
-            'PUB_CACHE': pubCacheDir.path,
-          },
-          workingDirectory: dummyDir.path,
-        );
+        final pr = await toolEnv.runUpgrade(dummyDir.path, false);
```
jonasfj, line 48: Given how simple this is, I'm not sure this is a great idea... is there any win here?

Does it help us with `flutter` support? or...

Just saying: `dart pub get` is the stable interface for making resolutions, wrapping it with a library might be overkill.

jonasfj (approved):
I have no problem with this, but does it really give us much... Other than making it rather to refactor `pana`?

app/lib/tool/test_profile/resolver.dart
isoos: It gives us Flutter support + gets us the `runtimeInfo.sdkVersion`. We could re-implement it slightly simpler than `pana`, but at the same time reuse is free.

As for the analysis the packages are in a clean state, there shouldn't be much difference between calling `pub get` or `pub upgrade`.

PR #4246, opened 2020-11-12
Search backend: read readme from asset.

isoos:
This is part of the asset migration.

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -12,17 +12,18 @@ import 'package:json_annotation/json_annotation.dart';
 import 'package:logging/logging.dart';
 
 import 'package:pub_dartdoc_data/pub_dartdoc_data.dart';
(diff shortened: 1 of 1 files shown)


jonasfj (approved):

app/lib/search/backend.dart
```diff
@@ -135,7 +138,7 @@ class SearchBackend {
       description: compactDescription(pv.pubspec.description),
       created: p.created,
       updated: pv.created,
-      readme: compactReadme(pv.readmeContent),
+      readme: compactReadme(readmeAsset?.textContent),
```
jonasfj, line 141: Should we bump runtime version? or something else that'll cause a new version of the search index and snapshot?

isoos: It is already bumped because of other changes, but it shouldn't matter: the values must be the same.

PR #4248, opened 2020-11-13
Configurable import source for test-profile.

isoos:
- `PubDevImportSource` for importing package versions and archive from pub.dev. It will be used in local development and in integration tests.
- `SemiRandomImportSource` for using a semi-random process to generate data in memory, without network access. It will be used in unit tests (with probably a bit more scope and variance).


app/lib/tool/test_profile/import_source.dart
```diff
@@ -0,0 +1,121 @@
...
+
+  /// Gets the archive bytes for [package]-[version].
+  Future<List<int>> getArchiveBytes(String package, String version);
+
+  /// Close resources that were opened during the sourcing of data.
+  Future<void> close();
+}
+
+/// Resolves and downloads data from pub.dev.
+class PubDevImportSource implements ImportSource {
```
jonasfj, line 32: Any reason to expose this type? Can't we make it private...

And have the constructors exposed as static function on `ImportSource`, like:
```dart
abstract class ImportSource {
  ...

  static ImportSource fromPubDev({@required String archiveCachePath}) {
    return _PubDevImportSource(archiveCachePath: archiveCachePath);
  }
}
```

And similarly for `SemiRandomImportSource`

jonasfj (approved):

app/lib/tool/test_profile/import_source.dart
isoos: Done.

app/lib/tool/test_profile/import_source.dart
```diff
@@ -0,0 +1,134 @@
...
+  /// Creates a source that resolves and downloads data from pub.dev.
+  static ImportSource fromPubDev({
+    @required String archiveCachePath,
+  }) {
+    return _PubDevImportSource(archiveCachePath: archiveCachePath);
+  }
+
+  /// Creates a source that generates data based on random seed, without any
+  /// network (or file) access.
+  static ImportSource semiRandom() {
```
jonasfj, line 39: Wouldn't it be nice to all the `pseudoRandom` or `arbitrary` or  `generatedTestData` or `autoGenerated` or...

PR #4253, opened 2020-11-17
Reduce search-page cache timeout.

isoos:


jonasfj (approved):
I'm guessing this aims to reduce the window for inconsistencies...

PR #4271, opened 2020-11-20
Remove visible search option for legacy packages.

isoos:


jonasfj (approved):

doc/help-search.md
```diff
@@ -36,3 +36,16 @@ The search UI also supports filters:
...
 specificity and the text match score of the search team.
+
+## Legacy packages
+
+`pub.dev` does not expose packages that support only Dart SDK 1.x in search results
+or in listing pages. One can access these packages through their package page
+(`/packages/<pkg>`) and `pub` client can download them.
+
+You can use the following options to search for these packages:
+- Add `legacy=1` query parameter to the URL ([example](https://pub.dev/packages?legacy=1)).
```
jonasfj, line 47: I suggest we don't document and remove the query parameter... I know that just leaves `is:legacy`, but IMO that's fine..

isoos: `is:legacy` doesn't work on itself - not sure why, need to investigate it.

jonasfj: Isn't it better to just file a bug for that and fix it independently, I don't think it's super-bad if that is temporarily broken.

isoos: Filed https://github.com/dart-lang/pub-dev/issues/4275

PR #4280, opened 2020-11-23
Basic tool to check search drift.

isoos:


app/bin/tools/check_drift.dart
```diff
@@ -0,0 +1,259 @@
...
+    tagsPredicate: TagsPredicate(requiredTags: [PackageTags.isFlutterFavorite]),
+  ),
+  SearchForm.parse(query: 'json'),
+  SearchForm.parse(sdk: 'dart', currentPage: 2),
+  SearchForm.parse(sdk: 'flutter', currentPage: 5),
+  SearchForm.parse(order: SearchOrder.created),
+  SearchForm.parse(order: SearchOrder.updated, currentPage: 3),
+];
+
+Future<void> main() async {
```
jonasfj, line 26: A `--help` option wouldn't hurt... and a few words on what this does.. Notably, it worth saying that it WON'T use any form of administrative access or modify server-side resources. It only does a series of `GET` requests, right?

jonasfj (approved):

PR #4281, opened 2020-11-24
Override the uploaded version's created timestamp in test profile.

isoos:
- a first step to make the test profile's database consistent
- `pub.dev` created timestamp are fetched via the `/api/packages/<pkg>/versions/<version>` endpoint
- auto-generated timestamps are calculated backwards from the referenced timestamp (which defaults to now)


jonasfj (approved):

isoos:
Closing this without merging, as offline discussion points to a different solution.

PR #4283, opened 2020-11-24
Rely less on pana's PackageDependency list.

isoos:


app/lib/search/backend.dart
```diff
@@ -154,19 +156,25 @@ class SearchBackend {
...
     final Map<String, String> dependencies = <String, String>{};
     view.allDependencies?.forEach((pd) {
       dependencies[pd.package] = pd.dependencyType;
     });
+    pubspec.dependencies.forEach((package) {
+      dependencies[package] = DependencyTypes.direct;
+    });
+    pubspec.devDependencies.forEach((package) {
+      dependencies[package] = DependencyTypes.dev;
+    });
```
jonasfj, line 177: Might you not want to re-order these? in case a package is both a dependency and a dev_dependency?

jonasfj (approved):

PR #4286, opened 2020-11-25
Consistent but somewhat random-looking ordering for test-profile package versions.

isoos:
I think it is important to have a random-looking order, otherwise tests that order by name or by updated times will be looking exactly the same, unable to differentiate between them.

### DIFF
--- a/app/lib/tool/test_profile/importer.dart
+++ b/app/lib/tool/test_profile/importer.dart
@@ -75,10 +75,13 @@ Future<void> importProfile({
   }
 
   // create versions
-  for (final testPackage in profile.packages) {
+  final orderedPackages =
+      _sortByShaHash<TestPackage>(profile.packages, (tp) => tp.name);
+  for (final testPackage in orderedPackages) {
     final packageName = testPackage.name;
     User lastActiveUser;
-    for (final versionName in testPackage.versions) {
+    final orderedVersions = _sortByShaHash(testPackage.versions);
+    for (final versionName in orderedVersions) {
       // figure out the active user
       final uploaderEmails = _potentialActiveEmails(profile, packageName);
       final uploaderEmail =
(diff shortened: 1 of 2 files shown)


app/lib/tool/test_profile/importer.dart
```diff
@@ -137,6 +140,21 @@ Future<void> importProfile({
   await source.close();
 }
 
+/// Sorts items in the order of their SHA1 hash, to ensure a deterministic, but
+/// random-looking time order for the entities inserted.
```
jonasfj, line 144:
```suggestion
/// arbitrary-looking time order for the entities inserted.
```

app/lib/tool/test_profile/importer.dart
```diff
@@ -75,10 +75,13 @@ Future<void> importProfile({
...
 
   // create versions
-  for (final testPackage in profile.packages) {
+  final orderedPackages =
+      _sortByShaHash<TestPackage>(profile.packages, (tp) => tp.name);
+  for (final testPackage in orderedPackages) {
     final packageName = testPackage.name;
     User lastActiveUser;
-    for (final versionName in testPackage.versions) {
+    final orderedVersions = _sortByShaHash(testPackage.versions);
```
jonasfj, line 83: Do we need to create versions in arbitrary order? that seems weird.

Wouldn't those normally be created in the natural order? Is there any reason, not to create them in the order they appear in the test-profile?

app/lib/tool/test_profile/importer.dart
```diff
@@ -75,10 +75,13 @@ Future<void> importProfile({
   }
 
   // create versions
```
jonasfj, line 77:
```suggestion
  // Create packages and package versions
```

Right?

app/lib/tool/test_profile/importer.dart
```diff
@@ -75,10 +75,13 @@ Future<void> importProfile({
...
-  for (final testPackage in profile.packages) {
+  final orderedPackages =
+      _sortByShaHash<TestPackage>(profile.packages, (tp) => tp.name);
+  for (final testPackage in orderedPackages) {
```
jonasfj, line 80: Why not create them in the order that they have in the test-profile?

Then the order is given when create the test-profile, or the thing that derives the test-profile can derive the order from when the packages were published? So that we have the same ordering as they have on the actual pub site?

isoos:
Offline discussion concluded that this is not the direction we want to take.

PR #4290, opened 2020-11-25
Refactor UploadSignerService registration

isoos:
- Part of #3711.
- Simplified the setup code using the authenticated client already present for the storage service override.
- Moved `fake_upload_signer_service.dart` into `lib/fake`, other fake-pub-server code will move there too.
- Removed duplicate registration for fake implementation.


app/lib/package/upload_signer_service.dart
```diff
@@ -28,6 +28,21 @@ UploadSignerService get uploadSigner =>
 void registerUploadSigner(UploadSignerService uploadSigner) =>
     ss.register(#_url_signer, uploadSigner);
 
+/// Creates an upload signer based on the current environment.
+Future<UploadSignerService> createUploadSigner(http.Client authClient) async {
+  if (envConfig.isRunningLocally) {
+    return ServiceAccountBasedUploadSigner();
```
jonasfj, line 34: Can we make `ServiceAccountBasedUploadSigner` internal or `@visibleForTestingOnly`? Similar for `IamBasedUploadSigner`

app/lib/package/upload_signer_service.dart
```diff
@@ -28,6 +28,21 @@ UploadSignerService get uploadSigner =>
 void registerUploadSigner(UploadSignerService uploadSigner) =>
     ss.register(#_url_signer, uploadSigner);
 
+/// Creates an upload signer based on the current environment.
+Future<UploadSignerService> createUploadSigner(http.Client authClient) async {
```
jonasfj, line 32: Should we change pattern and just call this: `setupUploadSigner` and have it call `registerUploadSigner` internally, instead of returning it?

Then we can make `registerUploadSigner` private.. perhaps that's one way to encapsulate things a bit.

jonasfj (approved):
just nits,

app/lib/package/upload_signer_service.dart
isoos: Done.

app/lib/package/upload_signer_service.dart
isoos: We also register the `FakeUploadSignerService` in the case of tests and fake pub server. As we move/integrate the fake pub server into the app project, I think we may have better patterns later on...

PR #4291, opened 2020-11-25
Move contentType header check outside of fake-pub-server.

isoos:
Integration tests will verify it on each request instead (also against the production server in post-deployment test).


jonasfj (approved):
We could also make an assertion for this in production... And just log something or use `assert()` to make it only fail in testing.

isoos:
> We could also make an assertion for this in production...

With this PR we will check this in `pkg/pub_integration` too, which will be used in the post-deployment verification. I don't think we'll need to check for this for every response in production.

PR #4293, opened 2020-11-25
Either canonical URL or noindex must be present in the headers.

### LINKED ISSUE #4136: Add canonical-link to all pages

isoos:
Fixes #4136.

### DIFF
--- a/pkg/pub_validations/lib/html/html_validation.dart
+++ b/pkg/pub_validations/lib/html/html_validation.dart
@@ -37,7 +37,7 @@ void validateHtml(Node root) {
     links = root.querySelectorAll('a');
     scripts = root.querySelectorAll('script');
   } else if (root is Document) {
-    _validateHead(root.querySelector('head'));
+    _validateCanonicalLink(root.querySelector('head'));
     elements = root.querySelectorAll('*');
     links = root.querySelectorAll('a');
     scripts = root.querySelectorAll('script');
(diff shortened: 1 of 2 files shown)


pkg/pub_validations/lib/html/html_validation.dart
```diff
@@ -101,6 +101,18 @@ void _validateHead(Element head) {
...
+  if (canonicalLinks.isEmpty) {
+    final robotsValues = head
+        .querySelectorAll('meta')
+        .where((elem) => elem.attributes['name'] == 'robots')
+        .map((elem) => elem.attributes['content'])
+        .expand((v) => v.split(' '))
+        .toSet();
+    if (!robotsValues.contains('noindex')) {
+      throw ArgumentError(
+          'When canonical URL is missing, noindex must be set.');
```
jonasfj, line 113: nit: `ArgumentError` is probably the wrong thing to throw an assertion fails.. maybe consider a custom exception for failed validation or just use `AssertionError` that's pretty close to a validation error...

pkg/pub_validations/lib/html/html_validation.dart
```diff
@@ -101,6 +101,18 @@ void _validateHead(Element head) {
   if (canonicalLinks.length > 1) {
     throw ArgumentError('More than one canonical link was specified.');
   }
+  if (canonicalLinks.isEmpty) {
```
jonasfj, line 104: nit: Maybe we should add a comment explaining: 
 (A) what is a canonical link,
 (B) What is a noindex meta tag,
 (C) Why should one of them be present

I don't disagree that it's reasonable, but it might be nice to explain future readers.

jonasfj (approved):

Issue #4113, opened 2020-10-08
Better handling of Package.updated


isoos:
`Package.updated` was originally updated only when a new version was published. However, we also update it if `publisherId`, `uploaders`, `isDiscontinued`, `isUnlisted` or other properties are updated. With that, the APIs listing by `updated` are not reflecting the true status of the last published state.

PR #4297, opened 2020-11-26
Tracking version published timestamps in Package.

isoos:
- #4113
- added backfill
- fields are updated the same way the key references are updated

### DIFF
--- a/app/lib/package/models.dart
+++ b/app/lib/package/models.dart
@@ -66,8 +66,14 @@ class Package extends db.ExpandoModel<String> {
   @db.ModelKeyProperty(propertyName: 'latest_version', required: true)
   db.Key latestVersionKey;
 
+  @db.DateTimeProperty()
+  DateTime latestPublished;
+
   @db.ModelKeyProperty(propertyName: 'latest_dev_version')
   db.Key latestPrereleaseVersionKey;
+
+  @db.DateTimeProperty()
+  DateTime latestPrereleasePublished;
 
   /// The publisher id (null, if the package does not have a publisher).
   @db.StringProperty()
@@ -179,12 +185,14 @@ class Package extends db.ExpandoModel<String> {
     if (latestVersionKey == null ||
         isNewer(latestSemanticVersion, newVersion, pubSorted: true)) {
       latestVersionKey = pv.key;
+      latestPublished = pv.created;
     }
 
     if (latestPrereleaseVersionKey == null ||
         isNewer(latestPrereleaseSemanticVersion, newVersion,
             pubSorted: false)) {
       latestPrereleaseVersionKey = pv.key;
+      latestPrereleasePublished = pv.created;
     }
   }
 
--- a/app/lib/package/models.dart
+++ b/app/lib/package/models.dart
@@ -63,15 +63,24 @@ class Package extends db.ExpandoModel<String> {
   @db.IntProperty(required: true)
   int likes;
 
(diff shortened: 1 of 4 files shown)


app/lib/package/models.dart
```diff
@@ -66,9 +66,15 @@ class Package extends db.ExpandoModel<String> {
...
   db.Key latestVersionKey;
 
+  @db.DateTimeProperty()
+  DateTime latestPublished;
+
   @db.ModelKeyProperty(propertyName: 'latest_dev_version')
   db.Key latestPrereleaseVersionKey;
 
+  @db.DateTimeProperty()
+  DateTime latestPrereleasePublished;
```
jonasfj, line 76:
```suggestion
  /// DateTime at which point the `PackageVersion` referenced in [latestPrereleaseVersionKey] was published.
  @db.DateTimeProperty()
  DateTime latestPrereleasePublished;
```

app/lib/package/models.dart
```diff
@@ -66,9 +66,15 @@ class Package extends db.ExpandoModel<String> {
   @db.ModelKeyProperty(propertyName: 'latest_version', required: true)
   db.Key latestVersionKey;
 
+  @db.DateTimeProperty()
+  DateTime latestPublished;
+
   @db.ModelKeyProperty(propertyName: 'latest_dev_version')
   db.Key latestPrereleaseVersionKey;
```
jonasfj, line 73:
```suggestion
  /// Reference to latest version of this package ordered by semantic versioning, including
  /// prerelease versions of this package.
  ///
  /// Note. This is **not** the _latest prerelease version_, it is the _latest version_ without deprioritization of
  /// prereleases. Hence, this might not be a prerelease version, if a newer non-prerelease exists.
  @db.ModelKeyProperty(propertyName: 'latest_dev_version')
  db.Key latestPrereleaseVersionKey;
```
Is this correct? I think this is worth documenting, because the property is NOT what it says :D

Also can this be `null`? We should document this.

app/lib/package/models.dart
```diff
@@ -66,9 +66,15 @@ class Package extends db.ExpandoModel<String> {
   @db.ModelKeyProperty(propertyName: 'latest_version', required: true)
   db.Key latestVersionKey;
 
+  @db.DateTimeProperty()
+  DateTime latestPublished;
```
jonasfj, line 70:
```suggestion
  /// [DateTime] when the [PackageVersion] in [latestVersionKey] was published.
  @db.DateTimeProperty()
  DateTime latestPublished;
```

app/lib/package/models.dart
```diff
@@ -66,9 +66,15 @@ class Package extends db.ExpandoModel<String> {
   @db.ModelKeyProperty(propertyName: 'latest_version', required: true)
   db.Key latestVersionKey;
```
jonasfj, line 67:
```suggestion
  /// Key referencing the [PackageVersion] for the latest version of this package, ordered by priority order of
  /// semantic versioning, hence, deprioritizing prereleases.
  @db.ModelKeyProperty(propertyName: 'latest_version', required: true)
  db.Key latestVersionKey;
```

jonasfj (approved):

app/lib/package/models.dart
isoos: The update is correct, thanks! Maybe we could migrate away from the old notion of storing the key on the entity, and with that also rename it?

I think it cannot be `null`, and we could probably mark it as such.

PR #4301, opened 2020-11-27
Deprecating PanaReport.pkgDependencies

isoos:
- Follow-up to #4283.
- Bumped runtime version, as this is a backward-incompatible change in the stored pana reports.


sigurdm (approved):
LGTM

Issue #4302, opened 2020-11-27
/api/dart2-package-name-completion-data


jonasfj:
```
GET /api/dart2-package-name-completion-data
{
  "packages": [
    ..., "retry", ... // ordered by popularity * likes * pub points
  ],
}
```
We filter out packages that:
 * only supports Dart 1.x
 * is flagged discontinued,
 * is flagged unlisted.

Results ordered by some combination of `popularity`, `likes` and `pub points`.
We do not formally promise what ordering the API has, nor promise that the ordering won't change.

Nor do we make the API paging, if there are more packages than we can list in some future, then we return the `N` most important packages ranked by popularity, likes and pub points; with `N` being a high number something like 10k or 100k.

Note:
 * it's important that we cache this response. I propose caching the gzipped response such that we can serve a large number of these requests rather efficiently.
 * We should ensure that we support gzip for this request, and either reject requests that do not have the header `accept-encoding: gzip`, or support serving without gzip too.
 * We could store the value of this in GCS and redirect to a public bucket, this would indeed be a very reasonable thing to do.
 * I'm open to better path suggestions than `/api/dart2-package-name-completion-data`, but I think it's worthwhile making sure that the API end-point is specific to:
   * Clients who want to do code-completion for package names,
   * Dart 2.x clients -- because Dart 1.x packages are irrelevant, and when one day Dart 3 becomes a thing we might want to use a new end-point for this (though I can be dissuaded from this argument).

jonasfj:
I could also be convinced we should call this something like:
 * `/api/package-name-completion`
 * `/api/package-names/completion-data`
 * `/api/completion/package-names-dart2`
 * `/api/package-name-completion`

sigurdm:
I don't think we need `dart2` in the name.

We know nothing about how dart3 will be - let's plan for it when we know more.

If a new endpoint is really  needed we can always call the new endpoint something with `dart3`.

`/api/package-name-completion` sounds good to me.

isoos:
> I don't think we need dart2 in the name.

+1

> /api/package-name-completion

I'd expect such endpoint to complete a package name that I provide via a query parameter. Anything against `/api/package-names/completion-data` or `/api/package-name-completion-data` ?

sigurdm:
> Anything against /api/package-names/completion-data or /api/package-name-completion-data ?

Nope - good point!

DanTup:
[...]

PR #4304, opened 2020-11-30
API for name completion data.

isoos:
- #4302
- Tested search responses on staging: 400-600 ms on the search instance to serve the content, a 200-500 ms more on the frontend instance to parse, transform and cache it. If we were to use a binary protocol for search, this could be reduced by about 50-150 ms.
- The uncompressed size is about 280k, the compressed is about 80k with the default gzip settings.
- Caching the gzipped content for 8 hours seems to be a good balance.
- Clients that do not accept gzip encoding are rejected with HTTP 406.
- Added to integration test (also checks on post-deployment) - as long as `retry` is a popular package, it won't fail :).

### DIFF
--- a/app/lib/frontend/handlers/custom_api.dart
+++ b/app/lib/frontend/handlers/custom_api.dart
@@ -3,7 +3,6 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:convert';
 import 'dart:io';
 
 import 'package:client_data/package_api.dart';
@@ -23,6 +22,7 @@ import '../../shared/exceptions.dart';
 import '../../shared/handlers.dart';
 import '../../shared/redis_cache.dart' show cache;
 import '../../shared/urls.dart' as urls;
+import '../../shared/utils.dart' show jsonUtf8Encoder;
 
 /// Handles requests for /api/documentation/<package>
 Future<shelf.Response> apiDocumentationHandler(
@@ -113,9 +113,9 @@ Future<shelf.Response> apiPackageNameCompletionDataHandler(
       skipCache: true,
     );
 
-    return gzip.encode(utf8.encode(json.encode({
+    return gzip.encode(jsonUtf8Encoder.convert({
       'packages': rs.packages.map((p) => p.package).toList(),
-    })));
+    }));
   });
 
   return shelf.Response(200, body: bytes, headers: {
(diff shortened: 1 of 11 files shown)


jonasfj:
>  200-500 ms more on the frontend instance to parse

Is that CPU time on the main thread of the frontend instance?

That'll make a lot of other requests much slower.. Should we consider making a `?compact=1` mode for search backend, where it only returns package names?

Maybe we should forbid `limit > 100` unless `compact=1` on the search backend?

isoos:
> Is that CPU time on the main thread of the frontend instance?

It was local benchmark.

> Should we consider making a ?compact=1 mode for search backend, where it only returns package names?

Good point, however, the result format would be somewhat different: instead of `"packages": [{"package": "<pkg>"}, {...}, ...]` we would have `"packageNames": ["<pkg>", "<pkg>", ]). I'd add that in a separate PR.

> Maybe we should forbid limit > 100 unless compact=1 on the search backend?

Not sure if this is necessary...

app/lib/frontend/handlers/custom_api.dart
```diff
@@ -90,6 +92,40 @@ Future<shelf.Response> apiPackageNamesHandler(shelf.Request request) async {
...
+/// Handles requests for
+/// - /api/package-name-completion-data
+Future<shelf.Response> apiPackageNameCompletionDataHandler(
+    shelf.Request request) async {
+  // only accept requests which allow gzip content encoding
+  final acceptsEncoding =
+      request.headers[HttpHeaders.acceptEncodingHeader] ?? '*';
+  if (!acceptsEncoding.contains('*') && !acceptsEncoding.contains('gzip')) {
+    throw NotAcceptableException('Client must accept gzip content.');
+  }
```
jonasfj, line 104: nit: What about an extension method on `Request.acceptEncoding(String) -> bool`

app/lib/frontend/handlers/custom_api.dart
```diff
@@ -90,6 +92,40 @@ Future<shelf.Response> apiPackageNamesHandler(shelf.Request request) async {
...
+      ServiceSearchQuery.parse(
+        tagsPredicate: TagsPredicate.regularSearch(),
+        limit: 20000,
+      ),
+      // Do not cache response at the search client level, as we'll be caching
+      // it in a processed form much longer.
+      skipCache: true,
+    );
+
+    return gzip.encode(utf8.encode(json.encode({
```
jonasfj, line 117: We should fuse these :D

jonasfj (approved):

PR #4309, opened 2020-12-01
Preconfigured upload signer service account (for staging).

isoos:
- #3711
- Verified on staging.

### DIFF
--- a/app/lib/shared/configuration.dart
+++ b/app/lib/shared/configuration.dart
@@ -101,9 +101,9 @@ class Configuration {
   final String gmailRelayImpersonatedGSuiteUser;
 
   /// The email of the service account which has access rights to sign upload
-  /// requests.
+  /// requests. The current service must be able to impersonate this account.
   ///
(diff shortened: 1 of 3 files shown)


app/lib/shared/configuration.dart
```diff
@@ -100,6 +100,15 @@ class Configuration {
   /// **Optional**, if omitted email sending is disabled.
   final String gmailRelayImpersonatedGSuiteUser;
 
+  /// The email of the service account which has access rights to sign upload
+  /// requests.
+  ///
+  /// Authorization requires the following IAM permission on the specified resource name:
+  /// - iam.serviceAccounts.signBlob
```
jonasfj, line 107: Might be worth mentioning that this is on the package bucket..

jonasfj (approved):

app/lib/shared/configuration.dart
jonasfj: and importantly that the service account running this service MUST be able to impersonate this account.

PR #4312, opened 2020-12-02
New copy to clipboard flow.

### LINKED ISSUE #4244: Show a copied message feedback after the copy button was pressed
The newly added feature (https://github.com/dart-lang/pub-dev/issues/3336) in pub.dev is awesome but it is quite strange to me when I tap a button and  don't get any visual feedback.

### Proposal

I propose to show a copied feedback message once the copy button is pressed. It is quite strange to me when I tap a button and  don't get any visual feedback.

### Suggestion

The message shown when we place the cursor over the copy button is quite useless as it just shows the package name and version which is already visible. I would remove it. I know it shows package and version in a format that is acceptable by `pubspec.yaml` but it still useless since we usually know that.

isoos:
- Fixes #4244.
- Adds browser test.

Screen Recording 2020-12-02 at 17 49 14

### DIFF
--- a/app/lib/frontend/templates/package.dart
+++ b/app/lib/frontend/templates/package.dart
@@ -243,9 +243,6 @@ String _renderPkgTitleContent(PackagePageData data) {
   return templateCache.renderTemplate('pkg/title_content', {
     'package': data.package.name,
     'version': data.version.version,
-    'show_prerelease':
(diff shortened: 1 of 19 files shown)


isoos:
The fadeout is faster than on the .gif, it seems as if it was slowed down during conversion... I'll deploy this to staging.

jonasfj (approved):

PR #4324, opened 2020-12-08
Audit entity model + backfill for package published events.

isoos:
- adding `AuditLogRecord` and specific constructor for the publish event
- separate bin tool + backfill library + tests
- also removing `bin/tools/backfill_history.dart`, as it is not going to happen


app/lib/audit/models.dart
```diff
@@ -0,0 +1,68 @@
...
+        'version `$version` '
+        'was published '
+        'by `${uploader.email}`.';
+    return AuditLogRecord()
+      ..id = createUuid()
+      ..created = created
+      ..expires = null // does not expire
+      ..kind = AuditLogRecordKind.packagePublished
+      ..agent = uploader.userId
+      ..summary = summary
```
isoos, line 58: Note: if we wanted to have a separate `data` JSON blob, it wouldn't be too difficult to add it here, with or without a formalized schema.

app/lib/audit/models.dart
```diff
@@ -0,0 +1,68 @@
...
+import '../shared/datastore.dart' as db;
+import '../shared/utils.dart' show createUuid;
+
+@db.Kind(name: 'AuditLogRecord', idType: db.IdType.String)
+class AuditLogRecord extends db.ExpandoModel<String> {
+  @db.DateTimeProperty(required: true)
+  DateTime created;
+
+  @db.DateTimeProperty()
+  DateTime expires;
```
jonasfj, line 17:
```suggestion
  /// [DateTime] after which a background tasks should delete this entity.
  ///
  /// `null`, if the entity does not expire, this is used for package publication records.
  @db.DateTimeProperty()
  DateTime expires;
```

app/lib/audit/models.dart
```diff
@@ -0,0 +1,68 @@
...
+
+  @db.StringProperty(required: true)
+  String kind;
```
jonasfj, line 20:
```suggestion
  /// String identifying the kind of event recorded.
  ///
  /// Allowed values are documented in [AuditLogRecordKind].
  @db.StringProperty(required: true)
  String kind;
```

app/lib/audit/models.dart
```diff
@@ -0,0 +1,68 @@
...
+
+  @db.StringProperty(required: true)
+  String agent;
```
jonasfj, line 23:
```suggestion
  /// [User.id] of the user who initiated / authorized the recorded action.
  ///
  /// This is a UUIDv4. If the user has been deleted, it is possible that this
  /// property may not match any [User] entity.
  @db.StringProperty(required: true)
  String agent;
```

TODO: Double check that `[User.id]` is the correct identifier.

app/lib/audit/models.dart
```diff
@@ -0,0 +1,68 @@
...
+
+  @db.StringProperty(required: true)
+  String summary;
```
jonasfj, line 26:
```suggestion
  /// Textual summary of the action that was performed in **markdown**.
  ///
  /// This summary should contain a full explanation of the action that was taken,
  /// including who authorized it, which user or resource was subject of the action.
  /// Users are identified by the email associated with their account at the point-in-time
  /// when the action was taken.
  ///
  /// The following properties are intended for presentation in a table when rendering an audit-log:
  ///  * [created]
  ///  * [kind]
  ///  * [agent] (left as `<deleted-user>` if user no longer exists)
  ///  * [summary]
  ///
  /// This can only use simple markdown formatting no block structures.
  @db.StringProperty(required: true, indexed: false)
  String summary;
```

I think there is somewhere else where we use "simple markdown without block structures", I think it's in publisher descriptions.

app/lib/audit/models.dart
```diff
@@ -0,0 +1,68 @@
...
+
+  @db.StringListProperty()
+  List<String> users;
```
jonasfj, line 29:
```suggestion
  /// List of [User.id] for the users involved in this record.
  ///
  /// This is the list of users who should have this record show up in their personal audit-log.
  /// This should include [agent], as well as any users who is subject of this action, that's
  /// it should include the user who did this action, and the user who is invited/removed.
  ///
  /// This property **MUST** only be used for querying. The contents of this property is
  /// **NOT** intended for consumption or presentation by other means. It's only here
  /// to make it easy to find [AuditLogRecord]s relevant for a specific user.
  @db.StringListProperty(required: true)
  List<String> users;
```

Any reason not to require this be at-least an empty-list?

TODO: Double check that `[User.id]` is the correct identifier.

app/lib/audit/models.dart
```diff
@@ -0,0 +1,68 @@
...
+
+  @db.StringListProperty()
+  List<String> packages;
+
+  @db.StringListProperty()
+  List<String> packageVersions;
+
+  @db.StringListProperty()
+  List<String> publishers;
```
jonasfj, line 38: TODO: documentation matching that which we have for `users`

app/lib/audit/models.dart
```diff
@@ -0,0 +1,68 @@
...
+    @required DateTime created,
+  }) {
+    final summary = 'Package `$package` '
+        'version `$version` '
+        'was published '
+        'by `${uploader.email}`.';
+    return AuditLogRecord()
+      ..id = createUuid()
+      ..created = created
+      ..expires = null // does not expire
```
jonasfj, line 55: We could also make the property non-nullable, and set the expiration to year 3000... It creates a y3k problem, but chances are it won't be our problem :rofl:

app/lib/audit/models.dart
```diff
@@ -0,0 +1,68 @@
...
+  factory AuditLogRecord.packagePublished({
+    @required User uploader,
+    @required String package,
+    @required String version,
+    @required DateTime created,
+  }) {
+    final summary = 'Package `$package` '
+        'version `$version` '
+        'was published '
+        'by `${uploader.email}`.';
```
jonasfj, line 51: hahaha, you already used markdown :D

isoos:
Updated `expires` to use year 9999 instead of null.

app/lib/audit/models.dart
```diff
@@ -0,0 +1,129 @@
...
+      ..summary = summary
+      ..users = [uploader.userId]
+      ..packages = [package]
+      ..packageVersions = ['$package/$version']
+      ..publishers = [];
+  }
+}
+
+abstract class AuditLogRecordKind {
+  static const packagePublished = 'package-published';
```
jonasfj, line 128:
```suggestion
  /// Event that a package was published.
  ///
  /// This can be an entirely new package or just a new version to an existing package.
  static const packagePublished = 'package-published';
```

jonasfj (approved):

app/lib/audit/models.dart
jonasfj: I won't object if you thinki it's a good idea to include a `data` property with untyped format.

Then if we have to use the `data`, we would have to use git history to see what we might have stored before, and we would have to investigate what data we have stored. But even in untyped form it would be easy to keep comments documenting when new properties were added to the untyped object..

I'm okay with that, just make sure the property is not indexed :D

PR #4328, opened 2020-12-09
Refactor HTTP-client tools into a separate library.

isoos:
I wanted to add `httpClientWithAuthorization` but `shared/utils.dart` seems to be overcrowded, let's create a new library for these. It is a component that the test-profile importer will use - eventually.

### DIFF
--- a/app/lib/tool/utils/http.dart
+++ b/app/lib/tool/utils/http.dart
@@ -1,0 +1,66 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:http/http.dart' as http;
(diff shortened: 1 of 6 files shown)


app/lib/tool/utils/http.dart
```diff
@@ -0,0 +1,66 @@
...
+    //       These are safe, retrying after partially sending data is more
+    //       sketchy, but probably safe in our application.
+    retries: retries ?? 5,
+  );
+}
+
+/// Returns an [http.Client] which sends a `Bearer` token as `Authorization`
+/// header for each request.
+http.Client httpClientWithAuthorization({
+  @required String bearerToken,
```
jonasfj, line 34: Generally, it's better to take  `Future<String> Function() tokenProvider`, because tokens become invalid over time, and so you can create a function that caches the previous token and knowns when it should refresh the token...

jonasfj (approved):

Issue #4338, opened 2020-12-11
Search results show "published" date, let's show "updated" instead


jonasfj:
Some users have reported being confused by the fact that our search results says "published: <date-time>", because they would rather have the date-time when the package was last updated.

Since "published: <date-time>" is actually the time the latest stable version was published, it's actually pretty close to what they expect.

But if there are pre-releases, perhaps it's fine to show the "Last Updated", as the newest timestamp for either _stable version_ or _latest prerelease_.

-------

In the following example, the prerelease is actually from `Dec 2, 2020`, so perhaps it's fair to show "updated" and newest timestamp from whether the prerelease or stable release was published. 

![3wpBTQcpVYEPLam](https://user-images.githubusercontent.com/149732/101907636-232ce500-3bbb-11eb-8fa8-51b9053974a6.png)

isoos:
How about addig both date like this? 

<img width="304" alt="Screenshot 2020-12-11 at 14 35 22" src="https://user-images.githubusercontent.com/4778111/101909668-1c53a180-3bbe-11eb-898c-daffcef853a8.png">

jayoung-lee:
Another suggestion I'd like to make is changing the current date format to "X days/months/years ago". 

Currently, users do mental calculation to translate "Jul 1, 2020" into "6 months ago".  Providing easier-to-read information about the package’s last updated date is a small change, but will reduce cognitive load!

isoos:
Closing this, as the "updated" part has been fixed, we have a separate issue to track the "X ago" timestamp.

PR #4343, opened 2020-12-14
Adjust Updated label on package listing page.

isoos:
- #4338
- Added changelog entry for backfilling the `latestPublished` and `latestPrereleasePublished` properties for later use here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ Important changes to data-models, configuration and migrations between each
 AppEngine version, listed here to ease deployment and troubleshooting.
 
 ## Next Release (replace with git tag when deployed)
+ * Run `app/bin/tools/backfill_package_fields.dart`.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@ Important changes to data-models, configuration and migrations between each
 AppEngine version, listed here to ease deployment and troubleshooting.
 
 ## Next Release (replace with git tag when deployed)
+ * Run `app/bin/tools/backfill_package_fields.dart`.
```
jonasfj, line 5: Why this? Have we changed anything in how these are derived?

jonasfj (approved):

CHANGELOG.md
isoos: We've added `latestPublished` and `latestPrereleasePublished` in the previous release, we'll need to backfill them in the current one. It is slightly related to this change (eventually we want to use these as updated dates), that's the reason I've added it.

PR #4349, opened 2020-12-15
Wrapping bearer token parsing in a callback method.

isoos:
- Pattern copied from #4332.
- I think this wrapper should be used in tests, instead of calling `registerAuthenticatedUser` (eventually all test should be migrated to it). It would decouple the tests one step further from the test models.

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -84,6 +84,28 @@ Future<User> requireAuthenticatedUser() async {
   return _authenticatedUser;
 }
 
+/// Authenticates with bearer [token] and populates [_authenticatedUser] with
+/// it, running in a new scope.
+///
+/// The [token] may be an oauth2 `access_token` or an openid-connect
+/// `id_token` (signed JWT).
+///
+/// When no associated User entry exists in Datastore, this method will create
+/// a new one. When the authenticated email of the user changes, the email
+/// field will be updated to the latest one.
+Future<R> withAuthorizationToken<R>(
+    String token, Future<R> Function() fn) async {
+  return await ss.fork(() async {
+    final user = token == null || token.isEmpty
+        ? null
+        : await accountBackend._authenticateWithBearerToken(token);
+    if (user != null) {
+      registerAuthenticatedUser(user);
+    }
+    return await fn();
+  }) as R;
+}
+
 /// Represents the backend for the account handling and authentication.
 class AccountBackend {
   final DatastoreDB _db;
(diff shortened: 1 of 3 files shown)


app/lib/account/backend.dart
```diff
@@ -84,6 +84,28 @@ Future<User> requireAuthenticatedUser() async {
...
+/// Authenticates with bearer [token] and populates [_authenticatedUser] with
+/// it, running in a new scope.
+///
+/// The [token] may be an oauth2 `access_token` or an openid-connect
+/// `id_token` (signed JWT).
+///
+/// When no associated User entry exists in Datastore, this method will create
+/// a new one. When the authenticated email of the user changes, the email
+/// field will be updated to the latest one.
+Future<R> withAuthorizationToken<R>(
```
jonasfj, line 96: Should this be `@visibleForTestingOnly` ?

app/lib/account/backend.dart
```diff
@@ -275,7 +297,7 @@ class AccountBackend {
   /// When no associated User entry exists in Datastore, this method will create
   /// a new one. When the authenticated email of the user changes, the email
   /// field will be updated to the latest one.
-  Future<User> authenticateWithBearerToken(String token) async {
+  Future<User> _authenticateWithBearerToken(String token) async {
```
jonasfj, line 300: Should we just rename this `withAuthorizationToken` ?

app/lib/account/backend.dart
```diff
@@ -84,6 +84,28 @@ Future<User> requireAuthenticatedUser() async {
   return _authenticatedUser;
 }
 
+/// Authenticates with bearer [token] and populates [_authenticatedUser] with
+/// it, running in a new scope.
```
jonasfj, line 88: Isn't it better to explain that `requireAuthenticatedUser() ` will not fail when called in the `Zone` that `fn` is executed in.

Because that's the external interface... `_authenticatedUser` is internal to this library.

jonasfj (approved):

app/lib/account/backend.dart
isoos: It is also used from out handler chain.

app/lib/account/backend.dart
isoos: It still fails if the user authentication failed. In such cases we don't block the requests. Maybe we should though...

isoos:
PTAL. I've updated the code to reject unsuccessful authentication attempts. The `requireAuthenticatedUser` does not change though...

jonasfj (approved):

PR #4375, opened 2020-12-22
Use dartcli from install tab

sigurdm:
Explain how to use `dart pub add`.

install

### DIFF
--- a/app/lib/frontend/templates/views/pkg/install_tab.mustache
+++ b/app/lib/frontend/templates/views/pkg/install_tab.mustache
@@ -16,28 +16,26 @@
 {{#use_as_a_library}}
 <h2>Use this package as a library</h2>
 <h3>1. Depend on it</h3>
-<p>Add this to your package's pubspec.yaml file:</p>
-<pre><code class="language-yaml">{{dependencies_key}}:
-  <strong>{{package}}: {{example_version_constraint}}</strong>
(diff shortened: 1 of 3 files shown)


app/lib/frontend/templates/views/pkg/install_tab.mustache
```diff
@@ -16,28 +16,26 @@
...
-</code></pre>
+<p>With Dart:</p>
+<pre><code class="language-shell"> $ dart pub add {{package}}</code></pre>
 {{/use_pub_get}}
 {{#use_flutter_packages_get}}
-  <p>with Flutter:</p>
-  <pre><code class="language-shell">$ <strong>flutter pub get</strong>
-</code></pre>
+<p>With Flutter:</p>
+<pre><code class="language-shell"> $ flutter pub pub add {{package}}</code></pre>
```
jonasfj, line 26: Maybe we shouldn't show this part

sigurdm: Yeah - maybe this is too early. We need `flutter pub add` to work first...

isoos:
`dart pub add` returns `Could not find a command named "add".` on the stable SDK. Maybe we should wait with this a bit?

isoos:
rebased / updated in #4687

PR #4387, opened 2020-12-31
Update Highlight.js to 10.5.0

parlough:
There's been a lot of fixes and improvements since the previous version.

Relating to Dart, the primary improvements are:

> - enh(dart) Fix empty block-comments from breaking highlighting (#2898) Jan Pilzer
> - enh(dart) Fix empty doc-comment eating next line Jan Pilzer
> - **enh(dart) Add late and required keywords, the Never built-in type, and nullable built-in types (#2550) Sam Rawlins**

You can refer to the entire changelog here: https://github.com/highlightjs/highlight.js/blob/master/CHANGES.md

As for the increased file length, that is how they distribute their minified bundle now. The languages included are the same as listed in the README.

**Corresponding dartdoc PR:** https://github.com/dart-lang/dartdoc/pull/2474

PR #4389, opened 2021-01-06
Switch profile image to input tag to allow users to tab to it

parlough:
Previously users could not tab to the profile image, so they perhaps could not easily see that information or log out.

This switches the tag to an image input so the user can tab to the element and open its dropdown just like the other navigation buttons. 

I also added alt labels to the profile image elements.

After the change:
image

### DIFF
--- a/app/lib/frontend/templates/views/shared/site_header.mustache
+++ b/app/lib/frontend/templates/views/shared/site_header.mustache
@@ -83,11 +83,11 @@
     </div>
     {{#is_logged_in}}
     <div class="nav-container nav-profile-container hoverable">
-      <img class="nav-profile-img nav-profile-image-desktop" src="{{& user_session.image_url}}" />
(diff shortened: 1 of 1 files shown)


isoos:
@parlough: This looks great, thanks! One thing this needs is passing tests: which is in this case the `test/frontend/templates_test.dart`. Please flip this flag, run the tests and revert the flag:
https://github.com/dart-lang/pub-dev/blob/master/app/test/frontend/templates_test.dart#L40
Then the golden files will be updated with your new changes.

parlough:
@isoos As far as I can tell, no goldens include the header when logged in, so this isn't part of it. I tried running the tests with the flap flipped and nothing changed.

I believe the app frontend tests are failing due to dartdoc updating `higlight.js` but pub not yet(https://github.com/dart-lang/pub-dev/pull/4387). 

Let me know if I'm missing something though, thanks!

isoos (approved):

PR #4395, opened 2021-01-08
updated dart pub get command

Abhishek01039:
`pub get` is deprecated.
https://dart.dev/tools/pub/cmd/pub-get

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -54,9 +54,9 @@ jobs:
       - run: dart --version
       - uses: actions/checkout@v2
       - id: pkg_web_css_pub_get
-        name: "pkg/web_css; dart pub get --no-precompile"
+        name: "pkg/web_css; pub get --no-precompile"
         working-directory: pkg/web_css
-        run: dart pub get --no-precompile
+        run: pub get --no-precompile
       - name: "pkg/web_css; dartfmt -n --set-exit-if-changed ."
         if: "steps.pkg_web_css_pub_get.conclusion == 'success'"
         working-directory: pkg/web_css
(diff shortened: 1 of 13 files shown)


.github/workflows/dart.yml
```diff
@@ -54,9 +54,9 @@ jobs:
       - run: dart --version
       - uses: actions/checkout@v2
       - id: pkg_web_css_pub_get
-        name: "pkg/web_css; pub get --no-precompile"
+        name: "pkg/web_css; dart pub get --no-precompile"
```
jonasfj, line 57: This is a generated file, as it there is some support for old versions this haven't been migrated yet, see `mono_repo`

jonasfj (changes requested):
If we avoid manual changes to `.github/workflows/dart.yml` then this might be  good change :D

.github/workflows/dart.yml
Abhishek01039: Ah Thank
@jonasfj

tool/ci.sh
```diff
@@ -47,11 +47,11 @@ for PKG in ${PKGS}; do
     exit 64
   fi
 
-  pub get --no-precompile || EXIT_CODE=$?
+  dart pub get --no-precompile || EXIT_CODE=$?
```
jonasfj, line 50: Top of this file says: `# Created with package:mono_repo v3.4.2`

I'm guessing there is little reason to modify it, maybe we should patch `mono_repo` to write more clearly not to modify such files, I assume `mono_repo` will overwrite it.. If not then changes in this file might make sense...

jonasfj (changes requested):

tool/ci.sh
Abhishek01039: Should I discard this changes? @jonasfj

jonasfj: yeah, let's not modify generated files.

Abhishek01039: ok sure Thanks.

Abhishek01039: Done @jonasfj

jonasfj (approved):

Abhishek01039:
Thanks @jonasfj

PR #4402, opened 2021-01-11
SDK constraint on the /versions page.

### LINKED ISSUE #4376: Display min Dart SDK on versions page
6gADX3EKEymzfmg

isoos:
Fixes #4376.

Desktop view:
Screenshot 2021-01-12 at 16 20 13

### DIFF
--- a/app/lib/frontend/templates/views/pkg/versions/version_row.mustache
+++ b/app/lib/frontend/templates/views/pkg/versions/version_row.mustache
@@ -8,7 +8,7 @@
     {{#has_sdk}}
     <code>{{ sdk }}</code>
     {{/has_sdk}}
-    </td>
+  </td>
   <td class="uploaded">{{short_created}}</td>
   <td class="documentation">
     <a href="{{& dartdocs_url}}"
--- a/app/lib/frontend/templates/views/pkg/versions/version_row.mustache
+++ b/app/lib/frontend/templates/views/pkg/versions/version_row.mustache
@@ -8,7 +8,6 @@
     {{#has_sdk}}
       {{ sdk.major }}.{{ sdk.minor }}
       {{#sdk.has_channel}}({{ sdk.channel }}){{/sdk.has_channel}}
-      <span class="sdk-code">{{ sdk.code }}</span>
     {{/has_sdk}}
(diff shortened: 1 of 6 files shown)


app/lib/frontend/templates/views/pkg/versions/version_row.mustache
```diff
@@ -4,6 +4,11 @@
 
 <tr data-version="{{version}}">
   <td class="version"><a href="{{& version_url}}">{{version}}</a></td>
+  <td class="sdk">
+    {{#has_sdk}}
+    <code>{{ sdk }}</code>
+    {{/has_sdk}}
+    </td>
```
parlough, line 11: Just a question: If it doesn't have an SDK does the SDK cell(td) even need to be included at all? Can the conditional be lifted up?

Either way closing `td` tag here can be unindented.

isoos: The most visited pages are likely to be modern packages which do have the SDK constraint and thus they will have the column there. While we could make the entire column hidden if none of the versions have constraint, the extra calculation and conditions will yield very little overall.

parlough (approved):

app/lib/package/model_properties.dart
```diff
@@ -87,6 +88,16 @@ class Pubspec {
...
   }
 
+  /// Returns the minimal SDK version for Flutter SDK or Dart SDK, in that order,
+  /// only if specified.
+  ///
+  /// Returns null if the constraint does not follow the `>=<version>` pattern.
+  MinSdkVersion get minSdkVersion {
+    _load();
+    return MinSdkVersion.tryParse('F', _inner.environment['flutter']) ??
+        MinSdkVersion.tryParse('D', _inner.environment['sdk']);
```
jonasfj, line 98: Didn't we agree that we don't need this?

isoos: Oh, I may have missed that part of the discussion. I'll remove it for now, it is not that hard to re-add later.

isoos:
Deploying new version to staging....

jonasfj (approved):

PR #4405, opened 2021-01-13
Focus search on / key.

### LINKED ISSUE #4390: Further search bar improvement is possible 🥰
Yo, just saying, further improvement is possible 🥰 
image
image
image

---

Consider making search box always open, consider adding shortcut to focus on it. Search is an essential thing, why to hide it?
image

_Originally posted by @orsenkucher in https://github.com/dart-lang/pub-dev/issues/3958#issuecomment-752639400_

isoos:
- Fixes #4390
- Also has a pointer mouse cursor when the input field is inactive, and changes to the default text selection cursor when it is focues.

### DIFF
--- a/app/lib/frontend/templates/views/shared/site_header.mustache
+++ b/app/lib/frontend/templates/views/shared/site_header.mustache
@@ -13,7 +13,7 @@
   {{#show_header_search}}
   <div class="site-header-search">
     <form action="/packages" method="GET">
-      <input class="site-header-search-input" name="q" placeholder="New search..." autocomplete="on" />
+      <input class="site-header-search-input" name="q" placeholder="Shortcut key: /" autocomplete="on" />
     </form>
   </div>
   {{/show_header_search}}
--- a/app/lib/frontend/templates/views/shared/site_header.mustache
+++ b/app/lib/frontend/templates/views/shared/site_header.mustache
@@ -13,7 +13,7 @@
   {{#show_header_search}}
   <div class="site-header-search">
     <form action="/packages" method="GET">
-      <input class="site-header-search-input" name="q" placeholder="Shortcut key: /" autocomplete="on" />
+      <input class="site-header-search-input" name="q" placeholder="New search..." autocomplete="on" />
     </form>
   </div>
   {{/show_header_search}}
--- a/pkg/web_app/lib/src/gtag_js.dart
+++ b/pkg/web_app/lib/src/gtag_js.dart
@@ -6,9 +6,17 @@
 library gtag_js;
 
 import 'package:js/js.dart';
+import 'package:meta/meta.dart';
 
 @JS('gtag')
 external dynamic _gtag(String type, String action, Map<String, dynamic> data);
 
-void gtagEvent(String action, Map<String, dynamic> data) =>
-    _gtag('event', action, data);
+void gtagEvent(
+  String action, {
+  @required String category,
+  @required String label,
+}) =>
+    _gtag('event', action, {
+      'event_category': category,
+      'event_label': label,
+    });
--- a/pkg/web_app/lib/src/gtag_js.dart
+++ b/pkg/web_app/lib/src/gtag_js.dart
@@ -1,0 +1,14 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+@JS('window')
+library gtag_js;
+
+import 'package:js/js.dart';
+
+@JS('gtag')
+external dynamic _gtag(String type, String action, Map<String, dynamic> data);
+
+void gtagEvent(String action, Map<String, dynamic> data) =>
+    _gtag('event', action, data);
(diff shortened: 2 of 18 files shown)


pkg/web_app/lib/src/search.dart
```diff
@@ -5,12 +5,28 @@
...
   _setEventForSearchInput();
   _setEventForFiltersToggle();
   _setEventForSortControl();
   _setEventForCheckboxChanges();
 }
 
+void _setEventForKeyboardShortcut() {
+  final inputElem = document.querySelector('input.site-header-search-input');
+  if (inputElem != null && inputElem is InputElement) {
+    document.onKeyPress.listen((e) {
```
jonasfj, line 18:
```suggestion
    window.onKeyPress.listen((e) {
```

jonasfj: Might be worth testing that this doesn't prevent us from writing `/` in other input fields on the page.

I think it will.. We probably have to test what target the event is hitting, otherwise you'll steal form field focus from any field the user is trying to write slash in.

app/lib/frontend/templates/views/shared/site_header.mustache
```diff
@@ -13,7 +13,7 @@
   {{#show_header_search}}
   <div class="site-header-search">
     <form action="/packages" method="GET">
-      <input class="site-header-search-input" name="q" placeholder="New search..." autocomplete="on" />
+      <input class="site-header-search-input" name="q" placeholder="Shortcut key: /" autocomplete="on" />
```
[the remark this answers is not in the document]
isoos: I'm not sure if I've hovered the mouse long enough over the input field to let it displayed. I'd rather have it in the placeholder...

pkg/web_app/lib/src/search.dart
isoos: Good catch! I've updated with checking the active element differently. PTAL.

pkg/web_app/lib/src/search.dart
```diff
@@ -5,12 +5,33 @@
...
+  if (inputElem != null && inputElem is InputElement) {
+    window.onKeyPress.listen((e) {
+      // Ignore keys other than the shortcut key:
+      if (e.key != '/') return;
+
+      // Only trigger focus when nothing is focused, so that we don't steal
+      // focus from other elements.
+      final active = document.activeElement;
+      if (active == null || active == document.body) {
+        inputElem.focus();
```
jonasfj, line 26: We could also ask if the current element is not an input / textarea / etc...

I think think there are many elements that take input. But it's easy to give focus to link or other thing, is it not?

app/lib/frontend/templates/views/shared/site_header.mustache
jonasfj, line 16: Let's remove the placeholder change.

pkg/web_app/lib/src/search.dart
```diff
@@ -5,12 +5,34 @@
...
+    window.onKeyPress.listen((e) {
+      // Ignore keys other than the shortcut key:
+      if (e.key != '/') return;
+
+      // Only trigger the input field and steal focus when nothing is focused,
+      // or when the focused element is not an input element.
+      final active = document.activeElement;
+      final isRestricted = active is InputElement || active is TextAreaElement;
+      if (!isRestricted) {
+        inputElem.focus();
```
jonasfj, line 27: Maybe: send an event to analytics when shortcut is triggered

isoos: Done. PTAL.

pkg/web_app/lib/src/gtag_js.dart
```diff
@@ -0,0 +1,14 @@
...
+@JS('window')
+library gtag_js;
+
+import 'package:js/js.dart';
+
+@JS('gtag')
+external dynamic _gtag(String type, String action, Map<String, dynamic> data);
+
+void gtagEvent(String action, Map<String, dynamic> data) =>
+    _gtag('event', action, data);
```
jonasfj, line 14: I don't know js binding so well, but don't have to call:
https://api.dart.dev/beta/2.12.0-133.2.beta/dart-js_util/jsify.html

Or create anonymous objects... at-least if we have nested objects?

jonasfj: We could also use named parameters to make this easier to use... less powerful, but we probably won't use all features anyways :D

pkg/web_app/lib/src/search.dart
```diff
@@ -4,13 +4,44 @@
...
+
+        // prevent the trigger character to get typed into the input field
+        e.preventDefault();
+
+        // notify analytics
+        gtagEvent(
+          'keyboard-shortcut-search',
```
jonasfj, line 36: I think this could just be `focus-search`.

I'm not sure it matters much.. but ordering seems to be `category > action > label` at least in the "top events" view :D
AEMqxTfBmS9gFVE

pkg/web_app/lib/src/search.dart
```diff
@@ -4,13 +4,44 @@
...
+          category: 'keyboard-shortcut',
+          label: 'search',
```
jonasfj, line 38: In other events we a `path:/packages/http` as generic label.

jonasfj (approved):

jonasfj:
@googlebot I consent.

Issue #4404, opened 2021-01-12
Support preview releases on pub.dev


mit-mit:
Currently we differentiate between two kinds of package versions on pub.dev:

  1. Stable versions: Package versions that use a stable version format, x.y.z
  2. Prerelease versions: Package versions that use a prerelease version format, x.y.z-prerelease_suffix

An example is the current `characters` package, which has a version 1.1.0 stable version and a 1.1.0-nullsafety.5 prerelease:

<img width="773" alt="Screenshot 2021-01-12 at 14 45 47" src="https://user-images.githubusercontent.com/13644170/104322170-e66b4b00-54e4-11eb-8a42-3d35d0ce172f.png">

It can be convenient to allow publishing future stable versions of a package (e.g. characters 1.1.0 stable)
prior to the corresponding stable Dart release (Dart 2.12 stable) being available. If that is done on the current pub.site, then the pub UI would suddenly recommend `characters: 1.1.0` as the main stable version to all consumers, despite this depending on a non-stable SDK.

The present bug proposes refining the pub UI to show three kinds of versions:

  1. Stable versions: Package versions that use a stable version format, x.y.z, AND depend on a stable Dart SDK
  1. Preview versions: Package versions that use a stable version format, x.y.z, AND depend on a Dart SDK which is higher than the current stable Dart SDK
  1. Prerelease versions: Package versions that use a prerelease version format, x.y.z-prerelease_suffix

Here is an example of how we would render this for characters after a `1.1.0` version has been published, which depends on Dart `2.12.0-200-0.beta`, in a situation where the most recent stable Dart SDK is `2.10`:

<img width="779" alt="Screenshot 2021-01-12 at 14 21 23" src="https://user-images.githubusercontent.com/13644170/104322180-e9fed200-54e4-11eb-80d9-1610bad4b0f9.png">

mit-mit:
Note: the screenshot above it a bit confusing; you would not see `Preview: 1.1.0` and `Prerelease: 1.1.0-nullsafety.5` together (as we'd continue to hide prereleases which are < than the most recent stable or preview release). But you might see `Preview: 1.1.0` and `Prerelease: 1.2.0-foobar.1`.

mit-mit:
Functionality note: We probably also need to update the version tags on the search pages:
[...]

PR #4432, opened 2021-01-21
Use neat periodic tasks for updating the preview versions.

isoos:
#4404

### DIFF
--- a/app/test/tool/neat_task/datastore_status_provider_test.dart
+++ b/app/test/tool/neat_task/datastore_status_provider_test.dart
@@ -35,7 +35,7 @@ void main() {
       expect(await p2.get(), [1, 2]);
       expect(await p2.set([3, 4]), isTrue);
       expect(await p2.get(), [3, 4]);
-      
+
       expect(await p1.set([1, 2]), isFalse);
     });
   });
--- a/app/test/tool/neat_task/datastore_status_provider_test.dart
+++ b/app/test/tool/neat_task/datastore_status_provider_test.dart
@@ -35,6 +35,8 @@ void main() {
       expect(await p2.get(), [1, 2]);
       expect(await p2.set([3, 4]), isTrue);
       expect(await p2.get(), [3, 4]);
+      
+      expect(await p1.set([1, 2]), isFalse);
     });
   });
 }
--- a/app/test/tool/neat_task/datastore_status_provider_test.dart
+++ b/app/test/tool/neat_task/datastore_status_provider_test.dart
@@ -1,0 +1,40 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+import 'package:pub_dev/shared/datastore.dart';
+import 'package:pub_dev/tool/neat_task/datastore_status_provider.dart';
+
+import '../../shared/test_services.dart';
+
+void main() {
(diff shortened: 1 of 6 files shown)


mit-mit:
@isoos can you describe what experience this would give us?

jonasfj:
We dynamically fetch information from `dart.dev` to detect what the latest stable version is.
We then periodically scan all packages and adjust "stable", "preview" and "prereleases" as the designated versions might change when a new version is published.

This doesn't happen with traffic switching because we don't store the "stable", "preview" and "prereleases" designations on a per-deployment basis -- that would probably introduce more latency.

Publishing a new stable release on `dart.dev` (in GCS), will start update "stable", "preview" and "prereleases" designations on pub.dev

mit-mit:
Ah, that makes sense. We usually publish on dart.dev/in GCS some hours, or even days, in advance of announcing.

What is the time span after publishing the new stable release before pub.dev will have completed the update?

app/test/tool/neat_task/datastore_status_provider_test.dart
```diff
@@ -0,0 +1,40 @@
...
+    testWithProfile('set status', fn: () async {
+      final p1 = DatastoreStatusProvider.create(dbService, 'task-id');
+      expect(await p1.set([1, 2]), isTrue);
+      expect(await p1.get(), [1, 2]);
+
+      final p2 = DatastoreStatusProvider.create(dbService, 'task-id');
+      expect(await p2.set([3, 4]), isFalse);
+      expect(await p2.get(), [1, 2]);
+      expect(await p2.set([3, 4]), isTrue);
+      expect(await p2.get(), [3, 4]);
```
jonasfj, line 37:
```suggestion
      expect(await p2.get(), [3, 4]);
      
      expect(await p1.set([1, 2]), isFalse);
```

jonasfj (approved):

isoos:
Currently we query the https://storage.googleapis.com/dart-archive/channels/stable/release/latest/VERSION endpoint. Not sure when this is being updated in the release process, but we query it every hour and may update the `/versions` list on the package page based on that. This PR adds a daily background task (that may run more frequently) to update individual Package entries in datastore. In an unlucky timing it may take 25 hours to get them updated.

mit-mit:
Would it be possible to manually "kick" the background task to make it run quicker?

jonasfj:
> Ah, that makes sense. We usually publish on dart.dev/in GCS some hours, or even days, in advance of announcing.

@athomas, when do we update the `https://storage.googleapis.com/dart-archive/channels/stable/release/latest/VERSION` file?

> What is the time span after publishing the new stable release before pub.dev will have completed the update?

We run the periodic task every 24 hours.

athomas:
> @athomas, when do we update the `https://storage.googleapis.com/dart-archive/channels/stable/release/latest/VERSION` file?
 
https://storage.googleapis.com/dart-archive/channels/stable/release/latest/VERSION is updated just before the release goes to dart.dev.

It's copied from:
https://storage.googleapis.com/dart-archive/channels/stable/raw/latest/VERSION (not exactly, because it's actually copied from the folder with the version number rather than `latest` but the `latest` URL is probably what you'd want to query).

Now, raw will also contain broken builds (assuming the build worked, but perhaps the tests didn't work, or the build failed on some platforms but succeeded on others), so it can give you a heads up but you don't want to just deploy raw to prod.

jonasfj:
> https://storage.googleapis.com/dart-archive/channels/stable/release/latest/VERSION is updated just before the release goes to dart.dev.


Thanks, that's perfect!

PR #4436, opened 2021-01-22
CLI tool to trigger version field updates + concurrency control.

isoos:
#4404


jonasfj (approved):
We could also call `NeatTaskScheduler.trigger()` on the instance that runs the tasks.. That might also update the status file -- I don't recall.

PR #4447, opened 2021-01-26
Wrap all isolate initialization code inside appengine service scopes.

### LINKED ISSUE #4443: _isolates.dart uses package:logging outside a zone created with `withServices`
Using logging outside `withServices` will dump stuff to stdout.

This runs in the root zone of a fresh isolate:
https://github.com/dart-lang/pub-dev/blob/1f506db7f01882de06aac147be6a814e8318cdc5/app/lib/service/entrypoint/_isolate.dart#L235-L245

Hence, there is nowhere to send log messages. We should wrap with `withServices` or at-least `withAppEngineServices`.
We probably should catch exceptions outside of this, but then it's probably best to simply print them to `stderr` manually, rather than using `package:logging` -- And it's probably preferable to avoid multi-line messages if possible (or at-least print a single coherent message in a single line).

isoos:
- Fixes #4443.
- I'm planning to do a follow-up that removes the stat-related messages.

### DIFF
--- a/app/lib/service/entrypoint/_isolate.dart
+++ b/app/lib/service/entrypoint/_isolate.dart
@@ -62,171 +62,173 @@ Future startIsolates({
   Future<void> Function(WorkerEntryMessage message) workerEntryPoint,
   Duration deadWorkerTimeout,
 }) async {
-  if (!envConfig.isRunningLocally) {
-    // The existence of this file may indicate an issue with the service health.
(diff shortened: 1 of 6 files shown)


jonasfj (approved):

isoos:
PTAL: implemented zone-local as discussed

jonasfj (approved):

Issue #4386, opened 2020-12-29
meta: Launch with happiness survey


jonasfj:
The following tasks is blocking launch with happiness survey:

 * [x] Migrate to Google Tag Manager (currently behind experimental)
 * [x] Remove HaTS from Google Tag Manager
 * [x] Add HaTS to Google Tag Manager (after deployment)
 * [x] Refactor `History` entities
   * Add `AuditLogRecord` with properties: `created`, `expires`, `kind`, `agent`,
     `summary`, `reason`, `users`, `packages`, `packageVersions`, `publishers`.
   * Expire `AuditLogRecord` entities after 60 days
   * Retain `AuditLogRecord` for `package-published` events indefinitely.
 * [x] Enabled administration of uploaders through the package admin page.
 * [x] Error page requiring a login before login when accepting an invitation should
   * say: "if you have no signed-in on pub.dev before a pub.dev profile will automatically be created upon sign-in"
   * in-line with [mockup](https://user-images.githubusercontent.com/149732/103291967-2282b400-49ed-11eb-85dd-dd8e81b278b3.png)

 * [x] Publisher creation page should state what a publisher can do ([mockup](https://user-images.githubusercontent.com/149732/103292005-3af2ce80-49ed-11eb-8f53-f734a3bf7c2b.png))

jonasfj:
Deployed today.

PR #4451, opened 2021-01-27
Updated text on the create publisher page.

isoos:
#4386
Screenshot 2021-01-27 at 16 45 49

### DIFF
--- a/app/lib/frontend/templates/views/publisher/create.mustache
+++ b/app/lib/frontend/templates/views/publisher/create.mustache
@@ -25,7 +25,7 @@
 <p>
   After you create a verified publisher, you will be the only member,
   and your email will be listed as the <i>public contact email</i> of
-  the publisher.
+  the publisher (you can change this later).
 </p>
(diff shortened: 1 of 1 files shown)


app/lib/frontend/templates/views/publisher/create.mustache
```diff
@@ -19,15 +19,33 @@
...
   or a collaborator on a property that someone else has verified. 
 </p>
 <p>
-  After you create a verified publisher, you can use the <i>publisher Admin page</i>
-  to invite other users to become members of the publisher.
-  Members of the publisher can transfer existing packages to the publisher using
-  the <i>package Admin page</i>.
+  After you create a verified publisher, you will be the only member,
+  and your email will be listed as the <i>public contact email</i> of
+  the publisher.
```
jonasfj, line 28: Reading it I wonder if we should add:
```suggestion
  the publisher (you can change this later).
```

jonasfj (approved):

Issue #4105, opened 2020-10-07
Having trouble setting up local development server


sumandipanshu:
**What is the problem?**
1. Exception: Unknown project id: &lt;my-project-id&gt;
2. DetailedApiRequestError(status: 403, message: &lt;my-service-account&gt; does not have storage.buckets.get access to the Google Cloud Storage bucket.)

**Steps to reproduce:**
1. Follow steps mentioned in [development.md](https://github.com/dart-lang/pub-dev/blob/master/doc/development.md)
2. If you use:
```terminal
export GCLOUD_PROJECT=<my-project-id>
export GCLOUD_KEY=<path-to-service-account-key.json>
dart bin/server.dart default
```
It results in problem#1(already mentioned above).
After digging into the source code, what i found maybe is the root cause of this issue. [Refer here.](https://github.com/dart-lang/pub-dev/blob/fae2152c81ca80bdd918f2850ff1650683cf20d1/app/lib/shared/configuration.dart#L237-L245)

3. If you use:
```terminal
export GCLOUD_PROJECT=dartlang-pub-dev
export GCLOUD_KEY=<path-to-service-account-key.json>
dart bin/server.dart default
```
It results in problem#2(already mentioned above).

**Screenshots:**
![problem1](https://user-images.githubusercontent.com/42720239/95382180-e1579800-0906-11eb-9833-bb2e9ba2687b.png)
![problem2](https://user-images.githubusercontent.com/42720239/95382185-e3215b80-0906-11eb-8324-80d5ccfb74ed.png)

**Note:**
1. My service account has following roles assigned:
     - Storage Admin
     - Storage Legacy Bucket Owner
     - Storage Object Admin
2. Tried on both access controls: Fine-grained and Uniform

jonasfj:
Hold on this is something we're working to improve.

gfb-47:
I'm having the same issue.

isoos:
Since the original report we've introduced the local development server that has no direct use of google cloud (`app/bin/fake_server.dart`). Please use that for local development:
https://github.com/dart-lang/pub-dev/blob/master/doc/development.md#local-development-without-external-services

PR #4471, opened 2021-02-02
Fake token authentication UI for local development.

isoos:
- stores token in the browser's local storage
- the sign-in button is reused for asking for the token
- the sign-out button works as expected
- follows the same session update mechanism as the normal google auth
- leaves out a few nice-to-haves (e.g. fake profile picture that doesn't brake layout), but function-wise completes the missing piece for #4105

### DIFF
--- a/pkg/web_app/lib/src/account.dart
+++ b/pkg/web_app/lib/src/account.dart
@@ -58,8 +58,15 @@ void setupAccount() {
   final metaElem =
       document.querySelector('meta[name="google-signin-client_id"]');
   final clientId = metaElem == null ? null : metaElem.attributes['content'];
+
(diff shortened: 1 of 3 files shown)


pkg/web_app/lib/src/account.dart
```diff
@@ -60,9 +60,8 @@ void setupAccount() {
   final clientId = metaElem == null ? null : metaElem.attributes['content'];
   if (clientId == null || clientId.isEmpty) {
```
jonasfj, line 61: Might it not be better to set something else here?

Like requiring that `clientId == 'fake-auth-local-dev'`

jonasfj (approved):

jonasfj (approved):

Issue #4464, opened 2021-02-02
which version of Flutter SDK is used to analyse


zmtzawqlp:
It seems that Pub.dev is not analysing  packages with Stable branch of Flutter. 

And we can't find which version is used to analyse the packages.(we can find it before).

zmtzawqlp:
for example,  https://pub.dev/packages/extended_image/score
This api is changed in Master branch but it's ok in Stable branch

![image](https://user-images.githubusercontent.com/16477333/106556986-c8c95880-655b-11eb-9af2-b7e764fdc794.png)

jonasfj:
We temporarily switched to using a beta version while null-safety is being shipped.

We probably should support latest dev release and stable release.

AlexV525:
> We probably should support latest dev release and stable release.

IMO it's not "we should", but "we must" instead. The worst thing about this change is hidden the current analyze version that publishers cannot do anything to save their UNIDENTIFIED packages. Basically we have two demands.
* Use stable for release version, use beta/dev/master for pre-release version, make it switchable or semantic.
* Display the current Dart/Flutter version that pub is using to analyze. This info disappeared after pub's interface updated last time.

PR #4481, opened 2021-02-03
Stable + preview ToolEnvironment (using the same SDKs).

isoos:
- follow-up on #4479, in preparation for #4464
- this only introduces the concept of deciding on stable/preview SDKs, but do not yet uses two different one (that is to be done in a subsequent PR)


AlexV525:
Any concept in preview SDKs? Ideally I can use `x.y.x-beta.0` with the beta channel, `x.y.z-dev.0` with the dev channel, etc.
(Here `dev` is the most common cases today, the second might be `pre` or `nullsafety`)

jonasfj (approved):

PR #4483, opened 2021-02-04
Cleanup PackageVersionPubspec

isoos:
Assuming that the `app/bin/tools/remove_packageversionpubspec.dart` has been run and Datastore no longer contains entries.


jonasfj (approved):
I did a quick check that we removed this, found that they weren't fixed the script and deleted them.

PR #4497, opened 2021-02-05
Tool to remove History entries from Datastore.

isoos:


jonasfj (approved):
Maybe a CHANGELOG.md updated?

isoos:
> Maybe a CHANGELOG.md updated?

There is another PR in the works that deprecates the `History` entity and I'd add the note for removal with that.

PR #4499, opened 2021-02-05
migrate to GTM

jonasfj:
Fixes step 1 in https://github.com/dart-lang/pub-dev/issues/4386


app/test/dartdoc/golden/dygraph_0.1.1_index.out.html
```diff
@@ -1,8 +1,8 @@
 <html lang="en">
   <head>
     <meta name="robots" content="noindex"/>
-    <script async="async" src="https://www.googletagmanager.com/gtag/js?id=UA-26406144-13"/>
-    <script src="/static/js/gtag.js"/>
+    <script async="async" src="https://www.googletagmanager.com/gtm.js?id=GTM-MX6DBN9"/>
+    <script src="/static/js/gtm.js?hash=fltt8rc82ob3timp0tvnk6b136c2fqn0"/>
```
jonasfj, line 5: @isoos we agree that it's better to have this hash here, or did we deliberately avoid this in dartdoc because it's generated?

jonasfj: Come to think, we probably don't want this...

isoos: I think this is better with the hash, as it allows better caching while it falls back to serve the content when needed.

isoos (approved):

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ AppEngine version, listed here to ease deployment and troubleshooting.
 
 ## Next Release (replace with git tag when deployed)
 
+ * Removed all usage of `app/static/js/gtag.js` it can be removed after a few
```
isoos, line 6: Please update to HEAD, I think the CHANGELOG has more content in it.

PR #4510, opened 2021-02-10
Use cache to reduce Datastore scans on Job initialization.

isoos:


app/lib/shared/redis_cache.dart
```diff
@@ -250,6 +250,21 @@ class CachePatterns {
...
       .withCodec(utf8)['/'];
+
+  /// Stores the flag that latest version of packages have been scanned for job entities.
+  Entry<bool> jobHistoryLatestScanned(String service) =>
+      jobHistoryPackageScanned(service, '');
+
+  /// Stores the flag that a [package]'s versions have been scanned for job entities.
+  Entry<bool> jobHistoryPackageScanned(String service, String package) => _cache
+      .withPrefix('job-history-package-scanned')
+      .withTTL(Duration(days: 7))
```
jonasfj, line 261: Won't this just cause the task to only run every 7 days.. nobody ever clears this cache...

Curious, couldn't we just do in a neat-periodic-task that runs every 7 days? Using a status that is runtimeVersioned..

Maybe I'm missing something, I'm not sure I entirely remember how all the job code workds :D

jonasfj (approved):

app/lib/shared/redis_cache.dart
isoos: The idea is that this job runs as soon as possible, in a distributed fashion (partitioned by packages).

A similar effect could be done via runtimeVersioned periodic task, although it would be an overkill to have a separate job for each package. I'd rather have this around for now.

Yes, the 7-days timeout is intentional, it seemed that once we had run it at the time of the new deployment, we might not want to run it too frequently...

Issue #4482, opened 2021-02-03
Migrate to neat-periodic-task for background tasks


jonasfj:
There are some here:

https://github.com/dart-lang/pub-dev/blob/9aaf7812e41834e26567138f1ad8c7e6efbba385/app/lib/service/entrypoint/frontend.dart#L133-L138

We have decent monitoring for neat-periodic-tasks.
- [x] frontend
- [x] analyzer
- [x] dartdoc
- [x] search
- [ ] review their frequency

PR #4519, opened 2021-02-15
Periodic task for GC of search snapshot and dartdoc SDKs.

isoos:
- #4482
- deletes old search snapshots
- deletes old dartdoc SDKs

### DIFF
--- a/app/lib/tool/neat_task/pub_dev_tasks.dart
+++ b/app/lib/tool/neat_task/pub_dev_tasks.dart
@@ -9,9 +9,11 @@ import 'package:neat_periodic_task/neat_periodic_task.dart';
 import '../../account/backend.dart';
 import '../../account/consent_backend.dart';
 import '../../audit/backend.dart';
+import '../../dartdoc/backend.dart';
 import '../../job/backend.dart';
 import '../../package/backend.dart';
 import '../../scorecard/backend.dart';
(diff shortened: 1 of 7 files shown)


app/lib/tool/neat_task/pub_dev_tasks.dart
```diff
@@ -52,7 +54,21 @@ void setupAnalyzerPeriodicTasks() {
 /// Setup the tasks that we are running in the dartdoc service.
 void setupDartdocPeriodicTasks() {
   _setupJobCleanupPeriodicTasks();
-  // TODO: migrate cleanup of the extracted SDK data
+
+  // Deletes the extracted dartdoc data from old SDKs.
+  _daily(
```
jonasfj, line 59: If this involves crawling around in GCS then maybe we don't need to run it daily?

jonasfj (approved):

app/lib/tool/neat_task/pub_dev_tasks.dart
isoos: At this point this is a cheap operation: scanning a list of objects in a directory with less than 100 items, and this way it will be done fewer times than right now (every worker startup would trigger it).

Let's review it again once we have all the periodic task migrated. I've added another bullet point to the tracking issue.

Issue #4458, opened 2021-02-01
/readiness_check failReason:"disk full"


jonasfj:
Happens for both `dartdoc` and `analyzer`.

isoos:
On some machines, the tool environment cycle's (which is capped either by 50 analysis or going over a threshold size for pub cache) reached over 100 as the counters in the log report. That part seems to be okay.

When there is a disk full readiness check failure, there is an accompanying log entry that goes along the lines:

> "Free disk space (570 MB) is lower than threshold value 800 MB. Reporting instance permanently unhealthy. Note: different instances will have a different threshold value."

Based on the logs, I think the process of handling the temporary directory for pub cache is okay, assuming that there is no hanging/timed out but not killed process that writes into it later.

We should further monitor the following:
- [ ] a separate base directory used only for the pub caches
- [ ] the system temp directory
- [ ] the tool directories (Dart and Flutter SDKs)

isoos:
The issue seems to be gone from the logs.

PR #4522, opened 2021-02-16
 Updated ToolEnvironment handling

isoos:
- running `git gc` regularly to reduce the probability of "disk full" issue #4458
- forcing singleton use with `pool` (makes initialization and cleanup unambigious)
- increased the threshold for cache reset
- added `CI=true` to the environment variables

### DIFF
--- a/app/lib/shared/tool_env.dart
+++ b/app/lib/shared/tool_env.dart
@@ -39,9 +39,6 @@ final _toolEnvTempDir = Directory.systemTemp.createTempSync('tool-env');
 /// Forcing callback processing into a single thread.
 final _pool = Pool(1);
 
-// The cached values of directory sizes (path -> size in bytes).
-Map<String, int> _sizes;
-
 _ToolEnvRef _current;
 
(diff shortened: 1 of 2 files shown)


app/lib/shared/tool_env.dart
```diff
@@ -63,91 +75,73 @@ class _ToolEnvRef {
...
+    dartSdkDir: envConfig.previewDartSdkDir,
+    flutterSdkDir: envConfig.previewFlutterSdkDir,
+    pubCacheDir: resolvedDirName,
+    environment: {
+      'CI': 'true',
+      'FLUTTER_ROOT': envConfig.previewFlutterSdkDir,
+    },
+  );
+  await _gitGc(envConfig.stableFlutterSdkDir);
+  await _gitGc(envConfig.previewFlutterSdkDir);
```
jonasfj, line 121: A comment explaining why we do this would be appreciated

jonasfj (approved):

PR #4524, opened 2021-02-16
DartdocRecord to store dartdoc content information in Datastore.

isoos:
- `uploading` / `ready` / `deleting` status is in-sync with the entry file stored in the storage bucket
- periodic task to remove entities with old runtime versions

- Planned status: `obsolete` - when a runtime version has a new entry file that is ready to serve content. Periodic task can be running to delete entities with obsolete status.
- Reducing the use of the entry files in the bucket will come only later.

### DIFF
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -110,7 +110,9 @@ class DartdocBackend {
     final record =
         DartdocRecord.fromEntry(entry, status: DartdocRecordStatus.uploading);
     // store record's upload status
-    await withRetryTransaction(_db, (tx) async => tx.insert(record));
+    await withRetryTransaction(_db, (tx) async {
+      tx.insert(record);
+    });
 
     // upload is in progress
     await uploadBytesWithRetry(
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -107,6 +107,11 @@ class DartdocBackend {
 
   /// Uploads a directory to the storage bucket.
   Future<void> uploadDir(DartdocEntry entry, String dirPath) async {
+    final record =
+        DartdocRecord.fromEntry(entry, status: DartdocRecordStatus.uploading);
+    // store record's upload status
+    await withRetryTransaction(_db, (tx) async => tx.insert(record));
+
     // upload is in progress
     await uploadBytesWithRetry(
         _storage, entry.inProgressObjectName, entry.asBytes());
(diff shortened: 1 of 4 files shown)


app/lib/dartdoc/backend.dart
```diff
@@ -375,9 +401,29 @@ class DartdocBackend {
   }
 
   Future<void> _deleteAll(DartdocEntry entry, {int concurrency}) async {
+    await withRetryTransaction(_db, (tx) async {
+      final r = await tx.lookupValue<DartdocRecord>(
+        _db.emptyKey.append(DartdocRecord, id: entry.uuid),
+        orElse: () => null,
+      );
+      if (r != null) {
+        r.status = DartdocRecordStatus.deleting;
```
jonasfj, line 410: Why do we need to change it to `deleting`?

This should have `'runtimeVersion < ', shared_versions.gcBeforeRuntimeVersion`, I don't think there any chance we'll try to use it, right?

So I don't think we need to write that we're currently deleting it.

It might make sense if we wanted to avoid concurrent deletion, but then we need something that removes this lock after a timeout... and if we use neat-periodic-task to schedule this, then there is rarely two cron jobs running at the same time.. Moreover, I'm not sure it's the end of the world if we accidentally did some concurrent deletion.

app/lib/dartdoc/models.dart
```diff
@@ -3,16 +3,112 @@
...
 import 'storage_path.dart' as storage_path;
 
 part 'models.g.dart';
 
+/// Status values for [DartdocRecord].
+abstract class DartdocRecordStatus {
+  static const uploading = 'uploading';
+  static const ready = 'ready';
+  static const deleting = 'deleting';
+}
```
jonasfj, line 23: Not sure, why we need `deleting`, maybe we just need a boolean `ready` (that also makes our model simpler)

app/lib/dartdoc/models.dart
```diff
@@ -3,16 +3,112 @@
...
+
+@db.Kind(name: 'DartdocRecord', idType: db.IdType.String)
+class DartdocRecord extends db.ExpandoModel<String> {
+  String get uuid => id;
```
jonasfj, line 27: unique identifier that identifies a specific execution of dartdoc.. maybe it could be called `runId`? and have a documentation string that says it's a uuid for a specific run of dartdoc.

jonasfj (approved):

app/lib/dartdoc/backend.dart
isoos: The same runtimeVersion could create a second content. When that happens, we can start to delete the older version, as we won't be serving it anyway. (This is one of the pending unmigrated periodic task.)

I don't think it is crucial to expire these deletes, as eventually it will be caught by the runtimeVersion-based cleanups.

app/lib/dartdoc/models.dart
isoos: It may be needed (not sure about it yet) by the serving part: if an entry becomes deleting, we shouldn't serve from that content anymore.

app/lib/dartdoc/models.dart
isoos: I like `runId`, not sure why we haven't used it before...

PR #4564, opened 2021-02-25
Link to extracted license file in the infobox.

isoos:


app/test/frontend/golden/pkg_show_version_page.html
```diff
@@ -233,8 +233,6 @@ <h3 class="title">Uploader</h3>
                 <a href="/packages?q=email%3Ahans%40juergen.com" title="Search packages from hans@juergen.com" rel="nofollow">hans@juergen.com</a>
               </span>
             </p>
-            <h3 class="title">License</h3>
-            <p>BSD (LICENSE.txt)</p>
```
jonasfj, line 237: Why is the license just gone here?

app/test/frontend/golden/pkg_show_page_publisher.html
```diff
@@ -290,6 +294,10 @@ <h3 class="title pkg-infobox-metadata">Metadata</h3>
             <a class="link" href="https://neon.example.dev/" rel="ugc">Homepage</a>
             <br/>
           </p>
+          <h3 class="title">License</h3>
+          <p>
+            <a href="/packages/neon/license">LICENSE</a>
+          </p>
```
jonasfj, line 300: Why wasn't this here before?

jonasfj (approved):

app/test/frontend/golden/pkg_show_version_page.html
isoos: Change in how the data is extracted. In the non-migrated tests (those without the test-profile) this is broken because inconsistencies in the data, but in the other case (those with migrated to test-profile) this is consistent, and thus it appears as new link.

PR #4583, opened 2021-03-04
Upgrade runtime SDK to 2.12.0

isoos:
- Only the SDK being used is updated, dependencies can be updated on a separate schedule.


sigurdm (approved):
LGTM

PR #4588, opened 2021-03-07
yaml based config

n0npax:
Some fixes to make spinning of the project easier


app/lib/shared/configuration.dart
```diff
@@ -255,10 +254,12 @@ class Configuration {
   factory Configuration.fromEnv(EnvConfig env) {
     if (env.gcloudProject == 'dartlang-pub') {
       return Configuration._prod();
-    } else if (env.gcloudProject == 'dartlang-pub-dev') {
-      return Configuration._dev();
+    } else if (env.gcloudProject.startsWith('dartlang-pub-dev')) {
```
jonasfj, line 257: I think it might be better to have a separate env var to indicate whether we want a development or production profile..

@isoos any thoughts here?

isoos: I think this is not a pattern that we'd like to establish. If somebody wants to run pub on appengine, they should use their own namespace and naming conventions (for the project, for the buckets, for everything basically).

It is more pronounced with the lack of popularity file, and I could imagine it may happen with more stuff in the future.

I'd rather have a solid no-dependency local development environment for contributors of pub.dev, and a separate pluggable-backend pub server that people can self-host.

isoos: As a consequence, I think I'll just split the current `development.md` into two, and frame the Appengine setup part as supporting documentation for pub-dev administrators only. Maybe we should also rename these configuration options too. wdyt?

n0npax: > I think this is not a pattern that we'd like to establish. If somebody wants to run pub on appengine, they should use their own namespace and naming conventions (for the project, for the buckets, for everything basically).

currently, the configuration seems to be hardcoded.
Developer need access to `dartlang-pub-dev` and it's subresources (buckets/datastore/...) and this seems impossible for not google/dart member.
To spin up `pub.dev` on my own I had to replace project name refs and debug to find out datastore is selected base on `env.gcloudProject`

Personally, I'd prefer to have logic like:
```
if env.gcloudProject == 'dartlang-pub':
    return config._prod()
else:
    return config._dev(env.gcloudProjec)
```

isoos: @n0npax: what do you think about using a separate config file? If you want to use your own custom AppEngine project, most of the values in the configuration needs to be updated, and it would be handled better via a config file (e.g. a `.yaml` that is parsed into a `Configuration` object via `json_serializable`). If you would like to do that, I can help you with the change.

At the same time, I'll refactor the documentation around local development a bit. I'm using the fake server locally for a time now, and I think it is ideal for small contributions. I'd like to emphasise that a bit more than currently is. I'd also do a separate documentation on how to run the project on your custom AppEngine project, but that seems to be a bit lower priority for me (compared to the immediately available fake server). If you want to do that setup, and willing to identify the blockers, I'm happy to get you through and update the code and/or the docs.

n0npax: The idea of `yaml` based config files sounds good. 
yes, I'll be happy to do refactor this PR to use it.

app/test/shared/configuration.dart
```diff
@@ -0,0 +1,22 @@
...
+import 'package:yaml/yaml.dart';
+
+import 'package:pub_dev/shared/configuration.dart';
+
+void main() {
+  test('Foo config from yaml file', () async {
+    final Configuration config =
+        Configuration.fromYaml('test/shared/test_data/foo_config.yaml');
+    final expectedValue = 'foo';
+    expect(config.projectId == expectedValue, isTrue);
```
isoos, line 17: in such places use instead direct value matcher:
```dart
expect(config.projectId, 'expected-value');
```

app/test/shared/test_data/foo_config.yaml
```diff
@@ -0,0 +1,6 @@
+projectId: foo
+packageBucketName: foo
+dartdocStorageBucketName: foo
+popularityDumpBucketName: foo
+searchSnapshotBucketName: foo
+backupSnapshotBucketName: foo
```
isoos, line 6: nit: add extra newline

doc/development.md
```diff
@@ -99,3 +100,11 @@ The key must have access to:
 If starting from a new cloud project, APIs for Cloud Data Store will have to be
 activated and it might be necessary to define indexes with
 `gcloud app deploy index.yaml`.
+
```
isoos, line 103: Let's remove this for now, and implement it in a separate PR. Instead of an extra setup step, it may be even better if we removed the hard failure if the file or the bucket is not there.

doc/development.md
```diff
@@ -76,13 +76,14 @@ dart pub global run mono_repo pub get
...
 To run the default application (web frontend) locally, do the following steps:
-```
+```bash
 cd app
 dart pub get
 export GCLOUD_PROJECT=<gcloud-project>
 export GCLOUD_KEY=<path-to-service-account-key.json>
 dart bin/server.dart default
 ```
+Any project with prefix `dartlang-pub-dev` will be considered as a `dev` env.
```
isoos, line 86: Let's remove this for now, we don't want to establish such pattern - it could lead to weird name collisions and name squatting.

app/lib/shared/configuration.dart
```diff
@@ -255,11 +269,9 @@ class Configuration {
   factory Configuration.fromEnv(EnvConfig env) {
     if (env.gcloudProject == 'dartlang-pub') {
       return Configuration._prod();
-    } else if (env.gcloudProject == 'dartlang-pub-dev') {
-      return Configuration._dev();
-    } else {
-      throw Exception('Unknown project id: ${env.gcloudProject}');
-    }
+    } else if (env.gcloudProject.startsWith('dartlang-pub-dev')) {
```
isoos, line 272: Let's remove this for now, we don't want to establish such pattern - it could lead to weird name collisions and name squatting.

app/lib/shared/configuration.dart
```diff
@@ -3,11 +3,17 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:io';
+import 'package:googleapis/run/v1.dart';
+import 'package:json_annotation/json_annotation.dart';
+import 'package:yaml/yaml.dart';
+import 'dart:convert';
```
isoos, line 9: please reorder: `dart:convert` before `dart:io`.

app/lib/shared/configuration.dart
```diff
@@ -3,11 +3,17 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:io';
+import 'package:googleapis/run/v1.dart';
```
isoos, line 6: Why do we need this?

isoos (changes requested):

n0npax:
I converted it to a draft, it's not ready for a review yet :)

n0npax:
Requested refactor:

Seems to +/- work as per

```bash
export PUB_CONFIG=config/dev-config-n0npax.yaml
dart bin/server.dart default                   

2021-03-10 21:48:49.621373 Info: pub: About to start frontend isolate #1...
2021-03-10 21:48:53.006187 Info: pub: Frontend isolate #1 started.
2021-03-10 21:48:53.028759 Info: pub.static_files: pkg/web_app build status: source: 2021-03-09 22:05:58.000, target: 2021-03-09 22:06:07.000
2021-03-10 21:48:53.039503 Info: pub.static_files: pkg/web_css build status: source: 2021-03-10 19:40:43.000, target: 2021-03-09 22:06:14.000
2021-03-10 21:48:53.043085 Info: pana: Running `pub get`...
2021-03-10 21:48:53.868975 Info: pana: Running `/bin/sh build.sh`...
2021-03-10 21:48:58.647051 Info: pub: Watching for resource changes...
2021-03-10 21:48:58.665627 Info: pub.popularity: Loading popularity data (init): gs://dartlang-pub-dev-306903--popularity/v3/popularity.json.gz
2021-03-10 21:48:59.388511 Info: pub.popularity: Popularity updated for 1 packages.
2021-03-10 21:48:59.392941 Info: pub.name_tracker: Scanning existing package names
2021-03-10 21:49:00.425466 Info: pub.name_tracker: Scanned initial package names (2) in 0:00:01.031423.
2021-03-10 21:49:00.425727 Info: pub.name_tracker: Monitoring new package creation....
```

Please consider it as high risk, as I've no idea if I break system in other place

n0npax:
CC: @isoos

app/config/dev-config-n0npax.yaml
```diff
@@ -0,0 +1,34 @@
+# dev config
```
isoos, line 1: we don't want to store developer-specific configuration in the repo, please remove this

app/config/dev-config-pub.yaml
```diff
@@ -0,0 +1,36 @@
+projectId: dartlang-pub-dev
```
isoos, line 1: Let's add this in two phase: in this PR we keep the hard-coded values in the `configuration.dart` as is, and also add this file + an additional test to check that the content of the file is indeed the same as the value in `configuration.dart`.

In a follow-up PR we can remove/cleanup the parts in `configuration.dart`.

isoos, line 1: Let's call this staging-config.yaml

app/config/dev-config.yaml
```diff
@@ -0,0 +1,34 @@
+# dev config
```
isoos, line 1: what's the difference between this and `dev-config-pub.yaml`?

app/config/prod-config.yaml
```diff
@@ -0,0 +1,38 @@
+# Configuration for production deployment.
```
isoos, line 1: Let's add this in two phase: in this PR we keep the hard-coded values in the configuration.dart as is, and also add this file + an additional test to check that the content of the file is indeed the same as the value in configuration.dart.

In a follow-up PR we can remove/cleanup the parts in configuration.dart.

app/lib/shared/configuration.dart
```diff
@@ -384,6 +306,7 @@ class EnvConfig {
       Platform.environment['TOOL_PREVIEW_FLUTTER_SDK'],
       frontendCount,
       workerCount,
+      Platform.environment['PUB_CONFIG'],
```
isoos, line 309: Let's call this `PUB_SITE_CONFIG` or `PUB_SERVER_CONFIG` in case we may have a similar config for the `pub` command line tool.

app/lib/shared/configuration.dart
```diff
@@ -253,13 +173,11 @@ class Configuration {
 
   /// Create a configuration based on the environment variables.
   factory Configuration.fromEnv(EnvConfig env) {
-    if (env.gcloudProject == 'dartlang-pub') {
-      return Configuration._prod();
-    } else if (env.gcloudProject == 'dartlang-pub-dev') {
-      return Configuration._dev();
-    } else {
-      throw Exception('Unknown project id: ${env.gcloudProject}');
+    if (File(env.configPath).existsSync()) {
```
isoos, line 176: Let's keep the defaults for `env.gcloudProject == ''`, as in the current method we are using the same `Dockerfile` and setup for both staging and prod. This may change in the future, but let's not change too many things at once.

app/lib/shared/configuration.dart
```diff
@@ -331,6 +249,9 @@ class Configuration {
 
 /// Configuration from the environment variables.
 class EnvConfig {
+  // Config Path points to configuratio file
```
isoos, line 252: The constructor parameter is the last one, while this is the first field. Let's synchronize (e.g. move this to the last field).

app/test/shared/configuration.dart
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+import 'package:pub_dev/shared/configuration.dart';
+
+void main() {
+  test('Foo config from yaml file', () async {
```
isoos, line 10: Let's also add another test:
- read the yaml file into `Configuration`
- serialize to JSON string, parse again
- serialize to JSON again and check if it is the same as the above.

app/test/shared/configuration.dart
```diff
@@ -0,0 +1,24 @@
...
+    final Configuration config =
```
isoos, line 11: Nit: no need for the type here.

app/pubspec.lock
```diff
@@ -7,7 +7,7 @@ packages:
       name: _discoveryapis_commons
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "0.2.0"
+    version: "0.2.1"
```
isoos, line 10: Please don't change the resolved versions.

app/lib/shared/configuration.dart
```diff
@@ -130,103 +137,16 @@ class Configuration {
...
-          permissions: AdminPermission.values,
-        ),
-        AdminId(
-          oauthUserId: '117672289743137340098',
-          email: 'assigned-tags-admin@dartlang-pub-dev.iam.gserviceaccount.com',
-          permissions: {AdminPermission.manageAssignedTags},
-        )
-      ],
-    );
+  factory Configuration.fromYaml(final String path) {
```
isoos, line 144: `fromYaml` -> `fromYamlFile`

app/lib/shared/configuration.dart
```diff
@@ -130,103 +137,16 @@ class Configuration {
...
+    final file = File(path);
+    final content = file.readAsStringSync();
+    final map =
+        json.decode(json.encode(loadYaml(content))) as Map<String, dynamic>;
+    return _$ConfigurationFromJson(map);
```
isoos, line 149: Call `Configuration.fromJson` instead, we are using that pattern in other places.

isoos (changes requested):

app/config/dev-config-n0npax.yaml
n0npax: sorry, I thought I removed this file, my bad

app/config/dev-config.yaml
n0npax: one is exposing everything only to localhost, the second one is using also appengine:
https://github.com/dart-lang/pub-dev/pull/4588/files#diff-b9dd2d6cb9598fc3f73907fc582d6d8e9f9efdf89438e2bbf646b2ab658031c5R20
lint 20 in example

app/lib/shared/configuration.dart
```diff
@@ -130,12 +142,47 @@ class Configuration {
...
+    return Configuration.fromJson(map);
+  }
+
   /// Create a configuration for production deployment.
   ///
   /// This will use the Datastore from the cloud project and the Cloud Storage
   /// bucket 'pub-packages'. The credentials for accessing the Cloud
   /// Storage is retrieved from the Datastore.
-  factory Configuration._prod() {
+  factory Configuration.prodConfig() {
```
n0npax, line 185: the method name was changed to simplify access from tests.

n0npax:
Sorry, it took soo long, but it was a crazy week at work,
Hope all issues are addressed

app/lib/shared/configuration.dart
```diff
@@ -2,11 +2,17 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:convert' show json;
 import 'dart:io';
 
 import 'package:collection/collection.dart' show UnmodifiableSetView;
+import 'package:equatable/equatable.dart';
```
isoos, line 9: Let's not add this dependency, we can use the serialized JSON string in the test.

isoos (changes requested):

app/lib/shared/configuration.dart
n0npax: ok, done :)

app/lib/shared/configuration.dart
```diff
@@ -130,12 +141,24 @@ class Configuration {
   /// The identifier of admins.
   final List<AdminId> admins;
 
+  factory Configuration.fromJson(Map<String, dynamic> json) =>
+      _$ConfigurationFromJson(json);
+  Map<String, dynamic> toJson() => _$ConfigurationToJson(this);
```
isoos, line 146: Nit: let's move this below the constructors.

app/lib/shared/configuration.dart
isoos: Let's add @visibleForTesting annotation to it.

app/lib/shared/configuration.dart
```diff
@@ -180,7 +203,7 @@ class Configuration {
   }
 
   /// Create a configuration for development/staging deployment.
-  factory Configuration._dev() {
+  factory Configuration.devConfig() {
```
isoos, line 206: Let's add @visibleForTesting annotation to it.

app/lib/shared/configuration.dart
```diff
@@ -130,12 +141,24 @@ class Configuration {
...
+    return Configuration.fromJson(map);
+  }
+
   /// Create a configuration for production deployment.
   ///
   /// This will use the Datastore from the cloud project and the Cloud Storage
   /// bucket 'pub-packages'. The credentials for accessing the Cloud
   /// Storage is retrieved from the Datastore.
-  factory Configuration._prod() {
+  factory Configuration.prodConfig() {
```
isoos, line 161: Nit: maybe call this `Configuration.prod`

app/lib/shared/configuration.dart
isoos, line 206: Nit: maybe call this `Configuration.staging`

app/lib/shared/configuration.dart
```diff
@@ -254,11 +277,19 @@ class Configuration {
...
   factory Configuration.fromEnv(EnvConfig env) {
     if (env.gcloudProject == 'dartlang-pub') {
-      return Configuration._prod();
+      return Configuration.prodConfig();
     } else if (env.gcloudProject == 'dartlang-pub-dev') {
-      return Configuration._dev();
+      return Configuration.devConfig();
     } else {
-      throw Exception('Unknown project id: ${env.gcloudProject}');
+      if (env.configPath.isEmpty) {
```
isoos, line 284: use `else if`(s) at a higher level and then another final `else`

isoos, line 284: Also check if env.configPath is null.

app/lib/shared/configuration.dart
```diff
@@ -395,13 +431,22 @@ class EnvConfig {
...
 class AdminId {
   final String oauthUserId;
   final String email;
 
   /// A set of strings that determine what operations the administrator is
   /// permitted to perform.
   final Set<AdminPermission> permissions;
+  factory AdminId.fromJson(Map<String, dynamic> json) =>
+      _$AdminIdFromJson(json);
+  Map<String, dynamic> toJson() => _$AdminIdToJson(this);
```
isoos, line 449: Nit: move this below the constructors.

isoos (changes requested):
Looking really close for completion, I'd like to ask for a few more changes though..

n0npax:
@isoos can I ask for a re-review, please?

isoos:
I think this looks good. /cc @jonasfj for further comments.

n0npax:
@jonasfj can I ask for a review, please?

n0npax:
@jonasfj can I ask for a review please? 🙏🏻

isoos:
@n0npax: IIRC we had an update in the prod config, could you please re-do the yaml and trigger the tests?

n0npax:
sure thing!

```zsh
dart test test/shared/configuration.dart -r expanded
00:00 +0: Foo config from yaml file
00:00 +1: Prod config from yaml file
00:00 +2: Staging config from yaml file
00:00 +3: All tests passed!
```

n0npax:
@jonasfj can I ask for a review, please?

n0npax:
@jonasfj can I ask for a review, please 🙏🏻

jonasfj (approved):
r+

jonasfj:
Let's land once todays deployment has gone through..

isoos (approved):

PR #4593, opened 2021-03-09
First attempt at LatestReleases

isoos:
- This is the very first step towards runtime-versioned previewVersion calculation.
- It seems that many parts of the codebase depend on these values, and it seems to be better to first refactor them to use the same access pattern.
- `Release` may contain the SDK constraint info later - if needed.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -175,6 +175,12 @@ class PackageBackend {
         .cast();
   }
 
+  /// Returns the latest releases info of a package.
+  Future<LatestReleases> latestReleases(Package package) async {
+    // TODO: implement runtimeVersion-specific release calculation
+    return package.latestReleases;
+  }
+
   /// Looks up a specific package version.
   ///
   /// Returns null if the version is not a semantic version or if the version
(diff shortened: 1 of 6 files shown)


app/lib/package/backend.dart
```diff
@@ -175,6 +175,12 @@ class PackageBackend {
         .cast();
   }
 
+  /// Returns the latest releases info of a package.
+  Future<LatestReleases> latestReleases(Package package) async {
+    // TODO: implement runtimeVersion-specific release calculation
+    return package.latestReleases;
```
jonasfj, line 181: So you're thinking we'll start storing `LatestReleases` as a runtimeVersioned entity? or...

I'm mostly wondering if we should make a property on the `Package` entity that is `runtimeVersioned` or a new entity that is runtimeVerisoned..

The new entity is certainly something we're already pretty skilled at doing.

isoos: > So you're thinking we'll start storing LatestReleases as a runtimeVersioned entity? or...

I think last time we've discussed this, we've considered using a map on the `Package` like this:
```
{
  '2021.02.02': {'stable': '1.0.0', 'preview': '2.0.0'},
  '2021.03.25': {'stable': '2.0.0'}
}
```

However, I think the separate entity has better isolation between releases, and may be worth doing it that way. I'd still use this new structure in both cases.

jonasfj: Okay, I'm down with keeping it in a separate entity, so long as we make sure to cache it whenever we load it :D

jonasfj (approved):

PR #4600, opened 2021-03-10
Fix GA events when using GTM

jonasfj:


static/js/gtm.js
```diff
@@ -8,38 +8,20 @@ window.dataLayer.push({
...
-    // https://support.google.com/analytics/answer/7478520?hl=en
-    if (elem.hasAttribute('href')) {
-      var updated = false;
-      var callbackFn = function () {
-        if (updated) return;
-        updated = true;
-        document.location = elem.href;
-      };
-      data.event_callback = callbackFn;
-      data.transport_type = 'beacon';
```
isoos, line 35: Removing this will break the tracking on links, wouldn't it?

jonasfj: Yeah, I think it might... but that doesn't work anymore regardless..

jonasfj: yes

isoos (approved):
You'll need to rebase to fix the versions test too.

isoos (approved):

Issue #4440, opened 2021-01-25
Migrate tar usage to package:tar


jonasfj:
I've done a decent review of https://pub.dev/packages/tar, it's all streaming in-memory.. Reasonably easy to use.
I propose that we:
 * [ ] Verify all tar archives on pub.dev can be read with `package:tar`
 * [ ] Migrate pub.dev to use `package:tar` read incoming archives, rejecting all archives that can be parsed by `package:tar`.

We could compare incoming archives against GNU tar too, to test for correctness in `package:tar`, but it looks reasonably well tested, so I propose we instead fix issues when users report them to us. So long as our rejection message about whatever was uploaded being an invalid tar archive gives users a pointer to where bugs should be reported we can roll `package:tar` serverside and undo it if something goes wrong.

isoos:
Work on this is blocked on `package:tar`'s dependency on Dart SDK 2.12, pub-dev is still on 2.10.

PR #4601, opened 2021-03-11
Use package:tar in test-profile and pkg/pub_package_reader's test.

isoos:
- #4440
- new interface for the `TarArchive` and separate implementation to use `package:tar`
- using it in `test-profile` to create the archive
- using it in `pkg/pub_package_reader` in the test
- also fixed a bug that didn't restricted the max content length on asset extraction


jonasfj (approved):
Is `useNative` just a temporary thing we remove later?

isoos:
> Is useNative just a temporary thing we remove later?

Yes, that's the plan. I'd keep two implementation while we verify that it is indeed identical, and then switch and remove the non-Dart version.

Issue #2142, opened 2019-03-21
Document API end-points, JSON schemas and stability expectations


jonasfj:
We should write a markdown document that documents APIs we offer, and outline the stability of these APIs.

Notably:
 * Get versions of a package (stable)
 * Get a package (stable)
 * Search for packages with JSON API (might be experimental)

Do we have other APIs that we want to expose to the world?

Note. Maybe we can declare the search API stable, it would certainly be useful to be able to list all packages and things like that. We don't have to make all features of the search API stable, ie. all filters that can be applied (I could imagine those changing over time).

isoos:
We do have a few more, I'll need to go through the code to get all of it.

I'm wondering if this could also be reflected in (a) tests, or documentation for tests, and (b) in the code that defines the routes. Assuming we migrate to use `shelf_router`, would it make sense to annotate the routes with something that reflects their stability?

jonasfj:
> Assuming we migrate to use shelf_router, would it make sense to annotate the routes with something that reflects their stability?

Yes, it would also make sense to write documentation comments that can be extracted into a reference for the API... But this all requires a lot of tooling, let's take on step at the time.

> We do have a few more, I'll need to go through the code to get all of it.

We don't have to document all APIs, ideally we document those that external parties depend on. Or the ones we want to offer people to program against. It's still okay to have internal API end-points for which we make no stability promises.

jingtuo:
> * experimental
Does the search API have more parameters, such as the number of pages per page, filtering packages like dart-io?

For example, if I search for json, I don't want to see dart:io in the results.

svenjacobs:
Is there an API endpoint for fetching the changelog(s) of a package?

isoos:
@svenjacobs: There is none at the moment (unless you count `/packages/<pkg>/changelog`). Curious: what's the use case you are looking for?

svenjacobs:
> Curious: what's the use case you are looking for?

I'm thinking of writing a graphical version of `pub outdated` that also displays all changelogs of outdated library versions starting from _current_ version (+1) to _upgradable_ giving a nice overview of what has changed.

sigurdm:
We have pub.dev/help/api

PR #4628, opened 2021-03-16
Documenting public APIs

isoos:
#2142

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -1,0 +1,70 @@
+# `pub.dev` API for developers
+
+## Stay informed
+
+This document describes the officially supported API of the `pub.dev` site.
+`pub.dev` may expose API endpoints that are available publicly, but unless
+they are documented here, we don't consider them as officially
+supported, and may change or remove them without any notice.
+
+Changes to the officially supported `pub.dev` API will be announced at [Dart announce][announce].
+
+## Hosted Pub Repository API
+
+`pub.dev` implements the [Hosted Pub Repository Specification V2][repo-v2],
+used by the `pub` command line client application.
+
+## Additional APIs
+
+### Package names for name completion
+
+**GET** `https://pub.dev/api/package-name-completion-data`
+
+**Headers:**
+* `accept-encoding: gzip`
+
+**Response**
+* `cache-control: public, max-age=28800`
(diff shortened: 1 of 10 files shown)


doc/api.md
```diff
@@ -0,0 +1,70 @@
...
+  "packages": [
+    "http",
+    "provider",
+    /* further package names */
+  ]
+}
+```
+
+The API returns the top package names on `pub.dev`. At the moment we
+limit the result length at the maximum of 20000 package names, which
```
sigurdm, line 42:
```suggestion
limit the result length at the maximum of 20000 package names, this
```

doc/api.md
```diff
@@ -0,0 +1,70 @@
...
+limit the result length at the maximum of 20000 package names, this
+limit may change in the future.
+
+The response is always a gzip-ed JSON content, and should be cached
+on the client side  for at least 8 hours between requests (as indicated
+by the `cache-control` header).
+
+Notes:
+ * Not all package names are included in this response.
+ * The inclusion criteria used by `pub.dev` may change without any early notice.
```
jonasfj, line 51:
```suggestion
The API returns the top package names on `pub.dev`. To reduce payload size the
result may not include all package names. The size limitation is subject to change.

The response is always a gzip-ed JSON content, and should be cached
on the client side  for at least 8 hours between requests (as indicated
by the `cache-control` header).

Notes:
 * Not all package names are included in this response.
 * The inclusion criteria used by `pub.dev` may change without notice.
```

doc/api.md
```diff
@@ -0,0 +1,70 @@
+# `pub.dev` API for developers
+
+## Stay informed
+
+This document describes the officially supported API of the `pub.dev` site.
+`pub.dev` may expose API endpoints that are available publicly, but unless
+they are documented here, we don't consider them as officially
+supported, and may change or remove them without any notice.
```
jonasfj, line 8:
```suggestion
supported, and may change or remove them without notice.
```

jonasfj (approved):

doc/api.md
jonasfj: Or something like htat... I would suggest avoiding terms like "We"...  And avoid being too specific in describing what the API does.

PR #4629, opened 2021-03-16
Fix URL rewrite in bad markdown.

### LINKED ISSUE #4627: [search] Link rewrite failed
Stack trace from search service:
```
pub.markdown: Link rewrite failed: src="https://github.com/jcxsoftware-dev/delayed_consumer_hud/raw/master/doc/delayed_consumer_hud_demo.gif"

Error:
    FormatException: Illegal scheme character (at character 4)
    src="https://github.com/jcxsoftware-dev/delayed_consumer_hud/raw/master/doc...
       ^
    

Stack:
    #0      _Uri._fail (dart:core/uri.dart:1538:5)
    #1      _Uri._makeScheme (dart:core/uri.dart:2075:9)
    #2      new _Uri.notSimple (dart:core/uri.dart:1369:18)
    #3      Uri.parse (dart:core/uri.dart:956:17)
    #4      _RelativeUrlRewriter._rewriteAbsoluteUrl (package:pub_dev/shared/markdown.dart:252:21)
    #5      _RelativeUrlRewriter._rewriteUrl (package:pub_dev/shared/markdown.dart:240:18)
    #6      _RelativeUrlRewriter.visitElementAfter (package:pub_dev/shared/markdown.dart:226:11)
    #7      Element.accept (package:markdown/src/ast.dart:52:15)
    #8      Element.accept (package:markdown/src/ast.dart:49:17)
    #9      _rewriteRelativeUrls.<anonymous closure> (package:pub_dev/shared/markdown.dart:81:32)
    #10     List.forEach (dart:core-patch/growable_array.dart:403:8)
    #11     _rewriteRelativeUrls (package:pub_dev/shared/markdown.dart:81:9)
    #12     markdownToHtml (package:pub_dev/shared/markdown.dart:52:11)
    #13     compactReadme (package:pub_dev/search/text_utils.dart:44:16)
    #14     SearchBackend.loadDocument (package:pub_dev/search/backend.dart:167:15)
    <asynchronous suspension>
    #15     StackZoneSpecification._registerUnaryCallback.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart)
    <asynchronous suspension>
```

isoos:
- Fixes #4627.
- Urls that do not parse will be removed alongside with the enclosing element.

### DIFF
--- a/app/lib/shared/markdown.dart
+++ b/app/lib/shared/markdown.dart
@@ -209,6 +209,7 @@ class _UnsafeUrlFilter implements m.NodeVisitor {
 class _RelativeUrlRewriter implements m.NodeVisitor {
   final String baseUrl;
   final String baseDir;
+  final _elementsToRemove = <m.Element>{};
   _RelativeUrlRewriter(this.baseUrl, this.baseDir);
 
   @override
@@ -226,8 +227,20 @@ class _RelativeUrlRewriter implements m.NodeVisitor {
       _updateAttributes(element, 'src', raw: true);
     }
     // remove children that are marked to be removed
-    element.children?.removeWhere((n) =>
-        n is m.Element && n.attributes.containsKey('pub-markdown-remove'));
+    if (element.children != null &&
+        element.children.isNotEmpty &&
+        _elementsToRemove.isNotEmpty) {
+      for (final r in _elementsToRemove.toList()) {
+        final index = element.children.indexOf(r);
+        if (index == -1) continue;
+
+        if (r.children != null && r.children.isNotEmpty) {
+          element.children.insertAll(index, r.children);
+        }
+        element.children.remove(r);
+        _elementsToRemove.remove(r);
+      }
+    }
   }
 
   void _updateAttributes(m.Element element, String attrName,
@@ -236,8 +249,7 @@ class _RelativeUrlRewriter implements m.NodeVisitor {
     if (newUrl != null) {
       element.attributes[attrName] = newUrl;
     } else {
-      element.attributes.remove(attrName);
-      element.attributes['pub-markdown-remove'] = 'true';
+      _elementsToRemove.add(element);
     }
   }
 
--- a/app/lib/shared/markdown.dart
+++ b/app/lib/shared/markdown.dart
@@ -219,17 +219,34 @@ class _RelativeUrlRewriter implements m.NodeVisitor {
 
   @override
   void visitElementAfter(m.Element element) {
+    // check current element
     if (element.tag == 'a') {
-      element.attributes['href'] = _rewriteUrl(element.attributes['href']);
+      _updateAttributes(element, 'href');
     } else if (element.tag == 'img') {
-      element.attributes['src'] =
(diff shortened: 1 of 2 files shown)


app/lib/shared/markdown.dart
```diff
@@ -219,18 +219,35 @@ class _RelativeUrlRewriter implements m.NodeVisitor {
...
-      element.attributes['src'] =
-          _rewriteUrl(element.attributes['src'], raw: true);
+      _updateAttributes(element, 'src', raw: true);
+    }
+    // remove children that are marked to be removed
+    element.children?.removeWhere((n) =>
+        n is m.Element && n.attributes.containsKey('pub-markdown-remove'));
+  }
+
+  void _updateAttributes(m.Element element, String attrName,
```
jonasfj, line 233:
```suggestion
  void _rewriteUrlAttribute(m.Element element, String attrName,
```

app/lib/shared/markdown.dart
```diff
@@ -219,18 +219,35 @@ class _RelativeUrlRewriter implements m.NodeVisitor {
...
+      {bool raw = false}) {
+    final newUrl = _rewriteUrl(element.attributes[attrName], raw: raw);
+    if (newUrl != null) {
+      element.attributes[attrName] = newUrl;
+    } else {
+      element.attributes.remove(attrName);
+      element.attributes['pub-markdown-remove'] = 'true';
```
jonasfj, line 240: Would it not be safer to create a `final _elementsToRemove = <m.Element>[]` list on the `_RelativeUrlRewriter` class.

That way we don't abuse attribute names... we just have to make sure never reuse an instance of `_RelativeUrlRewriter`, because we've made it stateful, but that's should be no problem..

Also, shouldn't we replace the element with it's contents... so that the link disappears, but the contents of the link does not?

This only happens when you have invalid markdown? or an invalid URL... right?
Basically, any link with a URL we can't parse will have this property.. so shouldn't we just remove the link and retain the contents?

Example:
```markdown
 * [My awesome thing](http://illegal@@thing)
```

Won't this strip away the text "My awesome thing"? When we probably just want to strip away the `<a>` tag, or perhaps just the `href` attribute.

app/test/shared/markdown_test.dart
```diff
@@ -217,6 +217,11 @@ void main() {
       expect(markdownToHtml('[a][b]', baseUrl: 'http://www.example.com/'),
           '<p>[a][b]</p>\n');
     });
+
+    test('bad image link with attribute', () {
+      expect(markdownToHtml('![demo](src="https://github.com/a/b/c.gif")'),
+          '<p></p>\n');
+    });
```
jonasfj, line 224: `Uri.tryParse("http://illegal@@thing") == null`

So another test case might be:
```markdown
[my illegal url](http://illegal@@thing)
```

app/lib/shared/markdown.dart
isoos: Goog point, I've updated the code and the tests to match this. PTAL

jonasfj (approved):

PR #4646, opened 2021-03-19
Throw proper error when pubspec.yaml can't be converted to json

### LINKED ISSUE #3692: Forbid uploads of packages with pubspec.yaml that can't be JSON
YAML can contain values that cannot be presented in JSON.
```
mykey:
  [1,2,3,4,5]: 'value of a composite key'
```
is a good example.

As custom keys are allowed in `pubspec.yaml` we need to:
 * Check that the `pubspec.yaml` can be represented as JSON, and,
 * Have a test case that publishing such a package will fail.

Today, this will presumably throw an unhandled exception when we try to convert the pubspec into JSON. This might be preferable to test this, and convert the exception into something more human friendly.

ramyak-mehra:
Fixes #3692

### DIFF
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -8,11 +8,14 @@ import 'package:meta/meta.dart';
 import 'package:pub_package_reader/src/yaml_utils.dart';
 import 'package:pubspec_parse/pubspec_parse.dart';
 import 'package:pub_semver/pub_semver.dart';
-import 'package:yaml/yaml.dart' show YamlException;
+import 'package:yaml/yaml.dart' show YamlException, loadYaml;
+import 'package:logging/logging.dart';
 
 import 'src/file_names.dart';
 import 'src/names.dart';
 import 'src/tar_utils.dart';
+
+final _logger = Logger('pub_package_reader');
 
 /// A validation issue in the package archive.
 class ArchiveIssue {
@@ -99,7 +102,6 @@ Future<PackageSummary> summarizePackageArchive(
   } on Exception catch (e) {
     issues.add(ArchiveIssue('Error parsing pubspec.yaml: $e'));
   }
-
   // Try again with lenient parsing.
   try {
     pubspec ??= Pubspec.parse(pubspecContent, lenient: true);
@@ -107,7 +109,8 @@ Future<PackageSummary> summarizePackageArchive(
     issues.add(ArchiveIssue('Error parsing pubspec.yaml: $e'));
     return PackageSummary(issues: issues);
   }
-
+  // Check wether the pubspecContent can be converted to JSON
+  issues.addAll(checkValidJson(pubspecContent));
   // Check whether the files can be extracted on case-preserving file systems
   // (e.g. on Windows). We can't allow two files with the same case-insensitive
   // name.
@@ -330,6 +333,18 @@ Iterable<ArchiveIssue> forbidGitDependencies(Pubspec pubspec) sync* {
         'Package dependency $name depends on a package with a different name',
       );
     }
+  }
+}
+
+Iterable<ArchiveIssue> checkValidJson(String pubspecContent) sync* {
+  try {
+    final map = loadYaml(pubspecContent) as Map;
+    json.decode(json.encode(map)) as Map<String, dynamic>;
+  } on JsonUnsupportedObjectError catch (_) {
+    yield ArchiveIssue(
+        'pubspec.yaml contains values that can\'t be converted to JSON.');
+  } on Exception catch (e, st) {
+    _logger.warning('Error while converting pubspec.yaml to JSON', e, st);
   }
 }
 
--- a/pkg/pub_package_reader/test/package_archive_test.dart
+++ b/pkg/pub_package_reader/test/package_archive_test.dart
@@ -483,4 +483,31 @@ $dependencies
       expect(requireNonEmptyLicense('LICENSE', 'BSD license'), isEmpty);
     });
   });
+
+  group('pubspec.yaml to json conversion', () {
+    test('pubspec.yaml to json conversion valid', () {
+      final pubspec = '''
+      name: provider     
+      environment:
+        sdk: ">=2.0.0 <3.0.0"
+      dependencies:
+        flutter:
+          sdk: flutter
+      ''';
+      expect(checkValidJson(pubspec), isEmpty);
+    });
+    test('pubspec.yaml to json conversion invalid', () {
+      final pubspec = '''
+      name: provider    
+      environment:
+        sdk: ">=2.0.0 <3.0.0"
+      dependencies:
+        flutter:
+          sdk: flutter
+      mykey:
+        [1,2,3,4,5]: 'value of a composite key'
+      ''';
+      expect(checkValidJson(pubspec), isNotEmpty);
+    });
+  });
 }


ramyak-mehra:
@googlebot I signed it!

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -177,6 +178,7 @@ Future<PackageSummary> summarizePackageArchive(
   //       https://github.com/dart-lang/pub/issues/2557
   // issues.addAll(forbidPreReleaseSdk(pubspec));
   issues.addAll(requireIosFolderOrFlutter2_20(pubspec, tar.fileNames));
+  issues.addAll(checkValidJson(pubspecContent));
```
isoos, line 181: Let's move this to line 99, after the pubspec checks.

pkg/pub_package_reader/test/package_archive_test.dart
```diff
@@ -483,4 +483,59 @@ $dependencies
       expect(requireNonEmptyLicense('LICENSE', 'BSD license'), isEmpty);
     });
   });
+
+  group('pubspec.yaml to json conversion', () {
+    test('pubspec.yaml to json conversion valid', () {
+      final pubspec = '''
```
isoos, line 489: Let's make these examples smaller. No need for version, authors, too many dependencies...

isoos (changes requested):

isoos:
@ramyak-mehra In general this looks good, only small nits to update.

ramyak-mehra:
@isoos  I think we need to update this [test](https://github.com/ramyak-mehra/pub-dev/blob/097741b44e6c1892b442b0b76501ee4b080f13c4/app/test/package/upload_test.dart#L381)

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -321,6 +322,16 @@ Iterable<ArchiveIssue> forbidGitDependencies(Pubspec pubspec) sync* {
   }
 }
 
+Iterable<ArchiveIssue> checkValidJson(String pubspecContent) sync* {
+  try {
+    final map = loadYaml(pubspecContent) as Map;
+    json.decode(json.encode(map)) as Map<String, dynamic>;
+  } on JsonUnsupportedObjectError catch (_) {
+    yield ArchiveIssue(
+        'pubspec.yaml contains values that can\'t be converted to json.   ');
```
isoos, line 331: Nit: remove extra spaces at the end.
Nit: replace `json` to `JSON`.

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -87,7 +88,7 @@ Future<PackageSummary> summarizePackageArchive(
   } on Exception catch (e) {
     issues.add(ArchiveIssue('Error parsing pubspec.yaml: $e'));
   }
-
+  issues.addAll(checkValidJson(pubspecContent));
```
isoos, line 91: What happens with the failing test if you move this to the line 99?

isoos (commented):
The failing test seems to be weird, it should throw an exception, but maybe an exception message would be different there.

pkg/pub_package_reader/lib/pub_package_reader.dart
ramyak-mehra: The test passed, but it was weird.

ramyak-mehra:
@isoos  I think we can get away with this fail or should I update the flutter version to 2.0.3. It was just updated.

isoos (commented):
LGTM, let's also wait for @jonasfj to review. The pending test will be fixed with #4643.

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -321,6 +321,16 @@ Iterable<ArchiveIssue> forbidGitDependencies(Pubspec pubspec) sync* {
...
 }
 
+Iterable<ArchiveIssue> checkValidJson(String pubspecContent) sync* {
+  try {
+    final map = loadYaml(pubspecContent) as Map;
+    json.decode(json.encode(map)) as Map<String, dynamic>;
+  } on JsonUnsupportedObjectError catch (_) {
+    yield ArchiveIssue(
+        'pubspec.yaml contains values that can\'t be converted to JSON.');
+  }
```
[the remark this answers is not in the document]
jonasfj: Actually, I like catching `JsonUnsupportedObjectError`.

But having a ` } catch (e) {` block which catches everything might also be nice...
We can make the two cache blocks do the same, but have the catch-all block print a warning using `package:logging`.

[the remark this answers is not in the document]
ramyak-mehra: First, I will catch JsonUnsupportedObjectError, and if there's any other ill add it to the logging. does this work?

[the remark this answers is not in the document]
ramyak-mehra: > This is a bit of a degenerate way of doing it... let's maybe catch any error :D
> 
> Also it might be interesting to write a one-off script and run this against all packages / all pubspec.yaml on pub.dev
> 
> * Packages names are available here: https://pub.dev/api/package-names
> * Find versions and tarballs for a package here: https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md

is there a constraint on using a specific language?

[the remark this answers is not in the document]
ramyak-mehra: @jonasfj I Wrote this script to download packages and validate them [script](https://gist.github.com/ramyak-mehra/802d60811914dd0b6472b4e11880d564) 
I think I  should use isolates to speed up the process but I am not sure how should I implement them. Possible implementations I can think of : 

- The isolate downloads the tar.gz files. Extract the pubspec content and returns it. Then the main thread validates the yaml and logs the entry somewhere.
- Any other ? happy to know.
- Completely ditch the isolates?

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -7,12 +7,14 @@ import 'dart:convert';
 import 'package:meta/meta.dart';
 import 'package:pubspec_parse/pubspec_parse.dart';
 import 'package:pub_semver/pub_semver.dart';
-import 'package:yaml/yaml.dart' show YamlException;
-
+import 'package:yaml/yaml.dart' show YamlException, loadYaml;
+import 'package:logging/logging.dart';
 import 'src/file_names.dart';
```
isoos, line 12: Nit: extra line before the non-package imports

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -321,6 +323,18 @@ Iterable<ArchiveIssue> forbidGitDependencies(Pubspec pubspec) sync* {
...
 
+Iterable<ArchiveIssue> checkValidJson(String pubspecContent) sync* {
+  try {
+    final map = loadYaml(pubspecContent) as Map;
+    json.decode(json.encode(map)) as Map<String, dynamic>;
+  } on JsonUnsupportedObjectError catch (_) {
+    yield ArchiveIssue(
+        'pubspec.yaml contains values that can\'t be converted to JSON.');
+  } on Exception catch (e) {
+    _logger.warning('Error while converting pubspec.yaml to JSON: $e');
```
isoos, line 334:
```dart
 } on Exception catch (e, st) {
    _logger.warning('Error while converting pubspec.yaml to JSON', e, st);
```

isoos:
@ramyak-mehra: please also rebase the PR on the master branch

pkg/pub_package_reader/lib/pub_package_reader.dart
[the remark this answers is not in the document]
isoos: Let's also add this script into this PR!
It could be under `pkg/pub_package_reader/tool/verify_pub_dev.dart`.

jonasfj (approved):

isoos (approved):

ramyak-mehra:
@isoos  I have not added this script in the PR https://gist.github.com/ramyak-mehra/802d60811914dd0b6472b4e11880d564
It's incomplete, I could complete this today just need to know how should I handle the fault packages, should I just log them or save the result in a file about all the packages?

isoos:
> It's incomplete, I could complete this today just need to know how should I handle the faulty packages, should I just log them or save the result in a file about all the packages?

We are planning to do a backward-compatibility test of the existing archives in the near future, this shall be automatically part of it, there is no need for a separate script.

Issue #4663, opened 2021-03-25
pub.dev is down (global outage)


timsneath:
![image](https://user-images.githubusercontent.com/2319867/112408932-da1a2000-8cd5-11eb-905b-a55b0a986ed0.png)

## Overview
At approximately 7pm, the pub infrastructure began to respond with a **HTTP 502 Site Error** message. The site infra is deployed to the `us-central` GCP region. This impacts core Flutter services, specifically:

- The pub.dev website itself, which presents the error above;
- Calls made by the `flutter` or `pub` commands (e.g. `flutter upgrade` or `pub upgrade`). 

Please do not reply to this bug with "me too" or +1 messages, it makes it harder for folk to track. Thanks!

## Updates
**7:10pm Pacific**
We're currently experiencing an outage on pub.dev, which appears to be related to a load balancer issue. We don't have an ETA for a resolution at this time; we're currently working to understand the issue.

**7:59pm Pacific**
The pub.dev site is still down. We have a Google on-call engineer currently investigating. We have not yet identified a root cause. 

**8:15pm Pacific**
We apologize for the inconvenience. We're seeing load balancer errors and are escalating to the appropriate team. Still no ETA, unfortunately, since we're still haven't determined the root cause.  

**8:27pm Pacific**
We have multiple Google Cloud engineers on-call investigating, but I'm sorry to report that we still don't have a root cause. We'll continue to post updates regularly. Thank you for your patience.

**9:00pm Pacific**
We are continuing to debug the problem. We have declared a Google escalated outage while we attempt to identify the root cause. Some folk have been successful using the Chinese mirror site at https://pub.flutter-io.cn. 

**9:20pm Pacific**
Again, apologies. 

**9:35pm Pacific**
We are currently exploring the theory that we have exceeded a quota, but that the error didn't show in the log. Paging an oncall team to try and increase the quota to see if this resolves. Again, this really sucks -- we recognize that it's a major inconvenience to you all, and we're feeling sick that we're down. Thank you for being patient with us :(

**9:45pm Pacific**
We have updated the quota and are resetting the VM instances, to see if we have successfully identified the root cause. 

**9:51pm Pacific**
We are seeing evidence of partially restored service.

**9:55pm Pacific**
The pub service appears to be fully restored.

**10:15pm Pacific**
[...]

PR #4672, opened 2021-03-26
Do not kill frontend isolate on uncaught exceptions.

isoos:
- #4663
- uncaught exceptions are only logged
- using `stderr` instead of `print` to make these errors stand out from the rest
- if the frontend isolate exits for a reason, it still restarts it with an ever-increasing delay

### DIFF
--- a/app/lib/service/entrypoint/_isolate.dart
+++ b/app/lib/service/entrypoint/_isolate.dart
@@ -78,8 +78,9 @@ Future startIsolates({
 
     int frontendStarted = 0;
 
-    /// The timestamp until errors won't cause frontend isolates to restart.
-    var frontendRestartProtection = DateTime.now();
+    /// The duration while errors won't cause frontend isolates to restart.
+    var restartProtectionOffset = Duration.zero;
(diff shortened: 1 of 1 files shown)


jonasfj:
Let's discuss this next week.. I see pros/cons I'm both directions.

For example: if some internal thing is failing persistently, and this causes 5xx responses, then it might be preferable to restart.

An unhandled exception at this levels typically implies undefined behavior, crashing and restarting is reliable way to recover.

Maybe we want to limit number of times we allow a restart. Like restart on first error, then avoid restarting due to unhandled errors for 5, 10, 20min, or something like that. That would degrade the service more gracefully.

isoos:
> Like restart on first error, then avoid restarting due to unhandled errors for 5, 10, 20min, or something like that. That would degrade the service more gracefully.

Done, with the `frontendStarted` as controlling for time. On the default service we'll have 4 isolates, so it will start by 4 min protection, then 5 minutes, ... up to 20 minutes maximum.

jonasfj (approved):
Just a suggestion... Overall this looks good.

PR #4674, opened 2021-03-26
Review of Timers and proposal of cached_value.dart

isoos:
- #4663
- `cached_value.dart` is a proposed pattern to handle synchronously accessed values with periodic updates and local cache.
- The pattern could be also used for slightly similar task like periodically saving the search snapshot.
- Identified uses of `Timer`s in the code, added a test to track them and marked the required migrations.
- Removed `Timer` use in `task_sources.dart`, as the log entry was used to track a scheduling issue which was since resolved.
- Moved event loop tracking outside of `utils.dart`, in order to prevent accidental use of `Timer` in `shared/utils.dart`.

### DIFF
--- a/app/lib/shared/cached_value.dart
+++ b/app/lib/shared/cached_value.dart
@@ -57,10 +57,11 @@ class CachedValue<T> {
   Duration get age => DateTime.now().difference(_lastUpdated);
   bool get isAvailable => _value != null && age <= _maxAge;
 
+  /// The cached value, may be null.
   T get value => _value;
 
   @visibleForTesting
-  set value(T v) {
+  void setValue(T v) {
     _value = v;
     _lastUpdated = DateTime.now();
   }
--- a/app/lib/shared/cached_value.dart
+++ b/app/lib/shared/cached_value.dart
@@ -7,6 +7,8 @@ import 'dart:async';
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 
+import 'monitoring.dart';
+
 final _logger = Logger('cached_value');
 
 typedef UpdateFn<T> = Future<T> Function();
@@ -89,9 +91,11 @@ class CachedValue<T> {
       }
     } catch (e, st) {
       if (age <= _maxAge) {
-        _logger.warning('Updating cached `$_name` value failed.', e, st);
+        _logger.pubWarning(
+            'cached_value', 'Updating cached `$_name` value failed.', e, st);
       } else {
-        _logger.shout('Updating cached `$_name` value failed.', e, st);
+        _logger.pubShout(
+            'cached_value', 'Updating cached `$_name` value failed.', e, st);
       }
     } finally {
       final c = _ongoingCompleter;
--- a/app/lib/shared/cached_value.dart
+++ b/app/lib/shared/cached_value.dart
@@ -67,6 +67,7 @@ class CachedValue<T> {
   }
 
   /// Updates the cached value.
+  @visibleForTesting
   Future<void> update() async {
     if (_closing) {
       throw StateError('Cache `$_name` is already closed.');
@@ -99,9 +100,18 @@ class CachedValue<T> {
     }
   }
 
-  void scheduleUpdates() {
+  /// Starts a periodic Timer to update the cached value.
+  ///
+  /// If this is the first call of [start], and initial value
+  /// was not set, this will also call [update].
+  Future<void> start() async {
     if (_closing) {
       throw StateError('Cache `$_name` is already closed.');
+    }
+    if (!isAvailable && _timer == null) {
+      await _update();
+      // ignore: invariant_booleans
+      if (_closing) return;
     }
     _timer ??= Timer.periodic(_interval, (timer) {
       _update();
--- a/app/lib/shared/cached_value.dart
+++ b/app/lib/shared/cached_value.dart
@@ -1,0 +1,75 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
+import 'package:retry/retry.dart';
+
+final _logger = Logger('cached_value');
+
+typedef UpdateFn<T> = Future<T> Function();
+
+/// A locally cached and periodically updated value that can be accessed
+/// synchronously.
+class CachedValue<T> {
+  final UpdateFn<T> _updateFn;
+  final RetryOptions _retryOptions;
+  Timer _timer;
+  T _value;
+  Completer _ongoingCompleter;
+  bool _initialized = false;
+
+  CachedValue({
+    @required UpdateFn<T> fn,
+    RetryOptions retryOptions,
+  })  : _updateFn = fn,
+        _retryOptions = retryOptions ??
+            RetryOptions(
+              delayFactor: Duration(seconds: 3),
+              maxAttempts: 3,
+            );
+
+  bool get isInitialized => _initialized;
+
+  T get value => _value;
+  set value(T v) {
+    _value = v;
+    _initialized = true;
+  }
+
+  /// Updates the cached value.
+  Future<void> update() async {
+    if (_ongoingCompleter != null) {
+      await _ongoingCompleter.future;
+      return;
+    }
+    _ongoingCompleter = Completer();
+    try {
+      await _retryOptions.retry(() async {
+        _value = await _updateFn();
+        _initialized = true;
+      });
+    } catch (e, st) {
+      _logger.warning('Updating cached value failed.', e, st);
+    } finally {
+      final c = _ongoingCompleter;
+      _ongoingCompleter = null;
+      c.complete();
+    }
+  }
+
+  void scheduleUpdates(Duration duration) {
+    _timer?.cancel();
+    _timer = Timer.periodic(duration, (timer) {
+      update();
+    });
+  }
+
+  Future<void> close() async {
+    _timer?.cancel();
+    _timer = null;
+  }
+}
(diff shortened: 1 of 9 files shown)


app/lib/shared/cached_value.dart
```diff
@@ -0,0 +1,75 @@
...
+        _retryOptions = retryOptions ??
+            RetryOptions(
+              delayFactor: Duration(seconds: 3),
+              maxAttempts: 3,
+            );
+
+  bool get isInitialized => _initialized;
+
+  T get value => _value;
+  set value(T v) {
```
jonasfj, line 38: Can we make the setter a method, and annotated it `visibleForTestingOnly`?

app/lib/shared/cached_value.dart
```diff
@@ -0,0 +1,75 @@
...
+
+  /// Updates the cached value.
+  Future<void> update() async {
+    if (_ongoingCompleter != null) {
+      await _ongoingCompleter.future;
+      return;
+    }
+    _ongoingCompleter = Completer();
+    try {
+      await _retryOptions.retry(() async {
```
jonasfj, line 51: Should we not avoid retrying here... and instead document that `updateFn` should retry as needed.

app/lib/shared/cached_value.dart
```diff
@@ -0,0 +1,75 @@
...
+class CachedValue<T> {
+  final UpdateFn<T> _updateFn;
+  final RetryOptions _retryOptions;
+  Timer _timer;
+  T _value;
+  Completer _ongoingCompleter;
+  bool _initialized = false;
+
+  CachedValue({
+    @required UpdateFn<T> fn,
```
jonasfj, line 26: Do we need other options than `fn` maybe it doesn't need to be a keyword parameter?

app/lib/shared/cached_value.dart
```diff
@@ -0,0 +1,75 @@
...
+  })  : _updateFn = fn,
+        _retryOptions = retryOptions ??
+            RetryOptions(
+              delayFactor: Duration(seconds: 3),
+              maxAttempts: 3,
+            );
+
+  bool get isInitialized => _initialized;
+
+  T get value => _value;
```
jonasfj, line 37: Can this be `null`? documentation comment needs to reflect this.

app/lib/shared/cached_value.dart
```diff
@@ -0,0 +1,75 @@
...
+    @required UpdateFn<T> fn,
+    RetryOptions retryOptions,
+  })  : _updateFn = fn,
+        _retryOptions = retryOptions ??
+            RetryOptions(
+              delayFactor: Duration(seconds: 3),
+              maxAttempts: 3,
+            );
+
+  bool get isInitialized => _initialized;
```
jonasfj, line 35:
```suggestion
  /// true, if a cached value is available.
  ///
  /// A cached value is not available, if no value has been loaded yet.
  /// Or if all attempts to load the value have failed.
  bool get isAvailable => _value != null;
```

app/lib/shared/cached_value.dart
```diff
@@ -0,0 +1,75 @@
...
+      c.complete();
+    }
+  }
+
+  void scheduleUpdates(Duration duration) {
+    _timer?.cancel();
+    _timer = Timer.periodic(duration, (timer) {
+      update();
+    });
+  }
```
jonasfj, line 69: Why doesn't the constructor just set this up? Or we could do it a static method that returns an instance of `CacheValue`.

Is there ever any point to creating an instance without scheduling updates? I suppose for testing.

Should we maybe just give `CacheValue` a method where it starts running and installs itself into the service scope?
So that it stops when the current service scope ends? Just brainstorming...

app/lib/shared/cached_value.dart
```diff
@@ -0,0 +1,75 @@
...
+import 'package:meta/meta.dart';
+import 'package:retry/retry.dart';
+
+final _logger = Logger('cached_value');
+
+typedef UpdateFn<T> = Future<T> Function();
+
+/// A locally cached and periodically updated value that can be accessed
+/// synchronously.
+class CachedValue<T> {
```
jonasfj, line 17: Could we expand documentation to say something like:
```
/// Wraps a locally cached value that is periodically updated, and can be accessed synchronously.
///
/// **Example**:
/// ```dart
/// final foo = CacheValue(() => retry(() async {
///   return fetchFooFromSomewhere();
/// }));
///
/// // Access foo, notice that it's important to be able to handle cases
/// // where foo isn't available, because it haven't been loaded yet.
/// if (foo.isAvailable) {
///  print(foo.someMethod());
/// }
/// ```
///
/// The `update` function should do have it's own retry logic.
/// `CacheValue` will eventually retry the `update` function when the next scheduled update happens.
```

jonasfj: We also need to determine a monitoring story here.

Should we maybe give `CacheValue` a few things during setup:
 * an `update` function which fetches a new value,
 * a `timeout` after which an old cached value becomes unavailable? is it ever preferable to return `null` if the cached value is really old?
 * a `name` that can be printed to logs and alerts,
 * an `interval` that says how frequently it should be updated,
 * Then `scheduleUpdates` might just be a  `startUpdates()` without any parameters.

This way, if you "goto references" for `CacheValue` constructor, then you can easily spot the options all cache values are configured with -- and it's easy to find the instance with a given name (say we get an alert that it's failing).

-------

From a monitoring perspective, all we can really do is maybe:
 * **warn** if `update` fails,
 * **shout** if `update` has failed so many times in a row that the cache value got older than `timeout` and, hence, has expired.

This way, for the youtube video listing we can say we have a `timeout` of 24 hours, and an update interval of 6 hours.
So if we fail to update 4 times in a row, then we'll get a **shout**, otherwise we only get warnings.

Similarly for name tracker... maybe we only care to shout if updating has failed for more than 30 minutes, assuming we update every 1-3 minutes.

jonasfj: (let's have a chat about how far we want to go with this next week)

isoos:
Updated with most of the suggestions + added test + `timeout` (alongside `maxAge`).

app/lib/shared/cached_value.dart
```diff
@@ -0,0 +1,118 @@
...
+      final c = _ongoingCompleter;
+      _ongoingCompleter = null;
+      c.complete();
+    }
+  }
+
+  void scheduleUpdates() {
+    if (_closing) {
+      throw StateError('Cache `$_name` is already closed.');
+    }
```
jonasfj, line 104: If `_timer == null && !isAvailable` then lets run an update immediately. Then we don't need to expose `update() ` (or we can make it visible for testing only).

app/lib/shared/cached_value.dart
jonasfj: Maybe making it a method will make it more obvious...

app/lib/shared/cached_value.dart
```diff
@@ -0,0 +1,118 @@
...
+    try {
+      _value = await _updateFn().timeout(_timeout);
+      if (_value != null) {
+        _lastUpdated = DateTime.now();
+      }
+    } catch (e, st) {
+      if (age <= _maxAge) {
+        _logger.warning('Updating cached `$_name` value failed.', e, st);
+      } else {
+        _logger.shout('Updating cached `$_name` value failed.', e, st);
```
jonasfj, line 92: Maybe we should include a simple pattern like: `[pub-ops-note:<code>]` where `<code>` is fairly static (low dimensionality).

In this case we could do: `[pub-ops-note:cached_value/update-failed]` we could also do `[pub-ops-note:cached_value/update-failed/$name]`, but maybe that's too high dimensionality.

It's pretty rudimentary, but robust, with it:
 * We can make a custom log metric that counts occurrences of `[pub-ops-note:<code>]` with `<code>` as tag on the metric.
    * This is easy to occasionally review, and if we see that a certain `<code>` happens regularly or frequently we can fix it.
 * We can search logs for `[pub-ops-note:` or filter to a specific `<code>` by searching for `[pub-ops-note:<code>]`.

It's not exactly as elegant as creating a custom metric, or sending custom tags in logs. But it's simple and robust.
Even if we print this to stderr, we can find it by searching.

app/lib/shared/monitoring.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:logging/logging.dart';
+
+extension LoggerExt on Logger {
+  void pubWarning(String code, String message, [Object error, StackTrace st]) {
```
jonasfj, line 8:
```suggestion
  /// Log a `[pub-notice:${code}` as warning.
  ///
  /// A custom log based metric will track how frequently each [code] happens.
  /// This is useful for logging non-critical errors that we can tolerate, but want to
  /// detect and remedy, should they occur frequently.
  ///
  /// [code] should be a hardcoded string to keep dimensionality manageable. 
  void pubNoticeWarning(String code, String message, [Object error, StackTrace st]) {
```

app/lib/shared/monitoring.dart
```diff
@@ -0,0 +1,15 @@
...
+    warning('[pub-notice:$code] $message', error, st);
+  }
+
+  void pubShout(String code, String message, [Object error, StackTrace st]) {
```
jonasfj, line 12:
```suggestion
  /// Log a `[pub-notice:${code}` with severity shout.
  ///
  /// A custom log based metric will track how frequently each [code] happens.
  /// This is useful for logging non-critical errors that we can tolerate, but want to
  /// detect and remedy, should they occur frequently.
  ///
  /// [code] should be a hardcoded string to keep dimensionality manageable. 
  void pubNoticeShout(String code, String message, [Object error, StackTrace st]) {
```

jonasfj (approved):

Issue #4675, opened 2021-03-27
"Failed to run dartdoc" shows empty output


Komposten:
My package [public_suffix](https://pub.dev/packages/public_suffix/) has lost 10 points because dartdoc failed to run in the latest analysis. When I check the score page it says "Running dartdoc failed with the following output" but no output is provided:

![image](https://user-images.githubusercontent.com/7335682/112724560-3cf8fa80-8f14-11eb-8321-7d40430dcbf0.png)

sethloco:
I had the same issue ([es_compression](https://pub.dev/packages/es_compression)) which seemed to happen around the time of the recent pub.dev outage.  Of course, this is just a guess.

However, I did a small update to pub.dev just a moment ago and it seems to be fixed

Komposten:
Yeah, it probably fixes itself when the analysis is run again so I'm not too worried about that. It's just that the output probably should say _something_ (whether it is actual dartdoc output or that the request to run the command failed or whatever). :)

sethloco:
Your absolutely right...empty error output indicates there is probably some room to improve:)

tiloc:
Same thing here for my just published package "faiadashu". No indication of a general outage. Also happens to colleagues of mine on a regular basis with their packages. I assumed it was because their package is huge (>10k files), but mine is very compact and it still happens.

isoos:
In case of `faiadashu` the log has an indication that no content was detected in the output: https://pub.dev/documentation/faiadashu/latest/log.txt

Not sure why that would be, the package looks fine otherwise, I'll investigate a bit more.

PR #4683, opened 2021-03-30
Display timeout and other failures on dartdoc report section.

isoos:
#4675

### DIFF
--- a/app/lib/dartdoc/dartdoc_runner.dart
+++ b/app/lib/dartdoc/dartdoc_runner.dart
@@ -9,6 +9,9 @@ import 'dart:io';
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 import 'package:pana/pana.dart' hide Pubspec, ReportStatus;
+import 'package:pana/pana.dart' as pana;
+// ignore: implementation_imports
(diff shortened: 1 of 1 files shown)


app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -657,3 +665,20 @@ DartdocReport _emptyReport() => DartdocReport(
       // TODO: add meaningful message for missing documentation on dartdoc
       documentationSection: null,
     );
+
+/// Creates a report section when running dartdoc failed to produce content.
+ReportSection _dartdocFailedSection(String abortMessage) {
```
jonasfj, line 670: What if we did this on the pana-side?

isoos: The original method `dartdocFailedSection` was/is in `pana`, but we don't use it inside `pana`. I used to think that this has a better place in `pana`, but only if we were to use it there too.

jonasfj: ah, nvm... this is because when running pana we don't know if running dartdoc has failed.

jonasfj (approved):

PR #4689, opened 2021-03-31
Global / runtimeVersioned neat task statuses.

isoos:


app/lib/tool/neat_task/pub_dev_tasks.dart
```diff
@@ -23,27 +23,39 @@ void setupFrontendPeriodicTasks() {
   // Deletes expired audit log records.
   _daily(
     name: 'delete-expired-audit-log-records',
+    isVersioned: false,
```
jonasfj, line 26: Can we use a more explaining name for this... like `isRuntimeVersioned`

app/lib/tool/neat_task/datastore_status_provider.dart
```diff
@@ -4,40 +4,75 @@
...
+import 'package:meta/meta.dart';
 import 'package:neat_periodic_task/neat_periodic_task.dart';
 import 'package:ulid/ulid.dart';
 
 import '../../shared/datastore.dart' as db;
+import '../../shared/versions.dart' show runtimeVersion;
 
+/// Tracks the status of the task.
+///
+/// The `id` of the entity is either `global/name` or `scope/name`.
```
jonasfj, line 16:
```suggestion
/// The `id` of the entity is either `global/<name>` or `<scope>/<name>`.
```

app/lib/tool/neat_task/datastore_status_provider.dart
```diff
@@ -4,40 +4,75 @@
...
 class NeatTaskStatus extends db.ExpandoModel<String> {
-  String get name => id;
+  /// The name of the task.
+  @db.StringProperty()
+  String name;
+
+  /// The scope of the task.
+  ///
+  /// TODO: cleanup entities without scope or name
+  /// TODO: make scope and name required: true
```
jonasfj, line 26: Is `"global"` a special scope for tasks that aren't scoped?

Also other than `runtimeVersion` what scopes might be we have?

Could we maybe document them here in a documentation comment..

jonasfj (approved):

app/lib/tool/neat_task/datastore_status_provider.dart
isoos: Yes, correct, global was meant for otherwise unscoped or unversioned tasks. Maybe `global` (or even `scope`) isn't the best name for it...

jonasfj: Perhaps we should just call the property for `runtimeVersion` and give it a comment that task which are not runtime versioned, have `NeatTaskStatus.runtimeVersion == '-'` (I think `-` is better than empty string, because then we can make it required, I think :D)

jonasfj (approved):

PR #4699, opened 2021-04-06
No-dependency tool to check access to pub domains.

isoos:


app/bin/tools/check_domain_access.dart
```diff
@@ -0,0 +1,39 @@
...
+    print('Iteration #${i + 1}/$iterations...');
+    for (final url in urls) {
+      await _checkUrlAccess(url);
+    }
+  }
+  print('Done.');
+}
+
+Future<void> _checkUrlAccess(String url) async {
+  final client = HttpClient();
```
jonasfj, line 32: Maybe create a single client and close it after all requests are done... otherwise the program hangs for a while, and we won't test connection reuse...

app/bin/tools/check_domain_access.dart
```diff
@@ -0,0 +1,39 @@
...
+  final rq = await client.getUrl(Uri.parse(url));
```
jonasfj, line 33: What if we do a quick: [`lookup`](https://api.dart.dev/stable/2.12.2/dart-io/InternetAddress/lookup.html) on the address of the URL to see what it is resolved to.

jonasfj: And maybe even do `Socket.connect` to see if we can reach it.

jonasfj: We should probably make this tool a bit smarter... So that if the HTTP connection fails, then it prints all the IPs for the host and whether or not connecting failed.

We probably also should try all urls, even if the first one fails. But then we have a decent tool for telling people that they likely have a firewall blocking `dart`. I don't see any other reason why `dart` shouldn't be able to open TCP connections.

Short of there being some bug in the Dart SDK that prevents TCP connections from working.

isoos:
Good point about the connect/socket test, I've added it, also updated the connection reuse part. PTAL.

jonasfj (approved):

PR #4710, opened 2021-04-09
Deploy without self-dependency.

### LINKED ISSUE #4550: Deploy without self-dependency
We currently deploy `pub.dev` using `dart pub get`, this might be undesirable in case we ever have to deploy while the service is down.

We could probably write a dart script in something like `tool/resolve-deps.dart` that:
 * Reads `app/pubspec.lock` to find package names and versions,
 * Download packages directly from Google Cloud Storage
    (`https://storage.googleapis.com/pub-packages/packages/<package>-<version>.tar.gz`)
 * Extracts tarballs into `.dart_tool/pub-dev-cache/` with same directory format as `PUB_CACHE`.
 * Calls `dart pub get --offline` with `PUB_CACHE=.dart_tool/pub-dev-cache/`

This might be so simple that we can do with simple bash scripting.

Then we can use this script in our dockerfiles, and avoid self-dependency during deployment.

isoos:
- Fixes #4550.
- We still depend on `tar` in the docker image, but that seems to be fine, as it happens only at build-time.


tool/populate_pub_cache.dart
```diff
@@ -0,0 +1,110 @@
...
+
+final _client = HttpClient();
+
+/// Fetches and populates a local .pub-cache directory without using `pub`
+/// client and `pub.dev` services.
+Future<void> main(List<String> args) async {
+  final baseDir = File(Platform.script.toFilePath()).parent.parent;
+  print('Base directory: $baseDir');
+
+  final pubCacheDir = Directory('${baseDir.path}/build/pub-cache');
```
jonasfj, line 15: Would it work if place it in `.dart_tool/pub-cache/`

tool/populate_pub_cache.dart
```diff
@@ -0,0 +1,110 @@
...
+  print('PUB_CACHE_DIR: ${pubCacheDir.path}');
+  pubCacheDir.createSync(recursive: true);
+
+  final multiVersions = <String, Set<String>>{};
+  final lockFiles = baseDir
+      .listSync(recursive: true)
+      .whereType<File>()
+      .where((f) => !f.path.contains('/pkg/code_coverage/'))
+      .where((f) => !f.path.contains('/pkg/pub_integration/'))
+      .where((f) => f.path.endsWith('pubspec.lock'));
```
jonasfj, line 25: hmm, what if we passed the list of lockfiles as arguments?

Or if we rename: `tool/populate_pub_cache.dart` to `tool/pub_get_offline.dart` and make it wrap `dart pub get --offline --no-precompile`.
So that we don't call ` dart pub get --offline --no-precompile` from `Dockerfile`, but only call `tool/pub_get_offline.dart` (for each package we want to `pub get`).

We can choose not to fetch a package if the `<package>-<version>/` folder already exists.

-------------------------

I'm mostly concerned that right now the list of packages we want to fetch dependencies for is specifies in two places: 
 * `tool/populate_pub_cache.dart`, and,
 * `Dockerfile`

tool/populate_pub_cache.dart
```diff
@@ -0,0 +1,110 @@
...
+      multiVersions.putIfAbsent(package, () => <String>{}).add(pvs[package]!);
+    }
+  }
+
+  final packages = multiVersions.keys.toList()..sort();
+  for (final package in packages) {
+    final versions = multiVersions[package]!.toList()..sort();
+    print('Found $package: $versions ...');
+    for (final version in versions) {
+      await _downloadInto(package, version, pubCacheDir);
```
jonasfj, line 39: Maybe wrap this with a rudimentary try / catch and retry...
We don't need exponential backoff, just sleep 3s before trying again.

tool/populate_pub_cache.dart
```diff
@@ -0,0 +1,110 @@
...
+/// Downloads the archive and extracts it into
+/// [pubCacheDir]/hosted/pub.dartlang.org/[package]-[version]/
+Future<void> _downloadInto(
+    String package, String version, Directory pubCacheDir) async {
+  final targetDir = Directory(
+      '${pubCacheDir.path}/hosted/pub.dartlang.org/$package-$version');
+  if (targetDir.existsSync() &&
+      File('${targetDir.path}/pubspec.yaml').existsSync()) {
+    print('- $package-$version already exists');
+    return;
```
jonasfj, line 88: What if the folder exists, but the `pubspec.yaml` does not?

tool/populate_pub_cache.dart
jonasfj: Also in the `catch` block it might be smart to delete the folder we expected this to be written to... since there could be a partially extracted package there..

tool/populate_pub_cache.dart
```diff
@@ -0,0 +1,110 @@
...
+  final rs = await rq.close();
+  if (rs.statusCode != 200) {
+    throw AssertionError('Unable to access archive of $package-$version.');
+  }
+  final bodyChunks = await rs.toList();
+
+  final process =
+      await Process.start('tar', ['-zxf', '-', '-C', targetDir.path]);
+  for (final chunk in bodyChunks) {
+    process.stdin.add(chunk);
```
jonasfj, line 102: You can also do: `rs.pipe(stdin)`

tool/populate_pub_cache.dart
isoos: I think `tar` will override it. Should we just clear the directory and force-download the packages into a clean state?

isoos:
Updated script to wrap around `pub get --offline`, and always create a clean `PUB_CACHE` directory.

tool/pub_get_offline.dart
```diff
@@ -0,0 +1,104 @@
...
+      '${pubCacheDir.path}/hosted/pub.dartlang.org/$package-$version');
+  if (targetDir.existsSync()) {
+    throw AssertionError('$package-$version already exists');
+  }
+  targetDir.createSync(recursive: true);
+  final rq = await _client.getUrl(Uri.parse(
+      'https://storage.googleapis.com/pub-packages/packages/${Uri.encodeComponent(package)}-${Uri.encodeComponent(version)}.tar.gz'));
+  final rs = await rq.close();
+  if (rs.statusCode != 200) {
+    throw AssertionError('Unable to access archive of $package-$version.');
```
jonasfj, line 95:
```suggestion
    throw Exception('Unable to access archive of $package-$version.');
```

tool/pub_get_offline.dart
```diff
@@ -0,0 +1,104 @@
...
+  }
+  final process =
+      await Process.start('tar', ['-zxf', '-', '-C', targetDir.path]);
+  await rs.pipe(process.stdin);
+  final exitCode = await process.exitCode;
+  if (exitCode != 0) {
+    throw AssertionError('Unable to extract archive of $package-$version.');
```
jonasfj, line 102:
```suggestion
    throw Exception('Unable to extract archive of $package-$version.');
```

tool/pub_get_offline.dart
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/// Downloads the archive and extracts it into
+/// [pubCacheDir]/hosted/pub.dartlang.org/[package]-[version]/
+Future<void> _downloadInto(
+    String package, String version, Directory pubCacheDir) async {
+  final targetDir = Directory(
+      '${pubCacheDir.path}/hosted/pub.dartlang.org/$package-$version');
+  if (targetDir.existsSync()) {
+    throw AssertionError('$package-$version already exists');
```
jonasfj, line 88:
```suggestion
    targetDir.deleteSync(recursive: true);
```

otherwise, retries won't work.

tool/pub_get_offline.dart
```diff
@@ -0,0 +1,104 @@
...
+    pubCacheDir.deleteSync(recursive: true);
+  }
+  pubCacheDir.createSync(recursive: true);
+
+  final pvs = _parsePubspecLockSync(File('$pkgDir/pubspec.lock'));
+  final packages = pvs.keys.toList()..sort();
+  for (final package in packages) {
+    final version = pvs[package]!;
+    print('Downloading $package: $version ...');
+    await _downloadInto(package, version, pubCacheDir);
```
jonasfj, line 28: Maybe some simple retries will make us happy:
```suggestion
    var retry = 0;
    while (true) {
      print('Downloading $package: $version (attempt: ${retry+1})');
      try {
        await _downloadInto(package, version, pubCacheDir);
      } catch (e) {
        if (retry >= 8) {
          rethrow;
        }
        await Future.delayed(Duration(seconds: 1));
        retry++;
      }
    }
```

jonasfj (approved):

PR #4718, opened 2021-04-12
Migrated Timer uses to cached value.

isoos:


app/lib/service/csp/backend.dart
```diff
@@ -19,30 +21,37 @@ CspBackend get cspBackend => ss.lookup(#_cspBackend) as CspBackend;
 
 /// Represents the backend for the CSP handling and related utilities.
 class CspBackend {
```
jonasfj, line 23: This ended up not working anyways... I'm okay with removing the feature too.

jonasfj (approved):

app/lib/service/csp/backend.dart
isoos: I'll remove it in a follow-up.

PR #4725, opened 2021-04-14
Copy to clipboard: strip the dollar-sign prefix from single-line shell commands.

### LINKED ISSUE #4716: Copying commands should not copy leading "$"
Example link : [https://pub.dev/packages/url_launcher/install](https://pub.dev/packages/url_launcher/install)

On copying the install command to clipboard, the leading dollar sign is also copied. This is inconvenient, because when pasting it into the shell, you need to go to the beginning of command and remove it.

On the example, clicking the copy but
```
 $ flutter pub pub add url_launcher
```
while it should be this
```
flutter pub pub add url_launcher
```

This minor enhancement will improve user experience.

isoos:
Fixes #4716.

### DIFF
--- a/app/lib/frontend/templates/views/pkg/install_tab.mustache
+++ b/app/lib/frontend/templates/views/pkg/install_tab.mustache
@@ -23,7 +23,7 @@
 {{/use_pub_get}}
 {{#use_flutter_packages_get}}
 <p>With Flutter:</p>
-<pre data-text-to-copy="flutter pub pub add {{ package }}"><code class="language-shell"> $ flutter pub pub add {{ package }}</code></pre>
+<pre data-text-to-copy="flutter pub add {{ package }}"><code class="language-shell"> $ flutter pub add {{ package }}</code></pre>
 {{/use_flutter_packages_get}}
 
 <p>
--- a/app/lib/frontend/templates/views/pkg/install_tab.mustache
+++ b/app/lib/frontend/templates/views/pkg/install_tab.mustache
@@ -5,7 +5,7 @@
 <h2>Use this package as an executable</h2>
 <h3>1. Install it</h3>
 <p>You can install the package from the command line:</p>
-<pre><code class="language-shell">$ <strong>dart pub global activate {{ package }}</strong>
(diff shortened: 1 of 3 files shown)


pkg/web_app/lib/src/hoverable.dart
```diff
@@ -108,7 +108,12 @@ void _setEventForPreCodeCopyToClipboard() {
     container.append(feedback);
 
     button.onClick.listen((_) async {
```
jonasfj, line 110: This seems like the wrong place to fix this..

Should we perhaps just remove `$ ` from the template instead?

Or we could add another data property like `data-text-to-copy="..."`, but magically deciding that texts starting with `$ ` should get special treatment seems liable to introduce other bugs in the future.

isoos: > Should we perhaps just remove $ from the template instead?

`dart.dev` has the same pattern for shell commands, I think we should follow that.

isoos: Updated to use data- attribute, PTAL.

app/lib/frontend/templates/views/pkg/install_tab.mustache
```diff
@@ -19,11 +19,11 @@
 <p>Run this command:</p>
 {{#use_pub_get}}
 <p>With Dart:</p>
-<pre><code class="language-shell"> $ dart pub add {{ package }}</code></pre>
+<pre data-text-to-copy="dart pub add {{ package }}"><code class="language-shell"> $ dart pub add {{ package }}</code></pre>
 {{/use_pub_get}}
 {{#use_flutter_packages_get}}
 <p>With Flutter:</p>
-<pre><code class="language-shell"> $ flutter pub pub add {{ package }}</code></pre>
+<pre data-text-to-copy="flutter pub pub add {{ package }}"><code class="language-shell"> $ flutter pub pub add {{ package }}</code></pre>
```
jonasfj, line 26: `fluttee pub add` should work now... No need for pub pub

jonasfj (approved):

app/lib/frontend/templates/views/pkg/install_tab.mustache
isoos: Updated.

jonasfj (approved):

PR #4741, opened 2021-04-20
More archive surface tests.

isoos:


pkg/pub_package_reader/lib/src/archive_surface.dart
```diff
@@ -0,0 +1,51 @@
...
+        'Package archive is too large (size > $maxArchiveSize).');
+    return;
+  }
+
+  int uncompressedLength = 0;
+  try {
+    uncompressedLength = await file
+        .openRead()
+        .transform(gzip.decoder)
+        .fold<int>(0, (length, element) => length + element.length);
```
jonasfj, line 37: A safer alternative is to use [`limitChunkedStream`](https://pub.dev/documentation/chunked_stream/latest/chunked_stream/limitChunkedStream.html) with a `maxSize` and then handle the `MaximumSizeExceeded` exception.

Because a tiny gzip stream can contain many GB, and this still causes the full stream to be inflated. While it shouldn't eat all memory, it could still take up a lot of CPU.. and effectively take a very long time..

jonasfj (approved):

pkg/pub_package_reader/lib/src/archive_surface.dart
isoos: Ack, doing it in a follow-up PR.

Issue #4745, opened 2021-04-23
Move 'Package of the Week' section to the top


jayoung-lee:
Less than 5% of the visitors of the landing page scroll down to reach the bottom, where the `Package of the Week` section is located. 

But I think the videos should be located on the top of the page and be easier to find, because they are more intriguing than the cards of featured packages, as the list gets updated more frequently! 

Can we consider moving the section to the top (right below the search section) to make it easier to find? 

If this decision needs data to back it up, we can consider adding click events (GA) to the features package cards as well as the video cards to see which content is more appealing ( = #clicks / #seen). We can even try A/B testing to see whether moving the section up has any bad consequences :)

jonasfj:
I think it's easy to move. So more than driving engagement, I think it could be worth doing just to keep things fresh. We can always move it back down a bit.

But let's perhaps wait with moving it up until there's a few more videos to pick from...

----------

In the meantime, maybe we should introduce a GA click event to see how much the cards are used.

jayoung-lee:
Waiting a bit to have a few more videos to pick from sounds good to me. Thanks!

sigurdm:
I think we might have enough videos for this now!

Let's try doing it.

PR #4746, opened 2021-04-23
Track a bit more landing page events.

isoos:
#4745


jonasfj (approved):

jonasfj:
not abbrevating might be nice.. it'll make the metrics easier to read... but it's a nit :D

PR #4765, opened 2021-05-03
Limit file count and total length in pkg/pub_dartdoc.

isoos:
- https://github.com/dart-lang/dartdoc/issues/2615
- written files that are inside the output directory are kept in memory with a strict limit (10k files, 10MB total size)
- this is a better approach than overriding only the `DartdocFileWriter`, as the later misses the static files that are copied directly
- updated pub-dartdoc-data to use the same resource provider pattern
- files are written to the disk at the end of the process

### DIFF
--- a/pkg/pub_dartdoc/bin/pub_dartdoc.dart
+++ b/pkg/pub_dartdoc/bin/pub_dartdoc.dart
@@ -6,20 +6,33 @@ import 'package:dartdoc/dartdoc.dart';
 import 'package:dartdoc/options.dart';
 
 import 'package:pub_dartdoc/pub_data_generator.dart';
+import 'package:pub_dartdoc/src/pub_hooks.dart';
 
 Future<void> main(List<String> arguments) async {
-  final config = await parseOptions(pubPackageMetaProvider, arguments);
+  // pub hooks
+  final pubResourceProvider =
+      PubResourceProvider(pubPackageMetaProvider.resourceProvider);
+  final pubMetaProvider =
+      PubPackageMetaProvider(pubPackageMetaProvider, pubResourceProvider);
+
+  // dartdoc initialization
+  final config = await parseOptions(pubMetaProvider, arguments);
   if (config == null) {
     throw ArgumentError();
   }
+  pubResourceProvider.setOutput(config.output);
+
   final packageConfigProvider = PhysicalPackageConfigProvider();
   final packageBuilder =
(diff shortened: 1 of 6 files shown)


pkg/pub_dartdoc/bin/pub_dartdoc.dart
```diff
@@ -6,20 +6,33 @@ import 'package:dartdoc/dartdoc.dart';
 import 'package:dartdoc/options.dart';
 
 import 'package:pub_dartdoc/pub_data_generator.dart';
+import 'package:pub_dartdoc/src/pub_hooks.dart';
 
 Future<void> main(List<String> arguments) async {
-  final config = await parseOptions(pubPackageMetaProvider, arguments);
+  // pub hooks
+  final pubResourceProvider =
+      PubResourceProvider(pubPackageMetaProvider.resourceProvider);
```
jonasfj, line 14: Where is `pubPackageMetaProvider` coming from??

jonasfj: nvm, https://pub.dev/documentation/dartdoc/latest/dartdoc/pubPackageMetaProvider.html

I thought it was missing :D

pkg/pub_dartdoc/lib/src/pub_hooks.dart
```diff
@@ -0,0 +1,195 @@
...
+import 'package:dartdoc/dartdoc.dart';
+import 'package:path/path.dart' as p;
+import 'package:watcher/watcher.dart';
+
+const _maxFileCount = 10000;
+const _maxTotalLengthBytes = 10 * 1024 * 1024;
+
+/// Creates an overlay file system with binary file support on top
+/// of the input sources.
+class PubResourceProvider implements ResourceProvider {
```
jonasfj, line 24: Why not just create a proper overlay, instead of relying on output folder detection?

When writing we always write to `_memoryResourceProvider`, when reading we read from `_memoryResourceProvider` and if not present there, then we check: `_defaultProvider`.

I guess the complicated part is when listing folders, we'll need to wrap and merge them... But it would be less magic..

We can also leave that as a TODO... (I'm guessing we can almost just steal the overlay implementation from analyzer and binary file support to it, why not?

jonasfj (approved):

pkg/pub_dartdoc/lib/src/pub_hooks.dart
isoos: ++

PR #4769, opened 2021-05-04
Limit the number of files in a package archive.

isoos:
Also limits the total length via `package:tar` parsing.

### DIFF
--- a/pkg/pub_package_reader/lib/src/tar_utils.dart
+++ b/pkg/pub_package_reader/lib/src/tar_utils.dart
@@ -83,10 +83,17 @@ abstract class TarArchive {
   }
 
   /// Creates a new instance by scanning the archive at [path].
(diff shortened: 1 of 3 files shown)


jonasfj (approved):

pkg/pub_package_reader/lib/src/tar_utils.dart
```diff
@@ -171,15 +187,33 @@ class _PkgTarArchive extends TarArchive {
...
+    int totalLengthBytes = 0;
     while (await reader.moveNext()) {
       final entry = reader.current;
+
+      fileCount++;
+      if (maxFileCount != null && fileCount > maxFileCount) {
+        throw Exception('Maximum file count reached: $maxFileCount.');
+      }
+
+      totalLengthBytes += entry.size;
```
jonasfj, line 211: :rocket:

PR #4776, opened 2021-05-06
Migrate pkg/pub_package_reader to null safety.

isoos:


isoos:
@jonasfj I have a commit that makes the `PackageSummary.issues` non-nullable, I will push it to the PR, but I'm a bit undecided on how it should look like. If you have a preference there, I'll update it accordingly...

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -44,7 +44,7 @@ class PackageSummary {
   final List<String>? libraries;
 
   PackageSummary({
-    this.issues,
+    List<ArchiveIssue>? issues,
```
jonasfj, line 47: You could also do: `issues = []` (I think)...

But this is fine too... The benefit of doing the other thing is that you can't accidentally pass null as `issues`.

jonasfj (approved):
:ship:

PR #4796, opened 2021-05-12
WIP: Script to check .tar.gz archives with package:tar parsing.

isoos:
- Marked as work-in-progress, because I don't think we'll need to merge it.
- Uses `package:iso` instead of `package:isolate`, as the later is not null-safe yet.
- Uses non-reused isolates, as otherwise we've got OOMs, not sure if it is because `package:iso` or we have a leak somewhere.
- Scanned 34% of the package archives, the extraction is 100% match so far.
- Early version of this code revealed #4791


jonasfj (approved):
LGTM,
I agree we probably don't need to merge this.

PR #4819, opened 2021-05-20
Use package:tar for dartdoc's package.tar.gz

isoos:


isoos:
Note: started testing on staging.

jonasfj (approved):

PR #4836, opened 2021-05-28
Store reports on ScoreCard too.

isoos:
- Starts to store reports if they are <32 KiB. Reports larger than 16 KiB are logged with warning or shout.
- Starts to use the reports stored on the `ScoreCard`, if they are available on the entity, otherwise falls back to use the `ScoreCardReport` ones.
- Once we have confirmed that this is not causing issues, we'll have a release that doesn't store the reports separately, and eventually we can remove them.


isoos:
Pushed to staging to test it with a new `runtimeVersion`.

PR #4839, opened 2021-05-28
Log the total number of entries deleted in one of the background tasks.

isoos:
- We should not have these files in the storage bucket anymore.
- Before deleting the background task, I want to be sure that we don't need it anymore.


jonasfj (approved):

jonasfj:
It might be a while before this runs through... GCS load from dartdoc makes this job fail sometimes...

PR #4845, opened 2021-06-01
Store reports on ScoreCard too.

isoos:
- Starts to store reports if they are <32 KiB. Reports larger than 16 KiB are logged with warning or shout.
- Starts to use the reports stored on the ScoreCard, if they are available on the entity, otherwise falls back to use the ScoreCardReport ones.
- Once we have confirmed that this is not causing issues, we'll have a release that doesn't store the reports separately, and eventually we can remove them.

### DIFF
--- a/app/lib/scorecard/backend.dart
+++ b/app/lib/scorecard/backend.dart
@@ -314,7 +314,7 @@ class ScoreCardBackend {
         final size = scr.reportJsonGz.length;
         if (size > _reportSizeDropThreshold) {
           // TODO: replace it with meaningful content
-          _logger.shout(
+          _logger.reportError(
               '${scr.reportType} report exceeded size threshold ($size > $_reportSizeWarnThreshold)');
           continue;
         }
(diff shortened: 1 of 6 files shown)


app/lib/scorecard/backend.dart
```diff
@@ -292,6 +308,31 @@ class ScoreCardBackend {
...
       );
 
+      // store compressed report on the ScoreCard entity too
+      for (final scr in reportEntities) {
+        if (scr == null || scr.reportJsonGz == null) continue;
+        final size = scr.reportJsonGz.length;
+        if (size > _reportSizeDropThreshold) {
+          // TODO: replace it with meaningful content
+          _logger.shout(
+              '${scr.reportType} report exceeded size threshold ($size > $_reportSizeWarnThreshold)');
```
jonasfj, line 318: Consider using something like:
```dart
extension CustomException on Logger {
  reportError(String message) =>
       this.shout(message, Exception(message), StackTrace.current);
}
```
So errors show up in stack driver error reporting.

jonasfj (approved):

PR #4854, opened 2021-06-04
null-safety badge on the versions listing page

### LINKED ISSUE #4825: Add null safety badge in front of versions in the Versions tab
Recently a user comment that it's difficult to find out which version supports null safety. They had to click every version in order to know if it is null safe or not. 

Can we consider adding an icon or a badge in front of versions in the Versions tab, so this sort of search/need is better supported?

isoos:
- Fixes #4825.
- Because the mobile view has not enough horizontal space, I've removed it from there.

Screenshot 2021-06-04 124224

### DIFF
--- a/app/lib/frontend/templates/package_versions.dart
+++ b/app/lib/frontend/templates/package_versions.dart
@@ -116,6 +116,7 @@ String renderVersionTableRow(PackageVersion version, String downloadUrl) {
     'package': version.package,
     'version': version.version,
     'version_url': urls.pkgPageUrl(version.package, version: version.version),
+    'has_opted_into_null_safety': version.pubspec.hasOptedIntoNullSafety,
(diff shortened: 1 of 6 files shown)


isoos:
/cc @mit-mit on the design change
I've also pushed it to staging, will be there in less than 30 mins.

mit-mit:
I find the grey color confusing. It can be read as "null safety disabled".

I think I'd rather make it the same blue, and then just have the tooltip explain: "Package version is opted into null safety".

isoos:
> I think I'd rather make it the same blue, and then just have the tooltip explain: "Package version is opted into null safety".

Done.

mit-mit:
LGTM

jonasfj (approved):

PR #4856, opened 2021-06-04
update mono_repo

kevmoo:


kevmoo:
CC @mit-mit

kevmoo:
Feel free to squash/merge and delete the branch if you're happy

app/mono_pkg.yaml
```diff
@@ -6,7 +6,7 @@ stages:
   - smoke_test:
     - group:
         - dartfmt
```
mit-mit, line 8: I find these group entry names confusing, as they still use the format commands.

Can we use more general terms such as 'formatting', 'analysis', and 'tests' ?

kevmoo: need to update mono_repo for that.

Remember, this was based on the travis syntax invented by natalie...~5 years ago

jonasfj (approved):

PR #4864, opened 2021-06-08
Limit SDK ranges on upload.

### LINKED ISSUE #4860: Over-application of "DART 2 INCOMPATIBLE" label
(Split out of https://github.com/flutter/flutter/issues/84109)

httpss://pub.dev/packages/responsive_sizer/versions/3.0.1 is marked as "DART 2 INCOMPATIBLE". Looking at the change, it appears that this is the result of a bad search/replace on the package version that resulted in the SDK constraint changing from

```
 sdk: '>;=2.12.0 <3.0.0' 
```

to

```
sdk: '>=2.12.0 <3.0.1'
```

The latter is obviously a very bad idea, and pub should warn about it (which maybe it did and the author ignored it), but it's not incompatible with Dart 2; the use of that message actively confused the package author about what was wrong.

isoos:
Fixes #4860.

### DIFF
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -332,7 +332,7 @@ Iterable<ArchiveIssue> checkSdkVersionRange(Pubspec pubspec) sync* {
   }
 
   // Dart 3 version accepted with valid upper constraint.
-  if (sdk is VersionRange && sdk.max != null && sdk.allows(_firstDart3Pre)) {
+  if (sdk.allows(_firstDart3Pre)) {
     yield ArchiveIssue(
         'The SDK constraint allows Dart 3.0.0, this is not allowed because Dart 3 does not exist.');
   }
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -328,7 +328,7 @@ Iterable<ArchiveIssue> checkSdkVersionRange(Pubspec pubspec) sync* {
 
   // No known version exists.
   if (sdk.intersect(_preDart3).isEmpty) {
-    yield ArchiveIssue('SDK constraint is outside of current Dart SDKs.');
+    yield ArchiveIssue('SDK constraint does not allow any known Dart SDK.');
   }
 
   // Dart 3 version accepted with valid upper constraint.
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -325,7 +325,7 @@ Iterable<ArchiveIssue> checkSdkVersionRange(Pubspec pubspec) sync* {
 
   // Dart 3 version accepted with valid upper constraint.
   if (sdk is VersionRange && sdk.max != null && sdk.allows(_firstDart3Pre)) {
-    yield ArchiveIssue('SDK constraint accepts Dart 3.');
+    yield ArchiveIssue('The SDK constraint allows Dart 3.0.0, this is not allowed because Dart 3 does not exist.');
   }
 }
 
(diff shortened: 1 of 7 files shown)


pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -307,6 +308,27 @@ Iterable<ArchiveIssue> checkAuthors(String pubspecContent) sync* {
...
 
+final _preDart3 = VersionConstraint.parse('<3.0.0');
+final _firstDart3Pre = Version.parse('3.0.0').firstPreRelease;
+
+/// Checks if the version range is acceptable by current SDKs.
+Iterable<ArchiveIssue> checkSdkVersionRange(Pubspec pubspec) sync* {
+  final sdk = pubspec.environment?['sdk'];
+  if (sdk == null) return;
+  // We still allow pubspec without SDK constraints.
+  if (sdk.isAny) return;
```
jonasfj, line 319: This is only relevant for Dart 1.x, and there is afaik no reason they couldn't be published with an SDK constraint.

Should we not just force that upon Dart 1.x users... We could do a quick scan of pubspec published past year to see if anyone still does this..

feel free to just file that as a separate issue, I see no reason to block this...

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -307,6 +308,27 @@ Iterable<ArchiveIssue> checkAuthors(String pubspecContent) sync* {
...
+
+  // No known version exists.
+  if (sdk.intersect(_preDart3).isEmpty) {
+    yield ArchiveIssue('SDK constraint is outside of current Dart SDKs.');
+  }
+
+  // Dart 3 version accepted with valid upper constraint.
+  if (sdk is VersionRange && sdk.max != null && sdk.allows(_firstDart3Pre)) {
+    yield ArchiveIssue('SDK constraint accepts Dart 3.');
```
jonasfj, line 328:
```suggestion
    yield ArchiveIssue('The SDK constraint allows Dart 3.0.0, this is not allowed because Dart 3 does not exist.');
```

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -307,6 +308,27 @@ Iterable<ArchiveIssue> checkAuthors(String pubspecContent) sync* {
...
+  }
```
jonasfj, line 329: Why not simply condense this down to `if (sdk.allows(_firstDart3Pre)) {` ?

isoos: It will trigger a reject with SDK constraint `>=1.1.0`.

pkg/pub_package_reader/lib/pub_package_reader.dart
isoos: Updated validation to require min and max ranges. PTAL.

pkg/pub_package_reader/lib/pub_package_reader.dart
isoos: With the range requirement this is now just `sdk.allows()`.

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -307,6 +308,36 @@ Iterable<ArchiveIssue> checkAuthors(String pubspecContent) sync* {
...
+      sdk.max == null ||
+      sdk.max!.isAny) {
+    yield ArchiveIssue(
+        'Dart SDK constraint with min and max range must be specified.');
+    return;
+  }
+
+  // No known version exists.
+  if (sdk.intersect(_preDart3).isEmpty) {
+    yield ArchiveIssue('SDK constraint is outside of current Dart SDKs.');
```
jonasfj, line 331: Can we improve the message to say: SDK constraint does not allow any known Dart SDK.

jonasfj (approved):

PR #4873, opened 2021-06-10
update dependencies

kevmoo:


isoos:
Was there any reason for the upgrade? The current dependency override mix is a bit fragile, and e.g. the integration tests are failing because of some error, I'd rather keep it to the required ones.

kevmoo:
I was digging into...something. Just curious why things are outdated.

No big push. Happy to close out. Mostly curious if tests would just pass...

Issue #4878, opened 2021-06-14
admin API for listing / adding / removing uploaders and publishers on packages


jonasfj:
For moderation it might be wise to build out admin API for managing uploaders and publishers on packages.

PR #4884, opened 2021-06-17
Admin API to get the list of package uploaders.

isoos:
#4878


app/lib/shared/configuration.dart
```diff
@@ -461,4 +465,7 @@ enum AdminPermission {
 
   /// Permission to remove a package.
   removePackage,
+
+  /// Permission to get/set the uploaders of a package.
+  managePackageUploaders,
```
jonasfj, line 470: Maybe we should call it `managePackageOwnership`

jonasfj (approved):

PR #4887, opened 2021-06-17
New HTML rendering

isoos:


app/lib/frontend/dom/dom.dart
```diff
@@ -0,0 +1,160 @@
...
+  Node fragment(List<Node> children);
+
+  /// Creates a DOM Element.
+  Node element(
+    String tag, {
+    String id,
+    List<String> classes,
+    Map<String, String> attributes,
+    List<Node> children,
+  });
```
jonasfj, line 29: Consider using `Iterable<T>` in interfaces instead of `List<T>`.
```suggestion
  Node fragment(Iterable<Node> children);

  /// Creates a DOM Element.
  Node element(
    String tag, {
    String id,
    Iterable<String> classes,
    Map<String, String> attributes,
    Iterable<Node> children,
  });
```

app/lib/frontend/dom/dom.dart
```diff
@@ -0,0 +1,160 @@
...
+  final List<Node> children;
+
+  _StringElement(this.tag, this.attributes, this.children);
+
+  @override
+  void writeHtml(StringSink sink) {
+    sink.write('<$tag');
+    if (attributes != null) {
+      for (final e in attributes.entries) {
+        sink.write(' ${e.key}="${_attributeEscape.convert(e.value)}"');
```
jonasfj, line 133: This is probably a good place to point out that it's important to escape `e.key` too ? :D

Or have something in the constructor that throws an `ArgumentError` if `attributes` contains an illegal key.

app/lib/frontend/dom/dom.dart
```diff
@@ -0,0 +1,160 @@
...
+    for (final node in children) {
+      (node as _StringNode).writeHtml(sink);
+    }
+  }
+}
+
+class _StringElement extends _StringNode {
+  final String tag;
+  final Map<String, String> attributes;
+  final List<Node> children;
```
jonasfj, line 124: May I propose that we store this as `List<_StringNode>` and that we do the cast in the constructor?

app/lib/frontend/dom/dom.dart
```diff
@@ -0,0 +1,160 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+
+final _attributeEscape = HtmlEscape(HtmlEscapeMode.attribute);
+
+/// The DOM context to use while constructing nodes.
+///
+/// Override this in browser.
+DomContext dom = StringDomContext();
```
jonasfj, line 12: nit...
```suggestion
DomContext html = StringDomContext();
```

Then things are written as `html.div` and `html.fragment`

app/lib/frontend/templates/package_misc.dart
```diff
@@ -207,14 +208,20 @@ String renderLabeledScores(PackageView view, {String version}) {
...
-  });
+  return dom.fragment([
+    dom.div(
+      classes: ['packages-score-value', if (value != null) '-has-value'],
+      children: [
+        dom.span(
+            classes: ['packages-score-value-number'],
+            children: [dom.text(value?.toString() ?? '--')]),
+        dom.span(
+            classes: ['packages-score-value-sign'], children: [dom.text(sign)]),
```
jonasfj, line 220: nit, but I would use tailing comma in these cases... Not that it's important :D 
```suggestion
        dom.span(
          classes: ['packages-score-value-number'],
          children: [dom.text(value?.toString() ?? '--')],
        ),
        dom.span(
          classes: ['packages-score-value-sign'],
          children: [dom.text(sign)],
        ),
```

jonasfj (approved):

jonasfj (approved):

PR #4891, opened 2021-06-18
Partial null-safety migration

isoos:
- single model references that are not really required are refactored
- `envConfig` is refactored to be migrated separately from the rest of `configuration.dart`
- otherwise regular null-safety migration


app/lib/frontend/dom/dom.dart
```diff
@@ -55,10 +57,10 @@ Node fragment(Iterable<Node> children) => dom.fragment(children);
...
   String tag, {
-  String id,
-  Iterable<String> classes,
-  Map<String, String> attributes,
-  Iterable<Node> children,
+  String? id,
+  Iterable<String>? classes,
+  Map<String, String>? attributes,
+  Iterable<Node>? children,
 }) =>
```
jonasfj, line 64: we could consider: 

```dart
void _method({String? optionalNotNull}) {}

const void Function({String optionalNotNull}) method = _method;

```

so `id` can accidentally be null...

app/lib/frontend/dom/dom.dart
```diff
@@ -124,15 +126,15 @@ class _StringDomContext extends DomContext {
...
+Map<String, String>? _mergeAttributes(
+    String? id, Iterable<String>? classes, Map<String, String>? attributes) {
   final hasClasses = classes != null && classes.isNotEmpty;
   final hasAttributes =
       id != null || hasClasses || (attributes != null && attributes.isNotEmpty);
   if (!hasAttributes) return null;
   return <String, String>{
     if (id != null) 'id': id,
-    if (hasClasses) 'class': classes.join(' '),
+    if (hasClasses) 'class': classes!.join(' '),
```
jonasfj, line 137:
```suggestion
    if (classes != null && classes.isNotEmpty) 'class': classes!.join(' '),
```

jonasfj (approved):

PR #4897, opened 2021-06-23
Do not display unknown publisher.

isoos:


jonasfj:
I thought we talked about giving the "unknown" text a tooltip that explain that the package author has not registered a publisher for this package, and hence...

As a means to encourage package authors to register a publisher for the their packages.

jonasfj (approved):
Caught up on the discussion :D

PR #4899, opened 2021-06-24
Migrate pkg/api_builder and pkg/web_app and most of app/ to null-safety.

isoos:
Most of the changes are straightforward (70% manual, 30% with `dart migrate`), some need to have follow-up cleanups:
- created `redis_cache_wrapper` to force typed access until `neat_cache` is being migrated
- `package:appengine` has non-nullable context, it's access need to be in try-cache block (e.g. `handler_helpers.dart`)

Most of `bin` and `test` is not migrated, we need to migrate `package:mustache` and `package:neat_cache` and/or `package:dartis` first.


isoos:
Deployed to staging, first tests look good.

jonasfj (approved):

PR #4905, opened 2021-06-30
Migrate search_tabs.mustache

isoos:
Some new pattern that I think will be useful:
- In `lib/frontend/templates/views/` I think we should replace the `.mustache` files with corresponding `.dart` files using `dom.dart`.
- In cases like this, where there is a reasonable self-contained data, we should use it alongside the rendering method.
- Let's start returning `String` (the serialized HTML) from these methods, and once we reach a better composition state during the migration, we move to use `Node` instead.


app/test/frontend/golden/search_page.html
```diff
@@ -131,21 +131,21 @@ <h3 class="foldable-button">
...
                 </a>
               </div>
             </div>
             <div class="search-controls-sdk search-controls-buttons">
               <span class="search-controls-label">SDK</span>
               <div class="list-filters">
-                <a class="filter " href="/dart/packages?q=foobar&amp;sort=top" title="Packages compatible with the Dart SDK">Dart</a>
-                <a class="filter " href="/flutter/packages?q=foobar&amp;sort=top" title="Packages compatible with the Flutter SDK">Flutter</a>
+                <a class="filter" href="/dart/packages?q=foobar&amp;sort=top" title="Packages compatible with the Dart SDK">Dart</a>
+                <a class="filter" href="/flutter/packages?q=foobar&amp;sort=top" title="Packages compatible with the Flutter SDK">Flutter</a>
```
jonasfj, line 146: nice cleanup of whitespace :D

app/lib/frontend/templates/views/shared/search_tabs.dart
```diff
@@ -2,32 +2,43 @@
...
-    this.text,
+    required this.active,
+    required this.href,
+    required this.title,
+    required this.text,
   });
+}
 
-  Map<String, dynamic> toJson() => _$SearchTabToJson(this);
+String renderSearchTabs(Iterable<SearchTab> tabs) {
```
jonasfj, line 22: Your suggested pattern is that in general things that produce `Node` exist inside a `views/*.dart` file...

But that in cases where we have a data-class we add the render methods next to the data-class.

I don't particularly mind either way, I think a good thing about `dom.dart` is that we can _go-to definition_ on all templates :D

--------

**Side note**:

Couldn't `SearchTab` just be a function that returns `Node`? or are we actually caching these somewhere, just curious?

Even if we do... it's entirely possible that it would just as fine to just cache the `Node` object.. not that we have to refactor this now -- it's probably best to do these things step by step.

I'm just wondering if in some cases, the intermediate data-class might just be functions that return `Node` which can be cached.

jonasfj (approved):

app/lib/frontend/templates/views/shared/search_tabs.dart
isoos: Maybe we could have a `Node` constructor that just takes the cached HTML output and keeps it, without parsing, emitting when needed. It could be used only in the caching layer...

Let's see how the template migration goes and we can figure out a good pattern for it.

PR #4909, opened 2021-07-02
Explore markdown parsing to detect screenshot images.

isoos:
- Self-linking image and alt text detection works.
- Inlined HTML + custom data value does not work yet, could be done via HTML post-processing, but both markdown processing and HTML sanitizer needs to have support for it. Maybe we should just drop this for now?

### DIFF
--- a/app/lib/shared/markdown.dart
+++ b/app/lib/shared/markdown.dart
@@ -2,8 +2,10 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:collection/collection.dart';
 import 'package:logging/logging.dart';
 import 'package:markdown/markdown.dart' as m;
+import 'package:meta/meta.dart';
 import 'package:pana/pana.dart' show getRepositoryUrl;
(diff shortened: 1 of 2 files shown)


app/lib/shared/markdown.dart
```diff
@@ -400,3 +400,86 @@ Version? _extractVersion(String? text) {
...
+  nodes.forEach((node) => node.accept(visitor));
+  return visitor.images;
+}
+
+/// Finds screenshot images.
+///
+/// When images are embedded in markdown, we could say that it is a screenshot image, if:
+/// - (1) the image is also a link to the image, or
+/// - (2) the image has an alt-text that includes the word "screenshot", or
+/// - (3) the image has a custom data-property only recognized by pub.dev.
```
jonasfj, line 430: We should only do one of these right?

isoos: We can do (1) and (2) easily with basic markdown support, and I think it is not controversial to support both. I think that (3) is more involved with markdown pruning the HTML markup's output, so I'd happily drop that for now.

PR #4914, opened 2021-07-05
Don't leave dartdoc report permanently empty when unknown error happens.

### LINKED ISSUE #4848: Published a package in April, still waiting to analyze
- I maintain a number of packages on pub.dev surrounding using the medical data standard called FHIR. 
- The base package can be found here: https://pub.dev/packages/fhir
- The last time I tried to publish it was apparently April 16th. It still states that it's [waiting to analyze](https://pub.dev/packages/fhir/score).
- I tried to update it to a new version because I've made changes, and the command line states that it's been uploaded, but pub.dev does not show it.

isoos:
- We retry non-successful Jobs after a day or so. If there was no report created on the previous attempt, creating an empty one will unblock the `awaiting` status for the package. 
- Fixes #4848.

### DIFF
--- a/app/lib/job/backend.dart
+++ b/app/lib/job/backend.dart
@@ -207,6 +207,7 @@ class JobBackend {
         final now = DateTime.now().toUtc();
         selected!
           ..state = JobState.processing
+          ..attemptCount = selected.attemptCount + 1
           ..processingKey = createUuid()
           ..lockedUntil = now.add(_defaultLockDuration);
         tx.insert(selected);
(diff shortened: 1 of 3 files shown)


jonasfj (approved):

app/lib/job/backend.dart
```diff
@@ -207,6 +207,7 @@ class JobBackend {
         final now = DateTime.now().toUtc();
         selected!
           ..state = JobState.processing
+          ..attemptCount = selected.attemptCount + 1
```
jonasfj, line 210: This requires us to bump runtimeVersion, right?

isoos: Yeah, we are doing that already for the current release.

PR #4921, opened 2021-07-06
Do not load uploader emails - except for admin page.

isoos:


app/lib/frontend/handlers/package.dart
```diff
@@ -264,6 +270,7 @@ Future<shelf.Response> packageAdminHandler(
     packageName: packageName,
     versionName: null,
     assetKind: null,
+    loadUploaderEmails: true,
```
jonasfj, line 273: Since this is the only place where we load uploader emails, why not remove it from `PackagePageData` and load here, and then pass it explicitly to `renderPkgAdminPage`?

app/lib/frontend/handlers/package.dart
```diff
@@ -328,8 +336,9 @@ Future<PackagePageData> loadPackagePageData(
   final analysisView = await analyzerClient.getAnalysisView(
       selectedVersion.package, selectedVersion.version!);
 
-  final uploaderEmails = await accountBackend
-      .getEmailsOfUserIds(package.uploaders as List<String>);
+  final uploaderEmails = loadUploaderEmails
```
jonasfj, line 339: Won't we need to update the documentation comment on `PackagePageData.uploaderEmails` to indicate that this being `null` doesn't necessarily mean that it has a publisher, because it might have been loaded with `loadUploaderEmails`.

See earlier comment about maybe removing `PackagePageData.uploaderEmails` and loading it in `packageAdminHandler` instead.. since that's the only place we use it.

jonasfj (commented):
I think we should consider refactoring it.. either:
 * Remove `PackagePageData.uploaderEmails` and move the fetching of uploaders; or;
 * Clearly document on `PackagePageData.uploaderEmails` the circumstances under which it might be `null` or empty-list and what the difference is.

app/lib/frontend/handlers/package.dart
isoos: Good point, I'll create a different PR for that.

PR #4927, opened 2021-07-08
Migrate template to code: pagination.

isoos:


isoos:
@jonasfj: maybe the template could be more compact if we also added a `Node? child` alongside the `Iterable<Node> children`, removing the list in the single-child templates?

jonasfj (approved):

Issue #4525, opened 2021-02-16
Display audit log


isoos:
- [x] for a package (only for admins, new tab?)
- [x] for a publisher (only for admins, new tab?)
- [x] for the logged-in user (new tab?)

kevmoo:
What would `for the logged-in user` mean? Who sees what where?

isoos:
> What would for the logged-in user mean? Who sees what where?

Similar to my-packages, my-publishers, we could have a my-events. You would see the events that you did or were part of (e.g. was invited).

kevmoo:
Makes sense!

jonasfj:
Yeah, we need to write up a short design document for this first, and drive it through privacy reviews.

jonasfj:
> Yeah, we need to write up a short design document for this first, and drive it through privacy reviews.

Done.

--------

Also we've decided to rename it to "Activity Log".

IchordeDionysos:
Is it released yet? :)

isoos:
@IchordeDionysos not yet, but hopefully soon

IchordeDionysos:
@isoos thanks, it's a great start already 😍

Would be great to know which activity logs are shown and what they mean.
Might make sense to provide a link to read about this somewhere on the log page.

isoos:
Until there is a documentation, you might want to take a look at this list:
https://github.com/dart-lang/pub-dev/blob/master/app/lib/audit/models.dart#L596-L663

PR #4955, opened 2021-07-16
Activity log UI

isoos:
- #4525 (package, publisher and logged in user's activity log).
- Also updates DOM rendering with `safeHtml` for markdown-provided HTML content.

### DIFF
--- a/app/lib/frontend/dom/dom.dart
+++ b/app/lib/frontend/dom/dom.dart
@@ -33,8 +33,8 @@ abstract class DomContext {
   /// Creates a DOM Text node.
   Node text(String value);
 
-  /// Creates a DOM node with raw unsafe HTML content.
-  Node rawUnsafeHtml(String value);
+  /// Creates a DOM node with unsafe raw HTML content.
+  Node unsafeRawHtml(String value);
 }
 
 void _verifyElementTag(String tag) {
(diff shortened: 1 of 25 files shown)


app/lib/frontend/dom/dom.dart
```diff
@@ -32,6 +32,9 @@ abstract class DomContext {
 
   /// Creates a DOM Text node.
   Node text(String value);
+
+  /// Creates a DOM node with safe HTML content.
+  Node safeHtml(String value);
```
jonasfj, line 37: This method doesn't sanitize html in any way... I think it's better to call this `rawHtml` rather than `safeHtml`.

The `rawHtmlString` string passed to `dom.rawHtml(rawHtmlString)` should be sanitized.

jonasfj: Just saying, this is a way to inject raw HTML, right?

Calling it `safeHtml` makes me think that it's "safe" to inject any HTML here...

Perhaps a better name is `rawUnsafeHtml` or ´unsafeRawHtml`, or something like that.. the method is inherently unsafe, right?

jonasfj (approved):

app/lib/frontend/dom/dom.dart
isoos: Good point, renamed.

isoos: One alternative would be to create `Node.markdownContent(String markdownSource)`, which would inside ensure it is safe... although it would need to accept our custom hooks...

isoos: I've renamed it to `rawUnsafeHtml`, but `unsafeRawHtml` may sound better if I read out the code. Let me know if you prefer the later.

PR #4958, opened 2021-07-19
Pagination for activity records.

isoos:
#4525

### DIFF
--- a/app/lib/audit/backend.dart
+++ b/app/lib/audit/backend.dart
@@ -3,10 +3,14 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:gcloud/service_scope.dart' as ss;
+import 'package:shelf/shelf.dart' as shelf;
 
 import '../shared/datastore.dart';
 
 import 'models.dart';
+
+/// The maximum number of entities to be loaded from Datastore in one batch.
+const _maxAuditLogBatchSize = 1000;
 
 /// Sets the audit backend service.
 void registerAuditBackend(AuditBackend backend) =>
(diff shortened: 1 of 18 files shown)


app/lib/audit/backend.dart
```diff
@@ -20,49 +24,66 @@ class AuditBackend {
...
+    DateTime? before,
+  ) async {
+    assert(propertyName == 'users' ||
+        propertyName == 'packages' ||
+        propertyName == 'packageVersions' ||
+        propertyName == 'publishers');
+    final query = _db.query<AuditLogRecord>()
+      ..filter('$propertyName =', value)
+      ..filter('created <=',
+          before ?? DateTime.now().toUtc().add(Duration(minutes: 5)))
```
jonasfj, line 39: Why add 5 minutes?

jonasfj: Can't we just make it strict less than?

jonasfj (approved):

app/lib/audit/backend.dart
```diff
@@ -73,3 +94,9 @@ class AuditBackend {
     );
   }
 }
+
+extension RequestHeaderBeforeExt on shelf.Request {
+  DateTime? get before => requestedUri.queryParameters.containsKey('before')
+      ? DateTime.tryParse(requestedUri.queryParameters['before']!)
```
jonasfj, line 100: Let's throw 400 if there is a parse error here..

isoos: I've updated the code to return a formatted 404. Right now we don't do such parsing (yet) in other places, e.g. https://pub.dev/packages?q=json&page=2x gets a valid response :(
I think it may not be important to spell out which parameter is wrong, so reusing the 404 may be fine?

PR #4961, opened 2021-07-20
Expose package's publisherId on the UI and in the data object.

isoos:


app/lib/audit/models.dart
```diff
@@ -150,10 +150,12 @@ class AuditLogRecord extends db.ExpandoModel<String> {
...
   }) {
-    final summary = 'Package `$package` '
-        'version `$version` '
-        'was published '
-        'by `${uploader.email}`.';
+    final summary = [
+      'Package `$package` version `$version` ',
+      'was published by `${uploader.email}`',
+      if (publisherId != null) ' (publisher `$publisherId`)',
+      '.',
```
jonasfj, line 157: Maybe we can write something like:
```suggestion
      'Package `$package` version `$version` ',
      'was published by `${uploader.email}`',
      if (publisherId != null) ' using publisher `$publisherId`',
      '.',
```
e.g.
> Package `foo` version `1.2.3` was published by `user@example.com` using publisher `example.com`.

or 

```suggestion
      'Package `$package` version `$version` ',
      if (publisherId != null) ' owned by publisher `$publisherId`',
      'was published by `${uploader.email}`',
      '.',
```
e.g.
> Package `foo` version `1.2.3` owned by publisher `example.com` was published by `user@example.com`.

jonasfj (approved):

app/lib/audit/models.dart
isoos: I like the "owned by publisher" better, updated the code to use that.

PR #4962, opened 2021-07-20
Short date format for activity log query parameter.

isoos:


app/lib/audit/backend.dart
```diff
@@ -93,14 +100,34 @@ class AuditBackend {
...
+        DateTime.utc(next.year, next.month, next.day).add(Duration(days: 1));
+    return nextDayStart.isBefore(last) && nextDayStart.isAfter(next)
+        ? nextDayStart.toIso8601String().split('T').first
+        : next.toIso8601String();
+  }
+
   /// Parses the `before` query parameter and returns the parsed timestamp.
   ///
   /// Returns a timestamp slightly into the future if the parameter is missing.
   /// Returns null if the query parameter is invalid.
```
jonasfj, line 115:
```suggestion
  /// Returns `null` if the query parameter is invalid.
```

Why not just throw an `ResponseException` with `400`?

app/lib/audit/backend.dart
```diff
@@ -93,14 +100,34 @@ class AuditBackend {
...
+      return now.add(const Duration(minutes: 5));
+    }
+    if (param.length == 10) {
+      final m = _shortBeforeFormat.matchAsPrefix(param);
+      if (m != null) {
+        final parsed = DateTime.utc(int.parse(m.group(1)!),
+            int.parse(m.group(2)!), int.parse(m.group(3)!));
+        if (parsed.year > 2000 && parsed.year <= now.year) {
+          return parsed;
+        }
```
jonasfj, line 128: Isn't it better to throw a `ResponseException` complaining that years before 2000 and after current year, is not allowed.

app/lib/audit/backend.dart
```diff
@@ -93,14 +100,34 @@ class AuditBackend {
     );
   }
 
+  @visibleForTesting
+  String nextTimestamp(DateTime last, DateTime next) {
+    final nextDayStart =
+        DateTime.utc(next.year, next.month, next.day).add(Duration(days: 1));
+    return nextDayStart.isBefore(last) && nextDayStart.isAfter(next)
```
jonasfj, line 107: Can `nextDayStart.isAfter(next)` ever be false?

If not maybe add an `assert(nextDayStart.isAfter(next));` (leaving the condition too, is not wrong, since we won't have asserts in production).

jonasfj (approved):

app/lib/audit/backend.dart
isoos: If the `next` timestamp is `00:00:00.000`, then they are equal, so that is just checking an unlikely edge case.

PR #4965, opened 2021-07-21
Migration of listing info + sort control templates.

isoos:


app/lib/frontend/templates/views/shared/listing_info.dart
```diff
@@ -0,0 +1,40 @@
...
+          if (searchQuery != null && searchQuery.isNotEmpty) ...[
+            d.text(' for search query '),
+            d.code(children: [d.text(searchQuery)]),
+          ],
+          if (ownedBy != null && ownedBy.isNotEmpty) ...[
+            d.text(' owned by '),
+            d.code(children: [d.text(ownedBy)]),
+          ],
+          if (messageMarkdown != null)
+            d.unsafeRawHtml(markdownToHtml(messageMarkdown)!),
```
jonasfj, line 34: May I propose that we make a `d.markdown` method?

This way, the `d.unsafeRawHtml` will only be used in a very few places, and it'll be easy to review.

jonasfj (approved):

PR #4969, opened 2021-07-22
Special DOM Node for markdown content.

isoos:


app/lib/frontend/dom/dom.dart
```diff
@@ -77,6 +82,9 @@ Node text(String value) => dom.text(value);
 /// Creates a DOM node with unsafe raw HTML content using the default [DomContext].
 Node unsafeRawHtml(String value) => dom.unsafeRawHtml(value);
 
+/// Creates a DOM node with markdown content using the default [DomContext].
+Node markdown(String text) => dom.markdown(text);
```
jonasfj, line 86: I don't think we need more than just:
```
Node markdown(String text) => unsafeRawHtml(markdownToHtml(text)!);
```

I don't think we need a context specific implementation for markdown.

Ideally, things should only be implemented on `dom` context if they behave differently in the client-side and server-side. I suspect markdown would have the same behavior, because we would probably use the same rendering logic.

jonasfj (approved):

PR #4972, opened 2021-07-23
Simpler way to specify dom children.

isoos:


app/lib/frontend/dom/dom.dart
```diff
@@ -289,6 +305,21 @@ Node ul({
       children: children,
     );
 
+Iterable<Node>? _children(Iterable<Node>? children, Node? child, String? text) {
+  if (children != null) {
+    assert(child == null);
```
jonasfj, line 310: Don't assert, throw an `ArgumentError`

app/lib/frontend/dom/dom.dart
```diff
@@ -289,6 +305,21 @@ Node ul({
       children: children,
     );
 
+Iterable<Node>? _children(Iterable<Node>? children, Node? child, String? text) {
+  if (children != null) {
+    ArgumentError.checkNotNull(child);
+    ArgumentError.checkNotNull(text);
```
jonasfj, line 311: Hmm... you probably want to check that it is null, hehe

app/lib/frontend/dom/dom.dart
```diff
@@ -289,6 +305,27 @@ Node ul({
       children: children,
     );
 
+Iterable<Node>? _children(Iterable<Node>? children, Node? child, String? text) {
+  if (children != null) {
+    if (child != null) {
+      throw ArgumentError('`child` is not null');
```
jonasfj, line 311:
```suggestion
      throw ArgumentError('Only one of `child`, `children` or `text` may be specified');
```

jonasfj (approved):

Issue #4807, opened 2021-05-18
experiment with splitting dartdoc generated content to use SSI


jonasfj:
Idea: split dartdoc generated to use server-side includes, like `<!--#include file="footer.html" -->` such that repeated content isn't duplicated in multiple files.

We should try to experiment with this in a post-processing step.. maybe just a one off script to run this over all packages and do a histogram on the size of the files that comes out..

Can they be stored in datastore? How many are there?

isoos:
On further experimenting, it turns out that `gunzip` does decode concatenated gzip chunks as a single file, however, Chrome displays only the first chunk. This old issue indicates that it was considered and rejected: https://bugs.chromium.org/p/chromium/issues/detail?id=20884

Looks like we might store it compressed, however, we'd need to decode it for concatenating. Or, store it compressed, without splitting the file?

jonasfj:
I'm not worried about having to decompress on the fly, we can test how slow gzip is, but my two cents says we can probably gzip between 100-300 MB/s... and we'll rarely have 5 MB.

Might be interesting to do a histogram over file-sizes, to see what the distribution is..

In any case, we probably don't have so many requests that we have to worry about decompression speed when serving the content. If we do, we cache the decompressed content in redis. Or even better we can use `dart:ffi` and load something like lz4 or zstd to do the compression/decompression -- lz4 is usually faster than network.

isoos:
This is now solved with dartdoc 6.3.0

PR #4979, opened 2021-07-26
Limit dartdoc output to 2GiB and 10M files.

isoos:
Follow-up to experimenting with #4807.


jonasfj (approved):
Should we consider throwing something other than an `AssertionError`... maybe throw  a `DocumentationTooBigException` or something else, which we can catch and create a nice error message / log using.

PR #4989, opened 2021-07-29
Pass customizer config to pkg/pub_dartdoc process.

isoos:
- #4807
- This PR stores the config file in the package directory. I don't think we'll need an extra command line argument for it, but if needed we can also do that later.


isoos:
As discussed, we may not need this after all.

PR #4991, opened 2021-07-29
Migrate pkg/labeled_scores.mustache

isoos:


app/lib/frontend/templates/views/pkg/labeled_scores.dart
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import '../../../dom/dom.dart' as d;
+
+d.Node renderLabeledScoresNode({
```
sigurdm, line 7: Perhaps name this without the `render`-prefix, and reserve that for the actual render to a string:
 ```suggestion
d.Node labeledScoresNode({
```

File: app/lib/frontend/templates/views/pkg/labeled_scores.dart [hunk 2]
```diff
@@ -0,0 +1,47 @@
...
+      ),
+      d.div(
+        classes: ['packages-score', 'packages-score-popularity'],
+        child: _renderLabeledScore('popularity', popularity, '%'),
+      ),
+    ],
+  );
+}
+
+d.Node _renderLabeledScore(String label, int? value, String sign) {
```
sigurdm, line 33: Could we make the `sign` param named to make the use-site a bit easier to understand?
```suggestion
d.Node _renderLabeledScore(String label, int? value, {required String sign}) {
```

sigurdm (approved):
LGTM with a few suggestions

app/lib/frontend/templates/views/pkg/labeled_scores.dart
isoos: I like that!

PR #4992, opened 2021-07-29
Migrate pkg/title_content.mustache

isoos:


app/lib/frontend/templates/views/pkg/title_content.dart
```diff
@@ -0,0 +1,36 @@
...
+            'data-ga-click-event': 'copy-package-version',
+          },
+          src: staticUrls.getAssetUrl('/static/img/content-copy-icon.svg'),
+          title: 'Copy "$package: ^$version" to clipboard',
+        ),
+        d.div(
+          classes: ['pkg-page-title-copy-feedback'],
+          children: [
+            d.span(classes: ['code'], text: '$package: ^$version'),
+            d.text('copied to clipboard'),
```
sigurdm, line 30: Do we need a space in front? I guess we don't since the tests pass....
```suggestion
            d.text(' copied to clipboard'),
```

sigurdm (approved):
LGTM

app/lib/frontend/templates/views/pkg/title_content.dart
isoos: That's a good point! I think the test pass, because we re-format the HTML and add a linebreak after the `<span>` in both cases.

PR #4993, opened 2021-07-30
Backend and API methods for retracted flag.

isoos:
- extended model object with `retracted` as the timestamp of retraction
- API similar to what we have for package options (+ handlers)
- backend methods to query the two required list (+ tests for them)
- backend method to update the retracted flag and timestamp (+ API tests)


jonasfj (approved):
@szakarias any comments, if not I suppose we should merge this and work on the UI changes.

jonasfj:
I'm guessing we still need:
 * [ ] UI for retraction in admin interface
 * [ ] Data in version listing API

PR #4994, opened 2021-07-30
Migrate the publisher_list.mustache template.

isoos:


app/lib/frontend/dom/dom.dart
```diff
@@ -92,10 +94,14 @@ Node a({
   String? text,
   String? href,
   String? rel,
+  String? target,
   String? title,
 }) {
-  final hasAttributes =
-      attributes != null || href != null || rel != null || title != null;
+  final hasAttributes = attributes != null ||
```
sigurdm, line 100: Do we really need to compute the `hasAttributes` here? It seems to me that _mergeAttributes handles an empty map just fine...

Also I think the `attributes` parameter could be made into a non-nullable with an empty map as default.

app/lib/frontend/templates/admin.dart
```diff
@@ -127,17 +127,19 @@ String renderAccountPublishersPage({
...
     tabs: [
       _myPackagesLink(),
       _myLikedPackagesLink(),
       Tab.withContent(
-          id: 'publishers',
-          title: 'My publishers',
-          contentHtml: publisherListHtml),
+        id: 'publishers',
+        title: 'My publishers',
+        contentHtml: publisherListNode.toString(),
```
sigurdm, line 141: Hmm, should we not represent the content as dom nodes, and wait with the rendering till we have everything? That probably belongs in another PR.

app/lib/frontend/templates/views/publisher/publisher_list.dart
```diff
@@ -0,0 +1,63 @@
...
+import '../../../../shared/urls.dart' as urls;
+import '../../../../shared/utils.dart';
+import '../../../dom/dom.dart' as d;
+
+d.Node renderPublisherListNode({
+  required List<PublisherSummary> publishers,
+  required bool isGlobal,
+}) {
+  return d.fragment([
+    if (isGlobal) d.element('h2', text: 'Publishers'),
```
sigurdm, line 15: We could consider helper functions for `h1`, `h2` etc.

sigurdm (approved):
LGTM

app/lib/frontend/dom/dom.dart
isoos: I think it is fine either way. IIRC we don't have many places where we emit an empty `<a>...</a>` tag without any attributes, so creating an additional map most of the time doesn't hurt much. But, on the other hand, it doesn't hurt to check either.  Maybe defer this decision a bit, and see what other tags we have and what is the default pattern?

app/lib/frontend/templates/admin.dart
isoos: That's exactly the plan, and as soon as the parent template will get migrated, it will be that way.

PR #5000, opened 2021-08-03
Migrate pkg/header.mustache

isoos:
- Fixes the missing `?hash` on the image link in `/static`
- Moves the `Prerelease:` label inside the `<span>`, making it consistent with the `Preview:` label.


sigurdm (approved):
LGTM

PR #5002, opened 2021-08-04
Updated integrity checks.

isoos:
- Consistency: references between `"` quotes, and sentences end with `.`.
- Removed cached data holders whenever they were unnecessary (reduces memory usage which may be relevant if we are running on a server).
- Updated a few checks, including the one where we are checking the one-to-many relationship between emails and User entities. This is now allowed, except for `@gmail.com` addresses.

### DIFF
--- a/app/lib/shared/integrity.dart
+++ b/app/lib/shared/integrity.dart
@@ -399,9 +399,10 @@ class IntegrityChecker {
       }
 
       final userId = like.userId!;
-      if (!_userToOauth.keys.contains(userId) ||
-          _deletedUsers.contains(userId)) {
-        yield 'Like entity with nonexisting or deleted user "$userId".';
+      if (!_userToOauth.keys.contains(userId)) {
+        yield 'Like entity with nonexisting user "$userId".';
+      } else if (_deletedUsers.contains(userId)) {
+         yield 'Like entity with deleted user "$userId".';
       }
 
       if (await _packageMissing(like.package!)) {
--- a/app/lib/shared/integrity.dart
+++ b/app/lib/shared/integrity.dart
@@ -20,7 +20,7 @@ class IntegrityChecker {
   final DatastoreDB _db;
   final int _concurrency;
 
-  final _userToOauth = <String, String>{};
+  final _userToOauth = <String, String?>{};
   final _oauthToUser = <String, String>{};
   final _deletedUsers = <String>{};
   final _invalidUsers = <String>{};
@@ -51,9 +51,7 @@ class IntegrityChecker {
     _logger.info('Scanning Users...');
     final gmailComEmails = <String>{};
     await for (User user in _db.query<User>().run()) {
-      if (user.oauthUserId != null) {
-        _userToOauth[user.userId!] = user.oauthUserId!;
-      }
+      _userToOauth[user.userId!] = user.oauthUserId;
       final email = user.email;
       if (email == null || email.isEmpty || !looksLikeEmail(email)) {
         yield 'User "${user.userId}" has invalid email: "${user.email}".';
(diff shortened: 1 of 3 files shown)


app/lib/shared/integrity.dart
```diff
@@ -19,296 +19,266 @@ final _logger = Logger('integrity.check');
...
-    await _checkUsers();
-    await _checkOAuthUserIDs();
-    await _checkPublishers();
-    await _checkPublisherMembers();
-    await _checkPackages();
-    await _checkVersions();
-    await _checkLikes();
-    await _checkModeratedPackages();
-    return _problems;
+  Stream<String> check() async* {
```
sigurdm, line 39: Would it make any sense to do some of these in parallel?

app/lib/shared/integrity.dart
```diff
@@ -19,296 +19,266 @@ final _logger = Logger('integrity.check');
...
+      //
+      // This is considered normal for non-gmail.com accounts, but should be impossible
+      // for gmail.com accounts, as they cannot be recreated.
+      if (email != null &&
+          email.endsWith('@gmail.com') &&
+          !gmailComEmails.add(email)) {
+        yield 'Email address "$email" is present at "${user.userId}" and another Users.';
       }
+
       if (user.isDeleted is! bool) {
```
sigurdm, line 73: Hmm with this kind of manual type-checking we could consider having schemas for these entries to automate this...

Not this PR of course

app/lib/shared/integrity.dart
```diff
@@ -426,50 +386,45 @@ class IntegrityChecker {
...
     _logger.info('Scanning Likes...');
 
-    await for (Like like in _db.query<Like>().run()) {
+    await for (final like in _db.query<Like>().run()) {
       if (like.packageName == null) {
-        _problems.add(
-            'Like entity for user ${like.userId} and ${like.package} has a '
-            '`packageName` property which is not at string ');
+        yield 'Like entity for user "${like.userId}" and package "${like.package}" has a '
+            '`packageName` property which is not at string.';
```
sigurdm, line 395:
```suggestion
            '`packageName` property which is not a string.';
```

app/lib/shared/integrity.dart
```diff
@@ -426,50 +386,45 @@ class IntegrityChecker {
...
-        .where((user) =>
-            (!_userToOauth.keys.contains(user) || _deletedUsers.contains(user)))
-        .forEach((user) {
-      _problems.add('Like entity with nonexisting or deleted user $user');
-    });
+      final userId = like.userId!;
+      if (!_userToOauth.keys.contains(userId) ||
+          _deletedUsers.contains(userId)) {
+        yield 'Like entity with nonexisting or deleted user "$userId".';
+      }
```
sigurdm, line 405: We might want to split this in two cases to make the reporting precise.
```suggestion
      if (!_userToOauth.keys.contains(userId)) {
        yield 'Like entity with nonexisting user "$userId".';
      } else if (_deletedUsers.contains(userId)) {
         yield 'Like entity with deleted user "$userId".';
      }
```

sigurdm (approved):
LGTM with a few suggestions

app/lib/shared/integrity.dart
isoos: In some case it may be good, but others rely on cached values from earlier methods.

app/lib/shared/integrity.dart
isoos: Hm, this may be coming from a different era, I think the current Dart SDK would just throw an exception if the value would be non-bool. I think we could just remove these checks for non-nullable values, wdyt?

sigurdm: We should probably verify what happens - then I'm all for removing the checks if they are redundant.
If null checks are still relevant we might again want to have a kind of schema-definition for each entity, ensuring that we check all non-nullable fields...

isoos: Agreed. I think we should do that in a follow-up.
https://github.com/dart-lang/pub-dev/issues/5004

Issue #4750, opened 2021-04-25
Package not getting analysed for > 1 day again.


tiloc:
Unfortunately, it is happening again... I could not find a button to re-open the original Issue #4694, so creating a new one.

_Originally posted by @tiloc in https://github.com/dart-lang/pub-dev/issues/4694#issuecomment-826400020_

isoos:
We are blocked on https://github.com/dart-lang/dartdoc/issues/2615

Luckily - for now -, that the retry (~a day later) did seem to run successfully.

filiph:
How often does pub.dev retry these? We have another report: #4766.

filiph:
I also seem to remember there was a way to look at logs for a package. Like, there was a link, something like `pub.dev/my_package/logs.txt` or something. I can't find the link now, though. It might help people see what's happening.

isoos:
> How often does pub.dev retry these? 

About once a day. I've updated the other report, it seems that we have a partially migrated traffic.

sigurdm:
It seems to be analyzed now.

PR #5003, opened 2021-08-04
Display a report section on uncaught dartdoc tool failure.

isoos:
#4750

### DIFF
--- a/app/lib/dartdoc/dartdoc_runner.dart
+++ b/app/lib/dartdoc/dartdoc_runner.dart
@@ -195,7 +195,7 @@ class DartdocJobProcessor extends JobProcessor {
           job,
           _emptyReport(
             title: 'Did not run dartdoc',
-            description: 'Package supports only old SDK.',
+            description: 'This package is not compatible with the current analysis SDK version',
           ));
       return JobStatus.skipped;
     }
--- a/app/lib/dartdoc/dartdoc_runner.dart
+++ b/app/lib/dartdoc/dartdoc_runner.dart
@@ -211,7 +211,7 @@ class DartdocJobProcessor extends JobProcessor {
           _emptyReport(
             title: 'Did not run dartdoc',
             description:
-                'Package version is older than two years and has newer release.',
+                'Package version is older than two years and has a newer release.',
           ));
       return JobStatus.skipped;
     }
--- a/app/lib/dartdoc/dartdoc_runner.dart
+++ b/app/lib/dartdoc/dartdoc_runner.dart
@@ -195,7 +195,8 @@ class DartdocJobProcessor extends JobProcessor {
           job,
           _emptyReport(
             title: 'Did not run dartdoc',
-            description: 'This package is not compatible with the current analysis SDK version',
+            description:
+                'This package is not compatible with the current analysis SDK version',
           ));
       return JobStatus.skipped;
     }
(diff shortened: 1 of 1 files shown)


app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -190,7 +191,12 @@ class DartdocJobProcessor extends JobProcessor {
     // We know that dartdoc will fail on this package, no reason to run it.
     if (packageStatus.isLegacy) {
       _logger.info('Package is on legacy SDK: $job.');
-      await _storeScoreCard(job, _emptyReport());
+      await _storeScoreCard(
+          job,
+          _emptyReport(
+            title: 'Did not run dartdoc',
+            description: 'Package supports only old SDK.',
```
sigurdm, line 198:
```suggestion
            description: 'This package is not compatible with the current analysis SDK version',
```

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -200,7 +206,13 @@ class DartdocJobProcessor extends JobProcessor {
     if (packageStatus.isObsolete) {
       _logger
           .info('Package is older than two years and has newer release: $job.');
-      await _storeScoreCard(job, _emptyReport());
+      await _storeScoreCard(
+          job,
+          _emptyReport(
+            title: 'Did not run dartdoc',
+            description:
+                'Package version is older than two years and has newer release.',
```
sigurdm, line 214:
```suggestion
                'Package version is older than two years and has a newer release.',
```

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -216,7 +228,12 @@ class DartdocJobProcessor extends JobProcessor {
         // These package versions are worth investigating, but we don't need
         // alerts on them.
         _logger.warning('Missing dartdoc report when error was present: $job.');
-        await _storeScoreCard(job, _emptyReport());
+        await _storeScoreCard(
+            job,
+            _emptyReport(
+              title: 'Failed to run dartdoc',
+              description: 'The dartdoc process timed out.',
```
sigurdm, line 235: Do we have a way to indicate the time it took? Should we?

app/lib/dartdoc/dartdoc_runner.dart
```diff
@@ -656,10 +673,26 @@ String _mergeOutput(ProcessResult pr, {bool compress = false}) {
...
+    documentationSection: ReportSection(
+      id: ReportSectionId.documentation,
+      title: documentationSectionTitle,
+      grantedPoints: 0,
+      maxPoints: 10,
+      summary: renderSimpleSectionSummary(
+        title: title,
+        description: description,
+        grantedPoints: 0,
+        maxPoints: 10,
```
sigurdm, line 693: Is maxPoints 10 the right thing here? Maybe it doesn't really matter...

sigurdm (approved):
LGTM

app/lib/dartdoc/dartdoc_runner.dart
isoos: This is a tricky one: we have dartdoc runs that neither compete, nor time out (the timeout doesn't kill the process). Once we use the proper kill signal to do it (in this new release), I hope this class of error is gone and we can remove this block.

sigurdm (approved):
LGTM

PR #5005, opened 2021-08-06
Partial migration of package infobox templates.

isoos:
- changes output to always include a formatted license name, or `unknown` if it is not recognized
- moves these parts to `views/pkg/info_box.dart`, which will be also the location of the migrated mustache template

### DIFF
--- a/app/lib/frontend/templates/views/pkg/info_box.dart
+++ b/app/lib/frontend/templates/views/pkg/info_box.dart
@@ -26,19 +26,21 @@ d.Node? dependencyListNode(Map<String, pubspek.Dependency>? dependencies) {
   if (dependencies == null) return null;
   final packages = dependencies.keys.toList()..sort();
   if (packages.isEmpty) return null;
-  return d.fragment(
-    packages.expand((p) {
-      final dep = dependencies[p];
-      var href = redirectPackageUrls[p];
-      String? constraint;
-      if (href == null && dep is pubspek.HostedDependency) {
-        href = urls.pkgPageUrl(p);
-        constraint = dep.version.toString();
-      }
-      return <d.Node>[
-        d.text(', '),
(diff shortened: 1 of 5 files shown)


app/lib/frontend/templates/views/pkg/info_box.dart
```diff
@@ -0,0 +1,44 @@
...
+      String? constraint;
+      if (href == null && dep is pubspek.HostedDependency) {
+        href = urls.pkgPageUrl(p);
+        constraint = dep.version.toString();
+      }
+      return <d.Node>[
+        d.text(', '),
+        href == null ? d.text(p) : d.a(href: href, title: constraint, text: p),
+      ];
+    }).skip(1), // skips the first comma (', ').
```
sigurdm, line 42: Nit: I don't like this - I don't have a really much better solution - maybe something like:
```
[for (var i = 0; i < packages.length; i++) ...[if (i != 0) ',', hrefOfPackage(packages[i]]]
```

`List` could really use a `intersperse` helper method...

sigurdm (approved):
LGTM

app/test/frontend/golden/pkg_show_page_discontinued.html
```diff
@@ -238,7 +238,9 @@ <h3 class="title">Documentation</h3>
             </p>
             <h3 class="title">License</h3>
             <p>
+              unknown (
```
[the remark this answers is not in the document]
isoos: `pkg_show_page.html` shows a package with a BSD license. The discontinued package doesn't get analysis, and with that, license detection is not run either, hence the page will display unknown.

app/lib/frontend/templates/views/pkg/info_box.dart
isoos: yeah, I am no fan either. I'll rewrite it to the for loop...

PR #5006, opened 2021-08-06
Datastore integrity checks as periodic task.

isoos:


app/bin/tools/check_integrity.dart
```diff
@@ -33,7 +33,7 @@ Future main(List<String> args) async {
 
   await withToolRuntime(() async {
     final checker = IntegrityChecker(dbService, concurrency: concurrency);
-    final problems = await checker.check().toList();
+    final problems = await checker.findProblems().toList();
     print('\nProblems detected: ${problems.length}\n');
```
sigurdm, line 37: Maybe print something else if the list is empty

app/lib/tool/neat_task/pub_dev_tasks.dart
```diff
@@ -27,6 +28,14 @@ void setupFrontendPeriodicTasks() {
     task: backfillNewFields,
   );
 
+  // Checks the Datastore integrity of the model objects.
+  _weekly(
```
sigurdm, line 32: Do we ensure that all weekly tasks don't run at the same time?

app/lib/tool/neat_task/pub_dev_tasks.dart
```diff
@@ -27,6 +28,14 @@ void setupFrontendPeriodicTasks() {
...
+    name: 'check-datastore-integrity',
+    isRuntimeVersioned: true,
+    task: () async =>
+        await IntegrityChecker(dbService, concurrency: 2).verifyAndLogIssues(),
```
sigurdm, line 36: What happens in case the integrity check crashes? What effect will it have on the service?

app/lib/shared/integrity.dart
```diff
@@ -35,8 +36,18 @@ class IntegrityChecker {
   IntegrityChecker(this._db, {int? concurrency})
       : _concurrency = concurrency ?? 1;
 
-  /// Runs integrity checks, and reports the list of problems.
-  Stream<String> check() async* {
+  /// Runs integrity checks, and reports the problems via a [Logger].
+  Future<void> verifyAndLogIssues() async {
+    var count = 0;
```
sigurdm, line 41: Should we log that we initiate the integrity check?

sigurdm (approved):
LGTM but wait landing until jonasfj had a say...

app/lib/tool/neat_task/pub_dev_tasks.dart
isoos: Yeah, `package:neat_periodic_task` does the heavy lifting there, and we have a Datastore-based state store for it.

app/lib/tool/neat_task/pub_dev_tasks.dart
isoos: Good question, I guess it will bring down the isolate + probably the VM, making that instance to reboot. Luckily there will be some time before it is being retried. I'm planning to add some memory logging before/after the periodic tasks are running, so that we will have some data there.

app/lib/shared/integrity.dart
isoos: the neat_periodic_task wrapper does it, and also its status (successful or failing)

app/lib/tool/neat_task/pub_dev_tasks.dart
sigurdm: Is there any chance we could move it to another isolate/process?

isoos: This is already a non-web-serving isolate. We could move this to a non-frontend service, e.g. `analyzer`, that way we don't risk bringing down the API...

isoos:
Pushing this to staging to check how it works.

isoos:
Staging looks good:
```
I 2021-08-09T16:04:23.975Z neat_periodic_task: ### [START] neat-periodic-task: "check-datastore-integrity" 
I 2021-08-09T16:34:40.731Z neat_periodic_task: ### [FINISHED] neat-periodic-task: "check-datastore-integrity" 
```
several issues were reported (which were fixed on prod)

app/lib/shared/integrity.dart
```diff
@@ -35,8 +36,18 @@ class IntegrityChecker {
...
+    await for (final problem in findProblems()) {
+      count++;
+      _logger.reportError(problem);
```
sigurdm, line 44: Should we prefix these to make them easier to filter for?

isoos: Wouldn't the logger's name enough for that?

PR #5008, opened 2021-08-06
Push non-nullable operator into some model objects.

isoos:


app/lib/publisher/models.dart
```diff
@@ -58,7 +58,7 @@ class Publisher extends db.ExpandoModel<String> {
 @db.Kind(name: 'PublisherInfo', idType: db.IdType.String)
 class PublisherInfo extends db.ExpandoModel<String> {
   /// The associated domain name of the publisher.
-  String? get publisherId => id;
+  String get publisherId => id!;
```
sigurdm, line 61: How can we be sure this check succeeds?

isoos: 1. After the entity is loaded from Datastore, it must have a non-null `id`.
2. Otherwise, we can't. But as you can see, we do force `publisherId` to non-null every time, we may do it in a single place instead.

sigurdm: Right, that makes sense!

sigurdm (approved):
LGTM

PR #5009, opened 2021-08-10
Force upgrade package:analyzer.

isoos:
I've found two cases where the 1.7.2+22.0.0 versions produced cast exceptions we have in the logs, while the 2.0.0+23.0.0 already fixed it. As in pub-dev only pana is using it in live code (otherwise we use it only in codegen), and pana is already upgraded on CI to use it, I think there is little to no risk for us to upgrade it before the toolchain would allow it.


isoos:
Ah, looks like this is not working yet, as codegen gets confused by many fields.

Issue #4858, opened 2021-06-07
Remove uploader emails from pub.dev


jonasfj:
Since we have successfully introduced publishers, emails is no longer the only form of supported identity on pub.dev

So let's try to stop displaying the "uploader" emails.

jonasfj:
Now we should also simply say "publisher: unknown", when there is uploaders are used:

![4A2bnis9sWz2nLa](https://user-images.githubusercontent.com/149732/120998983-e5370500-c788-11eb-88ba-33db7de8610b.png)

mzdm:
But how do you search now for all packages made by the same author? It was very intuitive that you could just click on the email address. And when I saw Unknown status I thought that the package was compromised / not validated.

PR #5010, opened 2021-08-10
Remove email: search query.

isoos:
Follow-up on #4858.


jonasfj (approved):
This won't break "my packages"?

isoos:
> This won't break "my packages"?

No, that's using a different field in `PackageDocument`. I'm preparing another PR for migrating that off of emails and use userIds.

PR #5011, opened 2021-08-10
Populate and use PackageDocument.uploaderUserIds for /my-packages.

isoos:
- Follow-up on #4858.
- This is a backward-compatible change: old frontend requests are served correctly, also client reaching an old search service instance will keep working the same way.
- After two new release we will be able to remove `uploaderEmails` from the search service.


sigurdm (approved):
LGTM
Do we alsowant a backfill?

isoos:
> Do we also want a backfill?

No, because this search snapshot is runtimeVersioned, and older deployed versions won't use it anyway. However, we want to have a `runtimeVersion` bump, I'll add it to this PR.

PR #5017, opened 2021-08-12
Deduplicate Youtube videoIds.

isoos:


app/lib/service/youtube/backend.dart
```diff
@@ -110,24 +113,33 @@ class _PkgOfWeekVideoFetcher {
...
                 (i.snippet?.description ?? '').trim().split('\n').first,
             thumbnailUrl: i.snippet!.thumbnails!.high!.url!,
-          ),
-        ));
+          );
+          if (videoIds.add(video.videoId)) {
+            videos.add(video);
+          } else {
+            _logger.warning(
+                'Duplicate videoId "${video.videoId}" on page $pageIndex.');
```
sigurdm, line 136: Should we also log the previous page this occurred on?

sigurdm (commented):
test?

app/lib/service/youtube/backend.dart
isoos: Do you think it is useful? We can't really use the information, not even the `pageIndex` much:
- the sort order is time descending, a new item will push the page indexes
- the pagination is based on continuation token.

isoos:
> test?

We have a sanity-check local test with local dev key provided:
https://github.com/dart-lang/pub-dev/blob/master/app/test/service/youtube/backend_test.dart#L15-L20

isoos:
Hm, actually, the test's expectation is not really great, because it has only a lower limit on the number of items. It seems we get much more than expected.

app/lib/service/youtube/backend.dart
sigurdm: Hmm - do we then want to log pageIndex at all?

Only thing I was curious about was if the duplicates could happen on the same page or not....

isoos: Ah, that's a good point!

isoos:
#5020 has the correct fix

PR #5021, opened 2021-08-13
Reduce risk of running periodic tasks

isoos:
- move tasks out of frontend instances and run them on analyzer instances
- log max memory use before and after the task, warn if it grew a larger amount


sigurdm (approved):
LGTM

Not sure how much we can trust `ProcessInfo.maxRss`, but I guess it would be interesting to see.

isoos:
> Not sure how much we can trust ProcessInfo.maxRss, but I guess it would be interesting to see.

If there is a trend, then it is an interesting information, the exact KiB change is less interesting.

PR #5023, opened 2021-08-13
Display private|admin lock sign on account-related tabs.

isoos:
image

(first underline was my mouse hover over the menu item, while the last is the active tab).

### DIFF
--- a/app/lib/frontend/templates/admin.dart
+++ b/app/lib/frontend/templates/admin.dart
@@ -64,7 +64,7 @@ String renderAccountPackagesPage({
     headerHtml: _accountDetailHeader(user, userSessionData),
     tabs: [
       Tab.withContent(
-          id: 'packages', title: 'My packages', contentHtml: tabContent),
+          id: 'my-packages', title: 'My packages', contentHtml: tabContent),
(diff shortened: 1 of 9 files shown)


app/lib/frontend/templates/admin.dart
```diff
@@ -64,7 +64,7 @@ String renderAccountPackagesPage({
     headerHtml: _accountDetailHeader(user, userSessionData),
     tabs: [
       Tab.withContent(
-          id: 'packages', title: 'My packages', contentHtml: tabContent),
+          id: 'my-packages', title: 'My packages', contentHtml: tabContent),
```
sigurdm, line 67: Will this change break existing links to that tab?

sigurdm: Perhaps just let the Tab.withContent take another parameter telling if it's private?

isoos: > Will this change break existing links to that tab?

No, tabs are no longer referenced by their ids, and these are private (authenticated) URLs anyway.

sigurdm (approved):
LGTM

PR #5026, opened 2021-08-16
Migrate account-related static templates.

isoos:
- removed the parameter(s) which were never set, and we are always using them with the default value
- with that, these templates could be just `.md` files... do we want to explore that route?


sigurdm:
> with that, these templates could be just .md files... do we want to explore that route?

I think that for now it is fine to express the dom nodes. I don't think it will be slower than processing md files, it will be more uniform, and it will be easier for us when some dynamic content should go into the template.

sigurdm (approved):
LGTM

PR #5034, opened 2021-08-20
more permissions to pub-moderation-admin

jonasfj:


sigurdm (approved):
LGTM

Should we also update: https://github.com/dart-lang/pub-dev/blob/19033f8154ca1f597ef5495acbc84a2bb368f16d/app/lib/shared/configuration.dart#L186 ?

isoos:
> Should we also update:

Yes, we should. At the moment the configuration is not yet migrated fully, and I think the code-based config is used. It is strange that the test didn't fail on changing the yaml.

isoos:
Facepalm moments: tests were not running, because the filename was not ending in `_test.dart` :(
Anyway, fixed that, plus small cosmetic changes, @sigurdm please take a look!

sigurdm (approved):
LGTM - though one test seems failing

jonasfj:
@googlebot I consent.

jonasfj:
Pretty sure @isoos has signed the CLA.

isoos:
@googlebot I consent.

PR #5037, opened 2021-08-23
Migrate consent/page.mustache

isoos:
I've started `material.dart` to collect the material-related templates into a single place. There will be inputs, dropdown and possibly other components when we migrate the rest of the templates.


sigurdm (approved):
LGTM

PR #5038, opened 2021-08-23
Reduce log level of max RSS periodic task messages.

isoos:
Last week's deployment showed that this may not be an issue at all, let's reduce the logging level so that we don't have too much noise.


sigurdm (approved):
LGTM

PR #5040, opened 2021-08-23
Increase the number of search service instances to 4 (autoscale up to 6)

isoos:
Also added a debug info about the reported CPU count to cross-check the number with the requested ones.


sigurdm (approved):
LGTM

PR #5069, opened 2021-09-01
Migrate pkg/install_tab.mustache

isoos:
- removed the numbered section header from before "Install it" (see [pana install page](https://pub.dev/packages/pana/install))
- otherwise no changes, maybe a bit simpler logic


app/lib/frontend/templates/views/pkg/install_tab.dart
```diff
@@ -0,0 +1,150 @@
...
+    d.h2(text: 'Use this package as an executable'),
+    d.h3(text: 'Install it'),
+    d.p(text: 'You can install the package from the command line:'),
+    d.pre(
+      attributes: {'data-text-to-copy': 'dart pub global activate $package'},
+      child: d.code(
+        classes: ['language-shell'],
+        child: d.strong(text: 'dart pub global activate $package'),
+      ),
+    ),
```
jonasfj, line 49: Why don't we make a helper method for `codeSnippet(language: ..., text: , child: , children: )`.

Then we don't have to write `pre` and `code`, and all the attributes.. ideally, we don't even need to supply `data-text-to-copy` if `text` is given..

app/lib/frontend/templates/views/pkg/install_tab.dart
```diff
@@ -0,0 +1,150 @@
...
+    d.h2(text: 'Use this package as a library'),
+    d.h3(text: 'Depend on it'),
+    d.p(text: 'Run this command:'),
+    if (usePubGet) _pubAddNode('Dart', 'dart $pubAddCommand'),
+    if (useFlutterPackagesGet) _pubAddNode('Flutter', 'flutter $pubAddCommand'),
+    d.p(children: [
+      d.text('This will add a line like this to your package\'s '
+          'pubspec.yaml (and run an implicit '),
+      d.code(
+          classes: ['language-shell'],
```
jonasfj, line 95: Does this have any effect, when it's not inside a pre?

jonasfj (approved):
Consider if we want a `codeSnippet` helper, or we can refactor to create such a function later.

app/lib/frontend/templates/views/pkg/install_tab.dart
isoos: Looks like hljs doesn't extend it with its own attributes, so I'll just remove it.

PR #5072, opened 2021-09-02
Extra debug info in pkg/pub_dartdoc process.

isoos:


pkg/pub_dartdoc/bin/pub_dartdoc.dart
```diff
@@ -11,7 +11,8 @@ import 'package:pub_dartdoc/pub_data_generator.dart';
 import 'package:pub_dartdoc/src/pub_hooks.dart';
 
 Future<void> main(List<String> arguments) async {
-// pub hooks
+  print('[${DateTime.now().toIso8601String()}] Starting...');
```
jonasfj, line 14: You don't want the line to also say what package / version it's running on?

jonasfj (approved):

pkg/pub_dartdoc/bin/pub_dartdoc.dart
isoos: All of this output goes to the `log.txt` which we know which package and version it is running on.

PR #5075, opened 2021-09-03
Migrate the detail page header template.

isoos:
There are a few changes in the template:
- `aria-pressed` was set to `true`, but it needs to be true only if it is active/on
- empty tags and likes `<div>`s are no longer emitted

Otherwise we have the usual attribute order changes.


sigurdm (approved):
LGTM

PR #5087, opened 2021-09-08
Migrate package admin page

isoos:
- first commit fixes the missing content from the golden file
- allows `viewBox` (cased) attribute on `svg` tag
- changes `<i>` to `<em>` in blocks that are now implemented with markdown
- we have the usual attribute order changes


app/lib/frontend/dom/dom.dart
```diff
@@ -399,6 +399,32 @@ Node li({
       children: _children(children, child, text),
     );
 
+/// Creates an `<option>` Element using the default [DomContext].
+Node option({
+  String? id,
+  Iterable<String>? classes,
+  Map<String, String>? attributes,
+  Iterable<Node>? children,
+  Node? child,
```
jonasfj, line 408: Does it ever make sense to have children or child, isn't `text` the only thing worth using?

app/lib/frontend/templates/views/pkg/admin_page.dart
```diff
@@ -0,0 +1,159 @@
...
+          children: [
+            d.p(
+                text: 'You can invite new uploader to this package. '
+                    'Once new uploaders accept the invitation, they have full administrative rights, '
+                    'with the following abilities:'),
+            d.ul(children: [
+              d.li(text: 'Transfer this package to a publisher'),
+              d.li(text: 'Upload new versions of this package'),
+              d.li(text: 'Invite and remove uploaders of this package'),
+            ]),
```
jonasfj, line 99: nit. Consider using markdown..

app/lib/frontend/dom/material.dart
```diff
@@ -246,3 +246,70 @@ d.Node dataTable<T>({
...
+          ),
+          d.div(classes: ['mdc-checkbox__ripple']),
+        ],
+      ),
+      d.label(attributes: {'for': id}, text: label),
+    ],
+  );
+}
+
+/// Renders a material dropdown / select component.
```
jonasfj, line 295: Let's document that `options` should be a list of `d.option`

jonasfj (approved):

app/lib/frontend/dom/dom.dart
isoos: I think material and other frameworks allow the creation of complex selects, with icons or formatted content. Not sure if they need rich content or data attributes for that, but let's keep that open.

app/lib/frontend/templates/views/pkg/admin_page.dart
isoos: Let's refactor that next time this part needs an update.

jonasfj (approved):

PR #5088, opened 2021-09-08
Migrate the detail/tabs.mustache template

isoos:
- refactored `Tab` a bit
- replacing `contentHtml` with `Node` will be after all of the tab contents are migrated and merged
- no change in the goldens


jonasfj (approved):

app/lib/frontend/templates/views/shared/detail/tabs.dart
```diff
@@ -0,0 +1,43 @@
...
+    ),
+    d.div(
+      classes: ['detail-container', 'detail-body-main'],
+      child: d.div(
+        classes: ['detail-tabs-content'],
+        children: tabs.where((t) => t.hasContent).map(
+              (t) => d.element(
+                'section',
+                classes: t.contentClasses,
+                child: d.unsafeRawHtml(t.contentHtml!),
```
jonasfj, line 37: This is a nice way to partially migrate... when we're done we can just search for `unsafeRawHtml` to find places we've missed...

PR #5090, opened 2021-09-09
Use lints/core.yaml in analysis_options.yaml override.

isoos:


app/lib/analyzer/pana_runner.dart
```diff
@@ -22,8 +22,8 @@ import '../shared/tool_env.dart';
 final Logger _logger = Logger('pub.analyzer.pana');
 
 @visibleForTesting
-final pedanticAnalysisOptionsYaml = File(p.join(static_files.resolveAppDir(),
-        'lib', 'analyzer', 'pedantic-analysis_options.1.8.0.yaml'))
+final defaultAnalysisOptionsYaml = File(p.join(static_files.resolveAppDir(),
```
sigurdm, line 25: Could we instead load this directly from package:lints with a resolvePackageUri?

sigurdm (approved):
LGTM

app/lib/analyzer/pana_runner.dart
isoos: I'll do that in a follow-up PR.

PR #5103, opened 2021-09-13
Disable report size trimming in ScoreCard.

isoos:


app/lib/scorecard/backend.dart
```diff
@@ -218,7 +218,9 @@ class ScoreCardBackend {
         if (size > _reportSizeDropThreshold) {
           _logger.reportError(
               '$reportType report exceeded size threshold ($size > $_reportSizeWarnThreshold)');
-          return true;
+          // Size trimming is temporarily disabled.
+          // TODO: re-enable after fixing the underlying issues with the verbose reports and package:appengine.
+          return false;
```
sigurdm, line 223: How about we instead make the truncated report something meaningful?
```
gzip.encode(json.encode(Report(sections: [])));
```

jonasfj: Further down we actually have:
```
      if (sizeCheck(ReportType.pana, scoreCard.panaReportJsonGz)) {
        // TODO: replace with something meaningful
        scoreCard.panaReportJsonGz = <int>[];
      }
```

What you are proposing seems meaningful to me :D


Or we just make a single artificial section that says: "Failed to run pana" or something like that.

jonasfj (commented):
Don't we need to do this? Or is enough that we have trimmed output from various commands?

app/lib/scorecard/backend.dart
isoos: I planned to have something like that (as the TODO suggests it), but there may be a side effect which I don't recall now, so I'd rather just turn it off for this release, and revisit what side effects we may have.

jonasfj (approved):

PR #5113, opened 2021-09-16
Prepare migration of detail page rendering.

isoos:
- using `Node` as parameters (refactored callers without much change in code structure)
- `<script>` content is passed-in as unescaped HTML

### DIFF
--- a/app/lib/frontend/dom/dom.dart
+++ b/app/lib/frontend/dom/dom.dart
@@ -459,6 +459,27 @@ Node pre({
       children: _children(children, child, text),
     );
 
+/// Creates a `<script>` Element using the default [DomContext].
+Node script({
+  String? id,
+  Iterable<String>? classes,
+  Map<String, String>? attributes,
+  Iterable<Node>? children,
+  Node? child,
+  String? text,
+  String? type,
+}) =>
+    dom.element(
+      'script',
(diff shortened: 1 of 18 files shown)


app/lib/frontend/dom/dom.dart
```diff
@@ -111,6 +111,15 @@ Node codeSnippet({
   );
 }
 
+/// Creates a DOM element with ld+json `<script>` content.
+Node ldJson(Map<String, dynamic> content) {
+  return script(
+    type: 'application/ld+json',
+    // TODO: check how this should be escaped
+    child: unsafeRawHtml(json.encode(content)),
+  );
```
jonasfj, line 120:
```suggestion
  final rawJson = json.encode(content);
  final rawHtml = rawJson.replaceMapped(
    // As rawJson can only contain the following characters inside of
    // a JSON string, we can always encode them as \u00xx
    // This ensures that html tags cannot be embedded.
    RegExp(r'</>'),
    (m) => r'\u00' + hex.encode(m[0]!.codeUnitAt(0)),
  );
  return script(
    type: 'application/ld+json',
    child: unsafeRawHtml(rawHtml),
  );
```

jonasfj (approved):

PR #5114, opened 2021-09-17
Migrate site footer

isoos:
- `class` attribute order changed in one place, it does not affect rendering
- the usual attribute order change in other places


sigurdm (approved):
LGTM

PR #5118, opened 2021-09-20
Migrate layout template

isoos:
- cleaner output (e.g. empty `class` attributes)
- the usual attribute reordering
- extended ld+json escape with `!` (merged the Node-creating method with the current util's escaping)


app/lib/frontend/dom/dom.dart
```diff
@@ -119,8 +119,10 @@ Node ldJson(Map<String, dynamic> content) {
   final rawHtml = rawJson.replaceAllMapped(
     // As rawJson can only contain the following characters inside of
     // a JSON string, we can always encode them as \u00xx
-    // This ensures that html tags cannot be embedded.
-    RegExp(r'[</>]'),
+    // This ensures that HTML tags cannot be embedded, e.g.:
+    // - <!-- to prevent HTML comments
+    // - < and </ to prevent <script> tags
+    RegExp(r'[</>!]'),
```
jonasfj, line 125: Is there anything else we should consider escaping? Maybe, non-ascii? Could someone inject null-bytes or other _interesting_ things here?

jonasfj (approved):

app/lib/frontend/dom/dom.dart
isoos: Agreed, maybe everything non-ascii could be escaped. I'll address it in a follow-up PR.

PR #5124, opened 2021-09-21
Migrate Tab.contentHtml to Tab.contentNode

isoos:


app/lib/frontend/templates/_utils.dart
```diff
@@ -8,32 +8,34 @@ import 'package:path/path.dart' as p;
...
   if (_isMarkdownFile(filename)) {
-    return markdownToHtml(content,
+    return d.unsafeRawHtml(
+      markdownToHtml(
+        content,
         baseUrl: baseUrl,
         baseDir: p.dirname(filename),
-        isChangelog: isChangelog)!;
+        isChangelog: isChangelog,
+      )!,
```
jonasfj, line 30: Can't we create a `d.safeHtml` which calls `sanitizeHtml`, then we don't have to call `sanitizeHtml` in `markdownToHtml`.

Or maybe we refactor these bits later... Just saying, we can slowly start to remove some of the calls to `d.unsafeRawHtml`, if we build things like `d.safeHtml`...

jonasfj (approved):

jonasfj:
I love all the `.toString()` calls that now disappear! :rocket:

app/lib/frontend/templates/_utils.dart
isoos: Yeah, we should certainly refactor some parts. We also have `unsafeRawHtml` in the announcement banner and in the consent description rendering. I think those are similar use-cases where we could use the sanitization on the HTML text.

PR #5149, opened 2021-09-30
Rewrite reports when size limit is exceeded.

isoos:


app/lib/package/model_properties.dart
```diff
@@ -107,7 +107,6 @@ class Pubspec {
   SdkConstraintStatus get _sdkConstraintStatus =>
       SdkConstraintStatus.fromSdkVersion(_inner.environment?['sdk'], name);
 
-  // TODO: migrate uses to SdkConstraintStatus.isDart2Compatible
```
sigurdm, line 110: Did we do this, or we no longer want to?

app/lib/scorecard/backend.dart
```diff
@@ -218,23 +219,68 @@ class ScoreCardBackend {
...
-        // TODO: replace with something meaningful
-        scoreCard.panaReportJsonGz = <int>[];
+      if (panaReport != null &&
+          sizeCheck(ReportType.pana, scoreCard.panaReportJsonGz)) {
+        scoreCard.updateReports(
+          panaReport: PanaReport(
+            timestamp: DateTime.now().toUtc(),
+            panaRuntimeInfo: panaReport.panaRuntimeInfo,
+            reportStatus: panaReport.reportStatus,
+            derivedTags: panaReport.derivedTags?.take(100).toList(),
```
sigurdm, line 237: Could this still be problematic if the individual tags are somehow very long?

app/lib/scorecard/backend.dart
```diff
@@ -218,23 +219,68 @@ class ScoreCardBackend {
...
+            flags: panaReport.flags,
+            urlProblems: panaReport.urlProblems?.take(10).toList(),
+          ),
+        );
       }
-      if (sizeCheck(ReportType.dartdoc, scoreCard.dartdocReportJsonGz)) {
-        // TODO: replace with something meaningful
-        scoreCard.dartdocReportJsonGz = <int>[];
+      if (dartdocReport != null &&
+          sizeCheck(ReportType.dartdoc, scoreCard.dartdocReportJsonGz)) {
```
sigurdm, line 266: Nit: Could we rename `sizeCheck` to something indicating what a true result means? Perhaps something like `reportIsTooBig`

app/lib/scorecard/backend.dart
```diff
@@ -218,23 +219,68 @@ class ScoreCardBackend {
...
+                        (s) => s.summary.length < 1000
+                            ? s
+                            : pana.ReportSection(
+                                id: s.id,
+                                title: s.title,
+                                status: s.status,
+                                grantedPoints: s.grantedPoints,
+                                maxPoints: s.maxPoints,
+                                summary: [
+                                  s.summary.substring(0, 900),
```
sigurdm, line 252: This is fine, but if we wanted to be really advanced we could take 500 chars from the beginning, and 400 from the end. The end might have some relevant information about what went wrong.

sigurdm: Can we run into unicode issues here when we do the raw substring? Not sure how important it is to handle, but I think this might split utf-16 4-byte encoded characters.

sigurdm (approved):
LGTM with a few questions

app/lib/package/model_properties.dart
isoos: Oh, that's an unrelated change I shall take it out from the PR. I was looking at TODOs in the code, and I picked it out first, as it seemed to be obsolete.

app/lib/scorecard/backend.dart
isoos: Yeah, this could also be problematic, I will rewrite this to track and limit length.

app/lib/scorecard/backend.dart
isoos: +1 on the beginning + end split. Unicode charcodes could be an issue, but IIRC it would not cause issues besides the character involved (which may be an issue we can ignore for now).

app/lib/scorecard/backend.dart
```diff
@@ -218,23 +219,68 @@ class ScoreCardBackend {
...
+            allDependencies: panaReport.allDependencies?.take(100).toList(),
+            licenseFile: panaReport.licenseFile,
+            report: pana.Report(
+              sections: panaReport.report?.sections
+                      .map(
+                        (s) => s.summary.length < 1000
+                            ? s
+                            : pana.ReportSection(
+                                id: s.id,
+                                title: s.title,
```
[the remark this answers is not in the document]
isoos: It's predefined in `pana`, but in theory they could also be too long. Limiting `id` and `title` with a reasonable length wouldn't hurt.

isoos:
PTAL.

Instead of trying to be smart about shortening the report, which makes the code really complex and full of edge cases (e.g. splitting the dependency table in half), I've chosen to create only a short version without reporting any of the original output.

The logs will contain the package and version info, and we should be able to recreate the issue if it arises.

PR #5155, opened 2021-10-04
Refactor SearchForm: explicit runtimes and platforms.

isoos:
For historical reasons we have `tagsPredicate` in `SearchForm`, the data holder of the user-facing search queries. This is not ideal, because it hides state that we need to act on while rendering the UI, and should use `TagsPredicate` only in the "compressed" form, when querying the search backend.

This is the first of the few PRs that will make the internal state explicit, eventually removing the `tagsPredicate` from `SearchForm`.


sigurdm (approved):
LGTM

PR #5156, opened 2021-10-04
Consider (approximate) singular/plural name as conflicting.

isoos:


sigurdm:
Code LGTM

I wonder about the error message you get when trying to publish. Will it tell what package-name you are too similar to?

sigurdm (approved):

isoos:
> I wonder about the error message you get when trying to publish. Will it tell what package-name you are too similar to?

The current message is: `Package name is too similar to another active or moderated package.`
I think that's good enough for now...

sigurdm:
cc @jonasfj might have opinions when he's back.

PR #5158, opened 2021-10-05
Upgrade dartdoc

isoos:
Looks like we are affected by the breaking API changes only a little bit.


isoos:
pushed to changing for further verification

sigurdm (approved):
LGTM

isoos:
new dartdoc on staging looks good

PR #5165, opened 2021-10-07
Atom feed return all published versions.

### LINKED ISSUE #5161: Feature request: Allow a custom limit for the Atom feed
### Issue
Currently pub.dev's atom feed is limited to the 10 most recent packages:

https://github.com/dart-lang/pub-dev/blob/e7faba418b1969c8f73d6a60f29a5ddbd510b170/app/lib/frontend/handlers/atom_feed.dart#L25

This leads to an increased number of times a reader has to poll the server as to not miss new packages.
Sometimes 5 minutes is all it takes for it to already be missing packages.

### Proposal
What I propose is adding a URL parameter to set the limit instead of hardcoding it.

* Create an optional `limit` parameter
* Not providing the parameter would default back to `10`
* The value could have a safe upper ceiling, say 100
* Setting `?limit=200` in this case should drop back to the defined upper limit, 100

isoos:
- feed returns the latest 100 package versions published
- a package may be present more than once (see changes in `atom_feed_test.dart`)
- fixes #5161

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -125,7 +125,7 @@ class PackageBackend {
 
   /// Retrieves package versions ordered by their published date descending.
   Future<List<PackageVersion>> latestPackageVersions(
-      {int offset = 0, int limit = 10}) async {
+      {int offset = 0, required int limit}) async {
     final query = db.query<PackageVersion>()
       ..order('-created')
       ..offset(offset)
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -123,11 +123,21 @@ class PackageBackend {
         .map((p) => p.name!);
   }
 
-  /// Retrieves package versions ordered by their latest version date.
+  /// Retrieves package versions ordered by their published date descending.
   Future<List<PackageVersion>> latestPackageVersions(
-      {int? offset, int? limit}) async {
-    final pkgPage = await latestPackages(offset: offset, limit: limit);
-    return lookupLatestVersions(pkgPage.packages);
+      {int offset = 0, int limit = 10}) async {
+    final query = db.query<PackageVersion>()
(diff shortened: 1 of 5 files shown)


app/lib/package/backend.dart
```diff
@@ -123,11 +123,21 @@ class PackageBackend {
         .map((p) => p.name!);
   }
 
-  /// Retrieves package versions ordered by their latest version date.
+  /// Retrieves package versions ordered by their published date descending.
   Future<List<PackageVersion>> latestPackageVersions(
-      {int? offset, int? limit}) async {
-    final pkgPage = await latestPackages(offset: offset, limit: limit);
-    return lookupLatestVersions(pkgPage.packages);
+      {int offset = 0, int limit = 10}) async {
```
sigurdm, line 128: I would make the limit required - I don't think there is any reasonable default.
```suggestion
      {int offset = 0, required int limit}) async {
```

app/lib/package/backend.dart
```diff
@@ -123,11 +123,21 @@ class PackageBackend {
         .map((p) => p.name!);
   }
 
-  /// Retrieves package versions ordered by their latest version date.
+  /// Retrieves package versions ordered by their published date descending.
   Future<List<PackageVersion>> latestPackageVersions(
-      {int? offset, int? limit}) async {
-    final pkgPage = await latestPackages(offset: offset, limit: limit);
-    return lookupLatestVersions(pkgPage.packages);
```
sigurdm, line 130: Do we use `lookupLatestVersions` anywhere else? Otherwise it might be dead code.

sigurdm (approved):
LGTM

I would clarify in the description that this does **not** allow for a custom limit as the title of the issue asks for. (Unless I am mistaken).

PR #5167, opened 2021-10-08
New search filters in the search service.

isoos:


app/lib/search/search_service.dart
```diff
@@ -283,6 +293,10 @@ class ServiceSearchQuery {
     final String? orderValue = uri.queryParameters['order'];
     final SearchOrder? order = parseSearchOrder(orderValue);
 
+    final minPoints =
```
sigurdm, line 296: Perhaps easier/more aligned to make this nullable like below, instead of having 0 be pseudo-null.

Issue #5189, opened 2021-10-18
Optimize /api/packages?page=<num>


isoos:
These requests are rather slow, for multiple reasons:
- cache timeout is 10 minutes, but requests are infrequent
- we are querying Datastore with offset + limit
- we are doing a second lookup to read all of the latest `PackageVersion` for these packages

We should consider:
- A: increase the cache timeout, allowing more inconsistencies between pages
- B: drop the latest version information from the response, making the content read part faster
- C: use an in-memory index (e.g. `NameTracker`) to list the packages in latest published order

PR #5190, opened 2021-10-18
Use cached package list from NameTracker instead of Datastore offsets.

isoos:
#5189

### DIFF
--- a/app/lib/package/name_tracker.dart
+++ b/app/lib/package/name_tracker.dart
@@ -20,6 +20,9 @@ class TrackedPackage {
   final String package;
   final String latestVersion;
   final DateTime lastPublished;
+
+  /// Derived from `Package.isVisible`, the flag indicates that the
+  /// package has not been withdrawn, and can be displayed.
   final bool isVisible;
 
   TrackedPackage({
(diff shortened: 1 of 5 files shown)


app/lib/package/name_tracker.dart
```diff
@@ -15,6 +15,39 @@ import 'models.dart';
 final _logger = Logger('pub.name_tracker');
 const _pollingInterval = Duration(minutes: 15);
 
+/// Minimal package information tracked by [NameTracker].
+class TrackedPackage {
+  final String package;
+  final String latestVersion;
+  final DateTime lastPublished;
+  final bool isVisible;
```
jonasfj, line 23: Can we add a documentation comment for this property, what is derived from?

app/lib/package/name_tracker.dart
```diff
@@ -52,6 +93,15 @@ class NameTracker {
     _names.remove(name);
   }
 
+  /// Returns the cached list of packages ordered by descending last published date.
+  /// Only the visible packages are present.
+  List<TrackedPackage> get packagesOrderedByLastPublishedDesc {
```
jonasfj, line 98:
```suggestion
  List<TrackedPackage> get visiblePackagesOrderedByLastPublished {
```

I'm not sure its worth store the sorted list in-memory, but I don't mind it... it's one more risk of cache issues, but probably not going to happen.

isoos: I think it is not a big overhead, and we don't need to sort the list every time it is accessed.

PR #5216, opened 2021-10-27
Package.versionCount: decrease on delete + integrity check.

isoos:


app/lib/admin/backend.dart
```diff
@@ -400,28 +400,29 @@ class AdminBackend {
...
       }
-      if (updatePackage) {
+      if (updateVersionReferences) {
         package!.lastVersionPublished = null;
         versions.where((v) => v.version != version).forEach((v) => package
             .updateVersion(v, dartSdkVersion: currentDartSdk.semanticVersion));
-        package.updated = DateTime.now().toUtc();
-        tx.insert(package);
       }
+      package!.versionCount = package.versionCount! - 1;
```
jonasfj, line 423: This doesn't seem like an idempotent operation... We can only decrease the version count if the same `tx` is deleting a `PackageVersion` entity.

If the request fails half way through and is tried, that might not be the case on the second try. Is that a fair analysis?

------

Unrelated should we printing in: https://github.com/dart-lang/pub-dev/blob/82861ec08e236f39f94bec0d71cf385a4b770479/app/lib/admin/backend.dart#L386

jonasfj (changes requested):

app/lib/admin/backend.dart
isoos: Good catch!

> We can only decrease the version count if the same tx is deleting a PackageVersion entity.

It was the same tx, but we should only decrease when we also known there is a delete. Fixed that in the commit + added tests for it.

app/test/admin/api_test.dart
```diff
@@ -270,6 +272,26 @@ void main() {
         moderatedPkg =
             await dbService.lookupOrNull<ModeratedPackage>(moderatedPkgKey);
         expect(moderatedPkg, isNull);
+
+        // calling remove second time must not affect updated or version count
```
jonasfj, line 276: Sweet!

jonasfj (approved):

PR #5221, opened 2021-10-28
Upgrade pana + preview Flutter SDK

isoos:
- copied `getRepositoryUrl` from `pana` as-is, we should extend it with fetching the default branch name dynamically
- removed the repository-linked license url from the schema.org metadata, using the license page instead

### DIFF
--- a/app/lib/shared/urls.dart
+++ b/app/lib/shared/urls.dart
@@ -19,7 +19,16 @@ const httpsApiDartDev = 'https://api.dart.dev/';
 const trustedUrlSchemes = <String>['http', 'https', 'mailto'];
 
 /// Extensions that are considered to be images.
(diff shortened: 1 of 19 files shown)


jonasfj (approved):

app/lib/shared/urls.dart
```diff
@@ -18,6 +18,16 @@ const httpsApiDartDev = 'https://api.dart.dev/';
 /// rejected and the URL mustn't be displayed.
 const trustedUrlSchemes = <String>['http', 'https', 'mailto'];
 
+/// Extensions that are considered to be images.
+final _imageExtensions = <String>{'.gif', '.jpg', '.jpeg', '.png'};
```
jonasfj, line 22: nit: webp, apng, avif, svg

jonasfj (approved):

PR #5225, opened 2021-10-29
Fix versions list order.

isoos:
- fixes the order on the package versions page
- fixes the order on the versions .json API endpoint


jonasfj (approved):
Curious, any reason not to change the ordering in the cached values instead?

I guess we would have to review all places it's used to see if it's okay. But did we check that other usages don't also expect a different ordering?

app/lib/frontend/handlers/package.dart
```diff
@@ -76,7 +79,10 @@ Future<shelf.Response> packageVersionsListHandler(
       final dartSdkVersion = await getDartSdkVersion();
       return renderPkgVersionsPage(
         data,
-        versions.versions,
+        // The cached version list contains the versions in semantically ascending order.
+        // By reversing that list, we will have them in semantically descending order,
+        // which is the preferred order for displaying on the versions page.
```
jonasfj, line 84: This should also be documented in the documentation comment for `packageBackend.listVersionsCached`, especially if it promises to always return the versions in said order.

isoos:
> Curious, any reason not to change the ordering in the cached values instead?

The cached value originally was for the `/api/packages/<pkg>` endpoint, and I didn't want to change that. Does the `pub` client have any assumption on the order of the versions?

> I guess we would have to review all places it's used to see if it's okay. But did we check that other usages don't also expect a different ordering?

Maybe we could use a different field names that returns it in specific orders, it would also help to address the comment about documentation. I'll update the PR.

app/lib/frontend/handlers/package.dart
isoos: It is documented on the `versions` field when we are accessing, I think that's a better place for it, but we can also repeat it at other places.

jonasfj (approved):

PR #5230, opened 2021-11-03
Additional log tags for search index efficiency tracking.

isoos:


jonasfj (approved):
LOL; we really should just do a client library for stackdriver metrics :D

Issue #5214, opened 2021-10-27
Core libraries look very "empty" in search results


mit-mit:
I suggest two changes:

* Add package descriptions: Just use the first paragraph of text from the core lib
* Add a 'Core library' tag (similar to the 'Null safety tag')

![Screen Shot 2021-10-27 at 10 36 15](https://user-images.githubusercontent.com/13644170/139030755-0ea40ce3-6339-4392-8f65-7d08abd3b680.png)

elbeicktalat:
Nice idea,
Of course they will be have no pub points or popularity indication.

PR #5231, opened 2021-11-03
Fetch/process SDK library descriptions for SDK search index.

isoos:
- #5214
- the `api.<sdk>.dev` dartdoc page is fetched, and the first paragraph is used for description

### DIFF
--- a/app/lib/search/dart_sdk_mem_index.dart
+++ b/app/lib/search/dart_sdk_mem_index.dart
@@ -5,7 +5,6 @@
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
-import 'package:retry/retry.dart';
 
 import '../shared/cached_value.dart';
 import 'backend.dart';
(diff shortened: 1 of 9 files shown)


app/lib/search/dart_sdk_mem_index.dart
```diff
@@ -61,13 +61,19 @@ Future<SdkMemIndex?> _createDartSdkMemIndex() async {
...
       () async {
         final index = SdkMemIndex.dart();
-        final uri = index.baseUri.resolve('index.json');
-        final rs = await http.get(uri);
-        if (rs.statusCode != 200) {
-          throw Exception('Unexpected status code for $uri: ${rs.statusCode}');
-        }
-        final content = DartdocIndex.parseJsonText(rs.body);
+        final content = DartdocIndex.parseJsonText(
+          await searchBackend.fetchSdkIndexContentAsString(
```
jonasfj, line 65: Using `httpRetryClient` inside `retry`...

app/lib/search/dart_sdk_mem_index.dart
```diff
@@ -61,13 +61,19 @@ Future<SdkMemIndex?> _createDartSdkMemIndex() async {
     return await retry(
       () async {
         final index = SdkMemIndex.dart();
-        final uri = index.baseUri.resolve('index.json');
-        final rs = await http.get(uri);
-        if (rs.statusCode != 200) {
-          throw Exception('Unexpected status code for $uri: ${rs.statusCode}');
-        }
-        final content = DartdocIndex.parseJsonText(rs.body);
+        final content = DartdocIndex.parseJsonText(
```
jonasfj, line 64: Should we perhaps do this as part of the build-step... and store the value in a `sdk-index.json` inside the Docker container, instead of creating it at runtime.

Similar to how we generate dart2js, sass -> css at build time. It's safer, faster and more reliable.

We could also hardcode a yaml file with dart / flutter SDK libraries/packages and descriptions. We'll have to manually update it, and we could make a test that compares to `api.dart.dev`.

I'm not sure it's wise to introduce this cross dependency, even if we supposedly fail gracefully it's probably best to be as disconnected as possible.

We could file this as an issue and solve it in a follow-up.

app/lib/search/dart_sdk_mem_index.dart
isoos: we are using `httpRetryClient` inside `fetchSdkIndexContentAsString`

app/lib/search/dart_sdk_mem_index.dart
isoos: I agree. There is a slight chance that the documentation is not available at the time of the deployment, so this should also have a fallback that will try to create the index after a day or so...

Will do in a follow-up...

jonasfj: I don't think we need a fallback... we could also generate the documentation into a json file when we bump `runtimeVersion`.

I'm arguing that we should consider not doing this at runtime ever, because it's fragile, it's better to have a deploy fail -- than to have failure at runtime.

app/lib/search/dart_sdk_mem_index.dart
jonasfj: Yes, I would generally discourage wrapping retries with more retries :D

Since the place we're calling `fetchSdkIndexContentAsString` (here) we are also inside a `retry` callback.

jonasfj (approved):

PR #5236, opened 2021-11-04
Patch pub integration

jonasfj:


pkg/pub_integration/test/pub_integration_test.dart
```diff
@@ -66,7 +66,6 @@ void main() {
         credentialsFileContent: fakeCredentialsFileContent(),
         invitedEmail: 'dev@example.org',
         inviteCompleterFn: inviteCompleterFn,
-        clientSdkDir: Platform.environment['PUB_INTEGRATION_CLIENT_SDK_DIR'],
```
isoos, line 69: This is being used in `pkg/pub_integration/tool/test-fake-with-local-dart-sdks.sh`, which I haven't really run in a long time, but it may be still useful.

jonasfj: What is it good for?

isoos (approved):

PR #5239, opened 2021-11-05
Add admin UI for package retraction

szakarias:


app/lib/frontend/templates/package_admin.dart
```diff
@@ -18,17 +18,20 @@ String renderPkgAdminPage(
...
       contentNode: packageAdminPageNode(
-        package: data.package!,
-        userPublishers: userPublishers,
-        uploaderUsers: uploaderUsers,
-      ),
+          package: data.package!,
+          userPublishers: userPublishers,
+          uploaderUsers: uploaderUsers,
+          retractableVersions: retractableVersions,
+          retractedVersions: retractedVersions),
```
jonasfj, line 34: nit: trailing comma makes for a nicer diff :D
```suggestion
          retractedVersions: retractedVersions,),
```

app/lib/frontend/templates/views/pkg/admin_page.dart
```diff
@@ -155,5 +159,64 @@ d.Node packageAdminPageNode({
...
+    if (requestContext.showPackageRetraction) ...[
+      d.h2(text: 'Package Version Retraction'),
+      d.div(children: [
+        d.markdown(
+            'You can retract a package version up to 7 days after publication.'),
+        d.markdown(
+            'This will not remove the package version, but warn developers using'
+            ' it and stop new applications from taking dependency on it without a dependency override.'),
+        d.markdown(
+            'You can restore a retracted package version if the version was retracted within the last 7 days.'),
```
jonasfj, line 171: Can't this be merged to a single markdown block...

with line breaks or something.. Just seems like it would be easier to write as:
```dart
      d.markdown('''
....

....
      ''')
```

Or
```
      d.markdown([
         '...',
         '',
         '....',
      ].join('\n'))
```

pkg/web_app/lib/src/account.dart
```diff
@@ -323,6 +340,37 @@ class _PkgAdminWidget {
...
+          'Are you sure you want to restore package version `$version`?'),
+      fn: () async {
+        print('before setVersionOption');
+        await client.setVersionOptions(pageData.pkgData!.package, version,
+            VersionOptions(isRetracted: false));
+      },
+      successMessage: text('Restoring complete. The page will reload.'),
+      onSuccess: (_) {
+        window.location.reload();
+      },
```
jonasfj, line 370: nit
```suggestion
      onSuccess: (_) => window.location.reload(),
```

pkg/web_css/lib/src/_form.scss
```diff
@@ -80,6 +80,6 @@
   }
 }
 
-#-admin-set-publisher-input {
+#-admin-set-publisher-input, #-admin-retract-package-version-input, #-admin-restore-retract-package-version-input{
```
jonasfj, line 83: Break lines... or maybe make a class called `-admin-dropdown` since we are using this on multiple dropdowns now..

jonasfj (approved):
LGTM

app/lib/frontend/templates/package_admin.dart
szakarias: done

pkg/web_app/lib/src/account.dart
szakarias: done

app/lib/frontend/templates/views/pkg/admin_page.dart
szakarias: The bottom approach does not seem to work, the top one does. However I think it is less readable, so unless you feel strongly about it, I'll leave it as is.

pkg/web_css/lib/src/_form.scss
szakarias: done

isoos (approved):

PR #5242, opened 2021-11-08
Better nullability for rpc helper method.

isoos:


pkg/web_app/lib/src/_dom_helper.dart
```diff
@@ -19,21 +19,21 @@ Future<R?> rpc<R>({
...
-  Future<R> Function()? fn,
+  Future<R?> Function()? fn,
 
   /// Message to show when the RPC returns without exceptions.
   required Element successMessage,
 
   /// Callback that will be called with the value of the RPC call, when it was
   /// successful.
-  FutureOr Function(R value)? onSuccess,
+  FutureOr Function(R? value)? onSuccess,
```
jonasfj, line 29: Why is this dynamic? Why not `FutureOr<void>` if there is no value...

pkg/web_app/lib/src/_dom_helper.dart
```diff
@@ -19,21 +19,21 @@ Future<R?> rpc<R>({
...
 
   /// Callback that will be called with the error object, when executing
   /// `fn` was not successful. The return value of this callback will be used
   /// to return from the method.
   ///
   /// If not specified, the error will be thrown instead.
-  FutureOr<R> Function(dynamic error)? onError,
+  FutureOr<R?> Function(dynamic error)? onError,
```
jonasfj, line 36:
```suggestion
  FutureOr<R?> Function(Object error)? onError,
```

pkg/web_app/lib/src/account.dart
```diff
@@ -231,7 +231,7 @@ class _PkgAdminWidget {
       return false;
     }
 
-    await rpc(
+    await rpc<void>(
```
jonasfj, line 234: This won't be inferred?

jonasfj (commented):
Just some drive by comments, feel free to ignore me :D

pkg/web_app/lib/src/account.dart
isoos: Not sure why, but the analyzer gave a hard error if we leave `<void>` out:
```
Couldn't infer type parameter 'R'.

Tried to infer 'Null' for 'R' which doesn't work:
  Parameter 'onSuccess' declared as     'FutureOr<void> Function(R?)?'
                        but argument is 'void Function(Object?)'.
The type 'Null' was inferred from:
  Parameter 'fn' declared as     'Future<R?> Function()?'
                 but argument is 'Future<Null> Function()'.

Consider passing explicit type argument(s) to the generic.
```

szakarias (approved):

PR #5249, opened 2021-11-08
Cleanup pkg/pub_integration CI and fix lints.

isoos:


isoos:
ah, forgot about the `test_data` directory, dropping the PR as it is not feasible at the moment.

PR #5256, opened 2021-11-10
Add retracted tag in package view

szakarias:


szakarias:
image

app/lib/tool/test_profile/models.dart
```diff
@@ -59,6 +59,7 @@ class TestPackage {
   final String? replacedBy;
   final bool? isUnlisted;
   final bool? isFlutterFavorite;
+  final List<String>? retractedVersions;
```
isoos, line 62: I think this is a simple and good enough solution for now, but I'd imagine if/when we'll have more flags for specific versions, it should go inside the `versions:` block.

app/lib/tool/test_profile/importer.dart
```diff
@@ -112,6 +112,13 @@ Future<void> importProfile({
               replacedBy: testPackage.replacedBy,
               isUnlisted: testPackage.isUnlisted,
             ));
+
+        if (testPackage.retractedVersions != null) {
+          testPackage.retractedVersions!.forEach((version) async {
```
isoos, line 117: Let's do a for loop over the versions, otherwise the async blocks may not complete until they are used e.g. in tests:

```
for (final version in testPackage.retractedVersions!) {
  await client.setVersionOptions(packageName, version, VersionOptions(isRetracted: true));
}
```

isoos (approved):
Small nit to change, otherwise good to go!

app/lib/tool/test_profile/importer.dart
szakarias: Good point, done.

PR #5257, opened 2021-11-10
Upgrade material web components to 7.0.0.

isoos:
- upgrading to 7.0.0 to enable tri-state checkboxes (indeterminate state)
- migrated dropdown template + access method, as it had a breaking change
- did not upgrade further to current versions, as manual tests revealed some further issues (TBD later)


isoos:
Note: deployed to staging, looks mostly okay to me.

jonasfj (approved):

PR #5258, opened 2021-11-11
Add lint package to devDependency overrides

passsy:
[`lint`](https://github.com/passsy/dart-lint/) - like `lints` - is only used as dev dependency and should displayed like that on pub.dev


jonasfj (approved):
Hmm, we probably should develop some pubspec field for stating this :D

But this is great for now :D

PR #5265, opened 2021-11-15
Update latest version references when retracting/restoring

szakarias:


app/lib/package/models.dart
```diff
@@ -235,12 +236,46 @@ class Package extends db.ExpandoModel<String> {
...
+  }
+
   /// Updates latest stable, prerelease and preview versions and published
   /// timestamp with the new version.
   void updateVersion(
     PackageVersion pv, {
     required Version dartSdkVersion,
+    bool includeRetracted = false,
   }) {
+    if (!includeRetracted && pv.isRetracted) return;
```
isoos, line 278: I'm wondering if this is needed here, or maybe at line 252? (see above)

app/lib/package/models.dart
```diff
@@ -235,12 +236,46 @@ class Package extends db.ExpandoModel<String> {
...
+    final oldStableVersion = latestSemanticVersion;
+    final oldPrereleaseVersion = latestPrereleaseSemanticVersion;
+    final oldPreviewVersion = latestPreviewSemanticVersion;
+
+    // reset field values
+    latestVersionKey = null;
+    latestPrereleaseVersionKey = null;
+    latestPreviewVersionKey = null;
+
+    versions.forEach((v) => updateVersion(v,
```
isoos, line 252: `versions.where((pv) => !pv.isRetracted).forEach()` ?

app/test/package/retracted_test.dart
```diff
@@ -164,5 +165,120 @@ void main() {
...
+      final origLatestPrereleasePublished = pkg.latestPrereleasePublished;
+
+      final client = createPubApiClient(authToken: adminAtPubDevAuthToken);
+
+      final orig = await client.getVersionOptions('oxygen', '1.2.0');
+      expect(orig.isRetracted, isFalse);
+      final origInfo = await client.packageVersionInfo('oxygen', '1.2.0');
+      expect(origInfo.isRetracted, isNull);
+
+// Retract latest
```
isoos, line 196: Nit: these comments are oddly indented.

isoos (approved):
Looks good, a few nits to change.

app/lib/package/models.dart
szakarias: Sure I can move that logic one level up, it's much nicer.

app/test/package/retracted_test.dart
szakarias: Done

Issue #5261, opened 2021-11-15
Explore non-blocking fonts


sigurdm:
Could make page-loading slightly smoother

jonasfj:
https://web.dev/font-display/

sigurdm:
This might still be relevant.

Need to check

PR #5266, opened 2021-11-15
font-display: swap for google fonts (for dartdoc pages)

isoos:
#5261

### DIFF
--- a/app/lib/dartdoc/customization.dart
+++ b/app/lib/dartdoc/customization.dart
@@ -208,7 +208,13 @@ class DartdocCustomizer {
       if (href != null &&
           href.startsWith('https://fonts.googleapis.com/') &&
           !href.contains('display=swap')) {
-        e.attributes['href'] = '$href&display=swap';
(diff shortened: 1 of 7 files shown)


jonasfj (approved):

app/lib/dartdoc/customization.dart
```diff
@@ -75,6 +75,7 @@ class DartdocCustomizer {
     _addAlternateUrl(head, canonical);
     _addAnalyticsTracker(head, body);
     _addGithubMarkdownStyle(head, body);
+    _replaceFontUrls(head);
```
jonasfj, line 78: Why is this needed?
https://github.com/dart-lang/dartdoc/blob/master/lib/templates/html/_head.html#L25

Is it just we want it for material icons too?

isoos:
Closing, as it is not recommended for icons to use the font-display: swap setting.
https://www.zachleat.com/web/font-display-icon-fonts/

PR #5268, opened 2021-11-15
Upgrade material web components to 13.0

isoos:
Looks like that the modal window issue I've experienced was maybe an issue with the locally cached version. After clearing the caches, I couldn't reproduce the issue that was blocking this upgrade previously.


isoos:
Or alternatively, it may be that the fixed template layout for the dropdowns also helped on the admin pages.
Anyway, pushed to staging for further testing.

jonasfj (approved):

PR #5273, opened 2021-11-16
Add input validation when retracting/restoring versions in admin UI

szakarias:


pkg/web_app/lib/src/account.dart
```diff
@@ -340,6 +340,10 @@ class _PkgAdminWidget {
   Future<void> _setRetracted() async {
     final version =
         materialDropdownSelected(_retractPackageVersionInput)?.trim() ?? '';
+    if (version.isEmpty) {
+      await _validateVersionSelection();
+    }
```
isoos, line 345: shouldn't this `return;`?

pkg/web_app/lib/src/account.dart
```diff
@@ -356,6 +360,10 @@ class _PkgAdminWidget {
     final version =
         materialDropdownSelected(_restoreRetractPackageVersionInput)?.trim() ??
             '';
+    if (version.isEmpty) {
+      await _validateVersionSelection();
+    }
```
isoos, line 365: same here...

pkg/web_app/lib/src/account.dart
szakarias: Yes of course. I was too quick when extracting the helper method!

isoos (approved):

PR #5276, opened 2021-11-16
Search section open state also considers referer URL.

isoos:


jonasfj (commented):
Won't this stop working once you click a new checkbox the second time?

isoos:
> Won't this stop working once you click a new checkbox the second time?

As long as you keep clicking inside the same section, it remains open, but switching sections will leave an now-inactive previously clicked section closed.

jonasfj (approved):

isoos:
Closing this as the client-side update prototype seems to be the way forward.

PR #5277, opened 2021-11-17
Link to retracted documentation

szakarias:


jonasfj (approved):

app/lib/frontend/templates/views/pkg/admin_page.dart
```diff
@@ -165,7 +165,7 @@ d.Node packageAdminPageNode({
       d.h2(text: 'Package Version Retraction'),
       d.div(children: [
         d.markdown(
-            'You can retract a package version up to 7 days after publication.'),
+            'You can [retract](https://dart.dev/go/pubretract) a package version up to 7 days after publication.'),
```
sigurdm, line 168: Perhaps we should also update doc/help-publishing.md ...

PR #5286, opened 2021-11-22
Strict version checks.

isoos:


pkg/pub_package_reader/test/integration_test.dart
```diff
@@ -66,5 +66,36 @@ dev_dependencies:
         )),
       );
     });
+
+    test('malformatted version', () async {
```
jonasfj, line 70: Should we put something like this into our test data in the future?

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -310,6 +311,44 @@ Iterable<ArchiveIssue> checkAuthors(String pubspecContent) sync* {
...
+Iterable<ArchiveIssue> checkStrictVersions(Pubspec pubspec) sync* {
+  Iterable<Version?> expandConstraint(VersionConstraint? constraint) {
+    if (constraint is VersionRange) {
+      return [constraint.min, constraint.max];
+    }
+    return const Iterable.empty();
+  }
+
+  Iterable<Version?> expandDependency(Dependency? dependency) {
+    if (dependency is HostedDependency) {
```
jonasfj, line 331: All dependency types can have `version` constraints... it's just a bug in `pubspec_parse` that it can't understand this...

This is a probably a temporary fix anyways, once we upgrade `pub_semver` we won't need this hack anymore.

(filed https://github.com/dart-lang/pubspec_parse/issues/77)

jonasfj (approved):

pkg/pub_package_reader/lib/pub_package_reader.dart
isoos: Added a TODO for it.

pkg/pub_package_reader/test/integration_test.dart
isoos: Any `ArchiveIssue` that we emit here is an upload-blocker, and we are testing this via full filesystem-based parsing. Nevertheless, we should make an integration test that tries to upload _dummy_pkg with the wrong internals.

isoos: Although... I guess at one point `pub` client itself will block the upload, so the usefulness of that test is probably only for a short time-window.

PR #5296, opened 2021-11-26
Correct published timestamps in Package entities.

isoos:


app/lib/tool/backfill/backfill_new_fields.dart
```diff
@@ -24,20 +24,79 @@ Future<void> _backfillPackages() async {
...
+        wasRace = true;
+      }
+    }
+
+    await for (final p in dbService.query<Package>().run()) {
+      if (await _packagePublishedTimestamps(p)) {
+        wasRace = true;
+      }
     }
   }
```
jonasfj, line 42: Use package retry throw an exception instead... and if the number of retries is exhausted then we handle that exception print an error and move on.

We should not sprinkle infinite loops around the code. It's better to try a number of times with exponential backoff and otherwise give up.

jonasfj (changes requested):

jonasfj (approved):

PR #5301, opened 2021-11-29
Refresh tokens in post deployment tests

jonasfj:
Newer versions of `dart pub` client will refuse to use `credentials.json` when `PUB_HOSTED_URL` is specified... so we have to refresh the access_token manually and use `dart pub token add $PUB_HOSTED_URL`.

### DIFF
--- a/pkg/pub_integration/lib/src/pub_tool_client.dart
+++ b/pkg/pub_integration/lib/src/pub_tool_client.dart
@@ -39,15 +39,17 @@ class DartToolClient {
     await Directory(tool._pubCacheDir).create(recursive: true);
 
     var creds = oauth2.Credentials.fromJson(credentialsFileContent);
-    final c = http.Client();
-    try {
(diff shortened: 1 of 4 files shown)


pkg/pub_integration/lib/src/pub_tool_client.dart
```diff
@@ -30,6 +38,18 @@ class DartToolClient {
     final tool = DartToolClient._(pubHostedUrl, tempDir.path);
     await Directory(tool._pubCacheDir).create(recursive: true);
 
+    var creds = oauth2.Credentials.fromJson(credentialsFileContent);
```
isoos, line 41: What do we do with this variable?

isoos (approved):

pkg/pub_integration/lib/src/pub_tool_client.dart
isoos: ++

PR #5303, opened 2021-11-30
Browser-based integration test for client-side search updates (Platforms + SDKs).

isoos:
- built on top of #5295
- doesn't handle the `null` issue with the 'Advanced' section yet.
- adds `/fake-test-profile` endpoint to the fake server, in order to make a large and diverse setup easier


pkg/pub_integration/test/search_update_test.dart
```diff
@@ -0,0 +1,188 @@
...
+              'packages': List.generate(
+                100,
+                (i) => {
+                  'name': i % 3 == 0 ? 'flutter_pkg_$i' : 'pkg_$i',
+                  if (i % 11 == 0) 'publisher': 'example.com',
+                },
+              ),
+            },
+          }));
+      await httpClient.get(Uri.parse('$origin/fake-update-search'));
```
jonasfj, line 47: Shouldn't we make some helper methods in `pub_puppeteer_helpers.dart` for calling things like: `/fake-update-search`

Ideally, such methods could have a documentation comment too, so that we can know what they do.

Example:
```dart

// Why not extend the PubApiClient generated by our code-gen?
class FakePubSiteClient extends PubApiClient {
  final http.Client _client;
  final String _baseUrl;

  FakePubSiteClient(String baseUrl, http.Client client) : _client = client, _baseUrl = baseUrl, super(baseUrl, client);

  /// Trigger update of search...... <what does actually mean>
  Future<void> fakeUpdateSearch() async {...}

  /// Load [profile] into the fake pub site.
  Future<void> loadTestProfile(Object profile) async {....}
}

```

Just an idea....  it could probably be merged with `PubHttpClient`, in fact... why not just use `PubHttpClient.forceSearchUpdate()` in this case here?

Then later we can dream about making `PubHttpClient` a subclass of the auto-generated `PubApiClient`.

pkg/pub_integration/test/search_update_test.dart
```diff
@@ -0,0 +1,188 @@
...
+            'flutter_pkg_51',
+            'pkg_13',
+            'pkg_62',
+            'pkg_34',
+            'pkg_28',
+            'pkg_94',
+            'flutter_pkg_66',
+            'flutter_pkg_0',
+            'pkg_73',
+          ]);
```
jonasfj, line 73: Why is this the case?

pkg/pub_integration/test/search_update_test.dart
```diff
@@ -0,0 +1,188 @@
...
+
+          // click Android
+          await page.click('#search-form-checkbox-platform-android');
+          await page.waitForNavigation(wait: Until.networkIdle);
+          final i2 = await listingPageInfo(page);
+          expect(i2.totalCount, 75);
```
jonasfj, line 79: Why does this happen? Why are the packages getting filtered, I don't see anything in the test profile definition...

jonasfj (approved):

pkg/pub_integration/test/search_update_test.dart
isoos: Maybe a better approach would be to do fake-update-search internally, as part of the test profile load. (updated)

pkg/pub_integration/test/search_update_test.dart
isoos: The test profile import uses a fake analysis by default, which assigns tags with a pseudorandom process (based on the hash of the package's name and sometimes version), with a few hardcoded patterns, e.g. `flutter_*` packages will get `sdk:flutter` tag assigned.
The test profile imported at the beginning creates 100 packages with semi-random tags, and adding and removing filters works because of the number of packages and their are kind of random.
(will add this as code comment)

PR #5309, opened 2021-12-01
Merge highlight.js + init via virtual StaticFile.

isoos:
- removes a very small request
- also moves highlighting to deferred execution


jonasfj (approved):

jonasfj:
We could also do a proper build script for `highlight.js`, see: https://highlightjs.readthedocs.io/en/latest/building-testing.html#building

And then tweak that build script to append our init.js... so we have the combination of these two files checked-in..
I would guess that's a tiny bit simpler, we have fewer things happening at runtime :D

isoos:
> I would guess that's a tiny bit simpler, we have fewer things happening at runtime :D

I think we - eventually - may have a build that creates a single CSS and a single JS file and we include only that. However, until then, virtual merger of these file is not that complex, and their reference checks is part of the templates tests. Note: code coverage metrics may be off once we do a blind merger of `pkg/web_app` and `pkg/web_css`...

PR #5318, opened 2021-12-03
Automatic lighthouse batch execution and total score.

isoos:


third_party/lighthouse/README.md
```diff
@@ -0,0 +1,31 @@
...
+dart run_suite.dart
+```
+
+Example output:
+
+```
+mobile:                                               avg [per, acc, bps, seo, pwa]
+
+/                                                      72 [ 83,  67,  92,  91,  30]
+/packages                                              71 [ 82,  78, 100,  75,  20]
```
jonasfj, line 17: May I suggest something like:
```
/packages, avg, 71
  , per, 82
  , acc, 78
  , bps, 100
  , seo, 75
  , pwa, 20
/packages/retry, avg, 73
....
```

So that we can copy/paste into a spreadsheet...

jonasfj (approved):

isoos:
As discussed, we won't merge this for now, as the dashboard has a good metric we can monitor.

PR #5329, opened 2021-12-07
Track deleted versions on the Package entity.

### LINKED ISSUE #5315: Deleted package-versions should be tombstoned
Otherwise they will be re-uploaded with different contents -> bad.

isoos:
Fixes #5315.

### DIFF
--- a/app/lib/admin/backend.dart
+++ b/app/lib/admin/backend.dart
@@ -379,7 +379,8 @@ class AdminBackend {
       final Key packageKey = _db.emptyKey.append(Package, id: packageName);
       final package = await tx.lookupOrNull<Package>(packageKey);
       if (package == null) {
-        print('Package $packageName does not exist.');
+        throw Exception(
+            'Package "$packageName" does not exists. Use full package removal without the version qualifier.');
       }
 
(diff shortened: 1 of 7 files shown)


app/lib/admin/backend.dart
```diff
@@ -379,15 +379,16 @@ class AdminBackend {
       final Key packageKey = _db.emptyKey.append(Package, id: packageName);
       final package = await tx.lookupOrNull<Package>(packageKey);
       if (package == null) {
-        print('Package $packageName does not exist.');
+        throw Exception(
+            'Package "$packageName" does not exists. Use full package removal without the version qualifier.');
```
jonasfj, line 383: What does using full package removal help?

jonasfj (approved):
We probably need admin API so we can remove version numbers from `Package.deletedVersions`, but it doesn't have to be in this PR.

app/lib/admin/backend.dart
isoos: It will run regardless of the `Package` entity existing. Removing single version assumes an existing `Package` entity, as it needs to modify entries on it.

PR #5333, opened 2021-12-08
Reduce multiple archive scans in package reader's content extraction.

isoos:
- Reduces IO time while publishing a new package version (and hopefully upload API request latencies).
- The IO could be further optimized with different tradeoffs (to be decided):
  - we scan the archive once and keep the file contents in memory,
  - we try to load as much of these as we can at the initial scan of the file, but that makes the extraction logic more complex.

### DIFF
--- a/pkg/pub_package_reader/lib/src/tar_utils.dart
+++ b/pkg/pub_package_reader/lib/src/tar_utils.dart
@@ -4,6 +4,7 @@
 
 import 'dart:convert';
 import 'dart:io';
+import 'dart:typed_data';
 
 import 'package:path/path.dart' as p;
 import 'package:tar/tar.dart';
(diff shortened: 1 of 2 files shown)


pkg/pub_package_reader/lib/src/tar_utils.dart
```diff
@@ -37,38 +38,63 @@ class TarArchive {
...
         );
 
-  /// Reads file content as String.
-  Future<String> readContentAsString(String name, {int maxLength = 0}) async {
-    final mappedName = _normalizedNames[name] ?? name;
+  /// Scans the tar archive and reads the content of the files identified by [names].
+  ///
+  /// For each file, if [maxLength] is reached, the content is returned up-to the
+  /// current byte buffer.
+  Future<Map<String, Uint8List>> scanAndReadFiles(
```
jonasfj, line 45: What about creating a visitor pattern... like:
```dart
/// Create a [PackageCheck] for a given [Pubspec].
typedef PackageCheckProvider = PackageCheck Function(Pubspec pubspec);

/// Abstraction for a check that can be executed on a package.
abstract class PackageCheck {
  /// Returns list of [ArchiveIssue] found in [path] for a given file.
  ///
  /// This will be called for all files before [checkMissingFiles] is called.
  Iterable<ArchiveIssue> checkFilePath(String path) => [];

  /// Returns `true` if this [PackageCheck] is interested in checking byte contents of the file at [path].
  ///
  /// Returning `true` will cause [checkFileByteContents] to be called with [path] and byte contents of the
  /// file. Invocations may arrive in any order.
  ///
  /// Checking files larger than 10 MiB will not work, and package will be rejected
  /// if [checkFile] returns `true` for a file larger than 10 MiB.
  bool shouldCheckFileBytes(String path) => false;

  /// Same as [shouldCheckFileBytes], but causes [checkFileStringContents] to be called instead.
  bool shouldCheckFileString(String path) => false;

  /// Returns list of [ArchiveIssue] found in file [contents] for file found at [path].
  Iterable<ArchiveIssue> checkFileByteContents(Sting path, Uint8List contents) => [];

  /// Returns list of [ArchiveIssue] found in file [contents] for file found at [path].
  ///
  /// The [contents] given may contain Unicode Replacement character `U+FFFD`
  /// in cases where the file contains invalid UTF-8.
  Iterable<ArchiveIssue> checkFileStringContents(Sting path, String contents) => [];

  /// Returns list of [ArchiveIssue] resulting from missing files. This is called after
  /// [checkFilePath] have been called for each file.
  Iterable<ArchiveIssue> checkMissingFiles();
}
```

Then each condition we want to enforce on a package can be implemented in a separate file as:
```dart

PackageCheck containsReadme(Pubspec pubspec) => _ContainsReadmePackageCheck(pubspec);

class _ContainsReadmePackageCheck extends PackageCheck {
  final Pubspec _pubspec;
  bool _seenReadme = false;

  _ContainsReadmePackageCheck(this._pubspec);

  @override
  Iterable<ArchiveIssue> checkFilePath(String path) {
    if (!_seenReadme && path == 'README.md') {
      _seenReadme = true;
    }
    return [];
  }

  @override
  Iterable<ArchiveIssue> checkMissingFiles() {
    if (!_seenReadme) {
       return [
          ArchiveIssue('missing README.md'),
       ];
    }
    return [];
  }
}
```

Also then we can test each `PackageCheck` dependently... and we can make a simple test framework... where we just write:
```dart

testPackageCheck(
  name: 'containsReadme finds missing README.md',
  check: containsReadme,
  files: {
    'pubspec.yaml': jsonEncode({
      'name': 'mypkg',
      'environment': {'sdk': '>= 2.12.0 <3.0.0'},
    }),
    'lib/mypkg.dart': 'void sayHello() => print("hello world");',
  },
  issues: allOf(
    isList,
    isNotEmpty,
    // Matches a list of strings where one of the strings contains: "missing README.md"
    anyElement(contains('missing README.md')),
  ),
);

testPackageCheck(
  name: 'containsReadme find README.md',
  check: containsReadme,
  files: {
    'pubspec.yaml': jsonEncode({
      'name': 'mypkg',
      'environment': {'sdk': '>= 2.12.0 <3.0.0'},
    }),
    'lib/mypkg.dart': 'void sayHello() => print("hello world");',
    'README.md': 'Hello world, this is mypkg!',
  },
  issues: isEmpty, // no issues expected here, at-least not from this checker! :D
);
```

jonasfj: Maybe some of this over-engineering.. But if we want a nice abstraction we could do something like this.

jonasfj (approved):

pkg/pub_package_reader/lib/src/tar_utils.dart
isoos: Added a tracking issue in #5336, I think this will be useful, but it may not be urgent.

PR #5335, opened 2021-12-08
Added package:indexed_blob

jonasfj:


pkg/indexed_blob/README.md
```diff
@@ -0,0 +1,15 @@
+# Indexed Blob
```
isoos, line 1: I think we should note here that in the blob each file is compressed individually, and the start and end positions are for their compressed bytes. Depending on how we do it, but later on we may need to extend this format later with the content length and possibly the media type for better http header support for these files.

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -0,0 +1,326 @@
...
+///       // File "lib/blob.dart" starts at 42 and ends at 80 in <blobId>.
+///       "blob.dart": "42:80"
+///     }
+///   }
+/// }
+/// ```
+///
+@sealed
+class BlobIndex {
+  final Uint8List _indexFile;
```
isoos, line 221: Note: I'm a bit worried that each path access is another partial parsing of the json, but maybe that's all we need. It is hard to tell how efficient this will be, let's go with it, and test it in practice. We have at least one alternative format that can be sharded and we could also try out if needed.

isoos (approved):
Overall, I think this looks good. Let's fix mono_repo and we should incorporate this in dartdoc processing.

Issue #5337, opened 2021-12-09
Package retraction should be logged and reflected in the "activity log"


sigurdm:
![image](https://user-images.githubusercontent.com/8613953/145367606-4d41b320-1323-4032-8a8d-15f5fdf1c874.png)

PR #5340, opened 2021-12-09
Log package retraction

szakarias:
Fixes https://github.com/dart-lang/pub-dev/issues/5337.

### DIFF
--- a/app/lib/audit/models.dart
+++ b/app/lib/audit/models.dart
@@ -153,6 +153,30 @@ class AuditLogRecord extends db.ExpandoModel<String> {
       ..publishers = [];
   }
 
+  factory AuditLogRecord.packageVersionOptionsUpdated({
+    required String package,
+    required String version,
+    required User user,
+    required List<String> options,
+  }) {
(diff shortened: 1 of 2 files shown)


app/lib/audit/models.dart
```diff
@@ -153,6 +153,30 @@ class AuditLogRecord extends db.ExpandoModel<String> {
...
+          'package `$package` version `$version`.'
+      ..data = {
+        'package': package,
+        'version': version,
+        'user': user.email,
+        'options': options,
+      }
+      ..users = [user.userId]
+      ..packages = [package]
+      ..packageVersions = [version]
```
isoos, line 176: The format of this field is unfortunately not documented, but it should be `'$package/$version'` as used in `packagePublished`. (The format is also used at different places e.g. `QualifiedVersionKey` class formalizes it).
Maybe add a note in the documentation of the field too?

isoos (changes requested):

isoos (approved):

PR #5362, opened 2021-12-15
Fix SDK URL in Dockerfile

jonasfj:


jonasfj:
We should probably fix the `tool/setup-dart.sh` script to just accept a version number as argument, maybe version number and channel to keep things simple.

Ideally, we would give it: version number, channel and sha256, as arguments.

isoos (approved):

PR #5369, opened 2021-12-16
Tool to upgrade the target Dart SDK in all mono_pkg.yaml

isoos:


jonasfj (approved):

jonasfj:
We could also try to make a PR for `mono_repo` and use `yaml_edit`.

PR #5377, opened 2021-12-22
Use pkg/indexed_blob for dartdoc content serving.

### LINKED ISSUE #5134: dartdoc content storage optimization: segmented .tar.gz compression
Another dartdoc content optimization (auxiliary to #4807) would be to create a `.tar.gz` file with separate `gzip` frames for each `TarEntry` bytestream segments. Instead of uploading the `.tar.gz` archive and its content separately, we build the archive with an external index (filename + offset + length), and store the index at a convenient place. (Sidenote: this could be similar to the CDXJ indexes of `.warc.gz` files, and stored in Datastore in batches of 100-1000 files, making it chunked, cacheable and still optimal for low-overhead lookups.)

The content serving won't change too much: we would still proxy dartdoc content through pub.dev, but instead of reading single files from the storage bucket, we would read a range of a larger file. In cases where the browser accepts the `gzip`-encoded content, we wouldn't need to uncompressed the range to serve the content.

However, the single-file approach is blocked on https://github.com/dart-lang/sdk/issues/47244, as the current default gzip decoding will stop at the first segment, and won't proceed to the next ones, making the segmented compression unable to read past the first file in the archive. In the absence of that, we could implement the solution with a separate file that is not served to the clients (compressed or not-compressed).

isoos:
- Closes #5134.
- Uses gzipped segments in the blob file.
- Some files are served outside of the blob (e.g. documentation archive, `log.txt`), and there is a fallback mechanism to also serve files from previously created entries.
- The storage bucket upload and also the cleanup process remains the same, but in effect it should be faster with fewer overall operations.
- The previous model stored and served the shared content (e.g. `shared-assets` directory) in a separate directory. For simplicity, this new model doesn't do this yet. We could later (a) implement it similarly, uploading those to a separate blob file, (b) serve those files through a shared cache.
- The blob file is unlikely to change in the later iterations, but the index file may change. To allow some reasonable upgrade path, it is now named as `index-v1.json`. We could later introduce a second index file and mark its existence in `DartdocEntry` - if needed.

### DIFF
--- a/app/lib/dartdoc/backend.dart
+++ b/app/lib/dartdoc/backend.dart
@@ -8,6 +8,7 @@ import 'dart:io';
 
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:gcloud/storage.dart';
+import 'package:indexed_blob/indexed_blob.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as p;
 import 'package:pool/pool.dart';
@@ -23,6 +24,17 @@ import '../shared/versions.dart' as shared_versions;
 
 import 'models.dart';
 import 'storage_path.dart' as storage_path;
+
+const archiveFilePath = 'package.tar.gz';
+const blobFilePath = 'blob-data.gz';
+const blobIndexV1FilePath = 'index-v1.json';
+const buildLogFilePath = 'log.txt';
+const uploadedFilePaths = [
+  archiveFilePath,
+  blobFilePath,
+  blobIndexV1FilePath,
+  buildLogFilePath,
+];
 
 final Logger _logger = Logger('pub.dartdoc.backend');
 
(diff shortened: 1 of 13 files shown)


app/lib/dartdoc/backend.dart
```diff
@@ -262,6 +274,25 @@ class DartdocBackend {
           () async => retry<FileInfo?>(
             () async {
               try {
+                if (uploadedFilePaths.contains(relativePath)) {
+                  final info = await _storage.info(objectName);
+                  return FileInfo(lastModified: info.updated, etag: info.etag);
+                }
```
jonasfj, line 280: It seems messy that we mix these two kinds of files in the same handler... Why do need this?

jonasfj (approved):

app/lib/dartdoc/backend.dart
```diff
@@ -24,6 +25,17 @@ import '../shared/versions.dart' as shared_versions;
...
 import 'storage_path.dart' as storage_path;
 
+const archiveFilePath = 'package.tar.gz';
+const blobFilePath = 'blob-data.gz';
+const blobIndexV1FilePath = 'index-v1.json';
+const buildLogFilePath = 'log.txt';
+const uploadedFilePaths = [
+  archiveFilePath,
+  blobFilePath,
+  blobIndexV1FilePath,
```
jonasfj, line 35: Why do these need to be exposed.

app/lib/dartdoc/backend.dart
```diff
@@ -24,6 +25,17 @@ import '../shared/versions.dart' as shared_versions;
 import 'models.dart';
 import 'storage_path.dart' as storage_path;
 
+const archiveFilePath = 'package.tar.gz';
+const blobFilePath = 'blob-data.gz';
+const blobIndexV1FilePath = 'index-v1.json';
+const buildLogFilePath = 'log.txt';
+const uploadedFilePaths = [
```
jonasfj, line 32: Does this `uploadedFilePaths` need to exported?

app/lib/dartdoc/models.dart
```diff
@@ -175,6 +175,12 @@ class DartdocEntry {
   /// The size of all the individual files, uncompressed.
   final int? totalSize;
 
+  /// The size of the compressed blob file.
+  final int? blobSize;
+
+  /// The size of the compressed blob index file.
+  final int? blobIndexSize;
```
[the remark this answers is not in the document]
isoos: Also, this data may be loaded from earlier runtime, which doesn't have this field.

app/lib/dartdoc/backend.dart
isoos: Yeah, the http handler doesn't try to serve these from the blob, rather it accesses them directly.

PR #5381, opened 2021-12-28
Fix inconsistent query text after search + back button.

isoos:
When navigating forward and back using both the old or the new search, we may end up with a page where the URL's query parameter is different from content inside the search form's input field. I've looked into other state handling methods, but it seems to me that neither of them work for such cases, however, we can use the `onPageShow` event to make sure the input field is consistent with the URL.


pkg/web_app/lib/src/search.dart
```diff
@@ -61,6 +61,15 @@ void _setEventForSearchInput() {
   });
 }
 
+void adjustQueryTextAfterPageShow() {
```
jonasfj, line 64: Can we add a comment explaining why this is necessary?

jonasfj (approved):

PR #5389, opened 2022-01-03
Remove emojis from description.

isoos:
- Recognizes only a limited set of emojis, but it may be a good-enough start for now. We can expand using a proper unicode-related library if this becomes a pressing issue.
- Tests to specifically allow CJK characters.


app/lib/tool/utils/emoji.dart
```diff
@@ -0,0 +1,82 @@
...
+  }
+  return String.fromCharCodes(newRunes);
+}
+
+/// Ranges were generated by the `_UnicodeEmojiSequences.getBasicRanges`
+/// method, with a merge threshold of 8. It collapses the ranges with a
+/// reasonable rate where we may include characters that are not necessarily
+/// emojies.
+///
+/// print(sequences.getBasicRanges(mergeThreshold: 8).map((e) => '$e\n').join());
```
sigurdm, line 29: How bad is mergeThreshold: 0 ?
I guess this is not a very performance-sensitive operation...

PR #5391, opened 2022-01-04
Use package:clock.

isoos:
`package:clock` will allow us to use arbitrary time in `test_profile` setup and also in tests that are time-sensitive.


kevmoo:
Would love a tiny bit of context in the commit message. 😄

isoos:
> Would love a tiny bit of context in the commit message. 😄

There isn't much to it: we have discussed it a while back that it is a good idea to have control over time in tests, and as we depend on it already, we may as well just use it directly.

jonasfj (approved):

isoos:
> Can we make a test that greps our source code for DateTime.now() ?

Yeah, it is in `app/test/shared/misc_source_files_test.dart`.

Issue #5368, opened 2021-12-16
GA analytics events for use of filters


jonasfj:
Let's add a `data-ga-click-event` attribute on search filter buttons, so that we can check if they are being used.
I think we are using client side updates with `History.pushState`, but it's unclear to me if GA will pick up those page changes.
But even if it does, it would be nice to know if the buttons are useful, and if the "advanced" section is frequently unfolded, etc.

![image](https://user-images.githubusercontent.com/149732/146345944-c4a5808d-1fc4-4275-b08d-4b105310381b.png)

cc @jayoung-lee

isoos:
It looks like we could pick these up via GTM triggers:
https://support.google.com/tagmanager/answer/7679322?hl=en

Would that work? I guess it is more reliable to have a single listener on these events than us trying to trigger it every time.

jayoung-lee:
Will the clicks on the filter items change the URL?

Actually, even if it does, I think [measuring the interaction events](https://developers.google.com/analytics/devguides/collection/analyticsjs/events) will provide more straightforward workflow & data for us. 

(If we use GTM, we need to test triggers every time the filter section is updated, and we also need to clean up the data to see what each before-after combination means.)

isoos:
> Will the clicks on the filter items change the URL?

Yes, changes the URL and also calls `pushState`.

Events would work too, any special category/action/label for them?

jayoung-lee:
> Event category is always 'click' for all the clicks.

It looks like we're just using event *action* to distinguish different events (Analytics 360 > Behavior > Events). Event *category* is always 'click' for all the clicks. 

What about using: 'filter-platform-[platform_name]', 'filter-sdk-[sdk_name]', etc?

isoos:
I think this should be active now. @jayoung-lee could you please confirm that this is working as expected?

jayoung-lee:
@isoos Yes I can see the `filter-*` events flowing in. Thank you!

PR #5396, opened 2022-01-05
Trigger a GTM event with in-page search filters.

isoos:
- #5368
- the action name is coming from the data attribute, so we have good control over it
- the GTM event has the same structure as we have in `gtm.js`
- separate events for going from unchecked->checked (`-on` postfix) vs. checked -> unchecked (`-off` postfix)

### DIFF
--- a/pkg/web_app/lib/src/gtm_js.dart
+++ b/pkg/web_app/lib/src/gtm_js.dart
@@ -17,14 +17,13 @@ void _pushMap(Map<String, dynamic> data) {
   _push(jsify(data));
 }
 
-/// Records a GTM [action] with the default [event] name and [category].
-void gtmCustomEventClick({
-  String event = 'custom-event',
-  String category = 'click',
+/// Records a GTM [action] with within a [category].
+void gtmCustomEvent({
+  required String category,
   required String action,
 }) {
   _pushMap({
-    'event': event,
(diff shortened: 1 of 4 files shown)


pkg/web_app/lib/src/gtm_js.dart
```diff
@@ -0,0 +1,33 @@
...
+external void _push(data);
+
+void _pushMap(Map<String, dynamic> data) {
+  _push(jsify(data));
+}
+
+/// Records a GTM [action] with the default [event] name and [category].
+void gtmCustomEventClick({
+  String event = 'custom-event',
+  String category = 'click',
```
jonasfj, line 23: Let's make category required, and have no default value.

pkg/web_app/lib/src/gtm_js.dart
```diff
@@ -0,0 +1,33 @@
...
+@JS('dataLayer.push')
+external void _push(data);
+
+void _pushMap(Map<String, dynamic> data) {
+  _push(jsify(data));
+}
+
+/// Records a GTM [action] with the default [event] name and [category].
+void gtmCustomEventClick({
+  String event = 'custom-event',
```
jonasfj, line 22: This should not be something that can be changed, we should just hardcode it.
It's also hardcoded in the trigger in google tag manager.

jonasfj (approved):

PR #5417, opened 2022-01-12
Escape ld+json content with manual JSON builder.

### LINKED ISSUE #5402: Error with encoding linked data for packages with double quotes in their descriptions
image

isoos:
- Fixes #5402.
- Using `json` encoder + second encoder doesn't work, as we want to double-encode some characters, but not the escape `\`. Detecting it correctly seems to need implementing a proper parser to do it right.
- Instead, this code emits the JSON structure `{`..`}`, `[`..`]`, `key: value` pairs directly, and also encodes the strings with unicode escaping for almost all of the characters, making sure that both the JSON and the HTML structures are escaped.

### DIFF
--- a/app/lib/frontend/dom/dom.dart
+++ b/app/lib/frontend/dom/dom.dart
@@ -136,7 +136,7 @@ Node codeSnippet({
 Node ldJson(Map<String, dynamic> content) {
   final sb = StringBuffer();
 
-  /// Build the JSON content by manually escaping most untrusted characters,
+  /// Build the JSON content by manually escaping dangerous characters,
   /// and also building the object and list structures.
   void write(dynamic value) {
     if (value is String) {
--- a/app/lib/frontend/dom/dom.dart
+++ b/app/lib/frontend/dom/dom.dart
@@ -169,7 +169,8 @@ Node ldJson(Map<String, dynamic> content) {
     } else if (value is bool || value is num || value == null) {
       sb.write(json.encode(value));
     } else {
-      throw ArgumentError('Unexpected value: $value');
+      throw ArgumentError(
+          'Value `$value` could not be translated to JSON, unexpected type: `${value.runtimeType}`.');
     }
   }
 
--- a/app/lib/frontend/dom/dom.dart
+++ b/app/lib/frontend/dom/dom.dart
@@ -16,8 +16,7 @@ final _elementRegExp = _attributeRegExp;
 // As we want to store rawJson inside a HTML Element, it is better
 // to escape all non-trusteded characters inside it. Non-trusted
 // characters must include `</!>` characters.
-final _ldJsonEscapedCharactersRegExp =
-    RegExp(r'[^0-9a-zA-Z ,@\:\{\}\[\]\.\-"]');
+final _ldJsonEscapedCharactersRegExp = RegExp(r'[^0-9a-zA-Z ,@\.\-]');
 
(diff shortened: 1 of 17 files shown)


app/lib/frontend/dom/dom.dart
```diff
@@ -135,17 +134,49 @@ Node codeSnippet({
...
+        if (i > 0) sb.write(',');
+        write(entries[i].key);
+        sb.write(':');
+        write(entries[i].value);
+      }
+      sb.write('}');
+    } else if (value is bool || value is num || value == null) {
+      sb.write(json.encode(value));
+    } else {
+      throw ArgumentError('Unexpected value: $value');
```
jonasfj, line 172: Better explain that the `content` passed to `ldJson` could not be translated to JSON, because it contains a value of type `value.runtimeType` (or something like that)

app/lib/frontend/dom/dom.dart
```diff
@@ -16,8 +16,7 @@ final _elementRegExp = _attributeRegExp;
 // As we want to store rawJson inside a HTML Element, it is better
 // to escape all non-trusteded characters inside it. Non-trusted
 // characters must include `</!>` characters.
-final _ldJsonEscapedCharactersRegExp =
-    RegExp(r'[^0-9a-zA-Z ,@\:\{\}\[\]\.\-"]');
+final _ldJsonEscapedCharactersRegExp = RegExp(r'[^0-9a-zA-Z ,@\.\-]');
```
jonasfj, line 19: Why this change?

I get that we need to escape `"`, but why any of the others?

jonasfj (approved):

app/lib/frontend/dom/dom.dart
jonasfj: Just thinking it's nice that URLs can be read even if escaping somehow fails...

app/test/frontend/golden/landing_page.html
```diff
@@ -221,6 +221,6 @@ <h1 class="home-block-title">Package of the Week</h1>
         <img class="inline-icon" src="/static/img/bug-report-white-96px.png?hash=mocked_hash_709947442" alt="bug report icon" width="20" height="20" title="Report an issue with this site"/>
       </a>
     </footer>
-    <script type="application/ld+json">{"@context":"http:\u002f\u002fschema.org","@type":"WebSite","url":"https:\u002f\u002fpub.dev\u002f","potentialAction":{"@type":"SearchAction","target":"https:\u002f\u002fpub.dev\u002fpackages\u003fq\u003d{search\u005fterm\u005fstring}","query-input":"required name\u003dsearch\u005fterm\u005fstring"}}</script>
+    <script type="application/ld+json">{"@context":"http\u003a\u002f\u002fschema.org","@type":"WebSite","url":"https\u003a\u002f\u002fpub.dev\u002f","potentialAction":{"@type":"SearchAction","target":"https\u003a\u002f\u002fpub.dev\u002fpackages\u003fq\u003d\u007bsearch\u005fterm\u005fstring\u007d","query-input":"required name\u003dsearch\u005fterm\u005fstring"}}</script>
```
jonasfj, line 224: @isoos are you sure we want to escape `:` in strings?

isoos: I don't mind escaping them. We won't have the `//` parts either, so URLs will be escaped, it is unlikely that they will be nice to read unparsed.

PR #5430, opened 2022-01-14
Block Zalgo in description on upload.

isoos:
The algorithm to detect the weird text is simple: when a text contains two subsequent diacritical mark, we consider it as zalgo-like and reject. We may fine-tune this later, if needed, but it seems to be a good sanity-check.

### DIFF
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -312,7 +312,16 @@ Iterable<ArchiveIssue> validatePackageVersion(Version? version) sync* {
 /// https://en.wikipedia.org/wiki/Zalgo_text
 Iterable<ArchiveIssue> validateZalgo(String field, String? text) sync* {
   if (text == null) return;
-  bool isDiacritic(int code) => code >= 0x0300 && code <= 0x036F;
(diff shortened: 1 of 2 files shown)


pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -307,6 +308,21 @@ Iterable<ArchiveIssue> validatePackageVersion(Version? version) sync* {
   }
 }
 
+/// Checks if the [text] has any weird characters like in
+/// https://en.wikipedia.org/wiki/Zalgo_text
+Iterable<ArchiveIssue> validateZalgo(String field, String? text) sync* {
+  if (text == null) return;
+  bool isDiacritic(int code) => code >= 0x0300 && code <= 0x036F;
```
sigurdm, line 315: What about "Combining Diacritical Marks Extended", "Combining Diacritical Marks , Supplement" , " for Symbols":

https://en.wikipedia.org/wiki/Combining_character (see "Unicode ranges")

isoos: Good point, extended the ranges.

sigurdm (commented):
LGTM

sigurdm (approved):

jonasfj (approved):

Issue #5334, opened 2021-12-08
Upgrade pub_semver


jonasfj:
This implies building a lenient-mode for existing packages with broken versions / version constraints.

We have two options:
 * (A) Hardcode list of package name / versions and manually patched pubspec to use.
 * (B) Translate broken format to canonical versions before reading `pubspec.yaml` from package versions published prior 2022.

PR #5434, opened 2022-01-17
Expose bad pubspec versions in integrity checks.

isoos:
- #5334
- This will create two log entry for each package version (one for `PackageVersion.pubspec`, another for `PackageVersionAsset[kind=pubspec].content`). We could group by these, if the amount of warnings is too high.
- The detection and reporting of the version parsing issue is independent of how we want to fix the versions themselves.


jonasfj (approved):
We are adding these warnings because want to fix it when loading, right?

isoos:
> We are adding these warnings because want to fix it when loading, right?

Yes, and also to detect it in case we have a future change that causes regression.

PR #5435, opened 2022-01-17
Verify publish_to field on upload.

isoos:


pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -308,6 +309,13 @@ Iterable<ArchiveIssue> validatePackageVersion(Version? version) sync* {
   }
 }
 
+/// Sanity check for matching `publish_to` field in `pubspec.yaml`.
+Iterable<ArchiveIssue> validatePublishTo(String? value) sync* {
+  if (value != null && value != 'pub.dartlang.org' && value != 'pub.dev') {
+    yield ArchiveIssue('Invalid `publish_to` value: `$value`.');
+  }
```
jonasfj, line 316: The value is supposed to be a URL, including https://

I propose that we forbid it from having any value at all.

Basically, we say that `publish_to` must not be present, or it must have a `null` value.

isoos: Good point on the URL format, and agreed, we can start by only accepting missing or `null` value. Updated the code + tests.

PR #5445, opened 2022-01-20
Backfilling from archive also updates PackageVersion.pubspec

isoos:
- Part of #5334, to make sure if we change the package reader, we'll get a consistent override/backfill.

### DIFF
--- a/app/lib/package/models.dart
+++ b/app/lib/package/models.dart
@@ -462,6 +462,7 @@ class PackageVersion extends db.ExpandoModel<String> {
     var changed = false;
     if (pubspecContentAsYaml != null) {
       final newPubspec = Pubspec.fromYaml(pubspecContentAsYaml);
+      // TODO: consider deep compare of the pubspec data
       if (pubspec!.jsonString != newPubspec.jsonString) {
         pubspec = newPubspec;
         changed = true;
--- a/app/lib/package/models.dart
+++ b/app/lib/package/models.dart
@@ -454,6 +454,22 @@ class PackageVersion extends db.ExpandoModel<String> {
     );
   }
 
+  /// Updates the current instance with the newly derived data.
+  /// Returns true if the current instance changed.
+  bool updateIfChanged({
+    required String? pubspecContentAsYaml,
+  }) {
+    var changed = false;
+    if (pubspecContentAsYaml != null) {
+      final newPubspec = Pubspec.fromYaml(pubspecContentAsYaml);
+      if (pubspec!.jsonString != newPubspec.jsonString) {
+        pubspec = newPubspec;
+        changed = true;
+      }
+    }
+    return changed;
+  }
+
   /// List of tags from the flags on the current [PackageVersion] entity.
   List<String> getTags() {
     return <String>[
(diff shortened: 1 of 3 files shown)


app/lib/package/models.dart
```diff
@@ -454,6 +454,22 @@ class PackageVersion extends db.ExpandoModel<String> {
...
 
+  /// Updates the current instance with the newly derived data.
+  /// Returns true if the current instance changed.
+  bool updateIfChanged({
+    required String? pubspecContentAsYaml,
+  }) {
+    var changed = false;
+    if (pubspecContentAsYaml != null) {
+      final newPubspec = Pubspec.fromYaml(pubspecContentAsYaml);
+      if (pubspec!.jsonString != newPubspec.jsonString) {
```
jonasfj, line 465: Shouldn't we use a different comparison? Like deep-equals after parsing from `jsonString`?

app/test/tool/backfill/backfill_packageversions_test.dart
```diff
@@ -97,5 +101,51 @@ void main() {
...
+            (await packageBackend.lookupPackageVersion('oxygen', '1.2.0'))!;
+        final asset = (await packageBackend.lookupPackageVersionAsset(
+          'oxygen',
+          '1.2.0',
+          AssetKind.pubspec,
+        ))!;
+        final originalPvContent = pv.pubspec!.jsonString;
+        final originalAssetContent = asset.textContent;
+
+        pv.pubspec = Pubspec.fromYaml('name: x');
```
jonasfj, line 118: Can we leave a comment with some explanation here, like:
```suggestion
        // Inject non-normalized version number in the pubspec asset.
        // This ensures that we can cleanup from:
        // https://github.com/dart-lang/pub_semver/pull/63
        pv.pubspec = Pubspec.fromYaml('name: x');
```

app/test/tool/backfill/backfill_packageversions_test.dart
```diff
@@ -97,5 +101,51 @@ void main() {
...
+        expect(
+          stat.toJson(),
+          {
+            'versionCount': 3,
+            'pvCount': 1,
+            'pvInfoCount': 0,
+            'pvAssetUpdatedCount': 1,
+            'pvAssetDeletedCount': 0
+          },
+        );
```
jonasfj, line 136: Isn't it a little fragile to test on this?

Shouldn't we only make assertions over `pvCount`, or maybe just not make an assertion here?

jonasfj (approved):

app/lib/package/models.dart
isoos: We could use deep comparison, but I don't think it will have much benefit over the current implementation.

The specific counts in the tests show us that if we feed the same `pubspec.yaml` file, we generate the same json string, and don't trigger an update with it. In practice the JSON string will keep the order of the keys.

We have similar comparison with JSON at least one more place in our codebase, so if we wanted to have deep copy, it could be used at multiple places, but I'd consider it a low-priority. I'll add a TODO for it.

jonasfj: A todo is fine :D

PR #5463, opened 2022-01-26
User entity integrity checks to load User before emitting error.

### LINKED ISSUE #5452: integrity problem is falsely reported when likes are made during the checks

isoos:
- Fixes #5452
- Applies the same verifications for all the `userId` references (except `PackageVersion.uploader`, which we retain indefinitely, and allow to reference a deleted `User`).
- When a `userId` is missing, we try to lookup it again.


sigurdm (approved):
LGTM

PR #5469, opened 2022-01-28
Remove try-catch around now-nullable traceId.

isoos:


isoos:
Looks like `context` is not null-aware yet, and this may not be safe to do.

PR #5471, opened 2022-01-28
Fix broken import.

isoos:
- #5460 was not based on the latest HEAD, and removed an import that was required by a parallel change.


sigurdm (approved):
LGTM

PR #5475, opened 2022-01-31
Enable new search UI.

isoos:
TBD: once the release is stable, we can remove the flag and update the templates tests.


jonasfj (approved):
LGTM, ping @mit-mit do we want to ship the new search UI this week?

PR #5494, opened 2022-02-03
Override invalid versions at package extraction + backfill.

isoos:
- #5334
- Uses `package:yaml_edit` to update the `pubspec.yaml`.
- Uses list of packages to narrow down the the `pubspec.yaml` content that may be updated.
- Uses created timestamp to exclude versions that were uploaded after the upload-time fix was deployed (even for the listed package names).
- Package names were queried through pub.dev APIs and checking their parsed pubspec content.
- Scheduled backfill.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -739,6 +739,7 @@ class PackageBackend {
         filename,
         maxContentLength: maxAssetContentLength,
         maxArchiveSize: UploadSignerService.maxUploadSize,
+        created: DateTime.now().toUtc(),
       );
       _logger.info('Package archive scanned in ${sw.elapsed}.');
       if (archive.hasIssues) {
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -739,7 +739,7 @@ class PackageBackend {
         filename,
         maxContentLength: maxAssetContentLength,
         maxArchiveSize: UploadSignerService.maxUploadSize,
-        created: DateTime.now().toUtc(),
+        created: clock.now().toUtc(),
       );
       _logger.info('Package archive scanned in ${sw.elapsed}.');
       if (archive.hasIssues) {
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -739,7 +739,7 @@ class PackageBackend {
         filename,
         maxContentLength: maxAssetContentLength,
         maxArchiveSize: UploadSignerService.maxUploadSize,
-        created: clock.now().toUtc(),
+        published: clock.now().toUtc(),
       );
       _logger.info('Package archive scanned in ${sw.elapsed}.');
       if (archive.hasIssues) {
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -15,6 +15,7 @@ import 'src/archive_surface.dart';
 import 'src/check_platforms.dart';
 import 'src/file_names.dart';
 import 'src/names.dart';
+import 'src/pubspec_content_override.dart';
 import 'src/tar_utils.dart';
 import 'src/yaml_utils.dart';
 
@@ -79,6 +80,12 @@ Future<PackageSummary> summarizePackageArchive(
   /// The maximum number of files in the archive.
   /// TODO: set this lower once we scan the existing archives
   int maxFileCount = 64 * 1024,
+
+  /// The timestamp when the archive was uploaded and accepted.
+  ///
+  /// If specified, some packages may get a transparent rewrite of their
+  /// assets to preserve backwards-compatible behavior on pub.dev.
+  DateTime? created,
 }) async {
   final issues = <ArchiveIssue>[];
 
@@ -119,7 +126,10 @@ Future<PackageSummary> summarizePackageArchive(
     return PackageSummary(issues: issues);
   }
 
-  final pubspecContent = await tar.readContentAsString(pubspecPath);
+  final pubspecContent = overridePubspecContentIfNeeded(
+    content: await tar.readContentAsString(pubspecPath),
+    created: created ?? DateTime.now().toUtc(),
+  );
   // Large pubspec content should be rejected, as either a storage limit will be
   // limiting it, or it will slow down queries and processing for very little
   // reason.
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -131,8 +131,8 @@ Future<PackageSummary> summarizePackageArchive(
     return PackageSummary(issues: issues);
   }
 
-  final pubspecContent = overridePubspecContentIfNeeded(
-    content: await tar.readContentAsString(pubspecPath),
+  final pubspecContent = overridePubspecYamlIfNeeded(
+    pubspecYaml: await tar.readContentAsString(pubspecPath),
     published: published ?? clock.now().toUtc(),
   );
   // Large pubspec content should be rejected, as either a storage limit will be
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -81,11 +81,15 @@ Future<PackageSummary> summarizePackageArchive(
   /// TODO: set this lower once we scan the existing archives
   int maxFileCount = 64 * 1024,
 
-  /// The timestamp when the archive was uploaded and accepted.
+  /// [DateTime] when the archive was uploaded / published.
   ///
-  /// If specified, some packages may get a transparent rewrite of their
-  /// assets to preserve backwards-compatible behavior on pub.dev.
-  DateTime? created,
+  /// Used to ensure that constraints in place when a package was published
+  /// are considered when summarizing assets. Ensuring we preserve
+  /// backwards-compatibility with packages that does not satisfy constraints
+  /// enforced at publish-time today.
+  ///
+  /// Defaults to [DateTime.now], if not specified.
+  DateTime? published,
 }) async {
   final issues = <ArchiveIssue>[];
 
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -4,6 +4,7 @@
 
 import 'dart:convert';
 
+import 'package:clock/clock.dart';
 import 'package:collection/collection.dart' show IterableExtension;
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as p;
@@ -132,7 +133,7 @@ Future<PackageSummary> summarizePackageArchive(
 
   final pubspecContent = overridePubspecContentIfNeeded(
     content: await tar.readContentAsString(pubspecPath),
-    created: created ?? DateTime.now().toUtc(),
+    published: published ?? clock.now().toUtc(),
   );
   // Large pubspec content should be rejected, as either a storage limit will be
   // limiting it, or it will slow down queries and processing for very little
--- a/pkg/pub_package_reader/lib/src/pubspec_content_override.dart
+++ b/pkg/pub_package_reader/lib/src/pubspec_content_override.dart
@@ -7,7 +7,7 @@ import 'package:pub_semver/pub_semver.dart';
 import 'package:yaml/yaml.dart' as yaml;
 import 'package:yaml_edit/yaml_edit.dart';
 
-final _logger = Logger('pubspec_content_override');
+final _logger = Logger('pubspec_yaml_override');
 
 /// The version regular expression that accepted any character as separator.
 final _lenientRegExp = RegExp(r'(\d+).(\d+).(\d+)' // Version number.
@@ -46,28 +46,28 @@ const _packagesWithBadVersions = <String>{
   'polymer_interop',
 };
 
-/// Override pubspec.yaml content if needed:
+/// Override pubspec.yaml if needed:
 ///
 /// - If the archive was created before 2022-01-01, we may need to update the
 ///   version number, version constraint and SDK constraints, as `pub_semver`
 ///   accepted separator characters other than `.`.
 ///   https://github.com/dart-lang/pub_semver/pull/63
-String overridePubspecContentIfNeeded({
-  required String content,
+String overridePubspecYamlIfNeeded({
+  required String pubspecYaml,
   required DateTime published,
 }) {
   // quick checks to skip overrides
   if (published.year >= 2022) {
-    return content;
+    return pubspecYaml;
   }
 
   try {
-    return _fixupBrokenVersionAndConstraints(content);
+    return _fixupBrokenVersionAndConstraints(pubspecYaml);
   } on FormatException catch (e, st) {
     _logger.info('Unable to parse pubspec.', e, st);
   }
-  // fallback: don't override content
-  return content;
+  // fallback: don't override
+  return pubspecYaml;
 }
 
 /// Fixes broken version and version constraints in [pubspecYaml] and returns a new YAML string.
--- a/pkg/pub_package_reader/lib/src/pubspec_content_override.dart
+++ b/pkg/pub_package_reader/lib/src/pubspec_content_override.dart
@@ -73,11 +73,11 @@ String overridePubspecYamlIfNeeded({
 /// Fixes broken version and version constraints in [pubspecYaml] and returns a new YAML string.
 ///
 /// Prior to [pub_semver#63] version numbers and version constraints were accepted with any
-/// character as separate, instead of requiring  `.` (dot). This occurred because the regular expression
-/// in `package:pub_semver` didn't escape dots as `\.`, but just contained `.` (dot, meaning any character).
+/// character as separator, instead of requiring  `.` (dot). This occurred because the regular expression
+/// in `package:pub_semver` didn't escape dots as `\.`, but just contained `.` (dot, meaning any character). (see https://github.com/dart-lang/pub_semver/commit/cee044a3dc8)
 ///
-/// Thus, package versions published prior to 2022-01-01, maybe contain _version numbers_,
-/// _version constraints_ and _SDK constraints_ that doesn't use `.` (dot) as separator. As removing
+/// Thus, package versions published prior to 2022-01-01, may contain _version numbers_,
+/// _version constraints_ and _SDK constraints_ that don't use `.` (dot) as separator. As removing
 /// these _package versions_ would break existing users, this function will rewrite the version numbers
 /// and version constraints to have a valid format equivalent to how older versions of `pub_semver`
 /// would have interpreted the version number / version constraints.
@@ -122,7 +122,8 @@ String _fixupBrokenVersionAndConstraints(String pubspecYaml) {
 
   final fixedPubspecYaml = editor.toString();
   if (hasBeenUpdated && fixedPubspecYaml == pubspecYaml) {
-    _logger.warning('Updating pubspec.yaml while fixing versions failed.');
+    _logger.warning(
+        'Updating pubspec.yaml in package:$name failed while fixing versions.');
   }
   return fixedPubspecYaml;
 }
--- a/pkg/pub_package_reader/lib/src/pubspec_content_override.dart
+++ b/pkg/pub_package_reader/lib/src/pubspec_content_override.dart
@@ -7,16 +7,9 @@ import 'package:pub_semver/pub_semver.dart';
 import 'package:yaml/yaml.dart' as yaml;
 import 'package:yaml_edit/yaml_edit.dart';
 
-final _logger = Logger('pubspec_yaml_override');
+import 'pub_semver_2.1.0_parser.dart';
 
-/// The version regular expression that accepted any character as separator.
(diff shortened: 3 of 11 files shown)


app/lib/package/backend.dart
```diff
@@ -739,6 +739,7 @@ class PackageBackend {
         filename,
         maxContentLength: maxAssetContentLength,
         maxArchiveSize: UploadSignerService.maxUploadSize,
+        created: DateTime.now().toUtc(),
```
jonasfj: didn't we migrate everything to `clock.now()` ?

hmm, funny a few lines up I see  `Stopwatch().` so I'm guessing not.

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,153 @@
...
+
+final _logger = Logger('pubspec_content_override');
+
+/// The version regular expression that accepted any character as separator.
+final _lenientRegExp = RegExp(r'(\d+).(\d+).(\d+)' // Version number.
+    r'(-([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?' // Pre-release.
+    r'(\+([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?' // Build.
+    );
+
+/// Packages with bad versions:
```
jonasfj, line 18: Just bad versions? Or also bad version constraints on dependencies, or bad version constraints in the SDK constraint?

Let's also document why these packages are hardcoded here, and why new packages are not going to be added to this list. Because we know check against this when uploading packages.

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,153 @@
...
+  'meta_types_redux',
+  'meta_types_redux_generator',
+  'mocktail',
+  'mocktail_image_network',
+  'polymer_interop',
+};
+
+/// Override pubspec.yaml content if needed:
+///
+/// - If the archive was created before 2022, we may need to update the versions,
```
jonasfj, line 50:
```suggestion
/// - If the archive was created before 2022-01-01, we may need to update the version number, version constraint and SDK constraints,
```

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,153 @@
...
+///   as pub_semver accepted separator characters other than `.`.
```
jonasfj, line 51: Let's add an example and link to https://github.com/dart-lang/pub_semver/pull/63

Guessing what we're doing here is going to be very hard in the future :D

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -79,6 +80,12 @@ Future<PackageSummary> summarizePackageArchive(
   /// The maximum number of files in the archive.
   /// TODO: set this lower once we scan the existing archives
   int maxFileCount = 64 * 1024,
+
+  /// The timestamp when the archive was uploaded and accepted.
+  ///
+  /// If specified, some packages may get a transparent rewrite of their
+  /// assets to preserve backwards-compatible behavior on pub.dev.
+  DateTime? created,
```
jonasfj, line 88: Why does this need to be optional? Can't we just patch all locations that call `summarizePackageArchive`?

Just a suggestion:
```suggestion
  /// [DateTime] when the archive was uploaded / published.
  ///
  /// Used to ensure that constraints in place when a package was published
  /// are considered when summarizing assets. Ensuring we preserve
  /// backwards-compatibility with packages that does not satisfy constraints
  /// enforced at publish-time today.
  ///
  /// Defaults to [DateTime.now], if not specified.
  DateTime? published,
```

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -119,7 +126,10 @@ Future<PackageSummary> summarizePackageArchive(
     return PackageSummary(issues: issues);
   }
 
-  final pubspecContent = await tar.readContentAsString(pubspecPath);
+  final pubspecContent = overridePubspecContentIfNeeded(
+    content: await tar.readContentAsString(pubspecPath),
+    created: created ?? DateTime.now().toUtc(),
```
jonasfj, line 131:
```suggestion
    created: created ?? clock.now().toUtc(),
```

If we want it to be optional..

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,153 @@
...
+String overridePubspecContentIfNeeded({
+  required String content,
```
jonasfj, line 53:
```suggestion
  required String pubspecYaml,
```

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,153 @@
...
+  required DateTime created,
+}) {
+  // quick checks to skip overrides
+  if (created.year >= 2022) {
+    return content;
+  }
+  if (_packagesWithBadVersions.every((name) => !content.contains(name))) {
+    return content;
+  }
```
jonasfj, line 62: Why not skip this and leave it for `_tryOverrideContent` to parse YAML and compare the name?

Is this really a performance critical path?

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
jonasfj: Let's link to https://github.com/dart-lang/pub_semver/pull/63 and give an example.

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,153 @@
...
+    }
+    return c ?? content;
+  } on FormatException catch (e, st) {
+    _logger.info('Unable to parse pubspec.', e, st);
+  }
+  // fallback: don't override content
+  return content;
+}
+
+String? _tryOverrideContent(String content) {
```
jonasfj, line 77:
```suggestion
/// Fixes broken version and version constraints in [pubspecYaml] and returns a new YAML string.
///
/// Prior to [pub_semver#63] version numbers and version constraints were accepted with any
/// character as separate, instead of requiring  `.` (dot). This occurred because the regular expression
/// in `package:pub_semver` didn't escape dots as `\.`, but just contained `.` (dot, meaning any character).
///
/// Thus, package versions published prior to 2022-01-01, maybe contain _version numbers_,
/// _version constraints_ and _SDK constraints_ that doesn't use `.` (dot) as separator. As removing
/// these _package versions_ would break existing users, this function will rewrite the version numbers
/// and version constraints to have a valid format equivalent to how older versions of `pub_semver`
/// would have interpreted the version number / version constraints.
///
/// [pub_semver#63]: https://github.com/dart-lang/pub_semver/pull/63
String _fixupBrokenVersionAndConstraints(String pubspecYaml) {
```

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,153 @@
...
+  final root = yaml.loadYaml(content);
+  if (root is! Map) {
+    return null;
```
jonasfj, line 80:
```suggestion
    _log.warning('Unable to parse YAML for package, in _fixupBrokenVersionAndConstraints!');
    return pubspecYaml; // return original as no changes can be made
```

We agree, that this should never happen, right?

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  // sanity check on the name again, now with parsed yaml
+  final name = root['name'];
+  if (name is! String || !_packagesWithBadVersions.contains(name)) {
+    return null;
```
jonasfj, line 86:
```suggestion
    return pubspecYaml; // no changes to be made
```

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,153 @@
...
+      final pre = match[5];
+      final build = match[8];
+      return Version(major, minor, patch, pre: pre, build: build).toString();
+    });
+    if (updated != c.value) {
+      editor.update(c.path, updated);
+    }
+  }
+
+  return editor.toString();
```
jonasfj, line 104:
```suggestion
  final fixedPubspecYaml = editor.toString();
  assert(fixedPubspecYaml != pubspecYaml);
  if (fixedPubspecYaml != pubspecYaml) {
    _log.warning('.....')
  }
  return fixedPubspecYaml;
```

app/lib/package/backend.dart
isoos: fixed (also indicated via a test :) )

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
isoos: removed

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,170 @@
...
+  'meta_types_json',
+  'meta_types_json_generator',
+  'meta_types_redux',
+  'meta_types_redux_generator',
+  'mocktail',
+  'mocktail_image_network',
+  'polymer_interop',
+};
+
+/// Override pubspec.yaml content if needed:
```
jonasfj, line 49:
```suggestion
/// Override pubspec.yaml [content] if needed:
```

I would suggest renaming the parameter to `pubspecYaml` or documenting it, because `content` could be many things.

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,170 @@
...
+    final updated = c.value.replaceAllMapped(_lenientRegExp, (match) {
+      final major = int.parse(match[1]!);
+      final minor = int.parse(match[2]!);
+      final patch = int.parse(match[3]!);
+      final pre = match[5];
+      final build = match[8];
+      return Version(major, minor, patch, pre: pre, build: build).toString();
+    });
+    if (updated != c.value) {
+      editor.update(c.path, updated);
```
jonasfj, line 112: If we do a modification, why don't we take `updated` and make:
```dart
try {
  Version.parse(updated);
} on FormatException catch (e, st) {
  _log.shout('Failed to fix broken version in package:$name', e, st);
  return pubspecYaml;
}
```
And similar for `VersionConstraint`.

It would be useful to assert that we're not producing garbage... because that might be harder to debug somewhere down the line.

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,170 @@
...
+/// in `package:pub_semver` didn't escape dots as `\.`, but just contained `.` (dot, meaning any character).
+///
+/// Thus, package versions published prior to 2022-01-01, maybe contain _version numbers_,
+/// _version constraints_ and _SDK constraints_ that doesn't use `.` (dot) as separator. As removing
+/// these _package versions_ would break existing users, this function will rewrite the version numbers
+/// and version constraints to have a valid format equivalent to how older versions of `pub_semver`
+/// would have interpreted the version number / version constraints.
+///
+/// [pub_semver#63]: https://github.com/dart-lang/pub_semver/pull/63
+String _fixupBrokenVersionAndConstraints(String pubspecYaml) {
```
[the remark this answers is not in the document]
isoos: We do have the `/packages/<pkg>/pubspec` endpoint that displays the original format, and I think it would be nice if these overrides would be still nicely visible there.

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,170 @@
...
+  var hasBeenUpdated = false;
+  for (final c in _detectVersionEditCandidates(root)) {
+    final updated = c.value.replaceAllMapped(_lenientRegExp, (match) {
+      final major = int.parse(match[1]!);
+      final minor = int.parse(match[2]!);
+      final patch = int.parse(match[3]!);
+      final pre = match[5];
+      final build = match[8];
+      return Version(major, minor, patch, pre: pre, build: build).toString();
+    });
```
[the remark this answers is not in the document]
isoos: Done, PTAL. I think the split+join by space gives enough power, we don't need more than that.

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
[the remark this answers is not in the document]
sigurdm: Should that endpoint ideally not display the original pubspec.yaml (with the errors)?

That would entail storing two pubspecs for each package, maybe that is overkill for such a small exposure...

isoos:
On top of the Dart SDK constraint override, I've added a single-line yaml comment in the beginning of the file to indicate that this is not the original content of the archive. wdyt? Should we add a property too so that the JSON in the API also has such indication?

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,228 @@
...
+              final versionStr =
+                  Version(major, minor, patch, pre: pre, build: build)
+                      .toString();
+              return '$range$versionStr';
+            }))
+        // Joining the parts by the same separator character.
+        .join(' ');
+    if (updated != c.value) {
+      try {
+        VersionConstraint.parse(updated);
```
jonasfj, line 137: Some candidates are versions and some are version constraints.

Couldn't we add this as a boolean on `_VersionEditCandidate` so that we check against `Version.parse` for versions?

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,228 @@
...
+    // no update
+  }
+
+  final fixedPubspecYaml = editor.toString();
+  if (fixedPubspecYaml == pubspecYaml) {
+    _logger.warning(
+        'Updating pubspec.yaml in package:$name failed while fixing versions.');
+  }
+
+  return '# overrides applied by pub.dev\n$fixedPubspecYaml';
```
jonasfj, line 179:
```suggestion
  return '# Compatibility rewrites applied by pub.dev\n$fixedPubspecYaml';
```

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
```diff
@@ -0,0 +1,228 @@
...
+  final name = root['name'];
+  if (name is! String || !_packagesWithBadVersions.contains(name)) {
+    return pubspecYaml; // return original as no changes needed
+  }
+
+  final editor = YamlEditor(pubspecYaml);
+  var hasBeenUpdated = false;
+  for (final c in _detectVersionEditCandidates(root)) {
+    final updated = c.value
+        // Version constraints may have more than one version expression, separated by space.
```
jonasfj, line 116: Is whitespace separation required? That is what this comment says, and the code relies on it.

jonasfj (commented):
Looks good, other than few comments.

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
isoos: Since the applied regexp is between `^` and `$`, I think we need to split by whitespaces. Or am I missing something here that makes this not necessary?

jonasfj: But you assume there is whitespace, which there might not be in the general case... right?

jonasfj: Why don't we copy/paste the functions from `pub_semver` then we're sure we have the same logic.
Here they are, ported to not depend on anything from `pub_semver` other than they create instances of `Version` and `VersionConstraint`.

```dart
import 'package:pub_semver/pub_semver.dart';

/// Parses a comparison operator ("<", ">", "<=", or ">=") at the beginning of
/// a string.
final _startComparison = RegExp(r'^[<>]=?');

/// The "compatible with" operator.
const _compatibleWithChar = '^';

/// Regex that matches a version number at the beginning of a string.
final _startVersion = RegExp(r'^' // Start at beginning.
    r'(\d+)\.(\d+)\.(\d+)' // Version number.
    r'(-([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?' // Pre-release.
    r'(\+([0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*))?'); // Build.

/// Like [_startVersion] but matches the entire string.
final _completeVersion = RegExp('${_startVersion.pattern}\$');

/// Creates a new [Version] by parsing [text].
Version _parsePossiblyBrokenVersion(String text) {
  final match = _completeVersion.firstMatch(text);
  if (match == null) {
    throw FormatException('Could not parse "$text".');
  }

  try {
    var major = int.parse(match[1]!);
    var minor = int.parse(match[2]!);
    var patch = int.parse(match[3]!);

    var preRelease = match[5];
    var build = match[8];

    return Version(major, minor, patch, pre: preRelease, build: build);
  } on FormatException {
    throw FormatException('Could not parse "$text".');
  }
}

/// Parses a version constraint.
///
/// Following rules from `package:pub_semver <= 2.1.0`.
VersionConstraint _parsePossiblyBrokenVersionConstraint(String text) {
  var originalText = text;

  void skipWhitespace() {
    text = text.trim();
  }

  skipWhitespace();

  // Handle the "any" constraint.
  if (text == 'any') return VersionConstraint.any;

  // Try to parse and consume a version number.
  Version? matchVersion() {
    var version = _startVersion.firstMatch(text);
    if (version == null) return null;

    text = text.substring(version.end);
    return Version.parse(version[0]!);
  }

  // Try to parse and consume a comparison operator followed by a version.
  VersionRange? matchComparison() {
    var comparison = _startComparison.firstMatch(text);
    if (comparison == null) return null;

    var op = comparison[0];
    text = text.substring(comparison.end);
    skipWhitespace();

    var version = matchVersion();
    if (version == null) {
      throw FormatException('Expected version number after "$op" in '
          '"$originalText", got "$text".');
    }

    switch (op) {
      case '<=':
        return VersionRange(max: version, includeMax: true);
      case '<':
        return VersionRange(
            max: version, includeMax: false, alwaysIncludeMaxPreRelease: true);
      case '>=':
        return VersionRange(min: version, includeMin: true);
      case '>':
        return VersionRange(min: version, includeMin: false);
    }
    throw FallThroughError();
  }

  // Try to parse the "^" operator followed by a version.
  VersionConstraint? matchCompatibleWith() {
    if (!text.startsWith(_compatibleWithChar)) return null;

    text = text.substring(_compatibleWithChar.length);
    skipWhitespace();

    var version = matchVersion();
    if (version == null) {
      throw FormatException('Expected version number after '
          '"$_compatibleWithChar" in "$originalText", got "$text".');
    }

    if (text.isNotEmpty) {
      throw FormatException('Cannot include other constraints with '
          '"$_compatibleWithChar" constraint in "$originalText".');
    }

    return VersionConstraint.compatibleWith(version);
  }

  var compatibleWith = matchCompatibleWith();
  if (compatibleWith != null) return compatibleWith;

  Version? min;
  var includeMin = false;
  Version? max;
  var includeMax = false;

  for (;;) {
    skipWhitespace();

    if (text.isEmpty) break;

    var newRange = matchVersion() ?? matchComparison();
    if (newRange == null) {
      throw FormatException('Could not parse version "$originalText". '
          'Unknown text at "$text".');
    }

    if (newRange.min != null) {
      if (min == null || newRange.min! > min) {
        min = newRange.min;
        includeMin = newRange.includeMin;
      } else if (newRange.min == min && !newRange.includeMin) {
        includeMin = false;
      }
    }

    if (newRange.max != null) {
      if (max == null || newRange.max! < max) {
        max = newRange.max;
        includeMax = newRange.includeMax;
      } else if (newRange.max == max && !newRange.includeMax) {
        includeMax = false;
      }
    }
  }

  if (min == null && max == null) {
    throw const FormatException('Cannot parse an empty string.');
  }

  if (min != null && max != null) {
    if (min > max) return VersionConstraint.empty;
    if (min == max) {
      if (includeMin && includeMax) return min;
      return VersionConstraint.empty;
    }
  }

  return VersionRange(
      min: min, includeMin: includeMin, max: max, includeMax: includeMax);
}
```

This is an adoption of what I used when detecting if we had broken versions in the first place.

If a version doesn't satisfy the new pattern (which isn't in a published version of pub_semver yet), then we need to parse with these functions. And use the canonicalized constraint, which `toString()` should create, because I don't pass `text` to the constructor.

isoos: Updated. I think the parsing with the spaces was equivalent, but you are right, better to use the snapshot of the pub_semver library. I've put it into a separate file, just in case we have more of these in the future, let's not mix them.

jonasfj (approved):

pkg/pub_package_reader/lib/src/pubspec_content_override.dart
isoos: > Afaik spaces are not required

TIL. yeah, better to stick with the existing code.

isoos:
I've deployed this to staging.

isoos:
I've confirmed this working and fixing the issue on staging.

Issue #5501, opened 2022-02-07
Cache-Control headers are missing for static file when signed-in


jonasfj:
This is only when signed in... URLs have content hash embedded in the URL, so we can cache it for a long time.

isoos:
The default cache control header for static resources is private. Should we update it to public too?
https://github.com/dart-lang/pub-dev/blob/master/app/lib/frontend/handlers/misc.dart#L151-L154

isoos:
This has been fixed.

PR #5506, opened 2022-02-08
Better override for cache headers in a user session.

isoos:
#5501

### DIFF
--- a/app/lib/shared/handler_helpers.dart
+++ b/app/lib/shared/handler_helpers.dart
@@ -277,9 +277,8 @@ shelf.Handler _userSessionWrapper(Logger logger, shelf.Handler handler) {
       // Responses for a user session are intended for a single user and must not
       // be stored by a shared, public cache. A private cache may store the response.
       //
-      // - We are keep parameters, e.g. max-age.
-      // - We are replacing `public` to `private`.
-      // - When the original header is empty, we set it to `private`.
+      // - replacing `public` to `private`
+      // - otherwise, keeping parameters, e.g. max-age
       final original = rs.headers[HttpHeaders.cacheControlHeader] ?? '';
       final parts = original
           .split(',')
@@ -287,8 +286,10 @@ shelf.Handler _userSessionWrapper(Logger logger, shelf.Handler handler) {
           .where((e) => e.isNotEmpty)
           .map((e) => e == 'public' ? 'private' : e)
           .toList();
-      final newValue = parts.isEmpty ? 'private' : parts.join(', ');
-      rs = rs.change(headers: {HttpHeaders.cacheControlHeader: newValue});
+      if (parts.isNotEmpty) {
+        final newValue = parts.join(', ');
+        rs = rs.change(headers: {HttpHeaders.cacheControlHeader: newValue});
+      }
     }
     return rs;
   };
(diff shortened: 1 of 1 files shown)


app/lib/shared/handler_helpers.dart
```diff
@@ -274,9 +274,21 @@ shelf.Handler _userSessionWrapper(Logger logger, shelf.Handler handler) {
...
+      //
+      // - We are keep parameters, e.g. max-age.
+      // - We are replacing `public` to `private`.
+      // - When the original header is empty, we set it to `private`.
+      final original = rs.headers[HttpHeaders.cacheControlHeader] ?? '';
+      final parts = original
+          .split(',')
+          .map((e) => e.trim())
+          .where((e) => e.isNotEmpty)
+          .map((e) => e == 'public' ? 'private' : e)
```
jonasfj, line 288: So if there is no `public` parameter, we don't insert `private` that seems risky.

This might work if everywhere we add cache control we also specify the `public` directive.

Perhaps we should think about this a bit more.. Do we really need to make static files `private`.

Might it not be better if we specify `Cache-Control` in each request handler? Where are we setting it now?

I'm not sure if this the best way to do it, just wondering if we don't risk creating some fragile logic here.

isoos: We set the cache-control header in two places: static files will get `private, max-age=[1 week]` (if the requested uri has a hash that matches the file's hash), documentation pages will get `private, max-age=[5 minutes]`. I think we could set both to `public`.

Yeah, maybe inserting `private` is a bit of a risk, I'll remove that.

jonasfj: I think what I'm trying to say is that I'm not sure rewriting `Cache-Control` headers we've already generated is a good idea.

If I understand it right, before this PR we just attached `Cache-Control: private` for all requests from anyone who is signed-in, right?

Is it correct that this is not a security requirement, merely a sanity check to ensure that when someone is signed-in we can't possibly set `Cache-Control: public` by accident. In other words, the old behavior is a safe-guard, right?

So we could simply remove the old behavior, right?
Or could we change all response handlers to explicitly set `Cache-Control` header?

Or maybe we implement a sanity check here, where we check that `Cache-Control: public, ...` is only allowed for `/static/` and `/documentation/` URLs. And if we see on anyother response then we fail 5xx. Assuming we're looking for a sanity check.

And/or as an extra safety precaution we can attach `Cache-Control: private` to any response that is authenticated and doesn't have an existing `Cache-Control` header.

------------------------

My concern here is mostly that we're not specifying cache-control in each request handler..

---------------

hmm, I guess I can also get onboard with:
 * always rewrite `Cache-Control` header to include `private` when authentication/cookie is present.
 * add `Cache-Control: private` if there is no header and authentication/cookie is present.

But then we should do proper `Cache-Control` header parsing..  in a function, or include the grammar in a comment:
```dart
     /// Parse [`Cache-Control`][1] header
     ///
     /// ```
     /// Cache-Control   = 1#cache-directive
     /// cache-directive = token [ "=" ( token / quoted-string ) ]
     /// ```
     /// [1]: https://datatracker.ietf.org/doc/html/rfc7234#section-5.2
```
Or something like that.

jonasfj (approved):

PR #5507, opened 2022-02-08
Cleanup SearchContext.isFlutterFavorite

isoos:
We no longer need a separate context for FF.


jonasfj (approved):
Maybe, update the changelog to ensure that we test this when we deploy.

isoos:
> Maybe, update the changelog to ensure that we test this when we deploy.

https://pub.dev/flutter/favorites already redirects.

isoos:
Also: similar cleanup is due for `SearchContext.dart()` and `SearchContext.flutter()`, but this was the one without the least amount of affected code parts.

PR #5521, opened 2022-02-16
deferred loading in pkg/web_app: api client, admin page widgets and package:intl

isoos:
- reduces main js size by about 25%
- test to track file size
- test to check if we have the proper library uses with `deferred as`
- further changes planned:
  - removing dependency on `package:intl`
  - refactoring the uses of `package:markdown` to be also deferred


pkg/web_app/lib/src/api_client/_authenticated_client.dart
```diff
@@ -2,23 +2,26 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'package:http/browser_client.dart';
-import 'package:http/http.dart';
+import 'dart:async';
+import 'dart:html';
+
+import 'package:http/browser_client.dart' as browser_client;
+import 'package:http/http.dart' as http;
```
jonasfj, line 9: What are changing in this file?

pkg/web_app/lib/src/api_client/api_client.dart
```diff
@@ -0,0 +1,43 @@
...
+  return Uri(
+    scheme: location.scheme,
+    host: location.host,
+    port: location.port,
+  ).toString();
+}
+
+/// The pub API client to use without credentials.
+PubApiClient get unauthenticatedClient =>
+    PubApiClient(_baseUrl, client: browser_client.BrowserClient());
```
jonasfj, line 26: Won't `http.Client()` work?

jonasfj (approved):

pkg/web_app/lib/src/api_client/_authenticated_client.dart
isoos: IIRC this is only a move, but it has some leftovers from an experiment where I've used it as part of `api_client.dart`.

pkg/web_app/lib/src/admin_pages.dart
```diff
@@ -0,0 +1,477 @@
...
+import 'dart:async';
+import 'dart:html';
+
+import 'package:_pub_shared/data/account_api.dart';
+import 'package:_pub_shared/data/package_api.dart';
+import 'package:_pub_shared/data/publisher_api.dart';
+
+import '_authentication_proxy.dart';
+import '_dom_helper.dart';
+import 'api_client/api_client.dart' deferred as api_client;
```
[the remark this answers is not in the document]
isoos: I think that using the same `api_client.dart` as deferred will only create a single split point, regardless of how many places use it. We should rather concentrate the heavy libraries after a split point barrier, so that they will be fetched together. Not sure how we can get to that, first, I'll remove the `package:intl` dependency, so that we don't need to worry about that...

PR #5527, opened 2022-02-18
Explicit and centralized cache-control headers.

isoos:
#5501

### DIFF
--- a/app/lib/frontend/handlers/headers.dart
+++ b/app/lib/frontend/handlers/headers.dart
@@ -13,13 +13,20 @@ class CacheHeaders {
   /// The maximum age while a response may be cached.
   final Duration maxAge;
 
-  CacheHeaders._(this.maxAge);
+  /// The cache storage category used for signed-in users.
+  /// `private` by default.
+  final String? signedInStorage;
+
+  CacheHeaders._(
+    this.maxAge, {
+    this.signedInStorage,
+  });
 
   Map<String, String> call() {
     final isSignedin = userSessionData != null;
     return <String, String>{
       HttpHeaders.cacheControlHeader: <String>[
-        isSignedin ? 'private' : 'public',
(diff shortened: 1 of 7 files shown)


app/lib/frontend/handlers/headers.dart
```diff
@@ -0,0 +1,47 @@
...
+class CacheHeaders {
+  /// The maximum age while a response may be cached.
+  final Duration maxAge;
+
+  /// The default cache storage category.
+  /// `public` by default
+  final String storage;
+
+  /// The cache storage category used for signed-in users.
+  /// `private by default
```
jonasfj, line 19:
```suggestion
  /// `private` by default
```

app/lib/frontend/handlers/headers.dart
```diff
@@ -0,0 +1,47 @@
...
+  final String signedInStorage;
+
+  CacheHeaders._(
+    this.maxAge, {
+    this.storage = 'public',
+    this.signedInStorage = 'private',
```
jonasfj, line 25: Since these are never used, why don't we just make private or constants, or just inline them in `call()`?

jonasfj (approved):

app/lib/shared/handler_helpers.dart
```diff
@@ -272,13 +272,7 @@ shelf.Handler _userSessionWrapper(Logger logger, shelf.Handler handler) {
...
       }
     }
-    shelf.Response rs = await handler(request);
-    if (userSessionData != null) {
-      // Indicates that the response is intended for a single user and must not
-      // be stored by a shared cache. A private cache may store the response.
-      rs = rs.change(headers: {HttpHeaders.cacheControlHeader: 'private'});
-    }
-    return rs;
+    return await handler(request);
```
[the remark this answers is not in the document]
isoos: updated, PTAL

PR #5540, opened 2022-03-01
Generic Admin API for tools under migration.

isoos:


app/lib/frontend/handlers/pubapi.dart
```diff
@@ -412,6 +412,13 @@ class PubApi {
   // **** Admin API
   // ****
 
+  @EndPoint.post('/api/admin/tools/<tool>/<args|[^]*>')
```
jonasfj, line 415: Why not use querystring for parameters?

```js
GET /api/admin/tools/
{
  "tools": [
    {
      "name": "merge-user",
      "parameters": {
        "mergeFromUserId": "UserId that should be merged into another user, this userId will be deleted",
        "mergeIntoUserId": "UserId that another user should be merged into",
      },
    },
    ...
  ],
}

POST /api/admin/tools/<tool>?param=value&...
{
  "success": true | false,
  "log": "some long text\nwith newlines explaining what happened."
}
```

In our code we could do something like:

```dart

/// Representation of an admin tool.
@sealed
class AdminTool {
  /// Short description of this tool.
  final String description;

  /// Map from parameter to documentation for the parameter,
  final Map<String,String> params;

  /// [run] this tool with [arguments] as a mapping from parameter name to argument.
  ///
  /// [log] allows messages to be returned to the caller.
  ///
  /// Returns true, if the tool was successful, false if the tool failed.
  /// If this throws an exception this will cause a 5xx response.
  final FutureOr<bool> Function(Map<String,String> arguments, void Function(String) log) run;

  AdminTool._({this.description, this.params, this.run});
  factory AdminTool({
    required String description,
    required Map<String,String> params,
    required FutureOr<bool> Function(Map<String,String> arguments, void Function(String) log) run,
  }) => AdminTool._({description, params, run});

  // Register of admin tools exposed.
  static List<AdminTool> tools = [
    mergeUser,
  ];
}

final mergeUser = AdminTool(
  description: 'Tool for merging two users',
  params: {
    'mergeFromUserId': "UserId that should be merged into another user, this userId will be deleted"
    'mergeIntoUserId': "UserId that another user should be merged into"
  },
  run: (Map<String,String> args, void Function(String) log) async {
    final mergeFromUserId = args['mergeFromUserId'] as String;
    final mergeIntoUserId = args['mergeIntoUserId'] as String;

    log('Merging users');
    //...
    return true;
  }
);

```

jonasfj: It's also possible I'm over thinking this... I think I'm a little worried about using ArgParser because if called incorrectly it might crash the server :see_no_evil:

jonasfj (approved):

PR #5548, opened 2022-03-03
Deferred load of page updater (used on search)

isoos:
- reduces main JS size by about 10% (33Kb)
- triggering the async load of the library at the time of detecting the update-enabled search controls, to reduce the latency of the update when the search actually happens
- increases the split parts count from 5 to 7


isoos:
Oh, apparently this is also referenced from the `main.dart`, I need to redesign it a bit before we can defer it.

PR #5549, opened 2022-03-03
Reducing JS size with more deferred libraries.

isoos:
- updated test to differentiate between accepted imports
- deferring `package:http` (~ 35Kb)
- deferring `package:markdown` (~35Kb)
- removing `package:collection` (not much change)
- overall: ~70 Kb reduction (-20%)

### DIFF
--- a/app/test/frontend/static_files_test.dart
+++ b/app/test/frontend/static_files_test.dart
@@ -153,7 +153,7 @@ void main() {
     test('script.dart.js size check', () {
       final file = cache.getFile('/static/js/script.dart.js');
       expect(file, isNotNull);
-      expect((file!.bytes.length / 1024).round(), closeTo(337, 1));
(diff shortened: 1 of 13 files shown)


isoos:
It turns out that the gains observed in #5548 was mostly deferring the `package:http` library.

jonasfj (approved):
I don't mind this... but I wonder if it wouldn't be preferable to split everything into 2 or 3 bundles...

If we make sure that markdown is only used from admin page, then isn't it enough to split admin page into a deferred?

isoos:
> I wonder if it wouldn't be preferable to split everything into 2 or 3 bundles...

That was the original plan, but it seems that we are using quite a lot of features everywhere, and there is no clear separation for only 2-3 bundles.

> If we make sure that markdown is only used from admin page, then isn't it enough to split admin page into a deferred?

Unfortunately we use it elsewhere too, for simplified formatting (yeah, we could probably do it differently too).

PR #5551, opened 2022-03-04
Allow all tag prefixes in search query and in Datastore.

isoos:
- Reduces the dependency on tags (fewer things to share with the JS web app).
- Worst case people will misstype their tags in queries and won't get any results.

### DIFF
--- a/app/lib/search/search_service.dart
+++ b/app/lib/search/search_service.dart
@@ -26,11 +26,6 @@ const int maxPageLinks = 10;
 /// The maximum length of the search query's text phrase that we'll try to serve.
 const _maxQueryLength = 256;
 
-/// The tag prefixes that we can detect in the user-provided search query.
-final _detectedTagPrefixes = <String>{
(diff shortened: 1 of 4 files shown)


jonasfj (approved):

app/lib/search/search_service.dart
```diff
@@ -512,10 +506,7 @@ class ParsedQueryText {
     final allPublishers = extractRegExp(_publisherRegexp);
     final publisher = allPublishers.isEmpty ? null : allPublishers.first;
 
-    final tagValues = extractRegExp(
-      _tagRegExp,
-      where: (tag) => _detectedTagPrefixes.any((p) => tag!.startsWith(p)),
-    );
+    final tagValues = extractRegExp(_tagRegExp);
```
jonasfj, line 509: Won't this cause problems for any keywords containing a colon?

isoos:
Discussed offline, we'll be keeping the known prefix list for now.

PR #5580, opened 2022-03-11
Extend code coverage with auxiliary packages in pkg/

isoos:


pkg/code_coverage/bin/code_coverage.dart
```diff
@@ -0,0 +1,69 @@
...
+        '--coverage',
+        rawOutputDir,
+      ],
+      workingDirectory: pkgDir,
+    );
+    print('Formatting coverage in pkg/$pkg...');
+    await Process.run(
+      'dart',
+      [
+        'pub',
```
jonasfj, line 53: Just do `dart run` not `dart pub run`

jonasfj (approved):

PR #5585, opened 2022-03-16
Local HTTP server for unit tests.

### LINKED ISSUE #5583: Make `pubHostedUrl` configurable in `PanaRunner`
Calling `processJobsWithPanaRunner` results in packages being downloaded from pub.dev. For test purposes, we should talk to the test profile for downloading archives.

isoos:
- Fixes #5583.
- Opens a new port for the frontend HTTP server and accepts request that are forwarded with the correct scoping to the local services.
- Does not force each local HTTP request through HTTP (will be done in a subsequent change).
- Adjusted tests to handle dynamic local ports.

### DIFF
--- a/app/lib/service/services.dart
+++ b/app/lib/service/services.dart
@@ -144,17 +144,6 @@ Future<R> withFakeServices<R>({
         primaryApiUri: frontendServerUri,
         primarySiteUri: frontendServerUri,
       );
-      serveRequests(frontendServer.server, (rq) async {
-        // TODO: find a way where we don't need to re-initialize all of the fake services.
-        return await withFakeServices(
-          fn: () async {
-            return await (createAppHandler()(rq));
-          },
-          configuration: configuration,
-          datastore: datastore,
-          storage: storage,
-        );
-      });
     }
     registerActiveConfiguration(configuration!);
 
@@ -166,6 +155,11 @@ Future<R> withFakeServices<R>({
         FakeUploadSignerService(configuration!.storageBaseUrl!));
     return await _withPubServices(() async {
       await youtubeBackend.start();
+      if (frontendServer != null) {
+        final frontendServerSubscription = frontendServer.server
+            .listen((rq) => handleRequest(rq, createAppHandler()));
+        registerScopeExitCallback(frontendServerSubscription.cancel);
+      }
       return await fn();
     });
   }) as R;
--- a/app/lib/service/services.dart
+++ b/app/lib/service/services.dart
@@ -10,6 +10,7 @@ import 'package:fake_gcloud/mem_storage.dart';
 import 'package:gcloud/service_scope.dart';
 import 'package:gcloud/storage.dart';
 import 'package:googleapis_auth/auth_io.dart' as auth;
+import 'package:shelf/shelf_io.dart';
 
 import '../account/backend.dart';
 import '../account/consent_backend.dart';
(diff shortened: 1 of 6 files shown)


app/lib/service/services.dart
```diff
@@ -105,14 +107,14 @@ Future<void> withServices(FutureOr<void> Function() fn) async {
...
-  required FutureOr<void> Function() fn,
+Future<R> withFakeServices<R>({
+  required FutureOr<R> Function() fn,
   Configuration? configuration,
   MemDatastore? datastore,
   MemStorage? storage,
 }) async {
   if (Zone.current[_pubDevServicesInitializedKey] == true) {
-    return await fork(() async => await fn());
+    return await fork(() async => await fn()) as R;
```
jonasfj, line 117: Can we make a utility function wrapping `fork`, called `saneFork` or maybe just `fork` and we place it somewhere we can import it from...

That way we won't need this cast every time we do something with `fork`.

app/lib/service/services.dart
```diff
@@ -123,15 +125,34 @@ Future<void> withFakeServices({
...
+
       // update configuration
       configuration = Configuration.test(
-          storageBaseUrl: 'http://localhost:${storageServer.port}');
+        storageBaseUrl: 'http://localhost:${storageServer.port}',
+        primaryApiUri: frontendServerUri,
+        primarySiteUri: frontendServerUri,
+      );
+      serveRequests(frontendServer.server, (rq) async {
+        return await withFakeServices(
```
jonasfj, line 147: Just curious, why do we need to call `withFakeServices` again?

app/test/frontend/handlers/_utils.dart
```diff
@@ -42,11 +43,13 @@ Future<shelf.Response> issueGet(
 Future<shelf.Response> issueHttp(
```
jonasfj, line 43: What does this method do?

app/test/frontend/handlers/account_test.dart
```diff
@@ -46,6 +47,7 @@ void main() {
         await issueGet(
           '/my-packages',
           headers: {'cookie': cookie},
+          host: activeConfiguration.primaryApiUri!.host,
         ),
```
jonasfj, line 51: Do we ever use `issueGet` for other `host` values?

If so, can we at-least just make two `issueGet` methods? I'm also okay, leaving this for later -- there's probably details here I'm not fully understanding :D

And it's fine to do things in steps.

jonasfj (approved):

app/lib/service/services.dart
isoos: The HTTP server's request processing would be in a new top-level scope, we need to re-initialize everything. It is possible that we could get the same result without it, but it may need tweaking how we subscribe to the server.

app/test/frontend/handlers/_utils.dart
isoos: Bridges to the shelf handlers without the overhead of the real HTTP connection and request parsing.
Now that we need the HTTP server anyway, I think we should migrate the code to call it directly.

szakarias (approved):
Very cool.

app/test/frontend/handlers/account_test.dart
isoos: > Do we ever use issueGet for other host values?

Yes, we have redirect tests where we have different behavior for different hosts, and we test it this way.

+1 on the migration/splitting this method...

app/lib/service/services.dart
jonasfj: But are we creating new instances of fake services... couldn't that have negative side-effects for stateful fake services?

isoos: The stateful parts (datastore and storage) are passed in, the rest is mostly cheap object creation that lives only for the request scope and will be discarded.

I also wish that we wouldn't need this, so I'll add a TODO and revisit it later.

jonasfj: great, because I plan on adding more stateful services :see_no_evil:

Issue #5578, opened 2022-03-10
Deprecate envConfig -- reduce configuration loaded from environment variables.


jonasfj:
A side from environment variables:
 * `GOOGLE_CLOUD_PROJECT`
 * `GAE_SERVICE`
 * `GAE_VERSION`

There is probably no need for us to load information from environment variables.

If we do, we can just use environment variable interpolated into the configuration, following https://github.com/dart-lang/pub-dev/pull/5577

Ideally, this means that we can delete most of the contents from:
https://github.com/dart-lang/pub-dev/blob/master/app/lib/shared/env_config.dart

PR #5588, opened 2022-03-16
Reduce the use of environment variables.

isoos:
#5578

### DIFF
--- a/app/config/dartlang-pub-dev.yaml
+++ b/app/config/dartlang-pub-dev.yaml
@@ -35,7 +35,7 @@ admins:
   permissions:
   - manageAssignedTags
 tools:
-  # Needs to be in sync with Dockerfile
+  # Needs to be in sync with Dockerfile, setup-dart.sh and setup-flutter.sh
   stableDartSdkPath: '/tool/stable/dart-sdk'
   stableFlutterSdkPath: '/tool/stable/flutter'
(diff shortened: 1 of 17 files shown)


isoos:
I'm a bit undecided which is better: hardcoding the values like the isolate counts, or moving them to config like the SDK paths. Any preference?

app/lib/service/entrypoint/frontend.dart
```diff
@@ -48,6 +48,8 @@ class DefaultCommand extends Command {
       logger: _logger,
       frontendEntryPoint: _main,
       workerEntryPoint: envConfig.isRunningLocally ? null : _worker,
+      frontendCount: envConfig.isRunningInAppengine ? 4 : 1,
+      workerCount: 1,
```
jonasfj, line 52: We can also put this in `Configuration`, but this is probably also fine.

app/config/dartlang-pub-dev.yaml
```diff
@@ -34,3 +34,9 @@ admins:
   email: assigned-tags-admin@dartlang-pub-dev.iam.gserviceaccount.com
   permissions:
   - manageAssignedTags
+tools:
+  # Needs to be in sync with Dockerfile
```
jonasfj, line 38: And needs to be in sync with `tool/setup-dart.sh` too, right?

jonasfj (approved):

PR #5595, opened 2022-03-18
git gc is no longer needed on local Flutter repository

isoos:
Previously, the version check in Flutter caused the git repository of the cloned Flutter SDK to grow (while it was fetching the extra commits, but not applying them). Since https://github.com/flutter/flutter/pull/76107 has been merged, and we started using both `--no-version-check` for all Flutter command in pana, and also set `CI=true` for analysis and dartdoc, there seems to be no need to keep doing `git gc`.


jonasfj (approved):
This might be worth noting in `CHANGELOG`.. just in case it causes surprises down the line.

PR #5598, opened 2022-03-21
Remove exported key-based upload signer.

isoos:
- #5578
- I think this was only used locally, which is now deprecated.


jonasfj (approved):

jonasfj:
Yeah, I'm pretty sure it was only used locally and we stopped using credentials locally.

PR #5604, opened 2022-03-22
Refactor PackageLinks

isoos:
- `repositoryBaseUrl` should use pana's `RepositoryUrl`, as it has better infer logic
- Instead of `PackageLinks`, we should eventually use URL analysis from `pana`, this is a first step to decouple it from `PackageVersion` internals, so that we won't use them by accident.


jonasfj (approved):

jonasfj:
Let's hold off merging for a day or two..

PR #5606, opened 2022-03-23
Abstractions for managing a pool of VMs executing tasks

jonasfj:
This should allow us to manage multiple separate pools of VMs on GCE. Including listing instances, and thus, recovering state and not forgetting to delete old instances.

Hmm, I guess special care must be taken to delete instances that are created by a previous `runtimeVersion`. Though this might not be important because preemptible instances can only last 24 hours or so.

### DIFF
--- a/app/lib/task/cloudcompute/cloudcompute.dart
+++ b/app/lib/task/cloudcompute/cloudcompute.dart
@@ -1,0 +1,91 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Interface describing an instance.
+abstract class CloudInstance {
+  /// Unique instance name.
+  String get name;
+
+  /// Zone this instance lives in.
+  String get zone;
+
+  /// [DateTime] of when the instance was created.
+  DateTime get created;
+
+  /// State of the instance.
+  InstanceState get state;
+
+  /// Representation for debugging purposes.
+  @override
+  String toString() => 'CloudInstance(${[
+        'name: $name',
+        'zone: $zone',
+        'created: $created',
+        'state: $state',
+      ].join(',')})';
+}
+
+/// Simplified instance state.
+///
+/// In practice we often don't care if an instance _provisioning_ or _staging_.
+/// If an instance is stuck in the [pending] state it's probably because it's
+/// not going to startup.
+///
+/// Similarly, we don't care if an instance is _booting_, _running_, or if the
+/// process running on the instance crashed without terminating the instance.
+/// In all these cases we want delete the instance if it gets stuck long
+/// enought to exhaust allowed execution time.
+///
+/// We also don't care if the instance is _suspending_, _suspended_, _stopping_,
+/// _stopped_ or _terminated_. In these the cases the instance should be
+/// deleted.
+enum InstanceState {
+  /// Instance is pending provisioning or in the process of being provisioned.
+  pending,
+
+  /// Instance is believed to be running (or booting).
+  running,
+
+  /// Instance has been terminated.
+  terminated,
+}
+
+/// Interface for controlling a cloud compute region.
+abstract class CloudCompute {
+  /// List of zones available.
+  List<String> get zones;
+
+  /// Generate an instance name.
+  String generateInstanceName();
+
+  /// Create an instance running [dockerImage] with given [arguments].
+  ///
+  /// The instance will be running in the given [zone] under the given [name].
+  /// Notice that [name] must be unique and will have implementation specific
+  /// limitations, using [generateInstanceName] is adviced.
+  ///
+  /// The _human readable_ [description] will attached to the instance in
+  /// Cloud Console and intended to assist humans inspecting the system.
+  ///
+  /// The future returned may take a long time to resolve (several minutes).
+  /// Any [Exception] thrown by this method should be considered a reason to
+  /// reduce the request rate on the given zone [zone].
+  Future<CloudInstance> createInstance({
+    required String zone,
+    required String name,
+    required String dockerImage,
+    required List<String> arguments,
+    required String description,
+  });
+
+  /// List instances.
+  Stream<CloudInstance> listInstances();
+
+  /// Delete the instance with [name] from [zone].
+  ///
+  /// If this operation fails, it may throw an [Exception] or discard the error.
+  /// Systems should assume deletion is best-effort and periodically list
+  /// instances to delete instances that are no longer needed.
+  Future<void> delete(String zone, String name);
+}
--- a/app/lib/task/cloudcompute/cloudcompute.dart
+++ b/app/lib/task/cloudcompute/cloudcompute.dart
@@ -1,4 +1,4 @@
-// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
@@ -36,8 +36,8 @@ abstract class CloudInstance {
 ///
 /// Similarly, we don't care if an instance is _booting_, _running_, or if the
 /// process running on the instance crashed without terminating the instance.
-/// In all these cases we want delete the instance if it gets stuck long
-/// enought to exhaust allowed execution time.
+/// In all these cases we want to delete the instance if it gets stuck long
+/// enough to exhaust the allowed execution time.
 ///
 /// We also don't care if the instance is _suspending_, _suspended_, _stopping_,
 /// _stopped_ or _terminated_. In these the cases the instance should be
@@ -68,10 +68,10 @@ abstract class CloudCompute {
   /// Notice that [instanceName] must be unique and will have implementation
   /// specific limitations, using [generateInstanceName] is adviced.
   ///
-  /// The _human readable_ [description] will attached to the instance in
+  /// The _human readable_ [description] will be attached to the instance in
   /// Cloud Console and intended to assist humans inspecting the system.
   ///
-  /// The future returned may take a long time to resolve (several minutes).
+  /// The [Future] returned may take a long time to resolve (several minutes).
   /// Any [Exception] thrown by this method should be considered a reason to
   /// reduce the request rate on the given zone [zone].
   Future<CloudInstance> createInstance({
--- a/app/lib/task/cloudcompute/cloudcompute.dart
+++ b/app/lib/task/cloudcompute/cloudcompute.dart
@@ -30,7 +30,7 @@ abstract class CloudInstance {
 
 /// Simplified instance state.
 ///
-/// In practice we often don't care if an instance _provisioning_ or _staging_.
+/// In practice we often don't care if an instance is _provisioning_ or _staging_.
 /// If an instance is stuck in the [pending] state it's probably because it's
 /// not going to startup.
 ///
--- a/app/lib/task/cloudcompute/cloudcompute.dart
+++ b/app/lib/task/cloudcompute/cloudcompute.dart
@@ -5,12 +5,14 @@
 /// Interface describing an instance.
 abstract class CloudInstance {
   /// Unique instance name.
-  String get name;
+  String get instanceName;
 
   /// Zone this instance lives in.
   String get zone;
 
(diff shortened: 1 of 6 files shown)


app/lib/task/cloudcompute/googlecloudcompute.dart
```diff
@@ -0,0 +1,502 @@
...
+class _PendingGoogleCloudInstance extends CloudInstance {
+  /// GCP zone this instance exists inside.
+  @override
+  final String zone;
+
+  @override
+  final String name;
+
+  @override
+  DateTime get created => clock.now();
```
isoos, line 119: Not sure if this is relevant anywhere, but I'd expect the `created` to be `final`, e.g. `late final created = clock.now();`

isoos (approved):
On a high level this looks good, let's see it in use :)

app/lib/task/cloudcompute/cloudcompute.dart
```diff
@@ -0,0 +1,91 @@
...
+/// process running on the instance crashed without terminating the instance.
+/// In all these cases we want delete the instance if it gets stuck long
+/// enought to exhaust allowed execution time.
+///
+/// We also don't care if the instance is _suspending_, _suspended_, _stopping_,
+/// _stopped_ or _terminated_. In these the cases the instance should be
+/// deleted.
+enum InstanceState {
+  /// Instance is pending provisioning or in the process of being provisioned.
+  pending,
```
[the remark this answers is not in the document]
jonasfj: I'm not sure `provisioning` is better, because we are in a state here where we're asking for a VM.

It's entirely possible the request could be denied, then we jump to `terminated`.

I've merged a few states because we don't want to distinguish between too many of them :D

app/lib/task/cloudcompute/cloudcompute.dart
```diff
@@ -0,0 +1,91 @@
...
+
+  /// Instance is believed to be running (or booting).
+  running,
+
+  /// Instance has been terminated.
+  terminated,
```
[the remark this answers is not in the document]
jonasfj: Yes, but as far as we are concerned there is no difference.

There is also a point at which the instance isn't visible in the API anymore.

app/lib/task/cloudcompute/googlecloudcompute.dart
jonasfj: But while pending the instance haven't been created yet.. I guess we could also use `DateTime(0)` as a special place-hoilder... or something... Or some future time...

But maybe, I should document this behavior.

app/lib/task/cloudcompute/cloudcompute.dart
```diff
@@ -0,0 +1,91 @@
...
+        'name: $name',
+        'zone: $zone',
+        'created: $created',
+        'state: $state',
+      ].join(',')})';
+}
+
+/// Simplified instance state.
+///
+/// In practice we often don't care if an instance _provisioning_ or _staging_.
```
sigurdm, line 31: Broken sentence.

```suggestion
/// In practice we often don't care if an instance is _provisioning_ or _staging_.
```

app/lib/task/cloudcompute/cloudcompute.dart
```diff
@@ -0,0 +1,91 @@
...
+/// If an instance is stuck in the [pending] state it's probably because it's
+/// not going to startup.
+///
+/// Similarly, we don't care if an instance is _booting_, _running_, or if the
+/// process running on the instance crashed without terminating the instance.
+/// In all these cases we want delete the instance if it gets stuck long
```
sigurdm, line 37:
```suggestion
/// In all these cases we want to delete the instance if it gets stuck long
```

app/lib/task/cloudcompute/cloudcompute.dart
```diff
@@ -0,0 +1,91 @@
...
+/// enought to exhaust allowed execution time.
```
sigurdm, line 38:
```suggestion
/// enough to exhaust the allowed execution time.
```

app/lib/task/cloudcompute/cloudcompute.dart
```diff
@@ -0,0 +1,91 @@
...
+  /// Generate an instance name.
+  String generateInstanceName();
+
+  /// Create an instance running [dockerImage] with given [arguments].
+  ///
+  /// The instance will be running in the given [zone] under the given [name].
+  /// Notice that [name] must be unique and will have implementation specific
+  /// limitations, using [generateInstanceName] is adviced.
+  ///
+  /// The _human readable_ [description] will attached to the instance in
```
sigurdm, line 68:
```suggestion
  /// The _human readable_ [description] will be attached to the instance in
```

app/lib/task/cloudcompute/cloudcompute.dart
```diff
@@ -0,0 +1,91 @@
...
+  /// Cloud Console and intended to assist humans inspecting the system.
+  ///
+  /// The future returned may take a long time to resolve (several minutes).
```
sigurdm, line 71:
```suggestion
  /// The [Future] returned may take a long time to resolve (several minutes).
```

app/lib/task/cloudcompute/cloudcompute.dart
```diff
@@ -0,0 +1,91 @@
...
+  /// reduce the request rate on the given zone [zone].
+  Future<CloudInstance> createInstance({
+    required String zone,
+    required String name,
+    required String dockerImage,
+    required List<String> arguments,
+    required String description,
+  });
+
+  /// List instances.
```
sigurdm, line 82:
```suggestion
```

app/lib/task/cloudcompute/cloudcompute.dart
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
```
sigurdm, line 1:
```suggestion
// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
```

Here and elsewhere

app/lib/task/cloudcompute/fakecloudcompute.dart
```diff
@@ -0,0 +1,156 @@
...
+
+final _log = Logger('pub.fakecloudcompute');
+
+@sealed
+class FakeCloudCompute extends CloudCompute {
+  var _nextInstanceId = 1;
+  final _instances = <FakeCloudInstance>{};
+
+  @override
+  final List<String> zones = List.unmodifiable(['zone-a', 'zone-b']);
```
sigurdm, line 19: Maybe:
```suggestion
  final List<String> zones =  const ['zone-a', 'zone-b'];
```

app/lib/task/cloudcompute/fakecloudcompute.dart
```diff
@@ -0,0 +1,156 @@
...
+    return instance;
+  }
+
+  @override
+  Stream<FakeCloudInstance> listInstances() => Stream.fromIterable(_instances);
+
+  @override
+  Future<void> delete(String zone, String name) async {
+    if (!_instances.any((i) => i.name == name && i.zone == zone)) {
+      // This is not really a problem, GoogleCloudCompute should might throw
```
sigurdm, line 67:
```suggestion
      // This is not really a problem, GoogleCloudCompute should throw
```

app/lib/task/cloudcompute/fakecloudcompute.dart
```diff
@@ -0,0 +1,156 @@
...
+    _instances.removeWhere((i) => i.name == name && i.zone == zone);
+  }
+
+  /// Change state of instance with [name] to [InstanceState.running].
+  void fakeStartInstance(String name) {
+    if (!_instances.any((i) => i.name == name)) {
+      throw StateError('instance "$name" does not exist');
+    }
+
+    final instance = _instances.firstWhere((i) => i.name == name);
```
sigurdm, line 85:
```suggestion
    final instance = _instances.firstWhereOrNull((i) => i.name == name);
    if (instance == null) {
      throw StateError('instance "$name" does not exist');
    }
```
(requires package:collection)

app/lib/task/cloudcompute/fakecloudcompute.dart
```diff
@@ -0,0 +1,156 @@
...
+    _instances.removeWhere((i) => i.name == name);
```
sigurdm, line 86:
```suggestion
    _instances.remove(instance);
```

app/lib/task/cloudcompute/googlecloudcompute.dart
sigurdm: I think you should either make `created` nullable (null when instance has not been created) or raise a `StateError` for pending instances.

app/lib/task/cloudcompute/googlecloudcompute.dart
```diff
@@ -0,0 +1,502 @@
...
+          name, 'name', 'must have a length between 1 and 63');
+    }
+    if (!_validInstanceNamePattern.hasMatch(name)) {
+      throw ArgumentError.value(
+          name, 'name', 'must match pattern: $_validInstanceNamePattern');
+    }
+    // Max argument string size on Linux is MAX_ARG_STRLEN = 131072
+    // In addition the maximum meta-data size supported by GCE is 256KiB
+    // We need a few extra bits, so we shall enforce a max size of 100KiB.
+    if (arguments.fold<int>(0, (a, b) => a + b.length) > 100 * 1024) {
```
sigurdm, line 227: Are we counting utf16 code-units or bytes here?

You might have to encode the strings to do correct counting (or ensure everything is ascii).

app/lib/task/cloudcompute/googlecloudcompute.dart
```diff
@@ -0,0 +1,502 @@
...
+
+    return c.stream;
+  }
+
+  CloudInstance _wrapInstance(Instance instance, String zone) {
+    DateTime created;
+    try {
+      created = DateTime.parse(instance.creationTimestamp ?? '');
+    } on FormatException {
+      // Print error and instance to log..
```
sigurdm, line 454: Is this a TODO?
```suggestion
      // TODO(jonasfj): Print error and instance to log..
```

app/test/task/cloudcompute/googlecloudcompute_test.dart
```diff
@@ -0,0 +1,86 @@
...
+        await Future.delayed(Duration(seconds: 1));
+      }
+    });
+  },
+      timeout: Timeout.parse('30m'),
+      skip: Platform.environment['GOOGLE_CLOUD_PROJECT'] != null &&
+              // Avoid running against production by accident
+              Platform.environment['GOOGLE_CLOUD_PROJECT'] != 'dartlang-pub'
+          ? false
+          : 'createGoogleCloudCompute testing requires GOOGLE_CLOUD_PROJECT');
```
sigurdm, line 85: I find this conditional hard to read. Perhaps:
```suggestion
      skip: Platform.environment['GOOGLE_CLOUD_PROJECT'] == null ||
              // Avoid running against production by accident
              Platform.environment['GOOGLE_CLOUD_PROJECT'] == 'dartlang-pub'
          ? 'createGoogleCloudCompute testing requires GOOGLE_CLOUD_PROJECT'
          : false);
```

sigurdm (approved):
Just a few nits

app/lib/task/cloudcompute/googlecloudcompute.dart
jonasfj: maybe... I don't think I want to handle the corner case, the `created` time is only suitable for GC'ing old instances.

app/lib/task/cloudcompute/cloudcompute.dart
jonasfj: I do love documenting everything... it tells people that at-least I meant this thing to be public.

I see so many things in where I don't know if the author meant for me to use the thing, and the first thing I have to do is read a lot of code to determine if the thing was even intended to be exposed. I mean for you to use this :D

app/lib/task/cloudcompute/googlecloudcompute.dart
jonasfj: Pretty sure it's bytes, so yes... It would be safer...

app/lib/task/cloudcompute/fakecloudcompute.dart
jonasfj: Hmm, maybe, but I don't trust it... :see_no_evil: 

The instance objects are immutable like it is in the other API...

PR #5608, opened 2022-03-23
Introducing `InvalidInputException.checkPackageName`

jonasfj:
This makes it easier to protect APIs against garbage package names.
Thus, we produce better error messages and protect ourselves against
weir behavior whenever we expect something that should be a valid
package name.

### DIFF
--- a/app/lib/shared/exceptions.dart
+++ b/app/lib/shared/exceptions.dart
@@ -15,7 +15,6 @@ library exceptions;
 
 import 'package:api_builder/api_builder.dart' show ApiResponseException;
 import 'package:pub_dev/shared/utils.dart';
-import 'package:pub_semver/pub_semver.dart';
 
 /// Base class for all exceptions that are intercepted by HTTP handler wrappers.
 abstract class ResponseException extends ApiResponseException {
--- a/app/lib/shared/exceptions.dart
+++ b/app/lib/shared/exceptions.dart
@@ -158,14 +158,19 @@ class InvalidInputException extends ResponseException {
     _check(_ulidPattern.hasMatch(value), () => '"$name" is not a valid ulid.');
   }
 
+  /// Pattern for both new and existing package names.
+  ///
+  /// This allows upper-case characters in package names, because that exist in
+  /// some existing packages.
+  static final _packageNamePattern = RegExp(r'^[a-zA-Z_][a-zA-Z0-9_]*$');
+
   /// Throw [InvalidInputException] if [package] is not an allowed package name.
   static void checkPackageName(String? package) {
     // TODO: Reuse logic from validatePackageName in pub_package_reader.dart
-    //       This is for new existing packages! Not only new packages!
+    //       This is for new and existing packages! Not only new packages!
     checkNotNull(package, 'package');
     checkStringLength(package!, 'package', minimum: 1, maximum: 64);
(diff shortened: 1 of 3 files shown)


app/lib/shared/exceptions.dart
```diff
@@ -157,14 +158,28 @@ class InvalidInputException extends ResponseException {
     _check(_ulidPattern.hasMatch(value), () => '"$name" is not a valid ulid.');
   }
 
-  static void checkSemanticVersion(String? version) {
+  /// Throw [InvalidInputException] if [package] is not an allowed package name.
+  static void checkPackageName(String? package) {
```
isoos, line 162: We already have `checkPackageVersionParams` in `package/backend.dart`, it seems like a good candidate to merge this with.

app/lib/shared/exceptions.dart
```diff
@@ -157,14 +158,28 @@ class InvalidInputException extends ResponseException {
...
+    // TODO: Reuse logic from validatePackageName in pub_package_reader.dart
+    //       This is for new existing packages! Not only new packages!
+    checkNotNull(package, 'package');
+    checkStringLength(package!, 'package', minimum: 1, maximum: 64);
+    checkMatchPattern(package, 'package', RegExp(r'^[a-zA-Z0-9_]+$'));
```
isoos, line 167: Let's extract these as separate top-level variables?

app/lib/shared/exceptions.dart
```diff
@@ -157,14 +158,28 @@ class InvalidInputException extends ResponseException {
     _check(_ulidPattern.hasMatch(value), () => '"$name" is not a valid ulid.');
   }
 
-  static void checkSemanticVersion(String? version) {
+  /// Throw [InvalidInputException] if [package] is not an allowed package name.
+  static void checkPackageName(String? package) {
+    // TODO: Reuse logic from validatePackageName in pub_package_reader.dart
+    //       This is for new existing packages! Not only new packages!
```
isoos, line 164: `new existing` should be only `existing`?

app/lib/shared/exceptions.dart
```diff
@@ -157,14 +158,28 @@ class InvalidInputException extends ResponseException {
...
+    checkNotNull(package, 'package');
+    checkStringLength(package!, 'package', minimum: 1, maximum: 64);
+    checkMatchPattern(package, 'package', RegExp(r'^[a-zA-Z0-9_]+$'));
+    checkMatchPattern(package, 'package', RegExp(r'^[a-zA-Z_]'));
```
isoos, line 168: I think this is a bit confusing compared to one line up, maybe a single regexp that handles both the beginning and the rest of the package name would be better.

app/lib/shared/exceptions.dart
jonasfj: I think it's only supposed to check the format of the name, so both new and existing.

isoos: `new AND existing` then?

isoos (approved):

app/lib/shared/exceptions.dart
jonasfj: hmm, I used this in `checkPackageVersionParams`, but it allows  `latest` as version...

Issue #5571, opened 2022-03-09
Improve seach engine


enzodanjour:
When you search for a package in pub.dev you may have cases where the relevant name is not listed, for example I only search for fluent, my goal is to find fluent find the package fluent_ui, but I don't find it at any time, and I I googled the exact term to find the package, image below detailing.
![Screenshot from 2022-03-09 13-58-35](https://user-images.githubusercontent.com/35856303/157496715-dc0e3ffd-ec50-4cfe-922b-0d3ede8e370d.png)

to find the modular flutter package, I typed in the term modular, but modular flutter is not listed.
![Screenshot from 2022-03-09 14-30-25](https://user-images.githubusercontent.com/35856303/157497744-9b6ace50-61b7-4d70-8b06-210f31b79132.png)

to find get_it I typed get, but the term get_it is not listed.
![Screenshot from 2022-03-09 14-33-23](https://user-images.githubusercontent.com/35856303/157498268-13deeaa1-6c94-4228-8627-420c608a7662.png)

isoos:
Thanks for the example, we'll look into the weights to improve the ranking.

isoos:
This has been deployed, and there is an improvement on the production site for the `fluent` search. We have similar reports [here](https://github.com/dart-lang/pub-dev/issues/5721#issuecomment-1116176972), and a common pattern seems to be that our current ranking under-weights popular packages.

enzodanjour:
@isoos Provider not showed on first itens.
![Screenshot from 2022-09-16 11-33-26](https://user-images.githubusercontent.com/35856303/190663981-b6bb6b10-234f-4a8d-8de2-9c724a32893a.png)

isoos:
@enzodanjour: it is the first item after the SDK-libraries for me.

leighajarett:
+1 this came up in research sessions at Flutter Vikings and it resonated with several developers

isoos:
Several changes have been deployed recently to improve these queries, I think they provider okay answers now. Feel free to reopen with these (or new) queries if you find the results unsatisfying or unexpected.

PR #5612, opened 2022-03-25
Compensate match score with partial overlaps in the query word and the package name.

isoos:
- fixes a bug in the ngram extraction
- should improve #5571, by adding more weight to prefix/postfix matches of the query word, even for longer package names
- actually improves test cases where previously we were expecting more packages to show up
- only the flutter_iap test seems to be a questionable change, as it exposes more package name with flutter, but maybe that is okay
- should not increase the search/matching complexity too much, although where previously we may have cut short on a non-matching word, we no longer do that


isoos:
Deployed this to staging, and it shows improvement on the `fluent` search query, but there is not much change on the other two described in the issue.

jonasfj (approved):

Issue #5586, opened 2022-03-16
Migrate to uniform access policy for buckets


jonasfj:
Created the following buckets:

* Staging:
  *   `dartlang-pub-dev-canonical-packages` (private)
  *   `dartlang-pub-dev-public-packages` (public)
  *   `dartlang-pub-dev-incoming-packages` (private)

* Production:
  *   `dartlang-pub-canonical-packages` (private)
  *   `dartlang-pub-public-packages` (public)
  *   `dartlang-pub-incoming-packages` (private)

-----------------------------------------------------

## Migration Stages

 * [x] Stage A: Use `incoming-packages` for uploading new packages.
    * We still need to look in the old bucket, because client might be in the middle of an upload while we migrate.
 * [x] Stage B: Use `incoming-packages` for uploading new packages.
    * Stop looking the old bucket for newly uploaded packages.
 * [x] Stage C: Store in `canonical-packages`
    * After upload packages are copied to `canonical-packages` before declaring package as accepted (before transaction).
      * Overwriting a tarball in `canonical-packages` is not allowed, but if the tarball exists and the datastore entity does not exist, then we let the upload proceed if the tarball is byte-to-byte equivalent to the tarball in `canonical-packages`.
    * All existing packages are copied to `canonical-packages` (using a background job, or one-off job)
    * background job syncs the canonical bucket to the `public-packages` **and** the old bucket.
       * In effect we treat the old bucket the same as we would treat the public bucket.
    * Requests are served from the old bucket.
 * [x] Stage D: Serve from `public-bucket`
 * [ ] Stage E: Stop updating old bucket and delete it.

isoos:
This has been completed.

PR #5615, opened 2022-03-28
New bucket names in the configuration.

isoos:
#5586

### DIFF
--- a/app/lib/shared/configuration.dart
+++ b/app/lib/shared/configuration.dart
@@ -49,6 +49,22 @@ const _fakeSiteAudience = 'fake-site-audience';
   disallowUnrecognizedKeys: true,
 )
 class Configuration {
+  /// The name of the Cloud Storage bucket to use for storing the uploaded
+  /// package archives.
+  ///
+  /// The bucket content policy should be private.
+  final String? canonicalPackagesBucketName;
+
+  /// The name of the Cloud Storage bucket to use for public package archive downloads.
+  ///
+  /// The bucket content policy should be public.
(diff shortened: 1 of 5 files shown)


app/lib/shared/configuration.dart
```diff
@@ -49,6 +49,22 @@ const _fakeSiteAudience = 'fake-site-audience';
...
 )
 class Configuration {
+  /// The name of the Cloud Storage bucket to use for storing the uploaded
+  /// package archives.
+  ///
+  /// The bucket content policy should be private.
+  final String? canonicalPackagesBucketName;
+
+  /// The name of the Cloud Storage bucket to use for public package archive downloads.
+  ///
```
jonasfj, line 59:
```suggestion
  ///
  /// This is the bucket which users are redirected to when they want to download package tarballs.
```

app/lib/shared/configuration.dart
```diff
@@ -49,6 +49,22 @@ const _fakeSiteAudience = 'fake-site-audience';
...
+  /// The bucket content policy should be public.
+  final String? publicPackagesBucketName;
+
+  /// The name of the Cloud Storage bucket to use for incoming package archives.
+  ///
```
jonasfj, line 64:
```suggestion
  ///
  /// When users are publishing packages using the `dart pub` client, they are given a signed-url
  /// which allows the client to upload a file. That signed-url points to an object in this bucket.
  /// Once the uploaded tarball have been verified, it can be copied to the canonical bucket.
```

jonasfj (approved):

PR #5616, opened 2022-03-28
Migrate moderation-related tools.

isoos:


app/lib/admin/tools/remove_publisher_and_block_all_members.dart
```diff
@@ -0,0 +1,53 @@
...
+import 'package:pub_dev/shared/datastore.dart';
+
+Future<String> executeRemovePublisherAndBlockAllMembers(
+    List<String> args) async {
+  if (args.isEmpty ||
+      args.length > 2 ||
+      (args.length == 2 && args[1] != 'delete')) {
+    return 'Remove publisher and blocks all members.\n'
+        '  <tools-command> <publisherId> # list publisher data\n'
+        '  <tools-command> <publisherId> delete # remove publisher and block members\n';
```
jonasfj, line 19: Might it not be safer to do:
```suggestion
        '  <tools-command> list <publisherId> # list publisher data\n'
        '  <tools-command> delete <publisherId> # remove publisher and block members\n';
```

jonasfj (approved):

PR #5621, opened 2022-03-29
Consolidate the use of Platform.environment into env_config.dart.

### LINKED ISSUE #5578: Deprecate envConfig -- reduce configuration loaded from environment variables.

isoos:
- closes #5578 (I think we've reduced the scope and there is not much left to do there)
- also added a test to check if `Platform.environment` is used elsewhere

### DIFF
--- a/app/lib/shared/env_config.dart
+++ b/app/lib/shared/env_config.dart
@@ -2,52 +2,70 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:convert';
 import 'dart:io';
 
-/// Configuration from the environment variables.
-final EnvConfig envConfig = EnvConfig._detect();
+import 'package:crypto/crypto.dart';
+import 'package:meta/meta.dart';
 
 /// Configuration from the environment variables.
-class EnvConfig {
+final envConfig = _EnvConfig();
+
+/// Configuration from the environment variables.
+class _EnvConfig {
   /// Service in AppEngine that this process is running in, `null` if running
   /// locally.
-  final String? gaeService;
+  late final _gaeService = Platform.environment['GAE_SERVICE'];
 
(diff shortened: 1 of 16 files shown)


app/lib/shared/env_config.dart
```diff
@@ -2,52 +2,70 @@
...
-      Platform.environment['GAE_SERVICE'],
-      Platform.environment['GAE_VERSION'],
-      Platform.environment['GAE_INSTANCE'],
-      Platform.environment['GOOGLE_CLOUD_PROJECT'],
-      Platform.environment['PUB_SERVER_CONFIG'],
-    );
-  }
+  late final _gaeInstance = Platform.environment['GAE_INSTANCE'];
+
+  late final googleCloudProject = Platform.environment['GOOGLE_CLOUD_PROJECT'];
```
jonasfj, line 30: Should we perhaps make these, either:
 * (A) non-nullable, and then default to empty-string; OR;
 * (B) keep them nullable, but make the empty string result in a null value.

IMO, most of these won't be null at runtime, and therefore (A) is a reasonable thing to do.
We can document that they will be empty string, if not specified.

In tests, and other cases where we want to skip based on whether or not an env var is specified we can just check if the value is `notEmpty`.

jonasfj (approved):

app/lib/shared/env_config.dart
isoos: I'm not sure about it yet. For some of it, it may make sense, for a few others, less so. I've added a TODO to get back to it later.

PR #5626, opened 2022-03-30
Hash in the path-segments for the caching header of the main.dart.js parts files

isoos:
- the cache-level etag is calculated from the individual file's etags
- the requested URL pattern is `/static/hash-<etag>/<rest>` parsed into `/static/<rest>` using public cache headers when the `<etag>` matches the current one
- right now only the main.dart.js files are inside that hashed path segment, other files will be migrated in a subsequent PR (it also involves changing a hash-verification test that needs to be updated)
- added a cache header test in puppeteer response evaluation to catch this happening again
- deployed and verified in staging, inclusion and admin works

### DIFF
--- a/app/lib/frontend/static_files.dart
+++ b/app/lib/frontend/static_files.dart
@@ -355,7 +355,7 @@ Future<void> updateWebCssBuild() async {
 /// It can parse the following formats:
 /// - /static/<url-hash>/path/to/resource
 /// - /static/path/to/resource?hash=<url-hash>
-/// 
+///
 /// TODO: remove after we no longer use url-hash
 class ParsedStaticUrl {
   final String? urlHash;
--- a/app/lib/frontend/static_files.dart
+++ b/app/lib/frontend/static_files.dart
@@ -86,6 +86,7 @@ Directory _resolveDir(String relativePath) =>
 /// Stores static files in memory for fast http serving.
 class StaticFileCache {
   final _files = <String, StaticFile>{};
+  String? _etag;
 
   StaticFileCache();
 
@@ -162,6 +163,7 @@ class StaticFileCache {
   @visibleForTesting
   void addFile(StaticFile file) {
     _files[file.requestPath] = file;
+    _etag = null;
   }
 
   @visibleForTesting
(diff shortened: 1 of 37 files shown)


app/lib/frontend/static_files.dart
```diff
@@ -333,3 +350,33 @@ Future<void> updateWebCssBuild() async {
     throw Exception(message);
   }
 }
+
+@visibleForTesting
+class ParsedStaticUrl {
```
jonasfj, line 355: Can this have a tiny documentation comment?

app/lib/frontend/static_files.dart
```diff
@@ -333,3 +350,33 @@ Future<void> updateWebCssBuild() async {
     throw Exception(message);
   }
 }
+
+@visibleForTesting
```
jonasfj, line 354: This is exported from this library and used outside testing, I see it used in `app/lib/frontend/handlers/misc.dart`

app/lib/frontend/static_files.dart
jonasfj: I'm guessing it because we need to handle both URL hash and query hash, during the migration?

jonasfj (approved):

app/lib/frontend/static_files.dart
```diff
@@ -170,6 +172,16 @@ class StaticFileCache {
...
   StaticFile? getFile(String requestPath) => _files[requestPath];
+
+  String get etag => _etag ??= _calculateEtagOfEtags().substring(0, 8);
+
+  String _calculateEtagOfEtags() {
+    final files = List<StaticFile>.from(_files.values);
+    files.sort((a, b) => a.requestPath.compareTo(b.requestPath));
+    final concatenatedEtags = files.map((f) => f.etag).join(' ');
+    final digest = crypto.sha256.convert(utf8.encode(concatenatedEtags));
+    return digest.bytes.map((b) => (b & 31).toRadixString(32)).join();
```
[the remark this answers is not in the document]
isoos: A bit more information encoded in the same string length (regardless of how long it is). In the end, I don't think it matters...

app/lib/frontend/static_files.dart
isoos: Yes, after we'd no longer use the request parameter hash, this will become much simpler.

app/lib/frontend/static_files.dart
isoos: Ah, removed.

PR #5630, opened 2022-03-31
Refactor TarballStorageNamer + TarballStorage initialization

isoos:
- #5586
- reduces the complexity of the object names (e.g. we won't use namespaces)
- reduces the duplication of `TarballStorage` initialization in admin backend

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -681,7 +681,7 @@ class PackageBackend {
     await requireAuthenticatedUser();
     final guid = createUuid();
     _logger.info('Starting semi-async upload (uuid: $guid)');
-    final object = _storage.tempObjectName(guid);
+    final object = tmpObjectName(guid);
     await data.pipe(_storage.bucket.write(object));
     return await publishUploadedBlob(guid);
   }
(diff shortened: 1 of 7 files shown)


app/lib/package/backend.dart
```diff
@@ -1377,30 +1376,22 @@ DerivedPackageVersionEntities derivePackageVersionEntities({
 /// Helper utility class for interfacing with Cloud Storage for storing
 /// tarballs.
 class TarballStorage {
```
jonasfj, line 1378: Should we just do away with this wrapper? Do we need it?

What if we just document the bucket layout when we create an instance of `Bucket`?
```dart
/// Bucket where package tarballs are stored in.
///
/// Organized as follows:
/// ```
/// gs:/<packageBucketName>/packages/<package>-<version>.tar.gz
/// ```
Bucket get packageBucket  => service_scope.... // get instance from service scrope.
```

Or something like that -- I don't recall the actual layout off the top of my head.

But if we write good documentation for the global `packageBucket` getter, then anywhere where you're accessing `packageBucket` you have a good chance find the documentation and check that it aligns with how you are using it.
Then we probably don't need a wrapper that encodes trivial rules.

jonasfj (approved):

PR #5635, opened 2022-04-01
Package publisher redirect route.

### LINKED ISSUE #5633: feature request: support an http redirect from a link for a package, to the page for that package's publisher
Hi - the use case here is that I'd like to be able to link to a package's publisher from the readme file of the package. I won't reliably know ahead of time who the publisher is - the source of record for that is pub.dev, and the publisher for a package can (infrequently) change.

For more context, shields.io now supports a 'publisher' badge for Dart - given a package, it'll return an svg w/ the name of the package's publisher. For example, for package:usage:

Screen Shot 2022-03-31 at 10 23 57 AM

I'd like to hyperlink from that badge to the pub page for the publisher. I currently have to hard-code the publisher link (`https://pub.dev/publishers/tools.dart.dev`). What I'd like to instead is link to a page for the package that would redirect to the publisher page; so a link like `https://pub.dev/packages/usage/publisher` or `https://pub.dev/packages/usage?redirect=publisher` (see also the comment: https://github.com/dart-lang/usage/pull/169#issuecomment-1083865213).

isoos:
FIxes #5633.

### DIFF
--- a/app/lib/frontend/handlers/package.dart
+++ b/app/lib/frontend/handlers/package.dart
@@ -399,3 +399,15 @@ Future<shelf.Response> listVersionsHandler(
     return createResponse(gzip.decode(body), isGzip: false);
   }
 }
+
+/// Handles requests for /packages/<package>/publisher
+Future<shelf.Response> packagePublisherHandler(
+    shelf.Request request, String package) async {
+  checkPackageVersionParams(package);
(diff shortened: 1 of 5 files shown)


devoncarew:
Can you ping this PR (or the above issue) when this is deployed to pub.dev? Thanks!

isoos:
It will be mid-next week, at the earlies.

PR #5636, opened 2022-04-01
Merge TarballStorage into PackageBackend.

### LINKED ISSUE #5632: Migrate away from TarballStorage
This abstractions is a relic of past abstractions, we're better off documenting the bucket layout and just use `Bucket` directly.

isoos:
Fixes #5632

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -62,6 +62,11 @@ PackageBackend get packageBackend =>
 /// Represents the backend for the pub site.
 class PackageBackend {
   final DatastoreDB db;
+
+  /// The Cloud Storage bucket to use for uploaded package content.
+  /// The following files are present:
+  /// - `packages/$package-$version.tar.gz` (package archive)
+  /// - `tmp/$guid` (incoming package archive that was uploaded, but not yet processed)
   final Bucket _bucket;
   final Storage _storage;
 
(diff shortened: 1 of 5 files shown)


app/lib/package/backend.dart
```diff
@@ -70,17 +62,13 @@ PackageBackend get packageBackend =>
...
-
-  PackageBackend(
-    DatastoreDB db,
-    TarballStorage storage, {
-    int? maxVersionsPerPackageOverride,
-  })  : db = db,
-        _storage = storage,
-        _maxVersionsPerPackage =
-            maxVersionsPerPackageOverride ?? maxVersionsPerPackage;
+  final Bucket _bucket;
```
jonasfj, line 65: Let's give this a documentation string, stating the layout of what is stored in here..

jonasfj (approved):

app/test/package/upload_test.dart
```diff
@@ -77,7 +82,9 @@ void main() {
           // Add one more byte than allowed.
           bigTarball.add([1]);
 
-          final sink = tarballStorage.bucket.write('tmp/$uploadId');
+          final bucket =
+              storageService.bucket(activeConfiguration.packageBucketName!);
+          final sink = bucket.write('tmp/$uploadId');
```
jonasfj, line 87: We don't have this in `testWithProfile` the storage service is now exposed on localhost... So we can actually do a multipart upload.

isoos: It was not exposed in the past, but I think this is exposed now, we just don't use it all the time (yet):
https://github.com/dart-lang/pub-dev/blob/master/app/lib/fake/server/fake_storage_server.dart#L37

Adding a TODO to migrate...

PR #5644, opened 2022-04-05
Remove duplicated scope exit callback on HTTP client.

isoos:
Regression introduced in #5629.

### DIFF
--- a/app/lib/service/services.dart
+++ b/app/lib/service/services.dart
@@ -73,7 +73,6 @@ Future<void> withServices(FutureOr<void> Function() fn) async {
       // retrying auth client for storage service
       final authClient = await auth
           .clientViaApplicationDefaultCredentials(scopes: [...Storage.SCOPES]);
-      registerScopeExitCallback(() async => authClient.close());
       final retryingAuthClient = httpRetryClient(innerClient: authClient);
       registerScopeExitCallback(() async => retryingAuthClient.close());
 


app/lib/service/services.dart
```diff
@@ -73,7 +73,6 @@ Future<void> withServices(FutureOr<void> Function() fn) async {
       // retrying auth client for storage service
       final authClient = await auth
           .clientViaApplicationDefaultCredentials(scopes: [...Storage.SCOPES]);
-      registerScopeExitCallback(() async => authClient.close());
```
jonasfj, line 76: This can't be it, can it?

https://pub.dev/documentation/googleapis_auth/latest/auth_io/clientViaApplicationDefaultCredentials.html

```dart
Future<AutoRefreshingAuthClient> clientViaApplicationDefaultCredentials({
  required List<String> scopes,
  Client? baseClient,
}) async {
  if (baseClient == null) {
    baseClient = Client();
  } else {
    baseClient = nonClosingClient(baseClient);
  }
```

isoos: The issue is that we close both the `authClient` and the `retryingAuthClient` (which calls `authClient.close()`). We need to close only one of them.

jonasfj: ah,

jonasfj (approved):

PR #5645, opened 2022-04-05
HTTP Client.close wrapper for retry and authenticated clients.

isoos:
Handles duplicate exit scope callbacks better (#5644).

### DIFF
--- a/app/lib/tool/utils/http.dart
+++ b/app/lib/tool/utils/http.dart
@@ -17,14 +17,14 @@ http.Client httpRetryClient({
   http.Client? innerClient,
   int? retries,
 }) {
-  return RetryClient(
+  return _MultiCloseClient(RetryClient(
     innerClient ?? http.Client(),
     when: (r) => _transientStatusCodes.contains(r.statusCode),
     // TOOD: Consider implementing whenError and handle DNS + handshake errors.
     //       These are safe, retrying after partially sending data is more
     //       sketchy, but probably safe in our application.
     retries: retries ?? 5,
-  );
+  ));
 }
 
 /// Returns an [http.Client] which sends a `Bearer` token as `Authorization`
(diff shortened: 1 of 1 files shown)


app/lib/tool/utils/http.dart
```diff
@@ -17,14 +17,14 @@ http.Client httpRetryClient({
   http.Client? innerClient,
   int? retries,
 }) {
-  return RetryClient(
```
jonasfj, line 20: hmm, when do we use this?

We should probably never use `RetryClient` as I don't think it retries if the connection fails when reading the response, I could be wrong...

Anyways, this is orthogonal.

app/lib/tool/utils/http.dart
```diff
@@ -67,3 +67,24 @@ class _AuthenticatedClient extends http.BaseClient {
     super.close();
   }
 }
+
+/// Allows the call of [close] multiple times without throwing an Exception.
+class _MultiCloseClient extends http.BaseClient {
```
jonasfj, line 72: So the counter argument to doing this is that now you risk one context closing the client used in another context.

When creating an instance of `_AuthenticatedClient` we can specify `_closeInnerClient = false` in the constructor.

app/lib/tool/utils/http.dart
isoos: We have 8 or 9 references already in pub-dev. While it doesn't retry on certain failures, it was still better than using nothing, so I wouldn't remove it unless we have a better solution.

app/lib/tool/utils/http.dart
isoos: Hm, I'm not sure if it is the same to not close the inner client. I agree, ideally, this shouldn't be necessary, but this seems to be slightly better than the obscure failure we were getting.

isoos:
Closing as this seems to be not needed.

PR #5650, opened 2022-04-06
Extra logging in services scope.

isoos:


app/lib/service/services.dart
```diff
@@ -236,8 +238,22 @@ Future<R> _withPubServices<R>(FutureOr<R> Function() fn) async {
...
+      }).run(
+        () async {
+          _logger.info('Starting services scope...');
+          try {
+            return await fn();
+          } catch (e, st) {
+            _logger.severe('Uncaught exception inside services scope.', e, st);
+            rethrow;
+          } finally {
+            _logger.info('Exiting services scope.');
```
jonasfj, line 253: Should this be a warning?

jonasfj (approved):

PR #5651, opened 2022-04-06
Properly detect deployment failures

jonasfj:
Using `#!/bin/bash -e` doesn't work when it is invoked using `bash tool/deploy.sh`.
So it's safer to use `set -e`.

### DIFF
--- a/tool/.deploy.sh
+++ b/tool/.deploy.sh
@@ -1,6 +1,11 @@
-#!/bin/bash -e
+#!/bin/bash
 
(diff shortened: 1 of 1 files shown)


isoos (approved):

jonasfj:
Maybe this wouldn't have been needed if we hadn't done:
https://github.com/dart-lang/pub-dev/blob/411388c7dfed7e474018e345f161a7ffa642b90b/tool/.deploy.yaml#L11-L12

PR #5652, opened 2022-04-06
Fix scope exit and don't close retry client.

isoos:


app/lib/service/services.dart
```diff
@@ -76,7 +76,8 @@ Future<void> withServices(FutureOr<void> Function() fn) async {
       final authClient = await auth
           .clientViaApplicationDefaultCredentials(scopes: [...Storage.SCOPES]);
       final retryingAuthClient = httpRetryClient(innerClient: authClient);
-      registerScopeExitCallback(() async => retryingAuthClient.close());
+      // TODO: investigate why this is called twice
+      // registerScopeExitCallback(() async => retryingAuthClient.close());
```
jonasfj, line 80: Shouldn't we only have to do one of these two things?

isoos:
This seems to be unnecessary for staging. Closing for now.

PR #5653, opened 2022-04-06
Prefer try/finally

jonasfj:


isoos (approved):

app/lib/shared/integrity.dart
```diff
@@ -62,17 +62,20 @@ class IntegrityChecker {
   /// Runs integrity checks, and returns the list of problems.
   Stream<String> findProblems() async* {
     _httpClient = httpRetryClient();
```
kevmoo, line 64: Any concerns that this could be re-entrant?

Not sure how to "lock" that – just a though.

jonasfj: rewrite it to a function instead of misusing a class.

PR #5660, opened 2022-04-11
Use the new incoming bucket for uploads and fallback to the default bucket on reading it.

isoos:
- #5586
- changing only the bucket, the file name pattern and everything else will be the same (makes fallback easier)
- fallback may be needed while the traffic is partially migrated or is being migrated, can be removed after the release is stable

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -62,18 +62,23 @@ PackageBackend get packageBackend =>
 /// Represents the backend for the pub site.
 class PackageBackend {
   final DatastoreDB db;
+  final Storage _storage;
 
   /// The Cloud Storage bucket to use for uploaded package content.
   /// The following files are present:
   /// - `packages/$package-$version.tar.gz` (package archive)
   /// - `tmp/$guid` (incoming package archive that was uploaded, but not yet processed)
   final Bucket _bucket;
-  final Storage _storage;
+
+  /// The Cloud Storage bucket to use for incoming package archives.
+  /// The following files are present:
+  /// - `tmp/$guid` (incoming package archive that was uploaded, but not yet processed)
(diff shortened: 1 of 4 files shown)


app/lib/package/backend.dart
```diff
@@ -62,18 +62,23 @@ PackageBackend get packageBackend =>
 /// Represents the backend for the pub site.
 class PackageBackend {
   final DatastoreDB db;
+  final Storage _storage;
 
   /// The Cloud Storage bucket to use for uploaded package content.
   /// The following files are present:
   /// - `packages/$package-$version.tar.gz` (package archive)
   /// - `tmp/$guid` (incoming package archive that was uploaded, but not yet processed)
```
jonasfj, line 70: Need to update this, right?

app/lib/package/backend.dart
```diff
@@ -679,7 +684,7 @@ class PackageBackend {
     final guid = createUuid();
     _logger.info('Starting semi-async upload (uuid: $guid)');
     final object = tmpObjectName(guid);
-    await data.pipe(_bucket.write(object));
+    await data.pipe(_incomingBucket.write(object));
```
jonasfj, line 687: I guess this is fine... but separately it would be probably be great if we got rid of the `upload` method..

Testing should just upload the fake bucket server, right?

jonasfj (approved):

app/lib/package/backend.dart
isoos: It may be true for the period when the traffic is migrated. Let's update this after this gets released and becomes stable.

app/lib/package/backend.dart
isoos: I'll remove it in a subsequent PR. It is used only in tests anyway...

Issue #5359, opened 2021-12-14
Support dual licensing in UI and possibly in search


isoos:
After https://github.com/dart-lang/pana/issues/997 gets implemented.

isoos:
We'll need to add the identified license files to the output of `pana` (alongside of screenshots or other assets), because the license file identification we do at upload time may not be consistent with the licenses identified later on at analysis time. The `/license` tab should show all of the license files identified during analysis.

PR #5666, opened 2022-04-14
Store and display multi-license information from pana.

isoos:
#5359

### DIFF
--- a/app/lib/frontend/templates/views/pkg/info_box.dart
+++ b/app/lib/frontend/templates/views/pkg/info_box.dart
@@ -34,13 +34,15 @@ d.Node packageInfoBoxNode({
   final version = data.version!;
   d.Node? license;
   if (data.versionInfo?.hasLicense ?? false) {
-    var licenseFile = data.scoreCard?.panaReport?.licenseFile;
-    if (licenseFile != null && data.scoreCard?.panaReport == null) {
-      licenseFile = LicenseFile('LICENSE', 'pending analysis');
+    final licenseFile = data.scoreCard?.panaReport?.licenseFile;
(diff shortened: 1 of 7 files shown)


app/lib/frontend/templates/views/pkg/info_box.dart
```diff
@@ -140,12 +142,14 @@ d.Node _linkAndBr(InfoBoxLink link) {
...
+  required List<License> licenses,
   required String licenseUrl,
 }) {
+  final paths = licenses.map((e) => e.path).toSet().toList();
+  final ids = licenses.map((e) => e.spdxIdentifier).toSet().toList();
   return d.fragment([
-    d.text('${licenseFile.shortFormatted} ('),
-    d.a(href: licenseUrl, text: licenseFile.path),
+    d.text('${ids.join(', ')} ('),
+    d.a(href: licenseUrl, text: paths.join(', ')),
```
jonasfj, line 152: There is only going to be one path here, right? unless we change the detection code... but then won't the URL be wrong?

jonasfj (approved):

app/lib/frontend/templates/views/pkg/info_box.dart
isoos: Unfortunately no, this is an issue. I've commented about this in the attached issue: we need to extract and store the file content via `pana`, otherwise we have a divergence (regardless of detecting one or multiple licenses), as the package reader code extracts only the first candidate and doesn't have any analysis in it. Or, alternatively, we should run the analysis at upload time, and re-run it periodically.

jonasfj: okay, that makes sense...

PR #5667, opened 2022-04-14
Updated publisher landing page.

isoos:


jonasfj (commented):
From staging site:
image

I wonder if we can't do this a little better... like make it wider... or have 9 boxes...

and did we need the "popular packages" heading? Should we change the tab name, and possibly just make the URL `/publisher/dart.dev` instead of `/publisher/dart.dev/packages`

jonasfj:
I guess we should discuss this a bit more more..

isoos:
Closing in favor of #5794.

PR #5676, opened 2022-04-20
Expose tags in score API

### LINKED ISSUE #5665: Make IsFlutterFavorite accessible via pub.dev API
An endpoint to retrieve package score information exists at https://pub.dev/api/packages/{package_name}/score, but the result does not include a field indicating whether the package is a Flutter Favorite. The addition of this field might make the response look something like:

```json
{
  "grantedPoints": 130,
  "maxPoints": 130,
  "likeCount": 6380,
  "popularityScore": 0.999690928759079,
  "lastUpdated": "2022-04-08T17:53:28.592757Z",
  "isFlutterFavorite": true
}
```

isoos:
Fixes #5665.

### DIFF
--- a/app/lib/frontend/handlers/custom_api.dart
+++ b/app/lib/frontend/handlers/custom_api.dart
@@ -231,27 +231,31 @@ Future<VersionScore> packageVersionScoreHandler(
     if (pkg == null) {
       throw NotFoundException.resource('package "$package"');
     }
-    var updated = pkg.updated;
-    final card = await scoreCardBackend.getScoreCardData(package, version);
-
-    // sanity check in case we have no card
-    if (card == null && version != null && version != 'latest') {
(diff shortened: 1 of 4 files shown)


app/lib/frontend/handlers/custom_api.dart
```diff
@@ -251,6 +251,7 @@ Future<VersionScore> packageVersionScoreHandler(
       maxPoints: card?.maxPubPoints,
       likeCount: pkg.likes,
       popularityScore: card?.popularityScore,
+      tags: card?.derivedTags,
```
jonasfj, line 254: Can we exposed both `derivedTags` and `assignedTags`?

isoos: Ah, good point, this doesn't contain the tags coming from the `Package` or `PackageVersion`... will update.

jonasfj (approved):

PR #5692, opened 2022-04-22
Migrate all API clients to HTTP-only in tests.

isoos:
Both keeping the zone and running inside an additional fork/zone was important to make this happen, but this will keep the services context.

### DIFF
--- a/app/lib/service/services.dart
+++ b/app/lib/service/services.dart
@@ -46,6 +46,7 @@ import '../service/youtube/backend.dart';
 import '../shared/configuration.dart';
 import '../shared/datastore.dart';
 import '../shared/env_config.dart';
+import '../shared/handler_helpers.dart';
 import '../shared/popularity_storage.dart';
 import '../shared/redis_cache.dart' show setupCache;
 import '../shared/storage.dart';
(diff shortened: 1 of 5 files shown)


app/lib/tool/utils/pub_api_client.dart
```diff
@@ -2,44 +2,25 @@
...
-    final apiClient = PubApiClient(
-      pubHostedUrl ?? 'http://localhost:0/',
-      client: httpClient,
-    );
-    return await fn(apiClient);
-  } finally {
-    httpClient.close();
-  }
+/// TODO: migrate callers to use [withHttpPubApiClient] instead.
+PubApiClient createLocalPubApiClient({String? authToken}) {
```
jonasfj, line 16: Can we annotate with `@visibleForTestingOnly` ?

Or call it `createTestingPubApiClient`?

It's not local when you use `activeConfiguration.primaryApiUri`

jonasfj (approved):

app/lib/service/services.dart
```diff
@@ -169,8 +170,12 @@ Future<R> withFakeServices<R>({
...
-            .listen((rq) => handleRequest(rq, createAppHandler()));
+        final handler =
+            wrapHandler(_logger, createAppHandler(), sanitize: true);
+        // Handle requests in the current zone, keeping the services context.
+        final currentZone = Zone.current;
+        final frontendServerSubscription = frontendServer.server.listen((rq) =>
+            currentZone.run(() => fork(() => handleRequest(rq, handler))));
         registerScopeExitCallback(frontendServerSubscription.cancel);
       }
       return await fn();
```
jonasfj, line 181: I think the problem is that things might be running after `fn()` returns.

jonasfj (changes requested):

app/lib/tool/utils/pub_api_client.dart
isoos: It may be local, if the current test configuration has a local primary API URI...
The annotation is a good point, added.

Issue #5684, opened 2022-04-21
Likes for some users missing


sigurdm:
We have a production issue.

We estimate  25% of users are affected and missing their "likes".

We are working on restoring these.

jonasfj:
State has been recovered.

PR #5693, opened 2022-04-22
Integrity check and periodic update for Package like counts.

isoos:
- #5684
- integrity check allows a small difference between observed and stored counts
- periodic task will update the counts when it seems there was no update on the Package entity while the scan was running

Note: right now we don't update the `Package.updated` field when we are updating the likes count. We could have better consistency check if we were to update it, or added a non-indexed additional timestamp to track the latest update of the likes count.

### DIFF
--- a/app/lib/shared/integrity.dart
+++ b/app/lib/shared/integrity.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:math' as math;
+
 import 'package:clock/clock.dart';
 import 'package:http/http.dart' as http;
 import 'package:logging/logging.dart';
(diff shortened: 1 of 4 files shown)


app/lib/shared/integrity.dart
```diff
@@ -521,6 +524,36 @@ class IntegrityChecker {
...
+      final updatedStored = p!.likes;
+      if (counted == updatedStored) {
+        continue;
+      }
+
+      final diff = counted - updatedStored;
+
+      // Allowing some difference to attribute for the likes created or removed
+      // between the package reads and the current counts.
+      if (diff.abs() <= 2) {
```
jonasfj, line 552: Let's consider allowing a difference of `max(3, counted * 0.10)`, so as to allow 10% inaccuracy.

jonasfj (approved):

PR #5696, opened 2022-04-25
Check if publisher transfer operation is idempotent.

isoos:


app/lib/package/backend.dart
```diff
@@ -549,6 +549,9 @@ class PackageBackend {
     final rs = await withRetryTransaction(db, (tx) async {
       final package = await db.lookupValue<Package>(key);
       final fromPublisherId = package.publisherId;
+      if (fromPublisherId == request.publisherId) {
+        return _asPackagePublisherInfo(package);
+      }
```
jonasfj, line 554:
```suggestion
      if (package.publisherId == request.publisherId) {
        // If desired publisherId is already the current publisherId, then we're already done.
        return _asPackagePublisherInfo(package);
      }
      final currentPublisherId = package.publisherId;
```

`currentPublisherId` might be better than `fromPublisherId` -- the other thing read a little concerning :D

jonasfj (approved):

jonasfj (approved):
r+

PR #5701, opened 2022-04-26
Verify double-slash URLs in fake server and tests.

isoos:


app/lib/fake/server/fake_default_service.dart
```diff
@@ -51,6 +51,9 @@ class FakePubServer {
 
           final server = await IOServer.bind('localhost', port);
           serveRequests(server.server, (request) async {
+            if (request.requestedUri.path.contains('//')) {
+              return shelf.Response.badRequest();
+            }
```
jonasfj, line 56: Let's not deviate from what the real server does, not when we don't have to... this could cause issues elsewhere.

In general an extra slash..

pkg/_pub_shared/lib/validation/html/html_validation.dart
```diff
@@ -74,6 +74,20 @@ void validateHtml(Node root) {
     }
   }
 
+  // All <a href=""> links must be valid and have only single slashes.
+  for (final elem in links) {
```
jonasfj, line 78: This will also be testing external links create by packages right, they can legitimately have double slashes...

pkg/pub_integration/lib/src/headless_env.dart
```diff
@@ -131,6 +131,11 @@ class HeadlessEnv {
         return;
       }
 
+      final uri = Uri.parse(rq.url);
+      if (uri.path.contains('//')) {
+        serverErrors.add('Double-slash URL detected: "${rq.url}".');
+      }
```
jonasfj, line 137: Why is this always wrong?

jonasfj (approved):

pkg/_pub_shared/lib/validation/html/html_validation.dart
jonasfj: Just saying, I think we don't need this...

isoos:
All of the assertions here are running only in our local tests, only the single integration test will try to load the newly published page from `pub.dev`, but even there we are in control of what goes on in the readme and links. I think it is fair to say that we shouldn't generate any link that has double slash, and checking for it seems to be not bad. Maybe the fake_server request check is a bit deviating from the real server with it, and we should remove that part, but the others I think are solid (and btw. they haven't trigger any test to fail).

jonasfj:
Okay, I'm down with keeping the other checks...

PR #5703, opened 2022-04-26
Update survey-helper.js

jayoung-lee:
I figured that the HaTS survey is triggered too often, and the reason is because the rate limiting was not working. I learned that we should set `enableTestingMode` to `false` from [this documentation](https://g3doc.corp.google.com/company/teams/uservoice/surveys/client/concepts.md?cl=head#testing-mode).

### DIFF
--- a/static/js/survey-helper.js
+++ b/static/js/survey-helper.js
@@ -14,7 +14,7 @@
   });
   helpApi.requestSurvey({
     triggerId: data.triggerId,
-    enableTestingMode: data.testingMode === 'true',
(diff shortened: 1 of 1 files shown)


jonasfj:
I don't see any difference here:
image

It should already be set to false. This is a screenshot using the chrome debugger by setting a break point.

jonasfj (changes requested):
I don't see what this changes.

jonasfj:
Please reopen if I missed anything.

PR #5712, opened 2022-04-28
Remove self(-update) link from osd.xml

isoos:
- The self-link is broken as the `/static/hash-[hash]` URLs are dynamically generated.
- The self-link is not easy to override, as the hash is generated based on the static files' content. Simpler to just remove this self-reference, which seem to be breaking it, but is not necessary at all.


jonasfj (approved):

isoos:
Closing as this is not the fix we want here.

PR #5714, opened 2022-04-28
Redirect path-hashed static URLs to the latest hash.

isoos:


app/lib/frontend/handlers/misc.dart
```diff
@@ -139,6 +139,10 @@ Future<shelf.Response> staticsHandler(shelf.Request request) async {
   final parsed = ParsedStaticUrl.parse(request.requestedUri);
   final staticFile = staticFileCache.getFile(parsed.filePath);
   if (staticFile != null) {
+    final pathHash = parsed.pathHash;
+    if (pathHash != null && pathHash != staticFileCache.etag) {
+      return redirectResponse(staticFile.pathHashedUrl(staticFileCache.etag));
```
jonasfj, line 144: Can we leave a comment saying why this is a good idea?

Also how will this play out doing traffic migration?

jonasfj (approved):

PR #5719, opened 2022-04-29
Fix and extend static files test.

isoos:
- follow-up to #5714.
- re-enabled skipped tests in CI, which were disabled in #5605
- fixed reference checks + added file existence checks via static handler requests
- added hash-path + caching tests

### DIFF
--- a/app/mono_pkg.yaml
+++ b/app/mono_pkg.yaml
@@ -8,14 +8,13 @@ stages:
         - format
         - analyze: --fatal-infos  .
   - unit_test:
-    - test: --total-shards 7 --shard-index 0
-    - test: --total-shards 7 --shard-index 1
-    - test: --total-shards 7 --shard-index 2
-    - test: --total-shards 7 --shard-index 3
-    - test: --total-shards 7 --shard-index 4
-    - test: --total-shards 7 --shard-index 5
-    - test: --total-shards 7 --shard-index 6
-    - test:  -t presubmit-only --run-skipped
+    - test: -P presubmit --total-shards 7 --shard-index 0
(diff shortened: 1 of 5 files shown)


app/mono_pkg.yaml
```diff
@@ -15,6 +15,7 @@ stages:
     - test: --total-shards 7 --shard-index 4
     - test: --total-shards 7 --shard-index 5
     - test: --total-shards 7 --shard-index 6
+    - test:  -t presubmit-only --run-skipped
```
jonasfj, line 18: Let's not run skipped, if something is skipped, it's skipped for a reason.

If want to filter enable/disable tests then make tags and let the test tags be disable by default and then explicitly enable the tags you want to run in CI.
See what I did for `fragile` in `dart_test.yaml`, see also:
https://github.com/dart-lang/test/blob/master/pkgs/test/doc/configuration.md#presets

It's not intuitive :see_no_evil:, but I think these tests can even be part of the sharded stuff, no need to create it's own thing...

```yaml
    - test: -P presubmit --total-shards 7 --shard-index 6
```

Then do `dart_test.yaml` as:
```yaml

  
tags:
  presubmit-only:
    skip: "Should only be run during presubmit"
    presets:
      presubmit:
        skip: false # Don't skip when running in presubmit
  fragile:
    skip: 'Only run fragile tests manually with `pub run test -P fragile`'
    presets:
      fragile:
        skip: false # Don't skip when running in -P fragile
```
I think that should work.

isoos: Updated the presubmit preset. I wasn't aware of it, test configs can indeed be non-trivial :)

jonasfj (approved):

PR #5726, opened 2022-05-02
Improve latency for first render: async gtm.js, lazy loading of some images.

### LINKED ISSUE #5724: Use loading="lazy" on images that are not on-screen initially
https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading#images_and_iframes

isoos:
Fixes #5724.

### DIFF
--- a/app/lib/frontend/templates/views/shared/layout.dart
+++ b/app/lib/frontend/templates/views/shared/layout.dart
@@ -38,7 +38,10 @@ d.Node pageLayoutNode({
               src: 'https://www.googletagmanager.com/gtm.js?id=GTM-MX6DBN9',
               async: true,
             ),
-            d.script(src: staticUrls.getAssetUrl('/static/js/gtm.js')),
+            d.script(
+              src: staticUrls.getAssetUrl('/static/js/gtm.js'),
+              async: true,
+            ),
             d.meta(charset: 'utf-8'),
             d.meta(httpEquiv: 'x-ua-compatible', content: 'ie=edge'),
             d.meta(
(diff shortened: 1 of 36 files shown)


app/lib/frontend/templates/views/shared/layout.dart
```diff
@@ -38,7 +38,10 @@ d.Node pageLayoutNode({
               src: 'https://www.googletagmanager.com/gtm.js?id=GTM-MX6DBN9',
               async: true,
             ),
-            d.script(src: staticUrls.getAssetUrl('/static/js/gtm.js')),
+            d.script(
+              src: staticUrls.getAssetUrl('/static/js/gtm.js'),
+              async: true,
+            ),
```
jonasfj, line 44: Can we come up with a basic test we want to do? And put that into `CHANGELOG.md`? (just an idea)

Maybe, just leave a note that we should run with GTM debugging to make it still works.


Also, I'm not sure if it's important that `dataLayer` is available before `https://www.googletagmanager.com/gtm.js?....` is loaded... I'm guessing probably not, since it probably replaces the `dataLayer` object (just guessing).

jonasfj (approved):

app/lib/frontend/templates/views/shared/layout.dart
isoos: I think it should be relatively easy to check if e.g. copy-to-clipboard events are collected and reported, added that to the changelog.

`dataLayer` initialization seems to be a common way for such trackers, so I think it doesn't really matter in which order they are loaded. We could bring our small `gtm.js` in the front if you think it is better.

Issue #4373, opened 2020-12-22
Add "Filter by license"


jayoung-lee:
Add “Filter by license” to the search results page for the developers working on commercial apps and want a permissive license (e.g., BSD, MIT). In a recent users study some users mentioned that they won't event want to look at packages that miss license information.

jayoung-lee:
We can consider exposing the license information to the search result, until SPDX license detection becomes available.

ChristianKleineidam:
When it comes to Filter by license, GPL style licenses would also be a category that exist and that people might want to filter out.

jonasfj:
We actually have a design for this now:

![image](https://user-images.githubusercontent.com/149732/152321664-499b162c-90aa-4efc-ae2b-8e1ff0720be8.png)

Should probably be spelled "License".

Options should probably be:
 * OSI Approved
 * More

With "More" expanding to a [subset of the licenses](https://github.com/dart-lang/pana/tree/master/lib/src/third_party/spdx/licenses) we can detect, or maybe just a high-level family of licenses.

Perhaps we can do an offline grep to identify common licenses and hard code the most popular licenses.


-------------------


Analyzing licenses using unstable API end-points:
```bash
get_license() { 
  curl -s "https://pub.dev/api/packages/$1/metrics" | jq .scorecard.panaReport.licenseFile.name -r ;
}
export -f get_license

curl -s https://pub.dev/api/package-names | jq .packages[] -r | parallel -j 50 get_license > /tmp/detected-licenses

cat /tmp/detected-licenses | sort | uniq -c | sort -n -r
```

Yields a list of licenses as follows:
```
  11176 MIT
   4378 unknown
   4351 null
   4124 BSD-3-Clause
   2116 Apache-2.0
    949 BSD-2-Clause
    420 GPL-3.0
     83 MPL-2.0
     73 LGPL-3.0
     69 Unlicense
     58 AGPL-3.0
     25 BSD-2-Clause-Views
     21 LGPL-2.1
      7 WTFPL
      7 EUPL-1.2
      6 CPL-1.0
      6 CC0-1.0
      5 OFL-1.1
      4 Zlib
      4 BSL-1.0
      3 EPL-2.0
      3 CC-BY-4.0
      3 Artistic-2.0
      2 OpenSSL
      2 MulanPSL-2.0
      2 MulanPSL-1.0
      2 Hippocratic-2.1
      2 CC-BY-SA-4.0
      2 AFL-3.0
      1 X11
[...]

PR #5728, opened 2022-05-02
Filter UI for OSI approved license.

isoos:
image

#4373


isoos:
As discussed, updated to have only the OSI approved checkbox for now, and we shall add other filters after a bit of experimenting...

jonasfj (approved):

Issue #5566, opened 2022-03-09
disable end-point used by `dart pub uploader add/remove`


jonasfj:
We should disable end-point used by `dart pub uploader add/remove`, and make it return a message that says to use the web-interface.

However, let's not rush to do this before Dart 2.17 when `dart pub uploader` is removed from the Dart SDK. I propose that we do this a few months after, when many people have upgraded.

isoos:
The UI-based invite uploader endpoint differs from the CLI's add uploader, but they share the remove uploader endpoint. Should we introduce a new one for the remove?

PR #5729, opened 2022-05-02
New remove-uploader endpoint for UI-only use.

isoos:
#5566

### DIFF
--- a/app/lib/frontend/handlers/pubapi.dart
+++ b/app/lib/frontend/handlers/pubapi.dart
@@ -165,6 +165,18 @@ class PubApi {
   ) async =>
       await packageBackend.removeUploader(package, email);
 
+  /// Remove an existing uploader.
+  ///
+  /// This is a different API endpoint than the pub CLI tool is using, because
+  /// we want to use a separate authentication mechanism going forward.
+  @EndPoint.post('/api/packages/<package>/remove-uploader')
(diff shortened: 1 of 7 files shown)


app/lib/frontend/handlers/pubapi.dart
```diff
@@ -165,6 +165,18 @@ class PubApi {
   ) async =>
       await packageBackend.removeUploader(package, email);
 
+  /// Remove an existing uploader.
+  ///
+  /// This is a different API endpoint than the pub CLI tool is using, because
+  /// we want to use a separate authentication mechanism going forward.
```
jonasfj, line 171: Why does that mean we need a new end-point?

isoos: I first thought that the authentication migration will change the credentials that will hit the endpoint (sessionId cookie + CSRF token), but now I've realized that we'll need to support both at the same time, so this doesn't necessarily need a new endpoint.

However, deprecating the pub tool endpoint (#5566) requires a new endpoint for the UI, and moving the email to the payload is more similar to the invite-uploader method.

jonasfj: Okay, I'm down with that...

jonasfj (approved):

PR #5731, opened 2022-05-03
Display license on the search result page.

isoos:
- #4373
- Also updated the package page's infobox label to use the same nicely formatted license text (at least for some of the common licenses).

image


app/lib/frontend/templates/views/pkg/license.dart
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:pub_dev/frontend/static_files.dart';
+
+import '../../../dom/dom.dart' as d;
+
+/// Nicely formatted labels for common SPDX identifiers.
```
jonasfj, line 9: Why, is there not a quality to using identifiers? I don't really mind either way.

app/lib/frontend/templates/views/pkg/license.dart
```diff
@@ -0,0 +1,46 @@
...
+String spdxLabel(String id) => _spdxLabels[id] ?? id;
+
+d.Node? packageListMetadataLicense(List<String>? spdxIdentifiers) {
+  if (spdxIdentifiers == null || spdxIdentifiers.isEmpty) {
+    return null;
+  }
+  final more = spdxIdentifiers.length - 1;
+  final label = [
+    spdxLabel(spdxIdentifiers.first),
+    if (more > 0) '+$more',
```
jonasfj, line 29: I'm a tiny bit worried how this can be interpreted, like what does `LGPL 2.1 +1` actually mean? 

Could someone read it as `LGPL 2.1+` and think it means `LGPL 2.1 or later`?

app/lib/frontend/templates/views/pkg/license.dart
isoos: I find it visually better (easier to read and parse) to have a space a lowercase starting letter in `BSD 3-clause` vs. `BSD-3-Clause`.

app/lib/frontend/templates/views/pkg/license.dart
isoos: hm... yeah, maybe we should just display all of it....

mit-mit:
Looks great, thanks!

app/lib/frontend/templates/views/pkg/license.dart
jonasfj: Or use 3 dots...

app/lib/frontend/templates/views/pkg/license.dart
isoos: For the search page results, I wouldn't put the raw identifiers there, as it is not indexed by Google, it is only read by humans.
For the package page, we could put the raw ids somewhere, maybe in a `title=""` tag perhaps?

mit-mit: Right, what I meant was that of someone copies the text from the search results and then runs that through Google search, I feel more confident that they'll get the right page if the text uses the SPDX identifier.

jonasfj (approved):

jonasfj (approved):

PR #5738, opened 2022-05-06
Reduce test fragility on fake pana analysis: consistent hashing.

isoos:
- The random in current fake pana analysis is fragile, because if we change any of it (adding/removing items), everything after that may change, as the next random int will be different.
- Instead, each component of the analysis should be "randomized" on its own, and this is now done with this consistent hashing of the package+version+field.
- Updated test to reflect this change, and also modified tests to become less fragile in the future, if any of it may change.

### DIFF
--- a/app/lib/fake/backend/fake_pana_runner.dart
+++ b/app/lib/fake/backend/fake_pana_runner.dart
@@ -2,8 +2,9 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:math';
+import 'dart:convert';
 
+import 'package:crypto/crypto.dart';
 import 'package:pana/pana.dart';
 import 'package:pub_semver/pub_semver.dart';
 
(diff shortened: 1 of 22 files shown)


jonasfj (approved):

app/lib/fake/backend/fake_pana_runner.dart
```diff
@@ -112,3 +114,26 @@ class FakePanaRunner implements PanaRunner {
...
+  final String version;
+  final _keys = <String>{};
+
+  _HashGenerator(this.package, this.version);
+
+  /// Returns the hash of the [key]. When [max] is present, only
+  /// ints between 0 and max (exclusive) will be returned.
+  ///
+  /// Throws [StateError] if it is called more than once with teh same [key].
+  int hashOf(String key, [int? max]) {
```
jonasfj, line 129: This is a bit of a hack, but works... maybe we can find a better name than `_HashGenerator` but that's all..
Or make functions:
```

int Function(String input, {int? max}) createHasher(String seed) => (input, {int? max}) {
   ....
}
```

PR #5751, opened 2022-05-11
Publisher.isBlocked field to prevent blocked/deleted publishers from reuse.

### LINKED ISSUE #5718: Implement blocking of publishers
Our current moderation tool for publishers deletes them.

That doesn't prevent recreation.

We should have a "blocked" flag on publishers preventing them from publishing and showing up.

isoos:
- Fixes #5718.
- Flag is applied at lookup, query and also update operations.
- Added backfill, but it can be used with the default value right away.
- Refactored membership and admin role tests to also include the blocked verification while the access right is being checked.

### DIFF
--- a/app/lib/admin/tools/block_publisher_and_all_members.dart
+++ b/app/lib/admin/tools/block_publisher_and_all_members.dart
@@ -46,9 +46,6 @@ Future<String> executeBlockPublisherAndAllMembers(List<String> args) async {
       final p = await tx.lookupValue<Publisher>(publisherKey);
       p.markForBlocked();
       tx.insert(p);
-      for (final m in members) {
-        tx.delete(publisherKey.append(PublisherMember, id: m.userId));
-      }
     });
     output.writeln('Blocked.');
     return output.toString();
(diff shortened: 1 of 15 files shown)


isoos:
One thing I've considered, but could not really decide: should we name it `isDeleted` instead? /cc @sigurdm

app/lib/admin/tools/block_publisher_and_all_members.dart
```diff
@@ -37,19 +36,21 @@ Future<String> executeRemovePublisherAndBlockAllMembers(
...
     }
 
     final publisherKey = dbService.emptyKey.append(Publisher, id: publisherId);
     await withRetryTransaction(dbService, (tx) async {
-      tx.delete(publisherKey);
+      final p = await tx.lookupValue<Publisher>(publisherKey);
+      p.markForBlocked();
+      tx.insert(p);
       for (final m in members) {
         tx.delete(publisherKey.append(PublisherMember, id: m.userId));
```
jonasfj, line 50: Is this a good idea, that makes it kind of non-reversible.

jonasfj (approved):

app/lib/admin/tools/block_publisher_and_all_members.dart
isoos: Hm, I wanted to do it in a follow-up change, but since we already change the admin tool name, this could be also changed here.

PR #5755, opened 2022-05-13
Integrity check for userId + robot agent(s).

isoos:


app/lib/account/agent.dart
```diff
@@ -0,0 +1,35 @@
...
+};
+
+/// Whether the [userId] is valid-looking,
+/// without namespace or other special value.
+bool isValidUserId(String userId) {
+  return _uuidRegExp.matchAsPrefix(userId) != null;
+}
+
+/// Whether the [agent] is a valid-looking identifier.
+bool isValidRobotAgent(String agent) {
```
jonasfj, line 20:
```suggestion
bool isKnownServiceAgent(String agent) {
```

app/lib/shared/integrity.dart
```diff
@@ -456,7 +464,7 @@ class IntegrityChecker {
     if (pv.uploader == null) {
       yield 'PackageVersion "${pv.qualifiedVersionKey}" has no uploader.';
     } else {
-      yield* _checkUserValid(
+      yield* _checkAgentValid(
         pv.uploader!,
```
jonasfj, line 468: We **critically** need to update documentation for this property.
All properties should specify what they contain, and what shape they have, when they may contain one thing and when they may contain something else..

Reading the model should provide sufficient documentation for writing the integrity check. It's always the model we look at when we write code -- we don't want to have to jump around the code to find other places.

isoos:
PTAL! Updated the documentation of both fields involved (the `AuditLogRecord.agent` and the `PackageVersion.uploader`).

jonasfj (approved):

PR #5764, opened 2022-05-17
Credential-less publishing: minimal UI, API and data model.

isoos:
- I've opted to do a separate API endpoint, as we will probably need to add more verification steps, and it is likely that we don't want to make the information public, vs. the other flags like `isDiscontinued`.
- However, reusing the options updated audit log record.
- The data and UI is minimal at this point: only using a single repository URL.


isoos:
Closing the PR as it is far from ready.

PR #5766, opened 2022-05-17
Credential-less publishing: minimal UI, API and data model.

isoos:
- A slightly different take than #5764.
- Created a new API endpoint to updated it, as we will probably need more verification as we go, and it is likely that we don't want to make the information public, vs. the other flags like `isDiscontinued`.
- However, reusing the options updated audit log record.
- The data and UI is minimal at this point: only using a single projectPath + the enabled checkbox.

### DIFF
--- a/app/lib/frontend/templates/views/pkg/admin_page.dart
+++ b/app/lib/frontend/templates/views/pkg/admin_page.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:pub_dev/frontend/request_context.dart';
+
 import '../../../../account/models.dart';
 import '../../../../package/models.dart';
 import '../../../../shared/urls.dart' as urls;
@@ -159,6 +161,35 @@ d.Node packageAdminPageNode({
         checked: package.isUnlisted,
       ),
     ],
+    if (requestContext.showAdminUIForCredentiallessPublishing)
+      d.fragment([
+        d.h2(text: 'Credential-less publishing'),
+        d.h3(text: 'GitHub'),
+        d.div(
+          classes: ['-pub-form-row'],
+          child: material.checkbox(
+            id: '-pkg-admin-credless-github-enabled',
+            label: 'Enable publishing from GitHub',
+            checked:
+                package.credentiallessPublishing.github?.isEnabled ?? false,
+          ),
+        ),
+        d.div(
+          classes: ['-pub-form-row'],
+          child: material.textField(
+            id: '-pkg-admin-credless-github-projectpath',
+            label: 'Project path',
+            value: package.credentiallessPublishing.github?.projectPath,
+          ),
+        ),
+        d.p(
+          child: material.button(
+            id: '-pkg-admin-credless-button',
+            label: 'Update',
+            raised: true,
+          ),
+        ),
+      ]),
     d.h2(text: 'Package Version Retraction'),
     d.div(children: [
       d.markdown(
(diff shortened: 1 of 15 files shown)


app/lib/package/backend.dart
```diff
@@ -50,6 +50,8 @@ final maxAssetContentLength = 128 * 1024;
 final _defaultMaxVersionsPerPackage = 1000;
 
 final Logger _logger = Logger('pub.cloud_repository');
+final _validGithubUserOrRepoRegExp =
```
sigurdm, line 53: Do we have a source for this regexp?

sigurdm (approved):

app/lib/frontend/templates/views/pkg/admin_page.dart
```diff
@@ -159,6 +161,35 @@ d.Node packageAdminPageNode({
         checked: package.isUnlisted,
       ),
     ],
+    if (requestContext.showAdminUIForCredentiallessPublishing)
+      d.fragment([
+        d.h2(text: 'Credential-less publishing'),
```
jonasfj, line 166:
```suggestion
        d.h2(text: 'Automated Publishing'),
```

app/lib/frontend/templates/views/pkg/admin_page.dart
```diff
@@ -159,6 +161,35 @@ d.Node packageAdminPageNode({
...
+            label: 'Enable publishing from GitHub',
+            checked:
+                package.credentiallessPublishing.github?.isEnabled ?? false,
+          ),
+        ),
+        d.div(
+          classes: ['-pub-form-row'],
+          child: material.textField(
+            id: '-pkg-admin-credless-github-projectpath',
+            label: 'Project path',
```
jonasfj, line 181:
```suggestion
            label: 'Repository',
```
Maybe, written as:
> Organization / Repository (e.g. `dart-lang/sdk`):

app/lib/package/backend.dart
jonasfj: We could do:
https://github.com/shinnn/github-username-regex

But it might be fine to just check that the `<org>/<repo>` string only contains a single slashes, and no `#` or something like that... and that neither `<org>` or `<repo>` is empty.

We don't really need to be super careful here, since github will only sign JWTs with the correct values.

app/lib/package/backend.dart
```diff
@@ -445,6 +447,54 @@ class PackageBackend {
...
+        final projectPath = github.projectPath?.trim() ?? '';
+        github.projectPath = projectPath;
+
+        InvalidInputException.check(!isEnabled || projectPath.isNotEmpty,
+            'The `projectPath` field must not be empty when enabled.');
+
+        if (projectPath.isNotEmpty) {
+          final parts = projectPath.split('/');
+          InvalidInputException.check(parts.length == 2,
+              'The `projectPath` field must follow the `<user>/<repository>` pattern.');
```
jonasfj, line 477:
```suggestion
              'The `organization/repository` field must follow the `<user>/<repository>` pattern.');
```

app/lib/package/backend.dart
```diff
@@ -445,6 +447,54 @@ class PackageBackend {
...
+      }
+
+      // finalize changes
+      p.credentiallessPublishing = body;
+      p.updated = clock.now().toUtc();
+      tx.insert(p);
+      tx.insert(AuditLogRecord.packageOptionsUpdated(
+        package: p.name!,
+        user: user,
+        options: ['credentialless-publishing'],
```
jonasfj, line 492:
```suggestion
        options: ['automated-publishing'],
```


I think we should make a separate audit-log entry, I think we talked about calling it `package-publication-authentication-changed` or something like it, see document..

I think the summary should include what org/repository is allowed to publish...

app/lib/package/backend.dart
isoos: If I understand the regexp correctly, we may be enabling slightly more than github allows, e.g. `-user/...`. I don't think it is super-important to get this exactly right, we only need to filter out obviously bad input.

PR #5779, opened 2022-05-24
Check oauth token audience and log them if they are not for the current scope.

isoos:


jonasfj (approved):

app/lib/package/backend.dart
```diff
@@ -1169,9 +1177,13 @@ class PackageBackend {
...
   Future<api.SuccessMessage> removeUploader(
-      String packageName, String uploaderEmail) async {
+    String packageName,
+    String uploaderEmail, {
+    AuthSource? authSource,
+  }) async {
+    authSource ??= AuthSource.client;
     uploaderEmail = uploaderEmail.toLowerCase();
-    final user = await requireAuthenticatedUser();
+    final user = await requireAuthenticatedUser(source: authSource);
```
jonasfj, line 1186: Can the client trigger this anymore?

jonasfj (approved):

app/lib/package/backend.dart
isoos: Old clients still can, we are not actively blocking on it yet.

Issue #5769, opened 2022-05-19
Allow publishing from github


sigurdm:

sigurdm:
Seems to be fixed now!

rrousselGit:
Awesome! Is there anywhere we can go to read about how to set it up?

AlexV525:
> Awesome! Is there anywhere we can go to read about how to set it up?

https://dart.dev/tools/pub/automated-publishing might work.

rrousselGit:
This is great, thanks!

liodali:
is there any documentation how we can setup it or used to publish flutter packages ?

PR #5793, opened 2022-05-26
Refactor package upload method.

isoos:
- In preparation for #5769.
- Removed `tarballUpload` indirection in `_performTarballUpload`.
- Group pre-transaction name checks in `_verifyPackageName`.
- Moved some in-transaction checks outside (when it doesn't affect the correctness: upload restrictions, reserved names).
- Reordered a few items.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -872,6 +872,13 @@ class PackageBackend {
     });
   }
 
+  /// Verify the package name defined in the newly uploaded archive file,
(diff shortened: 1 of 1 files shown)


app/lib/package/backend.dart
```diff
@@ -887,6 +872,40 @@ class PackageBackend {
     });
   }
 
+  Future<void> _verifyPackageName({
```
jonasfj, line 875: Can it have a small documentation comment?

jonasfj (approved):

PR #5795, opened 2022-05-27
Fetch JSON web keys from GitHub.

isoos:
#5769


app/lib/service/openid/jwks.dart
```diff
@@ -0,0 +1,97 @@
...
+  /// The thumbprint of the x.509 cert (SHA-1 thumbprint).
+  final String? x5t;
+
+  /// The modulus for the RSA public key.
+  @NullableUint8ListBase64Converter()
+  final Uint8List? n;
+
+  /// The exponent for the RSA public key.
+  @NullableUint8ListBase64Converter()
+  final Uint8List? e;
```
jonasfj, line 54: We can also keep it as a base64 encoded string, and create a `extension JsonWebKeyExt on JsonWebKey` with `Uint8List get nAsBytes`

But I'm also fine with this :D

app/lib/service/openid/jwks.dart
```diff
@@ -0,0 +1,97 @@
...
+      }
+      return base64.decode(json);
+    }
+    return null;
+  }
+
+  @override
+  String? toJson(Uint8List? object) {
+    if (object != null) {
+      return base64.encode(object);
```
jonasfj, line 93: Violates https://datatracker.ietf.org/doc/html/rfc7517#section-1.1, which says:
> BASE64URL(OCTETS) denotes the base64url encoding of OCTETS, per Section 2 of [[JWS](https://datatracker.ietf.org/doc/html/rfc7517#ref-JWS)].

Following JWS to https://www.rfc-editor.org/rfc/rfc7515.html#appendix-C, we get:
```c#
     static string base64urlencode(byte [] arg)
     {
       string s = Convert.ToBase64String(arg); // Regular base64 encoder
       s = s.Split('=')[0]; // Remove any trailing '='s
       s = s.Replace('+', '-'); // 62nd char of encoding
       s = s.Replace('/', '_'); // 63rd char of encoding
       return s;
     }
```

Documentation comment should probably reference these specifications.

app/lib/service/openid/jwks.dart
```diff
@@ -0,0 +1,97 @@
...
+    required this.e,
+  });
+
+  factory JsonWebKey.fromJson(Map<String, dynamic> json) =>
+      _$JsonWebKeyFromJson(json);
+
+  Map<String, dynamic> toJson() => _$JsonWebKeyToJson(this);
+}
+
+/// Converts bytes to BASE64-encoded String (nullable values).
```
jonasfj, line 73: Which base64 encoding, any?

Reference -> https://www.rfc-editor.org/rfc/rfc7515.html#appendix-C

Maybe we should call `NullableUint8ListToUrlSafeUnpaddedBase64Converter` :rofl:

app/lib/service/openid/jwks.dart
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  factory JsonWebKeyList.fromJson(Map<String, dynamic> json) =>
+      _$JsonWebKeyListFromJson(json);
+
+  Map<String, dynamic> toJson() => _$JsonWebKeyListToJson(this);
+}
+
+/// The JSON Web Key record.
+@JsonSerializable(includeIfNull: false, explicitToJson: true)
```
jonasfj, line 28: Need to reference the specification:
https://datatracker.ietf.org/doc/html/rfc7517
And list of registered parameters:
https://www.iana.org/assignments/jose/jose.xhtml#web-key-parameters

I would suggest supporting all possible parameters.


See also: https://github.com/google/webcrypto.dart/blob/master/lib/src/jsonwebkey.dart
We shouldn't use that, it's intended for communication with webcrypto in the browser :D

app/lib/service/openid/jwks.dart
```diff
@@ -0,0 +1,97 @@
...
+  final String? use;
+
+  // The unique identifier for the key.
+  final String? kid;
+
+  /// The family of cryptographic algorithms used with the key.
+  final String? kty;
+
+  /// The x.509 certificate chain. The first entry in the array is the certificate to use for token verification; the other certificates can be used to verify this first certificate.
+  final List<String>? x5c;
```
jonasfj, line 43: Fun, reading https://datatracker.ietf.org/doc/html/rfc7517#section-4.7 it seems this is not _unpadding url-safe base64_

Just interesting to know :D

app/lib/shared/redis_cache.dart
```diff
@@ -274,6 +275,17 @@ class CachePatterns {
         encode: (bool value) => value,
         decode: (d) => d as bool,
       ))['$service/$package'];
+
+  /// Stores the Github OpenID JSON Web Key list.
+  Entry<JsonWebKeyList> githubOpenIdWebKeyList() => _cache
+      .withPrefix('github-openid-web-key-list/')
+      .withTTL(Duration(minutes: 15))
```
jonasfj, line 282: Sad to see github responses doesn't carry a `Cache-Control` header allowing us to cache the response.
We should perhaps note that we are aware and decide to cache it for 15 minutes anyways. Just a comment.

app/lib/service/openid/jwks.dart
isoos: So we can just use `base64Url.encode()` instead.... we need to remove the trailing `=`s though...

jonasfj: Yes :D

I was just quoting the spec..

app/lib/service/openid/github_jwks.dart
```diff
@@ -0,0 +1,38 @@
...
+    try {
+      final configUri = Uri.parse(
+          'https://token.actions.githubusercontent.com/.well-known/openid-configuration');
+      final rs1 = await client.get(configUri);
+      if (rs1.statusCode != 200) {
+        throw Exception(
+            'Unexpected status code ${rs1.statusCode} while fetching $configUri');
+      }
+      final map1 = json.decode(rs1.body) as Map<String, dynamic>;
+      final jwksUri = Uri.parse(map1['jwks_uri'] as String);
```
jonasfj, line 25: may I propose we make an `OpenIDProviderConfiguration` or `OpenIdProviderMetadata` object too?

If we look at what github supplies:
```js
{
  // We should require that "issuer" matches the "iss" claim in the signed JWTs
  // we could even require that the issuer in this document matches something hardcoded in our code...
  "issuer": "https://token.actions.githubusercontent.com",
  // let's sanity check that jwks_uri is an HTTPS URL :D 
  "jwks_uri": "https://token.actions.githubusercontent.com/.well-known/jwks",
  "subject_types_supported": [
    "public",
    "pairwise"
  ],
  "response_types_supported": [
    "id_token" // we should require that this includes "id_token"
  ],
  "claims_supported": [
    // We should reject any JWT with claims not listed here:
    "sub",
    "aud",
    "exp",
    "iat",
    "iss",
    "jti",
    "nbf",
    "ref",
    "repository",
    "repository_id",
    "repository_owner",
    "repository_owner_id",
    "run_id",
    "run_number",
    "run_attempt",
    "actor",
    "actor_id",
    "workflow",
    "head_ref",
    "base_ref",
    "event_name",
    "ref_type",
    "environment",
    "job_workflow_ref"
  ],
  "id_token_signing_alg_values_supported": [
    // we should:
    //  * reject any JWK for other algorithms,
    //  * reject any JWT for other algorithms,
    "RS256"
  ],
  "scopes_supported": [
    "openid" // I guess we could require this includes "openid", spec says so: https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata
  ]
}
```

We can probably just make a generic structure to decode:
https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

And then a method / extension method to validate this against the specification.

And then a function to check that a given `OpenIDProviderConfiguration` object satisfies our hardcoded requirements for github actions.

Maybe we can create a `JwtValidator` object that takes a `OpenIDProviderConfiguration` and `jwkKeys` in the constructor and offers a mechanism to validate the JWTs. Just thinking... Just saying, we probably need to decode and cache the entire openid-configuration object.

isoos:
I've updated the models, the HTTP fetch and the caching to store the combined data of the provider and the keys. However, I think we should only limit the data structures to the ones we actually use (unless we want to implement a full-blown openid client, which I am not planning to at the moment :) ).
PTAL.

app/lib/service/openid/openid_models.dart
```diff
@@ -0,0 +1,145 @@
...
+
+  Map<String, dynamic> toJson() => _$OpenIdDataToJson(this);
+}
+
+/// The (partial) data structure of the `.well-known/openid-configuration` endpoint.
+///
+/// https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata
+///
+/// TODO: add all fields required for token verification, as described in:
+///       https://github.com/dart-lang/pub-dev/pull/5795#discussion_r884877050
```
jonasfj, line 34: Atleast the fields provided by github :D

jonasfj (approved):

app/lib/service/openid/openid_models.dart
isoos: Ack. I'll add them as part of a follow-up PR.

PR #5808, opened 2022-05-31
Finalize new incoming bucket: do not try fallback bucket on upload.

isoos:
#5586


jonasfj (approved):

jonasfj:
Maybe update the changelog

PR #5810, opened 2022-05-31
Migrate some tests to use the HTTP-based pub API client.

isoos:


app/test/account/like_test.dart
```diff
@@ -14,184 +15,232 @@ void main() {
...
-      await client2.unlikePackage(pkg1);
-
-      final rs11 = await client.getPackageLikes(pkg1);
-      expect(rs11.likes, 0);
-      final rs12 = await client2.getPackageLikes(pkg1);
-      expect(rs12.likes, 0);
+      await withHttpPubApiClient(
+        bearerToken: userAtPubDevAuthToken,
+        fn: (client) async {
+          await withHttpPubApiClient(
```
jonasfj, line 164: Maybe we should tweak `testWithProfile` to create a global `http.Client get testHttpClient` that is created/closed before/after tests are running.

That way we don't have to do: `withHttpPubApiClient`
We can just make a `final client = pubApiClient(bearerToken: ...);`

or something like that. I don't think we need to block this PR.

jonasfj (approved):

app/test/account/like_test.dart
isoos: Good idea!

PR #5812, opened 2022-06-01
JWT token parsing.

isoos:
#5769

### DIFF
--- a/app/lib/service/openid/jwt.dart
+++ b/app/lib/service/openid/jwt.dart
@@ -1,0 +1,66 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:typed_data';
+
+/// Pattern for a valid JWT, these must 3 base64 segments separated by dots.
+final _jwtPattern = RegExp(
+    r'^[a-zA-Z0-9+/=_-]{4,}\.[a-zA-Z0-9+/=_-]{4,}\.[a-zA-Z0-9+/=_-]{4,}$');
+
+/// Returns true if the format of the [token] could be a JWT based on the pattern.
+bool looksLikeJWT(String token) {
+  return _jwtPattern.hasMatch(token);
+}
+
+/// The decoded JWT token.
+class JWT {
+  final Map<String, dynamic> header;
+  final Map<String, dynamic> payload;
+  final Uint8List signature;
+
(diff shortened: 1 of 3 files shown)


app/lib/service/openid/jwt.dart
```diff
@@ -0,0 +1,66 @@
...
+final _jwtPattern = RegExp(
+    r'^[a-zA-Z0-9+/=_-]{4,}\.[a-zA-Z0-9+/=_-]{4,}\.[a-zA-Z0-9+/=_-]{4,}$');
+
+/// Returns true if the format of the [token] could be a JWT based on the pattern.
+bool looksLikeJWT(String token) {
+  return _jwtPattern.hasMatch(token);
+}
+
+/// The decoded JWT token.
+class JWT {
```
jonasfj, line 18: nit..
```suggestion
/// Representation of a JWT as defined by [RFC7519][1].
///
/// [1]: https://datatracker.ietf.org/doc/html/rfc7519
class JsonWebToken {
```
having a reference to the RFC is really nice.

app/lib/service/openid/jwt.dart
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:typed_data';
+
+/// Pattern for a valid JWT, these must 3 base64 segments separated by dots.
+final _jwtPattern = RegExp(
+    r'^[a-zA-Z0-9+/=_-]{4,}\.[a-zA-Z0-9+/=_-]{4,}\.[a-zA-Z0-9+/=_-]{4,}$');
```
jonasfj, line 10: https://datatracker.ietf.org/doc/html/rfc7519#section-2
says:
> The terms "JSON Web Signature ([JWS](https://datatracker.ietf.org/doc/html/rfc7519#ref-JWS))", "Base64url Encoding", "Header
   Parameter", "JOSE Header", "JWS Compact Serialization", "JWS
   Payload", "JWS Signature", and "Unsecured JWS" are defined by the JWS
   specification [JWS].

JWS, [RFC7515 Appendix C specifies](https://www.rfc-editor.org/rfc/rfc7515.html#appendix-C)
>    This appendix describes how to implement base64url encoding and
   decoding functions without padding based upon standard base64
   encoding and decoding functions that do use padding.

Hence, as far as I can see padding is not allowed.

-------------

Please consider including references to these RFCs in source code comment.

jonasfj (approved):

PR #5815, opened 2022-06-01
app: drop final use of pkg:http_retry

kevmoo:


isoos (approved):

isoos:
Running `pub get` is also needed, as the `ensure_build_test.dart` is failing.

PR #5823, opened 2022-06-03
Change copy mouse cursor to pointer.

### LINKED ISSUE #5813: Strange mouse cursor when hovering over READMEs' copy button
Hello! Hovering over a pub.dev code snippet's copy button adds a plus sign to your mouse cursor. This does not feel idiomatic: I would expect it to do the same mouse hover effect as a button (which is what GitHub's copy button does as well).

isoos:
- Fixes #5813.
- Also fixes generic copy icon in `<pre>` blocks.


sigurdm (approved):
LGTM

PR #5836, opened 2022-06-09
Upload package archive to canonical and public bucket too.

isoos:
- #5586
- The upload practically happens as part of the transaction callback, because if there is any issue with the storage buckets, the transaction won't get committed, and the failure state is still acceptable without further changes of other code paths.
- Kept the same objectname pattern (`packages/<package>-<version>.tar.gz`).
- When there is a canonical archive binary, and attempt to update it with the same content is not blocked (e.g. repeating a failed transaction), but different archive is not accepted.
- Integrity verification will check for the object's metadata in each bucket, but won't do a byte-to-byte comparison yet.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -76,10 +76,27 @@ class PackageBackend {
   /// - `tmp/$guid` (incoming package archive that was uploaded, but not yet processed)
   final Bucket _incomingBucket;
 
+  /// The Cloud Storage bucket to use for canonical package archives.
+  /// The following files are present:
+  /// - `packages/$package-$version.tar.gz` (package archive)
+  final Bucket _canonicalBucket;
+
+  /// The Cloud Storage bucket to use for public package archives.
+  /// The following files are present:
+  /// - `packages/$package-$version.tar.gz` (package archive)
+  final Bucket _publicBucket;
+
   @visibleForTesting
   int maxVersionsPerPackage = _defaultMaxVersionsPerPackage;
 
-  PackageBackend(this.db, this._storage, this._bucket, this._incomingBucket);
+  PackageBackend(
+    this.db,
+    this._storage,
+    this._bucket,
+    this._incomingBucket,
+    this._canonicalBucket,
+    this._publicBucket,
+  );
 
   /// Whether the package exists and is not withheld or deleted.
   Future<bool> isPackageVisible(String package) async {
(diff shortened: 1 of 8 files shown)


jonasfj:
> the transaction gets a rollback

Let's maybe just say the transaction doesn't get committed :D But yes.

app/lib/package/backend.dart
```diff
@@ -832,28 +849,47 @@ class PackageBackend {
...
+      final canonicalArchivePath =
+          tarballObjectName(pubspec.name, versionString);
+      final canonicalArchiveInfo =
+          await _canonicalBucket.tryInfo(canonicalArchivePath);
+      if (canonicalArchiveInfo != null) {
+        final file = File(filename);
+        if (canonicalArchiveInfo.length != await file.length()) {
+          throw PackageRejectedException.versionExists(
+              pubspec.name, versionString);
+        }
```
jonasfj, line 877: Isn't this covered by `byteToByteEquals` ?

I think it's extremely rare that this happens, so we don't have to optimize for cases where the length does match.

app/lib/package/backend.dart
```diff
@@ -931,6 +967,7 @@ class PackageBackend {
     required User user,
     required PackageSummary archive,
     required String guid,
+    required bool hasCanonicalArchiveObject,
```
jonasfj, line 970: This is just an optimization, right?

Because you don't want to overwrite if it already exists with the same bytes.
I think we can omit this optimization..? Won't it reduce the amount of code flows we have to test?

app/lib/package/backend.dart
```diff
@@ -1269,6 +1319,8 @@ class PackageBackend {
   Future<void> removePackageTarball(String package, String version) async {
     final object = tarballObjectName(package, version);
     await deleteFromBucket(_bucket, object);
+    await deleteFromBucket(_publicBucket, object);
+    await deleteFromBucket(_canonicalBucket, object);
```
jonasfj, line 1323: Maybe later we should split this into 2 methods:
 * removePublicPackageTarball
 * removeCanonicalPackageTarball

jonasfj (approved):

app/lib/package/backend.dart
isoos: This is a cheap test that gets a fast-path rejection without actually fetching the bytes from the bucket. (E.g. if a client retries an upload too much, we will spend fewer resources on the validation and rejection...)

app/lib/package/backend.dart
isoos: Good question on the testability. I've added it to reduce the time we are spending in the transaction: it is a work that is a no-op in the end, so an extra flag won't really hurt to check for it. We can remove it now (or maybe later too)..

app/lib/package/backend.dart
isoos: Maybe we remove only from the public bucket, and removing from the canonical is behind admin-only API?

isoos:
Instead of a separate PR, I've added the periodic public bucket update tool here, as it required the upload change already to be there to make sense in the tests.
https://github.com/dart-lang/pub-dev/pull/5845#issuecomment-1156507340

PTAL.

app/lib/package/backend.dart
isoos: Yeah, we did, but it has not been started yet.

PR #5838, opened 2022-06-10
Update invite text. Emails are no longer public

sigurdm:


app/test/frontend/golden/consent_package_uploader_authenticated.html
```diff
@@ -27,4 +20,4 @@
       .
     </li>
   </ul>
-</fragment>
+</fragment>
```
isoos, line 23: why remove the end of line?

isoos (approved):
Needs to fix the templates goldens, but otherwise looks good.

PR #5845, opened 2022-06-13
Backfill new buckets with archive files.

isoos:
- #5586
- Related to #5836, using the same helper methods to compare `ObjectInfo` values.
- Backfill reports the number of checks, copied objects, identical objects (only headers are checked) and invalid matches.
- Tests need to be updated to match changes in #5836, the merge order doesn't matter much.

### DIFF
--- a/app/lib/shared/storage.dart
+++ b/app/lib/shared/storage.dart
@@ -18,7 +18,13 @@ import 'package:pub_dev/shared/env_config.dart';
 import 'package:retry/retry.dart';
 
 import 'configuration.dart';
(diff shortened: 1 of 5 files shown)


jonasfj:
We need a separate job to copy from canonical to public, as a periodic background job. Just in cases where publishing fails halfway through... sort of to ensure eventually consistency, doesn't have to run more than once per week or so...

isoos:
> We need a separate job to copy from canonical to public, as a periodic background job. Just in cases where publishing fails halfway through... sort of to ensure eventually consistency, doesn't have to run more than once per week or so...

I'll add that as a separate PR.

PR #5848, opened 2022-06-15
Verify JWT signatures with OpenSSL CLI wrapper.

isoos:
- #5769
- only RSA + SHA256 is used for now
- `package:pem` is used to emit PEM-formatted keys for `openssl` (there is a slight tweaking for `RSA PRIVATE KEY`, I'll create an update for `package:pem`)
- `openssl` key-gen and signing is also implemented, in order to create larger tests

### DIFF
--- a/app/lib/service/openid/openssl_commands.dart
+++ b/app/lib/service/openid/openssl_commands.dart
@@ -3,108 +3,11 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:io';
-import 'dart:typed_data';
 
-import 'package:meta/meta.dart';
 import 'package:pana/pana.dart';
 import 'package:path/path.dart' as p;
 import 'package:pem/pem.dart';
 import 'package:pub_dev/shared/utils.dart';
-
-/// Randomly generated private and public RSA keypair.
-@visibleForTesting
-class RsaKeyPair {
-  final Uint8List privateKey;
-  final Uint8List publicKey;
-
-  RsaKeyPair(this.privateKey, this.publicKey);
-}
-
-/// Generates a new RSA keypair using `openssl`.
-@visibleForTesting
-Future<RsaKeyPair> generateRsaKeyPair({
-  int keySize = 2048,
-}) async {
-  return withTempDirectory((dir) async {
-    final privateKeyPath = p.join(dir.path, 'private.pem');
-    final publicKeyPath = p.join(dir.path, 'public.pem');
-    final pr1 = await runProc(
-      [
-        'openssl',
-        'genrsa',
-        '-out',
-        privateKeyPath,
-        '$keySize',
-      ],
-      timeout: Duration(seconds: 10),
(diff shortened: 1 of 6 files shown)


app/lib/service/openid/openssl_commands.dart
```diff
@@ -0,0 +1,145 @@
...
+class RsaKeyPair {
+  final Uint8List privateKey;
+  final Uint8List publicKey;
+
+  RsaKeyPair(this.privateKey, this.publicKey);
+}
+
+/// Generates a new RSA keypair using `openssl`.
+@visibleForTesting
+Future<RsaKeyPair> generateRsaKeyPair({
```
jonasfj, line 25: We don't strictly need this, can't we just hardcode a few key public keys into the test cases? :see_no_evil:

app/lib/service/openid/openssl_commands.dart
```diff
@@ -0,0 +1,145 @@
...
+    await inputFile.writeAsString(input);
+    final signatureFile = File(p.join(dir.path, 'input.sign'));
+    await signatureFile.writeAsBytes(signature);
+    final publicKeyFile = File(p.join(dir.path, 'public.pem'));
+    final publicPemContent = encodePemBlock(PemLabel.publicKey, publicKey);
+    await publicKeyFile.writeAsString(publicPemContent);
+    final pr = await runProc(
+      [
+        'openssl',
+        'dgst',
```
jonasfj, line 126: We could also consider:
https://www.openssl.org/docs/manmaster/man1/openssl-pkeyutl.html

But I don't see how it would be much different.

app/lib/service/openid/openssl_commands.dart
```diff
@@ -0,0 +1,145 @@
...
+      timeout: Duration(seconds: 10),
+    );
+    if (pr.exitCode != 0) {
+      throw Exception('Unable to run openssl:\n${pr.asJoinedOutput}');
+    }
+    return await outputFile.readAsBytes();
+  });
+}
+
+/// Verifies if [signature] is valid for [input], using the RSA public key.
```
jonasfj, line 109: Note that `publicKey` must be a DER encoded SubjectPublicKeyInfo following RFC 5280.

I think that's the thing that this is supposed to be.

I'm deciphering this using: https://datatracker.ietf.org/doc/html/rfc7468#section-4


-----------------

I think it's generally useful to write down what bits we expect to enter a variable.

jonasfj (approved):

app/lib/service/openid/openssl_commands.dart
sigurdm: And that also counts for the signing part. Let's keep this as small as possible.

isoos: Moved these to the `test/` subdirectory, because I think they may be needed in the near future for integration tests. Added a TODO to review/remove them later.

PR #5852, opened 2022-06-16
AuthenticatedAgent skeleton to prepare credential-less uploads.

isoos:
- #5769
- Publishing via JWT/service-agent is not yet enabled, and authorization returns `false` for it, but the PR prepares all the related code to handle either an authenticated `User` or an authenticated agent.
- `AuthenticatedAgent` is only a skeleton for now, will be updated once we connect the dots with JWT tokens from GitHub.

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -122,11 +122,16 @@ abstract class AuthenticatedAgent {
 /// Holds the authenticated Github Action information.
 class AuthenticatedGithubAction implements AuthenticatedAgent {
   final String label;
-  final JsonWebToken jwt;
+
+  /// OIDC `id_token` the request was authenticated with.
+  ///
+  /// The [agentId] of an [AuthenticatedAgent] have always been authenticated using the [idToken].
+  /// Hence, claims on the [idToken] may be used to determine authorization of a request.
+  final JsonWebToken idToken;
 
   AuthenticatedGithubAction({
     required this.label,
-    required this.jwt,
+    required this.idToken,
   });
 
   @override
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -12,6 +12,7 @@ import 'package:meta/meta.dart';
 // ignore: import_of_legacy_library_into_null_safe
 import 'package:neat_cache/neat_cache.dart';
 import 'package:pub_dev/account/agent.dart';
+import 'package:pub_dev/service/openid/jwt.dart';
 
 import '../package/models.dart';
 import '../shared/datastore.dart';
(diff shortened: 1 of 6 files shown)


app/lib/account/backend.dart
```diff
@@ -111,6 +112,59 @@ Future<User> requireAuthenticatedUser({AuthSource? source}) async {
   return user;
 }
 
+/// Holds the authenticated agent information.
+class AuthenticatedAgent {
+  final String agentId;
+  final String label;
+  final JsonWebToken jwt;
```
jonasfj, line 119:
```suggestion
  /// OIDC `id_token` the request was authenticated with.
  ///
  /// The [agentId] of an [AuthenticatedAgent] have always been authenticated using the [id_token].
  /// Hence, claims on the [id_token] may be used to determine authorization of a request.
  final JsonWebToken id_token;
```

app/lib/account/backend.dart
```diff
@@ -111,6 +112,59 @@ Future<User> requireAuthenticatedUser({AuthSource? source}) async {
   return user;
 }
 
+/// Holds the authenticated agent information.
+class AuthenticatedAgent {
```
jonasfj, line 116: Can we say what an `AuthenticatedAgent` is`

```suggestion
/// An [AuthenticatedAgent] represents a _service_ that has been authenticated
/// and which may be allowed to operate on specific resources on pub.dev
///
/// Examples:
///  * A GCP service account may authenticate using an OIDC `id_token`,
///  * A Github Action may authenticate using an OIDC `id_token`.
class AuthenticatedAgent {
```

jonasfj: Would the term `AuthenticatedService` be better, or `AuthenticateBot` since these are always bots or services.

I think `AuthenticateService` might be best since this really is intended to be _service accounts_ and github actions.

app/lib/account/backend.dart
jonasfj: Might even worthwhile documenting that:
 * The signature of the `id_token` have been verified.
 * The `audience` and `expiration` of the `id_token` has been verified.

app/lib/account/backend.dart
```diff
@@ -111,6 +112,59 @@ Future<User> requireAuthenticatedUser({AuthSource? source}) async {
...
+class AuthenticatedUserOrAgent {
+  final User? user;
+  final AuthenticatedAgent? agent;
+
+  AuthenticatedUserOrAgent.user(User this.user) : agent = null;
+  AuthenticatedUserOrAgent.agent(AuthenticatedAgent this.agent) : user = null;
+
+  late final bool isUser = user != null;
+  late final bool isNotUser = !isUser;
+  late final User asUser = user!;
```
jonasfj, line 138: Isn't it almost always better to do:
```dart
final u = authenticateAgent.user;
if (u != null) {
  print(u.agentId);
}
```

using `asUser` feels like a trap, it invites me to not create that `if != null`

app/lib/account/backend.dart
```diff
@@ -111,6 +112,59 @@ Future<User> requireAuthenticatedUser({AuthSource? source}) async {
...
+  late final bool isNotAgent = !isAgent;
+  late final AuthenticatedAgent asAgent = agent!;
+
+  late final userIdOrAgentId = isUser ? asUser.userId : asAgent.agentId;
+  late final emailOrLabel = isUser ? asUser.email! : asAgent.label;
+}
+
+/// Verifies the current bearer token in the request scope and returns the
+/// current authenticated user or a service agent with the available data.
+Future<AuthenticatedUserOrAgent> requireAuthenticatedUserOrAgent(
```
jonasfj, line 150: We could also just do:
```dart
/// Returns [AuthenticatedAgent] or [User].
Future<Object> requireAuthenticatedUserOrAgent(
```

Then calls have to do `if (... is User) {...`

app/lib/package/backend.dart
```diff
@@ -998,11 +1012,19 @@ class PackageBackend {
...
-            'for package ${package!.name}, rolling transaction back.');
-        throw AuthorizationException.userCannotUploadNewVersion(
-            user.email!, package!.name!);
+      } else {
+        final isAdmin = !await packageBackend.isPackageAdmin(
+          package!,
+          userOrAgent.user?.userId,
+          agent: userOrAgent.agent,
+        );
+        if (isAdmin) {
```
jonasfj, line 1021:
```suggestion
        if (!isAdmin) {
```

app/lib/package/backend.dart
```diff
@@ -998,11 +1012,19 @@ class PackageBackend {
       if (package == null) {
         _logger.info('New package uploaded. [new-package-uploaded]');
         package = Package.fromVersion(newVersion);
-      } else if (!await packageBackend.isPackageAdmin(package!, user.userId)) {
-        _logger.info('User ${user.userId} (${user.email}) is not an uploader '
-            'for package ${package!.name}, rolling transaction back.');
-        throw AuthorizationException.userCannotUploadNewVersion(
-            user.email!, package!.name!);
+      } else {
+        final isAdmin = !await packageBackend.isPackageAdmin(
```
jonasfj, line 1016:
```suggestion
        final isAdmin = await packageBackend.isPackageAdmin(
```

jonasfj: Move the negation or change the variable name to `isNotAdmin`

PR #5870, opened 2022-06-23
Relaxed processing of Youtube Playlist API response.

isoos:
- When a `PlaylistItem` fails, it is local to that specific item, other items may be processed and displayed.
- When the usual `high`-resolution thumbnail is missing, fall back to other thumbnails.
- When the `snippet` or `snippet.title` is missing, we still display the thumbnail.

### DIFF
--- a/app/lib/service/youtube/backend.dart
+++ b/app/lib/service/youtube/backend.dart
@@ -145,7 +145,7 @@ class _PkgOfWeekVideoFetcher {
               );
             } catch (e, st) {
               // this item will be skipped, the rest of the list may be displayed
-              _logger.pubNoticeWarning(
+              _logger.pubNoticeShout(
                   'youtube', 'Processing Youtube PlaylistItem failed.', e, st);
             }
             return null;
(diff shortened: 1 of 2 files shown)


app/lib/service/youtube/backend.dart
```diff
@@ -124,14 +128,29 @@ class _PkgOfWeekVideoFetcher {
...
+              return PkgOfWeekVideo(
+                videoId: i.contentDetails!.videoId!,
+                title: i.snippet?.title ?? '',
+                description:
+                    (i.snippet?.description ?? '').trim().split('\n').first,
+                thumbnailUrl: thumbnail!.url!,
+              );
+            } catch (e, st) {
+              // this item will be skipped, the rest of the list may be displayed
+              _logger.pubNoticeWarning(
```
jonasfj, line 148: Let's make it shout.. this shouldn't happen... If it does I suggest we filter out entries without a URL..

jonasfj (approved):

PR #5874, opened 2022-06-24
Add screenshot to search results

szakarias:


szakarias:
image

app/lib/frontend/templates/views/pkg/package_list.dart
```diff
@@ -45,8 +45,10 @@ d.Node _sdkLibraryItem(SdkLibraryHit hit) {
   ].join(' • ');
 
   return _item(
+    screenshots: [],
     url: hit.url!,
     name: hit.library!,
+    version: hit.version,
```
sigurdm, line 51: This is not related to screenshots - right?

@isoos is there a reason we did not have the version check here before?

app/lib/package/models.dart
```diff
@@ -855,25 +859,26 @@ class PackageView extends Object with FlagMixin {
...
+        flags: flags,
+        publisherId: publisherId,
+        isPending: isPending,
+        likes: likes,
+        grantedPubPoints: grantedPubPoints,
+        maxPubPoints: maxPubPoints,
+        tags: tags,
+        spdxIdentifiers: spdxIdentifiers,
+        apiPages: apiPages ?? this.apiPages,
+        screenshots: screenshots);
```
sigurdm, line 881: Nit: Add a trailing comma

sigurdm (approved):
LGTM

app/lib/frontend/templates/views/pkg/package_list.dart
isoos: It seems to be that this is required for the screenshot URL. Unfortunately, I don't think that `version` will be provided for Flutter SDK, but in such cases it won't matter as we will not have screenshots for such hits...

app/lib/frontend/templates/views/pkg/package_list.dart
```diff
@@ -146,15 +150,25 @@ d.Node _packageItem(
...
   required String name,
+  required String? version,
   required DateTime? newTimestamp,
   required d.Node? labeledScoresNode,
   required String description,
   required d.Node metadataNode,
   required d.Node? tagsNode,
   required List<_ApiPageUrl>? apiPages,
 }) {
+  final bool hasScreenshots = screenshots != null && screenshots.isNotEmpty;
```
isoos, line 164: maybe add `&& version != null` just in case, as the `getImageURL` uses `version!`

szakarias: done

sigurdm (approved):
LGTM

app/lib/frontend/templates/views/pkg/package_list.dart
```diff
@@ -189,12 +200,38 @@ d.Node _item({
...
+            d.img(
+                image: d.Image(
+                    alt: 'screenshot',
+                    width: null,
+                    height: null,
+                    src: thumbnailUrl)),
+            d.img(
+                classes: ['collections-icon'],
+                image: d.Image(
+                    height: null, // We handle sizing in css
```
isoos, line 229: I think the lighthouse analysis will downscore us if we don't put width and height here. Let's just repeat them here too, event though the CSS may have them as well.

app/lib/package/screenshots/backend.dart
```diff
@@ -67,4 +67,8 @@ class ImageStorage {
   }
 
   // TODO(zarah) add functionality to garbage collect bucket entries.
+
+  String getImageURL(String package, String version, String imageFilePath) {
+    return bucket.objectUrl([package, version, imageFilePath].join('/'));
```
isoos, line 72: Note to self: we may need to escape the `bucket.objectUrl` URLs.

Issue #5867, opened 2022-06-22
Add Date/Time to Packages' Changelog on Pub Dev

When viewing a Package's Changelog on pub.dev it would be great to see the date/time when the version was published.


grahamsmith:
![FVy55LzVUAAmzpV.jpeg](https://user-images.githubusercontent.com/1571527/174952215-bd06d73c-f701-496a-84d5-2089889f96db.jpeg)

Ref original tweet about this: 
https://twitter.com/whoisgraham/status/1539309744812216325?s=20&t=IG_e3iX0U3zbQNHIbiIFMQ

grahamsmith:
@mit-mit tagging you in per Tim's instructions 👍

sigurdm:
The changelog is presented directly from the actual uploaded package - currently there is nothing actually correlating it with the published versions existing on pub.dev.

There is the "versions" tab that presents the actual versions and the date they got uploaded.

We have talked about aligning these two into one when possible. There are a number of open questions, such as:

How do we ensure that all changelogs are parsing well and contain the right versions?

Which versions changelog should be used to show on the version tab? (latest / latest stable / latest stable plus the entry from each prerelease after that).
 
I think we want to do this, but it is not on our immediate roadmap.

sigurdm:
![image](https://user-images.githubusercontent.com/8613953/176150877-4830dc63-48dd-40e6-b991-1a696ff68fe4.png)

If you click on the "x days ago" you get to see the actual date.

grahamsmith:
Thanks for getting back to me.

I think this is a UX problem more than some complicated coding issue.

It's good to know that I can click to reveal the date.

I think it shows we need a bit of work as there's no indication that I can click to reveal.

Imagine there's bigger fish to fry but also this can't be a long hard issue to solve either.

sigurdm:
> I think it shows we need a bit of work as there's no indication that I can click to reveal.

Yeah - that's a good point, the discoverability is not great - I wonder how we could make this more obvious.

@isoos do you have good ideas? Perhaps it should show the date on hover-instead of click, like here on github:
![image](https://user-images.githubusercontent.com/8613953/176171630-9bfb79ba-8294-468a-ba5f-903dcb1ea624.png)

grahamsmith:
Cant we just print under the X days ago in a smaller font? Would be good for screen readers etc.

isoos:
> Perhaps it should show the date on hover-instead of click, like here on github:

[...]

PR #5881, opened 2022-06-28
Use aria-label on 'X ago' elements.

isoos:
- #5867
- While the specification would recommend to describe the clickable action in the `aria-label`, I think there is very little value to say "clicking this will reveal the real date", instead, we shall just emit the actual date.

### DIFF
--- a/app/lib/frontend/dom/dom.dart
+++ b/app/lib/frontend/dom/dom.dart
@@ -97,7 +97,10 @@ Node xAgoTimestamp(DateTime timestamp, {String? datePrefix}) {
   ].join(' ');
   return span(
     classes: ['-x-ago'],
-    attributes: {'title': title},
+    attributes: {
+      'title': title,
+      'aria-label': title,
+    },
     text: formatXAgo(clock.now().difference(timestamp)),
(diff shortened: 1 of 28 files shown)


isoos:
My understanding is that the screen readers may not read out loud the `title` attribute, and it may be better to use both the `title` and `aria-label` for such cases.

A rather good article about the topic and the label vs. title:
https://www.aditus.io/aria/aria-label/#aria-label-vs-title

sigurdm:
I tried activating a screen reader - and adding the aria-label via "inspect" - I was not able to get it to tell me the aria-label.

Also this "button" seems to not be available via tabbing through the different links. The link order also seems a bit off.

We should maybe start a bigger review of the accessibility of the page.

sigurdm:
Yeah - I also don't know.

sigurdm:
Github seems to use a `a` element

image

sigurdm:
In github the link is selectable with tab. But I  can't get the screenreader to read the actual date.

isoos:
Closing as #5892 is the preferred solution.

PR #5882, opened 2022-06-28
Emphasize 'X ago' spans with CSS styles.

isoos:
- as suggested in #5867:
- the dotted style is always present, on hovering we switch to a solid line, with slightly darker color


sigurdm (approved):
LGTM

PR #5883, opened 2022-06-29
Minimal ASN.1 encoder for RSA public keys to DER-encoded bytes (used by PEM)

isoos:
- #5769
- JWT providers share their public key using JSON Web Key(s) (JWK), and it contains the RSA key as modulus+exponent pair. We are using `openssl` to verify the JWT signature, and to do this, we need to convert the JWK data to PEM.
- The encoder is using only the minimally required operations to encode from RSA components to ASN.1 + PEM format.
- Uses the known public key (and its extracted components) from jwt.io as test.

### DIFF
--- a/app/lib/service/openid/asn1_encoder.dart
+++ b/app/lib/service/openid/asn1_encoder.dart
@@ -64,6 +64,7 @@ List<int> encodeSequence(Iterable<List<int>> parts) {
   ];
 }
 
+/// A constant value that identifies the data structure as RSA public key.
 const _objectIdentifierRsa = <int>[
   0x06,
   0x09,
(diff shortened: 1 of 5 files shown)


app/lib/service/openid/asn1_encoder.dart
```diff
@@ -0,0 +1,100 @@
...
+List<int> encodeSequence(Iterable<List<int>> parts) {
+  final totalLength = parts.map((e) => e.length).fold<int>(0, (a, b) => a + b);
+  return <int>[
+    0x30,
+    ...encodeLength(totalLength),
+    ...parts.expand((e) => e),
+  ];
+}
+
+const _objectIdentifierRsa = <int>[
```
sigurdm, line 67: Could you write a dartdoc for this?

sigurdm:
cc @jonasfj probably wants to read this also when he comes back.

PR #5888, opened 2022-06-30
Store archive sha256 hash on PackageVersion.

### LINKED ISSUE #5880: Add content-hashes to version listings
Current design proposal:

We put a hex-encoded SHA-256 sum of the archive.tar.gz next to the archive_url.

```
{
  "name": "<package>",
  "isDiscontinued": true || false, /* optional field, false if omitted */
  "replacedBy": "<package>", /* optional field, if isDiscontinued == true */
  "latest": {
    "version": "<version>",
    "retracted": true || false, /* optional field, false if omitted */
    "archive_url": "https://.../archive.tar.gz",
    "archive_sha256": "cf83e1357eefb8bdf1542850d6[...]81a538327af927da3e", /* NEW */
    "pubspec": {
      /* pubspec contents as JSON object */
    }
  },
  "versions": [
    {
      "version": "<package>",
      "retracted": true || false, /* optional field, false if omitted */
      "archive_url": "https://.../archive.tar.gz",
      "archive_sha256": "cf83e1357eefb8bdf15428[...]81a538327af927da3e", /* NEW */
      "pubspec": {
        /* pubspec contents as JSON object */
      }
    },
    /* additional versions */
  ]
}
```

isoos:
- fixes #5880.
- stores sha256 hash on new uploads
- backfill to read canonical archive files and calculate hash of already existing versions
- integrity check to verify hash length, and randomly verify actual file hash value (0.1% of the archives on each run)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ AppEngine version, listed here to ease deployment and troubleshooting.
  * Bumped runtimeVersion to `2022.06.27`.
  * NOTE: Relaxed processing of Youtube Playlist API response.
  * NOTE: Store and use verified URL and repository results from pana.
+ * NOTE: Started to populate `PackageVersion.sha256` and serve it in API response.
 
 ## `20220623t124000-all`
  * Bumped runtimeVersion to `2022.06.23`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ AppEngine version, listed here to ease deployment and troubleshooting.
  * Bumped runtimeVersion to `2022.06.27`.
  * NOTE: Relaxed processing of Youtube Playlist API response.
  * NOTE: Store and use verified URL and repository results from pana.
- * NOTE: Started to populate `PackageVersion.sha256` and serve it in API response.
+ * NOTE: Started to populate `PackageVersion.sha256`.
 
 ## `20220623t124000-all`
  * Bumped runtimeVersion to `2022.06.23`.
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ AppEngine version, listed here to ease deployment and troubleshooting.
  * Bumped runtimeVersion to `2022.06.27`.
  * NOTE: Relaxed processing of Youtube Playlist API response.
  * NOTE: Store and use verified URL and repository results from pana.
+ * NOTE: Started to populate `PackageVersion.sha256` and serve it in API response.
```
sigurdm, line 8: Maybe we start with the populating+backfilling+integrity check, and start serving in a separate PR - we don't want to promise anything in the API before we are certain about the client side of this.

app/lib/package/backend.dart
```diff
@@ -838,6 +840,9 @@ class PackageBackend {
           _incomingBucket.read(tmpObjectName(guid)), filename);
       _logger.info('Examining tarball content ($guid).');
       final sw = Stopwatch()..start();
+      final file = File(filename);
+      final fileBytes = await file.readAsBytes();
```
sigurdm, line 844: Consider making this streaming.

CHANGELOG.md
isoos: Officially we don't promise supporting `retracted` either - it is not part of the spec. Nevertheless, I'll remove it from this PR, exposing it isn't really an issue, it just made the testing slightly easier.

app/lib/package/backend.dart
isoos: It may work for the hashing, but `fileBytes` may be used later too: when there is an archive file in the canonical bucket, we need to reject the publishing if it differs from the current file. Unless streaming gives us a noticeable benefit, I'd rather keep it as-is.

PR #5904, opened 2022-07-08
Parse GitHub JWT payload.

isoos:
#5769

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -144,6 +144,9 @@ abstract class AuthenticatedAgent {
 /// Holds the authenticated Github Action information.
 class AuthenticatedGithubAction implements AuthenticatedAgent {
   @override
+  String get agentId => KnownAgents.githubActions;
+
+  @override
   final String visibleId;
 
   /// OIDC `id_token` the request was authenticated with.
@@ -156,13 +159,14 @@ class AuthenticatedGithubAction implements AuthenticatedAgent {
   /// pub-specific scope of the token.
   final JsonWebToken idToken;
 
+  /// The parsed, GitHub-specific JWT payload.
+  final GitHubJwtPayload payload;
+
   AuthenticatedGithubAction({
     required this.visibleId,
     required this.idToken,
+    required this.payload,
   });
-
-  @override
-  String get agentId => KnownAgents.githubActions;
 }
 
 /// Holds the authenticated user information.
@@ -186,16 +190,33 @@ Future<AuthenticatedAgent> requireAuthenticatedAgent(
   if (token == null || token.isEmpty) {
     throw AuthenticationException.authenticationRequired();
   }
-  if (JsonWebToken.looksLikeJWT(token)) {
-    final idToken = JsonWebToken.tryParse(token);
-    if (idToken != null) {
-      // TODO: check the audience and expiration
-      // TODO: skip JWT processing if it is not a recognized service agent
-      // TODO: check signature from JWKS
-      // TODO: when everything is verified, return the JWT token.
-    }
+  final authenticatedService = _tryAuthenticateGithubAction(token);
+  if (authenticatedService != null) {
+    return authenticatedService;
+  } else {
+    return AuthenticatedUser(await requireAuthenticatedUser(source: source));
   }
-  return AuthenticatedUser(await requireAuthenticatedUser(source: source));
+}
+
+AuthenticatedGithubAction? _tryAuthenticateGithubAction(String token) {
+  if (!JsonWebToken.looksLikeJWT(token)) {
+    return null;
+  }
+  final idToken = JsonWebToken.tryParse(token);
+  if (idToken == null) {
+    return null;
+  }
+  if (!idToken.payload.verifyTimestamps()) {
+    return null;
+  }
+  final payload = GitHubJwtPayload.tryParse(idToken.payload);
+  if (payload == null) {
+    return null;
+  }
+  // TODO: check the audience
+  // TODO: check signature from JWKS
+  // TODO: when everything is verified, return the JWT token.
+  return null;
 }
 
 /// Represents the backend for the account handling and authentication.
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -138,13 +138,13 @@ abstract class AuthenticatedAgent {
   ///  * For a regular user we display their `email`.
   ///  * For a service account we display a description.
   ///  * For automated publishing we display the service and the origin trigger.
-  String get formattedId;
+  String get visibleId;
 }
 
 /// Holds the authenticated Github Action information.
 class AuthenticatedGithubAction implements AuthenticatedAgent {
   @override
-  final String formattedId;
+  final String visibleId;
 
   /// OIDC `id_token` the request was authenticated with.
   ///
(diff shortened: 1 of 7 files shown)


isoos:
I'm not yet sure if this is a good way to do this, or should we rather just have helper methods without dedicated types. wdyt?

app/lib/service/openid/jwt.dart
```diff
@@ -123,7 +114,179 @@ Map<String, dynamic> _decodePart(String part) {
...
+        ref = map['ref'] as String,
+        refType = map['ref_type'] as String,
+        environment = map['environment'] as String,
+        super._(map);
+
+  factory GitHubJwtPayload(JwtPayload payload) {
+    final missing = payload.firstMissingKey(_keys);
+    if (missing != null) {
+      throw FormatException('Missing key: `$missing`.');
+    }
```
jonasfj, line 269: Why only the first?
```suggestion
    final missing = _keys.difference({...payload.keys}).sorted();
    if (missing.isNotEmpty) {
      throw FormatException('JWT from Github is missing following claims: ${missing.map((k) => '`$k`').join(', ')}.');
    }
```

app/lib/service/openid/jwt.dart
```diff
@@ -123,7 +114,179 @@ Map<String, dynamic> _decodePart(String part) {
...
+    'nbf',
+    'exp',
+    'aud',
+    'repository',
+    'repository_owner',
+    'event_name',
+    'ref',
+    'ref_type',
+    'environment',
+  ];
```
jonasfj, line 253: Why not use an set? if there is something important to the ordering, then that should be documented.
```suggestion
  static const _requiredClaims = <String>{
    'iat',
    'nbf',
    'exp',
    'aud',
    'repository',
    'repository_owner',
    'event_name',
    'ref',
    'ref_type',
    'environment',
  };
```

Set also allows us to do `_requiredClaims.difference({...payload.keys}).sorted()`, by using set difference operation.

app/lib/service/openid/jwt.dart
```diff
@@ -123,7 +114,179 @@ Map<String, dynamic> _decodePart(String part) {
...
+    final missing = payload.firstMissingKey(_keys);
+    if (missing != null) {
+      return null;
+    }
+    try {
+      return GitHubJwtPayload(payload);
+    } catch (_) {
+      return null;
+    }
+  }
```
jonasfj, line 283: Why make it any more complicated than this?
```suggestion
  static GitHubJwtPayload? tryParse(JwtPayload payload) {
    try {
      return GitHubJwtPayload(payload);
    } on FormatException {
      return null;
    }
  }
```

IMO we should only catch `FormatException` because that's the only thing `GitHubJwtPayload()` is allowed to throw (well, if we want to be perfectly right we should document that too).

app/lib/service/openid/jwt.dart
```diff
@@ -123,7 +114,179 @@ Map<String, dynamic> _decodePart(String part) {
...
+    }
+    if (exp != null && exp!.isBefore(now)) {
+      return false;
+    }
+    return true;
+  }
+}
+
+/// Parsed payload with the payload values GitHub sends with the token.
+class GitHubJwtPayload extends JwtPayload {
```
jonasfj, line 208: On topic: I don't know if we need this abstraction?

Are we going to make `JsonWebToken` type parameterized?

What are we going to return from the `requireAuthentication` method? when we don't return a user?

@isoos  Let's chat when you have time later this week.

app/lib/service/openid/jwt.dart
```diff
@@ -123,7 +114,179 @@ Map<String, dynamic> _decodePart(String part) {
...
+    } on FormatException {
+      rethrow;
+    } catch (_) {
+      throw FormatException('Unexpected value in JWT header.');
+    }
+  }
+}
+
+/// The parsed JWT payload.
+class JwtPayload extends UnmodifiableMapView<String, dynamic> {
```
jonasfj, line 150: For the record `JwtPayload` and `JwtHeader` are pretty nice ways of structuring things.

app/lib/service/openid/jwt.dart
isoos: `requireAuthentication` needs to return either an authenticated `User` or a service account description, as outlined (but not finalized) in https://github.com/dart-lang/pub-dev/pull/5852.

I'm not yet sure about this abstraction, but it feels like we should use something typed and verified over the plain payload data map.

app/lib/service/openid/jwt.dart
isoos: The reason for the catch-all was to catch type-cast exceptions too, which makes the entire initialization easier... but of course we could also verify the type before initializing the field :)

app/lib/service/openid/jwt.dart
```diff
@@ -123,7 +114,179 @@ Map<String, dynamic> _decodePart(String part) {
...
+DateTime? _parseSecondsIfNotNull(int? value) {
   if (value == null) return null;
-  return DateTime.fromMillisecondsSinceEpoch(value * 1000);
+  return DateTime.fromMillisecondsSinceEpoch(value * 1000, isUtc: true);
+}
+
+/// The parsed JWT header.
+class JwtHeader extends UnmodifiableMapView<String, dynamic> {
+  /// algorithm
+  final String? alg;
```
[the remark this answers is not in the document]
isoos: The difference is the eager vs. lazy evaluation of the token. E.g. if the value is an integer instead of a String, it should fail earlier...

app/lib/service/openid/jwt.dart
jonasfj: Maybe we merge this into `AuthenticatedGithubAction` so github related parts live there...

app/lib/account/backend.dart
```diff
@@ -138,13 +138,16 @@ abstract class AuthenticatedAgent {
   ///  * For a regular user we display their `email`.
   ///  * For a service account we display a description.
   ///  * For automated publishing we display the service and the origin trigger.
-  String get formattedId;
+  String get visibleId;
```
jonasfj, line 141: nit:
```suggestion
  String get displayId;
```

app/lib/service/openid/jwt.dart
```diff
@@ -123,7 +119,189 @@ Map<String, dynamic> _decodePart(String part) {
...
+          'Unexpected value in payload: `iss`: `${payload.iss}`.');
+    }
+    return GitHubJwtPayload._(payload);
+  }
+
+  static GitHubJwtPayload? tryParse(JwtPayload payload) {
+    try {
+      return GitHubJwtPayload(payload);
+    } on FormatException {
+      rethrow;
```
jonasfj, line 301: I'm confused, why not;
```suggestion
      return null;
```

app/lib/service/openid/jwt.dart
```diff
@@ -123,7 +119,189 @@ Map<String, dynamic> _decodePart(String part) {
...
+    }
+    if (exp != null && exp!.isBefore(now)) {
+      return false;
+    }
+    return true;
+  }
+}
+
+/// Parsed payload with the payload values GitHub sends with the token.
+class GitHubJwtPayload {
```
jonasfj, line 234: Let's put this in a separate file... maybe `service/openid/github.dart`

jonasfj (approved):

app/lib/service/openid/jwt.dart
isoos: Added a TODO for it, I'll do it in a subsequent change.

Issue #5807, opened 2022-05-31
Refactor `upload_test.dart` to use API and avoid internal backend


jonasfj:
Should we not be able to test publishing packages against the fake-server without accessing the internal `packageBackend` in testing?

This would move our tests closer to just being integration tests. But might that not be okay?

isoos:
Also related: https://github.com/dart-lang/pub-dev/pull/5810#discussion_r887758476

> Maybe we should tweak testWithProfile to create a global http.Client get testHttpClient that is created/closed before/after tests are running.
> That way we don't have to do: withHttpPubApiClient
> We can just make a final client = pubApiClient(bearerToken: ...);
> or something like that. I don't think we need to block this PR

PR #5905, opened 2022-07-08
Refactor API client creation in test: removing with* methods.

isoos:
- Related to #5807 and https://github.com/dart-lang/pub-dev/pull/5810#discussion_r887758476
- All tests are migrated, test_profiler's importer is still using the `withHttpApiClient` method.
- I think the scope-based self-closing client is better for the tests, not yet sure if the same is true for the importer.

### DIFF
--- a/app/lib/tool/utils/pub_api_client.dart
+++ b/app/lib/tool/utils/pub_api_client.dart
@@ -3,19 +3,16 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:gcloud/service_scope.dart';
-import 'package:meta/meta.dart';
 
 import '../../frontend/handlers/pubapi.client.dart';
 import '../../shared/configuration.dart';
 
 import 'http.dart';
 
-/// Creates an API client to the configured endpoint with [authToken].
-/// The client will be closed when the current scope is exited.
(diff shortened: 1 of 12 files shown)


app/lib/tool/utils/pub_api_client.dart
```diff
@@ -3,19 +3,16 @@
...
 
-/// Creates an API client to the configured endpoint with [authToken].
-/// The client will be closed when the current scope is exited.
+/// Creates an API client with [authToken] that uses the configured HTTP endpoints.
 ///
-/// TODO: migrate callers to use [withHttpPubApiClient] instead.
-@visibleForTesting
-PubApiClient createLocalPubApiClient({String? authToken}) {
+/// Services scopes are used to automatically close the client once we exit the current scope.
+PubApiClient createPubApiClient({String? authToken}) {
```
jonasfj, line 15: why not `@visibleForTesting`? 

is it because it cases issues in `app/test/shared/test_services.dart`, if so can we not resolve those issues by annotating things in `app/test/shared/test_services.dart` with `@visibleForTesting´?

isoos: Just a bad edit, restored.

PR #5910, opened 2022-07-19
Require a single LICENSE file (any variant), reject if there are multiple ones.

isoos:
- When a new archive has multiple license files, we'll reject with a message the lists all candidates.
- Optimized (and renamed) `firstFileNameOrNull`, with a slight change: if there are multiple file versions with the same lowercase form, previously we've used the first matching, now we are using the last one (due to a change from `for` loop to a map, where the last entry overrides the previous ones).


jonasfj:
I suggest we just require there to be a `LICENSE` file, and allow other files.

PR #5911, opened 2022-07-19
Required 'LICENSE' in the new versions.

### LINKED ISSUE #5826: Require LICENSE file
We should:
 * analyze how many packages does not contain a `LICENSE` file.
 * require the license file is `LICENSE`.

isoos:
- Fixes #5826.
- Optimized (and renamed) `firstFileNameOrNull`, with a slight change: if there are multiple file versions with the same lowercase form, previously we've used the first matching, now we are using the last one (due to a change from for loop to a map, where the last entry overrides the previous ones).

### DIFF
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -127,7 +127,7 @@ Future<PackageSummary> summarizePackageArchive(
   }
 
   // processing pubspec.yaml
-  final pubspecPath = tar.firstFileNameOrNull(['pubspec.yaml']);
+  final pubspecPath = tar.firstMatchingFileNameOrNull(['pubspec.yaml']);
   if (pubspecPath == null) {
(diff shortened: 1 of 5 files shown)


pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -671,16 +671,16 @@ Iterable<ArchiveIssue> requireIosFolderOrFlutter2_20(
 
 Iterable<ArchiveIssue> requireNonEmptyLicense(
     String? path, String? content) sync* {
-  if (path == null) {
-    yield ArchiveIssue('LICENSE file not found.');
+  if (path == null || path != 'LICENSE') {
+    yield ArchiveIssue('`LICENSE` file not found.');
```
jonasfj, line 675:
```suggestion
    yield ArchiveIssue('`LICENSE` file not found. All packages on pub.dev must contain a `LICENSE` file.');
```

jonasfj (approved):

PR #5917, opened 2022-07-22
Parse OpenID supported claims and algorithms metadata.

isoos:
- These can be used to quickly verify the token payload before more expensive operations.
- While the list of supported claims is not exhaustive, and items may be missing from the payload, we could shortcut authenticating JWT tokens that have too many different claims than the ones the provider supports.
- The supported algorithm list does not have such ambiguity, we shall just use it.

### DIFF
--- a/app/lib/service/openid/openid_models.dart
+++ b/app/lib/service/openid/openid_models.dart
@@ -45,9 +45,23 @@ class OpenIdProvider {
   @JsonKey(name: 'jwks_uri')
   final String jwksUri;
 
+  /// JSON array containing a list of the Claim Names of the Claims that the
+  /// OpenID Provider MAY be able to supply values for. Note that for privacy
+  /// or other reasons, this might not be an exhaustive list.
+  @JsonKey(name: 'claims_supported')
+  final List<String> claimsSupported;
+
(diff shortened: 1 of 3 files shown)


app/lib/service/openid/openid_models.dart
```diff
@@ -45,9 +45,23 @@ class OpenIdProvider {
...
+  /// JSON array containing a list of the Claim Names of the Claims that the
+  /// OpenID Provider MAY be able to supply values for. Note that for privacy
+  /// or other reasons, this might not be an exhaustive list.
+  @JsonKey(name: 'claims_supported')
+  final List<String> claimsSupported;
+
+  /// JSON array containing a list of the JWS signing algorithms (alg values)
+  /// supported by the OP for the ID Token to encode the Claims in a JWT. The
+  /// algorithm RS256 MUST be included.
+  @JsonKey(name: 'id_token_signing_alg_values_supported')
```
kevmoo, line 57: you might consider https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonSerializable/fieldRename.html – it's cleaner 😄

jonasfj (approved):

PR #5918, opened 2022-07-22
Remove PackageBackend.upload, use HTTP upload in tests.

### LINKED ISSUE #5807: Refactor `upload_test.dart` to use API and avoid internal backend

isoos:
Fixes #5807.


sigurdm (approved):
LGTM

PR #5922, opened 2022-07-26
Fix NPE in admin page initialization.

isoos:
This caused an exception in the browser's JS execution, on publisher pages that were not the `/admin` page. Apparently no functionality was affected, only the admin initialization code failed (which was not used on those pages anyway).


sigurdm (approved):
LGTM

PR #5924, opened 2022-07-26
Custom material components build for reduced static file sizes.

isoos:
Reduces static file sizes (about half MB overall reduction):
- JS: `460K -> 173K (-287K)`
- CSS: `463K -> 247K (-216K)`


sigurdm (approved):
LGTM

Issue #4158, opened 2020-10-19
EmailSenderFailed(500): Failed to send email, please retry later.


jonasfj:
I suppose we could increase the retry limit to 3 or 5, instead of only have 2 attempts.

``` 
pub.cloud_repository: Error post-processing package upload ....

Error:
    EmailSenderFailed(500): Failed to send email, please retry later.

Stack:
    #0      _GmailSmtpRelay.sendMessage (package:pub_dev/frontend/email_sender.dart:145:7)
    <asynchronous suspension>
    #1      PackageBackend._performTarballUpload (package:pub_dev/package/backend.dart:617:33)
    <asynchronous suspension>
    #2      PackageBackend.publishUploadedBlob.<anonymous closure> (package:pub_dev/package/backend.dart:502:29)
    <asynchronous suspension>
    #3      PackageBackend.publishUploadedBlob.<anonymous closure> (package:pub_dev/package/backend.dart)
    #4      withTempDirectory.<anonymous closure> (package:pub_dev/shared/utils.dart:48:16)
    #5      StackZoneSpecification._registerUnaryCallback.<anonymous closure>.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:126:26)
    #6      StackZoneSpecification._run (package:stack_trace/src/stack_zone_specification.dart:206:15)
    #7      StackZoneSpecification._registerUnaryCallback.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:126:14)
    #8      _rootRunUnary (dart:async/zone.dart:1198:47)
    #9      _CustomZone.runUnary (dart:async/zone.dart:1100:19)
    #10     _FutureListener.handleValue (dart:async/future_impl.dart:143:18)
    #11     Future._propagateToListeners.handleValueCallback (dart:async/future_impl.dart:696:45)
    #12     Future._propagateToListeners (dart:async/future_impl.dart:725:32)
    #13     Future._completeWithValue (dart:async/future_impl.dart:529:5)
    #14     Future._asyncCompleteWithValue.<anonymous closure> (dart:async/future_impl.dart:567:7)
    #15     StackZoneSpecification._run (package:stack_trace/src/stack_zone_specification.dart:206:15)
    #16     StackZoneSpecification._registerCallback.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:116:48)
    #17     _rootRun (dart:async/zone.dart:1190:13)
[...]

PR #5925, opened 2022-07-29
Store outgoing emails in Datastore to reduce loss of delivery.

isoos:
- related to #4158
- retries the email sending 5 times in the next ~15 hours, then drops the email
- TBD later: batching logic


app/lib/service/email/backend.dart
```diff
@@ -0,0 +1,137 @@
...
+        // successful send, deleting entry
+        tx.delete(o.key);
+        return true;
+      } catch (e, st) {
+        _logger.warning('Email sending failed.', e, st);
+
+        // failed to send, retry later
+        o.attempts++;
+        o.lastAttempted = clock.now().toUtc();
+        o.lastError = e.toString();
```
jonasfj, line 111: here is a problem. You don't know that this means we failed to send the email.

If we get a TCP connection failure after the thing has been sent, we risk going here.. while the email is sent.

app/lib/service/email/backend.dart
```diff
@@ -0,0 +1,137 @@
...
+  /// Returns true if the email was sent successfully.
+  Future<bool> trySendOutgoingEmail(String id) async {
+    return await withRetryTransaction(_db, (tx) async {
+      final o = await tx.lookupOrNull<OutgoingEmail>(
+          _db.emptyKey.append(OutgoingEmail, id: id));
+      if (o == null || o.isDead || !o.mayAttemptNow) {
+        return false;
+      }
+      try {
+        await emailSender.sendMessage(EmailMessage(
```
jonasfj, line 96: Maybe we should generate a `Message-ID` header.
See: https://www.rfc-editor.org/rfc/rfc5322#section-3.6.4

And guidelines from: https://www.jwz.org/doc/mid.html

That way, if we resend it, the server should be able to deduplicate the message (or at-least group them in the same conversation).

app/lib/service/email/models.dart
```diff
@@ -0,0 +1,76 @@
...
+  @db.DateTimeProperty(required: true)
+  DateTime? created;
+
+  /// The total number of attempts.
+  @db.IntProperty(required: true, indexed: false)
+  int attempts = 0;
+
+  /// The timestamp of the last attempt.
+  @db.DateTimeProperty()
+  DateTime? lastAttempted;
```
jonasfj, line 24: Why not `pendingAt`, which you can set to year 3000 when the message is dead and shouldn't be retried again.

app/lib/service/email/backend.dart
isoos: Done.

app/lib/tool/neat_task/pub_dev_tasks.dart
```diff
@@ -17,18 +16,37 @@ import '../../job/backend.dart';
...
 
 final _logger = Logger('pub_dev_tasks');
 
 /// Periodic task that are not tied to a specific service.
 void _setupGenericPeriodicTasks() {
+  // Tries to send pending outgoing emails.
+  _15mins(
+    name: 'send-outgoing-emails',
+    isRuntimeVersioned: false,
+    task: () async => await emailBackend.trySendAllOutgoingEmails(
```
jonasfj, line 37: Let's use a GlobalLock for sanity here...

app/lib/service/email/backend.dart
```diff
@@ -0,0 +1,159 @@
...
+      final o = await tx.lookupOrNull<OutgoingEmail>(
+          _db.emptyKey.append(OutgoingEmail, id: id));
+      if (o == null || o.isDead) {
+        return false;
+      }
+      if (!ignorePendingAt && !o.mayAttemptNow) {
+        return false;
+      }
+      try {
+        await emailSender.sendMessage(EmailMessage(
```
jonasfj, line 111: did you notice that you're sending emails inside a `withRetryTransaction` wrapper? That could have some negative implications.

isoos: Good point, I've separated the DB and email parts, PTAL.

app/lib/service/email/backend.dart
```diff
@@ -0,0 +1,169 @@
...
+    try {
+      await emailSender.sendMessage(EmailMessage(
+        uuid: entry.uuid,
+        EmailAddress(entry.fromEmail!),
+        [EmailAddress(entry.recipientEmail!)],
+        entry.subject!,
+        entry.bodyText!,
+      ));
+
+      await withRetryTransaction(_db, (tx) async {
```
jonasfj, line 119: How about increasing the `attempts` counter, etc.. before we send the email?

Otherwise, if we fail to update datastore we end up retrying it indefinitely, right?

app/lib/service/email/models.dart
```diff
@@ -0,0 +1,69 @@
...
+
+  /// The timestamp of the next attempt.
+  @db.DateTimeProperty(required: true)
+  DateTime? pendingAt;
+
+  /// The error message of the last failure.
+  @db.StringProperty(indexed: false)
+  String? lastError;
```
jonasfj, line 32: Do we need this?

app/lib/service/email/models.dart
```diff
@@ -0,0 +1,69 @@
...
+    // adding an extra
+    pendingAt = clock.now().toUtc().add(Duration(minutes: 1));
+  }
+
+  /// Whether a new attempt can be made to deliver the email.
+  bool get mayAttemptNow => isAlive && clock.now().isAfter(pendingAt!);
+
+  /// Whether we consider the outgoing email alive and try sending.
+  bool get isAlive => attempts < 2;
+  bool get isDead => !isAlive;
```
jonasfj, line 68:
```suggestion
  bool get isNotAlive => !isAlive;
```

So as to not make anyone think that `!isAlive` is a different state from `isDead`.

Honestly, I'd suggest we don't need both. There is one state and it's boolean.

app/lib/service/email/backend.dart
```diff
@@ -0,0 +1,175 @@
...
+    return successful;
+  }
+
+  /// Tries to send emails with the given [ids]. The
+  /// [OutgoingEmail] entry will be deleted after the
+  /// email was sent successfully.
+  ///
+  /// This method should be called right after the entries
+  /// are saved in the Datastore.
+  Future<void> trySendOutgoingEmails(Iterable<String> ids) async {
```
jonasfj, line 86: Could we make this accept a list of `OutgoingEmail`

app/lib/service/email/backend.dart
```diff
@@ -0,0 +1,175 @@
...
+      await _trySendOutgoingEmail(id, ignorePendingAt: true);
+    }
+  }
+
+  /// Tries to send email with the given [id]. The
+  /// [OutgoingEmail] entry will be deleted after the
+  /// email was sent successfully.
+  ///
+  /// Returns true if the email was sent successfully.
+  Future<bool> _trySendOutgoingEmail(
```
jonasfj, line 97: If instead of taking `id` we took `OutgoingEmail`, then we don't need `ignorePendingAt`.

Instead we skip sending if `OutgoingEmail.pendingAt` has changed since the instance passed in `OutgoingEmail`.

This way, by setting `OutgoingEmail.pendingAt` into the future, we'll have locked the row. Thus, there can be no race between the background process re-attempting to send emails, and the initial method invocation that tries to send emails.

Indeed, even if we ran two background processes at the same time, they would never try to send the same email.

app/lib/tool/neat_task/pub_dev_tasks.dart
```diff
@@ -20,18 +17,57 @@ import '../../job/backend.dart';
...
 
 /// Periodic task that are not tied to a specific service.
 void _setupGenericPeriodicTasks() {
+  // Tries to send pending outgoing emails.
+  _15mins(
+    name: 'send-outgoing-emails',
+    isRuntimeVersioned: false,
+    task: () async {
+      final aquireAbort = Completer();
+      final aquireTimer = Timer(Duration(minutes: 2), () {
```
jonasfj, line 44: 2 minutes? But you keep trying to send for 10 minutes?

I'm confused?

isoos: I thought that this is a timeout for the acquire operation, and once we have the lock, we use the 20 minutes expiration (but we keep a soft-limit of 10 minutes to send the queue, so will return the lock after about 10 minutes).
Isn't it the case?

app/lib/service/email/backend.dart
isoos: Done.

Note: I think we can't really rule out double sending the same email if we have two background processes running, but we are using now two independent locks that should minimize the risk of that.

jonasfj (approved):

PR #5929, opened 2022-08-01
Allow name alternatives (e.g. plural form) of moderated (deleted) packages.

### LINKED ISSUE #5928: Consider if the package name similarity check should stop taking moderated names into account.
I propose we stop taking moderated package names into account when doing naming similarity checks [here](https://github.com/dart-lang/pub-dev/blob/ca05a26847ffbbe138950ff31e6ee2a5be6601cb/app/lib/package/backend.dart#L924).

Currently if a package `foobar` is moderated we leave a tombstone preventing a package with the same name.

This is good because it prevents the case where someone depending on `foobar` gets an unrelated (or malicious) package on `dart pub get`.

But also the "similarity check" we have takes the name `foobar` into account making it impossible to publish:

* foobars
* foo_bar
* ...

I question the value of this:

* Existing dependents of `foobar` would not be prone to a spelling mistake.
* In most cases where we moderate a package, it has very little usage - making the risk of mistaking another package for it very little.

cc @jonasfj @isoos @szakarias

isoos:
Fixes #5928

### DIFF
--- a/app/lib/package/name_tracker.dart
+++ b/app/lib/package/name_tracker.dart
@@ -106,13 +106,20 @@ class NameTracker {
 
   void addReservedName(String name) {
     _reservedNames.add(name);
-    _addConflictingName(name);
+    _addConflictingName(name, skipAlternatives: true);
     _names.remove(name);
   }
 
-  void _addConflictingName(String name) {
-    for (final generated in _generateConflictingNames(name)) {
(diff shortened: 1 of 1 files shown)


sigurdm (approved):
LGTM but would like approval from @jonasfj before landing

sigurdm:
@jonasfj agreed offline.

PR #5932, opened 2022-08-02
Upgrade pana and use the repository resolveUrl when available.

isoos:


isoos:
Note: deploying this to staging for further tests. It affects only internal packages (as we are running repository verification only on internal ones).

sigurdm (approved):
LGTM

Issue #5573, opened 2022-03-10
Migrate to Google Identity Services library


jonasfj:
See https://developers.googleblog.com/2022/03/gis-jsweb-authz-migration.html

PR #5934, opened 2022-08-04
Prepare the use of Google Identity Services client library (CSP, experimental flag, init callback).

isoos:
- This is the boilerplate required for #5573, without any functionality.
- The new library will be enabled on staging, when the experimental flag is on. Should not affect prod, even with the experimental flag set.

### DIFF
--- a/app/lib/shared/handler_helpers.dart
+++ b/app/lib/shared/handler_helpers.dart
@@ -9,7 +9,6 @@ import 'package:appengine/appengine.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as path;
 import 'package:pub_dev/frontend/handlers/headers.dart';
-import 'package:pub_dev/shared/env_config.dart';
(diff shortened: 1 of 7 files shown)


app/lib/shared/handler_helpers.dart
```diff
@@ -129,11 +130,18 @@ shelf.Handler _requestContextWrapper(shelf.Handler handler) {
             !hasExperimentalCookie && // don't cache if experimental cookie is enabled
             userSessionData == null; // don't cache if a user session is active
 
+    // The use of the new Google Identity Services library is restricted to
+    // staging or running locally, and it is also behind the experimental flag.
+    final useGisSignIn = isExperimental &&
+        (envConfig.isRunningLocally ||
+            activeConfiguration.projectId == 'dartlang-pub-dev');
```
sigurdm, line 137: Should we somehow encapsulate this condition in a getter? `!isProduction`?...

sigurdm (approved):

Issue #5936, opened 2022-08-08
Admin Tool for adding members to a publisher


sigurdm:
We need such a tool.

sigurdm:
This is no longer urgent.

I think we eventually have such a tool, so I'll leave this open for now.

sigurdm:
We now need this again

budde377:
Started a solution in #6292, anything to get me unblocked! Merging my old account and a new account, removing the package from my publisher, moving my package to a new publisher are all acceptable.

isoos:
This has been done in #6293.

PR #5939, opened 2022-08-08
Admin tool to list publisher members.

isoos:
#5936

### DIFF
--- a/app/lib/admin/tools/publisher_member.dart
+++ b/app/lib/admin/tools/publisher_member.dart
@@ -1,0 +1,18 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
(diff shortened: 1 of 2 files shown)


app/lib/admin/tools/publisher_member.dart
```diff
@@ -0,0 +1,18 @@
...
+
+Future<String> executePublisherMember(List<String> args) async {
+  if (args.length != 1) {
+    return 'Tool to check publisher member info.\n'
+        '  <tools-command> <publisherId> - list current members\n';
+  }
+  final publisherId = args.single;
+
+  final members = await publisherBackend.listPublisherMembers(publisherId);
+  return members.map((e) => '${e.userId} ${e.role} ${e.role}\n').join();
```
sigurdm, line 17: I think the role got duplicated:
```suggestion
  return members.map((e) => '${e.userId} ${e.role}\n').join();
```

sigurdm (approved):
LGTM

PR #5940, opened 2022-08-08
Dynamically load Google Identity Service library.

isoos:
- #5573
- The `onload` introduced in #5934 won't work with our Content Security Policy, unless we enable `unsecure-inline`.
- The new script doesn't support the `?onload=<fn>` parameter that the old one did.
- CSP hashes/nonces does not work with `onload=fn()` attributes, so it seems we need to do this dynamically. If it later turns out that we don't need this callback, we can revert the library loading part to `<script>` element.

### DIFF
--- a/app/lib/frontend/templates/views/shared/layout.dart
+++ b/app/lib/frontend/templates/views/shared/layout.dart
@@ -121,14 +121,7 @@ d.Node pageLayoutNode({
             ),
             d.meta(
                 name: 'google-signin-client_id', content: oauthClientId ?? ''),
-            if (requestContext.experimentalFlags.useGisSignIn)
-              d.script(
-                src: 'https://accounts.google.com/gsi/client',
-                async: true,
-                defer: true,
-                onload: 'pubGsiClientInit()',
-              )
-            else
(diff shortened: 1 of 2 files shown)


ditman:
Hey peeps, I've been working in this migration from the flutter web side, and ended up with a JS-interop layer that you might be able to use (I think @kevmoo is integrating this with `googleapis_auth` on his end):

https://pub.dev/packages/google_identity_services_web

The new SDK doesn't have a programmatic "sign in" method, they want you to render a google sign in button (and I couldn't test that bit, because that doesn't work that well with Flutter UI), but I hope the rest of the library saves you some time.

(I implemented TrustedTypes to embed the script tag of the library, but I have to admit I didn't think about the "insecure-inline" CSP issue, I think this should be reported to the GIS team as a bug)

Feel free to ping me if I can be of assistance!

isoos:
@ditman: thank you, I'm sure this will be helpful for us!

isoos:
Closing as the agreed target solution will be a bit different (also #6405 removes most of the code that this builds on).

PR #5942, opened 2022-08-09
Serve archive_sha256 on client API endpoints.

isoos:
#5880

### DIFF
--- a/pkg/_pub_shared/lib/data/package_api.dart
+++ b/pkg/_pub_shared/lib/data/package_api.dart
@@ -196,6 +196,10 @@ class VersionInfo {
   @JsonKey(name: 'archive_url')
   final String? archiveUrl;
 
+  /// The SHA-256 hash of the canonical archive.
+  @JsonKey(name: 'archive_sha256')
+  final String? archiveSha256;
+
   /// This is an optional field of the API response, it may be `null` or omitted.
   final DateTime? published;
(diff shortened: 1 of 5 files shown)


pkg/_pub_shared/lib/data/package_api.dart
```diff
@@ -196,6 +196,10 @@ class VersionInfo {
   @JsonKey(name: 'archive_url')
   final String? archiveUrl;
 
+  /// The SHA-256 hash of the canonical archive.
+  @JsonKey(name: 'archive_sha256')
```
sigurdm, line 200: I don't know why we would ever get a null here, but if we have I think it would be better not to include the attribute in the json.

```suggestion
  @JsonKey(name: 'archive_sha256', includeIfNull: false)
```

pkg/_pub_shared/lib/data/package_api.dart
```diff
@@ -196,6 +196,10 @@ class VersionInfo {
   @JsonKey(name: 'archive_url')
   final String? archiveUrl;
 
+  /// The SHA-256 hash of the canonical archive.
```
sigurdm, line 199:
```suggestion
  /// The hex-encoded SHA-256 hash of the canonical archive.
```

jonasfj (approved):

PR #5943, opened 2022-08-09
Serve only reduced pubspec content on client API endpoints.

isoos:


sigurdm:
I'm afraid this is too breaking.

We don't know of all usages of the version-listings.

If we want to do this kind of optimization I'm afraid we have to introduce a new api end-point and update the client.

isoos:
Closing as we don't want to do it this way.

PR #5944, opened 2022-08-09
Integrity check ignore missing discontinued flag when package is blocked.

### LINKED ISSUE #5937: Integrity check for abandoned publisher should consider blocked packages
We have a package that triggers this integrity check:

"integrity.check: [pub-integrity-problem] Package "___" has an abandoned publisher, must be marked discontinued."

But the package was marked blocked (it was a spam package) so not sure we also need to mark it discontinued.

isoos:
Fixes #5937.


sigurdm (approved):
LGTM

Issue #4353, opened 2020-12-16
Add sponsorship information on pub.dev


timsneath:
Many package maintainers write their packages as a labor of love, but there's a growing movement to sponsor open source projects, whether to encourage folk in their work or give them a revenue stream sufficient to support their work as a part-time or full-time job. 

So far no single solution is dominant, but some examples of this are [GitHub Sponsors](https://github.com/sponsors), [Patreon](https://www.patreon.com/explore), [BuyMeACoffee](https://www.buymeacoffee.com/), as well as [tidelift](https://tidelift.com/).

[GitHub now supports](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/displaying-a-sponsor-button-in-your-repository) a `FUNDING.yml` page to allow repo authors to share their sponsorship options, and we should promote this on the pub website, since not everyone visits the underlying GitHub repo, and publishers deserve to have their funding model promoted.

/cc @escamoteur, who suggested this offline.

aliyazdi75:
This is an excellent idea. Unfortunately, I can see many good packages don't maintain anymore, and I think one of the main problems is that maintaining the packages is costly, especially for those pals who created a package to spread love in the open-source community without any company to sponsor them.

One of the suggestions could be adding funding like what `npm` did, for example:
```
30 packages are looking for funding
  run `npm fund` for details
```
Thanks for caring.

JaffaKetchup:
I would like to bring attention to this, especially the suggestion @aliyazdi75 made.

mit-mit:
The first thing needed here is to get the metadata. I suggest we expand the `pubspec.yaml` with a new field `funding`, which below it has a list of strings containing Urls, e.g:

```
funding:
  - "https://www.buymeacoffee.com/superduperwidget"
  - "https://github.com/sponsors/superwidget"
  - "https://patreon.com/superduperflutterwidget"
 ```
 
 We can then add that to the API, and show the links in the sidebar:
 
<img width="157" alt="Screen Shot 2022-08-08 at 13 37 06" src="https://user-images.githubusercontent.com/13644170/183414516-fd5c5203-7e89-4a99-8aff-7d18b1865efb.png">

JaffaKetchup:
I'm in support of something like that. GitHub already uses a similar suggestion.

isoos:
Instead of *donating*, maybe "Consider *supporting* this project:"?

sigurdm:
[...]

PR #5957, opened 2022-08-17
Minimal checks for top-level funding key.

isoos:
#4353


pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -718,3 +719,29 @@ Iterable<ArchiveIssue> checkScreenshots(
...
+      yield ArchiveIssue(
+          'Invalid `funding` value (`$item`): only URLs are allowed.');
+      continue;
+    }
+    final uri = Uri.tryParse(item.trim());
+    if (uri == null || uri.scheme != 'https') {
+      yield ArchiveIssue(
+          'Invalid `funding` value (`$item`): only `https` URLs are allowed.');
+      continue;
+    }
```
jonasfj, line 745: For sanity let's also limit the length of the URL. Let's just say no more than 255 characters or something like that?

We don't want garbage in here...

jonasfj (approved):

PR #5958, opened 2022-08-17
Add material FAB

szakarias:


app/lib/frontend/dom/material.dart
```diff
@@ -54,6 +54,44 @@ d.Node button({
   );
 }
 
+/// Renders a material floating action button (FAB) element.
+/// The FAB is 48x48px by default. If [fabMini] is `true` the FAB will be
+/// 40x40px.
+//
+// ignore: non_constant_identifier_names
+d.Node FAB({
```
jonasfj, line 62:
```suggestion
d.Node floatingActionButton({
```

jonasfj (approved):

app/lib/frontend/dom/material.dart
szakarias: Done.

third_party/material/app.scss
```diff
@@ -9,3 +9,5 @@
 @use '@material/ripple/mdc-ripple';
 @use '@material/select/mdc-select';
 @use '@material/textfield/mdc-text-field';
+@use "@material/fab";
```
isoos, line 12: Let's keep the alphabetical order, and also the detailed pattern here (`@material/fab/mdc-fab`?).

third_party/material/app.scss
```diff
@@ -9,3 +9,5 @@
...
+@include fab.core-styles;
```
isoos, line 13: Is this needed? The reference file uses only fab:
https://github.com/material-components/material-components-web/blob/master/packages/material-components-web/material-components-web.scss#L35

third_party/material/initialize.sh
```diff
@@ -14,6 +14,7 @@ EOF
 
 npm install "@material/auto-init@${VERSION}"
 npm install "@material/button@${VERSION}"
+npm install "@material/fab@${VERSION}"
```
isoos, line 17: Don't we also need a JS initialization for this? Maybe not, as I couldn't find any traces here:
https://github.com/material-components/material-components-web/blob/master/packages/material-components-web/index.ts

isoos (changes requested):

third_party/material/app.scss
szakarias: SG.

third_party/material/app.scss
szakarias: It seems not when `/mdc-fab` is added in the end of the `@use`. 

In https://material.io/components/buttons-floating-action-button/web#using-fabs it said 
```
@use "@material/fab";

@include fab.core-styles;
```

But ``` @material/fab/mdc-fab ``` seems to work so let's keep that. :-)

third_party/material/initialize.sh
szakarias: I think not. In my example it seems to work without initialization.

isoos (approved):

PR #5961, opened 2022-08-18
Display funding links in the package's info box.

isoos:
#4353


app/lib/frontend/templates/package.dart
```diff
@@ -88,10 +88,15 @@ d.Node renderPkgInfoBox(PackagePageData data) {
     addLink(dartdocsUrl, 'API reference', documentation: true);
   }
 
+  final fundingLinks = data.version!.pubspec!.funding.map((uri) {
+    // TODO: use `rel="nofollow noopener ugc"` if link verification failed
+    return InfoBoxLink(uri.toString(), uri.host, rel: 'ugc');
```
jonasfj, line 93: Why is this different?

app/lib/package/model_properties.dart
```diff
@@ -173,6 +173,21 @@ class Pubspec {
...
+    final list = _json!['funding'];
+    if (list is! List) {
+      return <Uri>[];
+    }
+    return list
+        .map((e) {
+          if (e is! String) return null;
+          return Uri.tryParse(e);
+        })
+        .whereType<Uri>()
```
jonasfj, line 188:
```suggestion
        .whereNotNull()
```

It exists in `package:collection`

app/lib/tool/test_profile/import_source.dart
```diff
@@ -146,7 +146,10 @@ class _AutoGeneratedImportSource implements ImportSource {
           'path': '${screenshot.path}',
           'description': '${screenshot.description}'
         }
-      ]
+      ],
+      'funding': [
```
jonasfj, line 150: Can we make optional, so that not all test packages have this? Maybe just the ones named `oxygen` ?

jonasfj (approved):

app/lib/tool/test_profile/import_source.dart
isoos: I'd like to semi-randomize it (same with screenshots and other items), but probably in a separate PR, because it would impact a lots of tests.

PR #5975, opened 2022-08-23
Connecting the JWT with the automatic publishing settings (but not enabling it yet)

isoos:
- #5769
- Simpler `isPackageAdmin`, as the concern for upload and the admin right is separate, treating it seaprately.
- GitHub OpenId data is not yet mockable in tests, planning a follow-up PR to do it.
- GitHub publishing ends with 403 (when we do not recognize something) or a PackageRejection, which reflects back the payload (for further debugging).

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -11,15 +11,16 @@ import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 // ignore: import_of_legacy_library_into_null_safe
 import 'package:neat_cache/neat_cache.dart';
-import 'package:pub_dev/account/agent.dart';
-import 'package:pub_dev/service/openid/jwt.dart';
 
 import '../package/models.dart';
+import '../service/openid/github_openid.dart';
+import '../service/openid/jwt.dart';
 import '../shared/datastore.dart';
 import '../shared/exceptions.dart';
 import '../shared/redis_cache.dart' show cache, EntryPurgeExt;
 import '../shared/utils.dart';
 
+import 'agent.dart';
 import 'auth_provider.dart';
 import 'models.dart';
 import 'session_cookie.dart' as session_cookie;
@@ -190,7 +191,7 @@ Future<AuthenticatedAgent> requireAuthenticatedAgent(
   if (token == null || token.isEmpty) {
     throw AuthenticationException.authenticationRequired();
   }
-  final authenticatedService = _tryAuthenticateGithubAction(token);
+  final authenticatedService = await _tryAuthenticateGithubAction(token);
   if (authenticatedService != null) {
     return authenticatedService;
   } else {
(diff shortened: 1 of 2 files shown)


app/lib/account/backend.dart
```diff
@@ -213,10 +215,19 @@ AuthenticatedGithubAction? _tryAuthenticateGithubAction(String token) {
   if (payload == null) {
     return null;
   }
-  // TODO: check the audience
-  // TODO: check signature from JWKS
-  // TODO: when everything is verified, return the JWT token.
-  return null;
+  if (payload.aud != 'https://pub.dev/') {
```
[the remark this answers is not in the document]
isoos: fixed

app/lib/account/backend.dart
```diff
@@ -213,10 +215,19 @@ AuthenticatedGithubAction? _tryAuthenticateGithubAction(String token) {
   if (payload == null) {
     return null;
   }
```
jonasfj, line 217: I propose we do an explicit `iss` check here.

I'm not sure I mind throwing `FormatException` in `GitHubJwtPayload.tryParse` if the issuer has mismatch, but it also seems a bit wrong to throw `FormatException` due to a validation issue.

This seems like a place where an issuer check should take place, so maybe just duplicate it here, or move it here for clarity.

app/lib/package/backend.dart
```diff
@@ -1007,22 +997,15 @@ class PackageBackend {
...
 
+      // If the package exists, check for authorization
+      if (package != null) {
+        await _checkUploadAuthorization(package!, agent);
+      }
+
       // If the package does not exist, then we create a new package.
       if (package == null) {
         _logger.info('New package uploaded. [new-package-uploaded]');
         package = Package.fromVersion(newVersion);
```
jonasfj, line 1008: I don't think we allow bots to publish the first version!

We probably should consider following this up with a mechanism for publishers to create packages without any versions (but that seems like a follow up).

app/lib/package/backend.dart
```diff
@@ -1140,6 +1123,50 @@ class PackageBackend {
...
+  Future<bool> _isGithubActionAllowed(
+      Package package, AuthenticatedGithubAction agent) async {
+    final githubPublishing = package.automatedPublishing.github;
+    if (githubPublishing == null) {
+      return false;
+    }
+    final isEnabled = githubPublishing.isEnabled ?? false;
+    if (!isEnabled) {
+      return false;
+    }
```
jonasfj, line 1151: nit
```suggestion
    if (githubPublishing == null || githubPublishing.isEnabled ?? false) {
      return false;
    }
```

jonasfj (approved):

app/lib/package/backend.dart
isoos: Bots won't be able to publish, because there won't be any `AutomatedPublishing` object on the non-existing/newly created `Package` entity. Maybe there could be another configuration elsewhere that pre-authorizes the agent to publish a new package, but that is for another day.

app/lib/account/backend.dart
jonasfj: Yes, my suggestion was that checking the issuer is a validation step, that should not happen in a `tryParse` function. It's all nits.

isoos: I'll do a follow-up PR, as I also have some other refactors pending.

PR #5978, opened 2022-08-24
Limit /api/documentation/<package> endpoint to the latest few versions.

isoos:


app/lib/frontend/handlers/custom_api.dart
```diff
@@ -45,8 +45,14 @@ Future<shelf.Response> apiDocumentationHandler(
     return jsonResponse({}, status: 404);
   }
 
+  // Limit versions to the latest few (sorted semantically).
+  final versionsToQuery = <String>{
+    versions.latest.version,
+    ...versions.versions.map((e) => e.version).toList().reversed.take(10),
```
jonasfj, line 51: What about trying a sanity limit like 30 ?

jonasfj (approved):

jonasfj (approved):

PR #5982, opened 2022-08-24
Upgrade dependencies

jonasfj:
In particular I want googleapis, it has new nice fields I want to set :D


isoos (approved):

isoos:
Maybe add a note in changelog if they seem to be relevant

Issue #5988, opened 2022-08-25
"Unlisted packages" or "Hidden packages" tab on publishers page


jonasfj:
There are some publishers that don't like unlisted / discontinued packages being displayed on the publishers page.

Perhaps we should hide them by default, and offer a "Unlisted Packages" tab, which we only display to admins of the publisher page. Not because unlisted packages are hidden from other users, just that don't show up in search package listings by default.

![image](https://user-images.githubusercontent.com/149732/186665510-ce8391b6-8925-401a-b06f-acafaeb1b7d6.png)

jonasfj:
cc/ @mit-mit

isoos:
The current `Packages` tab only lists the top 10 packages and then both the sorting control or the pagination would redirect the user to the search page (with `publisher:<x>` predicate). I am assuming this won't change for now (except we won't have the `show:hidden` predicate there).

Should the `Unlisted packages` include the ones that are `discontinued` too?

sigurdm:
> Should the Unlisted packages include the ones that are discontinued too?

IMO yes, you need a way of seeing all your packages.

isoos:
> IMO yes, you need a way of seeing all your packages.

So, let's call it `Hidden packages`, and the URL would be `/publishers/<publisherId>/hidden-packages` ?

mit-mit:
I think "hidden" sounds too strong; it may create the illusion that packages on that tab are actually hidden.

isoos:
> I think "hidden" sounds too strong; it may create the illusion that packages on that tab are actually hidden.

We have the same terminology on the search page, I think it won't cause much confusion. (Or the same confusion may be present on search too?)

PR #5991, opened 2022-08-26
Add is:unlisted tag to discontinued/legacy packages.

isoos:
Required for #5988.

### DIFF
--- a/pkg/_pub_shared/lib/search/tags.dart
+++ b/pkg/_pub_shared/lib/search/tags.dart
@@ -19,9 +19,6 @@ const allowedTagPrefixes = [
 
 /// Collection of package-related tags.
 abstract class PackageTags {
-  /// Package is marked discontinued | unlisted | legacy.
-  static const String isHidden = 'is:hidden';
-
   /// Package is shown, regardless of its hidden status.
   static const String showHidden = 'show:hidden';
 
--- a/pkg/_pub_shared/lib/search/tags.dart
+++ b/pkg/_pub_shared/lib/search/tags.dart
@@ -28,7 +28,7 @@ abstract class PackageTags {
   /// Package is shown, regardless of its discontinued status.
   static const String showDiscontinued = 'show:discontinued';
 
-  /// Package is marked unlisted.
+  /// Package is marked unlisted, discontinued, or is a legacy package.
   static const String isUnlisted = 'is:unlisted';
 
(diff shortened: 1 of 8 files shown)


jonasfj:
@isoos can't we repurpose the `is:unlisted` tag?

jonasfj:
We could also do `is:unlisted OR is:discontinued`.

But I think it might be okay to say that discontinued packages are unlisted.

jonasfj:
Actually, in the admin UI, you can't set `unlisted` if you've set `discontinue`. So I think it's fair if we argue that `discontinued` is a subsets of `unlisted`.

jonasfj:
Then we should probably also tweak the search filter is `include unlisted` in the advanced section.

isoos:
> Actually, in the admin UI, you can't set unlisted if you've set discontinue. So I think it's fair if we argue that discontinued is a subsets of unlisted.

But you can set `unlisted` without setting it discontinued, so repurposing it will lose some logical correctness.

> Then we should probably also tweak the search filter is include unlisted in the advanced section.

The current `show:hidden` enables both discontinued, unlisted and legacy (Dart 1.x). I think we should have a filter that enables "search everything" in one way or another.

jonasfj:
Maybe, we should update the UI for setting something as "discontinued" to say "set discontinue and unlisted", or something like that?

Maybe just in the explanatory text, such that people know that a discontinued package is always also unlisted.

Won't that remove the cognitive difference, and make `is:unlisted` a subsets of packages are that discontinued or unlisted?


-----------------

Context: I dislike the word "hidden", it gives connotations that makes a user think they are not visible to others, which is just not true. So using this word can cause surprises -- which in this context is a privacy issue.

isoos:
Refactored with the intent that we will change "Include hidden" to "Include unlisted".
PTAL.

pkg/_pub_shared/lib/search/tags.dart
```diff
@@ -19,9 +19,6 @@ const allowedTagPrefixes = [
 
 /// Collection of package-related tags.
 abstract class PackageTags {
-  /// Package is marked discontinued | unlisted | legacy.
-  static const String isHidden = 'is:hidden';
```
jonasfj, line 23: nit: Need to update the comment for `isUnlisted`.

jonasfj (approved):

PR #5992, opened 2022-08-26
Parse and use sort:<order> in search input text.

isoos:


app/lib/search/search_service.dart
```diff
@@ -244,10 +244,13 @@ class ServiceSearchQuery {
     return map;
   }
 
+  late final effectiveOrder = parsedQuery.order ?? order;
```
jonasfj, line 247: Can this have some documentation? what is `order` and `effectiveOrder`, do we ever need both?

jonasfj (approved):

app/lib/search/search_service.dart
isoos: We need it until the URL-parameter order is removed from the supported query.

PR #5997, opened 2022-08-29
Better handle ModeratedPackage names in NameTracker.

isoos:
- Do not generate conflicting names (including the reduced name) for moderated packages.
- Renamed fields and methods to make it explicit that they store the moderated names.
- Added test case for the current issue.

### DIFF
--- a/app/lib/package/name_tracker.dart
+++ b/app/lib/package/name_tracker.dart
@@ -86,7 +86,7 @@ class NameTracker {
   List<TrackedPackage>? _packagesOrderedByLastPublishedDesc;
 
   /// Names that are reserved due to moderated packages having these names.
-  final _reservedNames = <String>{};
+  final _moderatedNames = <String>{};
   final _conflictingNames = <String, String>{};
   final _firstScanCompleter = Completer();
   _NameTrackerUpdater? _updater;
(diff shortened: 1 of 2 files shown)


app/lib/package/name_tracker.dart
```diff
@@ -104,20 +104,20 @@ class NameTracker {
...
-    _names.remove(name);
+  void addModeratedName(String name) {
+    _moderatedNames.add(name);
+    final existed = _names.remove(name);
+    if (existed) {
+      // reset conflicting names
+      _conflictingNames.removeWhere((key, value) => value == name);
+      for (final name in _names) {
+        _addConflictingName(name);
+      }
```
jonasfj, line 115: Why is this necessary, are we cache state forever across runtimeVersions? that sounds dangerous.

Maybe we should fix the TODO, instead:
> /// TODO: support remove and re-scan package names every day or so.

isoos: This is a cache for the local instance only, and runtimeVersions don't play much role in it.

We could re-scan daily, however, there are alternatives:
- This reset of the values when a ModeratedPackage first appears to the instance.
- With a bit more memory, we could track multiple source names for each conflicting alternative, and then removal would be much cheaper, and doesn't need re-calculation.

As package moderation is not a common event, I think we are better off with this, but I could also imagine daily re-scan or maybe a bit more memory use would be better in complexity.

jonasfj: > As package moderation is not a common event, I think we are better off with this,

How is daily rescan not easier?

I just see a code path that rarely gets exercised.

Also we don't care if it takes 24 hours before a name can be reused...

isoos: > How is daily rescan not easier?

It will need to have a separate data structure that we build on scanning, and some maintenance around triggering the scan. Not super-complex, but more lines of code than now..

isoos:
Closing as #6017 will be better.

PR #6002, opened 2022-08-30
Testing the GitHub Actions OpenID token parser and verifier on GitHub Actions.

isoos:


.github/workflows/token-test.yml
```diff
@@ -0,0 +1,35 @@
...
+  pull_request:
+  schedule:
+    - cron: "0 0 * * 0"
+defaults:
+  run:
+    shell: bash
+env:
+  PUB_ENVIRONMENT: bot.github
+permissions:
+  id-token: write
```
jonasfj, line 17: We don't need it here, right?

app/test/service/openid/github_openid_test.dart
```diff
@@ -49,4 +54,70 @@ void main() {
...
+      // example `ref`: `refs/pull/38/merge`
+      expect(payload.ref, startsWith('refs/'));
+      expect(payload.ref, anyOf(contains('/pull/')));
+
+      // TODO: decide if we need to run checks on these values
+      // example `sub`: `repo:dart-lang/pub-dev:pull_request`
+      expect(token.payload['sub'], startsWith('repo:dart-lang/pub-dev:'));
+      expect(token.payload['repository_visibility'], 'public');
+      expect(token.payload['actor'], isNotEmpty);
+    });
```
jonasfj, line 121: Make it skip: if `ACTIONS_ID_TOKEN_REQUEST_URL` isn't set...

app/test/service/openid/github_openid_test.dart
```diff
@@ -49,4 +54,70 @@ void main() {
     expect(firstRsaKey.n, isNotEmpty);
     expect(firstRsaKey.e, isNotEmpty);
   });
+
+  if (Platform.environment['CI'] == 'true') {
```
jonasfj, line 58: use the tag presubmit only:
https://github.com/dart-lang/pub-dev/blob/master/app/dart_test.yaml#L2

app/test/service/openid/github_actions_id_token_test.dart
```diff
@@ -0,0 +1,82 @@
...
+    fn: () async {
+      expect(actionsIdTokenRequestUrl, isNotEmpty);
+      expect(actionsIdTokenRequestToken, isNotEmpty);
+
+      final uri = Uri.parse(actionsIdTokenRequestUrl);
+      final rs = await http.get(
+        uri.replace(
+          queryParameters: {
+            ...uri.queryParameters,
+            'audience': 'https://pub.dev',
```
jonasfj, line 33: Can we use a dummy audience? so we don't leak something that could potentially be used.

app/test/service/openid/github_actions_id_token_test.dart
```diff
@@ -0,0 +1,82 @@
...
+import 'package:test/test.dart';
+
+import '../../shared/test_services.dart';
+
+void main() {
+  final actionsIdTokenRequestUrl =
+      Platform.environment['ACTIONS_ID_TOKEN_REQUEST_URL'] ?? '';
+  final actionsIdTokenRequestToken =
+      Platform.environment['ACTIONS_ID_TOKEN_REQUEST_TOKEN'] ?? '';
+  testWithProfile(
```
jonasfj, line 21:
```suggestion

  // This test only works when running on Github Actions with `id-token: write` permissions.
  //
  // This test creates a id-token, not for talking to an external service,but for testing that we
  // are able to verify and validate the token.
  testWithProfile(
```

PR #6015, opened 2022-09-02
Show screenshot thumbnail on package page

szakarias:


szakarias:
image

isoos:
I'm wondering if the thumbnail would be better at the end of the metadata section. With that, we would start the info box with the scores and the publisher (both pub.dev-specific), and then continue with user-provided information, a short description and then the screenshots. wdyt?

app/lib/frontend/templates/views/pkg/screenshots.dart
```diff
@@ -0,0 +1,56 @@
...
+    children: [prev, imageContainer, next],
+  );
+}
+
+d.Node screenshotThumbnailNode(
+    String thumbnailUrl, List<String>? screenshotUrls) {
+  final collectionsIconWhite =
+      staticUrls.getAssetUrl('/static/img/collections_white_24dp.svg');
+  return d.div(attributes: {
+    'thumbnail-data': screenshotUrls!.reduce((a, b) => '$a,$b'),
```
isoos, line 46: maybe `screenshotUrls!.join(',')` ?

pkg/web_app/lib/src/screenshot_carousel.dart
```diff
@@ -57,7 +57,7 @@ void _setEventForScreenshot() {
   for (final thumbnail in thumbnails) {
     thumbnail.onClick.listen((event) {
       showElement(carousel);
-      images = thumbnail.attributes['data']!.split(',');
+      images = thumbnail.attributes['thumbnail-data']!.split(',');
```
isoos, line 60: I think this should rather be `data-<x>` attribute, that can be accessed via `Element.dataset['x']`.

app/lib/frontend/templates/views/pkg/screenshots.dart
szakarias: done

pkg/web_app/lib/src/screenshot_carousel.dart
szakarias: Definitely! :-) Done.

jonasfj (approved):

szakarias:
Landing this as per discussion, it is still behind experimental. We will probably move the placement of the thumbnail once we land on a final design.

PR #6017, opened 2022-09-02
Rewriting NameTracker: daily full scan, better handling ModeratedPackage.

isoos:
- An alternative to #5997.
- New timer for full daily re-scan of the Datastore for packages.
- Packages are scanned at startup and updates are running every 15 minutes + we may also run an update when a new version is published.
- Do not generate conflicting names (including the reduced name) for moderated packages.
- Added test case for the current issue.
-


jonasfj (approved):

app/lib/fake/server/fake_default_service.dart
```diff
@@ -44,7 +44,7 @@ class FakePubServer {
           }
 
           await generateFakePopularityValues();
-          nameTracker.startTracking();
+          await nameTracker.startTracking();
```
jonasfj, line 47: Should we have a timeout?

Do we need to handle exceptions here?

jonasfj: Let's make it a 5 minute timeout.

PR #6019, opened 2022-09-05
Adjust id-token test.

isoos:
Some of the items in #6002 are not the same in the `push` event as I've expected from the `pull_request` event, adjusting the test to handle it.

### DIFF
--- a/app/lib/service/openid/jwt.dart
+++ b/app/lib/service/openid/jwt.dart
@@ -270,7 +270,6 @@ class GitHubJwtPayload {
     'event_name',
     'ref',
     'ref_type',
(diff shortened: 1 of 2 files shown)


app/lib/service/openid/jwt.dart
```diff
@@ -270,7 +270,6 @@ class GitHubJwtPayload {
     'event_name',
     'ref',
     'ref_type',
-    'environment',
```
jonasfj, line 273: FYI, `aud` above is not a github specific claim, it's part of the spec.

jonasfj (approved):

PR #6030, opened 2022-09-08
Debug JWT timestamps in GitHub Actions test.

isoos:


app/lib/service/openid/jwt.dart
```diff
@@ -217,16 +217,26 @@ class JwtPayload extends UnmodifiableMapView<String, dynamic> {
...
-    if (iat != null && iat!.isAfter(now)) {
-      return false;
-    }
-    if (nbf != null && nbf!.isAfter(now)) {
-      return false;
-    }
-    if (exp != null && exp!.isBefore(now)) {
+
+    bool isNullOrAIsBeforeB(String name, DateTime? a, DateTime? b) {
+      if (a == null || b == null) {
```
jonasfj, line 222: We should never allow null, please leave a comment for us to not allow that in production.

jonasfj (approved):
I'm guessing `iat/nbf` always should be tested with an assumed 15 min drift... 

`exp` is probably less sensitive to clock drift..

app/lib/service/openid/jwt.dart
isoos: I've updated the code to reject if any is `null`.

PR #6032, opened 2022-09-09
Migrate from show:hidden -> show:unlisted.

isoos:
Part of #5988.


isoos:
Note: this is now deployed on staging.

jonasfj (approved):

PR #6033, opened 2022-09-09
Fix fake_time flaky test: skip periodic timer callback when it is no longer active.

isoos:
I have the suspicion that the condition + `continue` will fix the flakes, but couldn't really confirm it locally yet.

### DIFF
--- a/app/test/task/fake_time.dart
+++ b/app/test/task/fake_time.dart
@@ -252,6 +252,11 @@ class _FakeTime extends FakeTime {
 
     // Trigger the callback for the pending timer.
     for (final triggeredTimer in triggeredTimers) {
+      // Skip running periodic timers if they have been cancelled.
+      // TODO: review/refactor trigger sequence so that this doesn't happen
+      if (triggeredTimer._isPeriodic && !triggeredTimer.isActive) {
+        continue;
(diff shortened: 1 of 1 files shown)


app/test/task/fake_time.dart
```diff
@@ -252,6 +252,11 @@ class _FakeTime extends FakeTime {
 
     // Trigger the callback for the pending timer.
     for (final triggeredTimer in triggeredTimers) {
+      // Skip running periodic timers if they have been cancelled.
+      // TODO: remove if the exception doesn't show up in flaky tests
+      if (triggeredTimer._isPeriodic && !triggeredTimer.isActive) {
+        throw Exception('Periodic timer is no longer active when running it.');
```
jonasfj, line 258:
```suggestion
        throw StateError('Periodic timer is no longer active when running it.');
```
or `AssertionError` if this an internal error in the `fake_time` library.

jonasfj (approved):

app/test/task/fake_time.dart
isoos: Oh, since then I was able to confirm that this is indeed the case for the flakes, updating the PR soon.

PR #6034, opened 2022-09-12
JWT time check with 2 minutes threshold.

isoos:
- renamed method to `isTimely`
- token test uses only a 1-minute threshold, regular backend uses 2-minutes threshold - let's increase it only if we need to
- added `schedule` to the event names in token test (when it is triggered from cron)

### DIFF
--- a/app/lib/service/openid/jwt.dart
+++ b/app/lib/service/openid/jwt.dart
@@ -214,9 +214,8 @@ class JwtPayload extends UnmodifiableMapView<String, dynamic> {
   ///
   /// Returns `false` if the current timestamp is outside of the allowed range.
   /// If the timestamp is missing, we treat it as if it were expired/invalid.
-  bool isTimely({DateTime? now, Duration? threshold}) {
+  bool isTimely({DateTime? now, Duration threshold = Duration.zero}) {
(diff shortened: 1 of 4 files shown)


app/lib/service/openid/jwt.dart
```diff
@@ -214,30 +214,26 @@ class JwtPayload extends UnmodifiableMapView<String, dynamic> {
...
     bool isABeforeB(String name, DateTime? a, DateTime? b) {
       if (a == null || b == null) {
-        // TODO: remove debug message after the appropriate difference threshold is selected.
-        print('$name is missing.');
+        _logger.info('JWT does not have "$name" field.');
         return false;
       }
       if (a.isBefore(b) || a == b) {
         return true;
       }
```
jonasfj, line 228: Don't have two kinds of comparisons.

Just do:
```dart
return a.add(threshold).isBefore(b);
```

jonasfj: Having multiple comparisons, means that we'll have more code flows... and more options for flipping `<` with `>` and making a mistake. Let's keep it simple.

jonasfj (approved):

PR #6047, opened 2022-09-15
Prevent concurrent updates in NameTracker recent scans.

isoos:


sigurdm (commented):
It seems we are implementing a lot of reusable logic here...
Could we lean on neat_periodic_task to do the scheduling?

isoos:
> Could we lean on neat_periodic_task to do the scheduling

That is more about cluster-wide scheduling, this is about in-isolate coordination, but maybe I'm wrong.

sigurdm:
> That is more about cluster-wide scheduling, this is about in-isolate coordination, but maybe I'm wrong.

Ah - Ok I get it. We have to do this on every instance. Not once per deployment...
I guess we can go ahead with this, still I think the coordinating logic could be factored out somehow. Maybe using neat_periodic_task but with an in-memory backing (the default: https://github.com/google/dart-neats/blob/master/neat_periodic_task/lib/neat_periodic_task.dart#L160).

jonasfj (approved):

PR #6051, opened 2022-09-15
Single task blob index

jonasfj:


pkg/_pub_shared/lib/data/task_api.dart
```diff
@@ -9,50 +9,37 @@ part 'task_api.g.dart';
...
-/// This gives us some atomicity when uploading results, until the dartdoc index
-/// file is uploaded, we will keep using the old dartdoc blob file.
+/// It is important that the [blob] is uploaded before the [index], and that the
+/// contents of the index references the given [blobId]. The filename of [index]
+/// is static, while the filename of the [blob] is the randomized [blobId].
+/// This ensures that if two workers uploading results of the same package, then
+/// both will be able to upload the [blob] file, but when they upload the
+/// [index] file, one will overwrite the other. Thus, by encoding the [blobId]
+/// in [index] we ensure consistency. Using the mismatching [index]/[blob] when
+/// reading data is bound to yield garbage.
```
isoos, line 19: Note: is there a way in which we can prevent ourselves to misuse the blobId when reading the data?

isoos (approved):

pkg/_pub_shared/lib/data/task_api.dart
jonasfj: What do you mean?

PR #6056, opened 2022-09-16
Increase search cache TTL to 5 minutes.

isoos:


sigurdm:
What is the current TTL?

Do we have an estimate of how this will affect our redis memory usage? We are already using a sizable amount of our capacity.

isoos:
current TTL: 1 minute

> Do we have an estimate of how this will affect our redis memory usage? We are already using a sizable amount of our capacity.

Not sure, and we don't really have any breakdown on which part of the cache is taking how much memory. :(

jonasfj (approved):

PR #6076, opened 2022-09-23
Increase  thumbnail on package detail page and center image

szakarias:
Increase screenshot thumbnail size on the package detail page and center image in thumbnail if the ratio is not 1:1. The collections icon will stay in the bottom right corner.


szakarias:
image

szakarias:
[cast.webm](https://user-images.githubusercontent.com/3865266/191954049-413c2774-7fd7-4ee1-aa81-a688c5980780.webm)

pkg/web_css/lib/src/_list.scss
```diff
@@ -128,15 +128,26 @@
...
-  flex-grow: 0;
-  flex-shrink: 0;
-  flex-basis: 30px;
+  height: 30px;
   width: 30px;
-  align-self: flex-end;
-  margin-left: -30px;
   opacity: 70%;
   background-color: #aeaeae;
+  margin-top: auto;
```
isoos, line 140: maybe a `position: relative; left: -30px` could be used here, but I'm fine with this too

isoos (approved):

pkg/web_css/lib/src/_list.scss
szakarias: I cannot make it work with that, so I'll keep this since you are okay with this too.

PR #6085, opened 2022-09-27
Fixing OutgoingEmail entity, sending and retries.

### LINKED ISSUE #6082: Publishing fails with new backend because of too many entities being saved.
```
Error: ApplicationError: operating on too many entity groups in a single transaction.
    at GrpcDatastoreImpl.commit (package:appengine/src/grpc_api_impl/datastore_impl.dart:113:9)
    at StackZoneSpecification._registerUnaryCallback.<fn> (package:stack_trace/src/stack_zone_specification.dart:
```

isoos:
- Fixes #6082.
- All recipients are stored in the single `OutgoingEmail` entry.
- On a sending attempt, we set a random `claimId`, and clear it when the attempt is completed. A next attempt with an active `claimId` won't proceed, and the GC process will clear it.

### DIFF
--- a/app/lib/service/email/backend.dart
+++ b/app/lib/service/email/backend.dart
@@ -7,6 +7,7 @@ import 'dart:math';
 import 'package:clock/clock.dart';
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:logging/logging.dart';
+import 'package:pub_dev/shared/utils.dart';
 
 import '../../frontend/email_sender.dart';
 import '../../shared/datastore.dart';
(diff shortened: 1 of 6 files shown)


app/lib/service/email/backend.dart
```diff
@@ -62,84 +60,81 @@ class EmailBackend {
...
-      await withRetryTransaction(_db, (tx) async {
-        final o = await tx.lookupOrNull<OutgoingEmail>(key);
-        if (o == null) {
-          return;
-        }
-        o.lastError = e.toString();
+    await withRetryTransaction(_db, (tx) async {
+      final o = await tx.lookupOrNull<OutgoingEmail>(key);
+      if (o == null || o.claimId != claimId) {
+        return;
```
jonasfj, line 125: This needs to log something with level shout or something... because this is really bad!

app/lib/service/email/backend.dart
```diff
@@ -62,84 +60,81 @@ class EmailBackend {
...
+      }
+      for (final email in sent) {
+        o.recipientEmails?.remove(email);
+      }
+      if (o.recipientEmails?.isEmpty ?? false) {
+        tx.delete(key);
+      } else {
+        o.claimId = null;
```
jonasfj, line 133: don't you need to bump `pendingAt` ?

app/lib/service/email/backend.dart
```diff
@@ -62,84 +60,81 @@ class EmailBackend {
...
+    final claimId = createUuid();
     final entry = await withRetryTransaction(_db, (tx) async {
       final o = await tx.lookupOrNull<OutgoingEmail>(key);
-      if (o == null || o.isNotAlive) {
+      if (o == null || o.isNotAlive || o.claimId != null) {
         return null;
       }
       o.attempts++;
       o.lastAttempted = now;
       // retry after a random delay in the next 2-6 hours
```
jonasfj, line 94:
```suggestion
      // retry after a random delay in the next 2-6 hours, if and only if,
      // o.claimId has been cleared. We never retry sending emails if we don't know
      // if the email was sent or not (because we don't want to send it multiple times)
```

jonasfj (approved):

PR #6090, opened 2022-09-28
Expose GitHub JWT issue when used for publishing.

isoos:
#5769

### DIFF
--- a/app/lib/shared/exceptions.dart
+++ b/app/lib/shared/exceptions.dart
@@ -322,6 +322,10 @@ class AuthenticationException extends ResponseException {
   factory AuthenticationException.accessTokenInvalid() =>
       AuthenticationException._('Invalid `accessToken`.');
 
+  /// Signaling that the GitHub JWT token was missing a field or could not be authenticated.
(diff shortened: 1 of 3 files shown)


app/lib/shared/exceptions.dart
```diff
@@ -424,6 +428,11 @@ class AuthorizationException extends ResponseException {
         '`pub.dev` from verifying that you own the domain.',
       ].join('\n'));
 
+  /// Signaling that the GitHub JWT token was missing a field or could not
+  /// be authorized with the current configuration.
+  factory AuthorizationException.githubTokenIssue(String message) =>
+      AuthorizationException._('GitHub token issue: $message.');
```
jonasfj, line 434:
```suggestion
  /// Signaling that the GitHub JWT token could not
  /// be authorized with the current configuration.
  factory AuthorizationException.githubActionIssue(String reason) =>
      AuthorizationException._('The calling GitHub Action is not allowed to publish, because: $reason.\nSee https://dart.dev/go/publishing-from-github');
```

jonasfj (approved):
Just nits :D

PR #6091, opened 2022-09-29
Do not fail hard when decoding data from older ScoreCard.

isoos:


app/lib/scorecard/models.dart
```diff
@@ -93,7 +95,24 @@ class ScoreCard extends db.ExpandoModel<String> {
...
+  /// Tries to decode and return the data stored on the [ScoreCard].
+  ///
+  /// May return `null` when the stored data from an older runtime
+  /// is not compatible with the current runtime.
+  ScoreCardData? tryDecodeData() {
+    try {
+      return _toData();
+    } catch (e, st) {
+      if (runtimeVersion == versions.runtimeVersion) {
+        _logger.shout('Unable to decode current ScoreCard data.', e, st);
```
sigurdm, line 107: Just to make sure - will the log context include the affected package and version?

isoos: fixed

szakarias (approved):

PR #6094, opened 2022-09-30
Tag pattern with {{version}} for github publishing.

isoos:
- #5769

image

### DIFF
--- a/app/lib/frontend/templates/views/pkg/admin_page.dart
+++ b/app/lib/frontend/templates/views/pkg/admin_page.dart
@@ -269,7 +269,7 @@ d.Node _automatedPublishing(Package package) {
               label: 'Tag pattern',
               value: github?.tagPattern ?? '{{version}}',
             ),
-            d.markdown('`{{version}}` is the version number of the package. '
+            d.markdown('`{{version}}` will be substituted for the version number of the package. '
                 'For example, for tags like `v1.2.3` use `v{{version}}`, '
                 'and for `mypackage-1.2.3` use `mypackage-{{version}}`.'),
           ],
--- a/app/lib/frontend/templates/views/pkg/admin_page.dart
+++ b/app/lib/frontend/templates/views/pkg/admin_page.dart
@@ -269,7 +269,8 @@ d.Node _automatedPublishing(Package package) {
               label: 'Tag pattern',
               value: github?.tagPattern ?? '{{version}}',
             ),
-            d.markdown('`{{version}}` will be substituted for the version number of the package. '
+            d.markdown(
+                '`{{version}}` will be substituted for the version number of the package. '
                 'For example, for tags like `v1.2.3` use `v{{version}}`, '
                 'and for `mypackage-1.2.3` use `mypackage-{{version}}`.'),
           ],
--- a/app/lib/frontend/templates/views/pkg/admin_page.dart
+++ b/app/lib/frontend/templates/views/pkg/admin_page.dart
@@ -262,6 +262,19 @@ d.Node _automatedPublishing(Package package) {
           ),
         ),
         d.div(
+          classes: ['-pub-form-textfield-row'],
+          children: [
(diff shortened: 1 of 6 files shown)


app/lib/package/backend.dart
```diff
@@ -56,6 +56,8 @@ final _defaultMaxVersionsPerPackage = 1000;
 final Logger _logger = Logger('pub.cloud_repository');
 final _validGithubUserOrRepoRegExp =
     RegExp(r'^[a-z0-9\-\._]+$', caseSensitive: false);
+final _validGithubVersionPattern =
+    RegExp(r'^[a-z0-9\-\._]+$', caseSensitive: false);
```
sigurdm, line 60: Tiny nit, I don't think you have to escape the dot:
```suggestion
    RegExp(r'^[a-z0-9\-._]+$', caseSensitive: false);
```

app/lib/package/backend.dart
```diff
@@ -1156,14 +1169,38 @@ class PackageBackend {
...
+      throw AuthorizationException.githubActionIssue(
+          'publishing is only allowed from "tag" refType, this token has "${agent.payload.refType}" refType');
+    }
+    final expectedRefStart = 'refs/tags/';
+    if (!agent.payload.ref.startsWith(expectedRefStart)) {
+      throw AuthorizationException.githubActionIssue(
+          'publishing is only allowed from "refs/tags/*" ref, this token has "${agent.payload.ref}" ref');
+    }
+    final tagPattern = githubPublishing.tagPattern ?? '{{version}}';
+    if (!tagPattern.contains('{{version}}')) {
```
sigurdm, line 1182: This check seems redundant given the validation above. Maybe it is wise to keep it...

isoos: Yeah, in theory we shouldn't have this exception triggered, however, I'd rather have it there, in case we screw up e.g. and admin API that allows updating it

app/lib/frontend/templates/views/pkg/admin_page.dart
```diff
@@ -261,6 +261,19 @@ d.Node _automatedPublishing(Package package) {
...
         ),
+        d.div(
+          classes: ['-pub-form-textfield-row'],
+          children: [
+            material.textField(
+              id: '-pkg-admin-automated-github-tagpattern',
+              label: 'Tag pattern',
+              value: github?.tagPattern ?? '{{version}}',
+            ),
+            d.markdown('`{{version}}` is the version number of the package. '
```
jonasfj, line 272:
```suggestion
            d.markdown('`{{version}}` will be substituted for the version number of the package. '
```

app/lib/package/backend.dart
jonasfj: This is not an authorization error, it's a 500 error. One could argue misconfiguration, but if we're changing the format then everything configured before should be updated or disabled by us in a backfill job.

So this seems like an internal error to me.

app/lib/package/backend.dart
```diff
@@ -1156,14 +1169,38 @@ class PackageBackend {
...
+      throw AuthorizationException.githubActionIssue(
+          'configured tag pattern does not include `{{version}}`');
+    }
+    final expectedTagValue = tagPattern.replaceFirst('{{version}}', newVersion);
+    if (agent.payload.ref.substring(expectedRefStart.length) !=
+        expectedTagValue) {
+      throw AuthorizationException.githubActionIssue(
+          'publishing is configured to only be allowed from actions with a version pattern, '
```
jonasfj, line 1190:
```suggestion
          'publishing is configured to only be allowed from actions with specific ref pattern, '
```

jonasfj (approved):

app/lib/package/backend.dart
isoos: Updated to `ArgumentError`.

PR #6096, opened 2022-09-30
Run post-upload tasks outside of the request scope.

isoos:
- run post-upload tasks with a `Timer` that is scheduled with `Duration.zero`
- removed timeout from post-upload tasks (but kept severe log level if there is an issue)

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -1101,6 +1101,10 @@ class PackageBackend {
     return pv;
   }
 
+  /// The post-upload tasks are not critical and could fail without any impact on
+  /// the uploaded package version. Important operations (e.g. email sending) are
+  /// retried periodically, others (e.g. triggering re-analysis of dependent
+  /// packages) are only nice to have.
   Future<void> _postUploadTasks(
     Package? package,
     PackageVersion newVersion,
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -976,7 +976,7 @@ class PackageBackend {
       authorizedUploaders:
           uploaderEmails.map((email) => EmailAddress(email)).toList(),
     );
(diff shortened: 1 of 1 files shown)


app/lib/package/backend.dart
```diff
@@ -1079,43 +1079,60 @@ class PackageBackend {
...
+        final v = newVersion.qualifiedVersionKey;
+        _logger.severe('Error post-processing package upload $v', e, st);
+      }
+    });
+
     _logger.info('Post-upload tasks completed in ${sw.elapsed}.');
     return pv;
   }
 
+  Future<void> _postUploadTasks(
```
jonasfj, line 1104: Can we add a documentation comment... Maybe saying that none of these steps are critical, which they aren't, right?

What I'm saying is that: failure to do any of these things is not bad, just sad. Am I correct about that?

app/lib/package/backend.dart
```diff
@@ -1079,43 +1079,60 @@ class PackageBackend {
...
+        await _postUploadTasks(
+          package,
+          newVersion,
+          outgoingEmail,
+          prevLatestStableVersion: prevLatestStableVersion,
+          prevLatestPrereleaseVersion: prevLatestPrereleaseVersion,
+        );
+      } catch (e, st) {
+        final v = newVersion.qualifiedVersionKey;
+        _logger.severe('Error post-processing package upload $v', e, st);
```
jonasfj, line 1096: Where will this log stuff go when we do this?

jonasfj (approved):

sigurdm (approved):
LGTM

app/lib/package/backend.dart
isoos: I think this may be correlated with the request, and could go at the end of that batch, but I am just guessing here. It could be just a separate log entry.

PR #6098, opened 2022-10-03
Extra fields in audit log when publishing from GitHub Actions.

isoos:
#6094

### DIFF
--- a/app/lib/audit/models.dart
+++ b/app/lib/audit/models.dart
@@ -209,10 +209,14 @@ class AuditLogRecord extends db.ExpandoModel<String> {
     required DateTime created,
     String? publisherId,
   }) {
+    final uploaderFields = uploader.fields;
+    final formattedFields = uploaderFields == null
+        ? ''
+        : ' (${uploaderFields.entries.map((e) => '${e.key}: `${e.value}`').join(', ')})';
     final summary = [
       'Package `$package` version `$version`',
       if (publisherId != null) ' owned by publisher `$publisherId`',
-      ' was published by `${uploader.displayId}`.',
+      ' was published by `${uploader.displayId}`$formattedFields.',
     ].join();
     return AuditLogRecord()
       ..id = createUuid()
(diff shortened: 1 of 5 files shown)


app/lib/audit/models.dart
```diff
@@ -209,10 +209,14 @@ class AuditLogRecord extends db.ExpandoModel<String> {
...
   }) {
+    final uploaderFields = uploader.fields;
+    final formattedFields = uploaderFields == null
+        ? ''
+        : ' (${uploaderFields.entries.map((e) => '${e.key}: `${e.value}`').join(', ')})';
     final summary = [
       'Package `$package` version `$version`',
       if (publisherId != null) ' owned by publisher `$publisherId`',
-      ' was published by `${uploader.displayId}`.',
+      ' was published by `${uploader.displayId}`$formattedFields.',
```
jonasfj, line 219: This is a user-facing message, and security it's probably relatively important that users understand this message.
Might I suggest something along the lines of:
```markdown
Package `retry` version `1.2.3` was published from Github Actions running on revision `a2a2a2aa2a`
from the `google/dart-neats` repository.
The authenticating Github Actions flow was triggered by a user identified as `jonasfj` on Github (at the time of publishing).
```

isoos: Updated. PTAL.

app/lib/audit/models.dart
```diff
@@ -209,11 +209,37 @@ class AuditLogRecord extends db.ExpandoModel<String> {
...
     String? publisherId,
   }) {
-    final summary = [
+    final summaryParts = <String>[
       'Package `$package` version `$version`',
       if (publisherId != null) ' owned by publisher `$publisherId`',
-      ' was published by `${uploader.displayId}`.',
-    ].join();
+    ];
+    Map<String, dynamic>? fields;
```
jonasfj, line 216:
```suggestion
    var fields = <String, dynamic>{};
```

app/lib/audit/models.dart
```diff
@@ -226,6 +252,7 @@ class AuditLogRecord extends db.ExpandoModel<String> {
         'version': version,
         if (uploader is AuthenticatedUser) 'email': uploader.user.email,
         if (publisherId != null) 'publisherId': publisherId,
+        ...?fields,
```
jonasfj, line 255:
```suggestion
        ...fields,
```

jonasfj (approved):

PR #6099, opened 2022-10-03
Align pub_worker and pana_runner

jonasfj:


pkg/pub_worker/test/analyze_test.dart
```diff
@@ -43,14 +46,114 @@ void main() {
...
+          }),
+        ),
+        TarEntry.data(
+          TarHeader(name: 'lib/meta.dart'),
+          utf8.encode('''
+            class Sealed {}
+          '''),
+        ),
+      ]).transform(tarWriter).transform(gzip.encoder));
+    });
```
jonasfj, line 150: This is a moderate amount of madness, but it's really nice to have self contained tests like this :D

pkg/pub_worker/lib/payload.dart
```diff
@@ -21,10 +21,12 @@ class Payload {
...
-  ///  * `POST <callbackUrl>/api/tasks/<package>/<version>/finished`
+  ///  * `GET  <callbackUrl>/api/packages/<packages>`,
+  ///  * `POST <callbackUrl>/api/tasks/<package>/<version>/upload`, and,
+  ///  * `POST <callbackUrl>/api/tasks/<package>/<version>/finished`.
   ///
-  /// These requests must be authenticated with: `authorization: bearer <token>`.
+  /// The POST requests must be authenticated with:
+  ///         authorization: bearer <token>
   /// Using the `<token>` matching the `<version>` being reported.
   final String callbackUrl;
```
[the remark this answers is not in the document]
jonasfj: Yeah, we probably should. I had really envisioned it as just the callback urls... but that probably won't work :D

pkg/pub_worker/test/analyze_test.dart
jonasfj: Yeah, that's what I did before, but this wasn't so much work  --- and now it's nice and self-contained, no changes on pub.dev is going to break this test :D

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
```diff
@@ -33,26 +41,174 @@ Future<void> main(List<String> args) async {
...
+  //       we won't have to list versions again later.
+  final pubspec = await _fetchPubspec(
+    package: package,
+    version: version,
+    pubHostedUrl: pubHostedUrl,
+  );
+
+  // Discover installed Dart and Flutter SDKs.
+  // This reads sibling folders to the Dart and Flutter SDK.
+  // TODO: Install Dart / Flutter SDKs into these folders ondemand in the future!
```
jonasfj, line 55: This is a bit of a hack... we just read the SDKs from sibling directories of the default installed SDK.

I think this is a pretty cool hack :D

In the future maybe we can skip having an SDK in pub_worker at all, and instead we simply have a pre-built executable (which is faster to startup), and then we download the SDKs we need on-the-fly and cache them on the worker...

It's actually not that crazy, since we have to download the docker image containing the SDK anyways. Whether we download a large docker image or fetch an SDK tarball from GCS probably doesn't matter much.

Though if the flutter tarball is xz compressed we might need a gzipped thing :rofl:  or we'll spend a lot of time just decompressing bytes :see_no_evil:

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
```diff
@@ -78,7 +78,10 @@ Future<void> main(List<String> args) async {
   final toolEnv = await ToolEnvironment.create(
     dartSdkDir: dartSdk?.path,
     flutterSdkDir: flutterSdk?.path,
-    environment: {'CI': 'true'},
+    environment: {
+      'CI': 'true',
+      'PUB_CACHE': pubCache,
```
isoos, line 83: `ToolEnvironment.create` has `pubCacheDir` parameter that will populate this, let's use that instead.

isoos (approved):

PR #6103, opened 2022-10-04
UI and backend for enabling Google Cloud service account publishing.

isoos:
#5769

### DIFF
--- a/app/lib/frontend/templates/views/pkg/admin_page.dart
+++ b/app/lib/frontend/templates/views/pkg/admin_page.dart
@@ -304,10 +304,10 @@ d.Node _automatedPublishing(Package package) {
     ),
     d.h3(text: 'Publishing with Google Cloud Service account'),
     d.markdown(
-      'When publishing with a GCP _service account_ is enabled, the service account configured here'
-      'will be able to create temporary tokens that allows publishing of this package.'
-      'To learn more about publishing using a _service account_, see [dart.dev/go/automated-publishing](https://dart.dev/go/automated-publishing)'
-    )
+        'When publishing with a GCP _service account_ is enabled, the service account configured here '
+        'will be able to create temporary tokens that allows publishing of this package. '
+        'To learn more about publishing using a _service account_, see '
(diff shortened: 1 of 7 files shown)


app/lib/package/backend.dart
```diff
@@ -524,6 +525,22 @@ class PackageBackend {
...
+            googleCloud.serviceAccountEmail?.trim() ?? '';
+        googleCloud.serviceAccountEmail = serviceAccountEmail;
+
+        InvalidInputException.check(
+            !isEnabled || serviceAccountEmail.isNotEmpty,
+            'The service account email field must not be empty when enabled.');
+
+        if (serviceAccountEmail.isNotEmpty) {
+          InvalidInputException.check(isValidEmail(serviceAccountEmail),
+              'The service account email is not valid: `$serviceAccountEmail`.');
```
jonasfj, line 541: Let's require that they end with `.gserviceaccount.com`, see:
https://cloud.google.com/iam/docs/service-accounts#types

We can augment such an error message to say that: you can file a bug if your service account doesn't end with `.gserviceaccount.com`.

app/lib/frontend/templates/views/pkg/admin_page.dart
```diff
@@ -301,6 +302,34 @@ d.Node _automatedPublishing(Package package) {
         ),
       ],
     ),
+    d.h3(text: 'Publishing with Google Cloud Service account'),
```
jonasfj, line 305:
```suggestion
    d.h3(text: 'Publishing with Google Cloud Service account'),
    d.markdown(
      'When publishing with a GCP _service account_ is enabled, the service account configured here'
      'will be able to create temporary tokens that allows publishing of this package.'
      'To learn more about publishing using a _service account_, see [dart.dev/go/automated-publishing](https://dart.dev/go/automated-publishing)'
    )
```

jonasfj: Or something like that.. we probably need similar explanations / help-text for configuring github actions.

jonasfj (approved):

PR #6108, opened 2022-10-10
Automated publishing with a Google Cloud service account

isoos:
- part of #5769
- uses the data model from #6103
- also enables uploads from non-User authenticated agents
- at the end of the upload authorization steps, the publishing is auto-rejected with debug info
- the JWT token verification is done with the OpenId endpoints (and not with any of the Google-specific APIs)


app/lib/account/backend.dart
```diff
@@ -191,55 +220,93 @@ Future<AuthenticatedAgent> requireAuthenticatedAgent(
...
+    // As the uploader token's audience and the admin token's issuer and also
+    // their audience is the same, we only parse it as a non-user token, when
+    // the authentication source is from the pub client app (e.g. uploading a
+    // new package).
+    // At this point we don't fall back to authenticating the token as a user.
+    final payload = await _verifyAndParseToken(
+      idToken,
+      openIdDataFetch: fetchGoogleCloudOpenIdData,
+      payloadTryParse: GoogleCloudServiceAccountJwtPayload.tryParse,
+    );
```
jonasfj, line 275: We could also use the tokeninfo end-point, then we don't have to do the crypto here..

app/lib/audit/models.dart
```diff
@@ -234,6 +234,14 @@ class AuditLogRecord extends db.ExpandoModel<String> {
         if (sha != null) ' revision `$sha`',
         ' to the `$repository` repository.',
       ].join();
+    } else if (uploader is AuthenticatedGoogleCloudServiceAccount) {
+      fields = {
+        'email': uploader.payload.email,
+      };
+      summary = [
+        ...summaryParts,
+        ' was published by Google Cloud Service account `${uploader.payload.email}`.'
```
jonasfj, line 243:
```suggestion
        ' was published by Google Cloud service account: `${uploader.payload.email}`.'
```

jonasfj (approved):

PR #6109, opened 2022-10-10
Lenient HTTP retries for integrity check.

### LINKED ISSUE #6081: Integrity check should use retries
In particular when hitting 502 when trying to fetch an archive.

isoos:
- Fixes #6081.
- Retries HTTP requests on every exception or when `statusCode >= 500`

### DIFF
--- a/app/lib/tool/utils/http.dart
+++ b/app/lib/tool/utils/http.dart
@@ -13,17 +13,23 @@ final _transientStatusCodes = {
 };
 
 /// Creates a HTTP client that retries transient status codes.
+///
+/// When [lenient] is set, we retry on more errors and status codes.
 http.Client httpRetryClient({
   http.Client? innerClient,
   int? retries,
+  bool lenient = false,
 }) {
   return RetryClient(
     innerClient ?? http.Client(),
-    when: (r) => _transientStatusCodes.contains(r.statusCode),
-    // TOOD: Consider implementing whenError and handle DNS + handshake errors.
+    when: (r) =>
+        (lenient && r.statusCode >= 500) ||
+        _transientStatusCodes.contains(r.statusCode),
+    retries: retries ?? 5,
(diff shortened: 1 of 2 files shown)


app/lib/tool/utils/http.dart
```diff
@@ -13,17 +13,23 @@ final _transientStatusCodes = {
 };
 
 /// Creates a HTTP client that retries transient status codes.
+///
+/// When [lenient] is set, we retry on more errors and status codes.
 http.Client httpRetryClient({
   http.Client? innerClient,
   int? retries,
+  bool lenient = false,
```
jonasfj, line 21: This seems like a poorly documented hack.

How about we do retries correctly?

I don't think `RetryClient` will retry when the TCP connection is broken mid-request, so perhaps it better to refactor our retry logic to do something like:
```dart
await retryForHttp(() {
  final req = http.Request(..);
  final res = await client.send(req);
  res.throwIfNotOk();
  // ...
});

extension ThrowIfNotOk on http.Response {
  void throwIfNotOk() {
    if (statusCode < 200 || statusCode >= 300) {
      throw HttpResponseException(
         'TODO: Some message',
         response: this,
         intermittent: _transientStatusCodes.contains(statusCode),
      );
    }
  }
}

/// Same as [retry] except [retryIf] is specified so that it retries on HttpResponseException, TimeoutException, and IOException that could be caused by networking...
Future<T> retryForHttp(...) => retry(
  fn(),
  retryIf: (e) => (e is HttpResponseException && e.isIntermittent) || e is TimeoutException || 
     e is SocketException || e is TlsException || e is HttpException || e is WebSocketException,
  ...,
);
```

---------

The more I look at the HTTP code we have around, I'm leaning towards this being the only robust approach, thoughts?

Everything else feels like short term hacks that fixes a few cases, but doesn't cover all cases.

PR #6120, opened 2022-10-18
Remove dartdoc documentation from the search index.

isoos:
- Also removes entries from the search snapshot.


isoos:
> We still have symbols, right?

Yes.

PR #6138, opened 2022-10-28
Set flag for automated publishing to false...

szakarias:
in order to have the screenshots flag be the only feature enabled when you set the experimental flag.


isoos:
We'll need to also disable the integration test for this feature. However, I don't think this is necessary, because the only thing that will be enabled by this is the admin UI, which seems to be of low risk to expose.

szakarias:
Per discussion, we will do individual flags instead. Closing this.

PR #6139, opened 2022-10-28
Added run_id into audit-log

jonasfj:
Using runId to link the to Github Action run is probably better, then you can also see who signed off on the run (if environment is present).


isoos (approved):
LGTM, but you'll also need to update the test checking the upload message.

isoos (approved):

Issue #6140, opened 2022-10-31
Refactor auth flows


jonasfj:
Refactor auth flows like:
```
requireAuthenticatedAgent(); // is used to implemenet the ones below:
final agent = await requireAuthenticatedAdmin(permission);
final user = await requireAuthenticatedWebUser();
final agent = await requireAuthenticatedClient();
```
So we have some simple methods to use, that all rely on `requireAuthenticatedAgent`. We get rid of `AuthSource` and instead just hardcode this logic into the 3 helper functions for Admin/WebUser/Client

PR #6145, opened 2022-11-02
Remove AuthSource, use expected audience instead.

isoos:
- #6140
- Instead of the `AuthSource` abstraction, callers may do `expectedAudience` checks.
- The auth provider no longer checks for audience, it is pushed one level up (probably not the final place for such checks).
- Audience check cleanup in uploader invites: we no longer allow it from two sources.
- Renamed `requireAuthenticatedAgent` to `requireAuthenticatedClient`, as it is only used in the upload context.

### DIFF
--- a/app/lib/account/auth_provider.dart
+++ b/app/lib/account/auth_provider.dart
@@ -2,22 +2,16 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-enum AuthSource {
-  /// `pub` command-line client
-  client,
-
-  /// `pub.dev` user on the web UI
-  website,
-
-  /// administrator account
-  admin,
-}
-
 class AuthResult {
   final String oauthUserId;
   final String email;
+  final String audience;
 
-  AuthResult(this.oauthUserId, this.email);
+  AuthResult({
+    required this.oauthUserId,
+    required this.email,
+    required this.audience,
+  });
 }
 
 class AccountProfile {
@@ -36,7 +30,7 @@ abstract class AuthProvider {
   ///
   /// Returns null on any error, or if the token is expired, or the user is not
   /// verified.
-  Future<AuthResult?> tryAuthenticate(AuthSource source, String token);
+  Future<AuthResult?> tryAuthenticate(String token);
 
   /// Returns the profile information of a given access token.
   Future<AccountProfile?> getAccountProfile(String? accessToken);
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -28,8 +28,6 @@ import 'auth_provider.dart';
 import 'models.dart';
 import 'session_cookie.dart' as session_cookie;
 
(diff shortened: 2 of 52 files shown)


app/lib/account/backend.dart
```diff
@@ -88,16 +86,27 @@ UserSessionData? get userSessionData =>
 /// When no associated User entry exists in Datastore, this method will create
 /// a new one. When the authenticated email of the user changes, the email
 /// field will be updated to the latest one.
-Future<User> requireAuthenticatedUser({AuthSource? source}) async {
+Future<User> requireAuthenticatedUser({String? expectedAudience}) async {
```
jonasfj, line 89: Let's make it:
```dart
Future<AuthenticatedAgent> requireAuthenticatedAgent({Set<String>? authorizedAudiences}) async {
  authorizedAudiences ??= [
     // TODO: List of all audiences that may be allowed.
  ];
}

class AuthenticatedUser extends AuthenticatedAgent {
  final User user;
  final String audience; // TODO: Let's add this field
}

Future<AuthenticatedAgent> requireAuthenticatedClient() {
  return await requireAuthenticatedAgent(authorizedAudiences: [
    'https://pub.dev',
    // TODO: Audience from client with oauth
  ]);
}
```

app/lib/account/backend.dart
```diff
@@ -214,26 +223,22 @@ class AuthenticatedUser implements AuthenticatedAgent {
 
 /// Verifies the current bearer token in the request scope and returns the
 /// current authenticated user or a service agent with the available data.
-Future<AuthenticatedAgent> requireAuthenticatedAgent(
-    {AuthSource? source}) async {
+Future<AuthenticatedAgent> requireAuthenticatedClient() async {
   final token = _getBearerToken();
```
jonasfj, line 227: Can't we make a `requireAuthenticatedAgent` that is the only thing that calls `_getBearerToken`?

app/lib/account/auth_provider.dart
```diff
@@ -2,22 +2,16 @@
...
   final String oauthUserId;
   final String email;
+  final String audience;
 
-  AuthResult(this.oauthUserId, this.email);
+  AuthResult({
+    required this.oauthUserId,
+    required this.email,
+    required this.audience,
+  });
```
jonasfj, line 14: Ah, so we're moving audience check out of authentication check. I love it!

app/lib/account/backend.dart
isoos: I was planning to have something like this as the next step, to limit the number of moving parts in a single change.

app/lib/account/backend.dart
isoos: Planning to do in one of the next steps. I can add it to this PR, but I'm afraid it would be harder to verify the changes.

app/lib/account/backend.dart
jonasfj: okay, that's reasonable :D

jonasfj (approved):

Issue #6128, opened 2022-10-24
SMTP Authentication Failed (code: 334)


jonasfj:
```
Error: Authentication Failed (code: 334),
response: < eyJzdGF0dXMiOiI0MDAiLCJzY2hlbWVzIjoiQmVhcmVyIiwic2NvcGUiOiJodHRwczovL21haWwuZ29vZ2xlLmNvbS8ifQ==
``` 

The base64 string decodes to:
```js
{"status":"400","schemes":"Bearer","scope":"https://mail.google.com/"}
```

PR #6148, opened 2022-11-03
Email sender backoff period on authentication exception.

isoos:
- #6128
- Fetch new token earlier (increasing the 15 minute age threshold to 20 minutes).
- When email send fails with authentication exception, the token is cleared and the sender indicates a 2-minutes backoff before attempting to send messages again.


app/lib/service/email/backend.dart
```diff
@@ -118,6 +123,10 @@ class EmailBackend {
         sent.add(recipientEmail);
       } catch (e, st) {
         _logger.warning('Email sending failed (claimId="$claimId").', e, st);
+        if (emailSender.shouldBackoff) {
+          resetAttempt = true;
+          break;
+        }
```
jonasfj, line 129: I don't understand this.

Are we using `emailSender.shouldBackoff` as a return value?

That we should backoff is not an indicator that it's reasonable to try sending the email again, is it?

isoos: `emailSender` hides the `SmtpClientAuthenticationException` and returns an `EmailSenderException` instead. This was a quick hack to track the exception type, but I shall rather extend the `EmailSenderException` to indicate if the attempt should be reset.

isoos: Using the status code >= 500 now...

app/lib/service/email/backend.dart
```diff
@@ -116,6 +122,13 @@ class EmailBackend {
           entry.bodyText!,
         ));
         sent.add(recipientEmail);
+      } on EmailSenderException catch (e, st) {
+        _logger.warning('Email sending failed (claimId="$claimId").', e, st);
+        // Resetting attempt count when the failure seems to be infrastructure-related.
+        if (e.status >= 500) {
+          resetAttempt = true;
+          break;
+        }
```
jonasfj, line 131: Shouldn't we limit number of retries anyway? if there is consistent infrastructure issues, we want retries to exhaust.

isoos: We do limit the retries, however, that limit is applied on the single message send operation. If the backend wants to send the mail to 5 recipients, we have separate retries, and until this PR, we have tried to send them separately. The change here would be that after the first authentication-related failure we will not try to send the rest of the recipients, and also reset the attempt count on the datastore entity, so that we don't penalize the recipients just because we were not able to get an auth token.

jonasfj: If you reset the attempts count, then this can stay around forever, correct?

isoos: Good point, I agree. I can either remove, or introduce a small rule that limits the impact, e.g. reset only happens in the first few hours, after that we are bound by the attempt limits. I guess we can just remove now, and if there are issues, we introduce the reset with a time-limited rule.

jonasfj: Let's remove `resetAttempt` and then ship it

jonasfj (approved):

PR #6164, opened 2022-11-08
Fix failing tests.

isoos:
- GitHub added a new field.
- Template golden files were not updated with recent changes.

### DIFF
--- a/app/test/service/openid/github_openid_test.dart
+++ b/app/test/service/openid/github_openid_test.dart
@@ -39,6 +39,7 @@ void main() {
       'event_name',
       'ref_type',
       'environment',
+      'environment_node_id',
       'job_workflow_ref',
       'repository_visibility',
     ]);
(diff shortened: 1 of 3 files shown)


app/test/service/openid/github_openid_test.dart
```diff
@@ -39,6 +39,7 @@ void main() {
       'event_name',
       'ref_type',
       'environment',
+      'environment_node_id',
```
jonasfj, line 42: Wouldn't it be wise to just test that `data.provider.claimsSupported` is a super-set of the claims we rely on.

Why do we want to list claims here, when we don't even read them?

It would be more useful to assert that the claims we rely on are still supported. In in it's current form this test is rater fragile -- it'll break if github changes sorting. But if they add new claims that really doesn't affect us, having a test break without a lengthy documentation comment explaining that this is a zealous test. We risk that future maintainers will spend a lot of time trying to figure out why this test does what it does, and then spend time digging through code find out what relies on this thing.

jonasfj (approved):

app/test/service/openid/github_openid_test.dart
isoos: Agreed. I'll do a follow-up PR to fix it.

PR #6167, opened 2022-11-09
Fix: automated publishing isEnabled checks.

isoos:


app/lib/package/backend.dart
```diff
@@ -1201,7 +1201,7 @@ class PackageBackend {
   Future<void> _checkGithubActionAllowed(AuthenticatedGithubAction agent,
       Package package, String newVersion) async {
     final githubPublishing = package.automatedPublishing.github;
-    if (githubPublishing == null || (githubPublishing.isEnabled ?? false)) {
+    if (githubPublishing == null || !(githubPublishing.isEnabled ?? false)) {
```
jonasfj, line 1204: Maybe, this is easier to parse.
```suggestion
    if (githubPublishing?.isEnabled != true) {
```

jonasfj (approved):
Glade we caught this before launch :D

PR #6168, opened 2022-11-09
Refactor: more generic JWT claim tests.

isoos:
Note: the change revealed that GCP doesn't provide the `nbf` claim. We don't fail if it is missing, just surprising that it is not there.

### DIFF
--- a/app/lib/service/openid/gcp_openid.dart
+++ b/app/lib/service/openid/gcp_openid.dart
@@ -4,6 +4,7 @@
 
 import 'package:collection/collection.dart';
 import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
 import 'package:pub_dev/shared/redis_cache.dart';
 
 import 'jwt.dart';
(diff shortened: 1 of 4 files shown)


app/lib/service/openid/gcp_openid.dart
```diff
@@ -33,10 +33,10 @@ class GcpServiceAccountJwtPayload {
   /// The URL used as the `iss` property of JWT payloads.
   static const issuerUrl = 'https://accounts.google.com';
 
-  static const _requiredClaims = <String>{
+  static const requiredClaims = <String>{
```
jonasfj, line 36:
```suggestion
  @visibleForTesting
  static const requiredClaims = <String>{
```

app/lib/service/openid/github_openid.dart
```diff
@@ -54,7 +54,7 @@ class GitHubJwtPayload {
   /// The URL used as the `iss` property of JWT payloads.
   static const issuerUrl = 'https://token.actions.githubusercontent.com';
 
-  static const _requiredClaims = <String>{
+  static const requiredClaims = <String>{
```
jonasfj, line 57:
```suggestion
  @visibleForTesting
  static const requiredClaims = <String>{
```

jonasfj (approved):

PR #6175, opened 2022-11-11
Audience refactor: test+private API

isoos:
- #6140
- Fixed admin consent test: we shouldn't use pub client audience on admin-related tests. (The test skipped auth checks.)
- With that, the `requireAuthenticatedUser` doesn't need the `expectedAudience` in its public API, preparing the better integration of the auth flows.

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -86,7 +86,7 @@ UserSessionData? get userSessionData =>
 /// When no associated User entry exists in Datastore, this method will create
 /// a new one. When the authenticated email of the user changes, the email
 /// field will be updated to the latest one.
-Future<AuthenticatedUser> requireAuthenticatedUser() async {
+Future<AuthenticatedUser> requireAuthenticatedWebUser() async {
   return await _requireAuthenticatedUser(
       expectedAudience: activeConfiguration.pubSiteAudience);
 }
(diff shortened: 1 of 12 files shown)


app/lib/account/backend.dart
```diff
@@ -86,7 +86,12 @@ UserSessionData? get userSessionData =>
 /// When no associated User entry exists in Datastore, this method will create
 /// a new one. When the authenticated email of the user changes, the email
 /// field will be updated to the latest one.
-Future<AuthenticatedUser> requireAuthenticatedUser(
+Future<AuthenticatedUser> requireAuthenticatedUser() async {
```
jonasfj, line 89:
```suggestion
Future<AuthenticatedUser> requireAuthenticatedWebUser() async {
```

app/lib/account/backend.dart
```diff
@@ -86,7 +86,12 @@ UserSessionData? get userSessionData =>
...
+  return await _requireAuthenticatedUser(
+      expectedAudience: activeConfiguration.pubSiteAudience);
+}
+
+Future<AuthenticatedUser> _requireAuthenticatedUser(
```
jonasfj, line 94: I guess next step is renaming this `requiredAuthenticatedAgent` :partying_face:

jonasfj (approved):

app/lib/account/backend.dart
jonasfj: This is only for web users, right? (this can also be renamed in a follow-up, that only renames this).

isoos: Correct. I've renamed it.

app/lib/account/backend.dart
isoos: rename + restructure

PR #6169, opened 2022-11-09
Dartdoc alignment

jonasfj:

PR #6180, opened 2022-11-14
Dartdoc worker alignment

jonasfj:
Subset of https://github.com/dart-lang/pub-dev/pull/6169/ -- hoping it's self-contained :D

### DIFF
--- a/pkg/pub_worker/bin/pub_worker.dart
+++ b/pkg/pub_worker/bin/pub_worker.dart
@@ -6,9 +6,9 @@ import 'dart:async';
 import 'dart:convert' show json;
 import 'dart:io' show exit;
 
+import 'package:_pub_shared/data/task_payload.dart';
 import 'package:clock/clock.dart';
 import 'package:logging/logging.dart';
-import 'package:pub_worker/payload.dart';
 import 'package:pub_worker/src/analyze.dart' show analyze;
 
 void _printUsage() => print('Usage: pub_worker.dart <JSON_PAYLOAD>');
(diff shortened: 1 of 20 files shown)


pkg/pub_worker/bin/pub_worker.dart
```diff
@@ -20,6 +20,7 @@ Future<void> main(List<String> args) async {
     _printUsage();
     exit(1);
   }
+  print(args.first);
```
isoos, line 23: remove print?

pkg/pub_worker/lib/src/fetch_pubspec.dart
```diff
@@ -0,0 +1,69 @@
...
+        if (r.statusCode != 200) {
+          throw Exception(
+            'Failed to list versions, got ${r.statusCode} from "$u"',
+          );
+        }
+        return json.decode(r.body);
+      },
+      retryIf: (e) =>
+          e is _IntermittentHttpException ||
+          e is FormatException ||
```
isoos, line 41: Not sure: why `FormatException`?

isoos (approved):

pkg/pub_worker/bin/pub_worker.dart
jonasfj: Great catch, thanks,

pkg/pub_worker/lib/src/fetch_pubspec.dart
jonasfj: We could argue we shouldn't... I did it because the FormatException should never happen.
But if it does, it's probably because of a intermittent server issue, or a broken TCP connection or something like that.

PR #6181, opened 2022-11-14
Added static files needed for dartdoc embedding

jonasfj:
Subset of https://github.com/dart-lang/pub-dev/pull/6169 -- planning to use these files later.

We can hold off landing until we need the files, but getting the review done now is nice :D


third_party/dartdoc/README.md
```diff
@@ -0,0 +1,4 @@
+# Dartdoc resource files
+
+Vendoring resource files from:
+https://github.com/dart-lang/dartdoc/tree/master/lib/resources
```
isoos, line 4: Let's reference the `update-dartdoc-resources.sh` script here.

isoos (approved):

third_party/dartdoc/README.md
jonasfj: Will do in a follow up.

PR #6182, opened 2022-11-14
Added support for running fake workers in the background

jonasfj:
Subset of https://github.com/dart-lang/pub-dev/pull/6169 -- planning to use this later.

### DIFF
--- a/app/lib/fake/server/fake_analyzer_service.dart
+++ b/app/lib/fake/server/fake_analyzer_service.dart
@@ -9,7 +9,6 @@ import 'package:fake_gcloud/mem_datastore.dart';
 import 'package:fake_gcloud/mem_storage.dart';
 import 'package:gcloud/db.dart';
 import 'package:gcloud/service_scope.dart' as ss;
-import 'package:gcloud/service_scope.dart';
 import 'package:logging/logging.dart';
 import 'package:pub_dev/analyzer/handlers.dart';
 import 'package:pub_dev/analyzer/pana_runner.dart';
--- a/app/lib/fake/server/fake_analyzer_service.dart
+++ b/app/lib/fake/server/fake_analyzer_service.dart
@@ -65,7 +65,6 @@ class FakeAnalyzerService {
           (taskWorkerCloudCompute as FakeCloudCompute).startInstanceExecution();
           await taskBackend.backfillTrackingState();
           await taskBackend.start();
-          registerScopeExitCallback(taskBackend.stop);
 
           await ProcessSignal.sigint.watch().first;
 
--- a/app/lib/fake/server/fake_analyzer_service.dart
+++ b/app/lib/fake/server/fake_analyzer_service.dart
@@ -9,6 +9,7 @@ import 'package:fake_gcloud/mem_datastore.dart';
 import 'package:fake_gcloud/mem_storage.dart';
 import 'package:gcloud/db.dart';
 import 'package:gcloud/service_scope.dart' as ss;
+import 'package:gcloud/service_scope.dart';
 import 'package:logging/logging.dart';
 import 'package:pub_dev/analyzer/handlers.dart';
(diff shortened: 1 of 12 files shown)


app/lib/fake/server/fake_analyzer_service.dart
```diff
@@ -57,8 +61,14 @@ class FakeAnalyzerService {
           });
           _logger.info('running on port $port');
 
+          (taskWorkerCloudCompute as FakeCloudCompute).startInstanceExecution();
+          await taskBackend.backfillTrackingState();
+          await taskBackend.start();
+
           await ProcessSignal.sigint.watch().first;
 
+          await taskBackend.stop();
```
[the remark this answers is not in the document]
jonasfj: Love it, that's a great solution :D

app/lib/task/cloudcompute/fakecloudcompute.dart
```diff
@@ -104,6 +126,79 @@ class FakeCloudCompute extends CloudCompute {
...
+        _log.info('pub_worker exit code: $exitCode');
+      } finally {
+        // Don't terminate the instance if it's already deleted.
+        if (_instances.any((i) => i.instanceName == instance.instanceName)) {
+          fakeTerminateInstance(instance.instanceName);
+        }
+        _doneRunningInstance.complete();
+
+        // Start running the next instance
+        scheduleMicrotask(() async => await _keepExecutionLoopAlive());
```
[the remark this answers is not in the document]
jonasfj: It's possible the invariant might be harder to follow.

In the same sync block we release `_doneRunningInstance.complete();` we'd to stop working, or overwrite it another completer...

And we'd end up repeating some of the if-statements we have the top...

In the current configuration it's always safe to call `_keepExecutionLoopAlive`. We only have to be careful when calling ourselves, to make sure that we've released the previous lock (`_doneRunningInstance`).

app/lib/fake/server/fake_analyzer_service.dart
```diff
@@ -57,6 +62,11 @@ class FakeAnalyzerService {
           });
           _logger.info('running on port $port');
 
+          (taskWorkerCloudCompute as FakeCloudCompute).startInstanceExecution();
+          await taskBackend.backfillTrackingState();
+          await taskBackend.start();
+          registerScopeExitCallback(taskBackend.stop);
```
isoos, line 68: Oh, I've just found out that `services.dart` already has it on line 263. Do we still need it here?

app/lib/task/cloudcompute/fakecloudcompute.dart
```diff
@@ -104,6 +126,80 @@ class FakeCloudCompute extends CloudCompute {
...
+          mode: ProcessStartMode.inheritStdio,
+        );
+        final exitCode = await proc.exitCode;
+        _log.info('pub_worker exit code: $exitCode');
+      } finally {
+        // Don't terminate the instance if it's already deleted.
+        if (_instances.any((i) => i.instanceName == instance.instanceName)) {
+          fakeTerminateInstance(instance.instanceName);
+        }
+        // Must happen before we start the next instance.
```
isoos, line 173: Nit: can we do any assert here?

isoos (approved):

app/lib/task/cloudcompute/fakecloudcompute.dart
jonasfj: I'm not sure how, we start the next iteration in the very next line. And the next iteration will refuse to run if it's not completed.

app/lib/fake/server/fake_analyzer_service.dart
jonasfj: good catch... there is no reason not call it -- also no harm in doing it. Remove it!

isoos (approved):

PR #6184, opened 2022-11-14
Dartdoc customization alignment

jonasfj:
This is blocked in:
* https://github.com/dart-lang/pub-dev/pull/6183
* https://github.com/dart-lang/pub-dev/pull/6182
* https://github.com/dart-lang/pub-dev/pull/6181
* https://github.com/dart-lang/pub-dev/pull/6180


app/lib/frontend/handlers/documentation.dart
```diff
@@ -43,6 +45,39 @@ Future<shelf.Response> documentationHandler(shelf.Request request) async {
...
+    if (requestMethod != 'HEAD' && requestMethod != 'GET') {
+      // TODO: Should probably be "method not supported"!
+      return notFoundHandler(request);
+    }
+
+    final package = docFilePath.package;
+    var version = docFilePath.version!;
+    final path = docFilePath.path!;
+    // Redirect to /latest/ version if the version points to latest version
+    if (version != 'latest') {
```
isoos, line 59: This seems like very similar to the code below on line 86. Maybe merge the two and move up before the experimental conditional check?

app/lib/task/handlers.dart
```diff
@@ -1,29 +1,108 @@
...
+import 'package:pub_dev/package/backend.dart';
 import 'package:pub_dev/shared/exceptions.dart';
 import 'package:pub_dev/shared/handlers.dart';
+import 'package:pub_dev/shared/redis_cache.dart';
 import 'package:pub_dev/task/backend.dart';
 import 'package:shelf/shelf.dart' as shelf;
 
+const _safeMimeTypes = {
+  // Binary image files are generally safe to serve
+  'avif': 'image/avif',
```
isoos, line 14: I think it is annoying in `package:mime` that this extension-> expected content type mapping is not exposed (IIRC). Let's file an issue to expose these and also a TODO here to use it instead?

app/lib/task/handlers.dart
```diff
@@ -1,29 +1,108 @@
...
+    return htmlResponse(html);
+  }
 
-  final bytes = await taskBackend.dartdocPage(package, version, path);
-  if (bytes == null) {
+  // Handle any non-HTML request
+  final mime = _safeMimeTypes[ext];
+  if (mime == null) {
+    // TODO: Communicate that this file type is not allowed!
+    // Probably we should just do this in pub_worker and write something in the log
```
isoos, line 87: Also: will we prune the links to such file from the generated HTML?

app/lib/dartdoc/dartdoc_page.dart
```diff
@@ -0,0 +1,419 @@
...
+import 'package:collection/collection.dart';
+import 'package:html/dom.dart' show Element;
+import 'package:html/parser.dart' as html_parser;
+import 'package:pub_dev/frontend/dom/dom.dart' as d;
+import 'package:pub_dev/frontend/static_files.dart';
+import 'package:sanitize_html/sanitize_html.dart';
+
+import '../shared/urls.dart';
+
+class DartDocPageOptions {
```
isoos, line 14: This is very similar to `DocFilePath` in structure and intended use, let's extend that with new features! It will also help to merge the path check code paths.

app/lib/dartdoc/dartdoc_page.dart
```diff
@@ -0,0 +1,419 @@
...
+/// [left] | [content] | [right]
+///        |           |
+/// ------------------------------
+///           [footer]
+/// ```
+class DartDocPage {
+  final String title;
+  final String description;
+  final List<Breadcrumb> breadcrumbs;
+  final String left;
```
isoos, line 82: Let's add a note that these are HTML strings (in contrast with the other fields like title or description).

app/lib/dartdoc/dartdoc_page.dart
```diff
@@ -0,0 +1,419 @@
...
+            }
+          }
+          return Breadcrumb(
+            title: crumb.text,
+            href: href,
+          );
+        }).toList() ??
+        <Breadcrumb>[];
+
+    final L = document.querySelector('#dartdoc-sidebar-left');
```
isoos, line 151: Let's call this `left` (and similarly `R` -> `right`, `C` -> `content`), and we can call `sanitize(left?.innerHtml)` at the `DartDocPage` construction.

app/lib/dartdoc/dartdoc_page.dart
```diff
@@ -0,0 +1,419 @@
...
+      content: sanitize(content),
+    );
+  }
+
+  d.Node get _left => d.unsafeRawHtml(left);
+  d.Node get _right => d.unsafeRawHtml(right);
+  d.Node get _content => d.unsafeRawHtml(content);
+
+  d.Node _renderHead(DartDocPageOptions options) =>
+      d.element('head', children: [
```
isoos, line 187: Should we merge this with the pub-dev `head` rendering?

app/lib/dartdoc/dartdoc_page.dart
```diff
@@ -0,0 +1,419 @@
...
+          classes: ['material-symbols-outlined'],
+          attributes: {'role': 'button', 'tabindex': '0'},
+          text: 'menu',
+        ),
+        d.a(
+          href: '/',
+          classes: ['hidden-xs'],
+          child: d.img(
+            // TODO: Move this into a class
+            attributes: {'style': 'height: 30px; margin-right: 1em;'},
```
isoos, line 265: I hope that we have tests that prevent this, but if we don't have it yet, we should ;)

isoos (changes requested):

pkg/web_app/test/inner_html_test.dart
```diff
@@ -12,6 +12,12 @@ void main() {
     await for (final f in dir.list(recursive: true)) {
       if (f is File) {
         final content = await f.readAsString();
+        // ignore innerHtml from package:html in dartdoc_page.dart
+        if (content.contains('package:html') &&
+            !content.contains('dart:html') &&
+            f.path.endsWith('dartdoc_page.dart')) {
+          continue;
+        }
```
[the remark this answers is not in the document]
jonasfj: I wonder if we should rewrite this test as a lint, or as a package that uses `package:analyzer` to detect that `innerHtml` is only used as a getter, or only used on types other than those from `dart:html`.

app/lib/frontend/handlers/documentation.dart
```diff
@@ -43,6 +45,39 @@ Future<shelf.Response> documentationHandler(shelf.Request request) async {
     return redirectResponse('${request.requestedUri}/');
   }
   final String requestMethod = request.method.toUpperCase();
+
+  if (requestContext.experimentalFlags.showSandboxedOutput) {
+    if (requestMethod != 'HEAD' && requestMethod != 'GET') {
```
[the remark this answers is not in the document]
jonasfj: The existing code also does this. This is cargo culting.

We could throw `throw AssertionError` instead, I think it's reasonable to check that the method is as expected.

But yes, we could have an invariant that `documentationHandler` only gets called for `GET` and `HEAD` requests.
I think it's safer to say that an `xxxxHandler` function should generally check the method and return method-not-supported, response if the method isn'ẗ supported.
It's generally nicer than making it an internal error, and it's almost always correct :D

-------------

But yes, we can choose:
 * (A) `xxxxHandler` only supports some methods, and violating this is an internal error.
    * So we throw `AssertionError` if called with method not allowed (or `UnsupportedError`) -- In either case, it's an internal error.
    * The caller is responsible for checking the method (in this case `router.dart` does that).
    * Documentation comment must specify the supported methods -- I'm not sure we're good at writing these.
 * (B)  `xxxxHandler` checks the `request.method` and returns an method-not-supported error.
    * We can probably make this fairly simple by creating a `ResponseException` for method-not-supported. Then we just have `MethodNotSupportedException.checkGetOrHead(request)` function to call.

I think that in general it would be safer to go with (B), and that we should do this in a separate PR, I've filed https://github.com/dart-lang/pub-dev/issues/6205

app/lib/frontend/handlers/documentation.dart
jonasfj: I tried to keep it separate from the old code, because I want to replace the old logic.

app/lib/task/handlers.dart
jonasfj: You can just lookup the extension, but I intentionally want to limit this to safe list of safe-to-serve mimetypes.

Serving SVG is not safe because it can embed scripts..

app/lib/task/handlers.dart
jonasfj: Probably not. If want to do so we should build it into `dartdoc`.

Afaik, no problems exist here by default, it's only if we enable tools and/or raw html in dartdoc that this likely to become an issue.

app/lib/dartdoc/dartdoc_page.dart
jonasfj: afaik, we are serving this right now :see_no_evil:

app/lib/dartdoc/dartdoc_page.dart
jonasfj: I'm down for refactoring this when we remove the old code.
But are you sure we want to retain `DocFilePath`, wouldn't it be nicer to do parsing using `shelf_router`?

All that is needed besides that is adding `index.html` in some cases.

app/lib/dartdoc/dartdoc_page.dart
jonasfj: They look a bit different, I'm not sure we're quite ready for it. I realize there is a little duplication, but that's not new -- all the duplicated stuff was also duplicated in the old customization logic.

isoos (approved):

PR #6190, opened 2022-11-15
Reject unexpected MIME types in api_builder.

### LINKED ISSUE #6151: Follow-up on mime type restrictions in api_builder.
Once https://github.com/dart-lang/pub-dev/pull/6146 get published, we should check the logs and force content type checks.

isoos:
- Fixes #6151.
- No unexpected MIME type has been logged in the past few days.

### DIFF
--- a/pkg/api_builder/lib/api_builder.dart
+++ b/pkg/api_builder/lib/api_builder.dart
@@ -97,12 +97,13 @@ abstract class $utilities {
     T Function(Map<String, dynamic>) fromJson,
   ) async {
     try {
-      // TODO: Consider enforcing that requests should have 'Content-Type' set to
-      // 'application/json'. Notice that we should be careful as the CLI client,
-      // might be sending a different Content-Type.
-      if (!_expectedJsonMimeTypes.contains(request.mimeType)) {
-        _log.info('Unexpected MIME type: ${request.mimeType}', request.mimeType,
-            StackTrace.current);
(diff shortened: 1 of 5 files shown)


pkg/api_builder/lib/api_builder.dart
```diff
@@ -97,12 +97,13 @@ abstract class $utilities {
     T Function(Map<String, dynamic>) fromJson,
   ) async {
     try {
-      // TODO: Consider enforcing that requests should have 'Content-Type' set to
-      // 'application/json'. Notice that we should be careful as the CLI client,
-      // might be sending a different Content-Type.
-      if (!_expectedJsonMimeTypes.contains(request.mimeType)) {
-        _log.info('Unexpected MIME type: ${request.mimeType}', request.mimeType,
-            StackTrace.current);
+      final mimeType = request.mimeType;
```
jonasfj, line 100: How is `request.mimeType` handled. Specifically, will it parse something like `application/json; charset="utf-8"`.

jonasfj (approved):

pkg/api_builder/lib/api_builder.dart
isoos: If the type is malformed, accessing the mimeType will throw FormatException, which we catch and reject the request with 400. Otherwise we'll get the media type part of the content type header.

Issue #6195, opened 2022-11-16
publishing from actions should not allow creating new packages


jonasfj:
https://github.com/dart-lang/pub-dev/blob/b811cb472c77937d7ed21536c8d0192d4ab087a9/app/lib/package/backend.dart#L1049-L1052

This isn't an issue yet, since it's not enabled. But we don't want to allow creating new packages using a non-user entity.

I suggest adding `if (agent is! AuthenticatedUser) throw` in the if-statement where we create the package entity.

isoos:
It looks like that we do block service account uploads through the `_listAdminNotificationEmailsForPackage` checks, but it is non-trivial to conclude this just by reading the code.

jonasfj:
Shouldn't we make an explicit check?

isoos:
> Shouldn't we make an explicit check?

Yes, we will.

isoos:
This has been implemented and we a test for it with the right message.

PR #6196, opened 2022-11-17
New upload tests with service account token.

isoos:
#6195
#5769

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -546,7 +546,7 @@ class PackageBackend {
 
           InvalidInputException.check(
             serviceAccountEmail.endsWith('.gserviceaccount.com'),
-            'The service account email must end with `gserviceaccount.com`. '
+            'The service account email must end with `.gserviceaccount.com`. '
             'If you have a different service account email, please create an issue at '
             'https://github.com/dart-lang/pub-dev',
           );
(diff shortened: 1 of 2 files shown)


app/lib/package/backend.dart
```diff
@@ -1288,7 +1288,7 @@ class PackageBackend {
         serviceAccountEmail.isEmpty ||
         serviceAccountEmail != agent.payload.email) {
       throw AuthorizationException.serviceAccountPublishingIssue(
-          'publishing is not enabled for the "${agent.payload.email}" service account, it may be enabled for another email.');
+          'publishing is not enabled for the "${agent.payload.email}" service account, it may be enabled for another email');
```
jonasfj, line 1291:
```suggestion
          'publishing is not enabled for the "${agent.payload.email}" service account');
```

jonasfj (approved):
Can we have a test that a service account cannot create new packges.

isoos:
> Can we have a test that a service account cannot create new packges.

Added at line 234 in the upload_test.dart in this PR. (Also planning a similar test for Google action token just to be sure, but I have a conflicting change in the works so it gets a bit delayed.) The problem with the flow is that we have a confusing error message, which need to be fixed by solving #6195

jonasfj (approved):

jonasfj (approved):

PR #6198, opened 2022-11-18
Moving JWT token verification into AuthProvider

isoos:
- #6140
- With this change, the `AuthProvider` becomes responsible for both kinds of token parsing and verification. The default implementation is renamed to `DefaultAuthProvider`.
- Added a TODO to migrate the GCP service account token with the tokeninfo endpoint, and there may be other follow-up PRs (e.g. renaming `google_oauth2.dart` to `default_auth_provider.dart`?)

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -176,7 +176,8 @@ Future<AuthenticatedAgent?> _tryAuthenticateServiceAgent(String token) async {
   }
 
   Future<A> parseTokenPayload<A>(
-      A? Function(JwtPayload payload) payloadTryParse,) async {
+    A? Function(JwtPayload payload) payloadTryParse,
+  ) async {
     final payload = payloadTryParse(idToken.payload);
     if (payload == null) {
       throw AuthenticationException.tokenInvalid('unable to parse payload');
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -176,7 +176,7 @@ Future<AuthenticatedAgent?> _tryAuthenticateServiceAgent(String token) async {
   }
 
   Future<A> parseTokenPayload<A>(
-      A? Function(JwtPayload payload) payloadTryParse) async {
+      A? Function(JwtPayload payload) payloadTryParse,) async {
     final payload = payloadTryParse(idToken.payload);
     if (payload == null) {
       throw AuthenticationException.tokenInvalid('unable to parse payload');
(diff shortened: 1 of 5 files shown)


app/lib/account/google_oauth2.dart
```diff
@@ -21,19 +26,72 @@ final _logger = Logger('pub.account.google_auth2');
...
+    final signatureMatches = await idToken.verifySignature(openIdData.jwks);
+    if (!signatureMatches) {
+      throw AuthenticationException.tokenInvalid('invalid signature');
+    }
+  }
+
   /// Authenticate [token] as `access_token` or `id_token`.
   @override
-  Future<AuthResult?> tryAuthenticate(String token) async {
+  Future<AuthResult?> tryAuthenticateAsUser(String token) async {
```
jonasfj, line 94: Can't we merge `tryAuthenticateAsServiceToken` and `tryAuthenticateAsUser` into `tryAuthenticate` ?

jonasfj: `AuthResult` can just have two constructors, with some optional parameters... or we can have two subclasses of `AuthResult`.

jonasfj (approved):

app/lib/account/backend.dart
```diff
@@ -189,80 +170,40 @@ Future<AuthenticatedAgent> _requireAuthenticatedAgent() async {
...
-    // At this point we have confirmed that the token is a JWT token
-    // issued by GitHub. If there is an issue with the token, the
-    // authentication should fail without any fallback.
-    final payload = await _verifyAndParseToken(
-      idToken,
-      openIdDataFetch: fetchGithubOpenIdData,
-      payloadTryParse: GitHubJwtPayload.tryParse,
-    );
+  Future<A> parseTokenPayload<A>(
+      A? Function(JwtPayload payload) payloadTryParse) async {
```
jonasfj, line 179: trailing comma?
```suggestion
      A? Function(JwtPayload payload) payloadTryParse,) async {
```

app/lib/account/google_oauth2.dart
isoos: As discussed offline, this will be done in a follow-up.

app/lib/account/google_oauth2.dart
```diff
@@ -21,19 +26,72 @@ final _logger = Logger('pub.account.google_auth2');
...
-/// Provides OAuth2-based authentication through Google accounts.
-class GoogleOauth2AuthProvider extends AuthProvider {
+/// Provides OAuth2-based authentication through JWKS and Google account APIs.
+class DefaultAuthProvider extends AuthProvider {
   late http.Client _httpClient;
   late oauth2_v2.Oauth2Api _oauthApi;
 
-  GoogleOauth2AuthProvider() {
+  DefaultAuthProvider() {
     _httpClient = http.Client();
```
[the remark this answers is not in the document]
isoos: As discussed offline, this will be done in a follow-up.

PR #6199, opened 2022-11-18
Retrying HTTP clients in DefaultAuthProvider.

isoos:
- Follow-ups to #6198.
- Renamed `google_oauth2.dart` -> `default_auth_provider.dart`.
- Use HTTP retry client in `getAccountProfile` method.
- Use HTTP retry client for Oauth2 API.


app/lib/account/default_auth_provider.dart
```diff
@@ -249,26 +249,29 @@ class DefaultAuthProvider extends AuthProvider {
...
-      imageUrl: info.picture,
-    );
+      final oauth2 = oauth2_v2.Oauth2Api(authClient);
+      final info = await oauth2.userinfo.get();
+      return AccountProfile(
+        name: info.name ?? info.givenName,
+        imageUrl: info.picture,
+      );
+    } finally {
+      client.close();
```
jonasfj, line 273: I think we had an innerClient to ensure reuse of connections. Why do you not want that no

isoos: We did recreate `client` here, I haven't changed that part, only made the closing of it explicit. On the other hand, we could just reuse `_httpClient` if we are happy to have the same retry counts.

jonasfj (approved):

PR #6206, opened 2022-11-22
Emit PackageRejectedException when automated agent is trying to publish first version.

isoos:
- #5769
- Refactored `_listAdminNotificationEmailsForPackage`: separate check for package existence, making it explicit when agent is trying to publish a new package.
- Using `StateError` for the use case that shouldn't happen (inside and outside of `_listAdminNotificationEmailsForPackage`.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -645,8 +645,8 @@ class PackageBackend {
       return _asPackagePublisherInfo(preTxPackage);
     }
 
-    final preTxUploaderEmails = await _listAdminNotificationEmailsForPackage(
-        preTxPackage.name!, authenticatedUser);
+    final preTxUploaderEmails =
+        await _listAdminNotificationEmailsForPackage(preTxPackage);
     preTxPackage.publisherId == null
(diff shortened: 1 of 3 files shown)


jonasfj (approved):

app/lib/package/backend.dart
```diff
@@ -981,8 +981,24 @@ class PackageBackend {
...
+        throw PackageRejectedException.onlyUsersAreAllowedToUploadNewPackages();
+      }
+    } else {
+      uploaderEmails =
+          await _listAdminNotificationEmailsForPackage(existingPackage);
+    }
+    if (uploaderEmails.isEmpty) {
+      // should not happen
+      throw StateError(
+          'Package "${newVersion.package}" has no admin email to notify.');
```
jonasfj, line 999:
```suggestion
      // should not happen
      throw AssertionError(
          'Package "${newVersion.package}" has no admin email to notify.');
```

jonasfj (approved):

app/lib/package/backend.dart
jonasfj: nit, it's just that this isn't an object in the wrong state.

It's unreachable code, this should never happen, an invariant was violated.

PR #6211, opened 2022-11-23
Simpler refs/tags/* check in Github publishing.

isoos:


app/lib/package/backend.dart
```diff
@@ -1219,19 +1219,13 @@ class PackageBackend {
...
-    }
     final tagPattern = githubPublishing.tagPattern ?? '{{version}}';
     if (!tagPattern.contains('{{version}}')) {
       throw ArgumentError(
           'Configured tag pattern does not include `{{version}}`');
     }
     final expectedTagValue = tagPattern.replaceFirst('{{version}}', newVersion);
-    if (agent.payload.ref.substring(expectedRefStart.length) !=
-        expectedTagValue) {
+    if (agent.payload.ref != 'refs/tags/$expectedTagValue') {
```
jonasfj, line 1228: Love it!

PR #6215, opened 2022-11-23
Explicit upload authorization check before the upload transaction begins.

### LINKED ISSUE #6140: Refactor auth flows

isoos:
- Closes #6140.
- Previously it was not entirely clear when we have checked for uploader authorization (e.g. new packages and automated publishing). This moves the explicit check upfront, making it clear when a specific agent can publish.
- Removes duplicate checks for `isBlocked`.


app/lib/package/backend.dart
```diff
@@ -1168,7 +1159,19 @@ class PackageBackend {
   }
 
   Future<void> _checkUploadAuthorization(
```
jonasfj, line 1161:
```suggestion
  /// Throws a [ResponseException] if [agent] is **not** authorized to upload package.
  ///
  /// If [package] is null, this is an attempt to publish a new package, not a new version to an existing package.
  /// If [package] is not null, this is an attempt to publish [newVersion] of existing package.
  Future<void> _requireUploadAuthorization(
```

jonasfj (approved):

PR #6216, opened 2022-11-23
Enable sandboxing for all

jonasfj:
Maybe, we shouldn't enable tasks for all packages yet... This is the stuff I got working.

But we could also ship it, since the risk is fairly small.


app/lib/task/backend.dart
```diff
@@ -333,10 +337,10 @@ class TaskBackend {
   }) async {
     // Limit the number of packages we track during initial testing.
     // TODO: Remove this before going into production
-    if (envConfig.isRunningInAppengine &&
+    /*if (envConfig.isRunningInAppengine &&
```
isoos, line 340: remove these lines + comment?

app/lib/task/cloudcompute/googlecloudcompute.dart
```diff
@@ -103,6 +105,13 @@ CloudCompute createGoogleCloudCompute({
       'must only contain [a-z0-9_-]',
     );
   }
+  if (maxRunDuration.isNegative) {
```
isoos, line 108: Also check for null?

app/lib/task/scheduler.dart
```diff
@@ -67,7 +65,10 @@ Future<void> schedule(
 
       // If terminated or older than maxInstanceAge, delete the instance...
       final isTerminated = instance.state == InstanceState.terminated;
-      final isTooOld = instance.created.isBefore(clock.agoBy(_maxInstanceAge));
+      final maxInstanceAge = Duration(
+        hours: activeConfiguration.maxTaskRunHours,
+      );
+      final isTooOld = instance.created.isBefore(clock.agoBy(maxInstanceAge));
```
isoos, line 71: Nit: This may be easier to read if we had the time going forward, e.g. `instance.created.add(maxInstanceAge).isBefore(clock.now())`

app/test/task/cloudcompute/googlecloudcompute_test.dart
```diff
@@ -26,8 +26,14 @@ void main() {
       // Create CloudCompute instance
       final gce = createGoogleCloudCompute(
         project: envConfig.googleCloudProject!,
+        // NOTE: It's important to have a Cloud NAT for the default network.
+        //       This is probably only reasonable to do in a test project.
```
isoos, line 30: Not sure: why only in the test project?

isoos (approved):
LGTM, mostly nits left.
As the risk of something going sideways is very low, we can enable it for all packages. 🚀

app/lib/task/cloudcompute/googlecloudcompute.dart
jonasfj: It's not nullable, I could check for zero.

app/test/task/cloudcompute/googlecloudcompute_test.dart
jonasfj: We also have it in production project... but we can't test in that project :see_no_evil:

jonasfj:
Reduced it to only enable for all dart team packages. That makes it easier to keep track of logs too...

app/lib/dartdoc/dartdoc_page.dart
```diff
@@ -381,7 +382,9 @@ class DartDocPage {
 
   d.Node _renderBody(DartDocPageOptions options) =>
       d.element('body', attributes: {
-        'data-base-href': '../',
+        'data-base-href': p.relative('', from: p.dirname(options.path)) == '.'
```
isoos, line 385: Let's use an intermediate variable e.g.
```
final baseDir = p.dirname(options.path)) == '.'
            ? ''
            : p.relative('', from: p.dirname(options.path));
```

isoos (approved):

PR #6223, opened 2022-11-25
Simpler payload check + single audience verification in _tryAuthenticateServiceAgent.

isoos:
- removes `parseTokenPayload` as it doesn't add much value
- moves audience check into the auth provider(s) and keeps an assertion check in the caller

### DIFF
--- a/app/lib/account/default_auth_provider.dart
+++ b/app/lib/account/default_auth_provider.dart
@@ -58,11 +58,10 @@ class DefaultAuthProvider extends AuthProvider {
     }
 
     if (idToken.payload.iss == GcpServiceAccountJwtPayload.issuerUrl) {
-      // As the uploader token's audience and the admin token's issuer and also
-      // their audience is the same, we only parse it as a non-user token, when
-      // the authentication source is from the pub client app (e.g. uploading a
-      // new package).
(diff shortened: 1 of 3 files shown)


app/lib/account/default_auth_provider.dart
```diff
@@ -52,10 +57,7 @@ class DefaultAuthProvider extends AuthProvider {
       await _verifyToken(idToken, openIdDataFetch: fetchGithubOpenIdData);
     }
 
-    if (idToken.payload.iss == GcpServiceAccountJwtPayload.issuerUrl &&
-        idToken.payload.aud.length == 1 &&
-        idToken.payload.aud.single ==
-            activeConfiguration.externalServiceAudience) {
+    if (idToken.payload.iss == GcpServiceAccountJwtPayload.issuerUrl) {
       // As the uploader token's audience and the admin token's issuer and also
```
jonasfj, line 61: this comment needs help :D

jonasfj (approved):

app/lib/account/default_auth_provider.dart
isoos: 🙈 (fixed)

jonasfj (approved):

PR #6227, opened 2022-11-28
Move field verification to the front (at service token authorization).

isoos:


app/lib/package/backend.dart
```diff
@@ -1203,11 +1203,24 @@ class PackageBackend {
...
+    final tagPattern = githubPublishing.tagPattern ?? '';
+    if (!tagPattern.contains('{{version}}')) {
+      throw ArgumentError(
+          'Configured tag pattern does not include `{{version}}`');
+    }
+    final requireEnvironment = githubPublishing.requireEnvironment ?? false;
+    final environment = githubPublishing.environment;
+    if (requireEnvironment && (environment == null || environment.isEmpty)) {
+      throw ArgumentError('Missing or empty environment.');
+    }
```
jonasfj, line 1220: ArgumentError becomes a 5xx, right?

I guess that's fine. We agree this should never happen, right?

We could also throw `AssertionError` instead.

jonasfj (approved):

app/lib/package/backend.dart
isoos: Yeah, we should use AssertionError instead.

Issue #6187, opened 2022-11-15
Refactor consent backend to accept non-User agents


isoos:

isoos:
We've decided that we'll keep the User-based consent fields, and keep the admin `User` entities. I'll do another round to check if the current state is consistent and do a clean-up if required (e.g. TODOs).

PR #6234, opened 2022-11-28
Use AuthenticatedAgent in consent audit log record.

isoos:
#6187

### DIFF
--- a/app/lib/audit/models.dart
+++ b/app/lib/audit/models.dart
@@ -566,23 +566,25 @@ class AuditLogRecord extends db.ExpandoModel<String> {
   }
 
   factory AuditLogRecord.uploaderInvited({
-    required User user,
+    required AuthenticatedAgent agent,
     required String package,
     required String uploaderEmail,
   }) {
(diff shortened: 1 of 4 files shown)


app/lib/audit/models.dart
```diff
@@ -566,23 +566,25 @@ class AuditLogRecord extends db.ExpandoModel<String> {
   }
 
   factory AuditLogRecord.uploaderInvited({
-    required User user,
+    required AuthenticatedAgent agent,
```
jonasfj, line 569: Can this ever be called by a non-user? Maybe, an admin?
```suggestion
    required AuthenticatedUser agent,
```

jonasfj (approved):

app/lib/audit/models.dart
isoos: It can be a `AuthenticatedGcpServiceAccount` (which is the entire reason of the refactor in #6187), as admins are no longer users, and admin-triggered consents should be handled as a non-User accounts.

Issue #6233, opened 2022-11-28
GCP service account id_tokens do not have nbf


jonasfj:
jwt: JWT does not have "nbf" field.

in `isTimely`.

Let's not require `nbf` if not present in the token.

I think it also safe to assert that `iat <= nbf`, if both are defined.

PR #6235, opened 2022-11-28
Option to skip nbf check in JWT payload.

isoos:
#6233

### DIFF
--- a/app/lib/account/default_auth_provider.dart
+++ b/app/lib/account/default_auth_provider.dart
@@ -64,11 +64,7 @@ class DefaultAuthProvider extends AuthProvider {
       // without any fallback (e.g. authenticating the token as a User).
 
       // TODO: use the tokeninfo endpoint instead
-      await _verifyToken(
-        idToken,
-        openIdDataFetch: fetchGoogleCloudOpenIdData,
-        maySkipNbfCheck: true,
-      );
+      await _verifyToken(idToken, openIdDataFetch: fetchGoogleCloudOpenIdData);
     }
     return null;
   }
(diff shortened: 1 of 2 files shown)


app/lib/account/default_auth_provider.dart
```diff
@@ -64,16 +64,25 @@ class DefaultAuthProvider extends AuthProvider {
       // without any fallback (e.g. authenticating the token as a User).
 
       // TODO: use the tokeninfo endpoint instead
-      await _verifyToken(idToken, openIdDataFetch: fetchGoogleCloudOpenIdData);
+      await _verifyToken(
+        idToken,
+        openIdDataFetch: fetchGoogleCloudOpenIdData,
+        maySkipNbfCheck: true,
```
jonasfj, line 70: Do we need to make `nbf` required in the github case?

If signed by github or gcp, and it includes nbf or not includes nbf can't we trust it either way?

It's not like the caller can remove the claim if it was there.

app/lib/service/openid/jwt.dart
```diff
@@ -218,7 +218,11 @@ class JwtPayload extends UnmodifiableMapView<String, dynamic> {
   ///
   /// Returns `false` if the current timestamp is outside of the allowed range.
   /// If the timestamp is missing, we treat it as if it were expired/invalid.
-  bool isTimely({DateTime? now, Duration threshold = Duration.zero}) {
+  bool isTimely({
+    DateTime? now,
+    Duration threshold = Duration.zero,
+    bool maySkipNbfCheck = false,
```
jonasfj, line 224:
```suggestion
```

app/lib/service/openid/jwt.dart
```diff
@@ -229,8 +233,11 @@ class JwtPayload extends UnmodifiableMapView<String, dynamic> {
       return a.isBefore(b.add(threshold)) || a == b;
     }
 
+    final hasNbf = nbf != null;
+    final skipNbfCheck = !hasNbf && maySkipNbfCheck;
+
     return isABeforeB('iat', iat, now) &&
-        isABeforeB('nbf', nbf, now) &&
+        (skipNbfCheck || isABeforeB('nbf', nbf, now)) &&
```
jonasfj, line 240:
```suggestion
        (!hasNbf || isABeforeB('nbf', nbf, now)) &&
```

jonasfj (approved):

PR #6242, opened 2022-11-30
Fix 6237

jonasfj:
Fixing 6237

### DIFF
--- a/app/mono_pkg.yaml
+++ b/app/mono_pkg.yaml
@@ -8,13 +8,27 @@ stages:
         - format
         - analyze: --fatal-infos  .
   - unit_test:
-    - test: -P presubmit --total-shards 7 --shard-index 0
-    - test: -P presubmit --total-shards 7 --shard-index 1
-    - test: -P presubmit --total-shards 7 --shard-index 2
-    - test: -P presubmit --total-shards 7 --shard-index 3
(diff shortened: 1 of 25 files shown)


isoos (approved):

szakarias (approved):

app/mono_pkg.yaml
```diff
@@ -8,13 +8,27 @@ stages:
...
+      - test: -P presubmit --total-shards 7 --shard-index 3
+    - group:
+      - command: sudo apt-get install webp
+      - test: -P presubmit --total-shards 7 --shard-index 4
+    - group:
+      - command: sudo apt-get install webp
+      - test: -P presubmit --total-shards 7 --shard-index 5
+    - group:
+      - command: sudo apt-get install webp
+      - test: -P presubmit --total-shards 7 --shard-index 6
```
jonasfj, line 31: @isoos any concerns about this?

It's a bit of a hack, but now we actually run tests with screenshots doing their thing..

isoos: Looks acceptable, but I'd imagine that we should only have the `apt-get install` part only once.

jonasfj: I have no idea how to do this.

PR #6243, opened 2022-11-30
Refactor fake authentication to be much closer to the default authentication.

isoos:
- `BaseAuthProvider` contains the main logic, without the API calls, HTTP calls or `openssl` executions.
- `DefaultAuthProvider` implements the real calls, `FakeAuthProvider` implements in-memory responses with deterministic IDs.
- kept the `user-at-domain-dot-com?audience=fake-site-audience` pattern for simpler entry in the local fake server, but it is transformed into a GCP-like token inside `FakeAuthProvider`

### DIFF
--- a/app/lib/account/default_auth_provider.dart
+++ b/app/lib/account/default_auth_provider.dart
@@ -26,7 +26,7 @@ final _logger = Logger('pub.account.google_auth2');
 final _tokenInfoEndPoint = Uri.parse('https://oauth2.googleapis.com/tokeninfo');
 
 /// Provides OAuth2-based authentication through JWKS and Google account APIs.
-class DefaultAuthProvider extends AuthProvider {
+class DefaultAuthProvider extends BaseAuthProvider {
   late http.Client _httpClient;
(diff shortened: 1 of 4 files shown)


app/lib/account/default_auth_provider.dart
```diff
@@ -35,6 +35,78 @@ class DefaultAuthProvider extends AuthProvider {
...
+    return await oauth2.userinfo.get();
+  }
+
+  @override
+  Future<bool> verifyTokenSignature({
+    required JsonWebToken token,
+    required Future<OpenIdData> Function() openIdDataFetch,
+  }) async {
+    final openIdData = await openIdDataFetch();
+    return await token.verifySignature(openIdData.jwks);
```
jonasfj, line 87: We need caching here, right?

Perhaps we should pass `OpenIdData` and do the actual caching outside `auth_provider`?

isoos: The referenced/passed-in `openIdDataFetch()` functions do the caching.

jonasfj (approved):

PR #6246, opened 2022-12-02
Check valid (well-formatted) dependency names at upload.

isoos:


jonasfj (approved):

jonasfj:
This is a really good idea. I'd love for us to have more sanity checks like this.

PR #6248, opened 2022-12-05
Handle exhausted zone / quota in task scheduler

jonasfj:


app/lib/task/scheduler.dart
```diff
@@ -206,8 +207,40 @@ Future<void> schedule(
...
+          _log.severe(
+            'Quota exhausted trying to create $instanceName, banning all zones '
+            'for 10 minutes',
+            e,
+            st,
+          );
+
+          // Ban all zones for 10 minutes
+          for (final zone in compute.zones) {
+            zoneBannedUntil[zone] = clock.now().add(Duration(minutes: 10));
```
isoos, line 236: This may override a zone ban with ~30 minutes from line#219. Maybe move it behind a method that checks the existence of a ban and sets the later in time.

app/lib/task/scheduler.dart
```diff
@@ -194,6 +194,7 @@ Future<void> schedule(
       assert(description != null);
 
       scheduleMicrotask(() async {
+        var rollbackPackageState = false;
```
isoos, line 197: `var rollbackPackageState = true;`

app/lib/task/scheduler.dart
```diff
@@ -206,8 +207,40 @@ Future<void> schedule(
...
+          // zone for 15 minutes.
+          _log.info(
+            'zone resources exhausted, banning ${e.zone} for 30 minutes',
+            e,
+            st,
+          );
+          // Ban usage of zone for 30 minutes
+          zoneBannedUntil[e.zone] = clock.now().add(Duration(minutes: 30));
+
+          rollbackPackageState = true;
```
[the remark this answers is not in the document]
jonasfj: I wish we had a better way to do this sort of thing...

It's almost tempting to just have `} on Exception catch (e, st) {`
and then use `if (e is ZoneExhaustedException)` instead of  `} on ZoneExhaustedException catch (e, st) {`.

That would provide a more reasonable control flow.

app/lib/task/scheduler.dart
jonasfj: Great catch...

isoos (approved):
Needs to fix tests, otherwise lgtm.

PR #6254, opened 2022-12-07
Refactor: merge almost-identical PanaReport creation from Summary

isoos:


app/lib/scorecard/models.dart
```diff
@@ -245,6 +246,32 @@ class PanaReport {
...
 
+  static PanaReport fromSummary(
+    Summary? summary, {
+    required PackageStatus packageStatus,
+  }) {
+    final reportStatus =
+        summary == null ? ReportStatus.aborted : ReportStatus.success;
+    return PanaReport(
+      // TODO: use summary.createdAt once it lands
+      timestamp: clock.now().toUtc(),
```
jonasfj, line 257: This will be very wrong when it's not stored in scorecard. So let's land the pana fix soon.

jonasfj (approved):

app/lib/scorecard/models.dart
isoos: Yeah, let's land that first...

PR #6258, opened 2022-12-07
Upgrade pana version.

isoos:


jonasfj (approved):
But don't we also have a TODO we can now remove?

isoos:
> But don't we also have a TODO we can now remove?

It will be done with #6254 - I didn't want to deal with too many conflicts when merging it.

PR #6261, opened 2022-12-07
Tooling for better debugging pub_worker issues

jonasfj:
@isoos so far I'm able to reproduce the curious case of:
```
INFO: 2022-12-07 06:26:14.239509: Downloading package clean_router 0.3.6 from https://dartlang-pub-dev.appspot.com/packages/clean_router/versions/0.3.6.tar.gz
FINE: 2022-12-07 06:26:14.371480: Extracting .tar.gz stream to /tmp/pkg-ZOLSKA.
Unhandled exception:
FileSystemException: Cannot open file, path = '/tmp/pkg-ZOLSKA/example/' (OS Error: Is a directory, errno = 21)
#0      _File.open.<anonymous closure> (dart:io/file_impl.dart:356:9)
```
I still have no clue what is going wrong, but at-least I'm able to reproduce it locally :see_no_evil: 

-------------

However, I hope that with tooling like this we can troubleshoot weird issues locally. Possibly even running in a docker container allowing us to reproduce what would happen in the cloud with all details.

Most of this PR is just getting the tooling in place to support this.


pkg/pub_worker/lib/src/testing/server.dart
```diff
@@ -0,0 +1,296 @@
...
+import 'package:shelf/shelf_io.dart' as io;
+import 'package:shelf_router/shelf_router.dart';
+import 'package:tar/tar.dart';
+import 'package:yaml/yaml.dart';
+import 'package:http/http.dart' as http;
+import 'package:retry/retry.dart';
+
+part 'server.g.dart';
+
+class PubWorkerTestServer {
```
isoos, line 28: Would this be served better with running `../../app/bin/fake_server.dart` instead? Is there something limiting there?

isoos (approved):

PR #6263, opened 2022-12-08
Improvements in JWT/JWKS handling.

isoos:
- Updated comments.
- Sanity checks and additional verifications.

### DIFF
--- a/app/lib/service/openid/jwt.dart
+++ b/app/lib/service/openid/jwt.dart
@@ -114,6 +114,12 @@ class JsonWebToken {
   /// Verifies the token with the provided JSON Web Keys and
   /// returns `true` if the signature is valid.
   Future<bool> verifySignature(JsonWebKeyList jwks) async {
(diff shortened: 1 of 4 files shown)


app/lib/service/openid/jwt.dart
```diff
@@ -114,6 +114,12 @@ class JsonWebToken {
   /// Verifies the token with the provided JSON Web Keys and
   /// returns `true` if the signature is valid.
   Future<bool> verifySignature(JsonWebKeyList jwks) async {
+    // At this point we only reject the token if there is no `alg` specified
+    // in the header. If the key is not RS256 (the only one supported right now),
+    // it will get rejected later.
+    if (header.alg == null) {
+      return false;
+    }
```
jonasfj, line 122: Maybe we just do:
```suggestion
    if (header.alg != 'RS256') {
      return false;
    }
```

jonasfj (approved):

PR #6264, opened 2022-12-08
Upgrade dartdoc to 6.1.5

isoos:


srawlins:
Looking at the [commit history](https://github.com/dart-lang/dartdoc/commits/master), I think that is unintentional.

srawlins:
Ah, I'll have to look further back in the log to see what might have caused this. Going from 6.1.1 to 6.1.4.

isoos:
Checked the in-between versions, and it seems that it happened between 6.1.2 -> 6.1.3.

jonasfj (approved):

srawlins:
How can I reproduce this locally? Is there a `dart run test` that I can run?

isoos:
> How can I reproduce this locally? Is there a dart run test that I can run?

After updating the version here: 
https://github.com/dart-lang/pub-dev/blob/master/pkg/pub_dartdoc/pubspec.yaml#L11
I'm running this tool:
https://github.com/dart-lang/pub-dev/blob/master/pkg/pub_dartdoc/tool/update_golden_files.dart

srawlins:
Thanks! Bisecting leads me to https://github.com/dart-lang/dartdoc/commit/46f5f0d8d51371133b71779916add0d5563a13a8. I tested locally both with 2.18 and 2.19.0-edge.29c3fea3bf3aee14c22e628748fafbd2c33cc792. Very strange...

isoos:
> Bisecting leads me to dart-lang/dartdoc@46f5f0d. 

Maybe the commit just before it?
https://github.com/dart-lang/dartdoc/pull/3217

srawlins:
When I bisected, 51464a30fad5ff38f52f60602c0f02ff205edd0c was a good run, keeping the links intact.

srawlins:
OK, yes, this is because of the change in Dart 2.19-dev-something to change the hosted URL for packages in `.dart_tool/package_config.json`. I'm pretty sure this cross-linking will not be broken if this is all run with Dart 2.19. I'm unable to confirm this with `dart pub get` in various directories and the `[update_golden_files.dart](https://github.com/dart-lang/pub-dev/blob/master/pkg/pub_dartdoc/tool/update_golden_files.dart)` script; but it looks to be the case from dartdoc's own tests.

isoos:
> I'm pretty sure this cross-linking will not be broken if this is all run with Dart 2.19

Does that also mean that we shouldn't upgrade to 6.1.3 until we run on 2.19? (So we keep the links.)

srawlins:
I think I can fix-forward with a backwards-compatible fix. I cannot explain why I didn't conceive of this earlier, 🤦 .

https://github.com/dart-lang/dartdoc/pull/3267

isoos:
> I think I can fix-forward with a backwards-compatible fix. I cannot explain why I didn't conceive of this earlier, 🤦 .
> https://github.com/dart-lang/dartdoc/pull/3267

@srawlins: nice! What's the planned release schedule for it?

srawlins:
I have a bump to 6.1.5 here: https://github.com/dart-lang/dartdoc/pull/3274 hope to release today.

isoos:
> I have a bump to 6.1.5 here: dart-lang/dartdoc#3274 hope to release today.

It was fast, thank you! However, I've updated to 6.1.5, but it still has the same problem here.

srawlins:
Rats, I must have been testing it wrong. Would it be costly to wait until pub.dev is running on Dart 2.19? Or some Dart 2.19 pre-release?

jonasfj:
I see no reason to urgently update dartdoc in pub-dev, @isoos or am I missing something?

isoos:
> Rats, I must have been testing it wrong. Would it be costly to wait until pub.dev is running on Dart 2.19? Or some Dart 2.19 pre-release?

This is not blocking and we are not in rush to get this on pub.dev. Take your time and thank you for looking into it!

isoos:
I'm closing the PR, and will re-create when dartdoc will get this fixed.

isoos:
Note: some packages are affected by the lack of this upgrade: #6378

PR #6265, opened 2022-12-09
Add activity log separator between short-term and long-term records.

### LINKED ISSUE #6255: audit log entry, indicating we only retain publication events past 2 months
75GthwTGQWd8EoU

isoos:
Fixes #6255.

### DIFF
--- a/app/lib/frontend/templates/views/account/activity_log_table.dart
+++ b/app/lib/frontend/templates/views/account/activity_log_table.dart
@@ -25,6 +25,21 @@ d.Node activityLogNode({
 }
 
 d.Node _activityLogTableNode(AuditLogRecordPage activities) {
+  final lastShortTermIndex =
+      activities.records.lastIndexWhere((r) => r.isKeptShortTerm);
+  final shortTermRecords =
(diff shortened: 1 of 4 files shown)


app/lib/frontend/templates/views/account/activity_log_table.dart
```diff
@@ -25,6 +25,21 @@ d.Node activityLogNode({
...
+  final longTermRecords =
+      activities.records.skip(lastShortTermIndex + 1).toList();
+  final displaySeparator =
+      shortTermRecords.isNotEmpty && longTermRecords.isNotEmpty;
+  print([
+    shortTermRecords.length,
+    longTermRecords.length,
+    activities.records.length,
+    displaySeparator,
+  ]);
```
jonasfj, line 41: left over? :rofl:

jonasfj (approved):

app/lib/frontend/templates/views/account/activity_log_table.dart
isoos: Good catch, fixed.

PR #6269, opened 2022-12-12
Allow including Dart 3 in upper constraint

parlough:
Currently attempting to publishing packages (`dart pub publish`) with an upper constraint of `<4.0.0` gets the following error:

```
The SDK constraint allows Dart 3.0.0, this is not allowed because Dart 3 does not exist.
```

Now that the [Flutter templates use `<4.0.0`](https://github.com/flutter/flutter/commit/91568cc9b746046e86436e0ab96bb7df837c5c3f) and `dart create` [does so now too](https://github.com/dart-lang/sdk/commit/e3ab2fc4bd329c03769c327b41a4391edcf2b5c5), it seems to make sense to remove this limitation.

I'm not super familiar with pub though, so not sure if other changes are needed first related to package analysis. Feel free to close this PR if it isn't sufficient. Thanks!

\cc @mit-mit

### DIFF
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -439,8 +439,8 @@ Iterable<ArchiveIssue> checkStrictVersions(Pubspec pubspec) sync* {
           'Version value `$v` does not follow strict version pattern.'));
 }
 
-final _preDart3 = VersionConstraint.parse('<3.0.0');
-final _firstDart3Pre = Version.parse('3.0.0').firstPreRelease;
+final _preDart4 = VersionConstraint.parse('<4.0.0');
+final _firstDart4Pre = Version.parse('4.0.0').firstPreRelease;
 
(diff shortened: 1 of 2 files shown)


isoos:
LGTM but let @jonasfj also take a look

mit-mit:
I'm curious @parlough what are the repro steps for getting the message `The SDK constraint allows Dart 3.0.0, this is not allowed because Dart 3 does not exist.` ?

I tried, for example `flutter create + flutter pub get` and that doesn't trigger it despite that now using `<4.0.0`...

parlough:
@mit-mit This message is returned by pub.dev when trying to publish a package which has a `<4.0.0` upper constraint. It doesn't happen on a dry-run, but is instead happened after pushing to the server, hence the change here.

mit-mit:
Ah, got it, I misunderstood and thought this was a `pub get` resolution issue.

jonasfj (approved):
r+

PR #6270, opened 2022-12-12
Purge package cache before retrying package analysis.

isoos:
- An alternative workaround for https://github.com/dart-lang/pana/issues/1165 instead of https://github.com/dart-lang/pana/pull/1174. 
- The first pass on the analysis may fail due to version caching, but on retry (and only on retry) we invalidate the package cache, forcing the cached versions list to be repopulated before the 10-minute cache invalidation would happen automatically.

### DIFF
--- a/app/lib/analyzer/pana_runner.dart
+++ b/app/lib/analyzer/pana_runner.dart
@@ -69,37 +69,31 @@ class _PanaRunner implements PanaRunner {
     return await withToolEnv(
       usesPreviewSdk: packageStatus.usesPreviewAnalysisSdk,
       fn: (toolEnv) async {
-        try {
-          final PackageAnalyzer analyzer =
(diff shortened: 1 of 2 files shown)


isoos:
> Will we need to do something similar for the sandboxed worker?

I think we should have a similar thing, but right now workers don't have a second retry right after a failure. Added a TODO.

jonasfj (approved):

jonasfj:
For sandboxing I think it would be better to just set cache expiration short, and delay analysis for 5 minutes or something like that..

jonasfj:
Or, maybe we do a hack where we trick the pub client into attaching a querystring that indicates that we should by-pass the cache...

isoos:
> Or, maybe we do a hack where we trick the pub client into attaching a querystring that indicates that we should by-pass the cache...

At this point it would be only the pana HTTP client, maybe we add a user-agent header that indicates the purpose...

PR #6275, opened 2022-12-13
Explicit/expanded JWT timestamp checks and also iat <= nbf.

### LINKED ISSUE #6233: GCP service account id_tokens do not have nbf

isoos:
- Fixes #6233.
- Added `iat <= nbf` verification (when `nbf` exists).
- Added test cases for each check.

### DIFF
--- a/app/lib/service/openid/jwt.dart
+++ b/app/lib/service/openid/jwt.dart
@@ -228,17 +228,38 @@ class JwtPayload extends UnmodifiableMapView<String, dynamic> {
   }) {
     now ??= clock.now();
 
-    bool isABeforeB(String name, DateTime? a, DateTime? b) {
-      if (a == null || b == null) {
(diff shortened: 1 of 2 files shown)


isoos:
@jonasfj: The failing GitHub test (which has an example token from GitHub) has the following timestamps:
```
          'nbf': 1632492967, // Friday, September 24, 2021 2:16:07 PM
          'exp': 1632493867, // Friday, September 24, 2021 2:31:07 PM
          'iat': 1632493567, // Friday, September 24, 2021 2:26:07 PM
```
It turns out that `nbf` can be before `iat`... I think we should remove that condition from now.

Issue #6276, opened 2022-10-18
Use <code> inside <details><summary> in report


jonasfj:
![image](https://user-images.githubusercontent.com/149732/196389083-c00f3a0a-bfcb-41e2-b8f8-25fcf773953d.png)

Otherwise we get backticks, because markdown doesn't work. Or we might have to do an empty line before writing the summary for markdown for rendering.

PR #6277, opened 2022-12-14
Markdown tests for keeping track of <summary>-related issues.

isoos:
#6276

### DIFF
--- a/app/test/shared/markdown_test.dart
+++ b/app/test/shared/markdown_test.dart
@@ -255,6 +255,56 @@ void main() {
     });
   });
 
+  group('<summary>', () {
+    test('FIXME: <summary> must not exists in a list item', () {
+      expect(
+        markdownToHtml('- a<summary>b\n- c'),
+        '<ul>\n'
+        '<li>a<summary>b</summary></li>\n'
+        '<li>c</li>\n'
+        '</ul>\n',
+      );
+    });
+
+    test('FIXME: <summary> must not exists without <details>', () {
+      expect(
+        markdownToHtml('A <summary>b</summary>.'),
+        '<p>A </p><summary>b</summary>.<p></p>\n',
+      );
+    });
+
+    test('FIXME: <summary> should be the first child of details', () {
+      expect(
+        markdownToHtml('A <details> <b>bold</b> <summary>b</summary>.'),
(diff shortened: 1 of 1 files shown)


app/test/shared/markdown_test.dart
```diff
@@ -255,6 +255,56 @@ void main() {
...
+            'Consider [migrating](https://dart.dev/null-safety/migration-guide).\n'
+            '</details>',
+        '<details>\n'
+            '<summary>\n'
+            'Package language version (indicated by the sdk constraint `>=2.0.0-dev.48.0 <3.0.0`) is less than 2.12.\n'
+            '</summary>\n\n'
+            'Consider [migrating](https://dart.dev/null-safety/migration-guide).\n'
+            '</details>',
+      );
+    });
```
jonasfj, line 298: ??? aren't you just testing that two strings are equal?

app/test/shared/markdown_test.dart
```diff
@@ -255,6 +255,56 @@ void main() {
...
+  group('<summary>', () {
+    test('FIXME: <summary> must not exists in a list item', () {
+      expect(
+        markdownToHtml('- a<summary>b\n- c'),
+        '<ul>\n'
+        '<li>a<summary>b</summary></li>\n'
+        '<li>c</li>\n'
+        '</ul>\n',
+      );
+    });
```
jonasfj, line 267:
```suggestion
  testMarkdown(String name, String md, html) {
    test(name, () {
      // TODO: Normalize HTML before comparing
      expect(markdownToHtml(md), html));
    });
  }

    testMarkdown('FIXME: <summary> must not exists in a list item', 
       '- a<summary>b\n- c',
        '<ul>\n'
        '<li>a<summary>b</summary></li>\n'
        '<li>c</li>\n'
        '</ul>\n',
    );
```

Or even better maybe, improve our golden test framework and make it more generically reusable.

jonasfj (approved):
This is fine as is though.

app/test/shared/markdown_test.dart
isoos: Yeah, but it shouldn't be, hence the FIXME part in the test's name.

PR #6278, opened 2022-12-14
Fix search's name index: do not broaden the results on multiple words.

isoos:
- Fixes the bad search experience where count_of_hits('$a') was less than count_of_hits('$a $b') (e.g. "charts" vs "flutter charts").
- The main change is to evaluate each query term word separately, joining first on a document-level, and then getting the multiply at the query level.
- The name index is rewritten to get an optimized calculation of this method, the regular index is kept as-is for now, there would be a follow-up to also have a similar optimization.

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -315,15 +315,18 @@ class InMemoryPackageIndex implements PackageIndex {
         return aborted;
       }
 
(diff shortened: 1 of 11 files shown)


isoos:
This has been deployed to staging.

jonasfj (approved):

PR #6289, opened 2022-12-16
Tool for making a report of unique uploaders per month

sigurdm:


app/lib/account/backend.dart
```diff
@@ -562,6 +563,20 @@ class AccountBackend {
...
+    final query = _db.query<AuditLogRecord>();
+    query.filter('kind =', AuditLogRecordKind.packagePublished);
+    if (begin != null) {
+      query.filter('created >=', begin);
+    }
+    if (end != null) {
+      query.filter('created <', end);
+    }
+    return query.run();
+  }
```
jonasfj, line 578: We'll probably need a new entry in index.yaml for this to work... so either we test in staging or we try to guess what that'll be...

jonasfj (approved):

app/lib/account/backend.dart
sigurdm: Tried on staging, and you were absolutely right.

Added an entry in index.yaml.

jonasfj (approved):

PR #6290, opened 2022-12-16
Force-reload page if it was not updated after a sign-out.

### LINKED ISSUE #5970: Page should auto-refresh after sign-out.

isoos:
Fixes #5970.

### DIFF
--- a/pkg/web_app/lib/src/account.dart
+++ b/pkg/web_app/lib/src/account.dart
@@ -90,10 +90,14 @@ void _initWidgets() {
       .getElementById('-account-login')
       ?.onClick
       .listen((_) => authenticationProxy.trySignIn());
-  document
-      .getElementById('-account-logout')
-      ?.onClick
-      .listen((_) => authenticationProxy.signOut());
+  document.getElementById('-account-logout')?.onClick.listen((_) async {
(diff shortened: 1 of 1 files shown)


pkg/web_app/lib/src/account.dart
```diff
@@ -90,10 +90,14 @@ void _initWidgets() {
...
       ?.onClick
       .listen((_) => authenticationProxy.trySignIn());
-  document
-      .getElementById('-account-logout')
-      ?.onClick
-      .listen((_) => authenticationProxy.signOut());
+  document.getElementById('-account-logout')?.onClick.listen((_) async {
+    await authenticationProxy.signOut();
+    // force-reload page if it was not reloaded after signing out
+    await Future.delayed(Duration(seconds: 3));
```
sigurdm, line 96: Why 3 seconds?

1 should be more than enough, no?

jonasfj: Why not just reload?

isoos: I guess I don't trust that the `signOut` enough to complete with everything :) I'll remove it...

jonasfj (approved):

PR #6292, opened 2022-12-19
Add new "Add to publisher" admin tool

budde377:
Solves #5936

### DIFF
--- a/app/lib/admin/backend.dart
+++ b/app/lib/admin/backend.dart
@@ -32,6 +32,7 @@ import '../shared/datastore.dart';
 import '../shared/email.dart';
 import '../shared/exceptions.dart';
 import '../tool/utils/dart_sdk_version.dart';
+import 'tools/add_user_to_publisher.dart';
 import 'tools/block_publisher_and_all_members.dart';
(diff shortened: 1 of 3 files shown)


isoos:
@budde377: Thanks for this contribution! I have already started to work on a bit more involved changed that uses the invite flow to add new members. We think that it is the desired solution, so if I can do it until the next deployment, we'd go with that. (if not, we shall take this)

budde377:
Sure! Can you advise on a timeline or any other ways to unblock me and regain control of my packages?

isoos:
> Sure! Can you advise on a timeline or any other ways to unblock me and regain control of my packages?

Optimistic scenario: we may deploy a new version in the coming days, which will have this new admin tool one way or the other. Once the new release is stable with 100% traffic migrated, one of the admins will run the invite and then you accept and get control.

However, if something (unrelated to this change) goes wrong, it may be that it will be fixed only early January, since we won't have a new release next week.

budde377:
Appreciate the update this seems like a great timeline, thanks!

isoos:
Closing this in favor of #6293

PR #6294, opened 2022-12-20
Log admin when creating publisher

budde377:
When reviewing the code, I saw that the admin option wasn't used. Shouldn't this be used to log who's creating the publisher?


isoos:
The `user` field and the `admin` you have created should be the same. Maybe we could add an assert here, but it shouldn't have any difference at this part of the code.

PR #6295, opened 2022-12-20
Disable sandboxed analysis for now

jonasfj:


app/lib/task/backend.dart
```diff
@@ -322,6 +323,10 @@ class TaskBackend {
     String packageName, {
     bool updateDependants = false,
   }) async {
+    if (activeConfiguration.isProduction) {
```
sigurdm, line 326: Why only in production?

sigurdm (approved):
LGTM

isoos (approved):

PR #6308, opened 2022-12-23
Retry and reduce log level on SMTP handshake timeouts.

### LINKED ISSUE #6232: Timed out while waiting for greeting -- ensure retry
If we do retry, maybe this should just be a warning..
```
Error: Timed out while waiting for greeting.

at .connect ( package:mailer/src/smtp/smtp_client.dart:127 )
at StackZoneSpecification._registerUnaryCallback.<fn> ( package:stack_trace/src/stack_zone_specification.dart:125 )
```

isoos:
Fixes #6232.

### DIFF
--- a/app/lib/frontend/email_sender.dart
+++ b/app/lib/frontend/email_sender.dart
@@ -158,11 +158,7 @@ class _GmailSmtpRelay implements EmailSender {
       _accessToken = null;
       throw EmailSenderException.failed();
     } on MailerException catch (e, st) {
-      var level = Level.SEVERE;
-      if (e is SmtpNoGreetingException) {
-        level = Level.WARNING;
-      }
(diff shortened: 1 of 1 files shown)


app/lib/frontend/email_sender.dart
```diff
@@ -157,7 +158,11 @@ class _GmailSmtpRelay implements EmailSender {
       _accessToken = null;
       throw EmailSenderException.failed();
     } on MailerException catch (e, st) {
```
jonasfj, line 160: How about just doing:
```dart
} on SmtpNoGreetingException catch (e, st) {
  _logger_warning(...);
```

Printing a slightly different error message is nice too.

jonasfj (approved):

PR #6315, opened 2023-01-02
Fix tests and CI

isoos:
- Weekly CI test should run only sanity tags.
- `sanity` configuration didn't unblock `sanity` tags (copy-paste issue)
- Also fixing uploader count test by applying the timestamp just before the operation being run.


sigurdm (approved):
LGTM

jonasfj (approved):

Issue #6314, opened 2023-01-02
[screenshot] Images are not interactive and incorrectly rendered


AlexV525:
I'm trying to setup screenshots for https://pub.dev/packages/wechat_assets_picker, but the corresponding screenshots are not interactive and incorrectly rendered in pub.dev.

In the below image:
1. The image is not fully displayed on the screen.
2. The scroll bar on the page can still control the package page behind, rather than the image, which also means the image cannot be moved around.

![image](https://user-images.githubusercontent.com/15884415/210213970-80d964f0-1b65-49e8-a977-d2749bed6e0a.png)

### Additional context
Window size: 2370*1320
Browser env: Chrome 108.0.5359.95

PR #6320, opened 2023-01-04
Remove scrollbar in carousel and scale image according to window

szakarias:


pkg/web_app/lib/src/screenshot_carousel.dart
```diff
@@ -57,6 +57,7 @@ void _setEventForScreenshot() {
   for (final thumbnail in thumbnails) {
     thumbnail.parent!.onClick.listen((event) {
       showElement(carousel);
+      document.body!.style.overflow = 'hidden';
```
isoos, line 60: At one point we were planning to restrict such inlined styles with Content Security Policy headers. Let's use class names instead, e.g.
```
.overflow-hidden { overflow: hidden;}
.overflow-auto {overflow: auto;}
```
and
```
document.body!.classes.remove('overflow-auto');
document.body!.classes.add('overflow-hidden');
`````

szakarias:
Fixes https://github.com/dart-lang/pub-dev/issues/6314

PR #6339, opened 2023-01-12
Restrict minimum version on upload when SDK constraint allows Dart 3.

isoos:
Fixes https://github.com/dart-lang/pub/issues/3730


sigurdm (approved):
LGTM

jonasfj (changes requested):
Let's hold off for a second here, I don't think we actually need this.

We can discuss if we want to make this restriction, just because we think it's almost always wrong -- but I think it might be overly strict.

Perhaps we should just make it a publishing warning @sigurdm ?

sigurdm:
Yeah - I'm a bit on the fence - maybe a publishing warning is the right thing, but then again - as soon as we have any dart code in such a package the analyzer will warn us (from dart 3 onward) and for older darts we cannot change the warnings.
If there is no files then there should actually be no problem.

PR #6342, opened 2023-01-13
Cookie notice widget.

isoos:
- stores state in localStorage (if available), uses cookies as fallback
- JS and CSS is linked from dartdoc pages too


jonasfj (approved):
let's only use localstorage

PR #6343, opened 2023-01-16
Internal tag to indicate analysis status with the runtime version.

isoos:


app/lib/search/backend.dart
```diff
@@ -120,6 +118,10 @@ class SearchBackend {
       ...p.getTags(),
       ...pv.getTags(),
       ...?scoreCard?.panaReport?.derivedTags,
+      if (scoreCard?.runtimeVersion != null)
+        PackageVersionTags.hasAnalysisOnRuntime(scoreCard!.runtimeVersion!),
```
jonasfj, line 122: I don't mind, but this seems a bit messy...

Isn't easier to just write a bash script that fetches all the scorecards?

PR #6344, opened 2023-01-16
Hide previous image in carousel while waiting for the next

szakarias:


pkg/web_app/lib/src/screenshot_carousel.dart
```diff
@@ -37,8 +37,10 @@ void _setEventForScreenshot() {
   }
 
   void showImage(int index, UIEvent event) {
+    hideElement(imageElement!);
     event.stopPropagation();
```
isoos, line 41: Why does it happen after the hideElement? Wouldn't it be better before it?

szakarias: The stop propagation? I think it might not matter that much, but I can swap the two lines.

isoos (approved):

jonasfj (approved):

PR #6345, opened 2023-01-16
Strict case-sensitive file extraction from archive content.

### LINKED ISSUE #6341: We should not recognize `Example/` (capitalized example folder)
https://pub.dev/packages/shopify_storefront/versions/0.0.2/example shows an example, but our scoring doesn't recognize it: https://pub.dev/packages/shopify_storefront/versions/0.0.2/score

image

I suspect this has to do with capitalization of the folder name. We should align here.

I suggest we only recognize the lowercase version in both cases.

@jonasfj WDYT?

isoos:
Fixes #6341.

### DIFF
--- a/pkg/pub_package_reader/lib/src/file_names.dart
+++ b/pkg/pub_package_reader/lib/src/file_names.dart
@@ -2,25 +2,14 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-/// Returns common file name candidates for [base] (specified without any extension).
-List<String> textFileNameCandidates(String base) {
-  return <String>[
-    base,
-    '$base.md',
-    '$base.markdown',
-    '$base.mkdown',
-    '$base.txt',
-  ];
-}
-
-final List<String> changelogFileNames = textFileNameCandidates('changelog');
-
-final List<String> readmeFileNames = textFileNameCandidates('readme');
+final changelogFileNames = ['CHANGELOG.md'];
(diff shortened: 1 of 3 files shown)


isoos:
Background: The current package layout documentation is strict about the cases of `README.md` and `CHANGELOG.md`, but allows case-insensitive patterns for the example file: https://dart.dev/tools/pub/package-layout#examples

I think we should have only strict casing in the examples to (and also update the documentation about it). (+ update pana, as it still allows lowercase 'readme.md' IIRC).

pkg/pub_package_reader/lib/src/file_names.dart
```diff
@@ -2,25 +2,15 @@
...
-  return <String>[
-    base,
-    '$base.md',
-    '$base.markdown',
-    '$base.mkdown',
-    '$base.txt',
-  ];
-}
-
-final List<String> changelogFileNames = textFileNameCandidates('changelog');
```
sigurdm, line 16: Does this mean we accepted all sorts of names for the changelog before?

Will this break packages? Or is there another place where we require the changelog to be CHANGELOG.md?

isoos: > Does this mean we accepted all sorts of names for the changelog before?

possibly `ChaNgeLoG.mD` was accepted

> Will this break packages?

pub.dev won't display the above file as changelog for new uploads or when we re-extract assets

> Or is there another place where we require the changelog to be CHANGELOG.md?

pana (needs to be updated too)

sigurdm (approved):

Issue #6352, opened 2023-01-18
Packages that only support Dart 3 get a red "dart 2 incompatible" badge 


jakemac53:
See https://pub.dev/packages/build_runner/versions/2.4.0 as an example.

We should not show this badge for packages where the sdk constraint requires Dart 3.

kevmoo:
![image](https://user-images.githubusercontent.com/17034/213298383-aba4a847-0a6d-443e-acd7-3117dd1be546.png)

isoos:
This also impacts packages on search: when a stable, prerelease or preview version gets the 'Dart 2 incompatible' label (internally: `is:legacy`), it gets hidden from regular search results and one needs to enable it by adding `show:unlisted` as a search term.

Part of this will be fixed #6353, but we also need to change the search backend to include only positive search items from preview and prerelease versions.

PR #6355, opened 2023-01-20
Limit the search tags to use from prerelease and preview versions.

isoos:
Related fix for #6352.

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -88,22 +88,23 @@ class SearchBackend {
         await scoreCardBackend.getScoreCardData(packageName, pv.version!);
 
     // Find tags from latest prerelease and/or preview (if there one).
-    Future<Iterable<String>> loadTags(String version) async {
-      final tags = <String>{};
-      final prv =
+    Future<Iterable<String>> loadFutureTags(String version) async {
+      final futureVersion =
           await packageBackend.lookupPackageVersion(packageName, version);
-      prv?.getTags().forEach(tags.add);
-
-      final pra = await scoreCardBackend.getScoreCardData(packageName, version);
-      pra?.panaReport?.derivedTags?.forEach(tags.add);
-      return tags.where(isFutureVersionTag);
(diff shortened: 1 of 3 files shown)


app/lib/search/backend.dart
```diff
@@ -87,26 +87,24 @@ class SearchBackend {
...
 
-    // Find tags from latest prerelease and/or preview (if there one)
-    // This allows searching for tags with `<tag>-in-prerelease`.
-    // Example: `is:null-safe-in-prerelease`, or `platform:android-in-prerelease`
-    Future<List<String>> loadTags(String version) async {
-      final tags = <String>[];
+    // Find tags from latest prerelease and/or preview (if there one).
+    Future<Iterable<String>> loadTags(String version) async {
+      final tags = <String>{};
       final prv =
```
sigurdm, line 93: Just a nit - could we un-abbreviate `prv` and `pra`.

pkg/_pub_shared/lib/search/tags.dart
```diff
@@ -73,12 +76,27 @@ abstract class PackageVersionTags {
 
   /// Package version is compatible with Dart 3.
   static const String isDart3Ready = 'is:dart3-ready';
+
+  /// Version tags that provide a positive, forward-looking property
+  /// of a prerelease or preview version.
```
sigurdm, line 81: Can we extend this dartdoc with an explanation of why we only have the positive tags (eg. by explaining the consequence of having negative tags here).

sigurdm (approved):
LGTM

sigurdm:
Can we also add a regression test?

jonasfj (approved):
We could also choose not to use tags from prereleases in search.

That makes our logic simpler, and easier to understand.

isoos:
> We could also choose not to use tags from prereleases in search.

But then at the mids of the dart3 migration we will get this request again (as we did during the null-safety migration). Let's keep it, I think it is useful, but we need to filter the tags.

PR #6358, opened 2023-01-20
Upgrade pana and use PANA_CACHE in pkg/pub_worker

isoos:


isoos:
Note: I've pushed this to staging.

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@ Important changes to data models, configuration, and migrations between each
 AppEngine version, listed here to ease deployment and troubleshooting.
 
 ## Next Release (replace with git tag when deployed)
- * Bumped runtimeVersion to `2023.01.19`
+ * Bumped runtimeVersion to `2023.01.20`
```
sigurdm, line 5: I think we should strive to include more detail in the CHANGELOG.
Could we add a line about the PANA_CACHE?

sigurdm (approved):

PR #6376, opened 2023-01-27
JWT token from user session (behind experimental flag).

isoos:
- JWT token backed by a secret in Datastore entity using `HS256` algorithm
- emitted only when the experimental flag is set
- emitted only when a special header is set and response is a header value (reduces cross-origin attack surface)
- used in web_app (if enabled/available)

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -4,9 +4,11 @@
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:math';
 
 import 'package:clock/clock.dart';
 import 'package:collection/collection.dart';
+import 'package:crypto/crypto.dart';
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
@@ -16,6 +18,7 @@ import 'package:neat_cache/neat_cache.dart';
 import '../audit/models.dart';
 import '../service/openid/gcp_openid.dart';
 import '../service/openid/github_openid.dart';
+import '../service/openid/jwt.dart';
 import '../shared/configuration.dart';
 import '../shared/datastore.dart';
 import '../shared/exceptions.dart';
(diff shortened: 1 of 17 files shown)


app/lib/account/backend.dart
```diff
@@ -452,6 +465,73 @@ class AccountBackend {
     });
   }
 
+  /// Creates and signs a JWT token using HMAC-SHA256 signature.
```
jonasfj, line 468: We have many kinds of tokens. If this is a CSRF token, we need to consistently use that name. Like `CsrfToken` or something.

app/lib/tool/neat_task/pub_dev_tasks.dart
```diff
@@ -96,6 +96,13 @@ void _setupGenericPeriodicTasks() {
     task: () async => await accountBackend.deleteObsoleteSessions(),
   );
 
+  // Deletes expired tokens.
+  _daily(
+    name: 'delete-expired-tokens',
```
jonasfj, line 101: We cannot talk about generic tokens. That's too generic.

app/lib/account/models.dart
```diff
@@ -150,6 +150,35 @@ class UserSession extends db.ExpandoModel<String> {
...
+
+  /// The expiry in the Datastore should be more into the future than in the JWT,
+  /// as deleting the entry here invalidates the JWT, regardless of the expiry
+  /// there.
+  @db.DateTimeProperty(required: true)
+  DateTime? expires;
+
+  /// The Base64-encoded secret that we use for signing the JWT signature.
+  @db.StringProperty(indexed: false)
+  String? tokenHmacSecretBase64;
```
jonasfj, line 177: The idea with JWT is that you wouldn't need to store state on the server.

We could use a single secret for everything.

If we store a random tokenId on the server, then I don't see why would we need to sign a JWT.

isoos:
Closing this as we've decided to use another solution.

PR #6379, opened 2023-01-30
Upgrade Dart runtime to 2.19.0

isoos:


sigurdm:
Be aware that due to https://github.com/flutter/flutter/issues/119091 everything get's pub upgraded here. It might not be a problem, just beware.

jonasfj (approved):

PR #6381, opened 2023-01-30
Bump of dart-doc size-limit

sigurdm:
Re: https://github.com/dart-lang/dartdoc/issues/3311


isoos (approved):

jonasfj (approved):

jonasfj:
might also be worth mentioning in the CHANGELOG? Just in case we need to debug when we made this change as we review potential issues arising from it?

PR #6383, opened 2023-01-30
Tag integrity checks without any problem.

isoos:


app/lib/shared/integrity.dart
```diff
@@ -65,7 +65,10 @@ class IntegrityChecker {
       count++;
       _logger.warning('[pub-integrity-problem] $problem');
     }
-    _logger.info('Integrity check completed with $count issue(s).');
+    _logger.info([
+      'Integrity check completed with $count issue(s).',
+      if (count == 0) '[pub-integrity-no-problem]',
```
jonasfj, line 70:
```suggestion
      if (count == 0) '[pub-integrity-no-problems-found]',
```

jonasfj (approved):

Issue #6377, opened 2023-01-28
Chrome's instant search isn't working


ciriousjoker:
There is no example code, this is related to Google Chrome's instant search functionality.

1. On sites like npmjs.com, I can press tab to start searching:
<img width="1030" alt="Bildschirm­foto 2023-01-28 um 23 14 27" src="https://user-images.githubusercontent.com/16866223/215293487-e2a9c10f-002d-4f35-80e6-41c6d15c6575.png">

2. On pub.dev, this doesn't work:
<img width="1028" alt="Bildschirm­foto 2023-01-28 um 23 14 45" src="https://user-images.githubusercontent.com/16866223/215293495-b759a399-cf5b-473e-b1fc-d4eeace2a8b0.png">

3. In the past, it occasionally worked, but weirdly every time it did, it displayed a `_` at the end. That "feature" seems to be gone for quite a while, so I can't show a screenshot. It looked like the one on npmjs.com, but it displayed `pub.dev_` as the site being searched and Chrome tried to autocomplete my typed url with a `_` at the end as well.

isoos:
The strange thing is that we do have a good-looking OSD link in the HTML markup:

```
<link rel="search" type="application/opensearchdescription+xml" title="Dart packages" href="[/osd.xml](https://pub.dev/osd.xml)"/>
```

And the osd.xml file is mostly unchanged and was roughly the same over time: https://github.com/dart-lang/pub-dev/blob/master/static/osd.xml

If somebody is interested in digging deeper into the issue, any lead is appreciated!

ciriousjoker:
### Some notes:
- Instant search also doesn't work for me on https://translate.google.com/, so perhaps its an issue on my machine
- the xml file is not loaded at any point while browsing the website (checked via devtools)
- i can reach the xml file when navigating there manually
- issue still occurs in incognito mode without extensions

### Something really interesting:
In Safari, it somewhat works. It displays the "search pub.dev" prompt and i can enter a phrase, but it doesn't actually search for it. This is how it looks:
<img width="554" alt="Bildschirm­foto 2023-01-29 um 04 20 56" src="https://user-images.githubusercontent.com/16866223/215302837-887e38ce-0ad8-4344-94ed-bb1a5ac71afa.png">

@isoos perhaps the xml file has an error?

ciriousjoker:
What the fuck?
https://9to5google.com/2022/03/18/chrome-site-search-activate/

[...]

PR #6386, opened 2023-02-01
Use inlined OpenSearch URL query

isoos:
https://github.com/dart-lang/pub-dev/issues/6377 reported that the current OSD doesn't work with Safari, while Google Translate's work well: https://translate.google.com/opensearch.xml?hl=en

The difference is in the URL format, changing to one that seems to be accepted in more browsers.


jonasfj (approved):

jonasfj:
have someone tested this with Firefox?

isoos:
> have someone tested this with Firefox?

I couldn't get the current pub.dev's osd.xml to work. Maybe the new one...

PR #6399, opened 2023-02-06
Cleanup unused/obsolete tools, moved update package versions to admin/tools.

isoos:
- The `clear_package_properties.dart` should have two successors instead: in integrity tests we should check for unmapped properties, and in periodic tasks we should cleanup unmapped properties (when they are explicitly listed, and after an unused period of releases). Added TODO notes for both.
- `delete_all_staging.dart` has become obsolete I think, we have other tools for it.
- `update_package_versions.dart` moved to admin/tools


app/bin/tools/delete_all_staging.dart
```diff
@@ -1,155 +0,0 @@
-// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
-// for details. All rights reserved. Use of this source code is governed by a
-// BSD-style license that can be found in the LICENSE file.
-
-import 'dart:async';
-
-import 'package:args/args.dart';
```
jonasfj, line 7: @sigurdm do we have other tools for this?

sigurdm: Not that I know - I also don't know if this tool is still functional (I have never tried it).

sigurdm: Last when I restored to staging, I just imported on top - that is not really desirable, so we probably want a tool like this moving forward. However it could live in pub-ops.

isoos: > However it could live in pub-ops.

I'll migrate it too with this PR.

sigurdm (approved):
LGTM

PR #6406, opened 2023-02-07
Prevent non-canonical versions to be uploaded.

### LINKED ISSUE #6402: Reject non-canonical versions on publishing

isoos:
Fixes #6402.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -909,6 +909,7 @@ class PackageBackend {
         throw InvalidInputException.canonicalizeVersionError(
             pubspec.nonCanonicalVersion);
       }
+      // TODO: check this in pkg/pub_package_reader too
       if (versionString != pubspec.nonCanonicalVersion) {
         throw InvalidInputException.nonCanonicalVersion(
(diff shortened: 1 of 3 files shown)


app/lib/package/backend.dart
```diff
@@ -909,6 +909,10 @@ class PackageBackend {
         throw InvalidInputException.canonicalizeVersionError(
             pubspec.nonCanonicalVersion);
       }
+      if (versionString != pubspec.nonCanonicalVersion) {
+        throw InvalidInputException.nonCanonicalVersion(
+            pubspec.nonCanonicalVersion, versionString);
+      }
```
jonasfj, line 915: Why not put this in `summarizePackageArchive` ?

jonasfj (approved):

app/lib/package/backend.dart
jonasfj: I don't think computing the canonical version twice is particularly expensive.

Only concern is whether or not package reader can read existing archives that might violate this property.
But isn't there an explicit `lenient: true` mode to support that? (or do I remember incorrectly?)

PR #6408, opened 2023-02-08
Rename UserSession-related cookies, introducing ClientSession.

isoos:
- Renamed `UserSessionData` -> `SessionData` as both Datastore session entities may use it as-is (+ fields can be added).
- Renamed most generic session-related methods to have `UserSession` in it when they are `UserSession`-only.
- Added new `ClientSession` + expired cleanup.
- Added new cookies, setter method (not used yet) + deleting them linked with the current cookie.
- Added `SameSite=Strict` parameter to cookie builder.
- Exposed experimental status via `<meta>` header, returning from `pkg/web_app` authentication setup early when it is on.

### DIFF
--- a/app/lib/account/models.dart
+++ b/app/lib/account/models.dart
@@ -120,6 +120,47 @@ class LikeData {
   }
 }
 
+/// Tracks the strict session id (from cookie), optionally with
+/// authenticated userId and cached profile information.
+@db.Kind(name: 'ClientSession', idType: db.IdType.String)
+class ClientSession extends db.ExpandoModel<String> {
+  /// Same as [id].
+  /// This is a v4 (random) UUID String.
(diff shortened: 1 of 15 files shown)


app/lib/account/models.dart
```diff
@@ -120,6 +120,47 @@ class LikeData {
...
+  String get sessionId => id as String;
+
+  @db.StringProperty()
+  String? userId;
+
+  @db.StringProperty(indexed: false)
+  String? email;
+
+  /// The name of the user given by the authentication provider.
+  /// May be null, or could contain any arbitrary text.
```
jonasfj, line 138:
```suggestion
  /// May be null, or human readable name (specified by the user, in their profile).
```

jonasfj (approved):

PR #6421, opened 2023-02-14
Parsing the new client session cookies.

isoos:


app/lib/account/session_cookie.dart
```diff
@@ -89,6 +92,36 @@ String? parseUserSessionCookie(String? cookieString) {
   return sessionId;
 }
 
+/// Parses the cookie values and returns the status of client session cookies.
+ClientSessionCookieStatus parseClientSessionCookies(String? cookieString) {
```
jonasfj, line 96: If instead we did:
`parseClientSessionCookies(String? cookieString, String method)`

We can just return `sessionId`.

Even better, we could do: `parseClientSessionCookies(shelf.Request request)`

In cases where we have inconsistency can consider throwing a RequestException that clears the cookies?

Then we wouldn't have to return `needsReset`.

app/lib/account/session_cookie.dart
```diff
@@ -103,15 +136,32 @@ Map<String, Object> clearSessionCookies() {
...
     ],
   };
 }
+
+/// The session cookies present with the request, with the current session identifier.
+class ClientSessionCookieStatus {
+  final bool isPresent;
+  final bool isStrict;
+  final String? sessionId;
+  final bool needsReset;
```
jonasfj, line 157: Documentation comments, if we really want to carry around this much information.

If `needsReset` is true, then wouldn't it almost always be okay to throw a `ResponseException` that clears cookies?

jonasfj (approved):

app/lib/account/session_cookie.dart
isoos: I think it is better to keep this method at the cookie-parsing logic, and have another one which authorizes the cookie based on the method and the attempted operation.

PR #6423, opened 2023-02-15
Replace cookie-notice widget.

isoos:


isoos:
Note: this is currently deployed to staging, although it is only available on the regular pages, most dartdoc doesn't have it yet.

jonasfj (approved):

PR #6427, opened 2023-02-15
Skeleton for authentication with OpenID connect flow.

isoos:


app/lib/account/auth_provider.dart
```diff
@@ -41,6 +47,17 @@ abstract class AuthProvider {
...
+  /// Returns the URL that the user should visit for authentication.
+  Future<Uri> getOauthAuthenticationUrl({
+    required String state,
+    required String nonce,
+  });
+
+  /// Returns the OAuth2 profile information using the authentication [code].
+  Future<AuthResult?> tryAuthenticateOauthCode({
+    required String code,
+  });
```
jonasfj, line 59: Why not stick the nonce back in here?

So that the verification of it happens here? The `getOauthAuthenticationUrl` takes a `nonce`.

app/lib/account/default_auth_provider.dart
```diff
@@ -86,6 +93,69 @@ class DefaultAuthProvider extends BaseAuthProvider {
...
+      if (rs.statusCode != 200) {
+        return null;
+      }
+      final body = json.decode(rs.body);
+      // TODO: also expose access_token for domain verification calls
+      final idToken = body['id_token'] as String;
+      final auth = await _tryAuthenticateJwt(idToken);
+      if (auth == null) {
+        return null;
+      }
```
jonasfj, line 152: should we verify the `nonce` here? I think it's part of the OIDC flow. So probably we should have the `tryAuten...` method fail, if nonce is not matched.

app/lib/account/auth_provider.dart
```diff
@@ -8,11 +8,17 @@ class AuthResult {
   final String oauthUserId;
   final String email;
   final String audience;
+  final String? nonce;
```
jonasfj, line 11: Do we need nonce here? After OIDC flow have checked it, I don't think we need to check it again.

app/lib/account/auth_provider.dart
```diff
@@ -41,6 +45,18 @@ abstract class AuthProvider {
   /// Returns the profile information of a given access token.
   Future<AccountProfile?> getAccountProfile(String? accessToken);
 
+  /// Returns the URL that the user should visit for authentication.
+  Future<Uri> getOauthAuthenticationUrl({
+    required String state,
+    required String nonce,
+  });
+
+  /// Returns the OAuth2 profile information using the authentication [code].
```
jonasfj, line 54:
```suggestion
  /// Verifies authentication using [code] and returns the OAuth2 profile information.
  ///
  /// This is the final step of the sign-in flow started by redirecting the user
  /// to [getOauthAuthenticationUrl].
```

The important part is that it authenticates the user.

app/lib/account/default_auth_provider.dart
```diff
@@ -86,6 +93,73 @@ class DefaultAuthProvider extends BaseAuthProvider {
     final openIdData = await openIdDataFetch();
     return await token.verifySignature(openIdData.jwks);
   }
+
+  String _getOauthSiteAudience() =>
+      _oauthSiteAudience ?? activeConfiguration.pubSiteAudience!;
```
jonasfj, line 98: Why do we need two ways to specify this?

app/lib/account/default_auth_provider.dart
```diff
@@ -86,6 +93,73 @@ class DefaultAuthProvider extends BaseAuthProvider {
...
+
+  Future<String?> _getOauthSiteAudienceSecret(String audience) async {
+    return _oauthSiteAudienceSecret ??
+        await secretBackend.getCachedValue('${SecretKey.oauthPrefix}$audience');
```
jonasfj, line 102: Does `secretBackend` cache this in-memory, or only on redis?

app/lib/frontend/handlers/routes.dart
```diff
@@ -309,6 +309,16 @@ class PubSiteService {
   // **** Account, authentication and user administration
   // ****
 
+  /// Redirects user to the OAuth2 sign-in page.
+  @Route.get('/sign-in')
+  Future<Response> startSignIn(Request request) async =>
+      startSignInHandler(request);
+
+  /// Redirects user to the OAuth2 sign-in page.
+  @Route.get('/oauth/callback')
```
jonasfj, line 318:
```suggestion
  @Route.get('/sign-in/callback')
```

app/lib/account/default_auth_provider.dart
```diff
@@ -320,10 +397,31 @@ abstract class BaseAuthProvider extends AuthProvider {
       _logger.warning('JWT rejected, email_verified = "$emailVerified"');
       return null; // missing email (probably missing 'email' scope)
     }
+    final nonce = r['nonce'];
+    if (nonce != null && nonce is! String) {
+      _logger.warning('JWT rejected, bad `nonce`');
+      return null;
+    }
+    if (expectedNonce != null && nonce != expectedNonce) {
```
jonasfj, line 405: This should use fixed-time comparison. Otherwise it might be vulnerable to a timing attack.

It'd probably be hard to do in practice, since you probably can't create enough Google signed JWTs to launch an actual timing attack :D

app/lib/service/services.dart
```diff
@@ -177,7 +177,17 @@ Future<R> withFakeServices<R>({
...
 
     // register fake services that would have external dependencies
-    registerAuthProvider(FakeAuthProvider());
+
+    // Local override for OAuth services when developing for external servers.
+    if (envConfig.fakeOauthSiteAudience != null &&
+        envConfig.fakeOauthSiteAudienceSecret != null) {
+      registerAuthProvider(DefaultAuthProvider(
+        oauthSiteAudience: envConfig.fakeOauthSiteAudience,
+        oauthSiteAudienceSecret: envConfig.fakeOauthSiteAudienceSecret,
```
jonasfj, line 186: Why can't we put these in the config file?

Are we simply registering an oauth flow for use on localhost, or what is the plan?
Because I'm not sure it'd be horrible to registered that in the staging site, and check it into git in our config files.

app/lib/account/default_auth_provider.dart
isoos: Only in redis.

app/lib/service/services.dart
isoos: Removed these overrides.

app/lib/account/default_auth_provider.dart
isoos: Updated.

app/lib/account/default_auth_provider.dart
```diff
@@ -86,6 +88,72 @@ class DefaultAuthProvider extends BaseAuthProvider {
...
+      },
+    );
+  }
+
+  @override
+  Future<AuthResult?> tryAuthenticateOauthCode({
+    required String code,
+    required String expectedNonce,
+  }) async {
+    try {
```
jonasfj, line 126: Don't we also need to verify the `state`:
https://developers.google.com/identity/openid-connect/openid-connect#confirmxsrftoken

app/lib/frontend/handlers/account.dart
```diff
@@ -31,6 +33,56 @@ final _logger = Logger('account_handler');
...
+/// Handles GET /sign-in
+Future<shelf.Response> startSignInHandler(shelf.Request request) async {
+  if (!requestContext.experimentalFlags.useNewSignIn) {
+    return notFoundHandler(request);
+  }
+  final nonce = createUuid();
+  // TODO: get current session or create a new one
+  // TODO: store nonce on session
+  final oauth2Url = await authProvider.getOauthAuthenticationUrl(
+    state: '', // TODO: use meaningful state
```
jonasfj, line 45: We need a uuid for state too, I think:
https://developers.google.com/identity/openid-connect/openid-connect#confirmxsrftoken

I think it's common to also include a URL to which we want to redirect back to. But we can take that later.

jonasfj (approved):

Issue #6428, opened 2023-02-16
Delete unwanted objects from public bucket in background task


sigurdm:
We don't want to delete anything newer than 24 hours to avoid race conditions.
First implementation should just SHOUT, not actually delete.

PR #6432, opened 2023-02-17
Warn on possible leftover files in the public archive bucket.

isoos:
#6428

### DIFF
--- a/app/lib/tool/maintenance/update_public_bucket.dart
+++ b/app/lib/tool/maintenance/update_public_bucket.dart
@@ -76,21 +76,27 @@ Future<PublicBucketUpdateStat> updatePublicArchiveBucket({
 
     // Skip recently updated objects.
     if (publicInfo.age < ageCheckThreshold) {
+      // Ignore recent files.
       continue;
     }
 
     final canonicalInfo = await canonicalBucket.tryInfo(entry.name);
     if (canonicalInfo != null) {
+      // Warn if both the canonical and the public bucket has the same object,
+      // but it wasn't matched through the [PackageVersion] query above.
       if (publicInfo.age < ageCheckThreshold) {
+        // Ignore recent files.
         continue;
       }
       _logger.severe(
           'Object without matching PackageVersion in canonical and public buckets: "${entry.name}".');
       continue;
+    } else {
+      // The object in the public bucket has no matching file in the canonical bucket.
+      // We can assume it is stale and can delete it.
+      _logger.shout('Deleting object from public bucket: "${entry.name}".');
+      deleted++;
     }
-
-    _logger.shout('Deleting object from public bucket: "${entry.name}".');
-    deleted++;
   }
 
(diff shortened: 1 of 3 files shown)


app/lib/tool/maintenance/update_public_bucket.dart
```diff
@@ -38,6 +54,47 @@ Future<int> updatePublicArchiveBucket() async {
...
+    // Skip recently updated objects.
+    if (publicInfo.age < ageCheckThreshold) {
+      continue;
+    }
+
+    final canonicalInfo = await canonicalBucket.tryInfo(entry.name);
+    if (canonicalInfo != null) {
+      if (publicInfo.age < ageCheckThreshold) {
+        continue;
+      }
```
jonasfj, line 86: What is this good for?

app/lib/tool/maintenance/update_public_bucket.dart
```diff
@@ -38,6 +54,47 @@ Future<int> updatePublicArchiveBucket() async {
...
+      continue;
+    }
+
+    // Skip recently updated objects.
+    if (publicInfo.age < ageCheckThreshold) {
+      continue;
+    }
+
+    final canonicalInfo = await canonicalBucket.tryInfo(entry.name);
+    if (canonicalInfo != null) {
```
jonasfj, line 83:
```suggestion
    if (canonicalInfo == null) {
```

jonasfj (approved):

jonasfj:
Comments are probably worth fixing :D

app/lib/tool/maintenance/update_public_bucket.dart
isoos: This is actually intended. It checks for files that are both present in the public and canonical buckets, but were not checked through packageversion archives. I'll add a comment on the intent there.

jonasfj: ah, okay, so because it's on both, you think it's a weird. makes sense

app/lib/tool/maintenance/update_public_bucket.dart
```diff
@@ -38,6 +54,53 @@ Future<int> updatePublicArchiveBucket() async {
...
+      // Warn if both the canonical and the public bucket has the same object,
+      // but it wasn't matched through the [PackageVersion] query above.
+      if (publicInfo.age < ageCheckThreshold) {
+        // Ignore recent files.
+        continue;
+      }
```
jonasfj, line 90: Sure you don't want this?
```suggestion
      if (canonicalInfo.age < ageCheckThreshold) {
        // Ignore recent files.
        continue;
      }
```

jonasfj (approved):

app/lib/tool/maintenance/update_public_bucket.dart
isoos: Ah, good catch! thanks!

Issue #6069, opened 2022-09-22
Investigate reducing the amount of logging


sigurdm:
We might have log-entries that provide very little value, but make it harder to navigate the logs.

jonasfj:
When running locally or in CI? @sigurdm

jonasfj:
Or when running in production?

sigurdm:
I think we filed this issue related to production.

PR #6433, opened 2023-02-17
Reduce fake server logging when running locally or in CI.

isoos:
#6069

### DIFF
--- a/app/lib/fake/server/fake_server_entrypoint.dart
+++ b/app/lib/fake/server/fake_server_entrypoint.dart
@@ -6,7 +6,6 @@ import 'dart:convert';
 
 import 'package:args/command_runner.dart';
 import 'package:http/http.dart';
-import 'package:logging/logging.dart';
 import 'package:pub_dev/fake/backend/fake_dartdoc_runner.dart';
 import 'package:pub_dev/fake/backend/fake_pana_runner.dart';
 import 'package:pub_dev/fake/server/fake_analyzer_service.dart';
@@ -17,7 +16,6 @@ import 'package:pub_dev/fake/server/fake_storage_server.dart';
 import 'package:pub_dev/fake/server/local_server_state.dart';
 import 'package:pub_dev/frontend/static_files.dart';
 import 'package:pub_dev/shared/configuration.dart';
-import 'package:pub_dev/shared/env_config.dart';
 import 'package:pub_dev/shared/logging.dart';
 import 'package:pub_dev/task/cloudcompute/fakecloudcompute.dart';
 import 'package:pub_dev/tool/test_profile/import_source.dart';
@@ -68,18 +66,7 @@ class FakeServerCommand extends Command {
     final dataFile = argResults!['data-file'] as String?;
     final watch = argResults!['watch'] == true;
 
-    if (envConfig.isInsideCI) {
-      setupDebugEnvBasedLogging(extraDebug: 'fake_server pub.email');
-    } else {
-      Logger.root.onRecord.listen((r) {
-        print([
-          r.time.toIso8601String(),
-          r.toString(),
-          r.error,
-          r.stackTrace?.toString(),
-        ].where((e) => e != null).join(' '));
-      });
-    }
+    setupDebugEnvBasedLogging(defaultDebugEnvValue: 'fake_server pub.email');
 
     final state = LocalServerState();
     if (dataFile != null) {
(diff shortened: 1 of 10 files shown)


app/lib/fake/server/fake_server_entrypoint.dart
```diff
@@ -66,14 +68,18 @@ class FakeServerCommand extends Command {
...
+      setupDebugEnvBasedLogging(extraDebug: 'fake_server pub.email');
+    } else {
+      Logger.root.onRecord.listen((r) {
+        print([
+          r.time.toIso8601String(),
+          r.toString(),
+          r.error,
+          r.stackTrace?.toString(),
+        ].where((e) => e != null).join(' '));
+      });
```
jonasfj, line 81: Why do we need two different options here?

Can we just specify the `DEBUG` env variable in CI?

app/lib/shared/logging.dart
```diff
@@ -0,0 +1,84 @@
...
+/// operating on log names.
+///
+/// **Examples**:
+///  * `DEBUG='*'`, will show output from all loggers.
+///  * `DEBUG='pub.*'`, will show output from loggers with name prefixed 'pub.'.
+///  * `DEBUG='* -neat_cache'`, will show output from all loggers, except 'neat_cache'.
+///
+/// Multiple filters can be applied, the last matching filter will be applied.
+void setupDebugEnvBasedLogging({
+  String? extraDebug,
```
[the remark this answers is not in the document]
isoos: Replaced with `defaultDebugEnvValue` as a fallback value + added comment.

app/lib/fake/server/fake_server_entrypoint.dart
isoos: Removed the custom one, only using the single debug-based flow.

app/lib/fake/server/fake_server_entrypoint.dart
```diff
@@ -66,14 +66,7 @@ class FakeServerCommand extends Command {
...
 
-    Logger.root.onRecord.listen((r) {
-      print([
-        r.time.toIso8601String(),
-        r.toString(),
-        r.error,
-        r.stackTrace?.toString(),
-      ].where((e) => e != null).join(' '));
-    });
+    setupDebugEnvBasedLogging(defaultDebugEnvValue: 'fake_server pub.email');
```
jonasfj, line 69: This is because integration tests wouldn't work without this hack?

In that case, we could:
 * add a comment saying that we need to always have this by default for integration tests to pass.
 * We could specify `DEBUG=...` when running integration tests, since the integration tests actually launch the fake_server_entrypoint, right?

Specifically, why don't we just in:
https://github.com/dart-lang/pub-dev/blob/fee7f33c94bce8607eba24a6a80ba289d76309d5/pkg/pub_integration/lib/src/fake_pub_server_process.dart#L49-L67

specify the env var `DEBUG`, indeed we can specify it as:
```
  'DEBUG': 'fake_server pub.email ' + Platform.environment['DEBUG'] ?? ''
```

since it'll split with whitespace.

pkg/pub_integration/test/publisher_test.dart
```diff
@@ -31,7 +31,8 @@ void main() {
 
       Future<void> inviteCompleterFn() async {
         final pageUrl = await inviteUrlFuture.timeout(Duration(seconds: 30));
-        final pageUri = Uri.parse(pageUrl);
+        final pageUri = Uri.parse(
+            pageUrl.substring(pageUrl.indexOf('https://pub.dev/consent')));
```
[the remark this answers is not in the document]
isoos: Fixed.

app/lib/fake/server/fake_server_entrypoint.dart
isoos: Updated with `DEBUG` defaulting to `*` for locally running it.

pkg/pub_integration/lib/src/fake_pub_server_process.dart
```diff
@@ -64,6 +64,9 @@ class FakePubServerProcess {
         '--dartdoc-port=$dartdocPort',
       ],
       workingDirectory: pkgDir,
+      environment: {
+        'DEBUG': 'fake_server pub.email',
```
jonasfj, line 68:
```suggestion
        // Because we read the consent email from stdout.
        'DEBUG': 'fake_server pub.email',
```

jonasfj (approved):

PR #6441, opened 2023-02-21
Fix: expose headers on ResponseException in non-API responses.

isoos:


jonasfj (approved):

jonasfj:
When we use this, maybe we should add an integration test that covers it, if feasible.

PR #6444, opened 2023-02-22
Close CloudCompute service on scope exit.

isoos:
I think this is a contributing factor for the timeouts while running the code coverage.

### DIFF
--- a/app/lib/task/cloudcompute/cloudcompute.dart
+++ b/app/lib/task/cloudcompute/cloudcompute.dart
@@ -98,6 +98,9 @@ abstract class CloudCompute {
   /// Systems should assume deletion is best-effort and periodically list
   /// instances to delete instances that are no longer needed.
   Future<void> delete(String zone, String instanceName);
+
+  /// Cancel timers and free resources.
+  Future<void> close();
 }
 
 /// Exception thrown by [CloudCompute.createInstance] if resources in a cloud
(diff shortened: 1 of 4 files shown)


app/lib/task/cloudcompute/cloudcompute.dart
```diff
@@ -98,6 +98,9 @@ abstract class CloudCompute {
   /// Systems should assume deletion is best-effort and periodically list
   /// instances to delete instances that are no longer needed.
   Future<void> delete(String zone, String instanceName);
+
+  /// Cancel timers and free resources.
+  Future<void> close();
```
jonasfj, line 103: If the semantics of this method is that we "free resources and timers", then I'd expect to have no negative impact on anything.

If the idea that `CloudCompute` should not be used after calling `close()` then I think we should write this.

And I think we should ensure that all implementations of `CloudCompute` throws `StateError` after `close()` is called.

Additionally, we should have some documentation at class level or for each method documenting that when a `CloudCompute` instance is in the "closed"-state, then what happens? Also won't we have a "closing" state, where `close()` have been called, but current operations haven't finished yet.

Also unclear:
 * When does the `Future<void>` returned from `CloudCompute` resolve? 
 * Can we call `close()` multiple times?

jonasfj: I think it's better to find the place where we call: `.startInstanceExecution();` and then make sure that `.stopInstanceExecution` is called in a `registerScopeExitCallback`.

Rather than introducing a lot of new state in a high-level abstraction. Especially, since we don't need life-cycle management at this level (yet).

isoos:
Closing in favor of #6448.

PR #6445, opened 2023-02-22
Pana wrapper: log errors only with level > INFO

isoos:


isoos:
Closing in favor of #6446.

PR #6446, opened 2023-02-22
Reduce log-level in pana_wrapper

jonasfj:


pkg/pub_worker/lib/src/bin/pana_wrapper.dart
```diff
@@ -33,11 +33,23 @@ Future<void> main(List<String> args) async {
...
+    final prefix = '${rec.time} ${rec.level.name}:';
+    final printLinesWithPrefix = (String lines) {
+      for (final line in lines.split('\n')) {
+        print('$prefix $line');
+      }
+    };
+
+    printLinesWithPrefix(rec.message);
+    final e = rec.error;
+    if (e != null) {
```
isoos, line 47: `&& e.level > Level.INFO` ?

jonasfj: nah, now it'll be prefix INFO :D

PR #6451, opened 2023-02-24
Minimal session handling with new sign-in flow.

isoos:
- `state` is a `Map` encoded as the query parameters of an URI (after the `?` character)
- `UserSession.email` may also accept null in the future, although both `email` and `userId` is filled with `''` (empty string), and session handling logic also checks for that.
- removed `ClientSession` entity, its extra fields (e.g. expected `nonce`) is merged into `UserSession`
- fake service implements a minimal flow by taking a special `fake-email` parameter (only when running locally)
- implemented a minimal `go` target url and redirect

### DIFF
--- a/app/lib/account/default_auth_provider.dart
+++ b/app/lib/account/default_auth_provider.dart
@@ -2,7 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:convert' show json;
+import 'dart:convert';
 
 import 'package:clock/clock.dart';
 import 'package:googleapis/oauth2/v2.dart' as oauth2_v2;
@@ -112,7 +112,7 @@ class DefaultAuthProvider extends BaseAuthProvider {
           oauth2_v2.Oauth2Api.userinfoEmailScope,
           oauth2_v2.Oauth2Api.userinfoProfileScope,
         ].join(' '),
-        'state': Uri(queryParameters: state).toString().substring(1),
+        'state': encodeState(state),
         'nonce': nonce,
       },
     );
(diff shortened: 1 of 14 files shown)


isoos:
Note: the are still a few parts that are not implemented yet, but this is a sizeable part that can stand on its own.

app/lib/frontend/handlers/account.dart
```diff
@@ -57,31 +72,59 @@ Future<shelf.Response> signInCallbackHandler(shelf.Request request) async {
...
+  await accountBackend.updateClientSessionWithProfile(
+    sessionId: session.sessionId,
+    profile: profile,
+  );
+
+  // TODO: implement proper action on successful authentication
+  final go = state['go'];
+  if (go != null) {
+    // TODO: verify go
+    return redirectResponse(go);
```
jonasfj, line 117: Maybe, it would be wise to have some sanity checks here?

And not just redirect to anything.

app/lib/account/default_auth_provider.dart
```diff
@@ -112,7 +112,7 @@ class DefaultAuthProvider extends BaseAuthProvider {
           oauth2_v2.Oauth2Api.userinfoEmailScope,
           oauth2_v2.Oauth2Api.userinfoProfileScope,
         ].join(' '),
-        'state': state,
+        'state': Uri(queryParameters: state).toString().substring(1),
```
jonasfj, line 115: Should we not do something a bit more sane.

Like: `json + base64` or something?

isoos: Done.

app/lib/frontend/handlers/account.dart
jonasfj: nah,

jonasfj (approved):

PR #6452, opened 2023-02-27
Reject on upload: SDK lower constraint outside of the Dart 3 compatibility window.

### LINKED ISSUE #6450: Forbid publishing of packages with SDK constraint `>= 2.x <4.0.0`, where `x < 12`
Packages with SDK constraints that span 2.x where x < 12 and 3.x should not be allowed on pub.dev

isoos:
Fixes #6450.

### DIFF
--- a/pkg/pub_package_reader/lib/pub_package_reader.dart
+++ b/pkg/pub_package_reader/lib/pub_package_reader.dart
@@ -440,6 +440,8 @@ Iterable<ArchiveIssue> checkStrictVersions(Pubspec pubspec) sync* {
 
 final _preDart4 = VersionConstraint.parse('<4.0.0');
 final _firstDart4Pre = Version.parse('4.0.0').firstPreRelease;
+final _preNullSafety = VersionConstraint.parse('<2.12.0');
+final _postDart3 = VersionConstraint.parse('>=3.0.0-0');
 
 /// Checks if the version range is acceptable by current SDKs.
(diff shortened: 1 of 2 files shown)


pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -466,6 +468,12 @@ Iterable<ArchiveIssue> checkSdkVersionRange(Pubspec pubspec) sync* {
     yield ArchiveIssue(
         'The SDK constraint allows Dart 4.0.0, this is not allowed because Dart 4 does not exist.');
   }
+
+  if (!sdk.intersect(_preNullSafety).isEmpty &&
+      !sdk.intersect(_postDart3).isEmpty) {
+    yield ArchiveIssue(
+        'The SDK lower constraint is outside of the Dart 3 compatibility window (">=2.12.0").');
```
jonasfj, line 475:
```suggestion
        'An SDK lower bound constraint less than 2.12.0 is not compatible with Dart 3.\n'
        'Either bump the lower bound SDK constraint to `>=2.12.0`, or\n'
        'lower the upper bound SDK constraint to `<3.0.0`.');
```

jonasfj (approved):

pkg/pub_package_reader/lib/pub_package_reader.dart
jonasfj: Or something like that... we could possibly create the constraints and write them out to make it more obvious.

PR #6453, opened 2023-02-27
Topics upload validation

szakarias:


pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -760,3 +760,55 @@ Iterable<ArchiveIssue> checkFunding(String pubspecContent) sync* {
...
+
+    if (item.length > 32) {
+      yield ArchiveIssue(
+          'Invalid `topics` value (`$item`): name is too long (over 32 characters).');
+      continue;
+    }
+
+    if (topics.where((x) => x == item).length > 1) {
+      yield ArchiveIssue(
+          'Invalid `topics` value (`$item`): name must only be present once.');
```
jonasfj, line 802: This will be reported twice if we don't `break` here.

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -760,3 +760,55 @@ Iterable<ArchiveIssue> checkFunding(String pubspecContent) sync* {
...
+    }
+
+    final RegExp regExp = RegExp(r'^[a-z]([a-z0-9]|\-(?=[^\-]))+[a-z0-9]$');
+    if (!regExp.hasMatch(item)) {
+      yield ArchiveIssue(
+          'Invalid `topics` value (`$item`): must consist of lowercase '
+          'alphanumerical characters or dash (but no double dash), starting '
+          'with a-z and ending with a-z or 0-9.');
+      continue;
+    }
```
jonasfj, line 812: nit: you'd be better off splitting this into multiple `if` statements..
At-least `if (topic.contains('--')` is easier than having the complicated regexp, the rest is probably fine :D

jonasfj (approved):

pkg/pub_package_reader/lib/pub_package_reader.dart
szakarias: done.

pkg/pub_package_reader/lib/pub_package_reader.dart
szakarias: Added comments to the regexp

PR #6454, opened 2023-02-27
Emit, send and parse csrf-token.

isoos:


pkg/web_app/lib/src/deferred/http.dart
```diff
@@ -13,20 +15,61 @@ Client createAuthenticatedClient(Future<String?> Function() getToken) {
   return _AuthenticatedClient(getToken);
 }
 
-/// An [Client] which sends a `Bearer` token as `Authorization` header for each request.
-class _AuthenticatedClient extends BrowserClient {
-  final Future<String?> Function() _getToken;
+String? _getCsrfMetaContent() {
+  final values = document.head
+      ?.querySelectorAll('meta')
```
jonasfj, line 20: if you want to be fancy
```suggestion
      ?.querySelectorAll('meta[name="csrf-token"]')
```

jonasfj (approved):

PR #6466, opened 2023-03-01
Use existing session object - if present.

isoos:


app/lib/account/backend.dart
```diff
@@ -437,12 +437,32 @@ class AccountBackend {
     });
   }
 
-  /// Creates a new client session for pre-authorization secrets and
-  /// post-authorization user information.
-  Future<SessionData> createNewClientSession({
+  /// Updates an existing or creates a new client session for pre-authorization
+  /// secrets and post-authorization user information.
+  Future<SessionData> createOrUpdateClientSession({
```
jonasfj, line 442: What if the session already has a user associated? There be race conditions here.

isoos: I think sending the user for re-authentication is a valid use-case, so having a userId before the authentication is not an issue. However, that's a good point, we should handle conflicts when updating the profile with the authentication result. I guess it should be just an exception on the request + also invalidating the session (or all of the user's session?)

isoos: Updated code + test to reset the session when there is a new `userId` after authentication, otherwise we can keep the session. We could probably get away by resetting the session on every authentication, but I think it safe and also user friendly this way.

jonasfj (approved):

PR #6470, opened 2023-03-02
async loading of SessionData inside RequestContext, explicit dependency on rendering

isoos:
- async/lazy loading of `SessionData`
- prepares seamless migration to new session cookies
- makes rendering of pages with `SessionData` explicit (while previously the deepest part referenced the `requestContext` in the scope)
- fixed a template test that lacked session profile


isoos:
Closing this in favor of #6472. Let's do the on-demand loading and the explicit rendering after we have implemented the new sign-in.

PR #6472, opened 2023-03-03
Merged (sync) session data in RequestContext.

isoos:
- prepares seamless migration to new session cookies
- fixed a template test that lacked session profile

### DIFF
--- a/app/test/frontend/golden/pkg_admin_page.html
+++ b/app/test/frontend/golden/pkg_admin_page.html
@@ -50,8 +50,17 @@
         </form>
       </div>
       <div class="site-header-nav scroll-container">
-        <div class="nav-login-container">
-          <button id="-account-login" class="nav-main-button link">Sign in</button>
+        <div id="-account-profile" class="nav-container nav-my-container hoverable">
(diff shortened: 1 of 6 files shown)


isoos:
this is an alternative to #6470

app/test/frontend/golden/pkg_admin_page.html
```diff
@@ -50,8 +50,17 @@
         </form>
       </div>
       <div class="site-header-nav scroll-container">
-        <div class="nav-login-container">
-          <button id="-account-login" class="nav-main-button link">Sign in</button>
+        <div id="-account-profile" class="nav-container nav-my-container hoverable">
+          <button class="nav-main-button">My pub.dev</button>
```
jonasfj, line 54: Is this intentional? just asking.

jonasfj (approved):

app/test/frontend/golden/pkg_admin_page.html
isoos: Yes, the template test was rendering an admin page, and the rendering reflected the signed-in user. Fixed that in the templates_test.dart + golden update.

PR #6477, opened 2023-03-06
Double check entries on integrity problems.

### LINKED ISSUE #6455: Fix timing issues in integrity checks
Issues like this can be reported intermittently, let's fix that.
```
integrity.check: [pub-integrity-problem] Moderated package "****" also present in active packages.
integrity.check: [pub-integrity-problem] PackageVersionAsset "******/0.0.3/changelog" is not referenced from PackageVersionInfo.
integrity.check: [pub-integrity-problem] PublisherMember "*********" references a non-existing `publisherId`: ""*********.com".
```

isoos:
Fixes #6455.

### DIFF
--- a/app/lib/shared/integrity.dart
+++ b/app/lib/shared/integrity.dart
@@ -11,6 +11,7 @@ import 'package:gcloud/storage.dart';
 import 'package:http/http.dart' as http;
 import 'package:logging/logging.dart';
 import 'package:pool/pool.dart';
+import 'package:pub_dev/publisher/backend.dart';
 
 import '../account/agent.dart';
 import '../account/backend.dart';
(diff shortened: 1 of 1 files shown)


sigurdm (commented):
We have earlier used a strategy of ignoring entities that where recent (eg. <24h old) could that be a better fit here?

jonasfj (approved):

jonasfj:
If we data created timestamps, maybe we don't on membership entities.

isoos:
> We have earlier used a strategy of ignoring entities that where recent (eg. <24h old) could that be a better fit here?

I'm not sure: which entity check are you referring to?

I think having an explicit check is better than ignoring potential issues, especially as fake tests will be considered recent all the time...

PR #6482, opened 2023-03-07
Delete all session when using new signin.

isoos:


isoos:
Note: this is not called directly yet.
Note: we should consider to also check the CSRF token presence.

jonasfj (approved):

jonasfj:
Should we consider doing:
 * `GET /sign-out`
 * `GET /sign-in?action=switch-user

Should these be `POST` instead of `GET`?

isoos:
> `GET /sign-out` and `GET /sign-in?action=switch-user`

The first would open up an attack-vector to force-sign-out users by including an `` on the page. I'd even strengthen the current `DELETE` with CSRF check...

PR #6484, opened 2023-03-08
Support more parameters in OpenID Connect request.

isoos:


jonasfj (approved):

app/lib/account/auth_provider.dart
```diff
@@ -49,6 +49,8 @@ abstract class AuthProvider {
   Future<Uri> getOauthAuthenticationUrl({
     required Map<String, String> state,
     required String nonce,
+    required bool promptConsent,
```
jonasfj, line 52: I think you want an enum with `none, consent, select_account`

jonasfj: Or just pass a string, and do a:
```
if (['none', 'consent', 'select_account'].contains(prompt)) {
  throw ArgumentError('prompt', prompt);
}
```

isoos: I think we never want `none` (removing the option of prompt?) or `select_account` (why would we want to know of other accounts of the user?)... The `consent` is useful in cases where we want to force-refresh the authentication information though...

PR #6494, opened 2023-03-09
Verify redirect URL in the new sign-in flow.

isoos:


app/lib/account/models.dart
```diff
@@ -201,6 +201,9 @@ class SessionData {
   /// The time when the session will expire.
   final DateTime expires;
 
+  /// The time when the session was last authenticated.
+  final DateTime? authenticated;
```
jonasfj, line 205:
```suggestion
  final DateTime? authenticatedAt;
```

jonasfj (approved):

app/lib/account/models.dart
isoos: oh, that is for the other PR, I'll remove it

PR #6496, opened 2023-03-13
Re-authenticate sessions that don't have a recent authentication.

isoos:
- Created a shared method that handles both unauthenticated clients and sessions where authentication happened a long time ago.
- Updated some tests to use new session cookies and also testing the redirect (on one use case).

### DIFF
--- a/app/lib/account/models.dart
+++ b/app/lib/account/models.dart
@@ -149,6 +149,9 @@ class UserSession extends db.ExpandoModel<String> {
   @db.DateTimeProperty(required: true)
   DateTime? expires;
 
+  @db.DateTimeProperty(indexed: false)
+  DateTime? authenticatedAt;
+
   @db.StringProperty(indexed: false)
   String? csrfToken;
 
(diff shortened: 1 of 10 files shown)


app/lib/account/models.dart
```diff
@@ -201,6 +201,9 @@ class SessionData {
   /// The time when the session will expire.
   final DateTime expires;
 
+  /// The time when the session was last authenticated.
+  final DateTime? authenticated;
```
jonasfj, line 205:
```suggestion
  final DateTime? authenticatedAt;
```

app/lib/frontend/handlers/account.dart
```diff
@@ -410,3 +416,34 @@ Future<shelf.Response> accountMyActivityLogPageHandler(
...
+  if (needsReAuthentication) {
+    final requestedUri = request.requestedUri;
+    final goUri = Uri(
+      path: requestedUri.path,
+      query: requestedUri.hasQuery ? requestedUri.query : null,
+    );
+    final signInUri = Uri(path: '/sign-in', queryParameters: {
+      'go': goUri.toString(),
+    });
+    print(signInUri);
```
jonasfj, line 444: debug leftover?

jonasfj (approved):

PR #6498, opened 2023-03-13
Restrict default logger.

isoos:
I've found that local (and CI) logs are way too verbose to deal with them reasonably, let's restrict them to a minimum set.

### DIFF
--- a/app/lib/shared/logging.dart
+++ b/app/lib/shared/logging.dart
@@ -30,6 +30,8 @@ void setupDebugEnvBasedLogging() {
     return;
   }
   _envBasedLoggingSetupDone = true;
+  // The default value here is a hack for integration tests to work.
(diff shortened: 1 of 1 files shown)


app/lib/shared/logging.dart
```diff
@@ -31,7 +31,7 @@ void setupDebugEnvBasedLogging() {
   }
   _envBasedLoggingSetupDone = true;
   // ignore: invalid_use_of_visible_for_testing_member
-  final debugEnv = envConfig.debug ?? '*';
+  final debugEnv = envConfig.debug ?? 'fake_server pub.email';
```
jonasfj, line 34: Please leave a comment explaining that this is a hack for integration tests to work.

Or maybe, we should change the integration test fake email to use `print()` for the line it needs to output for the tests to work?

jonasfj (approved):

PR #6506, opened 2023-03-16
Reduced logging in fake/logging email sender.

isoos:


app/lib/frontend/email_sender.dart
```diff
@@ -56,8 +57,12 @@ class _LoggingEmailSender implements EmailSender {
         'from ${message.from} '
         'to ${message.recipients.join(', ')}';
 
+    final urls = _simpleUrlRegExp
+        .allMatches(message.bodyText)
+        .map((e) => e.group(0))
+        .toList();
     _logger.info('Not sending email (SMTP not configured): '
-        '$debugHeader\n${message.bodyText}.');
+        '$debugHeader${urls.map((e) => '\n$e').join('')}');
```
jonasfj, line 65: Since this is supposed to be used by the integration test... how about just doing:
 `print('[email-urks] ${urls.join(' ')}')`, then you won't have to parse the logging stuff.. And integration tests won't be broken because someone sets `DEBUG=` to a specific value.

jonasfj (approved):

app/lib/frontend/email_sender.dart
isoos: I'd rather have a final fix with file or HTTP-based communication, and if we did switch to print, it may be overlooked. Let's merge this as-is, and hopefully the fix would be not that complex.

PR #6509, opened 2023-03-17
Client-side warning for nearing session timeout.

isoos:
- new API call to get session information
- new `PageData` field to indicate on which pages the session expiry warning is interesting
- timer-based polling of session data that creates a dismissible panel on the top of the page


isoos:
Pushing this PR to staging...

pkg/_pub_shared/lib/data/page_data.dart
```diff
@@ -20,11 +20,13 @@ class PageData {
   final String? consentId;
   final PkgData? pkgData;
   final PublisherData? publisher;
+  final bool? sessionAware;
```
jonasfj, line 23: Can we give this a documentation comment?

pkg/web_app/lib/src/account.dart
```diff
@@ -23,13 +24,69 @@ late final useNewSignin = _newSigningMetaContent == '1';
...
+void _initSessionMonitor() {
+  // No need to monitor session on non-editable pages,
+  // where no action or state-change may be started.
+  if (!pageData.isSessionAware) {
+    return;
+  }
+
+  final checkFrequency = Duration(minutes: 5);
+  final sessionExpiresThreshold = Duration(minutes: 30);
+  final authenticationThreshold = Duration(minutes: 45);
```
jonasfj, line 44: 55 minutes should be fine, right?

pkg/web_app/lib/src/account.dart
```diff
@@ -23,13 +24,69 @@ late final useNewSignin = _newSigningMetaContent == '1';
...
 
+void _initSessionMonitor() {
+  // No need to monitor session on non-editable pages,
+  // where no action or state-change may be started.
+  if (!pageData.isSessionAware) {
+    return;
+  }
+
+  final checkFrequency = Duration(minutes: 5);
+  final sessionExpiresThreshold = Duration(minutes: 30);
```
jonasfj, line 43: let's leave this at 45 min

pkg/web_app/lib/src/account.dart
```diff
@@ -23,13 +24,69 @@ late final useNewSignin = _newSigningMetaContent == '1';
...
 }
 
+void _initSessionMonitor() {
+  // No need to monitor session on non-editable pages,
+  // where no action or state-change may be started.
+  if (!pageData.isSessionAware) {
+    return;
+  }
+
+  final checkFrequency = Duration(minutes: 5);
```
jonasfj, line 42: It might be better to just make a `while` loop and then do `await Future.delayed(sessionExpiresThreshold.subtract(now.subtract(authenticatedAt)))`

or something like that. Since there is no need to refresh in the meantime.

jonasfj (approved):

pkg/web_app/lib/src/account.dart
isoos: Added a TODO for it.

PR #6510, opened 2023-03-20
New tool: delete publisher.

isoos:


app/lib/admin/tools/delete_publisher.dart
```diff
@@ -0,0 +1,52 @@
...
+
+final _argParser = ArgParser()
+  ..addFlag('help', abbr: 'h', defaultsTo: false, help: 'Show help.')
+  ..addOption('publisher', help: 'name of publisher to delete');
+
+Future<String> executeDeletePublisher(List<String> args) async {
+  final argv = _argParser.parse(args);
+
+  if (argv['help'] as bool) {
+    return 'Delete a publisher using admin rights.\n${_argParser.usage}';
```
sigurdm, line 21: Can we add some details to the :

- Can only delete publishers with no packages.
- Will not leave a tombstone, the publisher can be recreated.

jonasfj (approved):

PR #6511, opened 2023-03-21
Disable like button for unathenticated users.

isoos:


sigurdm:
Do we have a tool-tip or other guidance for letting the user know they need to log in first?

jonasfj (approved):
I'm okay with this, but I would prefer that we do a popup, that asks the user to sign-in, and contains a link to /sign-in

Like: `<a href="/sign-in">Sign-in</a> to like packages`.

isoos:
> I'm okay with this, but I would prefer that we do a popup, that asks the user to sign-in, and contains a link to /sign-in

Agreed, but let's do that after we've removed the old signin from the code.

PR #6530, opened 2023-03-28
Cleanup of pkg/web_app

isoos:


pkg/web_app/lib/src/deferred/http.dart
```diff
@@ -9,10 +9,9 @@ import 'package:http/http.dart';
 
 export 'package:http/http.dart' show Client;
 
-/// Creates an authenticated [Client] that calls [getToken] to obtain an
-/// bearer-token to use in the `Authorization: Bearer <token>` header.
-Client createAuthenticatedClient(Future<String?> Function() getToken) {
-  return _AuthenticatedClient(getToken);
+/// Creates an authenticated [Client] that extends request with the CSRF header.
+Client createAuthenticatedClient() {
```
jonasfj, line 13: This is not an authenticated client... let's call it something like: ClientWithCsrf or something.

I think it's fine to land this, and rename it in a follow up PR :D

jonasfj (approved):

PR #6534, opened 2023-03-29
Cleanup old session cookie and its update API call.

isoos:


app/lib/frontend/handlers/account.dart
```diff
@@ -134,66 +127,6 @@ Future<ClientSessionStatus> getAccountSessionHandler(
...
-      changed: true,
-      expires: newSession.expires,
-      authenticatedAt: null,
-    ).toJson(),
-    headers: session_cookie.createUserSessionCookie(
-        newSession.sessionId, newSession.expires),
-  );
-}
-
 /// Handles DELETE /api/account/session
```
jonasfj, line 130: So in the future, it'd make a lot of sense if we move this to `/sign-out`

jonasfj (approved):

app/lib/frontend/handlers/account.dart
isoos: Maybe, however I don't see much benefit to gain there. Also, that could be an annoying attack if somebody sends you to the /sign-out link and now the session is signed out.

PR #6536, opened 2023-03-30
Better session status monitoring in web_app.

isoos:


pkg/web_app/lib/src/account.dart
```diff
@@ -21,10 +23,16 @@ void _initSessionMonitor() {
...
     return;
   }
+  // No need to monitor if the page was created with an unauthenticated session.
+  if (_signInButton != null) {
+    return;
+  }
 
-  final checkFrequency = Duration(minutes: 5);
-  final sessionExpiresThreshold = Duration(minutes: 45);
+  final defaultCheckDelay = Duration(minutes: 5);
```
jonasfj, line 31:
```suggestion
  final minCheckDelay = Duration(minutes: 5);
```

It is also used as minimum else where... we should never refresh more frequently than this.

PR #6549, opened 2023-04-03
Reduce integration test dependence on logging: file based emails, direct print.

isoos:


pkg/pub_integration/lib/src/fake_pub_server_process.dart
```diff
@@ -131,19 +126,25 @@ class FakePubServerProcess {
...
-}
-
-/// Matches the output line.
-typedef LineMatcher = bool Function(String line);
 
-class _LinePattern {
-  final LineMatcher matcher;
-  final completer = Completer<String>();
-  _LinePattern(this.matcher);
+  Future<List<Map<String, Object?>>> readAllEmails() async {
```
jonasfj, line 135: Can we leave a doc comment, saying what this returns?
```suggestion
  /// Returns a list of all emails sent by this [FakePubServer].
  ///
  /// Each email is a JSON object on the form:
  /// ```js
  /// {
  ///   "from": "<email>",
  ///   "uuid": "<message-id>", // optional
  ///   "recipients": ["<email>", ...,
  ///   "subject": "...",
  ///   "bodyText": "..."
  /// }
  /// ```
  Future<List<Map<String, Object?>>> readAllEmails() async {
```

app/lib/shared/email.dart
```diff
@@ -116,6 +116,16 @@ class EmailMessage {
       throw ArgumentError('Invalid uuid: `$uuid`');
     }
   }
+
+  Map<String, Object?> toJson() {
+    return {
+      if (uuid != null) 'uuid': uuid,
```
jonasfj, line 122: Wouldn't the field name `localMessageId` be more correct. This is:
https://en.wikipedia.org/wiki/Message-ID

-----------

Unrelated to this PR, but should we perhaps rename `EmailMessage.uuid` into `EmailMessage.localMessageId`?

Perhaps the docs for `localMessageId` should also link to https://www.jwz.org/doc/mid.html

jonasfj (approved):

app/lib/shared/email.dart
isoos: Renamed, added a TODO for a follow-up checking the recommendations on the link.

PR #6550, opened 2023-04-04
Started work on pub_integration TestUser + some cleanup.

isoos:
- refactored fake credential creation to follow the new API
- removed now-obsolete `fake_platform.js` and its initialization
- started work on fake `TestUser`, leaving the rest of it after other pending PRs get merged


pkg/pub_integration/lib/src/fake_test_user.dart
```diff
@@ -0,0 +1,103 @@
...
+import 'package:_pub_shared/pubapi.dart';
+import 'package:http/http.dart' as http;
+import 'package:pub_integration/src/fake_credentials.dart';
+import 'package:pub_integration/src/headless_env.dart';
+import 'package:pub_integration/src/pub_puppeteer_helpers.dart';
+import 'package:puppeteer/puppeteer.dart';
+
+import 'test_scenario.dart';
+
+class FakeTestUser implements TestUser {
```
jonasfj, line 14: Let's make this private. I don't see any reason why anyone should ever do `user is FakeTestUser`, is there a reason to do this?

pkg/pub_integration/lib/src/fake_test_user.dart
```diff
@@ -0,0 +1,103 @@
...
+    // TODO: implement readLatestEmail after #6549 lands.
+    throw UnimplementedError();
+  }
+
+  @override
+  Future<Map<String, Object?>> createCredentials() async {
+    return fakeCredentialsMap(email: email);
+  }
+
+  static Future<FakeTestUser> create({
```
jonasfj, line 41: I'd suggest making this a top-level `createFakeUser` function instead, so we don't have to make `FakeTestUser` a public class.

jonasfj (approved):

PR #6554, opened 2023-04-06
Rate limit package uploads (per package rates).

isoos:
- prevents new versions within a minute
- prevents more than 12 versions within an hour (5 minute / version)
- prevents more than 24 versions within a day (1 version / hour)
- check is conditional on a request context attribute, as we need to allow these events for the tests, separate tests to check the rates being limited
- these checks are based on the cached versions list, making it relatively cheap for the average use case


app/lib/package/backend.dart
```diff
@@ -1173,6 +1180,34 @@ class PackageBackend {
...
+          package.name!, 12, 'one hour');
+    }
+
+    final oneDayAgo = now.subtract(Duration(days: 1));
+    final lastDayCount =
+        publishedTimestamps.where((ts) => ts.isAfter(oneDayAgo)).length;
+    if (lastDayCount >= 24) {
+      throw PackageRejectedException.rateLimitReached(
+          package.name!, 24, 'one day');
+    }
```
jonasfj, line 1208:
```suggestion
    final existingVersions = await listVersionsCached(package.name!);
    
    // Count number of versions published since [someTimeAgo]
    int countPublishedSince(DateTime someTimeAgo) {
      return existingVersions
        .versions.map((v) => v.published)
        .whereNotNull()
        .where((ts) => ts.isAfter(someTimeAgo))
        .length;
    }
    
    if (countPublishedSince(clock.ago(minutes: 1)) >= 1) {
      throw PackageRejectedException.rateLimitReached(
        package.name!, 1, 'one minute');
    }

    if (countPublishedSince(clock.ago(hours: 1)) >= 12) {
      throw PackageRejectedException.rateLimitReached(
          package.name!, 12, 'one hour');
    }
    
    if (countPublishedSince(clock.ago(days: 1)) >= 24) {
      throw PackageRejectedException.rateLimitReached(
          package.name!, 24, 'one day');
    }
```

app/lib/frontend/request_context.dart
```diff
@@ -46,6 +47,9 @@ class RequestContext {
   /// The status of the client session cookie.
   final ClientSessionCookieStatus clientSessionCookieStatus;
 
+  /// Whether to check rate limits for this request.
+  final bool checkRateLimits;
```
jonasfj, line 51: How about making the rate-limit configurable?

So that it is specified in the config file... We have 3 limits... test probably only need to increase one or two of these limits.

jonasfj (approved):

app/lib/frontend/request_context.dart
jonasfj: > Maybe we could add @visibleForTesting on the setter of the flag?

It's not only read in testing.

isoos:
Implemented this with the new configuration in #6612.

PR #6563, opened 2023-04-12
Updated HeadlessEnv + FakeTestUser.

isoos:
- `HeadlessEnv` exposes fewer methods, mostly the `withPage()` is public.
- `TestUser` takes `HeadlessEnv`, as `puppeteer:Browser` won't be enough to handle all the customized logic around coverage and third-party access.
- Refactored / implemented to get the latest email body for `TestUser`.


pkg/pub_integration/lib/src/test_scenario.dart
```diff
@@ -68,7 +68,7 @@ abstract class TestUser {
 
   /// Get a headless browser where this test user is signed-in to their
   /// Google Account.
-  Browser get browser;
+  HeadlessEnv get browser;
```
jonasfj, line 71: How about just exposing:
```dart
Future<T> withBrowserPage<T>(FutureOr<T> Function(Page page) fn);
```

Do test scenario authors have any business calling `Browser.close()` anyways?


And perhaps make `serverErrors` and `clientErrors` expando properties on `Page`?

Ideally, then you can call `page.serverErrors.clear()` before returning from `withBrowserPage` if intentionally making a test that has a failure.

isoos: Done with the rename. As discussed offline, we don't need to expose serverErrors or clientErrors for the test cases.

isoos:
Applied #6568 here with new method signatures. PTAL.

pkg/pub_integration/lib/src/headless_env.dart
```diff
@@ -12,15 +12,15 @@ import 'package:puppeteer/puppeteer.dart';
...
   final String? _coverageDir;
   final Directory _tempDir;
-  final bool debug;
+  final bool _displayBrowser;
   Browser? _browser;
-  final clientErrors = <ClientError>[];
-  final serverErrors = <String>[];
-  late final bool trackCoverage =
+  final _clientErrors = <ClientError>[];
+  final _serverErrors = <String>[];
```
jonasfj, line 22: Unrelated to this PR, but shouldn't these be scoped to `withPage` ?

What if there is concurrent invocations of `withPage`, or what if an invocation fails and the test case ignores the error and makes another `withPage` invocation (then that'll fail for no reason).

jonasfj (approved):

pkg/pub_integration/lib/src/headless_env.dart
isoos: Good point, moving them into the withPage context.

PR #6568, opened 2023-04-14
Do not subclass TestUser

jonasfj:
Instead of subclassing, which just create classes that don't need to exist.

Thoughts?


isoos:
This was implemented in #6563.

PR #6569, opened 2023-04-17
Configuration for rate limits.

isoos:
I've implemented a parsed version along the lines of `X in Y minutes`, but in practice I think we'll end up with short, medium and long-term limits anyway, so I don't think we should have the complexity of parsing here. Keeping just a few integers seems to be enough.

### DIFF
--- a/app/config/dartlang-pub.yaml
+++ b/app/config/dartlang-pub.yaml
@@ -73,6 +73,6 @@ rateLimits:
     daily: 24
   - operation: package-published
     scope: user
-    burst: 16
-    hourly: 256
-    daily: 1024
+    burst: 20
+    hourly: 200
+    daily: 250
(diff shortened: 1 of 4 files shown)


app/config/dartlang-pub.yaml
```diff
@@ -65,3 +65,14 @@ tools:
...
+  - operation: package-published
+    scope: package
+    burst: 1
+    hourly: 12
+    daily: 24
+  - operation: package-published
+    scope: user
+    burst: 16
+    hourly: 256
+    daily: 1024
```
jonasfj, line 78:
```suggestion
    burst: 20
    hourly: 200
    daily: 250
```

jonasfj (approved):

app/config/dartlang-pub.yaml
jonasfj: we'll probably want to setup a meeting to discuss these limits.

jonasfj:
>  but in practice I think we'll end up with short, medium and long-term limits anyway, so I don't think we should have the complexity of parsing here. Keeping just a few integers seems to be enough.

Yeah, that makes sense.

PR #6570, opened 2023-04-17
Log when markdown rendering takes longer than 10 seconds.

isoos:


jonasfj (approved):

app/lib/shared/markdown.dart
```diff
@@ -53,19 +53,27 @@ String? markdownToHtml(
...
+      baseUrl: baseUrl,
+      baseDir: baseDir,
+    );
+    if (isChangelog) {
+      nodes = _groupChangelogNodes(nodes).toList();
+    }
+    return _renderSafeHtml(nodes,
+        inlineOnly: inlineOnly, disableHashIds: disableHashIds);
+  } finally {
+    if (sw.elapsed.inSeconds >= 10) {
```
jonasfj, line 73: I would even say 3 seconds, possibly 1 second seems bad.

PR #6572, opened 2023-04-18
Better logs indicating if a build was triggered

jonasfj:


jonasfj:
Just an attempt at simpler logs that are slightly more useful.

jonasfj:
I do wonder if it's smart to call `updateLocalBuiltFilesIfNeeded` from `_main`, as multiple of these may be started concurrently, thus, we risk concurrent builds.

isoos:
> I do wonder if it's smart to call updateLocalBuiltFilesIfNeeded from _main, as multiple of these may be started concurrently, thus, we risk concurrent builds.

Good point. It will be better if we wouldn't depend on it in the tests, e.g. have an exception for the validation rule that would otherwise throw. We will have a check for it in the integration tests anyway...

PR #6583, opened 2023-04-21
Reduce markdown parsing.

isoos:
- Dart code rendering is simple to replace with a pre+code block.
- The error handling in request processing has fewer risk if it uses direct rendering of the content.

### DIFF
--- a/app/lib/shared/handler_helpers.dart
+++ b/app/lib/shared/handler_helpers.dart
@@ -8,6 +8,7 @@ import 'dart:io';
 import 'package:appengine/appengine.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as path;
+import 'package:pub_dev/frontend/templates/misc.dart';
 import 'package:shelf/shelf.dart' as shelf;
 import 'package:shelf/shelf_io.dart' as shelf_io;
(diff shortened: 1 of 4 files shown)


app/lib/shared/handler_helpers.dart
```diff
@@ -171,22 +173,25 @@ shelf.Handler _logRequestWrapper(Logger logger, shelf.Handler handler) {
...
+          d.text('Please open an issue: '),
+          d.a(href: issueUrl, text: issueUrl),
+          d.p(text: 'Add these details to help us fix the issue:'),
+          d.codeSnippet(
+            language: 'text',
+            text: 'Requested URL: ${request.requestedUri}\n'
+                'Request ID: ${context.traceId}',
+          ),
+        ]),
+      ]);
```
jonasfj, line 190: Are we sure we don't want to prefer to use markdown instead of HTML... it's much harder to read this.

And the risk from rendering a bit of markdown is fairly tiny.

isoos: We could refactor this into a template method, so that it won't impact the readability of the error processing. I don't see it too bad to have it spelled out though...

PR #6601, opened 2023-04-26
Show topics in search results.

szakarias:


app/lib/frontend/templates/views/pkg/package_list.dart
```diff
@@ -230,7 +244,10 @@ d.Node _item({
         d.div(
           classes: ['packages-body'],
           children: [
-            d.div(classes: ['packages-description'], text: description),
+            d.div(
+              classes: ['packages-description'],
+              children: [d.span(text: description), ...topics],
```
isoos, line 249: I'm wondering if `<span>` is better here than `<p>`... the later would put the topics into a separate line though...

pkg/web_css/lib/src/_list.scss
```diff
@@ -241,6 +241,10 @@
     }
   }
 
+  .topics-tag {
+  padding: 4px 4px;
```
isoos, line 245: Let's also add `display: inline-block`, otherwise I'm not sure if the padding will get applied (or maybe it will, I'm just not sure).

isoos (approved):

app/lib/frontend/templates/views/pkg/package_list.dart
szakarias: We actually do want it on the same line, so I guess we keep it like this.

PR #6603, opened 2023-04-26
Always mock static files in app/ tests

isoos:
Instead of calling `updateLocalBuiltFilesIfNeeded` in every test file, they can all get the mocked version of static file cache, which is extended with the built files. The integration test will use the real built files, their presence will be tested.


app/lib/frontend/static_files.dart
```diff
@@ -113,6 +114,23 @@ class StaticFileCache {
...
   }
 
+  @visibleForTesting
+  factory StaticFileCache.forTests() {
+    final properCache = StaticFileCache.withDefaults();
+    final cache = StaticFileCache();
+    final paths = <String>{
+      ...properCache.keys,
+      '/static/css/style.css',
+      '/static/js/script.dart.js',
```
jonasfj, line 124: Why do these need special casing?

jonasfj (approved):

PR #6606, opened 2023-04-27
Remove duplicate dependency used for byte stream reading.

isoos:


app/lib/fake/server/fake_storage_server.dart
```diff
@@ -78,7 +78,7 @@ class FakeStorageServer {
       await for (final m in stream) {
         final disposition = _parse(m.headers['content-disposition']);
         final name = disposition['name']!;
-        final data = await readAsBytes(m);
+        final data = await readByteStream(m);
```
jonasfj, line 81: For this purpose there is also:
https://pub.dev/documentation/async/latest/async/collectBytes.html

I didn't use that in the `TaskBackend` because I wanted something I could give a maximum number of bytes.

jonasfj (approved):

PR #6612, opened 2023-05-01
Rate limit using AuditLogRecords.

isoos:


app/lib/audit/backend.dart
```diff
@@ -26,6 +26,9 @@ AuditBackend get auditBackend => ss.lookup(#_auditBackend) as AuditBackend;
 /// Represents the backend for the audit handling and authentication.
 class AuditBackend {
   final DatastoreDB _db;
+  _CachedRecords _cachedRecords =
+      _CachedRecords(clock.now().subtract(Duration(days: 1000)), []);
```
jonasfj, line 30:
```suggestion
  var _cachedRecords = Future.value(_CachedRecords(DateTime(0), []));
```

app/lib/audit/backend.dart
```diff
@@ -136,4 +139,56 @@ class AuditBackend {
...
+        .where((r) => now.difference(r.created!) < day)
+        .toList();
+
+    // update if current list is newer
+    if (_cachedRecords.timestamp.isBefore(now)) {
+      _cachedRecords = _CachedRecords(now, records);
+    }
+
+    return records;
+  }
```
jonasfj, line 186:
```suggestion
/// Returns the entries from the last day.
  ///
  /// Keeps the [_cachedRecords] fields updated, and lists only the entries
  /// up to the last query.
  Future<List<AuditLogRecord>> getEntriesFromLastDay() async {
    final now = clock.now().toUtc();
    final cachedRecords = await _cachedRecords.onError((e, _) {
      // Ignore error, we'll just not used the cached result.
      // If we don't ignore the error, then we won't be able to recover from an
      // error -- and, thus, will enter a state were all operations fail!
      return _CachedRecords(DateTime(0), []);
    });

    // fast track for requests within 5 seconds:
    if (now.difference(cachedRecords.updated) < Duration(seconds: 5)) {
      return cachedRecords.records;
    }

    _cachedRecords = Future.microtask(() async {
      var since = cachedRecords.updated;
      if (now.difference(since) > Duration(days: 1)) {
        since = now.subtract(Duration(days: 1));
      }
      // Always allow for a little clock drift
      since = since.subtract(Duration(minutes: 5));

      final query = dbService.query<AuditLogRecord>()
        ..filter('created >', since);
      final newRecords = await query.run().toList();

      // ;erge records from cachedRecords and newRecords
      final records = cachedRecords.records
          .where((r) =>
              now.difference(r.created!) < Duration(days: 1) &&
              newRecords.any((nr) => nr.id! == r.id!))
          .followedBy(newRecords)
          .toList(growable: false);

      return _CachedRecords(now, List.unmodifiable(records));
    });

    return (await _cachedRecords).records;
  }
```

app/lib/audit/backend.dart
```diff
@@ -136,4 +139,56 @@ class AuditBackend {
...
+}
+
+class _CachedRecords {
+  final DateTime timestamp;
+  final List<AuditLogRecord> records;
+
+  _CachedRecords(this.timestamp, this.records);
 }
```
jonasfj, line 194:
```suggestion
class _CachedRecords {
  final DateTime updated;
  final List<AuditLogRecord> records;

  _CachedRecords(this.updated, this.records);
}
```

app/lib/audit/backend.dart
```diff
@@ -136,4 +139,56 @@ class AuditBackend {
...
+  /// Keeps the [_cachedRecords] fields updated, and lists only the entries
+  /// up to the last query.
+  Future<List<AuditLogRecord>> getEntriesFromLastDay() async {
+    final now = clock.now().toUtc();
+    final cachedAge = now.difference(_cachedRecords.timestamp);
+
+    // fast track for requests within 5 seconds:
+    if (cachedAge.inSeconds < 5) {
+      return _cachedRecords.records;
+    }
```
jonasfj, line 154: I don't mind by then we have concurrent query operations... that's perhaps a bit much :rofl:

app/lib/service/rate_limit/rate_limit.dart
```diff
@@ -0,0 +1,150 @@
...
+/// if the limit has been exceeded.
+Future<void> verifyPackageUploadRateLimit({
+  required AuthenticatedAgent agent,
+  required String package,
+}) async {
+  final operation = AuditLogRecordKind.packagePublished;
+  if (agent.email != null) {
+    await _verifyRateLimit(
+      rateLimit: _getRateLimit(operation, RateLimitScope.user),
+      dataFilters: {'email': agent.email!},
```
jonasfj, line 25: Why not filter on the user-id ?

isoos: service accounts share the same agent id, but have different email that we can filter on

app/lib/shared/exceptions.dart
```diff
@@ -337,6 +337,20 @@ class OperationForbiddenException extends ResponseException {
...
+  RateLimitException({
+    required int maxCount,
+    required String windowAsText,
+  }) : super._(
+          429,
+          'RateLimit',
+          'The operation is blocked, as rate limit in the current window '
+              'has been reached ($maxCount in the $windowAsText). '
+              'Please try again later.',
+        );
```
jonasfj, line 351: We could opt to send a `Retry-After: $windowAsSeconds` header

app/lib/shared/redis_cache.dart
```diff
@@ -329,6 +329,22 @@ class CachePatterns {
...
+  /// until the block should be applied.
+  Entry<DateTime> rateLimitUntil({
+    required String entryKey,
+    required Duration window,
+  }) =>
+      _cache
+          .withPrefix('rate-limit-until/$entryKey/')
+          .withTTL(window)
+          .withCodec(utf8)
+          .withCodec(json)
```
jonasfj, line 343: the json codec is probably overkill.

app/lib/shared/redis_cache.dart
```diff
@@ -329,6 +329,22 @@ class CachePatterns {
             encode: (OpenIdData v) => v.toJson(),
             decode: (v) => OpenIdData.fromJson(v as Map<String, dynamic>),
           ))[configurationUrl];
+
+  /// When a rate limit is triggered, the cached entry holds the time
+  /// until the block should be applied.
+  Entry<DateTime> rateLimitUntil({
+    required String entryKey,
+    required Duration window,
```
jonasfj, line 337: Does `window` need to be here? Don't that just make it slower?

I think it's just `rateLimitUntil(String operation, RateLimitScope scope, String value)` where `value` is either a userid or a package name, depending on `scope`.

isoos: Done. (it can be used in the `set` method as TTL)

app/lib/shared/exceptions.dart
isoos: Done.

app/lib/audit/backend.dart
isoos: Implemented, but with a different code.

app/lib/audit/backend.dart
```diff
@@ -26,6 +26,9 @@ AuditBackend get auditBackend => ss.lookup(#_auditBackend) as AuditBackend;
 /// Represents the backend for the audit handling and authentication.
 class AuditBackend {
   final DatastoreDB _db;
+  var _cachedRecords = _CachedRecords(DateTime(0), []);
+  Future<void>? _cacheRecordsUpdateFuture;
```
jonasfj, line 30: Why not:
```suggestion
  Future<_CachedRecords> _cachedRecords = Future.value(_CachedRecords(DateTime(0), []));
```
Then you always honor the invariant by awaiting it, less risk of making a mistake.

jonasfj, line 30: At-least do:
```suggestion
  var _cachedRecords = _CachedRecords(DateTime(0), []);
  Future<void> _cacheRecordsUpdateFuture = Future.value(null);
```
So you never have to handle the null case

app/lib/audit/backend.dart
```diff
@@ -136,4 +139,68 @@ class AuditBackend {
...
+      if (cachedAge.inSeconds < 3) {
+        return _cachedRecords.records;
+      }
+
+      // we should have only one update running
+      _cacheRecordsUpdateFuture = _updateEntriesFromLastDay(
+        cachedAge: cachedAge,
+        oldRecords: _cachedRecords.records,
+      );
+      await _cacheRecordsUpdateFuture;
```
jonasfj, line 162: You need to await the future, then check if updates are necessary, and only then can you trigger an update. This will cause racing updates (not bad, just a waste of resources).

jonasfj: You won't have to check the timestamp twice, if you just do a `Future<_CachedRecords> _cachedRecords` and always await it.

app/lib/audit/backend.dart
```diff
@@ -136,4 +139,56 @@ class AuditBackend {
...
+    // merge records from cache and current query
+    final recordMap = <String, AuditLogRecord>{};
+    for (final r in _cachedRecords.records) {
+      recordMap[r.id!] = r;
+    }
+    for (final r in current) {
+      recordMap[r.id!] = r;
+    }
+    // filter records that are within the last day
+    final records = recordMap.values
```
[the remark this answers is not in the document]
jonasfj: We could also use `Set` instead of map... like:
```dart
final newIds = newRecords.map(r => r.id).toSet();
final records = [
  ...cachedRecords.records.where(r => !newIds.contains(r.id)),
  ...newRecords,
];
```
Or even as simple as:

```dart
final ids = <String>{};
final records = [
  ...cachedRecords.records,
  ...newRecords,
].where(r => ids.add(r.id!)).toList();
```

The ideal, would of course be something like:
```dart
extension UniqueList<T> on List<T> {
  List<T> uniqueBy<S>(S Function(T element) byKey) {
    final uniqueKeys = <S>{};
    return this.where(e => uniqueKeys.add(byKey(e))).toList();
  }
}

// then simply do:
final records = [
  ...cachedRecords.records,
  ...newRecords,
].uniqueBy(r => r.id!);
```

But downside of set/map is that we do allocations, and these are probably more expensive than just looping through the small list of newRecords many times..

app/lib/audit/backend.dart
isoos: I've found it confusing when reading the code: we would use the same variable for (a) getting the timestamp of its update (and the value) and (b) waiting for it to be updating. I'd rather keep the "update is happening right now" variable separate, easier to write a conditional around it.

app/lib/audit/backend.dart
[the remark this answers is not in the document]
isoos: Updated to use the current id set as a filter.

app/lib/service/rate_limit/rate_limit.dart
isoos: Updated to hopefully use this logic. PTAL.

app/lib/audit/backend.dart
```diff
@@ -136,4 +139,65 @@ class AuditBackend {
...
+
+    final query = dbService.query<AuditLogRecord>()
+      ..filter('created >', now.subtract(window));
+    final current = await query.run().toList();
+
+    // merge records from cache and current query
+    final currentIds = current.map((e) => e.id!).toSet();
+    final records = [
+      ...oldRecords.where((r) =>
+          !currentIds.contains(r.id!) && now.difference(r.created!) < day),
```
jonasfj, line 190: nit, not sure if it's formatted any better :D
```suggestion
      ...oldRecords
          .where((r) => !currentIds.contains(r.id!)),
          .where((r) => now.difference(r.created!) < day),
```

app/lib/service/rate_limit/rate_limit.dart
```diff
@@ -0,0 +1,149 @@
...
+    return;
+  }
+
+  final cacheKeyParts = [
+    rateLimit.operation,
+    rateLimit.scope.name,
+    if (package != null) 'package-$package',
+    if (userId != null) 'userId-$userId',
+  ];
+  final entryKey = Uri(pathSegments: cacheKeyParts).toString();
```
jonasfj, line 69:
```suggestion
  final cacheEntry = cache.rateLimitUntil([
    rateLimit.operation,
    rateLimit.scope.name,
    if (package != null) 'package-$package',
    if (userId != null) 'userId-$userId',
  ].join('/'));
  
  final current = await cacheEntry.get();
  if (current != null && current.isAfter(clock.now())) {
      throw RateLimitException(...
  }
```

Calling `auditBackend.getEntriesFromLastDay();` can incur some cost and pain, we should probably check the cache before we do an expensive operation.

I do see the problem that our cache doesn't contain the window or windowAsText for rate-limit that was applied.

**Proposal:** We make the `RateLimitException` serializable to json and put it in the cache :rofl:

app/lib/service/rate_limit/rate_limit.dart
```diff
@@ -0,0 +1,149 @@
...
+  return activeConfiguration.rateLimits?.firstWhereOrNull(
+    (r) => r.operation == operation && r.scope == scope,
+  );
+}
+
+Future<void> _verifyRateLimit({
+  required RateLimit? rateLimit,
+  String? package,
+  String? userId,
+}) async {
```
jonasfj, line 53:
```suggestion
}) async {
  assert(userId != null || package != null);
```

jonasfj (approved):

jonasfj (approved):

app/lib/service/rate_limit/rate_limit.dart
jonasfj: We can also leave this as TODO for later.

PR #6617, opened 2023-05-02
Add has:topic tag when the pubspec has entries.

isoos:


app/lib/package/model_properties.dart
```diff
@@ -173,6 +173,7 @@ class Pubspec {
   }
 
   late final List<Uri> funding = _inner.funding ?? const <Uri>[];
+  late final hasTopic = topics != null && topics!.isNotEmpty;
```
sigurdm, line 176: Could be
```suggestion
  late final hasTopic = topics?.isNotEmpty ?? false;
```
To avoid null-assert

Issue #6614, opened 2023-05-02
Let's add a "Dart 3 incompatible" badge


jonasfj:
![image](https://user-images.githubusercontent.com/149732/235615360-303500e3-78b0-4a37-908f-6fcadcb82d2a.png)

We probably need this when we switch to Dart 3.0 beta to do stable analysis.

jonasfj:
For packages that are pre-null-safety (lower bound SDK constraint `<2.12`) we don't want to run analysis.

We should give them a dummy analysis result.
 * zero score
 * tagged `is:dart3-incompatible`


Later, we can remove `is:dart3-incompatible` and make these `is:legacy`.

PR #6619, opened 2023-05-02
Tag and badge for Dart 3 incompatible packages.

isoos:
#6614


app/lib/package/model_properties.dart
```diff
@@ -135,6 +135,13 @@ class Pubspec {
               .intersect(VersionConstraint.parse('>=2.0.0'))
               .isEmpty);
 
+  late final isDart3Incompatible =
+      !supportsOnlyLegacySdk && // do not mix is:legacy with is:dart3-incompatible
+          _hasDartSdkConstraint &&
+          !_dartSdkConstraint!
+              .intersect(VersionConstraint.parse('<2.12.0'))
```
jonasfj, line 142: I'm guessing this is slightly better:
```suggestion
              .intersect(VersionConstraint.parse('<2.12.0-0'))
```

app/lib/tool/test_profile/import_source.dart
```diff
@@ -129,7 +129,7 @@ class _AutoGeneratedImportSource implements ImportSource {
     final hasRepository = hasher('hasRepository', max: 20) > 0;
     final isLegacy = version.contains('legacy');
 
-    final sdkConstraint = isLegacy ? '>=1.12.0 <2.0.0' : '>=2.6.0 <3.0.0';
+    final sdkConstraint = isLegacy ? '>=1.12.0 <2.0.0' : '>=2.13.0 <3.0.0';
```
jonasfj, line 132: It might be wise to do this in a separate PR. Not that I see anything scary in the golden files.

jonasfj (approved):

app/lib/tool/test_profile/import_source.dart
isoos: It will change the golden files either way: if we don't change this, the goldens will show the dart3-incompatible badge.

PR #6621, opened 2023-05-03
Add daily task to scan topics

szakarias:


app/lib/tool/neat_task/pub_dev_tasks.dart
```diff
@@ -157,6 +157,8 @@ void _setupGenericPeriodicTasks() {
     ),
   );
 
+  _daily(name: 'topics', isRuntimeVersioned: false, task: countTopics);
```
jonasfj, line 160:
```suggestion
  _daily(name: 'count-topics', isRuntimeVersioned: false, task: countTopics);
```

jonasfj: @isoos there is some try-catch sanity around this, right?

jonasfj (approved):

app/lib/tool/neat_task/pub_dev_tasks.dart
isoos: Yeah, it will setup it in a sane way.

szakarias: done

PR #6624, opened 2023-05-04
Add pub.dev/topics page

szakarias:


szakarias:
image

app/lib/frontend/handlers/misc.dart
```diff
@@ -67,6 +73,25 @@ Future<shelf.Response> readinessCheckHandler(shelf.Request request) async {
...
+        .readAsBytes(topicsJsonFileName);
+
+    topics = utf8JsonDecoder.convert(data) as Map<String, int>;
+  } on FormatException {
+    topics = {};
+  } on DetailedApiRequestError catch (e, st) {
+    if (e.status == 404) {
+      topics = {};
+    }
+  }
```
jonasfj, line 91:
```suggestion
  } on FormatException {
    topics = {};
    // TODO: Log error loading topics
  } on DetailedApiRequestError catch (e, st) {
    topics = {};
    if (e.status == 404) {
      // TODO: Log inability to load topics
    }
  }
```

jonasfj (approved):

app/lib/frontend/handlers/misc.dart
jonasfj: We can do the TODO later, I don't think that's particularly blocking.

If we want it we just need:
```dart
import 'package:logging/logging.dart' show Logger;

final _log = Logger('pub.handlers.misc');
```
at the top of the file.

szakarias: added loggin.

app/lib/frontend/templates/views/page/topics_list.dart
```diff
@@ -0,0 +1,42 @@
...
+d.Node renderTopicsList(Map<String, int> topics) {
+  return d.div(children: [
+    d.h1(text: 'Topics '),
+    topicsListNode(topics),
+  ]);
+}
+
+/// Renders the package list node of /topics page.
+d.Node topicsListNode(Map<String, int> topics) {
+  final sortedTopics = SplayTreeMap<String, int>.from(
```
isoos, line 20: Isn't this sorted in the file that we store in the bucket?

isoos, line 20: `topics.entries.toList()..sort((a, b) => -a.value.compareTo(b.value))` may be also good here, not sure which consumes less CPU

szakarias: No.

szakarias: Let's do this. It's also easier to read.

isoos (approved):

PR #6630, opened 2023-05-08
cache topics page

szakarias:


sigurdm (approved):
lgtm

app/lib/frontend/handlers/misc.dart
```diff
@@ -76,24 +76,29 @@ Future<shelf.Response> readinessCheckHandler(shelf.Request request) async {
...
+        _log.severe('Failed to load topics.json, error : ', e, st);
+      }
     }
-  }
-  return htmlResponse(renderTopicsPage(topics));
+
+    return renderTopicsPage(topics);
+  });
+
+  return htmlResponse(html!);
```
sigurdm, line 101: Not sure why `cache.get()` returns a nullable... I guess that's just the way it is.

PR #6637, opened 2023-05-09
Status for API docs

jonasfj:
This is not perfect, but once migrated we can put this into the template... no need to fetch it from clientside.

Or maybe we want to load whether docs was successfully generated... I'm open to discussing this.
Perhaps it's something we could include in the pana summary?

Also current UI doesn't have good support for: "we chose not to process this version".

### DIFF
--- a/app/lib/frontend/handlers/custom_api.dart
+++ b/app/lib/frontend/handlers/custom_api.dart
@@ -46,13 +46,13 @@ Future<shelf.Response> apiDocumentationHandler(
     final status = await taskBackend.packageStatus(package);
     return jsonResponse({
       'name': package,
-      'versions': status.versions
-          .map((s) => {
-                'version': s.version,
-                'status': s.status == TaskPackageVersionStatus.pending ||
-                        s.status == TaskPackageVersionStatus.running
+      'versions': status.versions.entries
+          .map((e) => {
+                'version': e.key,
+                'status': e.value == PackageVersionStatus.pending ||
+                        e.value == PackageVersionStatus.running
                     ? 'pending'
-                    : (s.status == TaskPackageVersionStatus.failed
+                    : (e.value == PackageVersionStatus.failed
                         ? 'failed'
                         : 'completed'),
               })
--- a/app/lib/frontend/handlers/custom_api.dart
+++ b/app/lib/frontend/handlers/custom_api.dart
@@ -38,10 +38,11 @@ Future<shelf.Response> apiDocumentationHandler(
     return jsonResponse({}, status: 404);
   }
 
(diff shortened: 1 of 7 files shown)


app/lib/frontend/handlers/custom_api.dart
```diff
@@ -36,6 +38,27 @@ Future<shelf.Response> apiDocumentationHandler(
     return jsonResponse({}, status: 404);
   }
 
+  if (requestContext.experimentalFlags.showSandboxedOutput) {
+    if (!await packageBackend.isPackageVisible(package)) {
```
isoos, line 42: This could go outside of the `if` experimental block.

app/lib/task/backend.dart
```diff
@@ -900,6 +905,45 @@ class TaskBackend {
...
+          var status = TaskPackageVersionStatus.completed;
+          if (s.attempts > 0 && s.attempts < taskRetryLimit) {
+            // attempts > 0 && attempts < taskRetryLimit  ==> pending
+            status = TaskPackageVersionStatus.pending;
+          } else if (s.scheduled
+              .add(Duration(days: 31))
+              .isBefore(clock.now())) {
+            // scheduled + 31 days < now                  ==> pending
+            status = TaskPackageVersionStatus.pending;
+          } else if (s.attempts == 0 && s.scheduled == initialTimestamp) {
```
isoos, line 929: Maybe move this pending to the top of the if conditional? It seems like this is the default state.

app/lib/task/backend.dart
```diff
@@ -900,6 +905,45 @@ class TaskBackend {
...
+            // attempts == 0 && scheduled == init         ==> pending
+            status = TaskPackageVersionStatus.pending;
+          } else if (s.attempts >= taskRetryLimit) {
+            // attempts >= taskRetryLimit                 ==> failed
+            status = TaskPackageVersionStatus.failed;
+          } else {
+            // attempts == 0                              ==> completed
```
isoos, line 936: Btw. I find this a bit confusing that we are using the attempt count as a status indicator. It may be worth to add a separate flag that gets updated?

isoos (approved):
Mostly nits.

app/lib/task/backend.dart
jonasfj: Perhaps, but if we want to change the state, I think we should do this later.

jonasfj: Now I remember we I didn't do this... obviously, after I started to try.

You cannot store the state when the state changes under you. A task is pending, if it haven't been analyzed in a certain amount of time. So storing field that says `state = "completed"` will be wrong a some point in time in the future.

PR #6657, opened 2023-05-15
Merging fake server process and headless env in integration tests.

isoos:
- This is one step towards having all integration tests into a somewhat uniform format.
- Note: This still uses a separate browser process for each user session, that change is coming in a follow-up PR.


pkg/pub_integration/lib/src/fake_test_scenario.dart
```diff
@@ -0,0 +1,75 @@
...
+
+import 'fake_pub_server_process.dart';
+import 'fake_test_user.dart';
+import 'headless_env.dart';
+import 'test_scenario.dart';
+
+/// The timeout factor that should be used in integration tests.
+final testTimeoutFactor = 6;
+
+class FakeTestScenario {
```
jonasfj, line 19:
```suggestion
class TestContextProvider {
```

Or `TestRunner`?

PR #6673, opened 2023-05-19
Integration test's uploader management on the UI instead of API calls.

isoos:


pkg/pub_integration/lib/src/pub_puppeteer_helpers.dart
```diff
@@ -136,6 +136,44 @@ extension PubPageExt on Page {
...
+        clicked = true;
+        break;
+      }
+    }
+    if (!clicked) {
+      throw Exception('Email "$email" was not found in the uploaders list.');
+    }
+    await waitAndClickOnDialogOk(waitForOneResponse: true);
+    await waitAndClickOnDialogOk();
+    await _waitForModelHidden();
```
jonasfj, line 174: Can we do comments like:
```suggestion
    // Click ok in the dialog to confirm
    await waitAndClickOnDialogOk(waitForOneResponse: true);
    // Click ok on the popup saying it was done
    await waitAndClickOnDialogOk();
    await _waitForModelHidden();
```

jonasfj (approved):

PR #6684, opened 2023-05-23
Increase instance count of search instances.

isoos:


pkg/pub_package_reader/lib/src/names.dart
```diff
@@ -39,6 +39,7 @@ const reservedWords = <String>{
   'implements',
   'import',
   'in',
+  'inline',
```
sigurdm, line 42: Does this change belong?

isoos: It is a CI issue that needs to be fixed, either here or another one.

isoos: removed, added new PR https://github.com/dart-lang/pub-dev/pull/6686

sigurdm (approved):

PR #6688, opened 2023-05-23
Move ngram out if inner loop

jonasfj:
Does this make sense

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -480,15 +480,15 @@ class PackageNameIndex {
     final singularWord = word.length <= 3 || !word.endsWith('s')
         ? word
         : word.substring(0, word.length - 1);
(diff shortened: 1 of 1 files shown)


app/lib/search/mem_index.dart
```diff
@@ -480,15 +480,15 @@ class PackageNameIndex {
     final singularWord = word.length <= 3 || !word.endsWith('s')
         ? word
         : word.substring(0, word.length - 1);
+    final parts = singularWord.length <= 3
```
sigurdm, line 483: nit: Should this have been lazy (late)? we only use it if we get past the first condition inside the loop.

isoos: I'd think that most cases will trigger it anyway, but it wouldn't hurt to have that calculated only on-demand.

PR #6690, opened 2023-05-23
Faster iteration and Map creation.

isoos:


app/lib/search/token_index.dart
```diff
@@ -97,32 +97,22 @@ class Score {
...
       return this;
     }
-    final result = <String, double>{};
-    for (String key in _values.keys) {
-      final value = _values[key]!;
-      if (value < threshold) continue;
-      result[key] = value;
-    }
+    final result = Map<String, double>.fromEntries(
+        _values.entries.where((entry) => entry.value >= threshold!));
```
jonasfj, line 101: The `threshold!` null check assertion shouldn't be necessary.
We could `final t = threshold!` on the line before, if the analyzer can't get it from `if(threshold == null) return this`.

app/lib/search/token_index.dart
```diff
@@ -97,32 +97,22 @@ class Score {
...
-  Score project(Iterable<String> keys) {
-    final result = <String, double>{};
-    for (String key in keys) {
-      final value = _values[key];
-      if (value == null) continue;
-      result[key] = value;
-    }
+  Score project(Set<String> keys) {
+    final result = Map<String, double>.fromEntries(
+        _values.entries.where((entry) => keys.contains(entry.key)));
```
jonasfj, line 108: Iteration over `keys` might be faster, if the set is significant smaller?

jonasfj (approved):

PR #6693, opened 2023-05-23
Install libwebp from binary release

szakarias:


tool/setup-webp.sh
```diff
@@ -0,0 +1,30 @@
...
+    mkdir -p ${TARGET_DIRECTORY}
+    if [ "$?" != "0" ]; then
+        exit -1
+    fi
+}
+
+function download_and_extract() {
+    curl --retry 8 -L ${LIBWEBP_TAR_URI} | \
+      tar -xzv --strip-components=2 -C ${TARGET_DIRECTORY} --\
+      libwebp-${LIBWEBP_VERSION}-linux-x86-64/bin
```
jonasfj, line 25:
```suggestion
    curl --retry 8 -L "${LIBWEBP_TAR_URI}" | \
      tar -xzv --strip-components=2 -C "${TARGET_DIRECTORY}" --\
      "libwebp-${LIBWEBP_VERSION}-linux-x86-64/bin"
```

tool/setup-webp.sh
```diff
@@ -0,0 +1,30 @@
...
+
+}
+# Entry
+ensure_target_directory
+download_and_extract
```
jonasfj, line 30: Let's add a test that checks that `webp` and `gif2webp` are present and executable in the `TARGET_DIRECTORY`.

If not, print an error and exit non zero.

jonasfj (approved):

tool/setup-webp.sh
jonasfj: that way, if in the future the contents of the tar.gz changes it we'll get an error.

tool/setup-webp.sh
szakarias: done

tool/setup-webp.sh
szakarias: done

Issue #6692, opened 2023-05-23
trigram matching in search is extremely slow


jonasfj:
When we call `searchWord` for each word in the search query, and we then create `parts` as trigrams for each word.
And for each word for each package for each trigram in word, we do `String.contains`, it'll be slow.

https://github.com/dart-lang/pub-dev/blob/84a249aa64905b30775fb7febc78ab9988b092ca/app/lib/search/mem_index.dart#L492-L496

Side note: I'm not even sure this is sound, we don't convert the words to lowercase before splitting them into trigrams. So... those with uppercase will never match anything.

Perhaps we should do something much simpler here, and accept poor scores in favor of speed.

-----------------

If we wanted to search package names with trigrams we should probably convert all words into a single set of trigrams. Then represent that with `Set<String>` or a sorted list of `Uint32List` and for each package name do: 
```dart
for (var i = 0; i < packageName.length - 2; i++) {
  if (trigramSet.contains(packageName.substring(i, i + 3))) {
     // match
  }
}
```

or something even smarter.

-----------------

As what we have has severe degenerate performance, I'm okay with simply making our index dumber and scoring dumber, in favor of faster results.

@isoos should we perhaps consider aborting after each word in:
```dart
      for (final word in words) {
        final nameScore =
            _packageNameIndex.searchWord(word, packages: packages);
        final descr =
            _descrIndex.searchWords([word], weight: 0.90, limitToIds: packages);
        final readme = _readmeIndex
            .searchWords([word], weight: 0.75, limitToIds: packages);
        coreScores.add(Score.max([nameScore, descr, readme]));
      }
```

In `InMemoryPackageIndex._searchText` ?

sigurdm:
Can we close this for now?

isoos:
Let's close it!

PR #6694, opened 2023-05-24
Improve name index: lowercase + trigram forward index.

isoos:
- #6692
- simpler trigram parsing method (skips `Set` creation, keeps repeated items, but doesn't really affect the end result)
- trigram forward index with pre-parsed trigram set (created when name is added first time)
- some preprocessing for input and indexed names (lowercase + removing `_`)

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -445,11 +445,12 @@ class _TextResults {
 /// A simple (non-inverted) index designed for package name lookup.
 @visibleForTesting
 class PackageNameIndex {
+  final _data = <String, _PkgNameData>{};
+
   /// Maps package name to a reduced form of the name:
   /// the same character parts, but without `-`.
-  final _namesWithoutGaps = <String, String>{};
-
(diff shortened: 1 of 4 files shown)


app/lib/search/mem_index.dart
```diff
@@ -474,23 +478,26 @@ class PackageNameIndex {
...
+        continue;
+      }
+      if (entry.collapsed.contains(collapsedWord)) {
         values[pkg] = 1.0;
         continue;
       }
       var matched = 0;
       for (final part in parts) {
-        if (withoutGaps.contains(part)) {
+        if (entry.trigrams.contains(part)) {
```
jonasfj, line 500: Should we consider set intersection?

jonasfj (approved):

app/lib/search/mem_index.dart
jonasfj: also, the sane thing is probably an inverse index.

isoos: Set intersection produced worse results.
inverse index may work, but yeah, live updates are not in favor of it

PR #6696, opened 2023-05-24
Minimal improvement in search predicate iterators.

isoos:
#6692


app/lib/search/search_service.g.dart
```diff
@@ -43,7 +43,7 @@ Map<String, dynamic> _$PackageDocumentToJson(PackageDocument instance) =>
       'created': instance.created?.toIso8601String(),
       'updated': instance.updated?.toIso8601String(),
       'readme': instance.readme,
-      'tags': instance.tags,
+      'tags': instance.tags.toList(),
```
jonasfj, line 46: Are we missing the non-generated part of this?

isoos: Good catch, reverting it.

I was experimenting with converting the `List<String> tags` into a `Set` or `HashSet`, but neither provided enough difference to make it into the final PR.

PR #6703, opened 2023-05-26
Remove priority package name hit from sort hot path.

isoos:
- This has only a minor, but consistent effect when the priority hit is in the search result.
- Tried to also implement top-k search with `SplayTreeMap/Set` or keeping a shorter sorted list, and while those also provided some benefits, it made some queries worse.


jonasfj (approved):

isoos:
I'm dropping this as I did an error while benchmarking (left the dartdoc cutoff on). After the correct benchmark (and also the fix) this doesn't show improvement on the original code, and sometimes it takes longer to run.

PR #6716, opened 2023-05-30
Switch on sandbox

jonasfj:
Reverts dart-lang/pub-dev#6713 which reverted sandboxed output.


isoos (approved):
let's also make a note in changelog?

PR #6721, opened 2023-05-31
Update search snapshot periodically in a worker isolate.

isoos:
- reduces the load on the frontend isolate by not blocking the CPU when snapshot save happens
- also moves old snapshot GC into the worker isolate

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -76,7 +76,7 @@ class SearchBackend {
   ///
   /// When other process has the claim, the loop waits a minute before
   /// attempting to get the claim.
-  Future<void> updateSnapshotInForeverLoop() async {
+  Future<Never> updateSnapshotInForeverLoop() async {
     final lock = GlobalLock.create(
       'update-search-snapshot',
       expiration: Duration(minutes: 20),
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -82,20 +82,14 @@ class SearchBackend {
       expiration: Duration(minutes: 20),
     );
     for (;;) {
-      final claim = await lock.tryClaim();
-      if (claim == null) {
-        await Future.delayed(Duration(minutes: 1));
-        continue;
-      }
-
       try {
-        await doCreateAndUpdateSnapshot(claim);
+        await lock.withClaim((claim) async {
+          await doCreateAndUpdateSnapshot(claim);
+        });
       } catch (e, st) {
         _logger.warning('Snapshot update failed.', e, st);
       }
-
-      // cleanup
-      await claim.release();
+      await Future.delayed(Duration(minutes: 1));
     }
   }
 
@@ -104,22 +98,13 @@ class SearchBackend {
     GlobalLockClaim claim, {
     Duration sleepDuration = const Duration(minutes: 2),
   }) async {
-    final claimUpdateBefore = Duration(minutes: 10);
     final firstClaimed = clock.now();
-    final workUntil = firstClaimed.add(Duration(days: 1));
 
-    Future<void> refreshClaimIfNeeded() async {
-      final now = clock.now();
-      if (now.isAfter(workUntil)) {
-        // not updating claim anymore
-        return;
-      }
-      if (claim.expires.difference(now) < claimUpdateBefore) {
-        await claim.refresh();
-      }
-    }
+    // The claim will be released after a day, another process may
+    // start to build the snapshot from scratch again.
+    final workUntil = clock.now().add(Duration(days: 1));
 
-    // create snapshot from scratch
+    // creating snapshot from scratch
     final snapshot = SearchSnapshot();
     Future<void> updatePackage(String package, DateTime? updated) async {
       // Skip if the last document timestamp is before [updated].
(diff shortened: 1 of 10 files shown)


isoos:
PTAL, re-did the implementation to use global lock + continuous updates.

app/test/search/backend_test.dart
```diff
@@ -23,5 +25,34 @@ void main() {
...
+        _GlobalLockClaim(clock.now().add(Duration(seconds: 3))),
+        sleepDuration: Duration(milliseconds: 300),
+      );
+      await snapshotStorage.fetch();
+      expect(snapshotStorage.documents, isNotEmpty);
+    });
   });
 }
+
+class _GlobalLockClaim implements GlobalLockClaim {
```
jonasfj, line 42:
```suggestion
class _FakeGlobalLockClaim implements GlobalLockClaim {
```

app/lib/service/entrypoint/search.dart
```diff
@@ -72,3 +73,11 @@ Future _main(FrontendEntryMessage message) async {
 
   await runHandler(_logger, searchServiceHandler);
 }
+
+Future _worker(WorkerEntryMessage message) async {
+  message.protocolSendPort.send(WorkerProtocolMessage());
+  await popularityStorage.start();
+  setupSearchPeriodicTasks();
+  unawaited(searchBackend.updateSnapshotInForeverLoop());
+  await Completer().future; // never completes
```
jonasfj, line 82: Just curious, why not do?
```suggestion
  await searchBackend.updateSnapshotInForeverLoop();
```

app/lib/search/backend.dart
```diff
@@ -60,9 +61,139 @@ void registerPackageIndex(PackageIndex index) =>
...
+  /// the snapshot after every 10 minutes.
+  ///
+  /// When other process has the claim, the loop waits a minute before
+  /// attempting to get the claim.
+  Future<void> updateSnapshotInForeverLoop() async {
+    final lock = GlobalLock.create(
+      'update-search-snapshot',
+      expiration: Duration(minutes: 20),
+    );
+    for (;;) {
```
jonasfj, line 84:
```suggestion
    while (true) {
```

app/lib/search/backend.dart
```diff
@@ -60,9 +61,139 @@ void registerPackageIndex(PackageIndex index) =>
...
+      try {
+        await lock.withClaim((claim) async {
+          await doCreateAndUpdateSnapshot(claim);
+        });
+      } catch (e, st) {
+        _logger.warning('Snapshot update failed.', e, st);
+      }
+      await Future.delayed(Duration(minutes: 1));
```
jonasfj, line 92:
```suggestion
      // Wait for 1 minutes for sanity, before trying again.
      await Future.delayed(Duration(minutes: 1));
```

I assume this is just for sanity? to avoid a run away loop?

app/lib/search/backend.dart
```diff
@@ -60,9 +61,139 @@ void registerPackageIndex(PackageIndex index) =>
...
+      } on RemovedPackageException catch (_) {
+        snapshot.remove(package);
+      }
+    }
+
+    // initial scan of packages
+    await for (final package in dbService.query<Package>().run()) {
+      if (package.isNotVisible) continue;
+      if (!claim.valid) break;
+      await updatePackage(package.name!, null);
```
jonasfj, line 132: Why not pass in `updated` doesn't the `package.updated` property hold this value?

app/lib/search/backend.dart
```diff
@@ -60,9 +61,139 @@ void registerPackageIndex(PackageIndex index) =>
...
+    // start to build the snapshot from scratch again.
+    final workUntil = clock.now().add(Duration(days: 1));
+
+    // creating snapshot from scratch
+    final snapshot = SearchSnapshot();
+    Future<void> updatePackage(String package, DateTime? updated) async {
+      // Skip if the last document timestamp is before [updated].
+      // 1-minute window is kept to reduce clock-skew.
+      if (updated != null) {
+        final lastTimestamp = snapshot.documents![package]?.timestamp;
```
jonasfj, line 113: timestamp I can spot from the type, but it's the last time that this package was updated in search index, right?
```suggestion
        final lastUpdated = snapshot.documents![package]?.timestamp;
```

app/lib/search/backend.dart
```diff
@@ -60,9 +61,139 @@ void registerPackageIndex(PackageIndex index) =>
...
+        lastUploadedSnapshotTimestamp = snapshot.updated!;
+      }
+
+      await Future.delayed(sleepDuration);
+    }
+  }
+
+  Future<Map<String, DateTime>> _queryRecentlyUpdated(
+      DateTime lastQueryStarted) async {
+    final updatedThreshold = lastQueryStarted.subtract(Duration(minutes: 1));
```
jonasfj, line 172:
```suggestion
    final updatedThreshold = lastQueryStarted.subtract(Duration(minutes: 5));
```

jonasfj: I think it'd be wise to allow for some drift. My thinking was 15min, but 5 might be okay, 1 is a bit small... I'm just guessing :rofl:

jonasfj (approved):

app/lib/search/backend.dart
```diff
@@ -143,6 +279,19 @@ class SearchBackend {
...
+      p.updated,
+      scoreCard?.updated,
+    ].fold<DateTime?>(
+      null,
+      (p, v) {
+        if (v == null) return p;
+        if (p == null) return v;
+        return v.isAfter(p) ? v : p;
+      },
+    );
```
jonasfj, line 293: @isoos consider `whereNotNull()` and maybe even `max`, see:
https://pub.dev/documentation/collection/latest/collection/IterableComparableExtension.html


```suggestion
    // select the latest entity updated timestamp (when available)
    final sourceUpdated = [
      p.updated,
      scoreCard?.updated,
    ].whereNotNull().max;
```

Issue #6708, opened 2023-05-30
Seems some users of a publisher gets emails much faster than others


sigurdm:
Could be that some emails fail and get retried...

isoos:
In the past month or two I have all emails delivered shortly after the publishing event, the huge delay seems to be gone now.

PR #6722, opened 2023-05-31
Move email sending outside of a Timer scope.

isoos:
#6708


jonasfj (approved):
I don't think this changes anything.

PR #6734, opened 2023-06-07
Use only a single browser for multiple user profiles.

isoos:


pkg/pub_integration/lib/src/headless_env.dart
```diff
@@ -9,43 +9,6 @@ import 'package:_pub_shared/validation/html/html_validation.dart';
...
-  Future<void> close() async {
-    while (_profiles.isNotEmpty) {
-      await _profiles.removeLast().close();
-    }
-  }
-}
-
 /// Creates and tracks the headless Chrome environment, its temp directories and
 /// and uncaught exceptions.
 class HeadlessEnv {
```
jonasfj, line 14: I'd love it if we could give this class a descriptive name.

Maybe, Just `HeadlessBrowser` or just `TestBrowser` or something?

pkg/pub_integration/test/fake_sign_in_test.dart
```diff
@@ -46,8 +46,8 @@ void main() {
       final browserSession = await fakeTestScenario.createAnonymousTestUser();
       String? firstSessionId;
       // sign-in page
-      await browserSession.withBrowserPage(
-        (page) async {
+      await browserSession.withBrowserPage((page) async {
+        {
           final rs = await page.gotoOrigin('/sign-in?fake-email=user@pub.dev');
```
jonasfj, line 51: Isn't the idea that `withBrowserPage` will return a browser instance that is already signed-in for the given user?

pkg/pub_integration/test/fake_sign_in_test.dart
```diff
@@ -46,8 +46,8 @@ void main() {
       final browserSession = await fakeTestScenario.createAnonymousTestUser();
```
jonasfj, line 46: This is a `user` object, right? Sorry -- if the variable name confuses me :D

jonasfj (approved):

pkg/pub_integration/lib/src/headless_env.dart
isoos: Well, it should be called `BrowserWithCoverageTrackingAndCustomResponses`, but I would rather have a shorter name. `TestBrowser` is closer than just `HeadlessBrowser`, so renaming it for now...

PR #6739, opened 2023-06-08
Reduce String->int[] conversion when caching dartdoc content from sandbox.

isoos:


jonasfj (approved):

app/lib/shared/handlers.dart
```diff
@@ -64,11 +64,12 @@ shelf.Response jsonResponse(
 }
 
 shelf.Response htmlResponse(
-  String content, {
+  Object /* String | List<int> */ content, {
```
jonasfj, line 67: We could also make two methods?

jonasfj (approved):

PR #6746, opened 2023-06-09
Extend session expires/expired messages.

### LINKED ISSUE #6740: Banner text: "Your session will expire soon" is not informative/helpful
I got the banner by leaving an open admin tab for a while, and was unsure:

- what session the message was talking about (i did not explicitly start a session) - what is soon?
- what should I do when I see this warning? - not very actionable, is there any way to reset session without losing data?
- what are the consequences of an expired session?

I believe the Code is here:
https://github.com/dart-lang/pub-dev/blob/7707373028394fae78e046a6196cb4e745919eb6/pkg/web_app/lib/src/account.dart#L58

image

isoos:
Fixes #6740.

### DIFF
--- a/pkg/web_app/lib/src/account.dart
+++ b/pkg/web_app/lib/src/account.dart
@@ -53,7 +53,7 @@ void _initSessionMonitor() {
     String? displayMessage;
     if (expires == null || authenticatedAt == null) {
       displayMessage =
-          'Your session has expired: you may need to reload the page to resume editing.';
+          'Your session has expired: reload the page before making further administrative actions.';
     } else if (expires.isBefore(now.add(sessionExpiresThreshold)) ||
         authenticatedAt.isBefore(now.subtract(authenticationThreshold))) {
       displayMessage =
--- a/pkg/web_app/lib/src/account.dart
+++ b/pkg/web_app/lib/src/account.dart
@@ -52,10 +52,12 @@ void _initSessionMonitor() {
     final authenticatedAt = status.authenticatedAt;
     String? displayMessage;
     if (expires == null || authenticatedAt == null) {
-      displayMessage = 'Your session has expired.';
+      displayMessage =
+          'Your session has expired: you may need to reload the page to resume editing.';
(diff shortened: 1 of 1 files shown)


sigurdm (approved):

sigurdm:
I would still be a bit surprised as a user to know that I even initiated a session

Is it true that the session starts when going to the admin page?

pkg/web_app/lib/src/account.dart
```diff
@@ -52,10 +52,12 @@ void _initSessionMonitor() {
     final authenticatedAt = status.authenticatedAt;
     String? displayMessage;
     if (expires == null || authenticatedAt == null) {
-      displayMessage = 'Your session has expired.';
+      displayMessage =
+          'Your session has expired: you may need to reload the page to resume editing.';
```
sigurdm, line 56: Could it be?
```suggestion
          'Your admin session has expired: you may need to reload the page to resume editing.';
```

sigurdm: Don't you **have** to reload?
```suggestion
          'Your session has expired: reload the page before making further administrative actions.';
```

isoos:
> Don't you have to reload?

I don't think we've tested it yet, but the session handling is designed in a way that you could have opened two tabs, and if you are active on one tab, the other won't expire... so if your tab is closing on expiry, but you resume activity on the other, the next session status update refresh API call in the background would reflect that, and the message will disappear.

sigurdm:
> I don't think we've tested it yet, but the session handling is designed in a way that you could have opened two tabs, and if you are active on one tab, the other won't expire... so if your tab is closing on expiry, but you resume activity on the other, the next session status update refresh API call in the background would reflect that, and the message will disappear.

Let's try having a more clear messaging, and not according for an extremely rare case of two admin tabs, one keeping the other alive.

Just reloading the tab with the warning will always restart the session? I think that serves the 99.9%

Btw. Would it be possible to insert a link in the banner that reloads the session without a page-reload?

sigurdm:
Yeah - we probably should not spend too much time here.

Thanks for improving the message!

PR #6748, opened 2023-06-12
Preprocess dartdoc

jonasfj:


pkg/web_app/lib/src/autocompletion.dart
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:html';
+
+final _widgets = {
+  'autocompletion': _createAutocompletionWidget,
```
isoos, line 9: should this go in as a separate PR?

jonasfj: yes, this thing is not done yet :see_no_evil:

isoos (approved):

pkg/pub_dartdoc_data/lib/dartdoc_page.dart
```diff
@@ -0,0 +1,196 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
```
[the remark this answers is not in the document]
jonasfj: Yeah, it could be either place... Ideally, we probably want to do this in a background process anyways, so it can move back into `app/` later.

PR #6752, opened 2023-06-13
Code cleanup in pub_integration.

isoos:
- removing `FakeGoogleUser`, unused
- renamed `headless_env.dart` -> `test_browser.dart`
- merged `fake_test_user.dart` into `fake_test_context_provider.dart`, as it is the only place where it is being called


pkg/pub_integration/lib/src/fake_test_context_provider.dart
```diff
@@ -65,11 +67,88 @@ class TestContextProvider {
...
+  required String email,
+  required TestBrowser testBrowser,
+  required FakeEmailReaderFromOutputDirectory fakeEmailReader,
+  List<String>? scopes,
+}) async {
+  late PubApiClient api;
+  await testBrowser.withPage(fn: (page) async {
+    await page.fakeAuthSignIn(email: email, scopes: scopes);
+    api = await _apiClientHttpHeadersFromSignedInSession(page);
+  });
```
jonasfj, line 89: Maybe, we should do: `testBrowser.createSession` which returns a new incognito session.

Then we can use that session in `withBrowserPage` and in `api`, and we won't have to call `fakeAuthSignIn` more than once.. (I think an incognito mode session can have multiple pages/tags).

I think when we do actual integration tests we won't want to do the sign-in process more than once per user. It'll have non-trival overhead, as in it'll be slow.

jonasfj (approved):

pkg/pub_integration/lib/src/fake_test_context_provider.dart
isoos: Good point. I'll try to do it in a follow-up PR.

PR #6764, opened 2023-06-15
Add model and backend for security advisories

szakarias:


app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,77 @@
...
+      if (vulnerability.published != null) {
+        newAdvisory.published = DateTime.parse(vulnerability.published!);
+      } else {
+        newAdvisory.published = DateTime.now();
+      }
+
+      if (vulnerability.affected != null) {
+        newAdvisory.affectedPackages =
+            vulnerability.affected!.map((a) => a.package.name).toList();
+      }
```
jonasfj, line 67:
```suggestion
      newAdvisory.affected = (advisory.affected ?? []).map((a) => a.package.name).toList();
```

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,77 @@
...
+      if (oldAdvisory != null && oldAdvisory.modified!.isAfter(modified)) {
+        return oldAdvisory;
+      }
+
+      final newAdvisory = SecurityAdvisory()
+        ..id = vulnerability.id
+        ..modified = modified
+        ..parentKey = _db.emptyKey
+        ..osvJsonBlob = vulnerability;
+
```
jonasfj, line 52: Do we need to make sure that there no other advisories using `advisory.id` as an alias?

Shouldn't we also check that IDs listed as aliases here are not an id used for another advisory?

jonasfj (commented):
I think it'd be wise to do some extra checks during ingestion?

Overall this looks good.

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,77 @@
...
+      return query.run().toList();
+    });
+  }
+
+  Future<SecurityAdvisory?> lookupById(String id) async {
+    final key = _db.emptyKey.append(SecurityAdvisory, id: id);
+    return _db.lookupOrNull<SecurityAdvisory>(key);
+  }
+
+  Future<SecurityAdvisory> injestSecurityAdvisory(OSV vulnerability) async {
```
sigurdm, line 38: Did you mean
```suggestion
  Future<SecurityAdvisory> ingestSecurityAdvisory(OSV vulnerability) async {
```

szakarias: yes!

app/lib/service/security_advisories/models.dart
```diff
@@ -0,0 +1,259 @@
...
+  }
+
+  @override
+  bool validate(db.ModelDB db, Object? value) {
+    return super.validate(db, value) && (value == null || value is OSV);
+  }
+}
+
+@JsonSerializable(includeIfNull: false)
+class OSV {
```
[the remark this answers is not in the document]
szakarias: done

app/lib/service/security_advisories/models.dart
```diff
@@ -0,0 +1,259 @@
...
+part 'models.g.dart';
+
+@db.Kind(name: 'SecurityAdvisory', idType: db.IdType.String)
+class SecurityAdvisory extends db.Model<String> {
+  String get name => id!;
+
+  /// A list of IDs of the same vulnerability in other databases, including
+  /// the main id.
+  @db.StringListProperty()
+  List<String> aliases = <String>[];
```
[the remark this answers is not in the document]
szakarias: done

app/lib/service/security_advisories/models.dart
```diff
@@ -0,0 +1,259 @@
...
+
+import 'dart:convert' show json;
+
+import 'package:json_annotation/json_annotation.dart';
+
+import '../../shared/datastore.dart' as db;
+
+part 'models.g.dart';
+
+@db.Kind(name: 'SecurityAdvisory', idType: db.IdType.String)
```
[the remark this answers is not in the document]
szakarias: done

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,77 @@
...
+    return await withRetryTransaction(_db, (tx) async {
+      final modified = DateTime.parse(vulnerability.modified);
+      final oldAdvisory = await lookupById(vulnerability.id);
+
+      if (oldAdvisory != null && oldAdvisory.modified!.isAfter(modified)) {
```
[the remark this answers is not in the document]
szakarias: yup, we can do that, but then it is more readable to us  `isAtOrAfter`()

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,77 @@
...
+      final newAdvisory = SecurityAdvisory()
+        ..id = vulnerability.id
+        ..modified = modified
+        ..parentKey = _db.emptyKey
+        ..osvJsonBlob = vulnerability;
+
+      newAdvisory.aliases = [vulnerability.id];
+      if (vulnerability.aliases != null) {
+        newAdvisory.aliases.addAll(vulnerability.aliases!);
+      }
```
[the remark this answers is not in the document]
szakarias: Done.

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,77 @@
...
+
+      if (vulnerability.published != null) {
+        newAdvisory.published = DateTime.parse(vulnerability.published!);
+      } else {
+        newAdvisory.published = DateTime.now();
+      }
```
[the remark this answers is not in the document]
szakarias: As we talked about I would only do checks that are strictly necessary so we don't end up with errors on runtime. Ideally we should have the exact same advisories as those on Github. Otherwise I fear our service will not be useful.

app/lib/service/security_advisories/backend.dart
szakarias: As discussed, we will not do this for now. The plan is to import only import advisories from OSV, so potential deduplication is handled there.

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,77 @@
...
+
+  SecurityAdvisoryBackend(this._db);
+
+  Future<List<SecurityAdvisory?>> lookupSecurityAdvisories(
+      String package) async {
+    return await withRetryTransaction(_db, (tx) async {
+      final query = _db.query<SecurityAdvisory>()
+        ..filter('affectedPackages =', package);
+      return query.run().toList();
+    });
```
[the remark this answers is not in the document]
isoos: I don't think we need (or should) do transactions here.
Maybe caching could be useful?

app/lib/service/security_advisories/backend.dart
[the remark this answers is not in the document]
szakarias: Added some checks.

app/lib/service/security_advisories/backend.dart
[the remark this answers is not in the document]
szakarias: Done

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  Future<SecurityAdvisory?> lookupById(String id) async {
+    final key = _db.emptyKey.append(SecurityAdvisory, id: id);
+    return _db.lookupOrNull<SecurityAdvisory>(key);
```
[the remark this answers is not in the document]
szakarias: Skipping cache here for now. We might only be using the other lookup.

app/lib/service/security_advisories/models.dart
```diff
@@ -0,0 +1,270 @@
...
+  /// The time this entry was last modified.
+  @db.DateTimeProperty(required: true)
+  DateTime? modified;
+
+  /// A list of affected package names.
+  @db.StringListProperty()
+  List<String>? affectedPackages = <String>[];
+
+  @OSVProperty(required: true)
+  OSV? osvJsonBlob;
```
jonasfj, line 43: Why not just?
```suggestion
  OSV? osv;
```

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,113 @@
...
+      return query.run().map((e) => e.osvJsonBlob!).toList();
+    }))!;
+  }
+
+  Future<SecurityAdvisory?> lookupById(String id) async {
+    final key = _db.emptyKey.append(SecurityAdvisory, id: id);
+    return _db.lookupOrNull<SecurityAdvisory>(key);
+  }
+
+  List<String> validateAdvisoryErrors(OSV osv) {
```
jonasfj, line 45: Can't we do unit tests for this function.

We could even make this a helper function in a separate file, you don't need to setup the server to do this at all.

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,113 @@
...
+    final errors = <String>[];
+
+    if (DateTime.parse(osv.modified)
+        .isAfter(DateTime.now().add(Duration(hours: 1)))) {
```
jonasfj, line 49:
```suggestion
        .isAfter(clock.now().add(Duration(hours: 1)))) {
```

from `package:clock`, that way we can fake if we ever want to.

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,113 @@
...
+
+  Future<SecurityAdvisory?> lookupById(String id) async {
+    final key = _db.emptyKey.append(SecurityAdvisory, id: id);
+    return _db.lookupOrNull<SecurityAdvisory>(key);
+  }
+
+  List<String> validateAdvisoryErrors(OSV osv) {
+    final errors = <String>[];
+
+    if (DateTime.parse(osv.modified)
```
jonasfj, line 48: This can throw `FormatException` either we should document it, or handle it.

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,113 @@
...
+      errors.add('Invalid modified date, cannot be a future date.');
+    }
+
+    if (osv.id.length > 255) {
+      errors.add('Invalid id, id too long (over 255 characters).');
+    }
+
+    final invalids = <int>[];
+
+    osv.id.runes.forEach((element) {
```
jonasfj, line 59:
```suggestion
    // Check that [osv.id] consists of printable ASCII.
    osv.id.runes.forEach((element) {
```

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,113 @@
...
+      if (element < 32 || element > 126) {
+        invalids.add(element);
+      }
+    });
+
+    if (invalids.isNotEmpty) {
+      errors.add(
+          'Invalid id, contains non-printable ASCII code points: $invalids.');
```
jonasfj, line 67: I don't think you need to say what is invalid.

Just say: `The "id" property must be printable ASCII"`

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    return errors;
+  }
+
+  bool validateAdvisory(OSV osv) {
+    final errors = validateAdvisoryErrors(osv);
+    errors.forEach((error) => _logger.shout('[advisory-malformed]: $error'));
+    return errors.isEmpty;
+  }
```
jonasfj, line 81: Is both `validateAdvisoryErrors` and `validateAdvisory` intended to be public and used outside this backend?

Can we make them private or annotate them `@visibleForTestingOnly`.

jonasfj: Are they both internal or both public? or?

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,113 @@
...
+
+  Future<SecurityAdvisory?> ingestSecurityAdvisory(OSV osv) async {
```
jonasfj, line 83: Should we document some of the caveats and assumptions we're making?

Is this documentation correct? Are there other things we keep in mind.
```suggestion
  /// Stores the security advisory from [osv] in datastore.
  ///
  /// Overwrites existing advisory with the same [osv.id], if [osv] is newer.
  ///
  /// If [osv.id] is already listed as `alias` for another advisory, no action will
  /// be taken to resolve this. Instead both advisories will be stored and served.
  /// It's assumed that security advisory database owners take care to keep the
  /// security advisories sound, and that inconsistencies are intentional.
  Future<SecurityAdvisory?> ingestSecurityAdvisory(OSV osv) async {
```

app/lib/service/security_advisories/backend.dart
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    if (json.encode(osv.toJson()).length > 1024 * 500) {
+      errors.add('OSV too large (larger than 500 kB)');
+    }
+
+    return errors;
+  }
+
+  bool validateAdvisory(OSV osv) {
```
jonasfj, line 77:
```suggestion
  bool isValidAdvisory(OSV osv) {
```

Consider documenting what and why we validate... Maybe like:
```suggestion
  /// Sanity checks to ensure that we can store, lookup and update the advisory.
  ///
  /// We do validate all fields, instead we aim to simply ensure that the advisory
  /// is sufficiently sound that we can store it, look it up and update it in the future.
  bool isValidAdvisory(OSV osv) {
```

jonasfj (approved):
just nits

app/lib/service/security_advisories/backend.dart
szakarias: done.

app/lib/service/security_advisories/models.dart
szakarias: done

app/lib/service/security_advisories/backend.dart
szakarias: done

app/lib/service/security_advisories/backend.dart
szakarias: Done.

app/lib/service/security_advisories/backend.dart
szakarias: private

app/lib/service/security_advisories/backend.dart
szakarias: done

app/lib/service/security_advisories/backend.dart
szakarias: Added comment, and renamed

app/lib/service/security_advisories/backend.dart
szakarias: done.

app/lib/service/security_advisories/backend.dart
szakarias: Added todo.

PR #6765, opened 2023-06-15
Pool-bound forEach for Stream and Iterable

isoos:
- replaced most Pool use in app/
- also limiting the number of pending operations in the pool

### DIFF
--- a/app/lib/shared/utils.dart
+++ b/app/lib/shared/utils.dart
@@ -370,18 +370,19 @@ extension StreamBoundedForEach<T> on Stream<T> {
     FutureOr<void> Function(T) eachFn,
   ) async {
     final pool = Pool(concurrency);
-    final futures = Queue<Future>();
-    // keeps the size of the futures limited, allows some difference in the run times
-    final maxQueueSize = concurrency * 2;
+    final futures = <Future>[];
     try {
       await for (final item in this) {
         final f = pool.withResource(() => eachFn(item));
         futures.add(f);
-        while (futures.length > maxQueueSize) {
-          await futures.removeFirst();
+        if (futures.length >= concurrency) {
+          await Future.wait(futures);
+          futures.clear();
         }
       }
-      await Future.wait(futures);
+      if (futures.isNotEmpty) {
+        await Future.wait(futures);
+      }
     } finally {
       await pool.close();
     }
--- a/app/lib/shared/utils.dart
+++ b/app/lib/shared/utils.dart
@@ -16,6 +16,7 @@ import 'package:logging/logging.dart';
 // ignore: implementation_imports
 import 'package:mime/src/default_extension_map.dart' as mime;
 import 'package:path/path.dart' as p;
+import 'package:pool/pool.dart';
 import 'package:pub_semver/pub_semver.dart' as semver;
 import 'package:stream_transform/stream_transform.dart';
 
@@ -360,4 +361,39 @@ bool fixedTimeEquals(String a, String b) {
     result |= a.codeUnitAt(i) ^ b.codeUnitAt(i);
   }
   return result == 0;
+}
+
+extension StreamBoundedForEach<T> on Stream<T> {
+  /// Executes items using [eachFn] with maximum [concurrency].
+  Future<void> boundedForEach(
+    int concurrency,
+    FutureOr<void> Function(T) eachFn,
+  ) async {
+    final pool = Pool(concurrency);
+    final futures = Queue<Future>();
+    // keeps the size of the futures limited, allows some difference in the run times
+    final maxQueueSize = concurrency * 2;
+    try {
+      await for (final item in this) {
+        final f = pool.withResource(() => eachFn(item));
+        futures.add(f);
+        while (futures.length > maxQueueSize) {
+          await futures.removeFirst();
+        }
+      }
+      await Future.wait(futures);
+    } finally {
+      await pool.close();
+    }
+  }
+}
+
+extension IterableBoundedForEach<T> on Iterable<T> {
+  /// Executes items using [eachFn] with maximum [concurrency].
+  Future<void> boundedForEach(
+    int concurrency,
+    FutureOr<void> Function(T) eachFn,
+  ) async {
+    await Stream.fromIterable(this).boundedForEach(concurrency, eachFn);
+  }
 }
--- a/app/lib/shared/utils.dart
+++ b/app/lib/shared/utils.dart
@@ -16,7 +16,6 @@ import 'package:logging/logging.dart';
 // ignore: implementation_imports
 import 'package:mime/src/default_extension_map.dart' as mime;
 import 'package:path/path.dart' as p;
-import 'package:pool/pool.dart';
 import 'package:pub_semver/pub_semver.dart' as semver;
 import 'package:stream_transform/stream_transform.dart';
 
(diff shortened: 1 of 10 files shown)


app/lib/shared/utils.dart
```diff
@@ -361,3 +362,38 @@ bool fixedTimeEquals(String a, String b) {
...
+    final futures = Queue<Future>();
+    // keeps the size of the futures limited, allows some difference in the run times
+    final maxQueueSize = concurrency * 2;
+    try {
+      await for (final item in this) {
+        final f = pool.withResource(() => eachFn(item));
+        futures.add(f);
+        while (futures.length > maxQueueSize) {
+          await futures.removeFirst();
+        }
```
jonasfj, line 382: I don't understand this.

Reading: https://pub.dev/documentation/pool/latest/pool/Pool/close.html

I think it's sufficient to just do `await pool.close()`.

```dart
    final pool = Pool(concurrency);
    try {
      await for (final item in this) {
        unawaited(pool.withResource(() async {
          await eachFn(item);
        }));
      }
    } finally {
      await pool.close();
    }
```

maybe actually want to do:

```
    Object? err;
    StackTrace? stackTrace;
    final pool = Pool(concurrency);
    try {
      await for (final item in this) {
        unawaited(pool.withResource(() async {
          try {
            await eachFn(item);
          } catch (e, st) {
            // Take the first error
            if (err != null) {
              err = e;
              stackTrace = st;
            }
          }
        }));
      }
    } finally {
      await pool.close();
      if (err != null) {
        rethrow err, st; // if something like this is possible.
      }
    }
```

app/lib/shared/utils.dart
```diff
@@ -361,3 +362,38 @@ bool fixedTimeEquals(String a, String b) {
...
+    final pool = Pool(concurrency);
+    final futures = Queue<Future>();
+    // keeps the size of the futures limited, allows some difference in the run times
+    final maxQueueSize = concurrency * 2;
+    try {
+      await for (final item in this) {
+        final f = pool.withResource(() => eachFn(item));
+        futures.add(f);
+        while (futures.length > maxQueueSize) {
+          await futures.removeFirst();
```
jonasfj, line 381: Shouldn't this be a `Future.race` ?

app/lib/shared/utils.dart
```diff
@@ -361,3 +362,38 @@ bool fixedTimeEquals(String a, String b) {
...
+        }
+      }
+      await Future.wait(futures);
+    } finally {
+      await pool.close();
+    }
+  }
```
jonasfj, line 388:
```suggestion
    Future<void> boundedForEach(int concurrency, FutureOr<void> Function(T) each) async {
    try {
      var N = 0;
      var Object? err;
      var StackTrace? stackTrace;
      var blocked = Completer()..complete();
      await for (final item in this) {
        scheduleMicroTask(() async {
          try {
            await each(item);
          } catch (e, st) {
            if (err == null) {
              (err, stackTrace) = (e, st);
            }
          } finally {
            N--;
            if (!blocked.isCompleted) {
              blocked.complete();
            }
          }
        });
        N++;
        if (N >= concurrency) {
          blocked = Completer();
          await blocked.future;
        }
      }
    } finally {
      while (N > 0) {
        blocked = Completer();
        await blocked.future;
      }
    }
  }
```

isoos: I think this will schedule all of the async task upfront (e.g. for 1 million entries it may create 1 million microtask before the first completes). The benefit of the queue + blocking is that it won't have more than the bound condition at once (assuming that the original iteration pauses when the async processing is being blocked).

app/lib/shared/utils.dart
isoos: Not sure: what's that, in which library?

app/lib/shared/utils.dart
```diff
@@ -361,3 +362,38 @@ bool fixedTimeEquals(String a, String b) {
...
+    final maxQueueSize = concurrency * 2;
+    try {
+      await for (final item in this) {
+        final f = pool.withResource(() => eachFn(item));
+        futures.add(f);
+        while (futures.length > maxQueueSize) {
+          await futures.removeFirst();
+        }
+      }
+      await Future.wait(futures);
```
jonasfj, line 384: If the first future throws, then this will never run and any subsequent future that throws will cause an uncaught exception, right?

isoos:
```
Future<void> main() async {
  try {
    await Future.wait(List.generate(
        10,
        (i) => Future.microtask(() {
              throw Exception('$i');
            })));
  } catch (e) {
    print('caught $e');
  }
}
```
Will catch the first exception, no further exception is uncaught. The `Future.wait` documentation says:

> If any future completes with an error, then the returned future completes with that error. If further futures also complete with errors, those errors are discarded.

app/lib/shared/utils.dart
isoos: I'm not sure if that's what we want:

> The returned future is completed with the result of the first future in futures to report that it is complete, whether it's with a value or an error. The results of all the other futures are discarded.

app/lib/shared/utils.dart
jonasfj: I think `discarded` means that errors are not unhandled.

isoos: This is correct. Therefore, it is safe to use `Future.wait`, or am I reading something wrong?

jonasfj: yes, but `await futures.removeFirst()` is not safe, and `Future.wait` might never be called.

isoos: Updated to always use `Future.wait` with a simple condition. This is suboptimal for the concurrency, but maintaining exactly the given number of concurrently running functions requires more complex code, and we don't really need that level of optimization yet.

app/lib/shared/utils.dart
```diff
@@ -361,3 +362,39 @@ bool fixedTimeEquals(String a, String b) {
...
+    final futures = <Future>[];
+    try {
+      await for (final item in this) {
+        final f = pool.withResource(() => eachFn(item));
+        futures.add(f);
+        if (futures.length >= concurrency) {
+          await Future.wait(futures);
+          futures.clear();
+        }
+      }
```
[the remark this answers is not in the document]
jonasfj: We could also use `Future.any()` and then find a way to remove the completed future from the list of `futures`.

But this is much harder and more expensive, there is really no reason to keep a list of all `futures`, that's just a lot of unnecessary allocations :D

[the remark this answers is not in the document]
isoos: Given the alternative, I find the `Future.any` + removing the completed Future from the list more appealing. I think the overhead of an update in a small list is probably very little compared to the async processing we are doing, I'd rather keep the for-loop simple. Let's explore that first before going in for the semaphore version.

app/lib/shared/utils.dart
```diff
@@ -361,3 +362,40 @@ bool fixedTimeEquals(String a, String b) {
...
+    final pool = Pool(concurrency);
+    final futures = <Future>[];
+    try {
+      await for (final item in this) {
+        final f = pool.withResource(() => eachFn(item));
+        futures.add(f);
+        while (futures.length > concurrency) {
+          await Future.any(futures.map(
+            (e) => e.whenComplete(() => futures.remove(e)),
+          ));
```
jonasfj, line 381: You end up doing this `.whenComplete` a LOT of times.

Since you'll do this `O(n^2)` times, is that really necessary?

We'll remove each future from the list `O(n^2)` times. Given that list is `n` long and the time to remove an item from the list is probably `O(n)`, I'm guessing this will be something like `O(n^3)`.

app/lib/shared/utils.dart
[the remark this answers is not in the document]
jonasfj: We could make this nicer using a `Condition` like:
```dart
/// A [Condition] allows micro-tasks to [wait] for other micro-tasks to
/// [notify].
///
/// [Condition] is a concurrency primitive that allows one micro-task to
/// wait for notification from another micro-task. The [Future] return from
/// [wait] will be completed the next time [notify] is called.
final class Condition {
  var _completer = Completer<void>();

  /// Complete all futures previously returned by [wait].
  ///
  /// Calls to [wait] after this call, will not be resolved, until the next time
  /// [notify] is called.
  void notify() {
    if (!_completer.isCompleted) {
      _completer.complete();
    }
  }

  /// Returns a [Future] that will complete the next time [notify] is called.
  ///
  /// This will always return an unresolved [Future]. Once [notify] is called
  /// the future will be completed, and any new calls to [wait] will return a
  /// new future. This future will also be unresolved, until [notify] is called.
  ///
  /// The [Future] return from this condition will never throw.
  Future<void> get wait {
    if (_completer.isCompleted) {
      _completer = Completer();
    }
    return _completer.future;
  }
}

```

isoos:
I think we could close this in favor of https://github.com/dart-lang/async/pull/249

jonasfj:
@isoos yes, but feel free to just copy paste from https://github.com/dart-lang/async/pull/249, it might be a while before it lands :D

PR #6766, opened 2023-06-16
Separate scorecard cache + 2h TTL

isoos:


app/lib/scorecard/backend.dart
```diff
@@ -124,9 +124,11 @@ class ScoreCardBackend {
         return null;
       }
     }
-    final cacheEntry = onlyCurrent || showSandboxedOutput
+    final cacheEntry = onlyCurrent
         ? null
-        : cache.scoreCardData(packageName, packageVersion);
+        : (showSandboxedOutput
+            ? cache.scoreCardData2(packageName, packageVersion)
```
jonasfj, line 130: I think we need to also store something in `scoreCardData2`.

Also we probably need purge from within the task backend.

isoos: The store is automatic in this method.
I've added calls for the purge.

jonasfj: > The store is automatic in this method.

The `    if (showSandboxedOutput) {` will return directly, without storing something in the cache.

isoos: fixed, PTAL

jonasfj (approved):

app/lib/scorecard/backend.dart
```diff
@@ -166,6 +168,11 @@ class ScoreCardBackend {
         ),
         panaReport: PanaReport.fromSummary(summary, packageStatus: status),
       );
+      // only full cards will be stored in cache
+      if (cacheEntry != null && data.isCurrent && data.hasAllReports) {
```
jonasfj, line 172: We purge when updating, so why not cache incomplete stuff?

PR #6770, opened 2023-06-19
Cheaper transaction in TaskBackend.

isoos:
- the most important part is to use `packageBackend.listVersionsCached` instead of querying the `PackageVersion`s inside the transaction
- also moving out resource loading outside of it, I think it won't really add much difference

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -3,6 +3,7 @@ import 'dart:convert';
 import 'dart:io';
 import 'dart:typed_data';
 
+import 'package:_pub_shared/data/package_api.dart';
 import 'package:_pub_shared/data/task_api.dart' as api;
 import 'package:chunked_stream/chunked_stream.dart'
     show readByteStream, MaximumSizeExceeded;
@@ -18,6 +19,7 @@ import 'package:indexed_blob/indexed_blob.dart' show BlobIndex, FileRange;
 import 'package:logging/logging.dart' show Logger;
 import 'package:pana/models.dart' show Summary;
 import 'package:pool/pool.dart' show Pool;
+import 'package:pub_dev/package/backend.dart';
 import 'package:pub_dev/package/models.dart';
 import 'package:pub_dev/package/upload_signer_service.dart';
 import 'package:pub_dev/scorecard/backend.dart';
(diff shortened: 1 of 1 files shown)


app/lib/task/backend.dart
```diff
@@ -328,26 +330,22 @@ class TaskBackend {
     String packageName, {
     bool updateDependants = false,
   }) async {
-    var lastVersionCreated = initialTimestamp;
-    await withRetryTransaction(_db, (tx) async {
-      final pkgKey = _db.emptyKey.append(Package, id: packageName);
+    final package = await packageBackend.lookupPackage(packageName);
```
jonasfj, line 333: I think the idea with `trackPackage` is that it will be called as soon as we detect changes to a package... loading from cache will sort of defeat the purpose won't it?

I do wonder if we call this frequently? Or do we only call it once per publishing event. Checkout `_scanForPackageUpdates` it tracks which events we've seen, right?

app/lib/task/backend.dart
```diff
@@ -328,26 +330,22 @@ class TaskBackend {
...
+    if (package == null) {
+      return; // assume package was deleted!
+    }
+    final lastVersionCreated = package.lastVersionPublished ?? initialTimestamp;
+    final packageVersions =
+        await packageBackend.listVersionsCached(packageName);
```
jonasfj, line 339: This defeats the purpose of detecting changes :D

Sure cache, should be up to date, but it might not be.
That's fine when serving API or website, because those only need to eventually be consistent. If we use cache here, we won't eventually be consistent, until a weekly background task runs `backfillTrackingState` -- I think that's a long time to wait.

If the tracking logic crashes for a while, then it's nice that `backfillTrackingState` will eventually ensure consistency.

jonasfj (commented):
I think not using a transaction in `trackPackage` will cause problems.

app/lib/task/backend.dart
isoos: I think version list queries are expensive. Why not use more frequent cheaper status checks than do the costly queries every time? The background task could run daily or like every 8 hours...

app/lib/task/backend.dart
isoos: Note: we could also cross-reference this last published timestamp with the one on the `Package` entity, discarding the cache if they differ.

app/lib/task/backend.dart
jonasfj: The cheap thing will usually not return the version that was just published.

`trackPackage` is usually called when we've detected that a package has changed. This is exactly the time at which using the cache will give an outdated result.

Look at `_scanForPackageUpdates` it tracks updates, and only calls `trackPackage` for every change.

app/lib/task/backend.dart
```diff
@@ -971,8 +972,9 @@ List<Version> _versionsToTrack(
     // Consider 5 latest major versions, if any:
     ...packageVersions
         // Ignore prereleases and retracted versions
-        .where((pv) => !pv.isRetracted && !pv.semanticVersion.isPreRelease)
-        .map((pv) => pv.semanticVersion)
+        .where((pv) => !(pv.retracted ?? false))
+        .map((pv) => Version.parse(pv.version))
```
jonasfj, line 976: Why use `pv.version` over `pv.semanticVersion` ? I probably don't understand the difference.

app/lib/task/backend.dart
```diff
@@ -971,8 +972,9 @@ List<Version> _versionsToTrack(
...
+        .where((v) => !v.isPreRelease)
```
jonasfj, line 977: I'd love this fix, can we split it out? But curious why not use `pv.isRetracted`?

app/lib/task/backend.dart
```diff
@@ -611,6 +604,8 @@ class TaskBackend {
 
     String? zone, instance;
     bool isInstanceDone = false;
+    final summary = await panaSummary(package, version);
+    final dartdocIndex = await dartdocFile(package, version, 'index.html');
```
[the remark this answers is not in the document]
jonasfj: True, we could also write a custom method for fetching both, that would slightly cheaper... especially we only use the dartdoc file to check for existence.

But that seems like a follow up. Keeping it simple if fine!

app/lib/task/backend.dart
jonasfj: > The cache, esp. this version list cache should be purged

"should be"

-----------

> I think we may call it more than once per publishing event,

I'd be curious to see how?

app/lib/task/backend.dart
isoos: We are switching base classess here: `PackageVersion` did have `isRetracted`, the same field is `VersionInfo.retracted` and is nullable. Similarly: `PackageVersion` had `semanticVersion`, `VersionInfo` does only `version`.

app/lib/task/backend.dart
isoos: Yeah, I'll do a split out PR.

PR #6774, opened 2023-06-20
Generate dartdoc content archive in pub_worker (package.tar.gz).

isoos:


pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
```diff
@@ -180,6 +181,35 @@ Future<void> _dartdoc({
   // This helps if there is a timeout along the way.
   await Directory(tmpOutDir).rename(p.join(outputFolder, 'doc'));
   _log.info('Finished post-processing');
+
+  _log.info('Creating .tar.gz archive');
+  Stream<TarEntry> _list() async* {
+    final finalDocDir = Directory(p.join(outputFolder, 'doc'));
```
jonasfj, line 187: This doesn't contain html files... it contains `.html` files that actually are JSON files.

`Directory(docDir)` is more like it. See post-processing above.

isoos: Ah, good catch, fixed.

jonasfj (approved):

PR #6790, opened 2023-06-26
Expand the fake JWT token timestamp range.

isoos:
There is a slight chance that the strict validity window may cause test flakes on CI, there is not much drawback to extend it a bit.


sigurdm:
Seems there is still a failure...

isoos:
> Seems there is still a failure...

Yeah :(
Merging anyway, as it seems to eliminate one class of errors (or we had just different flakes consistently).

PR #6795, opened 2023-06-30
Refactored isolate runner

isoos:
- (hopefully) not much functional change, but log messages may be changing slightly
- refactored message handling into a separate class
- unified entry and ready messages for different kind of isolates
- small side effect of the unification: worker isolates have the same restart delay as the frontend ones (less initial delay the increases over time)
- tests for regular and error-based exits and restarts
- a further goal of this is to prepare the search service for starting a new isolate for each new snapshot in a controlled way (TBD in a subsequent PR)

### DIFF
--- a/app/lib/service/entrypoint/_isolate.dart
+++ b/app/lib/service/entrypoint/_isolate.dart
@@ -21,267 +21,144 @@ import 'tools.dart';
 
 final _random = Random.secure();
 
-class FrontendEntryMessage {
-  final int frontendIndex;
+class EntryMessage {
   final SendPort protocolSendPort;
   final SendPort aliveSendPort;
+  final SendPort statsSendPort;
 
-  FrontendEntryMessage({
-    required this.frontendIndex,
+  EntryMessage({
     required this.protocolSendPort,
     required this.aliveSendPort,
+    required this.statsSendPort,
   });
 }
 
-class FrontendProtocolMessage {
+/// Message sent from the isolate to indicate that it is ready with the initialization.
+class ReadyMessage {
   final SendPort? statsConsumerPort;
 
-  FrontendProtocolMessage({
+  ReadyMessage({
     this.statsConsumerPort,
   });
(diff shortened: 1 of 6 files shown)


app/lib/service/entrypoint/_isolate.dart
```diff
@@ -21,329 +21,342 @@ import 'tools.dart';
 
 final _random = Random.secure();
 
-class FrontendEntryMessage {
-  final int frontendIndex;
+class EntryMessage {
```
jonasfj, line 24: documentation comment would be lovely

app/lib/service/entrypoint/_isolate.dart
```diff
@@ -21,329 +21,342 @@ import 'tools.dart';
...
+  EntryMessage({
     required this.protocolSendPort,
     required this.aliveSendPort,
+    required this.statsSendPort,
   });
 }
 
-class FrontendProtocolMessage {
+/// Message sent from the isolate to indicate that it is ready with the initialization.
+class ReadyMessage {
```
jonasfj, line 37: You could do:
```dart
sealed class Message {}
final ReadyMessage extends Message {
  final String stuff;
  ReadyMessage(this.stuff);
}
final EntryMessage extends Message {}
final DebugMessage extends Message {}
```

then you can do:
```dart
_protocolReceivePort.listen((e) {
 if (e is! Message) {
   return;
 }
 switch(e) {
   case ReadyMessage(stuff: var s):
     print(s);
   case EntryMessage():
   case DebugMessage():
 }

});
```

No need though, just promoting new Dart 3 features :rofl:

jonasfj (approved):

app/lib/service/entrypoint/_isolate.dart
isoos: I think this is nice, possibly in a follow-up after this PR proved that it works as expected :)

PR #6800, opened 2023-07-04
Load sessions on non-primary host too.

isoos:
I think this may be a blocker for post-deployment integration tests.


jonasfj (approved):

app/lib/frontend/request_context.dart
```diff
@@ -80,7 +80,7 @@ Future<RequestContext> buildRequestContext({
   // Parse client session cookie status, which can be present at any kind of request.
   SessionData? sessionData;
   final clientSessionCookieStatus = parseClientSessionCookies(cookies);
-  if (isPrimaryHost && clientSessionCookieStatus.isPresent) {
```
jonasfj, line 83: Actually, isn't this here because there are multiple domains?

PR #6805, opened 2023-07-04
Use cached versions list instead of querying.

isoos:
- This is the larger part of #6770.

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -3,6 +3,7 @@ import 'dart:convert';
 import 'dart:io';
 import 'dart:typed_data';
 
+import 'package:_pub_shared/data/package_api.dart';
 import 'package:_pub_shared/data/task_api.dart' as api;
 import 'package:chunked_stream/chunked_stream.dart'
     show readByteStream, MaximumSizeExceeded;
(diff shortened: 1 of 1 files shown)


app/lib/task/backend.dart
```diff
@@ -329,26 +331,30 @@ class TaskBackend {
...
+    if (lastCachedPublished == null ||
+        lastCachedPublished.isBefore(lastVersionCreated)) {
+      await purgePackageCache(packageName);
+      packageVersions = await packageBackend.listVersionsCached(packageName);
+    }
+    // Determined the set of versions to track
+    final trackedVersions =
+        _versionsToTrack(package, packageVersions.versions).map(
+      (v) => v.canonicalizedVersion, // add extra sanity!
+    );
```
jonasfj, line 353: If I'm in a race condition against another process running or publishing `trackPackage`, then I kind of have to reload the facts I'm using create my mutation inside `withRetryTransaction`.

If `withRetryTransaction` is retried due a conflicting mutation, then how do I know that the data I loaded before my transaction is still valid?


---------------------

Maybe, if `trackPackage` is called from `backfillTrackingState`, we could use cache values to check if mutations are necessary.
But:
 * Only when called from `backfillTrackingState`,
 * Only using cache value to conclude that an update isn't necessary.

We rarely have to do actual updates. I think consistency issues are a bit more concerning than extra compute spent once per week.

app/lib/task/backend.dart
```diff
@@ -329,26 +331,30 @@ class TaskBackend {
     String packageName, {
     bool updateDependants = false,
   }) async {
-    var lastVersionCreated = initialTimestamp;
-    await withRetryTransaction(_db, (tx) async {
-      final pkgKey = _db.emptyKey.append(Package, id: packageName);
+    final package = await packageBackend.lookupPackage(packageName);
```
jonasfj, line 334: If we're willing to use a cached or potentially old value, then we can just take it as a parameter, instead of `packageName`.

I didn't do that because I wanted to reload the `Package` entity if the transaction has a conflict.

isoos:
Offline discussion concluded that we don't want to change this for now. If/when this is a problem, we can use the cached list + some kind of cache invalidation, but we are not hard-pressed for it.

PR #6811, opened 2023-07-06
Create empty blob index in cache when task index is missing.

isoos:


pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -143,13 +154,11 @@ class IndexedBlobBuilder {
...
     w.addKey('index');
-    _addMap(w, _index);
+    _addMap(w, index);
 
     w.endObject();
 
-    final bytes = utf8.encoder.convert(out.toString());
-
-    return BlobIndex.fromBytes(bytes);
+    return utf8.encoder.convert(out.toString());
```
jonasfj, line 161:
```suggestion
    return utf8.encode(out.toString());
```

isoos: Keeping the original code as it gets a proper Uint8List, while the other one needs casting.

PR #6817, opened 2023-07-06
Check (cached) package status before attempting to load version blob index.

isoos:
- Built on top of #6816.
- using the name `forRuntimeVersion` instead of `runtimeVersion`, because otherwise we would need to prefix the imported library
- purging the cache entries with the older `runtimeVersion` values seems to be something we could do, but doesn't really need to, leaving it out for now

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -38,6 +38,7 @@ import 'package:pub_dev/task/models.dart'
         PackageState,
         PackageStateInfo,
         PackageVersionStateInfo,
+        PackageVersionStatus,
         initialTimestamp,
         maxTaskExecutionTime;
 import 'package:pub_dev/task/scheduler.dart';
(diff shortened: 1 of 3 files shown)


app/lib/task/backend.dart
```diff
@@ -738,28 +739,33 @@ class TaskBackend {
   Future<BlobIndex?> _taskResultIndex(String package, String version) async =>
       await cache.taskResultIndex(package, version).get(() async {
         // Try runtimeVersions in order of age
-        var path = '$runtimeVersion/$package/$version/index.json';
-        List<int>? bytes;
         for (final rt in acceptedRuntimeVersions) {
```
jonasfj, line 742: Hmm,

We could also avoid caching packageStatus per runtimeVersion, instead we could choose to only load the one for the current runtimeVersion.

If the version is not going to be analyzed by the current runtimeVersion, we should never load an analysis done by an old runtimeVersion, because that analysis will not eventually become consistent. That analysis is permanently stale.

So maybe we should just check the `packageState` of the current package, before we do:
`for (final rt in acceptedRuntimeVersions) {`

That'll probably shortcut must of the lookups, all of the rest of the lookups should eventually be able to find a result from the current runtimeVersion.

isoos:
This became obsolete

PR #6818, opened 2023-07-07
Don't try to load index if we don't consider the version for analysis.

isoos:


app/lib/task/backend.dart
```diff
@@ -737,8 +738,20 @@ class TaskBackend {
...
+        }
         // Try runtimeVersions in order of age
         for (final rt in acceptedRuntimeVersions) {
+          // skip current runtime if it is a know failure or hasn't been started yet.
+          if (rt == runtimeVersion) {
+            final versionStatus = status.versions[version]!.status;
+            if (versionStatus == PackageVersionStatus.failed) {
+              continue;
+            }
+          }
```
jonasfj, line 754: I'd suggest moving this outside the loop.

If the current runtimeVersion failed, then failure is intentional. We should only fallback when we haven't processed the current version, because it's pending.

app/lib/task/backend.dart
```diff
@@ -737,8 +738,24 @@ class TaskBackend {
   /// package name, version and randomized blobId.
   Future<BlobIndex?> _taskResultIndex(String package, String version) async =>
       await cache.taskResultIndex(package, version).get(() async {
+        // Don't try to load index if we don't consider the version for analysis.
+        final status = await packageStatus(package);
+        if (!status.versions.containsKey(version)) {
+          return BlobIndex.empty(blobId: '');
+        }
```
jonasfj, line 745: nit:
```suggestion
        if (!status.versions.containsKey(version) ||
            status.versions[version] == PackageVersionStatus.failed) {
          return BlobIndex.empty(blobId: '');
        }
```

jonasfj (approved):

app/lib/task/backend.dart
jonasfj: Then you can remove one condition from the loop

PR #6847, opened 2023-07-21
Fake task analysis using current fake pana and dartdoc.

isoos:
- This is a requirement for #6834, submitting it separately.
- The largest change seems to be the retracted version template.

### DIFF
--- a/app/lib/fake/backend/fake_dartdoc_runner.dart
+++ b/app/lib/fake/backend/fake_dartdoc_runner.dart
@@ -52,26 +52,29 @@ class FakeDartdocRunner implements DartdocRunner {
     required bool useLongerTimeout,
     required String outputDir,
   }) async {
-    final random = Random('$package/$version'.hashCode);
-
-    // basic content, existence checked by the job processor
-    await File(p.join(outputDir, 'index.html'))
-        .writeAsString('<html><head></head><body>index.html</body></html>');
-    // JS search index, existence checked by the job processor
-    await File(p.join(outputDir, 'index.json')).writeAsString(json.encode({}));
-    // pub search data
-    final pubData = PubDartdocData(
-      coverage: Coverage(documented: random.nextInt(21), total: 20),
-      apiElements: [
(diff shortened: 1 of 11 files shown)


app/lib/fake/backend/fake_pub_worker.dart
```diff
@@ -0,0 +1,254 @@
...
+
+              Future<void> addFileAsStringGzipped(
+                  String path, String content) async {
+                final stream =
+                    Stream.fromIterable([gzip.encode(utf8.encode(content))]);
+                await builder.addFile(path, stream);
+              }
+
+              for (final e in dartdocFiles.entries) {
+                await addFileAsStringGzipped(e.key, e.value);
```
jonasfj, line 128: The `.html` files created by `pub_worker` are actually JSON files :rofl: 

This happened when we moved the postprocessing/sanitizing into the sandbox

app/lib/fake/backend/fake_dartdoc_runner.dart
```diff
@@ -52,26 +52,29 @@ class FakeDartdocRunner implements DartdocRunner {
...
+    }
 
     return DartdocRunnerResult(
       args: ['fake_dartdoc', '--input', pkgPath, '--output', outputDir],
       processResult: PanaProcessResult(0, 0, 'OK', ''),
     );
   }
 }
+
+Map<String, String> fakeDartdocFiles(String package, version) {
```
jonasfj, line 67: wouldn't it be easier to copy / paste this, than have cross dependencies...

app/lib/fake/backend/fake_pub_worker.dart
jonasfj: We can probably just hardcode some JSON that fits the DartDocPage pattern

isoos: Updated. PTAL.

jonasfj (approved):

PR #6849, opened 2023-07-24
Upgrade dependencies (incl. appengine and markdown).

isoos:


jonasfj (approved):

isoos:
Removed upgrades in pkg/web_css, because sass seems to be failing.

PR #6850, opened 2023-07-24
Move retries into package:appengine

jonasfj:


isoos (approved):

isoos:
Let's merge this first (needs a fix in the import test), updated #6849 to match the same format to reduce conflict.

PR #6863, opened 2023-07-27
Explicitly access uncached data when finalizing task upload.

isoos:
Fixes the cache issue introduced by #6804

### DIFF
--- a/app/lib/shared/redis_cache.dart
+++ b/app/lib/shared/redis_cache.dart
@@ -534,9 +534,9 @@ extension EntryPurgeExt<T> on Entry<T> {
       return value;
     }
     if (ttl != null) {
-      return (await get(create, ttl))!;
+      return await get(create, ttl);
     }
-    return (await get(create))!;
+    return await get(create);
   }
 }
 
--- a/app/lib/shared/redis_cache.dart
+++ b/app/lib/shared/redis_cache.dart
@@ -508,7 +508,7 @@ class _ConnectionRefreshingCacheProvider<T> implements CacheProvider<T> {
       .timeout(_defaultCacheWriteTimeout, onTimeout: () => null);
 }
 
-extension EntryPurgeExt on Entry {
+extension EntryPurgeExt<T> on Entry<T> {
   /// Datastore.query is only eventually consistent, and when we are changing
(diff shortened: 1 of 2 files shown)


app/lib/task/backend.dart
```diff
@@ -741,56 +750,75 @@ class TaskBackend {
...
-      await cache.taskResultIndex(package, version).get(() async {
-        // Don't try to load index if we don't consider the version for analysis.
-        final status = await packageStatus(package);
-        if (!status.versions.containsKey(version)) {
-          return BlobIndex.empty(blobId: '');
+  Future<BlobIndex?> _taskResultIndex(
+    String package,
+    String version, {
+    bool skipCached = false,
+  }) async {
```
jonasfj, line 757:
```dart
extension NeatCacheEntry<T> on Entry<T> {
  // Get or create a value.
  Future<T> obtain(
    Future<T> Function() create, {
    Duration? ttl,
    bool purgeCache = false,
  }) async {
    if (purgeCache) {
      final value = create();
      if (ttl != null) {
        await set(value, ttl);
      } else {
        await set(value);
      }
      return value;
    }
    if (ttl != null) {
      return (await get(create, ttl))!;
    }
    return (await get(create))!;
  }
}
```

Filed https://github.com/google/dart-neats/issues/214 suggesting we add this upstream.

jonasfj (approved):
Maybe, we should do the extension method just keep things clean.

If I add the method upstream later, we'll get an easy issue to fix when we upgrade, so that's not a huge problem.

app/lib/shared/redis_cache.dart
```diff
@@ -517,6 +517,27 @@ extension EntryPurgeExt on Entry {
...
+        await set(value, ttl);
+      } else {
+        await set(value);
+      }
+      return value;
+    }
+    if (ttl != null) {
+      return (await get(create, ttl))!;
+    }
+    return (await get(create))!;
```
jonasfj, line 539:
```suggestion
      return await get(create, ttl);
    }
    return await get(create);
```

jonasfj (approved):

Nishapar (approved):

PR #6867, opened 2023-07-27
Allow long word breaks by default in the base CSS.

### LINKED ISSUE #6861: Text over expanding when using a long description of a publisher
URL: https://pub.dev/publishers/rapougnac.moe/packages

I was playing around with the description of my publisher, and added a long text, and it seems there's no line break of word wrap to prevent the text to go out of the page, which leads to a long horizontal white page.

image

I tested it on FF and Chrome.

isoos:
Fixes #6861.


isoos:
Note: deploying this to staging for further verification.

isoos:
Staging looks fine with this change.

PR #6872, opened 2023-07-28
Controlled renew of an isolate within a group.

isoos:
- This will be used by the search service.
- Renewing first creates a new isolate, and then closes the older ones. Right now it only wait for a specific duration, later on it may wait for the pending requests to complete before closing the older isolates.

### DIFF
--- a/app/lib/service/entrypoint/_isolate.dart
+++ b/app/lib/service/entrypoint/_isolate.dart
@@ -50,13 +50,14 @@ class DebugMessage extends Message {
 /// isolate execute the same code).
 ///
 /// TODO: The runner will handle the cross-group communication of the isolates.
-class IsolateRunner {
+@visibleForTesting
+class IsolateCollection {
   final Logger logger;
   var _closing = false;
 
   final _groups = <IsolateGroup>[];
 
-  IsolateRunner({
+  IsolateCollection({
     required this.logger,
   });
 
@@ -99,7 +100,7 @@ class IsolateRunner {
 /// Once an isolate starts, it is expected to run indefinitely. When it exits,
 /// either by completion or uncaught exception, a new isolate will be started.
 class IsolateGroup {
-  final IsolateRunner runner;
+  final IsolateCollection runner;
   final String kind;
   final Future<void> Function(EntryMessage message) entryPoint;
   final Duration? deadTimeout;
@@ -202,7 +203,7 @@ class IsolateGroup {
   }
 }
 
-/// Starts an [IsolateRunner] with the default isolate configuration
+/// Starts an [IsolateCollection] with the default isolate configuration
 /// (when specified).
 ///
 /// After starting the isolates, the method waits for terminating
@@ -219,7 +220,7 @@ Future runIsolates({
   await withServices(() async {
     _verifyStampFile();
     try {
-      final runner = IsolateRunner(logger: logger);
+      final runner = IsolateCollection(logger: logger);
       if (frontendEntryPoint != null) {
         await runner.startGroup(
           kind: 'frontend',
@@ -274,7 +275,7 @@ void _verifyStampFile() {
 /// autokill timer.
 class _Isolate {
   /// Parent runner that owns this group
-  final IsolateRunner parent;
+  final IsolateCollection parent;
   final IsolateGroup group;
   final Logger logger;
   final String id;
--- a/app/lib/service/entrypoint/_isolate.dart
+++ b/app/lib/service/entrypoint/_isolate.dart
@@ -49,76 +49,145 @@ class IsolateRunner {
   final Logger logger;
   var _closing = false;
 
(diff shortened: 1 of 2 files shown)


app/lib/service/entrypoint/_isolate.dart
```diff
@@ -45,83 +46,168 @@ class DebugMessage extends Message {
...
+  Future<void> start(int count) async {
+    for (var i = 0; i < count; i++) {
+      await _startOne();
+    }
+  }
+
+  /// Starts [count] new isolates, and after a [wait] duration,
+  /// closes the old ones.
+  Future<void> renew({
+    required int count,
```
jonasfj, line 135: Would this count not always be the same ?

app/lib/service/entrypoint/_isolate.dart
```diff
@@ -45,83 +46,168 @@ class DebugMessage extends Message {
...
+    required Duration wait,
```
jonasfj, line 136: Wouldn't it be safer to wait to tell the isolates to shutdown when they are not handling any more search messages...
and then have a timeout for sanity, so that if the timeout expires we hard kill.
For this wouldn't need a configurable timeout, just a hardcoded sanity value and a shout into logs.

app/lib/service/entrypoint/_isolate.dart
```diff
@@ -45,83 +46,168 @@ class DebugMessage extends Message {
   DebugMessage(this.text);
 }
 
+/// Runs the collection of different isolate groups (where a group of
+/// isolate execute the same code).
+///
+/// TODO: The runner will handle the cross-group communication of the isolates.
 class IsolateRunner {
```
jonasfj, line 53: Is this perhaps an `IsolateCollection` instead?

app/lib/service/entrypoint/_isolate.dart
```diff
@@ -45,83 +46,168 @@ class DebugMessage extends Message {
   DebugMessage(this.text);
 }
 
+/// Runs the collection of different isolate groups (where a group of
+/// isolate execute the same code).
+///
+/// TODO: The runner will handle the cross-group communication of the isolates.
```
jonasfj, line 52:
```suggestion
/// TODO: The runner will handle the cross-group communication of the isolates.
@visibleForTesting
```

jonasfj (approved):

app/lib/service/entrypoint/_isolate.dart
```diff
@@ -45,83 +46,168 @@ class DebugMessage extends Message {
...
+  IsolateGroup({
+    required this.runner,
+    required this.kind,
+    required this.entryPoint,
+    required this.deadTimeout,
+    this.skipWaitBetweenRestarts = false,
+  });
+
+  /// Starts [count] new isolates.
+  Future<void> start(int count) async {
```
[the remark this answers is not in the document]
isoos: `renew` also calls this method

app/lib/service/entrypoint/_isolate.dart
```diff
@@ -45,83 +46,168 @@ class DebugMessage extends Message {
...
+  }
+}
+
+/// Starts, monitors, stops or restarts isolates that run the same code.
+///
+/// Once an isolate starts, it is expected to run indefinitely. When it exits,
+/// either by completion or uncaught exception, a new isolate will be started.
+class IsolateGroup {
+  final IsolateRunner runner;
+  final String kind;
```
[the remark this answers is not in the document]
isoos: I thought that the search frontend -> index isolate could identify the target group through this `kind` property.

app/lib/service/entrypoint/_isolate.dart
```diff
@@ -45,83 +46,168 @@ class DebugMessage extends Message {
...
+    required this.kind,
+    required this.entryPoint,
+    required this.deadTimeout,
+    this.skipWaitBetweenRestarts = false,
+  });
+
+  /// Starts [count] new isolates.
+  Future<void> start(int count) async {
+    for (var i = 0; i < count; i++) {
+      await _startOne();
```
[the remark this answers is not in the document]
isoos: I think it is better to spread it out mostly for caching: the first isolate's startup could populate the cache, so the subsequent ones only need to access only redis, and not query Datastore too.

app/lib/service/entrypoint/_isolate.dart
isoos: In the search case it would be always 1.

app/lib/service/entrypoint/_isolate.dart
isoos: Yeah, that's probably better to do. I though that we can skip the complexity by giving it enough time for surely completing the requests (and the rest should already timed out by the calling HTTP handlers).

PR #6881, opened 2023-08-04
Calculate sandboxed PackageView.isPendig from task status.

### LINKED ISSUE #6878: "Pending analysis" status shows "Unknown Platforms" instead
I remember the tag should displayed as `PENDING ANALYSIS`, but we got `UNKNOWN PLATFORMS` instead.

image

isoos:
Fixes #6878.


jonasfj (approved):
We'll need to bump runtimeVersion, if we haven't already, other caches will have issues -- as we've extended the model.

PR #6884, opened 2023-08-07
Add unit tests for osv validation

szakarias:


jonasfj (approved):

app/lib/service/security_advisories/backend.dart
```diff
@@ -44,8 +45,8 @@ class SecurityAdvisoryBackend {
     return _db.lookupOrNull<SecurityAdvisory>(key);
   }
 
-  // TODO(zarah): add unit test for this.
-  List<String> _validateAdvisoryErrors(OSV osv) {
+  @visibleForTesting
+  List<String> validateAdvisoryErrors(OSV osv) {
```
jonasfj, line 49: this could just be a static function, then you won't need `testWithProfile` but can just use `test(` -- it's much faster and simpler.

szakarias: Made it top-level.

PR #6893, opened 2023-08-09
Admin API to trigger task processing.

isoos:
Also removes search index trigger, as we will soon not use updates like that.

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -1000,6 +1000,23 @@ class TaskBackend {
       await processPayload(payload);
     }
   }
+
+  /// Trigger a one-off priority bump for [packageName].
+  ///
+  /// Intended to be used for admin actions, not intended for normal operations.
+  Future<void> adminBumpPriority(String packageName) async {
+    // Ensure we're up-to-date.
+    await trackPackage(packageName);
+
+    await withRetryTransaction(_db, (tx) async {
+      final stateKey = PackageState.createKey(_db, runtimeVersion, packageName);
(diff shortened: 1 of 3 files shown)


app/lib/task/backend.dart
```diff
@@ -427,7 +428,7 @@ class TaskBackend {
               attempts: 0,
             ),
         });
-      state.derivePendingAt();
+      state.derivePendingAt(isHighPriority: isHighPriority);
```
jonasfj, line 431:
```suggestion
      state.pendingAt = initialTimestamp;
```

jonasfj: We have an initial value, and there is a comment  for why it is so.

I'm not sure we ever set `pendingAt` to it anywhere.. so it might be fine to do this.

isoos: Updated it inside the `derivePendingAt`. Maybe it is not the best place for it, but then we'd need to `if (highPrio) {setToZero()} else {derivePendingAt()}` for the two places in this pr, and I think it has better read-flow this way.

jonasfj: I still don't believe this will work.

```dart
      // Stop transaction, if there is no changes to be made!
      if (untrackedVersions.isEmpty && deselectedVersions.isEmpty) {
        return false;
      }
```

will probably cause problems...

Why not just: call `trackPackage(packageName)` then load `PackageState` and set `pendingAt = initialTimestamp`.
It's a bit cleaner than moving logic into a function that's now going to be half broken, because we probably can't keep track of all the concerns it has to deal with.

jonasfj (approved):

PR #6896, opened 2023-08-09
updatePublicArchiveBucket should handle missing objects

jonasfj:


isoos (approved):

app/lib/tool/maintenance/update_public_bucket.dart
```diff
@@ -46,18 +46,27 @@ Future<PublicBucketUpdateStat> updatePublicArchiveBucket({
 
   final objectNamesInPublicBucket = <String>{};
 
+  var errCount = 0;
```
isoos, line 49: this is not used.

isoos (approved):

PR #6902, opened 2023-08-11
Search index in an isolate.

isoos:
- Using `spawnUri` for the index isolate, to reduce memory issues.
- Most messages are sent encoded as JSON, but if they are known to go inside the same heap, they may be sent as objects to keep efficiency.
- Tracks pending request-reply pairs, and isolate renewal waits for pending messages to complete (up to a threshold, where isolate will be killed regardless of the pending messages).
- New behavior: once the `SearchIndex.isReady()` signal turns true, it keeps at that, the index is no longer queried.

### DIFF
--- a/app/lib/service/entrypoint/_isolate.dart
+++ b/app/lib/service/entrypoint/_isolate.dart
@@ -405,7 +405,7 @@ class _Isolate {
   }) async {
     final ready = Completer();
     _protocolSubscription = _protocolReceivePort.listen((event) {
-      final e = Message.decode(event);
+      final e = Message.fromObject(event);
       if (e is ReadyMessage && !ready.isCompleted) {
         _readyMessage = e;
         ready.complete();
--- a/app/lib/service/entrypoint/_isolate.dart
+++ b/app/lib/service/entrypoint/_isolate.dart
@@ -68,10 +68,11 @@ class IsolateCollection {
       kind: kind,
       entryPoint: entryPoint,
       spawnUri: spawnUri,
+      count: count,
       deadTimeout: deadTimeout,
     );
     _groups.add(group);
-    await group.start(count);
+    await group.start();
     return group;
   }
 
@@ -96,6 +97,7 @@ class IsolateGroup {
   final String kind;
   final EntryPointFn? entryPoint;
   final Uri? spawnUri;
+  final int count;
   final Duration? deadTimeout;
   final bool skipWaitBetweenRestarts;
 
@@ -113,12 +115,13 @@ class IsolateGroup {
     required this.kind,
     required this.entryPoint,
     required this.spawnUri,
+    required this.count,
     required this.deadTimeout,
     this.skipWaitBetweenRestarts = false,
   });
 
   /// Starts [count] new isolates.
-  Future<void> start(int count) async {
+  Future<void> start() async {
     for (var i = 0; i < count; i++) {
       await _startOne();
     }
@@ -127,7 +130,6 @@ class IsolateGroup {
   /// Starts [count] new isolates, waits for the pending requests to get processed,
   /// and after a maximum [wait] duration, closes the old ones.
   Future<void> renew({
-    required int count,
     required Duration wait,
   }) async {
     final isolatesToClose = [..._isolates];
@@ -136,7 +138,7 @@ class IsolateGroup {
       i.markedForReplace = true;
     }
     // start new isolates
-    await start(count);
+    await start();
 
     await Future.delayed(wait);
 
@@ -292,8 +294,7 @@ Future runIsolates({
           deadTimeout: null,
         );
         indexRenewSubscription = indexRenewTrigger?.listen((_) {
-          indexGroup.renew(
-              count: 1, wait: indexRenewTimeout ?? Duration(minutes: 5));
+          indexGroup.renew(wait: indexRenewTimeout ?? Duration(minutes: 5));
         });
       }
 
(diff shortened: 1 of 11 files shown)


isoos:
Deploying the latest to staging...

app/lib/service/entrypoint/_messages.dart
```diff
@@ -0,0 +1,172 @@
...
+
+/// Message sent to a different isolate group for processing.
+class RequestMessage extends Message {
+  /// The `kind` of the target isolate group.
+  final String kind;
+
+  /// The payload (request) object.
+  final Object payload;
+
+  /// The port where we expect the respnse.
```
jonasfj, line 93:
```suggestion
  /// Port to which a [ReplyMessage] must be sent when the request is processed.
  ///
  /// The sender of the request may close the [replyPort], if the sender has decided
  /// timeout and not wait any further.
```

app/lib/service/entrypoint/_messages.dart
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/// Wraps the request/reply message sending logic.
+Future<Object?> handleRequestReply({
+  required SendPort requestSendPort,
+  required String kind,
+  required Object payload,
+  required Duration timeout,
+}) async {
+  final replyRecievePort = ReceivePort();
```
jonasfj, line 162: I think we need to close this in a `finally`, otherwise we're leaking ports.

I'm not sure how well GC'ing works on ports, it's probably one of things that it's helpful to close.

Especially, since you're not cancelling the listening operation. It's unlikely that the GC will ever be able to see that all instances of the [SendPort] have been GC'ed. I'm just guessing here. But these can be sent to FFI stuff, so our isolate will probably be unable to die naturally, it'll always sit around waiting for a message on the port, if a timeout happens.

isoos: Good point, let's do it explicitly, although, implicitly we are already doing it I think:

"A receive port is closed by canceling its subscription." + "Internally the method cancels its subscription after the first element." (`RecievePort.first`)

app/lib/service/entrypoint/_messages.dart
```diff
@@ -0,0 +1,265 @@
...
+  /// Port to which any object may be sent periodically. After the first message
+  /// is sent, on each message, a timer will be set, and if no other message arrives,
+  /// the isolate will be killed and a new one will be started. The Timer's duration
+  /// is set at the isolate initialization and may depend on the task that the isolate
+  /// is doing.
+  ///
+  /// These message tell the isolate owner that the isolate isn't stuck.
+  /// In particular, this aims to allow the `IsolateRunner` to terminate
+  /// the isolate if it gets stuck in an infinite-loop.
+  final SendPort aliveSendPort;
```
jonasfj, line 124: Why do we need two ports? We can cleanup later, I'm fine with that.

But I'm just asking: Could we do one, since we're already differentiating after deserializing messages sent on the `protocolSendPort` ?

app/lib/service/entrypoint/_messages.dart
jonasfj: If the message never comes and there is a timeout, then `RecievePort.first` will never be resolved, and it'll never cause the subscription to be canceled.

app/lib/service/entrypoint/_isolate.dart
```diff
@@ -139,23 +123,67 @@ class IsolateGroup {
     }
   }
 
-  /// Starts [count] new isolates, and after a [wait] duration,
-  /// closes the old ones.
+  /// Starts [count] new isolates, waits for the pending requests to get processed,
+  /// and after a maximum [wait] duration, closes the old ones.
   Future<void> renew({
     required int count,
```
jonasfj, line 129: Is it not extremely dangerous that we can pass a different `count` to `renew` and `start`.

What is the behavior if this happens? is this fully thought through? or are we assuming that `count` will always have same value when a specific instance of `IsolateGroup` is called?

app/lib/service/entrypoint/_isolate.dart
```diff
@@ -139,23 +123,67 @@ class IsolateGroup {
...
+        i.markedForReplace == false &&
+        i._readyMessage?.requestSendPort != null);
+    if (last == null) {
+      logger.warning('No active isolate to process request.');
+      e.replyPort.send(
+          ReplyMessage.error('No isolate to process request.').encodeAsJson());
+      return;
+    }
+
+    final intermediatePort = ReceivePort();
```
jonasfj, line 176: More ports that are only closed if a message is actually sent to it. If a reply never arrives then it'll never be closed.

app/lib/service/entrypoint/_messages.dart
isoos: Yes we can (and shall) merge them eventually (or remove the aliveSendPort and the autokill timer)

app/lib/service/entrypoint/_messages.dart
```diff
@@ -0,0 +1,265 @@
...
+  ///
+  /// ```js
+  /// {
+  ///   reply: {
+  ///     error: <string>,
+  ///     result: <object>,
+  ///   }
+  /// }
+  /// ```
+  static Message decodeFromMap(Map<String, dynamic> map) {
```
jonasfj, line 66:
```suggestion
  static Message fromObject(Object? message) {
```

jonasfj (approved):

app/lib/service/entrypoint/_messages.dart
jonasfj: In case we want it to handle cases like `Message` and `Map<String, Object?>`.

PR #6906, opened 2023-08-14
Use SVG logo in header for increased clarity

parlough:
The existing PNG appears slightly blurry to me.

These screenshots don't really do it justice, but it's still noticable:

**Previously with PNG:**
Header image using PNG

**Now with SVG:**
Header image using SVG


isoos:
Looks good, here we also have the templates_tests failure.

parlough:
Thanks for the heads up.

I also had to remove the old file itself to fix another test since it now unused. Do you think that is reasonable or should I add it as an exception in that test?

isoos:
> I also had to remove the old file itself to fix another test since it now unused. Do you think that is reasonable or should I add it as an exception in that test?

Removal is fine, if we need the file again, it will be in git history.

jonasfj (approved):

isoos (approved):

PR #6907, opened 2023-08-14
Update Dart 3 compatible badge text

parlough:
Now that the release has occurred and analysis is running with a Dart 3+ SDK, the "expected" should no longer be needed.


isoos:
There seems to be a non-flake test failure in templates_test around the `activity-0-date` change, that needs to get fixed.

PR #6910, opened 2023-08-14
Use only pana's repository verification to display links.

isoos:
As we are now running repository verification for all packages in the sandbox, there is no reason to keep the pubspec-based fallback.


jonasfj (approved):

jonasfj:
So if packages aren't analyzed we'll have nothing?

Are we using the latest analysis for this, when displaying older versions? (should we?)

PR #6917, opened 2023-08-15
Sanity check pana report size in pub_worker.

isoos:
Ideally we would do this after the upload happened, but it won't hurt to check before it either.

### DIFF
--- a/pkg/pub_worker/lib/src/bin/pana_wrapper.dart
+++ b/pkg/pub_worker/lib/src/bin/pana_wrapper.dart
@@ -3,8 +3,8 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:convert' show json;
-import 'dart:io' show Directory, File, Platform, exit;
+import 'dart:convert' show json, utf8;
+import 'dart:io' show Directory, File, Platform, exit, gzip;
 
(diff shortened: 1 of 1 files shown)


pkg/pub_worker/lib/src/bin/pana_wrapper.dart
```diff
@@ -138,6 +142,30 @@ Future<void> main(List<String> args) async {
...
   final updatedSummary = summary.change(report: updatedReport);
 
+  // sanity check on pana report size
+  final reportSize =
+      gzip.encode(utf8.encode(json.encode(summary.toJson()))).length;
+  if (reportSize > _reportSizeDropThreshold) {
+    summary = Summary(
+      createdAt: summary.createdAt,
+      runtimeInfo: summary.runtimeInfo,
+      tags: ['has:pana-report-exceeds-size-threshold'],
```
jonasfj, line 152: Should we do a similar thing here:
https://github.com/dart-lang/pub-dev/blob/a10d45071d0ee399e3b9847fda2dce0dff357683/pkg/pub_worker/lib/src/analyze.dart#L278-L283

jonasfj: Not in this PR :D

jonasfj (approved):

PR #6920, opened 2023-08-16
Update security advisory model

szakarias:


app/test/service/security_advisory/testdata/example_advisory.json
```diff
@@ -0,0 +1,196 @@
...
+    }
+  ],
+  "schema_version": "1.4.0",
+  "severity": [
+    {
+      "type": "CVSS_V3",
+      "score": "CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:C/C:L/I:L/A:N"
+    }
+  ]
+}
```
jonasfj, line 196: missing line break at the end of the file.

app/test/service/security_advisory/security_advisory_test.dart
```diff
@@ -1,15 +1,26 @@
...
+import 'package:path/path.dart' as path;
 import 'package:pub_dev/service/security_advisories/backend.dart';
 import 'package:pub_dev/service/security_advisories/models.dart';
+import 'package:pub_dev/shared/utils.dart';
 import 'package:test/test.dart';
 
 import '../../shared/test_services.dart';
 
 void main() {
+  testWithProfile('Parse advisory', fn: () async {
```
jonasfj, line 13: Don't test with profile, if you're not really testing services with fake data.

```suggestion
  test('Parse advisory', fn: () async {
```

You could consider making a `app/test/service/security_advisory/osv_test.dart` with a few simple test vectors, I think having the real one from `package:http` is great.

if you want you an also just hardcode in a few fake ones to exercise the code. But this is probably not needed.

jonasfj: If you make it a separate `osv_test.dart` file, you won't have many other test cases in the file, so you could just make a big multi-line string with the JSON instead of having a file.

jonasfj (approved):

PR #6941, opened 2023-08-19
Dynamic `x-ago` calculation

AlexV525:
We recently rolled the new mirror website for pub.dev: https://pub-web.flutter-io.cn.

As a mirror site, it uses a caching mechanism that will cache the whole static page and won't update it until the package gets upgraded (by checking the `/api/packages`). In this case, the published time is fixed and won't follow the regular update as pub.dev does.

To help the caching, and maybe avoid constantly updating pages on pub.dev too (it's not just about the date but also likes, however that's another part of improvement which is not addressed here), the PR adds the ability to write the published date as `data-timestamp` attribute and update the actual `x-ago` text when setup scripts, so the text will always be up-to-date.

---

<details>

</details>

### DIFF
--- a/pkg/_pub_shared/lib/format/x_ago_format.dart
+++ b/pkg/_pub_shared/lib/format/x_ago_format.dart
@@ -1,0 +1,21 @@
+/// Formats an [age] duration with `<amount> <unit> ago` or `in the last hour`.
+String formatXAgo(Duration age) {
+  if (age.inDays > 365 * 2) {
+    final years = age.inDays ~/ 365;
+    return '$years years ago';
+  }
+  if (age.inDays > 30 * 2) {
+    final months = age.inDays ~/ 30;
+    return '$months months ago';
+  }
+  if (age.inDays > 1) {
+    return '${age.inDays} days ago';
+  }
+  if (age.inHours > 1) {
+    return '${age.inHours} hours ago';
+  }
+  if (age.inHours == 1) {
+    return '${age.inHours} hour ago';
+  }
+  return 'in the last hour';
+}
--- a/pkg/web_app/lib/src/x_ago.dart
+++ b/pkg/web_app/lib/src/x_ago.dart
@@ -1,0 +1,17 @@
+import 'dart:html';
+
+import 'package:_pub_shared/format/x_ago_format.dart';
+
+void setupXAgo() {
+  document.querySelectorAll('a.-x-ago').forEach((e) {
+    final DateTime actual = DateTime.parse(e.getAttribute('data-timestamp')!);
+    e.text = formatXAgo(DateTime.now().toUtc().difference(actual));
+    e.onClick.listen((event) {
+      event.preventDefault();
+      event.stopPropagation();
+      final text = e.text;
+      e.text = e.getAttribute('title');
+      e.setAttribute('title', text!);
+    });
+  });
+}
--- a/app/lib/frontend/dom/dom.dart
+++ b/app/lib/frontend/dom/dom.dart
@@ -109,6 +109,7 @@ Node xAgoTimestamp(DateTime timestamp, {String? datePrefix}) {
     attributes: {
       'aria-label': 'Switch between date and elapsed time.',
       'aria-role': 'button',
+      'data-timestamp': timestamp.toUtc().toIso8601String(),
     },
     text: formatXAgo(clock.now().difference(timestamp)),
   );
--- a/app/lib/frontend/dom/dom.dart
+++ b/app/lib/frontend/dom/dom.dart
@@ -4,10 +4,11 @@
 
 import 'dart:convert';
 
+import 'package:_pub_shared/format/x_ago_format.dart';
 import 'package:clock/clock.dart';
 
 import '../../shared/markdown.dart';
-import '../../shared/utils.dart' show formatXAgo, shortDateFormat;
+import '../../shared/utils.dart' show shortDateFormat;
 
 final _attributeEscape = HtmlEscape(HtmlEscapeMode.attribute);
 final _attributeRegExp = RegExp(r'^[a-z](?:[a-z0-9\-\_]*[a-z0-9]+)?$');
--- a/app/lib/frontend/templates/views/pkg/package_list.dart
+++ b/app/lib/frontend/templates/views/pkg/package_list.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:_pub_shared/format/x_ago_format.dart';
 import 'package:_pub_shared/search/search_form.dart';
 import 'package:_pub_shared/search/tags.dart';
 import 'package:clock/clock.dart';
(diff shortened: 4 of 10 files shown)


pkg/_pub_shared/lib/format/x_ago_format.dart
```diff
@@ -0,0 +1,21 @@
+/// Formats an [age] duration with `<amount> <unit> ago` or `in the last hour`.
```
isoos, line 1: Please add license header here.

pkg/web_app/lib/src/x_ago.dart
```diff
@@ -0,0 +1,17 @@
+import 'dart:html';
```
isoos, line 1: This seems to be a too small method to be moved into a separate file. Let's keep it in `hoverable.dart` (otherwise it requires license header).

isoos (changes requested):
In general this looks good to me. However, you'll need to re-generate golden files in `templates_test.dart` and in `task/end2end_test.dart` (flipping a const value in the beginning of the files and re-running the tests).

pkg/web_app/lib/src/x_ago.dart
AlexV525: It's worth splitting and you'll not need to remember that there is a hoverable configuration for `.-x-ago` in some other places.

AlexV525:
> However, you'll need to re-generate golden files in `templates_test.dart` and in `task/end2end_test.dart` (flipping a const value in the beginning of the files and re-running the tests).

@isoos Could you indicate more preparation steps for it? I ran tests (/app/tests) on Windows locally and got massive exceptions with most of the same traces.

```
package:pub_dev/frontend/static_files.dart 55:3    resolveAppDir
package:pub_dev/frontend/static_files.dart 60:20   resolveStaticDirPath
package:pub_dev/frontend/static_files.dart 99:29   new StaticFileCache.withDefaults
package:pub_dev/frontend/static_files.dart 119:41  new StaticFileCache.forTests
test\shared\test_services.dart 53:56               testWithProfile.<fn>.<fn>
package:pub_dev/service/services.dart 204:22       withFakeServices.<fn>.<fn>
===== asynchronous gap ===========================
package:pub_dev/service/services.dart 308:20       _withPubServices.<fn>.<fn>.<fn>
===== asynchronous gap ===========================
package:pub_dev/service/services.dart 301:12       _withPubServices.<fn>
===== asynchronous gap ===========================
package:pub_dev/service/services.dart 192:12       withFakeServices.<fn>
===== asynchronous gap ===========================
package:pub_dev/service/services.dart 152:10       withFakeServices
===== asynchronous gap ===========================
test\shared\test_services.dart 51:5                testWithProfile.<fn>
===== asynchronous gap ===========================
test\shared\utils.dart 29:14                       scopedTest.<fn>.<fn>

Exception: Unknown script: file:///C:/Users/AlexV525/AppData/Local/Temp/dart_test.kernel.8e008f6d/test.dart_22.dill
```

AlexV525:
ping @isoos

AlexV525:
ping @isoos

jonasfj (commented):
I don't really have any objections, though it might be nice to compute this on the server, to offer better support for clients without JS.

And more importantly avoid any risk of flickering after loading.

AlexV525:
> And more importantly avoid any risk of flickering after loading.

It won't flicker the pub.dev (until we no longer update pages periodically). And it should be updated immediately.

>  though it might be nice to compute this on the server, to offer better support for clients without JS.

I'd appreciate it if you have better ideas!

AlexV525:
cc @chenglu to share more details

chenglu:
We are using nginx reverse host proxy of the pub.dev web UI.

isoos:
> We are using nginx reverse host proxy of the pub.dev web UI.

In that case I think we would rather keep the current way of rendering and labels. Even if you are 2-3 hours late in caching the page, it is not the end of the world to have a bad "-ago" label, but I think it would be probably cached only for minutes... isn't it?

AlexV525:
We'll cache the page until it gets updated (version upgrade), which is why I purposed the change. Otherwise we'll cost too much for caches.

AlexV525:
The reason why we want a dynamic -ago is users and we cannot tell whether the package got updated correctly until we compare changes between pub.dev.

chenglu:
Displaying the correct package update time on the page will help developers choose better packages. If it can be calculated on the "front end", it will be a great help to the CN devs. Thanks to the team!

isoos:
I could see a compromise here that could be good for both pub.dev and any kind of proxy:

The JS updater is one-time only, it either updates the page on load or it doesn't update it at all (no period timer). The decision could be based on the domain, or on a timestamp we put there as page data: if the timestamp is recent (e.g. within 2 hours), the updater won't change anything. Otherwise we keep the current granularity (no minutes), and keep it simple for pub.dev.

wdyt?

isoos:
> Displaying the correct package update time on the page will help developers choose better packages.

Note: this could be done by more frequent proxy fetches too. We could also explore if we could help that in another way.

AlexV525:
Does the "JS Updater" indicate the script I'm changing?

> The decision could be based on the domain, or on a timestamp we put there as page data: if the timestamp is recent (e.g. within 2 hours), the updater won't change anything.

As I imagine, this adds complexity to scripts which would check for more conditions (involving domains). Let's say you'll only update pages once their metadata changes (likes, publishers, etc.) rather than visit-based, with the PR, you don't have to consider when pages will be updated, no matter which domain or updated time. (Or I got the wrong idea here, please correct me then)

isoos:
> We'll hold the cache until the package gets updated (version upgrade by checking /api/packages periodically), which is why I proposed the change. Otherwise, we'll cost too much for caches.

I think that won't give you when an analysis is updated for the package. If we could handle `if-modified-since` headers better for all pages, would that be better for your costs?

AlexV525:
> I think that won't give you when an analysis is updated for the package. If we could handle `if-modified-since` headers better for all pages, would that be better for your costs?

It's the most correct method but in this case, the published date (x-ago) is still outdated until the page is modified (which is counter-intuitive if you have to explain this).

cc @chenglu to check if we already have `if-modified-since` integration.

isoos:
@AlexV525 @chenglu: any update on the use of `if-modified-since`?

AlexV525:
> @AlexV525 @chenglu: any update on the use of `if-modified-since`?

We have no progress on it because it involves multiple systems coordination.

As far as I'm concerned, it would be hard to trace or support fields if they don't. So I might still want to stick to the solution provided here.

chenglu:
The `If-Modified-Since` header helps the server determine whether the client-side cache is outdated. If not outdated, a `304` is returned to save network bandwidth and improve access speed. 

To ensure faster access, the user-faced-front of the China mirror is a CDN service. I understand that we might set the CDN's cache to either not cache or have a shorter caching time, then rely on the server-side 304 to decide whether to deliver the CDN cached data to the user or re-fetch data from the origin. 

Unfortunately, our CDN provider informed us that most CDN providers do not offer such support during back-to-origin requests. Therefore, while this strategy might be effective if users access the origin directly, the significance of `If-Modified-Since` seems not very substantial under the current architecture. 

Please feel free to point it out if I am wrong, thanks!

@isoos Thanks for your time taking care of this case for all the CN devs!

AlexV525:
> Closing this as #7166 has been merged. Not sure when it gets deployed though, I'll ping this thread once it is out.

@isoos Thanks!

PR #6946, opened 2023-08-21
Fix spelling mistake on help index page

parlough:


parlough:
I keep forgetting to update the goldens, I'm sorry. Should be good now :)

Issue #6953, opened 2023-08-22
[calendar_date_picker2] package shows [UNKNOWN PLATFORMS] after new release


theideasaler:
I am the creator/publisher of package calendar_date_picker2 (https://pub.dev/packages/calendar_date_picker2).

I have released the new version of the package today and I saw it passed all testing and got 140 points not too long after the release.

But I just found the package is now showing [UNKNOWN PLATFORMS].

![IMG_8230](https://github.com/dart-lang/pub-dev/assets/17869748/6cd11c12-83a8-42b2-ae17-279c87f9d5af)

isoos:
@theideasaler: Thanks for reporting this. We are aware of a bug that affects newly uploaded packages, and there is a fix that should land soon. You don't need to do anything, the analysis, scores and platform tags should correct themselves automatically.

natsuk4ze:
@ Could you please change the tile of issue more versatile? like a "Analytics stuck in PENDING"

theideasaler:
I can see the analysis points are back. 

But looks like the package's position among search results has completely changed

isoos:
> But looks like the package's position in search results has completely changed

Could you please give us more details on that?

theideasaler:
> > But looks like the package's position in search results has completely changed
> 
> 
> 
> Could you please give us more details on that?

It used to be the first search result when search by keyword "date picker" or "calendar picker" by using Search Relevance filter.

And it was on the first page when search by keyword "range picker", "multi picker"

isoos:
> It used to be the first search result when search by keyword "date picker" or "calendar picker" by using Search Relevance filter.

The search index has not picked up the score change yet, it still has the pre-analysis empty score. It should correct itself automatically, I'll look into reasons why it may be missing out on it.

isoos:
We've updated how we handle this, it has been deployed, this should be fixed now.

PR #6954, opened 2023-08-22
Use the latest (version with) finished analysis when displaying generic scorecard, serving /documentation/<pkg>/latest/ or in search.

### LINKED ISSUE #6933: Documentation goes dead for a while after a package release
After a package is released, it takes some amount of time to analyze it and rebuild the documentation. For example, at the time of writing, `sass` 1.66.0 was released to pub 2 hours ago and is still listed as [PENDING ANALYSIS]. During this time, _no_ API documentation is available for this package. This is a bad experience for users of the package, and it breaks any automated link checking as well. Please only take down the old documentation once the new documentation is ready to serve.

isoos:
- Should fix #6933 and #6953
- When no version value is specified, instead of loading the latest stable version (and its assets), search, scorecard and `documentation/<pkg>/latest` will first access the latest version with any analysis finished, and use that version to load its assets.
- When the analysis (and dartdoc) completes for a newly uploaded package, the caches will be purged and the new version will be displayed.

### DIFF
--- a/app/lib/scorecard/backend.dart
+++ b/app/lib/scorecard/backend.dart
@@ -88,14 +88,16 @@ class ScoreCardBackend {
         return null;
       }
 
+      // Get the scorecard with the latest version available with finished analysis.
+      final card = await scoreCardBackend.getScoreCardData(package, null);
+
+      // Load the version with the analysis above, or the latest version if no analysis
+      // has been finished yet.
       final releases = await packageBackend.latestReleases(p);
       final version = releases.stable.version;
-      final pvFuture = packageBackend.lookupPackageVersion(package, version);
-      final cardFuture = scoreCardBackend.getScoreCardData(package, version);
-      await Future.wait([pvFuture, cardFuture]);
+      final pv = await packageBackend.lookupPackageVersion(
+          package, card?.packageVersion ?? version);
 
-      final pv = await pvFuture;
-      final card = await cardFuture;
       return PackageView.fromModel(
         package: p,
         releases: releases,
@@ -112,7 +114,8 @@ class ScoreCardBackend {
     bool onlyCurrent = false,
   }) async {
     if (packageVersion == null || packageVersion == 'latest') {
-      packageVersion = await packageBackend.getLatestVersion(packageName);
+      packageVersion = await taskBackend.latestFinishedVersion(packageName) ??
+          await packageBackend.getLatestVersion(packageName);
       if (packageVersion == null) {
         // package does not exists
         return null;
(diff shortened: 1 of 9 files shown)


app/lib/task/backend.dart
```diff
@@ -1024,6 +1027,40 @@ class TaskBackend {
...
+          continue;
+        }
+        if (finishedVersions.length == 1) {
+          return finishedVersions.single;
+        }
+        final bestVersion = finishedVersions
+            .map((e) => Version.parse(e))
+            .reduce((a, b) =>
+                compareSemanticVersionsDesc(a, b, true, true) <= 0 ? a : b);
+        return bestVersion.toString();
```
jonasfj, line 1058: Should we split this into a utility function called `latestVersion(Iterable<Version> versions)` so that it can have a few unit tests -- this is a great place to be off by one :D

app/lib/task/backend.dart
```diff
@@ -1024,6 +1027,40 @@ class TaskBackend {
...
+        final finishedVersions = state.versions?.entries
+            .where((e) => e.value.finished)
+            .map((e) => e.key)
+            .toList();
+        if (finishedVersions == null || finishedVersions.isEmpty) {
+          continue;
+        }
+        if (finishedVersions.length == 1) {
+          return finishedVersions.single;
+        }
```
jonasfj, line 1053: There is no need for this special case, is there?

app/lib/scorecard/backend.dart
```diff
@@ -96,7 +97,8 @@ class ScoreCardBackend {
     bool onlyCurrent = false,
   }) async {
     if (packageVersion == null || packageVersion == 'latest') {
-      packageVersion = await packageBackend.getLatestVersion(packageName);
+      packageVersion = await taskBackend.latestFinishedVersion(packageName) ??
+          await packageBackend.getLatestVersion(packageName);
```
jonasfj, line 101: Let's also update the documentation, it was never obvious what `latest` does, or that it was an allowed value.

Is `getScoreCardData` only used by search?

app/test/task/task_test.dart
```diff
@@ -675,6 +678,59 @@ void main() {
           TestUser(email: 'admin@pub.dev', likes: []),
         ],
       ));
+
+  group('latest analyzed version', () {
```
jonasfj, line 682: I'd be fine splitting this file in two.

Maybe we just put all the fake_time testing `task_faketime_test.dart` ?

jonasfj:
This also needs a bump of runtimeVersion

isoos:
> This also needs a bump of runtimeVersion

Doing that it another PR.

app/lib/task/backend.dart
```diff
@@ -1024,6 +1027,33 @@ class TaskBackend {
...
+          continue;
+        }
+        final bestVersion = state.versions?.entries
+            .where((e) => e.value.finished)
+            .map((e) => Version.parse(e.key))
+            .latestVersion;
+        if (bestVersion == null) {
+          continue;
+        }
+        return bestVersion.toString();
```
jonasfj, line 1051:
```suggestion
        if (bestVersion != null) {
          return bestVersion.toString();
        }
```

jonasfj (approved):
Make sure we bump runtime version along with this.

PR #6964, opened 2023-08-23
Fix SDK index parsing - use new fields and content.

### LINKED ISSUE #6956: dart:* packages return 404 on web
Clicking on a package, like dart:html, returns 404.

Eg: https://api.dart.dev/stable/3.1.0/

isoos:
Fixes #6956.


jonasfj (approved):
We won't need to support older versions for Dart/Flutter or are both SDKs failing parse at the moment?

isoos:
> We won't need to support older versions for Dart/Flutter or are both SDKs failing parse at the moment?

We were always aiming for the latest SDK docs, no need for older ones. Both of them are failing.

PR #6970, opened 2023-08-25
Remove isolate autokill feature.

isoos:
This is no longer relevant for since the job isolates are gone, and the other isolate types are fine if they are not going to get killed and restarted based on it.


isoos:
Let's not remove this, it looks useful.

jonasfj (approved):

PR #6976, opened 2023-08-28
Fix parsing of snake case fields in OSVs

szakarias:


app/test/service/security_advisory/security_advisory_test.dart
```diff
@@ -15,7 +15,48 @@ void main() {
...
+      expect(osv.affected!.first.ranges!.first.events!.last.fixed, '0.13.3');
+      expect(osv.affected!.first.versions!.length, 113);
+      expect(osv.affected!.first.versions![100], '0.8.2');
+      expect(osv.affected!.first.databaseSpecific!['source'],
+          'https://github.com/github/advisory-database/blob/main/advisories/github-reviewed/2022/05/GHSA-4rgh-jx4f-qfcq/GHSA-4rgh-jx4f-qfcq.json');
+
+      expect(osv.severity!.length, 1);
+      expect(osv.severity!.first.type, 'CVSS_V3');
+      expect(osv.severity!.first.score,
+          'CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:C/C:L/I:L/A:N');
```
jonasfj, line 59: Have you considered `DeepEquality().equals(OSV.fromJson(...).toJson(), ...)` asan added option...

I think these other things are fine to test too! :rofl:

jonasfj (approved):

app/test/service/security_advisory/security_advisory_test.dart
szakarias: The serialisation is not unique so I don't think this is applicable.

PR #6985, opened 2023-08-30
Reduce isolate count of frontend isolates (increase autoscaling of default service.

isoos:


jonasfj (approved):
I love for us to ship this as a single change, can we hold of landing it?

PR #6992, opened 2023-08-31
Overriding runtimeVersions in tests.

isoos:
I've found that the least amount of API change is to repeatedly call `fn` and inside that depend on what to do. It also makes sure that all the caches, the http server and everything is using the set acceptable runtime versions.


app/test/shared/test_services.dart
```diff
@@ -66,25 +68,35 @@ void testWithProfile(
   bool processJobsWithFakeRunners = false,
   Pattern? integrityProblem,
   dynamic skip,
+  List<List<String>>? runtimeVersions,
```
jonasfj, line 71: Please elaborate on what this does in the documentation comment.

jonasfj (approved):

app/test/shared/test_services.dart
jonasfj: We agree, that this completely changes how the test is run, right?

Suddenly, you're running more than once. That might be very surprising.

Also why is a list of lists?

isoos: List of lists: in each iteration we provide a new list of `acceptedRuntimeVersions`. If it were a single list, we couldn't change it arbitrarily. 

Tests running more than once: I agree, it is a non-trivial run behavior. The other change I have in mind would wrap the runtimeversions and the callback to run in a class and provide that as a list. E.g.

```
class RuntimeAndFn {
  final List<String> runtimeVersions;
  final Future<void> Function() fn;

  RuntimeAndFn({this.runtimeVersions, this.fn});
}

void testWithRuntimes(List<RuntimeAndFn> list) ....
```

Would that be better?

jonasfj: Yeah, something like `testWithRuntimes` or... we should split `testWithProfile` into multiple things can be combined.

```dart
void main() {
  test('something', () async {
    await withServiceScope(() async {
      await startFakeServices(); // launch fake storage and fake datastore, what else?

      // Launch the service once
      await withServiceScope(() async {
        await startServices(/* TODO: specify acceptedRuntimeVersions*/);

        // Load data into fake storage and fake datastore
        await loadProfile();

        // TODO: Assert some test properties
      });

      // Launch the service again
      await withServiceScope(() async {
        await startServices();

        // TODO: Assert some other test properties
      });
    });
  });
}
```

Using the idea would be that when `withServiceScope` returns, then everything that started within that service scope stops.

So we can start the server with the same same fake storage more than once, but nesting service scopes.

With primitives like this could easily make `testWithServiceScope` or `testWithProfile` or `testWithRuntimeVersionEvolution`.

(side note: I think `testWithRuntimeVersionEvolution` is better than `testWithRuntimes`, because you're really testing the evolution of runtime versions over time).

--------------------

I don't know if this is sane, we could also just make `testWithRuntimeVersionEvolution` and not make all the primitives, a little code duplication isn't the worst. The primitives is mostly if we wanted to avoid duplication I guess, we already have a bit of duplication in `testWithFakeTime` :see_no_evil:


--------------------

Alternatively, what we could do to test runtimeVersion fallback, is:
 * Whenever we create a new runtimeVersion, we:
   * Import a specific profile
   * Save data from fake storage/datastore to a `test/testdata/runtime-version-data/<runtimeVersion>.json` file
   * Check this `test/testdata/runtime-version-data/` into git
   * Delete `test/testdata/runtime-version-data/<rt>.json` for any `rt` that isn't in the `acceptedRuntimeVersions` list anymore.
 * Create a test function call `testRuntimeVersionEvolution(fn)` where `fn` gets called with `test/testdata/runtime-version-data/` loaded into fake datastore/storage and services running.

Or maybe, we only need the previous runtimeVersion.

isoos: Storing the runtime-specific version as json may be useful on its own, but then we need to do it every time we upgrade the runtime. I treat it as a separate concern here.

I'll try to refactor the test for better primitives, I think into a class that can be reused in parts or as a whole...

isoos: Updated. PTAL.

app/test/shared/test_services.dart
```diff
@@ -55,6 +58,70 @@ Future<void> withRuntimeVersions(
...
 }
 
+/// The stored and shared state of the appengine context (including datastore,
+/// storage and cloud compute).
+class FakeAppengineEnv {
+  final _storage = MemStorage();
+  final _datastore = MemDatastore();
+  final _cloudCompute = FakeCloudCompute();
+
+  /// Starts a new environment and runs [fn] in it.
```
jonasfj, line 68:
```suggestion
  /// Create a service scope with fake services and run [fn] in it.
  ///
  /// All services will terminate when [fn] returns.
  ///
  /// Calling [run] more than once, is allowed. Services will not survive between
  /// invocations, but datastore, storage and cloud compute will retain state.
```

jonasfj (approved):
Love it!

PR #7007, opened 2023-09-05
Upgrade dartdoc to 6.3.0 (+ code cleanup)

isoos:
I'm unsure about `base-href`, and `using-base-href`, but as the dartdoc JS code seems to depend on it, we may expect changes in the future, so better to store it for now.


pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
```diff
@@ -179,7 +179,9 @@ Future<void> _dartdoc({
     final suffix = file.path.substring(docDir.length + 1);
     final targetFile = File(p.join(tmpOutDir, suffix));
     await targetFile.parent.create(recursive: true);
-    if (file.path.endsWith('.html')) {
+    final isDartDocPage =
+        file.path.endsWith('.html') && !file.path.endsWith('-sidebar.html');
```
jonasfj, line 183: So we're not going to sanitize the sidebar?

I suspect we need a `DartDocSideBar` class.

pkg/pub_dartdoc_data/lib/dartdoc_page.dart
```diff
@@ -52,13 +52,22 @@ final class DartDocPage {
   /// Sanitized HTML for the [content] pane.
   final String content;
 
+  final String? baseHref;
+  final String? usingBaseHref;
+  final String? aboveSidebarUrl;
+  final String? belowSidebarUrl;
```
jonasfj, line 58: If you look in the documentation string for `DartDocPage` there is no sidebars above or below.

Also can we have documentation comments for these.

If they are used by client side to load HTML files, then clearly we should explain this.

jonasfj (commented):
Given that you're modifying files in `third_party/dartdoc/resources/` it's safe to assume that documentation from existing runtimeVersions may not work.

Probably changes are so small it's not a huge issue, but let's always assume that presence or absence of some div/class/id might break the scripts.

Similarly, modifying `DartDocPage` likely means we cannot reuse data from previous runtimeVersions. I'd certainly suggest that we don't try.

It's easy to empty the `acceptableRuntimeVersions` list, bump runtimeVersion and wait 24 hours after deployment for most things to be processed before migrating traffic.

jonasfj:
Okay, after chat I'm convinced we don't need to empty `acceptableRuntimeVersions` for this change.

But sanitizing the sidebar seems worthwhile.

jonasfj (approved):

PR #7024, opened 2023-09-07
Do not redirect to /latest/ doc URL and keep the canonical URL in-sync with the URL segment

isoos:
- follow-up to #6994
- with this change, the canonical URL inside the HTML meta headers will always render the same URL that the page is served on
- we no longer redirect `/d*/<pkg>/<v>/` to `/d*/<pkg>/latest/` if the request was for a specific version (even though it may be the latest version)

### DIFF
--- a/app/lib/task/handlers.dart
+++ b/app/lib/task/handlers.dart
@@ -42,6 +42,7 @@ Future<shelf.Response> handleDartDoc(
   shelf.Request request,
   String package,
   String version,
+  String segment,
   String path,
 ) async {
   InvalidInputException.checkPackageName(package);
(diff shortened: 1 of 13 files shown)


isoos:
Note: not sure why the end2end test fails in unexpected ways. I may bisect the PR into two parts...

isoos:
> Note: not sure why the end2end test fails in unexpected ways. I may bisect the PR into two parts...

Found the bug in the cache key. PTAL.

app/lib/task/handlers.dart
```diff
@@ -42,6 +42,7 @@ Future<shelf.Response> handleDartDoc(
   shelf.Request request,
   String package,
   String version,
+  String segment,
```
jonasfj, line 45: I see how we need to pass this around, but we need to call it something else.

`segment` is a name nobody can guess what is.
If we can't come up with a self-explanatory name, then we should document it.


We could also make this a `bool hasLatestInUrl`, right?
Because `segment == version || segment == 'latest'`, right? (I could be wrong here?, there might be other cases where this isn't the case, in particular when caching a redirect).

jonasfj (approved):

Issue #6357, opened 2023-01-20
Admin tool for "dissolving" a publisher


sigurdm:
Can eg. be used if original owners have lost control with the domain.

- All users in the publisher become uploaders on each package in the publisher.
- Should delete the publisher (leave a tombstone)

sigurdm:
Perhaps this should actually be two tools:

* A tool for removing a package from publisher, and make users in publisher uploader
* A tool for deleting an empty publisher (already exists).

PR #7030, opened 2023-09-11
Admin tool to remove a package from its publisher

sigurdm:
The package gets the publishers owners as uploaders

https://github.com/dart-lang/pub-dev/issues/6357


app/lib/admin/tools/remove_package_from_publisher.dart
```diff
@@ -0,0 +1,46 @@
...
+import 'package:pub_dev/package/models.dart';
+import 'package:pub_dev/publisher/backend.dart';
+import 'package:pub_dev/shared/datastore.dart';
+
+final _argParser = ArgParser()
+  ..addOption('package', help: 'The package to update.')
+  ..addOption('publisher-id', help: 'The `publisherId` to set.')
+  ..addFlag('help', abbr: 'h', defaultsTo: false, help: 'Show help.');
+
+Future<String> executeRemovePackageFromPublisher(List<String> args) async {
```
jonasfj, line 18: Would be nicer to make admin actions:
https://github.com/dart-lang/pub-dev/blob/master/app/lib/admin/actions/actions.dart

We don't have CLI for admin-actions yet, but it should be easy to build.

jonasfj (approved):

app/lib/admin/tools/remove_package_from_publisher.dart
sigurdm: Ah yes, I forgot you landed these. Much nicer!

Done

app/lib/admin/tools/remove_package_from_publisher.dart
```diff
@@ -0,0 +1,46 @@
...
+  }
+  final currentPublisherMembers =
+      (await publisherBackend.listPublisherMembers(currentPublisherId));
+
+  await withRetryTransaction(dbService, (tx) async {
+    final pkg = await tx.lookupValue<Package>(package.key);
+    pkg.publisherId = null;
+    pkg.uploaders = currentPublisherMembers.map((e) => e.userId).toList();
+    pkg.updated = clock.now().toUtc();
+    tx.insert(pkg);
```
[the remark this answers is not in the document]
sigurdm: Should we instead keep a separate log of all admin-actions and their output?

isoos (approved):

app/lib/admin/tools/remove_package_from_publisher.dart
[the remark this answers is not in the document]
jonasfj: > Hmm, what should it say?

The action taken and that `support@pub.dev` did it.

app/lib/admin/actions/remove_package_from_publisher.dart
```diff
@@ -0,0 +1,68 @@
...
+    options: {
+      'package': 'name of the package to remove from its current publisher'
+    },
+    summary:
+        'Removes <package> from its current publisher, and makes that publisher\'s owners uploaders',
+    description: '''
+Removes <package> from its current publisher, and makes that publisher\'s owners
+ uploaders.
+
+Fails if that package currently has no uploader.
```
jonasfj, line 24: This doesn't make sense, should it be negated?

app/lib/admin/actions/remove_package_from_publisher.dart
```diff
@@ -0,0 +1,68 @@
...
+
+Example: given a publisher example.com with two members a@example.com and
+ b@example.com, and apackage:foo in the publisher.
```
jonasfj, line 27:
```suggestion
 b@example.com, and a package:foo in the publisher.
```

jonasfj (approved):

app/lib/admin/tools/remove_package_from_publisher.dart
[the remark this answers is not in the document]
sigurdm: Done

app/lib/admin/actions/remove_package_from_publisher.dart
sigurdm: Good catch, should be publisher

PR #7033, opened 2023-09-12
Refactor: non-nullable fields in PackageView.

isoos:


isoos:
Note: `PackageView` is only cached/stored with the current runtime version cache key, there is no fallback requirement.

jonasfj (approved):

PR #7036, opened 2023-09-13
Add sync time to security advisory model

szakarias:


app/lib/service/security_advisories/backend.dart
```diff
@@ -79,6 +79,7 @@ class SecurityAdvisoryBackend {
       }
 
       if (!_isValidAdvisory(osv)) return null;
+      final syncTime = clock.now();
```
jonasfj, line 82: I think this is sound, but it might be slightly more elegant to pass `syncTime` into `ingestSecurityAdvisory` such that all advisories from the same zip gets the same `syncTime`.

jonasfj (approved):

app/lib/service/security_advisories/backend.dart
szakarias: done

PR #7043, opened 2023-09-15
Fix dartdoc page template + test for unexpected differences.

isoos:


app/test/dartdoc/dartdoc_page_test.dart
```diff
@@ -0,0 +1,352 @@
...
+import 'package:html/parser.dart' as html_parser;
+import 'package:path/path.dart' as p;
+import 'package:pub_dev/dartdoc/dartdoc_page.dart';
+import 'package:pub_dev/frontend/static_files.dart';
+import 'package:stream_transform/stream_transform.dart';
+import 'package:test/test.dart';
+import 'package:xml/xml.dart';
+
+void main() {
+  group('DartDocPage rendering', () {
```
jonasfj, line 18: If this test does what I think it does, then it deserves 5 lines explaining what it is.

Maybe consider if you need both group and test, since you only have a single test() anyways.

You can probably just do the setup and teardown in a try/finally...

jonasfj: Like how is someone opening this test file because it fails supposed to guess what it's trying to do?

This is the thing that test if output from dartdoc pass through DartDocPage rendering pipeline is the same as output from dartdoc. Maybe just explain: What we're testing, why, and how we're doing it (perhaps even explain why we're doing it this way).

We are doing this because we want to updated the DartDocPage rendering pipeline when dartdoc templates change, right?

jonasfj (approved):

PR #7051, opened 2023-09-19
Start using like and popularity score inside the search snapshot.

isoos:
The search index should not synchronously blocking on the like normalization (in a follow-up PR once we have started using this).

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -165,9 +165,9 @@ class SearchBackend {
     if (!claim.valid) {
       return;
     }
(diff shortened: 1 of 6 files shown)


app/lib/search/backend.dart
```diff
@@ -166,6 +167,7 @@ class SearchBackend {
     }
 
     // first complete snapshot, uploading it
+    snapshot.updateLikeScores();
```
jonasfj, line 170: Move this above the comment that says we're uploading it.

jonasfj (approved):

PR #7053, opened 2023-09-19
Migrate create-publisher to action

sigurdm:
The --amin arg was never used, so it is removed


app/lib/admin/actions/create_publisher.dart
```diff
@@ -0,0 +1,91 @@
...
+import 'package:pub_dev/admin/actions/actions.dart';
+import 'package:pub_dev/audit/models.dart';
+import 'package:pub_dev/publisher/models.dart';
+import 'package:pub_dev/shared/datastore.dart';
+
+final createPublisher = AdminAction(
+    name: 'create-publisher',
+    options: {
+      'publisher': 'name of publisher to create',
+      'member': 'email of user to add',
```
isoos, line 17: maybe call it `--member-email` so we can also use `member-user-id` later if needed?

isoos (approved):

PR #7057, opened 2023-09-21
Concurrent data loader for package pages.

isoos:
- loads every entity only a single time
- awaits them concurrently instead of sequentially
- the loaded entities are passed-in into methods that may use them if the cache is empty

Note: current `/debug` endpoint suggests about `90th` percentile to be in the 90-110 ms range, `99th` percentile in the 200-300 ms range.

### DIFF
--- a/app/lib/frontend/handlers/package.dart
+++ b/app/lib/frontend/handlers/package.dart
@@ -334,28 +334,36 @@ Future<PackagePageData> loadPackagePageData(
 
   final latestReleasesFuture = packageBackend.latestReleases(package);
 
-  final isLikedFuture = Future(
-    () async => requestContext.isNotAuthenticated
-        ? false
-        : await likeBackend.getPackageLikeStatus(
-                requestContext.authenticatedUserId!, package.name!) !=
-            null,
-  );
+  final isLikedFuture = Future(() async {
+    if (requestContext.isNotAuthenticated) {
+      return false;
+    }
+    final likeStatus = await likeBackend.getPackageLikeStatus(
+      requestContext.authenticatedUserId!,
+      package.name!,
+    );
+    return likeStatus != null;
+  });
 
   final selectedVersionFuture =
       packageBackend.lookupPackageVersion(packageName, versionName!);
   final versionInfoFuture =
       packageBackend.lookupPackageVersionInfo(packageName, versionName);
 
-  final assetFuture = Future(() async => assetKind == null
-      ? null
+  final assetFuture = assetKind == null
+      ? Future.value(null)
       : packageBackend.lookupPackageVersionAsset(
-          packageName, versionName!, assetKind));
+          packageName,
+          versionName,
+          assetKind,
+        );
 
-  final isAdminFuture = Future(() async => requestContext.isNotAuthenticated
-      ? false
-      : await packageBackend.isPackageAdmin(
-          package, requestContext.authenticatedUserId!));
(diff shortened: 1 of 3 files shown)


app/lib/frontend/handlers/package.dart
```diff
@@ -324,61 +325,77 @@ Future<shelf.Response> packageActivityLogHandler(
...
-  final package = await packageBackend.lookupPackage(packageName);
-  if (package == null || package.isNotVisible) {
-    // note: this should not happen, as isVisible check is done before calling the method
-    throw NotFoundException.resource('package "$package"');
-  }
+  final packageName = package.name!;
+  versionName ??= package.latestVersion;
+
+  final latestReleasesFuture =
+      Future(() => packageBackend.latestReleases(package));
```
jonasfj, line 336: Why not just:
```suggestion
  final latestReleasesFuture = packageBackend.latestReleases(package);
```

app/lib/frontend/handlers/package.dart
```diff
@@ -324,61 +325,77 @@ Future<shelf.Response> packageActivityLogHandler(
...
+
+  final isLikedFuture = Future(
+    () async => requestContext.isNotAuthenticated
+        ? false
+        : await likeBackend.getPackageLikeStatus(
+                requestContext.authenticatedUserId!, package.name!) !=
+            null,
+  );
+
+  final selectedVersionFuture = Future(
```
jonasfj, line 346: This future may go unawaited! It's possible that `getScoreCardData` won't await it!

jonasfj: In general the fact that we have to trace it that way suggests that we need to do this differently.

app/lib/frontend/handlers/package.dart
```diff
@@ -324,61 +325,72 @@ Future<shelf.Response> packageActivityLogHandler(
...
+
+  final assetFuture = Future(() async => assetKind == null
+      ? null
+      : packageBackend.lookupPackageVersionAsset(
+          packageName, versionName!, assetKind));
+
+  final isAdminFuture = Future(() async => requestContext.isNotAuthenticated
+      ? false
+      : await packageBackend.isPackageAdmin(
+          package, requestContext.authenticatedUserId!));
```
jonasfj, line 358: nit: How about?
```suggestion
  final isAdminFuture = requestContext.isNotAuthenticated
      ? Future.value(false)
      : await packageBackend.isPackageAdmin(
          package,
          requestContext.authenticatedUserId!,
        );
```

trailing comma, and using `Future.value(false)` is a tiny bit easier to read (perhaps).

jonasfj (approved):

app/lib/frontend/handlers/package.dart
```diff
@@ -5,7 +5,7 @@
 import 'dart:async';
 import 'dart:io';
 
-import 'package:_pub_shared/data/advisories_api.dart';
+import 'package:_pub_shared/data/advisories_api.dart' as advisories_api;
```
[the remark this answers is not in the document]
isoos: Naming: I think handlers.dart is not specialized to serve only advisories, it may be better to keep the full library alias.
Conflicts: `Package` is a conflict.

app/lib/frontend/handlers/package.dart
isoos: updated this and the above suggestions as a separate commit, as we had merge conflicts

PR #7061, opened 2023-09-25
Use Set.from(tags) in search tags matching.

isoos:
Improves the tag matching latency (overall latency gain on the slow-query benchmark is about 10%).


isoos:
Note also considered and benchmarked: `SplayHashSet` and sort list + binary search.

jonasfj (approved):

PR #7069, opened 2023-09-27
Chat from build

sigurdm:
This should send a chat message when the build is complete/failed taking away the need for monitoring the build.


tool/.deploy.sh
```diff
@@ -4,8 +4,18 @@
...
 # Always exit on error
 set -e
+
+# Echos The message and sends it to team chat
+message() {
+  echo $1
+  CHAT_ID=AAAAkQUOtE8
+  THREAD_KEY=$TAG_NAME
+  DATA='{"text": "'"$1"'"}'
+  curl -H 'Content-Type: application/json' -X POST https://chat.googleapis.com/v1/spaces/$CHAT_ID/messages?key=$CHAT_KEY\&token=$CHAT_TOKEN\&threadKey=$THREAD_KEY --data $DATA
```
jonasfj, line 14: Can we have double quotes for strings? don't know if we need to remove the backslashes around &
```suggestion
  curl -H 'Content-Type: application/json' -X POST "https://chat.googleapis.com/v1/spaces/$CHAT_ID/messages?key=$CHAT_KEY\&token=$CHAT_TOKEN\&threadKey=$THREAD_KEY" --data $DATA
```

I personally use:
https://www.shellcheck.net/
Via: https://github.com/vscode-shellcheck/vscode-shellcheck

jonasfj (approved):

tool/.deploy.sh
sigurdm: > https://www.shellcheck.net/

What a nice tool!

PR #7072, opened 2023-09-28
Reduce test flakes by force-closing dialog.

isoos:


pkg/pub_integration/lib/src/pub_puppeteer_helpers.dart
```diff
@@ -221,6 +223,31 @@ extension PubPageExt on Page {
...
+    }
+
+    // return if the dialog is no longer present
+    final dialogs = await $$('.mdc-dialog');
+    if (dialogs.isEmpty) {
+      return;
+    }
+
+    // click on the neutral area of the page
+    await mouse.click(Point(1, 1));
```
jonasfj, line 244: Would it not be better to pick a neutral element and click that? I'm not sure if this is supported.

jonasfj (approved):

pkg/pub_integration/lib/src/pub_puppeteer_helpers.dart
isoos: The role of the mouse click here is to click on the background element that makes the UI behind the modal dialog darker. We could figure out the exact element where the material dialog component listens to this click event, but then it may be an implementation detail in that widget, and may change in the future. I think it is better to click on the UI outside of the dialog (and 1,1 is usually outside of that).

PR #7079, opened 2023-09-29
Fix status inconsistencies on package score page.

isoos:
- Previously we had certain rules that prevented analysis, and those rules were duplicated on the score page, explaining why the analysis was skipped. Right now this leads to inconsistent pages like https://pub.dev/packages/unscrambler/score, and the new rules should fix that.
- Also removing `FlagMixin` as it has little relevancy at this point.


szakarias (approved):
LGTM

PR #7086, opened 2023-10-04
Export /api/packages/<pkg> JSON documents to storage bucket

isoos:


app/lib/package/export_api_to_cdn.dart
```diff
@@ -0,0 +1,239 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
sigurdm, line 1: Should we call the file `export_api_to_bucket` instead? As the cdn is somewhat orthogonal.

app/lib/package/export_api_to_cdn.dart
```diff
@@ -0,0 +1,239 @@
...
+Future<void> gcPackageApiCdn(Bucket bucket) async {
+  final versions = <String>{};
+
+  // Objects in the bucket are stored under the following pattern:
+  //   `current/api/<package>`
+  //   `<runtimeVersion>/api/<package>`
+  // Thus, we list with `/` as delimiter and get a list of runtimeVersions
+  await for (final d in bucket.list(prefix: '', delimiter: '/')) {
+    if (!d.isDirectory) {
+      _logger.warning('bucket should not contain any top-level object');
```
sigurdm, line 219: Please mention `$d` in the message for easier debugging.

app/lib/package/export_api_to_cdn.dart
```diff
@@ -0,0 +1,239 @@
...
+      continue;
+    }
+
+    // Remove trailing slash from object prefix, to get a runtimeVersion
+    assert(d.name.endsWith('/'));
```
sigurdm, line 224: Are we running with asserts enabled in production?

If we are, this should probably just print an error and skip the item instead of crashing?

app/lib/package/export_api_to_cdn.dart
```diff
@@ -0,0 +1,239 @@
...
+  String package,
+  Bucket bucket,
+) async {
+  final data = await retry(() => packageBackend.listVersions(package));
+  final rawBytes = jsonUtf8Encoder.convert(data.toJson());
+  final gzippedBytes = gzip.encode(rawBytes);
+
+  for (final objectName in _cdnObjectNames(package)) {
+    final info = await bucket.tryInfo(objectName);
+    // Skip upload if was updated recently, and the bytes length and md5 hash matches.
```
sigurdm, line 149: Why do we reupload after 12 hours? If the hash matches we should still be good - no?

app/lib/package/export_api_to_cdn.dart
```diff
@@ -0,0 +1,239 @@
...
+      () => Stream.value(gzippedBytes),
+      metadata: ObjectMetadata(
+        contentType: 'application/json',
+        contentEncoding: 'gzip',
+      ),
+    );
+  }
+}
+
+Future<void> _deletesPackageFromBucket(String package, Bucket bucket) async {
```
sigurdm, line 174:
```suggestion
Future<void> _deletePackageFromBucket(String package, Bucket bucket) async {
```

app/lib/package/export_api_to_cdn.dart
```diff
@@ -0,0 +1,239 @@
...
+    final f = pool.withResource(() => update(mp.name!, false));
+    futures.add(f);
+  }
+  await Future.wait(futures);
+  futures.clear();
+  if (!claim.valid) {
+    return;
+  }
+
+  await for (final package in dbService.query<Package>().run()) {
```
sigurdm, line 98: Will a package ever be in both ModeratedPackage and Package at the same time?

app/lib/package/export_api_to_cdn.dart
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    // Check if the runtimeVersion should be GC'ed
+    if (shouldGCVersion(rtVersion)) {
+      versions.add(rtVersion);
+    }
+  }
+
+  for (final v in versions) {
+    await deleteBucketFolderRecursively(bucket, '$v/', concurrency: 4);
```
sigurdm, line 237: Just an aside (not for this CL) perhaps we can use batch requests https://cloud.google.com/storage/docs/batch to delete files faster.

app/lib/package/export_api_to_cdn.dart
```diff
@@ -0,0 +1,239 @@
...
+import 'models.dart';
+
+final Logger _logger = Logger('export_api_to_cdn');
+
+/// The default concurrency to upload API JSON files to the CDN bucket.
+const _defaultUploadPackageApisToCdnBucketConcurrency = 8;
+
+List<String> _cdnObjectNames(String package) => [
+      '$runtimeVersion/api/packages/$package',
+      'current/api/packages/$package',
```
sigurdm, line 31: I forgot our strategy wrt. ensuring that only the "current" deployment writes to `current`... how does that work?

sigurdm (commented):
Let's also hear from @jonasfj on this :)

app/lib/package/export_api_to_cdn.dart
isoos: Maybe we don't need to when hashes match, but I'd rather be wrong by writing it more than we needed.

app/lib/package/export_api_to_cdn.dart
isoos: It won't at the same time, at least in the long run. But as the queries happen with some delay, a package may appear in both for a short time.

app/lib/package/export_api_to_cdn.dart
isoos: IIRC every deployment should write to `current/` and we think that it will be fine.

jonasfj (approved):

app/lib/package/export_api_to_bucket.dart
```diff
@@ -0,0 +1,244 @@
...
+import 'models.dart';
+
+final Logger _logger = Logger('export_api_to_cdn');
+
+/// The default concurrency to upload API JSON files to the CDN bucket.
+const _defaultUploadPackageApisToCdnBucketConcurrency = 8;
+
+List<String> _cdnObjectNames(String package) => [
+      '$runtimeVersion/api/packages/$package',
+      'current/api/packages/$package',
```
jonasfj, line 31: We should probably also have:
`/api/package-name-completion-data`

isoos:
@jonasfj: PTAL. I've updated the code, applying most of what we have discussed. The `package-name-completion-data` may be be added tomorrow.

app/lib/package/export_api_to_bucket.dart
```diff
@@ -0,0 +1,297 @@
...
+      final eventsSince = lastQueryStarted.subtract(Duration(minutes: 5));
+      await for (final event in _queryRecentPkgUpdatedEvents(eventsSince)) {
+        if (!claim.valid) {
+          break;
+        }
+        final f = pool.withResource(() async {
+          if (!claim.valid) {
+            return;
+          }
+          await _processPkgUpdated(event);
```
jonasfj, line 161: I'd suggest moving the contents of `_processPkgUpdate` in-line here.

Then you don't need `_pkgLastUpdated` to be a member, it can just be a variable in the scope of this function. That makes it much easier to track, and in some weird world where `incrementalPkgScanAndUpload` is invoked twice, concurrently, those two invocations won't interfere with each other by virtue of messing with `_pkgLastUpdated`.

isoos: We may move `_processPkgUpdated` inline, but we will need to update the (daily) full scan to repeat most of its logic. Which would be fine if we don't want to skip the update based on a timestamp... which is probably fine, as that process may not be triggered on the same machine at all.

app/lib/package/export_api_to_bucket.dart
```diff
@@ -0,0 +1,297 @@
...
+    await pool.close();
+  }
+
+  /// Uploads the package version API response bytes to the bucket, mirroring
+  /// the endpoint name in the file location.
+  Future<void> _uploadPackageToBucket(String package) async {
+    final data = await retry(() => packageBackend.listVersions(package));
+    final rawBytes = jsonUtf8Encoder.convert(data.toJson());
+    final gzippedBytes = gzip.encode(rawBytes);
+
```
jonasfj, line 191: Why not compute md5 here?

app/lib/package/export_api_to_bucket.dart
```diff
@@ -0,0 +1,297 @@
...
+    final md5Hash = md5.convert(bytes).bytes;
+    if (md5Hash.length != info.md5Hash.length) {
+      return false;
+    }
+    for (var i = 0; i < md5Hash.length; i++) {
+      if (md5Hash[i] != info.md5Hash[i]) {
+        return false;
+      }
+    }
+    return true;
```
jonasfj, line 283: I'd suggest moving the comparison logic for comparing two byte strings that contain a hash into a `fixedTimeComparison` function -- or even better, used a fixed-time comparison function from somewhere else.

app/lib/package/export_api_to_bucket.dart
jonasfj: But the daily full scan should not interfere with `_pkgLastUpdated`, and certainly shouldn't populate it. That'll just burn an awful lot of memory for no purpose.

Especially, if it's not running on the instance that has the lock and is currently scanning.

I think my point was that `_pkgLastUpdated` should be private to `incrementalPkgScanAndUpload`. You can also just pull those bits out of `_processPkgUpdated`

app/lib/package/export_api_to_bucket.dart
isoos: Changed to fixed time. Although, as this is called in a background process (or maybe later *after* an upload was successful, asynchronously), I don't see why timing has any relevance here.

app/lib/package/export_api_to_bucket.dart
isoos: We do that in `_isSameContent`. Oh, we may do it twice, that's true...

isoos: Added lazy calculation.

app/lib/package/export_api_to_bucket.dart
isoos: Agreed and done.

app/lib/package/export_api_to_bucket.dart
jonasfj: Fixed time is not relevant, it's just always seems wrong to compare hashes in non-fixed time :D

PR #7091, opened 2023-10-06
Fix dark mode diff highlight.

### LINKED ISSUE #7082: [doc] Inappropriate `diff` text color under the dark mode
Visit https://pub.dev/documentation/dio/latest/topics/Migration%20Guide-topic.html and switch to the dark mode.

image

isoos:
Fixes #7082.

### DIFF
--- a/third_party/css/github-markdown.css
+++ b/third_party/css/github-markdown.css
@@ -397,7 +397,7 @@
 
 .dark-theme .hljs .hljs-deletion,
 .dark-theme .hljs .hljs-addition {
-  /* TODO: Investigate if this should be fixed in dartdoc's styles.css: the dark theme is not applied there. */
+  /* TODO(https://github.com/dart-lang/pub-dev/issues/7101): Investigate if this should be fixed in dartdoc's styles.css: the dark theme is not applied there. */
   /*       https://github.com/dart-lang/pub-dev/issues/7101 */
   color: black;
 }
--- a/third_party/css/github-markdown.css
+++ b/third_party/css/github-markdown.css
@@ -393,6 +393,7 @@
   /* TODO: Investigate if this should be fixed in dartdoc's styles.css: the `pre` rule has precendece over the dark theme there. */
   background-color: inherit;
 }
+
 .dark-theme .hljs .hljs-deletion,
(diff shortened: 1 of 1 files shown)


srawlins:
Great question thanks for asking, @isoos !!! It looks like we implement the themes purely with CSS variables, see [this section](https://github.com/dart-lang/dartdoc/blob/main/lib/resources/styles.css#L36). And then down below, we have a lot of stylings on `.hljs*` stuff, [starting here](https://github.com/dart-lang/dartdoc/blob/main/lib/resources/styles.css#L250).

It looks like the fix over there might be a new declaration like this?

```css
.hljs-deletion, .hljs-addition {
  color: --main-text-color; /* or --main-h-color? */
}
```

`--main-text-color` is `#111111` and `--main-h-color` is `black`. 🤷 

CC @parlough

third_party/css/github-markdown.css
```diff
@@ -393,6 +393,11 @@
   /* TODO: Investigate if this should be fixed in dartdoc's styles.css: the `pre` rule has precendece over the dark theme there. */
   background-color: inherit;
 }
+.dark-theme .hljs .hljs-deletion,
```
sigurdm, line 396: newline above this line?
```suggestion

.dark-theme .hljs .hljs-deletion,
```

third_party/css/github-markdown.css
```diff
@@ -393,6 +393,11 @@
...
+.dark-theme .hljs .hljs-addition {
+  /* TODO: Investigate if this should be fixed in dartdoc's styles.css: the dark theme is not applied there. */
```
sigurdm, line 398: Could we leave an issue number with the TODO? Otherwise I fear it might never happen.

sigurdm (approved):
LGTM with nits

third_party/css/github-markdown.css
isoos: Done.

sigurdm: I would write like:
```suggestion
  /* TODO(https://github.com/dart-lang/pub-dev/issues/7101): Investigate if this should be fixed in dartdoc's styles.css: the dark theme is not applied there. */
```

PR #7095, opened 2023-10-06
Refactor InMemorySearchIndex initialization.

isoos:
- Removes the need to call `markReady`.
- Folds most use of `addPackage(s)` methods into the constructor, in preparation to refactor the index to use a single-step initialization.


sigurdm (approved):
LGTM

PR #7099, opened 2023-10-09
Track latest advisory on the Package entity.

szakarias:


app/lib/service/security_advisories/backend.dart
```diff
@@ -93,20 +96,65 @@ class SecurityAdvisoryBackend {
             osv.published != null ? DateTime.parse(osv.published!) : modified
         ..syncTime = syncTime;
 
-      tx.queueMutations(
-        // This is an upsert
-        inserts: [newAdvisory],
-      );
+      final packages = await _lookupAffectedPackages(newAdvisory, tx);
```
sigurdm, line 99: What happens when there are too many affected packages?

Do we just never get a warning in the client?

Could we update the packages outside the transaction or rather in several smaller transactions?

app/lib/service/security_advisories/backend.dart
```diff
@@ -93,20 +95,73 @@ class SecurityAdvisoryBackend {
...
+      if (packages.length > 50) {
```
jonasfj, line 99: Just do `advisory.affectedPackages.length` before calling `_lookupAffectedPackages(newAdvisory, tx);`

otherwise, if there's 500 affected packages you'll try to load them all, and then conclude that you don't want to do anything.

app/lib/service/security_advisories/backend.dart
jonasfj: > Could we update the packages outside the transaction or rather in several smaller transactions?

Then it's not eventually consistent. We'd need more infrastructure to ensure evential consistency.

Let's hope we never get one that has more than 50 affected packages. And cross the bridge when that happens.

app/lib/service/security_advisories/backend.dart
szakarias: done

jonasfj (approved):

PR #7100, opened 2023-10-09
Use only popularity from search snapshot when serving search queries.

isoos:
As we have started populating the `PackageDocument` fields in the previous runtime version, it is safe to only use them, and remove the fallback.


sigurdm (approved):
LGTM

PR #7103, opened 2023-10-10
Reduce sort() in API search index processing.

isoos:
- follow-up to #7094
- deduplicated parsing (`_apiDocPkg`)
- limited per-package sorted list for API package pages (instead of full sort on the entire doc page matches)
- ~5-9% latency improvement on the latest "slow query" example set

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -184,7 +184,8 @@ class InMemoryPackageIndex {
       packageHits = packageHits.map((ps) {
         final apiPages = textResults.topApiPages[ps.package]
             // TODO: extract title for the page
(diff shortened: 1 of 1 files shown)


app/lib/search/mem_index.dart
```diff
@@ -184,7 +184,8 @@ class InMemoryPackageIndex {
       packageHits = packageHits.map((ps) {
         final apiPages = textResults.topApiPages[ps.package]
             // TODO: extract title for the page
```
szakarias, line 186: Nit: Do we want to elaborate(or make an issue?) on this todo since we are touching the surrounding code?

szakarias (approved):

PR #7108, opened 2023-10-11
Simpler likeScore calculation in search index.

isoos:
- removes _LikeTracker, as we no longer update the index
- calls like score updates on initialization if the scores haven't been updated yet (should happen in testing only)


szakarias (approved):
LGTM

PR #7109, opened 2023-10-11
Pre-sorted lists for likes, popularity and points.

isoos:
Reduces memory allocations and `sort()` calls when serving a query with such sort orders.


isoos:
Hm, something is off with some test results, not sure why they are affected, pulling this PR for now.

PR #7121, opened 2023-10-13
Backfilling automated publishing field.

isoos:


app/lib/package/backend.dart
```diff
@@ -1575,6 +1575,24 @@ class PackageBackend {
...
+  /// [Package.automatedPublishingField].
+  Future<void> backfillAutomatedPublishingField() async {
+    await for (final p in db.query<Package>().run()) {
+      // ignore: deprecated_member_use_from_same_package
+      if (p.automatedPublishingJson == null) {
+        continue;
+      }
+      await withRetryTransaction(db, (tx) async {
+        final pkg = await tx.lookupValue<Package>(p.key);
+        pkg.automatedPublishing = pkg.automatedPublishing;
```
sigurdm, line 1591: Shouldn't this be:
```suggestion
        pkg.automatedPublishing = pkg.automatedPublishingJson;
```

app/lib/package/backend.dart
```diff
@@ -1575,6 +1575,24 @@ class PackageBackend {
...
+
+  // ignore: deprecated_member_use_from_same_package
+  /// Migrates the deprecated [Package.automatedPublishingJson] JSON field
+  /// and makes sure that the only related value is in the
+  /// [Package.automatedPublishingField].
+  Future<void> backfillAutomatedPublishingField() async {
+    await for (final p in db.query<Package>().run()) {
+      // ignore: deprecated_member_use_from_same_package
+      if (p.automatedPublishingJson == null) {
+        continue;
```
sigurdm, line 1587: Can there be cases where both fields are non-null?
Should we avoid updating in that case?

app/lib/package/backend.dart
isoos: The `automatedPublishing` field has already a logic to access the json content if available, and that's the consolidated view the users are seeing.

app/lib/package/backend.dart
isoos: There may be such cases, however, we already ignore one of them. Better to consolidate on the field that is being used anyway.

app/lib/package/backend.dart
sigurdm: Ah! nice

sigurdm (approved):

jonasfj (approved):

PR #7124, opened 2023-10-16
Verify that index.json has all data needed for pub-data.json

isoos:
- Moving `DartdocIndex` and its processing to `pub_dartdoc[_data]`, to get a reasonable separation from the rest of the app.
- This proves only on a single example that the coverage and the index information can be reasonably created from the `index.json`, which means we have no further requirement from `dartdoc` to include for us.
- The next step could be to move this code into `pana` and run `dartdoc` as part of the analyis.


pkg/pub_dartdoc/test/kind_names_test.dart
```diff
@@ -0,0 +1,17 @@
...
+import 'package:pub_dartdoc_data/dartdoc_index.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('kind names', () {
+    final expected = <int, String>{};
+    for (final k in Kind.values) {
+      expected[k.index] = k.toString();
+    }
+    expect(kindNames, expected);
```
jonasfj, line 15: Is this supposed to test that `kindNames` as defined in `pkg/pub_dartdoc_data/lib/dartdoc_index.dart` matches `Kind.values` ?

Can we add a comment.

-------------------

The reason you don't want to use `Kind.values` directly, is that you are not sure that `k.toString()` will work in production?

jonasfj (approved):

pkg/pub_dartdoc/test/kind_names_test.dart
isoos: The reason was that this parsing code should not depend on `package:dartdoc` in the long term (if we want to call SDK-specific dartdoc and/or non-customized dartdoc), but we still need these values, and matching it here as a separate test seemed like a good in-between.

PR #7126, opened 2023-10-18
Link to and display score/log.txt

### LINKED ISSUE #7115: Expose task logs on score page.

isoos:
- Fixes #7115.
- Slight improvements over the experimental implementation (more checks, `text/plain` instead of `plain/text`).

### DIFF
--- a/app/lib/frontend/handlers/routes.dart
+++ b/app/lib/frontend/handlers/routes.dart
@@ -365,14 +365,6 @@ class PubSiteService {
   // **** Experimental task end-points
   // ****
 
-  @Route.get('/experimental/task-log/<package>/<version>/')
-  Future<Response> taskLog(
-      Request request, String package, String version) async {
-    checkPackageVersionParams(package, version);
-    return redirectResponse(
-        '/packages/$package/versions/$version/score/log.txt');
-  }
-
   @Route.get('/experimental/task-summary/<package>/<version>/')
   Future<Response> taskSummary(
       Request request, String package, String version) async {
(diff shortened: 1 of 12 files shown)


app/lib/frontend/handlers/routes.dart
```diff
@@ -360,20 +369,8 @@ class PubSiteService {
...
-    if (!await packageBackend.isPackageVisible(package) ||
-        (await packageBackend.lookupPackageVersion(package, version)) == null) {
-      return Response.notFound('no such package');
-    }
-
-    final log = await taskBackend.taskLog(package, version);
-    return Response.ok(
-      log ?? 'no log',
-      headers: {'content-type': 'plain/text'},
-    );
```
jonasfj, line 376: I propose that we just remove this end-point, or at that we at-least leave a TODO for it to be deleted in the future.

app/lib/frontend/templates/views/pkg/score_tab.dart
```diff
@@ -53,21 +54,31 @@ d.Node scoreTabNode({
...
+          ],
+        ),
+        _reportNode(report),
+        if (toolEnvInfo != null) toolEnvInfo,
+        d.p(
+          classes: ['analysis-info'],
+          children: [
+            d.text('You can check the '),
+            d.a(href: 'score/log.txt', text: 'analysis log'),
+            d.text(' for more details.'),
```
jonasfj, line 78: nit:

```suggestion
            d.text('Check '),
            d.a(href: 'log.txt', text: 'analysis log'),
            d.text(' for details.'),
```

I think it says that same, just with fewer words, right?

jonasfj (approved):

PR #7137, opened 2023-10-23
Pre-compute overallScore hit ordered list in search.

isoos:
- This change will help default pagination requests and also a slight benefit for text queries.
- Updated test scores to reflect that we shall return the unadjusted raw score, instead of the multiplication-optimized one.

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -25,7 +25,7 @@ class InMemoryPackageIndex {
   final TokenIndex _readmeIndex = TokenIndex();
   final TokenIndex _apiSymbolIndex = TokenIndex();
 
-  /// Adjusted score takes the overall score and transform
+  /// Adjusted score takes the overall score and transforms
   /// it linearly into the [0.4-1.0] range.
   final _adjustedOverallScores = <String, double>{};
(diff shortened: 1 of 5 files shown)


app/lib/search/mem_index.dart
```diff
@@ -24,6 +24,11 @@ class InMemoryPackageIndex {
   final TokenIndex _descrIndex = TokenIndex();
   final TokenIndex _readmeIndex = TokenIndex();
   final TokenIndex _apiSymbolIndex = TokenIndex();
+
+  /// Adjusted score takes the overall score and transform
```
szakarias, line 28: transform -> transforms

app/lib/search/mem_index.dart
```diff
@@ -201,18 +214,18 @@ class InMemoryPackageIndex {
     );
   }
 
-  Score _getOverallScore(Iterable<String> packages) {
-    final values = Map<String, double>.fromEntries(packages.map((package) {
-      final doc = _packages[package]!;
+  /// Updated the overall score both on [PackageDocument] and in the [_adjustedOverallScores] map.
```
szakarias, line 217: Updated -> update?

szakarias (approved):

PR #7147, opened 2023-10-29
Provide `id-token` to example workflow

### LINKED ISSUE #7146: Include id-token permission in "Example workflow"
Currently, when enabling automated publishing in `pub.dev` panel, the "example workflow" does not include the `id-token` permission.

Without this permission, the GitHub Action is going to fail.

The documentation is correct in https://dart.dev/tools/pub/automated-publishing, and we should make it correct here too.

image

mateusfccp:
Provide the `permission` configuration with `id-token: write` to the example workflow.

Fixes #7146.


isoos:
I think this is not strictly needed, see my comment in the linked issue. And the example that the UI gives should be the minimum requirement, and maybe link to more documentation. Maybe the website documentation could be better phrased with the `permissions` block though...

isoos (approved):

PR #7155, opened 2023-11-01
Redirect missing /documentation/ urls to the closest finished/analyzed version

isoos:
- redirect happens with best effort:
  - if the requested `version` has finished analysis, it is returned
  - first the earliest newer finished version is looked up `1.0.0 -> 1.0.2` before `1.1.0`.
  - if there is no such version, then the latest (earlier) is selected, e.g. `2.1.0 -> 1.4.0`.
- handling of `latest` urls is unchanged
- task status now keeps the latest finished task version entry, increasing the URL stability when a new version is published
- added test for each version + publishing a new version
- cleaning up TODO in `latestFinishedVersion`

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -1074,7 +1074,7 @@ class TaskBackend {
       for (final rt in acceptedRuntimeVersions) {
         final key = PackageState.createKey(_db, rt, package);
         final state = await dbService.lookupOrNull<PackageState>(key);
-        // skip states where the entry was created, but no analysis has not finished yet
+        // Skip states where the entry was created, but the analysis has not finished yet.
         if (state == null || state.hasNeverFinished) {
           continue;
         }
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -1063,7 +1063,7 @@ class TaskBackend {
 
   /// Returns the closest version of the [package] which has a finished analysis.
   ///
-  /// If [version] or new exists with finished analysis, it will be preferred, otherwise
+  /// If [version] or newer exists with finished analysis, it will be preferred, otherwise
   /// older versions may be considered too.
   ///
   /// Returns `null` if no such version exists.
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -1091,7 +1091,7 @@ class TaskBackend {
         final newerCandidates =
             candidates.where((e) => isNewer(semanticVersion, e)).toList();
         if (newerCandidates.isNotEmpty) {
-          // return the earliest finished that is newer than [version].
+          // Return the earliest finished that is newer than [version].
           return newerCandidates
               .reduce((a, b) => isNewer(a, b) ? a : b)
               .toString();
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -25,7 +25,7 @@ import 'package:pub_dev/shared/exceptions.dart';
 import 'package:pub_dev/shared/redis_cache.dart';
 import 'package:pub_dev/shared/storage.dart';
 import 'package:pub_dev/shared/utils.dart'
-    show canonicalizeVersion, VersionIterableExt;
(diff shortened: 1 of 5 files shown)


app/lib/task/backend.dart
```diff
@@ -1048,16 +1055,48 @@ class TaskBackend {
...
-          }
+      }
+      return '';
+    });
+    return (cachedValue == null || cachedValue.isEmpty) ? null : cachedValue;
+  }
+
+  /// Returns the closest version of the [package] which has a finished analysis.
+  ///
+  /// If [version] or new exists with finished analysis, it will be preferred, otherwise
```
sigurdm, line 1066:
```suggestion
  /// If [version] or newer exists with finished analysis, it will be preferred, otherwise
```

app/lib/task/backend.dart
```diff
@@ -1048,16 +1055,48 @@ class TaskBackend {
...
+  ///
+  /// Returns `null` if no such version exists.
+  Future<String?> closestFinishedVersion(String package, String version) async {
+    final cachedValue =
+        await cache.closestFinishedVersion(package, version).get(() async {
+      final semanticVersion = Version.parse(version);
+      for (final rt in acceptedRuntimeVersions) {
+        final key = PackageState.createKey(_db, rt, package);
+        final state = await dbService.lookupOrNull<PackageState>(key);
+        // skip states where the entry was created, but no analysis has not finished yet
```
sigurdm, line 1077:
```suggestion
        // Skip states where the entry was created, but the analysis has not finished yet.
```

app/lib/task/backend.dart
```diff
@@ -1048,16 +1055,48 @@ class TaskBackend {
...
+        if (candidates == null || candidates.isEmpty) {
+          continue;
+        }
+        if (candidates.contains(semanticVersion)) {
+          return version;
+        }
+        final newerCandidates =
+            candidates.where((e) => isNewer(semanticVersion, e)).toList();
+        if (newerCandidates.isNotEmpty) {
+          // return the earliest finished that is newer than [version].
```
sigurdm, line 1094: nit
```suggestion
          // Return the earliest finished that is newer than [version].
```

app/lib/task/backend.dart
```diff
@@ -1048,16 +1055,48 @@ class TaskBackend {
...
+          return newerCandidates
+              .reduce((a, b) => isNewer(a, b) ? a : b)
+              .toString();
         }
+        return candidates.latestVersion!.toString();
       }
       return '';
```
sigurdm, line 1101: Just to clarify my understanding: Given that you return empty string here representing "computed, but found no value" can `cachedValue` ever be `null` below?

app/lib/task/backend.dart
```diff
@@ -1048,16 +1055,48 @@ class TaskBackend {
...
+          continue;
+        }
+        final candidates = state.versions?.entries
+            .where((e) => e.value.finished)
+            .map((e) => Version.parse(e.key))
+            .toList();
+        if (candidates == null || candidates.isEmpty) {
+          continue;
+        }
+        if (candidates.contains(semanticVersion)) {
```
sigurdm, line 1088: I think the logic is correctly implemented. But I think you could compute the version in a single iteration through the list.
```pseudocode
var result = null;
for (final candidate in candidates) {
  if (result == null) {
    result = candidate;
  } else if (candidate > version) {
    if (result < version) {
      result = candidate;
    }
  } else {
    if (candidate > result) {
      result = candidate;
    }
  }
}
```

Not sure it's really better....

sigurdm (approved):
Just a few nits

szakarias (approved):

app/lib/task/backend.dart
isoos: Correct. `null` values won't be cached, triggering recompute.

app/lib/task/backend.dart
isoos: I don't mind either way. If the current code is fine, let's keep it.

Issue #7002, opened 2023-09-04
youtube backend should use shared cache


jonasfj:
Probably all `CachedValue` should be forbidden. We should be a simple pattern that stores data in a cache entry in redis.

PR #7161, opened 2023-11-02
Remove CachedValue and cleanup of SDK search indexes.

isoos:
- We no longer need periodic updates in the search index. There is a slight benefit of retrying and re-fetching after a few minutes, in case the initial HTTP GET was unsuccessful, but we can get more reliable results with an index.json fetched at the time of the container image is created. (to be decided later)
- We should migrate off of CachedValue (see #7002).
- Also removed the wrapper classes, we no longer need their complexity.

### DIFF
--- a/app/lib/search/dart_sdk_mem_index.dart
+++ b/app/lib/search/dart_sdk_mem_index.dart
@@ -35,7 +35,7 @@ SdkMemIndex? get dartSdkMemIndex =>
 /// a search index from it.
 ///
 /// Returns `null` when the loading of `index.json` failed, or when there
-/// was an error parsing the faile or building the index.
+/// was an error parsing the file or building the index.
 Future<SdkMemIndex?> createDartSdkMemIndex() async {
   try {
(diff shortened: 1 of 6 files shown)


app/lib/search/dart_sdk_mem_index.dart
```diff
@@ -23,47 +21,22 @@ const dartSdkLibraryWeights = <String, double>{
...
-Future<SdkMemIndex?> _createDartSdkMemIndex() async {
+/// The active Dart SDK in-memory index.
+SdkMemIndex? get dartSdkMemIndex =>
+    ss.lookup(#_dartSdkMemIndex) as SdkMemIndex?;
+
+/// Tries to load Dart SDK's dartdoc `index.json` and build
+/// a search index from it.
+///
+/// Returns `null` when the loading of `index.json` failed, or when there
+/// was an error parsing the faile or building the index.
```
szakarias, line 38: The comment is unfinished

szakarias (approved):
LGTM

PR #7166, opened 2023-11-06
Update x-ago timestamps on page load.

isoos:
- Replaces / re-implements #6941, to solve the issue of the proxied/cached pages.
- Notable differences from the original PR:
  - uses milliseconds-since-epoch as data format, to reduce the time required to parse the timestamp on the client,
  - checks the data-attribute input and parsing,
  - updates the text/title only when it changed.

Note: I have simulated the updates locally with increased frequency, and it was mostly okay. Maybe the "in the last hour" -> "1 hour ago", "days" -> "months", "months" -> "years" updates were a bit noticeable when I was looking closely, but it didn't feel like too much change on the UI. I think we should try it and if it turns to be bothering the users, we can still remove / reschedule it. (One idea there is to only update it after days of stale cache.)

### DIFF
--- a/app/lib/frontend/dom/dom.dart
+++ b/app/lib/frontend/dom/dom.dart
@@ -4,10 +4,11 @@
 
 import 'dart:convert';
 
+import 'package:_pub_shared/format/x_ago_format.dart';
 import 'package:clock/clock.dart';
 
 import '../../shared/markdown.dart';
-import '../../shared/utils.dart' show formatXAgo, shortDateFormat;
+import '../../shared/utils.dart' show shortDateFormat;
 
 final _attributeEscape = HtmlEscape(HtmlEscapeMode.attribute);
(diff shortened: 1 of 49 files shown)


sigurdm:
Just a thought - Could we use exponential backoff for the timer?

Such that we update after 5 minutes, then 10 minutes etc.

Given that we use less and less granularity, the updates also need to be less and less frequent...

isoos:
> Just a thought - Could we use exponential backoff for the timer?

Good point! Although I implemented with the an every-increasing 5-minutes, as exponential will eliminate it too soon in practice. Not sure how frequent it is to keep a pub page open (and active) for weeks though...

isoos:
After an offline discussion I've removed the timer-parts, let's see how the on-load update works first.

sigurdm (approved):

PR #7175, opened 2023-11-09
Use pana-integrated dartdoc instead of pkg/pub_dartdoc.

isoos:
- Upgraded to latest pana.
- Main changes are in `pkg/pub_worker`.
- Includes updated tests and golden files from #7167 (see last commit for specific changes).


isoos:
Looks like this also fixes the flaky test outcomes in `test/task/end2end_test.dart`, with the root cause of `pkg/pub_dartdoc` inconsistent documentation output for `values` property in enums.

szakarias (approved):
LGTM

PR #7176, opened 2023-11-10
Off-request async task queue for local deferred tasks.

isoos:
- Handles post-upload deferred tasks and deferred cache purges.
- May also handle cached-value updates for `CachedValue` once their `Timer`-based approach is removed.


isoos:
Note: I'm not sure why the task/end2end test fails, but contrary to the previous flaky failures, this is now consistent, which is hope that we may find and fix a root cause of that flakiness. Investigating.

szakarias (approved):
LGTM

isoos:
Note: I think the root cause of the flaky tests are somewhere inside our `pkg/pub_dartdoc`: it seems that it sometimes generates documentation for the `enum`'s `values` property, sometimes it does not. I haven't see this fluctuation with the `dartdoc`-integrated `pana` yet, but it may be there too: #7175

PR #7181, opened 2023-11-14
Include advisoriesUpdated timestamp in listVersions result

szakarias:


app/lib/package/backend.dart
```diff
@@ -757,6 +757,7 @@ class PackageBackend {
       replacedBy: pkg.replacedBy,
       latest: latest.toApiVersionInfo(),
       versions: packageVersions.map((pv) => pv.toApiVersionInfo()).toList(),
+      advisoriesUpdated: pkg.latestAdvisory,
```
isoos, line 760: nit: should we make `advisoriesUpdated` `final`?

isoos (approved):

app/lib/package/backend.dart
szakarias: done

PR #7182, opened 2023-11-14
Hide invalid popularity scores.

isoos:


app/lib/shared/popularity_storage.dart
```diff
@@ -131,6 +142,13 @@ class _PopularityData {
 
   String get dateRange =>
       '${first?.toIso8601String()} - ${last?.toIso8601String()}';
+
+  /// Indicates that the data has no time range, or the range that is not
+  /// covering a full day, or that is more than a month old.
```
szakarias, line 147: Update comment: Indicates that the data has no time range, or that the range is not covering a full day, or that it is more than a month old.

szakarias (approved):
LGTM

PR #7184, opened 2023-11-15
Reduce integration test races with lockfile.

isoos:


isoos:
This revealed the following error:
```
Exception: dart pub get failed in app
  Resolving dependencies...
  
  Deletion failed, path = '/home/runner/.pub-cache/hosted/pub.dev/build_runner-2.4.6' (OS Error: No such file or directory, errno = 2)
```
It is possible that this is a result of race conditions in concurrent test execution.

szakarias (approved):

PR #7211, opened 2023-11-21
Upgrade package:puppeteer in pkg/pub_integration

### LINKED ISSUE #7199: upgrade package:puppeteer in pkg/pub_integration
While upgrading the SDK in #7197, it turns out our integration test fails with newer puppeteer+chrome (it may not be related to the current change, we haven't upgraded puppeteer for a while now). We should investigate and upgrade when possible.

isoos:
- Fixes #7199.
- Apparently the issue was local to my environment, I needed to upgrade my chrome instance.
- Also added a fix to reduce CI flakes by waiting for layout to be updated (different) for the specified elements.

### DIFF
--- a/pkg/pub_integration/lib/src/pub_puppeteer_helpers.dart
+++ b/pkg/pub_integration/lib/src/pub_puppeteer_helpers.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 import 'dart:async';
+import 'dart:io';
 
 import 'package:puppeteer/puppeteer.dart';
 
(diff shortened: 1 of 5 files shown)


pkg/pub_integration/lib/src/pub_puppeteer_helpers.dart
```diff
@@ -187,6 +192,39 @@ extension PubPageExt on Page {
     await _waitForModelHidden();
   }
 
+  /// Waits until the given selectors are visible and their UI top-left position
```
sigurdm, line 195: Sounds completely crazy that we have to make a hack like this. Should this not be built into the framework somehow?

sigurdm (approved):

PR #7217, opened 2023-11-23
Use redis cache for announcement and YouTube API response.

isoos:
- part of #7002
- `top_packages.dart` already uses search client's caching of responses
- popularity storage doesn't need redis-based caching
- announcement cache is invalidated on updating the value
- added YouTube redis cache for the API response object

### DIFF
--- a/app/lib/search/top_packages.dart
+++ b/app/lib/search/top_packages.dart
@@ -118,6 +118,7 @@ CachedValue<List<PackageView>> _cachedValue(
 }) {
   return CachedValue<List<PackageView>>(
     name: id,
+    // the redis-cached search results have 5 minutes TTL
     interval: Duration(minutes: 15),
     maxAge: Duration(hours: 24),
     updateFn: () async {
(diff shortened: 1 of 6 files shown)


app/lib/search/top_packages.dart
```diff
@@ -118,6 +118,7 @@ CachedValue<List<PackageView>> _cachedValue(
 }) {
   return CachedValue<List<PackageView>>(
     name: id,
+    // the redis-cached search results have 5 minutes TTL
```
sigurdm, line 121: How does the 5 minutes relate to the 15 minutes interval?

isoos: If we have a valid value locally, we don't initiate new search queries for the top packages for 15 minutes, and after that we have a good chance that we have a <5 minutes old value in the cache (we have 40-72 isolates asking for the  same data).

We could reduce the 15 minutes to 5 minutes, but in practice doesn't add more value, as the search index itself won't get updated only at every 15 minutes...

sigurdm: Could you put these kind of arguments in the comments?

It currently reads a bit funny without the context:

```
// the redis-cached search results have 5 minutes TTL
interval: Duration(minutes: 15), // whyyy?
```

sigurdm (approved):

PR #7239, opened 2023-12-01
Persistent connection for email sending.

isoos:


app/lib/frontend/email_sender.dart
```diff
@@ -168,6 +174,30 @@ class _GmailSmtpRelay implements EmailSender {
...
+            PersistentConnection(
+              await _getSmtpServer(),
+              timeout: Duration(seconds: 15),
+            ),
+          );
+        },
+      );
+    }
+    return _connection!;
+  }
```
jonasfj, line 199: Always return early, and do `return _connection = ` then you won't need null assertions :D
```suggestion
  Future<_GmailConnection> _getConnection() async {
    final old = _connection == null ? null : await _connection;
    if (old != null && !old.isExpired) {
      return old;
    }
    return _connection = Future.microtask(() async {
      if (old != null) {
        try {
          await old.connection.close();
        } catch (e, st) {
          _logger.warning('Unable to close SMTP connection.', e, st);
        }
      }
      return _GmailConnection(
        PersistentConnection(
          await _getSmtpServer(),
          timeout: Duration(seconds: 15),
        ),
      );
    });
  }
```

app/lib/frontend/email_sender.dart
```diff
@@ -160,6 +165,7 @@ class _GmailSmtpRelay implements EmailSender {
     } on SmtpClientAuthenticationException catch (e, st) {
       _logger.shout('Sending email failed due to invalid auth: $e', e, st);
       _backoffUntil = clock.now().add(Duration(minutes: 2));
+      _connection = null;
```
jonasfj, line 168: That doesn't close the underlying TCP connection, in fact we leak it

app/lib/frontend/email_sender.dart
```diff
@@ -141,11 +143,14 @@ class _GmailSmtpRelay implements EmailSender {
...
       await retry(
-        () async => send(
-          _toMessage(message),
-          await _getSmtpServer(),
-          timeout: Duration(seconds: 15),
-        ),
+        () async {
+          final c = await _getConnection();
+          try {
+            await c.connection.send(_toMessage(message));
```
jonasfj, line 149: We might not want to reuse the connection if we get a `TimeoutException` or `IOException` or `SmtpClientAuthenticationException` or `MailerException` (not sure about the last one)

isoos: Updated with a `_forceReconnect` flag to always stop the current connection and create a new one when there was any issue with it. It may be that we could do better and sometimes continue using it, but I think it is better to keep it simple.

isoos:
Note: updated PR with multiple connections for different senders.

jonasfj (approved):

PR #7240, opened 2023-12-04
Better accessibility for youtube links.

### LINKED ISSUE #7224: Youtube video links on front page should be more accessible
The links should turn "active" (red play icon) when receiving keyboard focus, not only on mouse-over. (Or we drop the active/inactive distinction)

If we can hide the play icon and the "state" from the accessibility tree we should, it doesn't add anything to a visually impaired user in terms of navigation. Perhaps it can be implemented in terms of css somehow...

isoos:
- Fixes #7224.
- overlay images are now decorative
- uses `:focus-within` to handle tab focus the same way as hovering


szakarias (approved):
Coolio

PR #7248, opened 2023-12-04
Code cleanup: removing obsolete code.

isoos:
- `ScoreCard` is not used anymore as Datastore entity.
- `randomizeStream` is not used since job-processing was removed.


sigurdm (approved):
LGTM

Do we need to do any clean-up of existing ScoreCard entities?

PR #7249, opened 2023-12-04
x-ago rendering: updated aria label and role

### LINKED ISSUE #7247: The x-ago link is hard to use with a screen reader
It reads 'switch between...' . It would perhaps be easier if it just read both the 'x ago' and the conrete date. 

image

image

isoos:
- Fixes #7247.
- Added a separate `role="button"` attribute to it, not sure if it helps but the computed attribute said `role: link`.


sigurdm:
Can we get this up on staging to check it out?

isoos:
> Can we get this up on staging to check it out?

Deployed to staging.

sigurdm (approved):

PR #7254, opened 2023-12-05
Accessible sort control.

### LINKED ISSUE #7243: Sort by control on search result page should be reachable with keyboard focus
And the drop-down should be usable with keyboard-only interactions.

isoos:
- Fixes #7243
- Switching `div`s to `button`s, adjusting CSS styles, and also using the `&:focus-within` selector.

### DIFF
--- a/app/lib/frontend/templates/views/shared/sort_control.dart
+++ b/app/lib/frontend/templates/views/shared/sort_control.dart
@@ -17,13 +17,16 @@ d.Node sortControlNode({
         attributes: {'title': selected.tooltip},
         children: [
           d.text('Sort by '),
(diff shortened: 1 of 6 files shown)


isoos:
Note: I've deployed this to staging.

szakarias:
It does not quite have the behaviour I would expect. Reaching the dropdown, I would want to navigate with arrows, not continuing with tabs. Maybe it can be done with an HTML select?

szakarias:
Let's merge this, but keep the issue to track the enhanced behaviour.

szakarias (approved):
Let's merge this, but keep the issue to track the enhanced behaviour.

PR #7255, opened 2023-12-05
Update the accessibility label of icons.

### LINKED ISSUE #7225: Alt-text for elements should not include the element type
Example: "verified publisher icon", it is enough to say "verified publisher".

isoos:
Fixes #7225

### DIFF
--- a/app/lib/frontend/templates/views/pkg/package_list.dart
+++ b/app/lib/frontend/templates/views/pkg/package_list.dart
@@ -228,7 +228,7 @@ d.Node _item({
                   image: d.Image(
                     src:
                         staticUrls.getAssetUrl('/static/img/schedule-icon.svg'),
-                    alt: 'icon indicating recent time',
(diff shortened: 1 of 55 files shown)


app/lib/frontend/templates/views/pkg/package_list.dart
```diff
@@ -228,7 +228,7 @@ d.Node _item({
                   image: d.Image(
                     src:
                         staticUrls.getAssetUrl('/static/img/schedule-icon.svg'),
-                    alt: 'icon indicating recent time',
+                    alt: 'new package',
```
sigurdm, line 231: Maybe:
```suggestion
                    alt: 'recently uploaded package',
```

isoos: I think "recently created package" may be slightly better, otherwise it could be misread with "recently updated package". Updating...

Issue #7263, opened 2023-12-06
Canonicalize topics using a YAML file


jonasfj:
It would be nice if we could merge topics that are effectively the same. For example, some packages are tagged `widget` while others are tagged `widgets`.

I propose that we make a `canonical-topics.yaml` file somewhere in the repository, and the simply map topics from aliases to canonical topic (when parsing the `topics` section in `pubspec.yaml`).

```yaml
# canonical-topics.yaml
topics:
- topic: widget
  # "description" serves as documentation for next time we need to review a PR
  # proposing that we add another alias
  # This can further be used in:
  #  * https://pub.dev/topics
  #  * tooltips when displaying a topic-link on pub.dev
  description: |
    Packages that contain Flutter widgets.
  # "aliases" is a list topics that should be replaced by "widget"
  aliases:
  - widgets
  - flutter-widget
```

@szakarias, @sigurdm, @isoos, thoughts?

### Pull-request to `canonical-topics.yaml`

The file itself should contain a comment explaining that pull-requests to this file are encouraged.
We'll review them as they come in.

Obviously, it'll easily take a week or two from a PR is merged until we make a deployment and the changes takes effect. But this should be acceptable.

Similarly, I don't think we should be concerned about the PR volume.
There are entire security advisory databases managed using PRs, and hopefully it's not every topic that will have aliases -- though I could imagine some people would enjoy writing descriptions for topics :D

### Name of the file `canonical-topics.yaml`

We could call this file something else.
A topic isn't more or less canonical because it's listed in this file.

Package authors shouldn't be preferring to use topics listed in this file.

The purpose of the file is really just to merge topics that mean the exact same thing.

Making up your own topic for your own framework or new thing that you're inventing is still a good thing to do.

Perhaps we should called it `merged-topics.yaml` or even simpler just `topics.yaml`. As we could even allow pull-request that have no other purpose than adding a description for the topic.

Description could be pretty awesome to have on `https://pub.dev/topics` and would also be great to have in tooltips when display a topic on pub.dev.

### Location of `canonical-topics.yaml`

I'm not entirely sure where we should put this file. Ideally somewhere it can live for a long time. This could be `dart-lang/pub/doc`, or we could just put it in `dart-lang/pub-dev/doc` or `dart-lang/pub-dev/config` or ?

### Creating a `dart.dev/go/canonical-topics.yaml`

I'd like to further propose that we make short-link: `dart.dev/go/canonical-topics.yaml`.
[...]

PR #7265, opened 2023-12-07
Introducing topics.yaml

jonasfj:
See https://github.com/dart-lang/pub-dev/issues/7263

### DIFF
--- a/app/lib/shared/count_topics.dart
+++ b/app/lib/shared/count_topics.dart
@@ -142,14 +142,4 @@ extension on YamlMap {
     }
     return value;
   }
-
-  Iterable<String> expectListOfString(String key) sync* {
-    for (final entry in expectList(key).nodes) {
-      if (entry.value case final String v) {
-        yield v;
-      } else {
-        throw SourceSpanFormatException('expected a string', entry.span);
-      }
-    }
-  }
 }
--- a/app/lib/shared/count_topics.dart
+++ b/app/lib/shared/count_topics.dart
@@ -2,15 +2,27 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:collection';
+import 'dart:io';
+
+import 'package:collection/collection.dart';
 import 'package:gcloud/storage.dart';
+import 'package:logging/logging.dart';
+import 'package:path/path.dart' as p;
 import 'package:pub_dev/package/backend.dart';
 import 'package:pub_dev/package/models.dart';
 import 'package:pub_dev/shared/configuration.dart';
 import 'package:pub_dev/shared/datastore.dart';
 import 'package:pub_dev/shared/storage.dart';
 import 'package:pub_dev/shared/utils.dart';
+import 'package:source_span/source_span.dart';
+import 'package:yaml/yaml.dart';
+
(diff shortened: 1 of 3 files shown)


doc/topics.yaml
```diff
@@ -0,0 +1,68 @@
...
+# However, if mistakes are made these changes are reversible.
+# And we should not be afraid of accepting that sometimes a single topic can
+# have multiple meaning, even if this makes the topic hard to use.
+#
+# The editorial guidelines are intended to evolve as we gain more experience
+# merging/managing topics.
+topics:
+- topic: widget
+  description: >-
+    Packages that contain Flutter widgets.
```
isoos, line 66: Nit: I think it may be better to keep these on a single line, mostly for formatting, but also to encourage some limit on the length of it.

doc/topics.yaml
```diff
@@ -0,0 +1,64 @@
...
+# ```
+#
+#
+# Contributing
+# ------------
+#
+# You are welcome to submit pull-request with additional aliases, canonical
+# topics and/or improved topic descriptions. 
+#
+# Please do limit pull-request to propose one topic!
```
sigurdm, line 46:
```suggestion
# Please limit pull-requests to propose one topic per PR!
```

doc/topics.yaml
```diff
@@ -0,0 +1,64 @@
...
+#   aliases:
+#   - <aliases-topic>
+#   - ...
+# ```
+#
+#
+# Contributing
+# ------------
+#
+# You are welcome to submit pull-request with additional aliases, canonical
```
sigurdm, line 43:
```suggestion
# You are welcome to submit pull-requests with additional aliases, canonical
```

sigurdm (approved):
We should probably link to this file from documentation somewhere.

Maybe even from https://pub.dev/topics ...

app/lib/shared/count_topics.dart
```diff
@@ -32,3 +44,112 @@ Future<void> countTopics() async {
...
+    _log.shout('failed to load doc/topics.yaml', e, st);
+
+    // This is sad, but we can just ignore it!
+    return UnmodifiableListView(<CanonicalTopic>[]);
+  }
+}();
+
+/// True, if [topic] is formatted like a valid topic.
+bool isValidTopicFormat(String topic) =>
+    RegExp(r'^[a-z0-9-]{2,32}$').hasMatch(topic) &&
```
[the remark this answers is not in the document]
jonasfj: I think it lives in pkg/pub_package_reader, I think this is so trivial I just duplicated it rather than add another dependency.

We can change it if we want to. This is really just to test the configuration file.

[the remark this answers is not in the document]
jonasfj: It has a `Iterable<ArchiveIssue> checkTopics(String pubspecContent)` function, which IMO should be private (or in `src/`) because the only thing `pub_package_reader` should expose is `Future<PackageSummary> summarizePackageArchive(...)`.

So even if there was a good method we did export that we could use, I think it would be wrong to do so. `pub_package_reader` is for reading packages and validating them.

One could argue we should have a separate package or library somewhere called `topic_validation.dart` -- but we don't, and I'm not sure where it should live :D

szakarias (approved):
LGTM

Issue #7270, opened 2023-12-08
Use separate config directories for separate analysis SDKs.


isoos:
- Original report: https://github.com/dart-lang/pub-dev/issues/7264#issuecomment-1845926093
- We need at minimum a separate config directory for preview SDKs, and potentially a separate directory for each SDKs. This is a breaking change in how we configure and run `pana`.

PR #7271, opened 2023-12-08
Use separate config directory for preview SDKs.

isoos:
- Should handle #7270 better.
- This is not the final fix, only a workaround, tracking the proper fix (pana changes) in #7270.

### DIFF
--- a/Dockerfile.worker
+++ b/Dockerfile.worker
@@ -17,18 +17,24 @@ USER worker:2000
 COPY --chown=worker:worker . /home/worker/pub-dev
 WORKDIR /home/worker/pub-dev
 
+# A config directory for preview SDKs.
+# TODO(https://github.com/dart-lang/pub-dev/issues/7270): Use per-SDK config directory.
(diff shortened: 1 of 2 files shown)


sigurdm:
I guess this will work, though we need to test it manually first.

It might be a simpler fix to simply let pana filter the output from `flutter --version --machine` to exclude any line starting with:
 "╔", " ║" or " ╚" (see https://github.com/flutter/flutter/blob/487bd69078058629f5b6fa8ecbc411f0d27ad894/packages/flutter_tools/lib/src/reporting/first_run.dart#L13)

isoos:
> I guess this will work, though we need to test it manually first.

I was able to confirm that the new setting within the docker did not produce the error that we were seeing before.

> It might be a simpler fix to simply let pana filter the output from flutter --version --machine to exclude any line starting with:
"╔", " ║" or " ╚" (see

It could be too. I think with this workaround in place I'd rather focus the pana changes to be more long-term fixes.

PR #7288, opened 2023-12-11
Throw exception when isolate exits without renewal indication.

isoos:


app/lib/service/entrypoint/_isolate.dart
```diff
@@ -119,6 +119,9 @@ class IsolateRunner {
     _isolates.add(isolate);
     unawaited(isolate.done.then((_) async {
       _isolates.remove(isolate);
+      if (!_closing && !isolate.markedForRenew) {
+        throw Exception('$id exited unexpectedly.');
```
sigurdm, line 123: Should we indicate the value of `_closing` and `isolate.markedForRenew` as there is two different ways we can get here?

sigurdm (approved):

PR #7301, opened 2023-12-13
End2end test for dockerized pub_worker build and execution.

isoos:


isoos:
Note: the test seem to fail on GitHub when we are trying to run the build container. Not yet sure if it is a permission or environment problem, skipping it on CI, it is still worthwhile to run it locally.

isoos:
Note: this test seems to be ranging between 5-10 minutes on CI. I think it is worth it regardless.

sigurdm:
> Note: this test seems to be ranging between 5-10 minutes on CI. I think it is worth it regardless.

Can we make it run in its own check, such that it finishes in parallel with other tests?

isoos:
> Can we make it run in its own check, such that it finishes in parallel with other tests?

I think it is already run in parallel (through random chance as we don't have much tests in `pkg/pub_worker` to bundle with). And if this is too long, we should probably consider investing in (1) making the image building part fetch the SDKs concurrently, and (2) reuse the docker image between runs that don't touch the `pub_worker` directory.

sigurdm (approved):

Issue #7122, opened 2023-10-13
Redesign the packages pages to include installation cmd directly on the first page


ebenjs:
URL: https://pub.dev/

It will be nice to redesign the packages pages to include directly the installation command just like `npm` or `crates.io` do. Sometimes we just need to copy quickly this command and paste it into our terminal to install the package. We don't need to explicitly click on the install tab before being able to see the package installation command.

sigurdm:
We have the button here that copies what you need to paste into pubspec.yaml:
![image](https://github.com/dart-lang/pub-dev/assets/8613953/aa1a25be-b1d7-4520-986d-e9047cfb32b3)

Maybe we should also have a place for `dart pub add <packagename>`...

ebenjs:
"Maybe we should also have a place for dart pub add <packagename>"; yes this is what I'm talking about.

ebenjs:
@sigurdm Can you have a look #7305

PR #7305, opened 2023-12-14
Added one click install command feature

ebenjs:
This PR is targeting issue #7122 
---

We're used to having the ability to easily install packages by copying the install code and pasting it directly into the console. This is especially helpful when you need to quickly install a package. Popular platforms like https://www.npmjs.com/package/axios, https://crates.io/crates/http, and https://pypi.org/project/http1/ all employ this design.

Currently, there's something similar in the header section of the package page after the package name. However, this command only allows you to copy text that's intended to be pasted in the pubspec.yml file and manually run 'pub get install.'

Therefore, I suggest adding this feature as it would be a time-saving feature for all developers.

Here is a screenshot

Screenshot from 2023-12-14 16-28-44

<details>

</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,8 @@ build/
 key.json
 
 node_modules/
+
+# test profile files are also not needed
+
+/app/dev-data-file.jsonl
+test-profile.yml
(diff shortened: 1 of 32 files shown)


Levi-Lesches:
I like the concept -- a lot of packages know they have an `installation` tab but still include the command on the front page because it's what people will see first. But I think longer package names will be cut off if used on the sidebar. What about here? 

image

sigurdm:
We could have the copy button open a drop down, so you could copy either for the pubspec or the command line.

Something like

image

Issue #7308, opened 2023-12-15
Modals should "trap focus"


sigurdm:
Currently the keyboard focus runs through the background controls when using TAB while the following modals are up:

* screenshot carousel
* Confirmation dialog in admin tab

Not sure about the best technique for this. But know at least the "page setup" dialog in google docs achieves this.

isoos:
After a bit of investigation, I don't see any widespread standard that we could use, rather only the following two workarounds:
- catch keyboard and focus events, and if focus gets or would get outside of the modal dialog, we redirect it, preventing the default behavior
- disable all other focusable elements

After experimenting the later (and learning that you can disable `<link>` stylesheets in the `head` element), I think overall it is a more robust way than trying to fight the browser and prevent the default behavior. Cleaning up and sending a PR soon.

PR #7309, opened 2023-12-15
Trap modal focus by disabling all focusable elements outside of its content.

isoos:
#7308

### DIFF
--- a/pkg/web_app/lib/src/_dom_helper.dart
+++ b/pkg/web_app/lib/src/_dom_helper.dart
@@ -212,8 +212,10 @@ bool _isInsideContent(Element e, Element content) {
 /// called to revert its original state.
 Function _disableFocusability(Element e) {
   final isLink = e.tagName.toLowerCase() == 'a';
+  final hasTabindex = e.hasAttribute('tabindex');
   final attributesToSet = <String, String>{
-    if (isLink) 'tabindex': '-1' else 'disabled': 'disabled',
+    if (isLink || hasTabindex) 'tabindex': '-1',
+    if (!isLink) 'disabled': 'disabled',
     'aria-hidden': 'true',
   };
   final attributesToRestore =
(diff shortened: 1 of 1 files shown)


isoos:
Note: deploying to staging.

sigurdm:
Should also be enabled for the screenshot carousel.

If I try this on the admin page the drop-down selector "select a publisher" still gets focus

This approach seems very fragile (what if we forget a selector?) but I guess there is no native html functionality to support this directly...

sigurdm:
I think `:focus-within` can be used for determining which elements are inside the modal...
https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within
https://caniuse.com/?search=focus-within

isoos:
> This approach seems very fragile (what if we forget a selector?)

Worst case: we have a rogue focus somewhere. Commented on the issue the two main approaches that we could take, and I think this is the more graceful failure we could have. The other approach (catching and redirecting the focus) seems to be more brittle.

> If I try this on the admin page the drop-down selector "select a publisher" still gets focus

Good catch, my local test didn't have a publisher. Fixed the `tabindex` handling, as material select had some unexpected use of it.

> Should also be enabled for the screenshot carousel.

I wanted to address that separately, and also wanted to verify if this doesn't introduce something that is broken/fragile with integration test on CI.

isoos:
`:focus-within` is great for styling (we are using it at other places), but I don't see how it could be used for focus trapping, other than the listening on all focus events and rewriting the default focus behavior.

sigurdm:
> `:focus-within` is great for styling (we are using it at other places), but I don't see how it could be used for focus trapping, other than the listening on all focus events and rewriting the default focus behavior.

Yeah, I meant it only for the `_isInsideContent` calculation, not as a way to disable focus in general. :/

sigurdm:
> Worst case: we have a rogue focus somewhere. 

Yeah that's probably not too bad.

> Commented on the issue the two main approaches that we could take, and I think this is the more graceful failure we could have. The other approach (catching and redirecting the focus) seems to be more brittle.

I agree!

sigurdm (approved):

PR #7328, opened 2023-12-21
Trigger API exporter after a package is uploaded.

isoos:


app/lib/package/export_api_to_bucket.dart
```diff
@@ -178,6 +178,11 @@ class ApiExporter {
     await pool.close();
   }
 
+  /// Updates the API files after a package has been successfully uploaded.
+  Future<void> updateAfterPackageUpload(String package, String version) async {
```
jonasfj, line 182: nit: How about calling it `updatePackageVersion` or `trackPackageVersion`

Is there any reason this shouldn't be called from all places where we make changes to a version that might affect the exported API files?

Just saying, the "afterPackageUpload" part makes this API appear specific to package uploads. But isn't it the same API we'll use when `isRetracted` is modified?

If we want a method for each "concern", then maybe we should make it `notifyPackageVersionPublished`? And establish a convention that anyone calling `ApiExporter` to notify about changes, should call a `notify...` method.

I'm not sure we need to care about why we're being notified though. It's probably sufficient to have a single public `updatePackage` method, or maybe two methods one for package and one for package-version. But I suspect just `updatePackage` is fine.

jonasfj (approved):

app/lib/package/export_api_to_bucket.dart
isoos: Good point about the notification, we should probably replace the current random direct calls with some kind of hierarchical ownerships or notifications. Renamed to `updatePackageVersion` for now.

PR #7335, opened 2023-12-28
Single-step SDK install in docker container.

isoos:
- I think this is much better than the two-step setup approach. It also enables parallelization as we no longer use the shared parent directory for temporary files.
- Also updated `setup-dart.sh` with an option to download SDKs based on hashes (and use the release directory for versioned ones).


jonasfj (approved):

jonasfj:
This is MUCH nicer :D

PR #7336, opened 2023-12-29
Foldable widgets: added keyboard navigation, removed (broken) animation.

### LINKED ISSUE #7286: Cannot unfold scoring details with keyboard
They don't receive keyboard focus when using tab
image

isoos:
- Fixes #7286
- I think the animation was not working for a while now, not sure when it stopped working. Removing it, as it did trigger an unnecessary reflow, and it slightly messes with keyboard navigation. (I've tried `visibility: hidden` and `transform: scaleY` alternatives, but they were not working as expected or were breaking accessibility again).
- Added keyboard focus to the folding icon. It is somewhat larger on the score page, but really small on the search sidebar. It should be a follow-up to review the sizes and the outline area for them.

### DIFF
--- a/pkg/web_app/lib/src/foldable.dart
+++ b/pkg/web_app/lib/src/foldable.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:html';
 import 'dart:math' show min, max;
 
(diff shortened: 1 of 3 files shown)


pkg/web_app/lib/src/foldable.dart
```diff
@@ -65,30 +51,31 @@ void _setEventForFoldable() {
...
+
     // listen on trigger events
     h.onClick.listen((e) async {
-      // Toggle state.
-      final isActive = foldable.classes.toggle('-active');
-      await update(isActive);
+      e.preventDefault();
+      await toggle();
+    });
+    h.onKeyDown.where((e) => e.key == 'Enter').listen((e) async {
```
sigurdm, line 76: Perhaps also spacebar?

pkg/web_app/lib/src/foldable.dart
```diff
@@ -22,41 +23,26 @@ void _setEventForFoldable() {
...
-      // Reset content state as hidden.
-      content.style.maxHeight = '0px';
-
-      // Wait one animation frame before trigger the full height content.
-      await window.animationFrame;
-      content.style.maxHeight = '${contentHeight}px';
-
       num scrollDiff = 0;
       if (scrollContainer != null) {
+        // Wait one animation frame before measurements.
```
sigurdm, line 34: Not sure I understand all this scrolling logic...

Do we really need it?

sigurdm:
I'd like to see this on staging also

pkg/web_app/lib/src/foldable.dart
isoos: This is for the mobile view: when the navigation menu is at the bottom, and it is clicked, there is no indication that something has been displayed under the fold. Scrolling to it makes the new content visible.

pkg/web_app/lib/src/foldable.dart
isoos: I was wondering about this, whether we should add `Space` or not, and if we add it here, we should add it at other places too. Created a new issue to track it: https://github.com/dart-lang/pub-dev/issues/7341

pkg/web_app/lib/src/foldable.dart
sigurdm: Can we just use [scrollIntoView](https://caniuse.com/?search=scrollIntoView)?

isoos: I think we have tried it and it worked with full-page elements, but had issues with embedded-scroll of the navigation. It may be worth to re-evaluate it again, a lot of time has passed since.

isoos: https://github.com/dart-lang/pub-dev/issues/7342 to track it separately

sigurdm (approved):

PR #7345, opened 2024-01-02
Slightly simpler scrollTo logic for foldable components.

### LINKED ISSUE #7342: Use scrollIntoView instead of custom scroll logic for foldable components.

isoos:
- Follow-up to #7336.
- Closes #7342.
- I was not satisfied with `scrollIntoView`'s behavior, but the multi-step scrolling can be simpler (or we may remove it entirely if we wanted to).

### DIFF
--- a/pkg/web_app/lib/src/foldable.dart
+++ b/pkg/web_app/lib/src/foldable.dart
@@ -29,7 +29,6 @@ void _setEventForFoldable() {
         return;
       }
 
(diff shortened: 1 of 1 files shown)


sigurdm:
Could you put this on staging?

isoos:
> Could you put this on staging?

Done.

sigurdm (approved):

PR #7348, opened 2024-01-04
Keyboard focusability for copy-to-clipboard icons.

### LINKED ISSUE #7285: No keyboard focus on copy buttons
The copy icons cannot be reached without using a mouse

image

isoos:
Fixes #7285


sigurdm (approved):

sigurdm (commented):
Would it work to use a `<button>` element wrapping the image to get some of this behaviour for free?

isoos:
> Would it work to use a <button> element wrapping the image to get some of this behaviour for free?

I think it is likely that we'd get it for free, but we'd need to create/adjust the button+image CSS styles. It would be worth if we do that for all the clickable icons on the site (e.g. folding icons).

PR #7349, opened 2024-01-04
Red ribbon for the staging site.

### LINKED ISSUE #7347: Staging should have visual distinction
It is too easy to be confused when on the staging site.

We could add a corner banner similar to the flutter debug one...

isoos:
Fixes #7347.
image

### DIFF
--- a/app/lib/dartdoc/dartdoc_page.dart
+++ b/app/lib/dartdoc/dartdoc_page.dart
@@ -7,6 +7,7 @@ import 'package:collection/collection.dart';
 import 'package:path/path.dart' as p;
 import 'package:pub_dev/frontend/dom/dom.dart' as d;
 import 'package:pub_dev/frontend/static_files.dart';
+import 'package:pub_dev/shared/configuration.dart';
 
 import '../shared/urls.dart';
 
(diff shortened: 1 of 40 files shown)


isoos:
Note: deployed to staging, should be online in ~20 minutes.

sigurdm:
I love it. Only nit is that it makes it hard to click the avatar. Is there a way to make it invisible to mouse-events?

sigurdm (approved):

isoos:
> Only nit is that it makes it hard to click the avatar.

I think this is more of a feature than a bug. It makes it more apparent that there is something with this site that is not normal.

sigurdm:
perhaps [pointer-events](https://caniuse.com/?search=pointer-events): none

sigurdm:
> I think this is more of a feature than a bug. 

Well it makes the staging site hard to use, not a great feature... I think the visible impression is clear enough

sigurdm (approved):

isoos:
> perhaps pointer-events: none

Yeah, this helped to fix it. (though it still selects the text, but that's a small issue)

PR #7358, opened 2024-01-08
Fix redirect URL for flutter_web_plugins.

### LINKED ISSUE #7357: `flutter_web_plugins` redirect to 404
Typing the following url: https://pub.dev/packages/flutter_web_plugins

redirects to: https://api.flutter.dev/flutter/package-flutter_web_plugins_flutter_web_plugins/package-flutter_web_plugins_flutter_web_plugins-library.html

which gives the following error:
Screen Shot 2024-01-08 at 10 30 06 AM

----
Please let me know if this should be filed elsewhere.

isoos:
Fixes #7357.

### DIFF
--- a/app/lib/package/overrides.dart
+++ b/app/lib/package/overrides.dart
@@ -48,7 +48,7 @@ const redirectPackageUrls = <String, String>{
   'flutter_test':
       'https://api.flutter.dev/flutter/flutter_test/flutter_test-library.html',
   'flutter_web_plugins':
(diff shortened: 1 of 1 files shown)


sigurdm (approved):
I guess this is better than the 404, but i really think these package redirects are deeply confusing...

isoos:
> I guess this is better than the 404, but i really think these package redirects are deeply confusing...

I agree. We should probably review our redirects / overrides / reserved names every once in a while...

Issue #7330, opened 2023-12-21
Package name ... is too similar to another active package


kwado-tech:
**The Issue**
####
I previously published a couple of packages to pub.dev and trying to update these package (running `flutter pub publish`) resulted to the below error message. 

I tried re-authenticating pub.dev with the same credentials used in publishing the said packages a number of times but this resulted to the same error message each time.

```text
Package name `l10n_mapper_annotation` is too similar to another active package: `l10n_mapper_annotation` (https://pub.dev/packages/l10n_mapper_annotation).
```

This seems to be an issue as republication (running `flutter pub publish`) should update the said packages with the new specified version-number, updates and fixes.

Below are the listed packages
- https://pub.dev/packages/l10n_mapper_annotation
- https://pub.dev/packages/l10n_mapper_generator

isoos:
@kwado-tech: Does the problem still exist? I'm trying to recreate the `l10_mapper_annotation` moderation + publishing this locally, and I cannot reproduce the error message.

kwado-tech:
@isoos The problem still exists. I just re-authenticated pub with two of my package publisher-credentials to publish updates to https://pub.dev/packages/l10n_mapper_annotation but got the same error massage.

I recently applied for the removal of a similar name package before publishing these. Probably this might be related or be of help?

isoos:
@kwado-tech: I was able to reproduce parts of the issue now, and not yet sure why this inconsistency was able to happen, but it looks like we have a tombstone conflict in the database. Did you have `l10n_mapper_annotation` published, deleted (by admins) and then re-published?

kwado-tech:
Hey @isoos, whats the progress on this issue? I see two related PRs merged but not sure what a conclusive resolve is.

isoos:
@kwado-tech: there will be a new site release this week, if it gets stable, one of the admins will remove/merge the Datastore entity that caused the issue.

kwado-tech:
Thanks @isoos I'll be on a lookout for this

sigurdm:
Should be fixed now. Please try again!

isoos:
Note: thinking about it, it may be that `nameTracker` caches do not detect the removal of the `ModeratedPackage` entity, and success may depend on whether the instance has been restarted since the removal.

kwado-tech:
Thanks @isoos but apparently, I still notice the same error indicated fix. Perhaps this is could be a resulting from `nameTracker` cache not invalidate/restarted?

[...]

PR #7359, opened 2024-01-08
Admin action to fix lingering ModeratedPackage entities.

isoos:
The tool is hopefully a one off to fix the integrity issues around #7330.


sigurdm (approved):

sigurdm:
This is good
Maybe we will at some point need a tool for just removing a tombstone (not merging it into existing package) - but let's take it at that point

PR #7360, opened 2024-01-08
Migrate delete-publisher to action framework

sigurdm:


app/test/shared/test_services.dart
```diff
@@ -133,6 +134,7 @@ class FakeAppengineEnv {
 
 /// Registers test with [name] and runs it in pkg/fake_gcloud's scope, populated
 /// with [testProfile] data.
+@isTest
```
isoos, line 137: what does this do?

PR #7361, opened 2024-01-08
Visible focus for like buttons during keyboard navigation.

### LINKED ISSUE #7294: Like button doesn't get visible keyboard focus
It is focused as part of the TAB-order, and can be toggled with space, but there is no visible indication that it has focus.

image

Also, after toggling the like-status by pressing space, it loses focus. It would be better if focus stayed on the same control.

isoos:
Fixes #7294.

image

### DIFF
--- a/pkg/web_css/lib/src/_detail_page.scss
+++ b/pkg/web_css/lib/src/_detail_page.scss
@@ -179,11 +179,17 @@ $detail-tabs-tablet-width: calc(100% - 240px);
     text-transform: uppercase;
     white-space: nowrap;
     display: flex;
-    margin-top: -2px;
+    margin-top: -4px;
+    border: 1px solid transparent;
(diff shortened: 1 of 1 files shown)


sigurdm:
I would much prefer this was implemented with a `<button>` with the system default focus (in my browser focus is usually blue and with slightly rounded corners, but this is black and square).

image

vs

image


Also it is a bit funky that the focus is around icon *and* text but with the mouse you can only click the icon...

(the button should probably be surrounding both the icon and the text...)

isoos:
> I would much prefer this was implemented with a <button>

Agreed, filed a follow-up tracking issue for this and similar components that need `<button>` instead of the current component.
https://github.com/dart-lang/pub-dev/issues/7362

sigurdm (approved):

Issue #7381, opened 2024-01-11
Store unique identifiers in AuditLogRecord.agentId


isoos:
```
service:gcp-service-account:<oauthClientId>
service:github-actions:<organization>/<repository
```

PR #7384, opened 2024-01-12
Expand service account agentIds.

isoos:
#7381

### DIFF
--- a/app/lib/account/agent.dart
+++ b/app/lib/account/agent.dart
@@ -88,7 +88,7 @@ abstract class AuthenticatedAgent {
 
 /// Holds the authenticated Github Action information.
 ///
-/// The [agentId] has the following format: `service:github-actions:organization/repository`
+/// The [agentId] has the following format: `service:github-actions:<organization>/<repository>`
 class AuthenticatedGithubAction implements AuthenticatedAgent {
   @override
   late final agentId = [
--- a/app/lib/account/agent.dart
+++ b/app/lib/account/agent.dart
@@ -18,6 +18,11 @@ abstract class KnownAgents {
   // Agent for pub admin actions.
   static const pubSupport = 'support@pub.dev';
 
+  static const _prefixValues = [
+    '$githubActions:',
+    '$gcpServiceAccount:',
+  ];
+
   static const _values = <String>{
     githubActions,
     gcpServiceAccount,
@@ -33,7 +38,8 @@ bool isValidUserId(String userId) {
 
 /// Whether the [agent] is a valid-looking identifier.
 bool isKnownServiceAgent(String agent) {
-  return KnownAgents._values.contains(agent);
+  return KnownAgents._values.contains(agent) ||
+      KnownAgents._prefixValues.any((prefix) => agent.startsWith(prefix));
 }
 
 /// Whether the [agent] is a valid-looking actor.
@@ -81,9 +87,14 @@ abstract class AuthenticatedAgent {
 }
 
 /// Holds the authenticated Github Action information.
+///
+/// The [agentId] has the following format: `service:github-actions:organization/repository`
 class AuthenticatedGithubAction implements AuthenticatedAgent {
   @override
-  String get agentId => KnownAgents.githubActions;
+  late final agentId = [
+    KnownAgents.githubActions,
+    if (payload.hasValidRepository) payload.repository,
+  ].join(':');
 
   @override
   String get displayId => KnownAgents.githubActions;
@@ -111,9 +122,14 @@ class AuthenticatedGithubAction implements AuthenticatedAgent {
 }
 
 /// Holds the authenticated Google Cloud Service account information.
+///
+/// The [agentId] has the following format: `service:gcp-service-account:user@examplec.com`
 class AuthenticatedGcpServiceAccount implements AuthenticatedAgent {
   @override
-  String get agentId => KnownAgents.gcpServiceAccount;
+  late final agentId = [
+    KnownAgents.gcpServiceAccount,
+    email,
+  ].join(':');
 
   @override
   String get displayId => email;
--- a/app/lib/account/agent.dart
+++ b/app/lib/account/agent.dart
@@ -37,6 +37,10 @@ abstract class KnownAgents {
   static const pubSupport = 'support@pub.dev';
 
   /// Returns an agentId in the format of `service:github-actions:<repositoryOwnerId>/<repositoryId>`
+  ///
+  /// Note: [repositoryOwnerId] and [repositoryId] are opaque identifiers.
+  /// These are **NOT** the same as the repository name, these identifiers
+  /// remain stable when organizations and repositories are renamed!
   static String githubActionsAgentId({
     required String repositoryOwnerId,
     required String repositoryId,
--- a/app/lib/account/agent.dart
+++ b/app/lib/account/agent.dart
@@ -30,18 +30,19 @@ abstract class KnownAgents {
   // Agent for pub admin actions.
   static const pubSupport = 'support@pub.dev';
 
-  /// Returns an agentId in the format of `service:github-actions:<org/repo>`
+  /// Returns an agentId in the format of `service:github-actions:<repositoryOwnerId>/<repositoryId>`
   static String githubActionsAgentId({
-    required String repository,
+    required String repositoryOwnerId,
+    required String repositoryId,
   }) {
-    return [genericGithubActions, repository].join(':');
+    return [genericGithubActions, '$repositoryOwnerId/$repositoryId'].join(':');
   }
 
-  /// Returns an agentId in the format of `service:gcp-service-account:<user@example.com>`
+  /// Returns an agentId in the format of `service:gcp-service-account:<oauthUserId>`
   static String gcpServiceAccountAgentId({
-    required String email,
+    required String oauthUserId,
   }) {
-    return [genericGcpServiceAccount, email].join(':');
+    return [genericGcpServiceAccount, oauthUserId].join(':');
   }
 
   static const _agentIdPrefixes = [
@@ -114,11 +115,13 @@ abstract class AuthenticatedAgent {
 
 /// Holds the authenticated Github Action information.
 ///
-/// The [agentId] has the following format: `service:github-actions:<organization>/<repository>`
+/// The [agentId] has the following format: `service:github-actions:<repositoryOwnerId>/<repositoryId>`
 class AuthenticatedGithubAction implements AuthenticatedAgent {
   @override
-  late final agentId =
-      KnownAgents.githubActionsAgentId(repository: payload.repository);
+  late final agentId = KnownAgents.githubActionsAgentId(
+    repositoryOwnerId: payload.repositoryOwnerId,
+    repositoryId: payload.repositoryId,
+  );
 
   @override
   String get displayId => KnownAgents.genericGithubActions;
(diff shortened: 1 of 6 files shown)


app/lib/account/agent.dart
```diff
@@ -81,9 +87,14 @@ abstract class AuthenticatedAgent {
...
 
 /// Holds the authenticated Github Action information.
+///
+/// The [agentId] has the following format: `service:github-actions:organization/repository`
 class AuthenticatedGithubAction implements AuthenticatedAgent {
   @override
-  String get agentId => KnownAgents.githubActions;
+  late final agentId = [
+    KnownAgents.githubActions,
+    if (payload.hasValidRepository) payload.repository,
```
jonasfj, line 96: There should always be a `repository` property, right?

Wouldn't it be better to throw an `ArgumentError` in the constructor instead.
This shouldn't happen, if it does aren't we fine with a 500 error.

We could perhaps check when creating a `AuthenticatedGithubAction` that a repository property is present. And refuse to make a `AuthenticatedGithubAction` instance, if not.

If we ever need to support actions that don't have a repository associated, then we can make a `AuthenticatedGithubAction` and a `AuthenticatedGithubRepositoryAction extends AuthenticatedGithubAction`. But we probably don't need both today.

In either case, producing the wrong value seems like a worse option, than return 500.
If it's because we don't know if the repository property is always there, then I think we should test whether it is. I suspect it is always there:
https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect

app/lib/account/agent.dart
```diff
@@ -18,6 +18,11 @@ abstract class KnownAgents {
   // Agent for pub admin actions.
```
jonasfj, line 18: Since:
```
  static const githubActions = 'service:github-actions';
  static const gcpServiceAccount = 'service:gcp-service-account';
```

are no longer just static values, but also prefixes. I think it's critically important that they each have documentation comment explaining.
 * Summary line what they are
 * Description of the form the take, e.g. `service:github-actions:<org>/<repository>`.
 * Description of the form they used to have `service:github-actions`.
 * If we support both forms for new entries, then we critically need to document when and why each of the two forms are used. As suggested further down, I propose we only allow the form `service:github-actions:<org>/<repository>` for new actions.

app/lib/account/agent.dart
```diff
@@ -33,7 +38,8 @@ bool isValidUserId(String userId) {
 
 /// Whether the [agent] is a valid-looking identifier.
 bool isKnownServiceAgent(String agent) {
```
jonasfj, line 40: What is a "KnownServiceAgent" I don't think I understand the concept anymore.

Should it instead be called "ServiceAgent" or `AutomatedServiceAgent` or ?

All this method really says is: "This is not a user".

app/lib/account/agent.dart
```diff
@@ -81,9 +87,14 @@ abstract class AuthenticatedAgent {
 }
 
 /// Holds the authenticated Github Action information.
+///
+/// The [agentId] has the following format: `service:github-actions:organization/repository`
```
jonasfj, line 91:
```suggestion
/// The [agentId] has the following format: `service:github-actions:<organization>/<repository>`
```
or similar :rofl:

app/lib/account/agent.dart
```diff
@@ -111,9 +122,14 @@ class AuthenticatedGithubAction implements AuthenticatedAgent {
 }
 
 /// Holds the authenticated Google Cloud Service account information.
+///
+/// The [agentId] has the following format: `service:gcp-service-account:user@examplec.com`
```
jonasfj, line 126:
```suggestion
/// The [agentId] has the following format: `service:gcp-service-account:<user@examplec.com>`
```

app/lib/account/agent.dart
```diff
@@ -111,9 +122,14 @@ class AuthenticatedGithubAction implements AuthenticatedAgent {
...
 class AuthenticatedGcpServiceAccount implements AuthenticatedAgent {
   @override
-  String get agentId => KnownAgents.gcpServiceAccount;
+  late final agentId = [
+    KnownAgents.gcpServiceAccount,
+    email,
+  ].join(':');
```
jonasfj, line 132: hmm, we could consider using `oauthUserId`.

It's less readable, and not very nice. So I'm inclined to also go with email.

We could also use `repository_id` in the github agent id. Then it would be the same identifier across renames, but it's also a bit crazy -- and impossible to read.

On the flip side: the UUID used for user_id isn't very readable either.

I guess I'm wondering what is it we want agentId to be?

--------------

We should probably also update the documentation `AuthenticatedAgent.agentId` regardless.

isoos: > We could also use repository_id in the github agent id. Then it would be the same identifier across renames, but it's also a bit crazy -- and impossible to read.

We are allowing the publishing based on the `<org>/<repo>` name (as stored in the package `AutomatedPublishing` object), so it may seem better to use that as the agent id. We could store the ids in the attached `data` map though

.> On the flip side: the UUID used for user_id isn't very readable either.

Yeah, but at least that is the only UUID we are using and the lookup to email is straightforward. So there is very little benefit of changing that now.

app/lib/account/agent.dart
isoos: This method is effectively only used in integrity checks: https://github.com/dart-lang/pub-dev/blob/master/app/lib/shared/integrity.dart#L692-L696

I think this is slightly more than "This is not a user", it also verifies that the `agentId` string on the audit log record is valid. We could remove it if you think it is just better to use `isValidUserId` in the above use in the integrity check, but I'd rather expand on the documentation instead.

app/lib/account/agent.dart
isoos: Yeah, I'm undecided on this too. It may be better to use a more opaque id, especially if we know that it won't change in the future. `repositoryId` may change: people may delete their repository and recreate it, using the same name, but resulting in a different id. It is less likely with the oauthUserId. So maybe use that here?

app/lib/account/agent.dart
isoos: renamed + added documentation. PTAL.

app/lib/account/agent.dart
isoos: Crazy idea; what about `<org>/<repo>/<repoId>` and similarly `<email>/<oauthUserId>`?

isoos:
@jonasfj as discussed offline, I've updated the PR to use oauthUserId and repository ids instead of the visible labels.

app/lib/account/agent.dart
```diff
@@ -12,41 +12,74 @@ import 'models.dart';
...
+///
+/// Older identifiers may only describe the type of the agent.
 abstract class KnownAgents {
-  static const githubActions = 'service:github-actions';
-  static const gcpServiceAccount = 'service:gcp-service-account';
+  /// Non-specific agent - only specifies it is from GitHub Actions.
+  static const genericGithubActions = 'service:github-actions';
+
+  /// Non-specific agent - only specifies it is from GCP Service Account.
+  static const genericGcpServiceAccount = 'service:gcp-service-account';
```
jonasfj, line 28:
```suggestion
  /// Non-specific agent - only specifies it is from GitHub Actions.
  ///  
  /// Deprecated and should not be used for new audit-log entries.
  /// This value is still present in some older audit-log entries.
  static const genericGithubActions = 'service:github-actions';

  /// Non-specific agent - only specifies it is from GCP Service Account.
  ///
  /// Deprecated and should not be used for new audit-log entries.
  /// This value is still present in some older audit-log entries.
  static const genericGcpServiceAccount = 'service:gcp-service-account';
```

app/lib/account/agent.dart
```diff
@@ -12,41 +12,74 @@ import 'models.dart';
...
 
-  static const _values = <String>{
-    githubActions,
-    gcpServiceAccount,
+  /// Returns an agentId in the format of `service:github-actions:<repositoryOwnerId>/<repositoryId>`
+  static String githubActionsAgentId({
+    required String repositoryOwnerId,
+    required String repositoryId,
+  }) {
+    return [genericGithubActions, '$repositoryOwnerId/$repositoryId'].join(':');
```
jonasfj, line 38: nit: feel free to ignore :D

```suggestion
    return '$_GithubActionsPrefix:$repositoryOwnerId/$repositoryId';
```

and then have:
```
/// Prefix for `agentId`s on the form:
/// `service:github-actions:<repositoryOwnerId>/<repositoryId>`
final _GithubActionsPrefix = 'service:github-actions'
```

because `genericGithubActions` indicates a generic agentId, not a prefix.
And for all intents and purposes `genericGithubActions`is a historical value, not a prefix.

--------

Or even simpler, just hardcode it:
```suggestion
    return 'service:github-actions:$repositoryOwnerId/$repositoryId';
```

app/lib/account/agent.dart
```diff
@@ -12,41 +12,74 @@ import 'models.dart';
...
+  static const genericGcpServiceAccount = 'service:gcp-service-account';
+
   // Agent for pub admin actions.
   static const pubSupport = 'support@pub.dev';
 
-  static const _values = <String>{
-    githubActions,
-    gcpServiceAccount,
+  /// Returns an agentId in the format of `service:github-actions:<repositoryOwnerId>/<repositoryId>`
+  static String githubActionsAgentId({
```
jonasfj, line 34:
```suggestion
  /// Returns an agentId in the format of `service:github-actions:<repositoryOwnerId>/<repositoryId>`
  ///
  /// Note: [repositoryOwnerId] and [repositoryId] are opaque identifiers.
  /// These are **NOT** the same as the repository name, these identifiers
  /// remain stable when organizations and repositories are renamed!
  static String githubActionsAgentId({
```

Just to reduce risk that someone plugs in the wrong value.

jonasfj (approved):

app/lib/account/agent.dart
isoos: Updated with the separate field names (I think it is better as we also need to repeat the prefix in another place).

app/lib/account/agent.dart
```diff
@@ -104,16 +158,31 @@ class AuthenticatedGithubAction implements AuthenticatedAgent {
...
 }
 
+void _checkRepository(String repository) {
+  InvalidInputException.check(repository.trim().isNotEmpty,
+      'The JWT from GitHub must have a non-empty `repository`.');
+  final parts = repository.split('/');
+  InvalidInputException.check(
+      parts.length == 2, 'The JWT from GitHub must have a valid `repository`.');
+  InvalidInputException.check(parts.every((e) => e.isNotEmpty),
+      'The JWT from GitHub must have a valid `repository`.');
```
jonasfj, line 176: This will produce a 400 error if they are not caught.

It's unclear who is responsible for ensuring that these properties are satisfied.

I think we have the following options:

**Option A:** The caller is responsible for ensuring that these properties are satisfied. If they are not the caller should have return 401 or 403, that's irrelevant to the discussion of `AuthenticatedGithubAction`, because as far as `AuthenticatedGithubAction` is concerned the caller should always check these properties are correct. If this is the behavior we want we should:
 * (A.1) Throw `ArgumentError`.
 * (A.2) Document the requirements on the input in the constructor for `AuthenticatedGithubAction`.

**Option B:** Whatever happen should never be able to happen. Then throw `AssertionError`. No need to check anything, or document it. But if this ever happens it's must be because some invariant is wrong and it implies a bug in our understanding of our own code.
 * (B.1) Throw `AssertionError`.

**Option C:** This function is part of the authentication logic and the security of the application depends on this logic. It seems a bit confusing that the constructor for an `AuthenticatedGithubAction` object contain elements critical to the authentication flow. If this is the case, then:
  * (C.1) Throw `AuthenticationException` which becomes a 401.

**Option D:** This function is part of the authorization logic, again this seems a bit confusing. Especially, since the constructor for `AuthenticatedGithubAction` doesn't have the context to know what the caller is trying to do (so how would it determine if the caller is authorized to do something). But if this was the case:
 * (D.1) Throw `AuthorizationException` which becomes a 403.

**Option E:** This function is actually parsing content from the JWT; thus, it might be appropriate to throw a `FormatException`.
 * (E.1) Throw `FormatException`.
 * (E.2) Document in the constructor that it throws a `FormatException` (all exceptions thrown should be documented, errors need not be)
 * (E.3) Catch the `FormatException` everywhere we call the constructor.

Throwing `InvalidInputException` causes a 400 error, I don't see how that makes sense. If this is due to invalid input, it is not invalid input on the part of the caller, but on the part of the JWT passed from Github.

I suspect that (A) or (B) or (E) is the correct choice.

jonasfj (approved):

PR #7399, opened 2024-01-16
Remove docs report joining from fake analyzer.

isoos:
This is an artifact from the past where we had separate pana and dartdoc report and needed to join them. Removing it will also allow removing further methods from the pana API.


sigurdm (approved):
Seems there's still some failing tests...

PR #7419, opened 2024-01-19
Updated dartdoc search.html rendering: redirects, title and noindex.

### LINKED ISSUE #7406: Page title should include search terms when searching on pub.dev/documentation

isoos:
- Fixes #7406.
- I think redirects are important so that keep the URL index clean. TBD: maybe we should also do it for other pages, e.g. `/documentation/package/latest/library.html?x=1` should redirect to `library.html`.
- As the results are rendered via JS from `dartdoc`, we should always `noindex` the pages.
- Also updated the page title.


isoos:
Note: pushed to staging.

sigurdm (approved):

PR #7420, opened 2024-01-19
Set ARIA role and label for header navigation.

### LINKED ISSUE #7415: The top bar should have the 'navigation' aria role
image

isoos:
Fixes #7415.

Note: we had a HTML validation test that stated: we shouldn't use `role` attribute on non-interactive components, this PR added and exemption for the site header `div`. I see a lot of examples on Mozilla's ARIA documentation pages where the `role` is on a `div`, so it seems to be fine.

### DIFF
--- a/app/lib/frontend/templates/views/shared/site_header.dart
+++ b/app/lib/frontend/templates/views/shared/site_header.dart
@@ -55,6 +55,10 @@ d.Node siteHeaderNode({
           'site-header-nav',
           'scroll-container',
         ],
+        attributes: {
+          'role': 'navigation',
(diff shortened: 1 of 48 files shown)


sigurdm:
> Note: we had a HTML validation test that stated: we shouldn't use role attribute on non-interactive components, 

Good point. I don't know how to align these requirements in general.

But for this case could we use the [`<nav>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) element instead of an explicit role.

isoos:
Closing in favor of #7464.

PR #7421, opened 2024-01-19
Advanced filter: is:plugin

### LINKED ISSUE #7403: Create "advanced" search checkbox for "is:plugin"
Context: https://github.com/dart-lang/pana/issues/640

isoos:
Fixes #7403.

### DIFF
--- a/app/lib/frontend/templates/views/pkg/index.dart
+++ b/app/lib/frontend/templates/views/pkg/index.dart
@@ -159,6 +159,13 @@ d.Node _searchFormContainer({
                 searchForm: searchForm,
                 title: 'Show only packages compatible with Dart 3.',
               ),
+              _tagBasedCheckbox(
+                tagPrefix: 'is',
+                tagValue: 'plugin',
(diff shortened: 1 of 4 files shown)


isoos:
Note: started to push to staging.

isoos:
> Seems there are no plugins found on staging: https://staging.pub.dev/packages?q=is%3Aplugin

It seems that the analysis on staging is lagging behind. Maybe we should really handpick a few packages that get some priority there.

> Also I think we need to somehow describe better how we define 'plugin'. It is not clear enough that this doesn't just mean "flutter packages"... Potentially we could link to https://docs.flutter.dev/packages-and-plugins/developing-packages#plugin

Not sure: how should we expose that information?

sigurdm:
> Not sure: how should we expose that information?

Not sure either. Maybe we call the filter "Flutter plugin packages". To distinguish it from other packages...

 @jonasfj any better ideas?

jonasfj (approved):

PR #7438, opened 2024-01-25
Verify production deployment with CLI callback-driven runtime parameters.

isoos:
Note: `withBrowserPage` is not yet implemented.


isoos:
Note: I've implemented the browser session + cookies too. I think this could be the CLI tool that we run to verify the deployment.

In this form it is not possible to use it for local/fake server, but if we remove the gmail-specific code and use only a callback for getting the last email, it may work locally too.

jonasfj (approved):
I see no harm in landing this, but we'll probably have to rework some of the callbacks when we implement them.

PR #7440, opened 2024-01-29
Escape relative path for dartdoc page urls.

isoos:
Also removes `omitTrailingSlash` as we are not using it, and its logic is entirely driven by the presence of `relativePath`.

### DIFF
--- a/app/lib/shared/urls.dart
+++ b/app/lib/shared/urls.dart
@@ -154,9 +154,15 @@ String pkgDocUrl(
   String? version,
   bool includeHost = false,
   String? relativePath,
-  bool omitTrailingSlash = false,
   bool isLatest = false,
 }) {
+  final parsedRelativePathSegments = relativePath == null
+      ? const <String>[]
+      : Uri.parse(relativePath)
+          .pathSegments
+          .where((e) => e.isNotEmpty)
+          .toList();
+
   if (isLatest || version == null) {
     version = 'latest';
   }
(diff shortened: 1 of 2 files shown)


sigurdm (approved):
Could you make a test that these indeed gets escaped?

app/lib/shared/urls.dart
```diff
@@ -154,30 +154,32 @@ String pkgDocUrl(
...
   ];
   final baseUri = includeHost ? _siteRootUri : _pathRootUri;
   String url = baseUri.resolveUri(Uri(pathSegments: segments)).toString();
 
-  if (relativePath != null) {
-    url = p.join(url, relativePath);
-  } else if (!omitTrailingSlash) {
-    url = '$url/';
-  }
   if (url.endsWith('/index.html')) {
```
sigurdm, line 178: Could this be done better by testing the last segment?

isoos: It could work with something like this:
```
    ...parsedRelativePathSegments
        // filter out the last segment if it is `index.html` 
        .whereIndexed((index, element) =>
            index != parsedRelativePathSegments.length - 1 ||
            element != 'index.html'),
```

However, I don't find it easier to read. Maybe `if (segments.isNotEmpty && segments.last == 'index.html`) segments.removeLast();` could work, but then it will make the trailing slash check a bit more complex. Which would you prefer?

sigurdm: I was thinking something like segments.removeLast();

Also the trailing-slash check could probably be done by checking if the last segment is empty.

But I'm not sure I know the perfect solution. Just all these string-based uri operations make me a bit worried.

isoos: Updated, PTAL.

app/lib/shared/urls.dart
```diff
@@ -154,32 +154,34 @@ String pkgDocUrl(
   String? version,
   bool includeHost = false,
   String? relativePath,
-  bool omitTrailingSlash = false,
   bool isLatest = false,
 }) {
+  final relativePathSegments = relativePath == null
+      ? const <String>[]
+      : Uri.parse(relativePath)
```
sigurdm, line 161: Are we sure relativePath is always relative? Otherwise we might get some weird results here.

sigurdm (approved):
Nice! LGTM

app/lib/shared/urls.dart
isoos: `relativePath` comes from the "rest" part of the requested URL parsing after `/documentation/<package>/<version>/<relativePath>`.

Issue #1987, opened 2019-01-24
tarballs should be served with 'content-disposition: attachment' from GCS


jonasfj:
Tarballs should always be served with:
```http
content-disposition: attachment
```

This is done by setting metadata in GCS:
https://cloud.google.com/storage/docs/metadata#mutable

----
There is no rush on this... This came out of the security review, but on a separate domain as we redirect to downloads.

PR #7444, opened 2024-01-30
Set content-disposition: attachment on objects in public bucket

sigurdm:
Fixes https://github.com/dart-lang/pub-dev/issues/1987


isoos (approved):
nice!

isoos (approved):
I think this is a good way to do it. We could apply the metadata inside the upload transaction, but it is also just fine to do it post-upload + backfill.

sigurdm:
Yeah, I would hate the transaction to fail because of the content-disposition not being set.

PR #7446, opened 2024-01-31
Use browser's history when displaying/hiding the infobox on mobile view.

### LINKED ISSUE #7428: Navigating back from metadata popup should go back to package-page
Instead it goes back to from before the package page.

isoos:
- Fixes #7428
- Unfortunately this increased the complexity of the view toggle a bit, but history API is not really friendly in this regard.

### DIFF
--- a/pkg/web_app/lib/src/mobile_nav.dart
+++ b/pkg/web_app/lib/src/mobile_nav.dart
@@ -35,6 +35,7 @@ void _setEventForDetailMetadataToggle() {
   int? origX, origY;
 
   var isVisible = false;
+  // ignore: cancel_subscriptions
   StreamSubscription? stateSubscription;
   final currentTitle = document.head?.querySelector('title')?.text?.trim();
   final currentUrl = window.location.toString();
--- a/pkg/web_app/lib/src/mobile_nav.dart
+++ b/pkg/web_app/lib/src/mobile_nav.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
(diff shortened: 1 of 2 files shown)


pkg/web_app/lib/src/mobile_nav.dart
```diff
@@ -33,21 +34,62 @@ void _setEventForDetailMetadataToggle() {
...
+        }, currentTitle ?? '', null);
       } else {
-        // restore scroll position
-        window.scrollTo(origX, origY);
-        origX = null;
+        window.history.back();
       }
+
+      // only listen to state events after the first initialization
+      stateSubscription ??= window.onPopState.listen((event) async {
```
jonasfj, line 81: This won't work well, if you:
 1. Navigate to `pub.dev/packages/retry`
 2. Toggle the "details / meta-data" view
 3. Click the repository link (navigating to `github.com/...`)
 4. Click the back button.
 5. The "details / meta-data" view will not be active, leaving users disappointed.

Am I wrong? (I haven't tested the theory)

We could say that (5) is an acceptable downside.

------------------

We could also use `pub.dev/packages/retry#details` to toggle the "details / meta-data" view. Storing the state in `document.location.hash`.

We might be able to use the [`hashchange` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/hashchange_event) then. Though it might be just as complicated.

------------------

An alternative is to create a separate page: `pub.dev/packages/<package>/details` that contains the sidebar. It'll be a bit sad, because it won't work when you resize the window.

isoos: I've deployed the PR to staging, and it seems to me that clicking through 1-5 is working as expected, at least on Chrome.

PR #7450, opened 2024-02-01
Disable email notification on package published.

isoos:


sigurdm (approved):

isoos:
The pub_worker docker test failure seems to be unrelated (happened at another PR too, one of the packages does not get max score).

sigurdm:
> The pub_worker docker test failure seems to be unrelated (happened at another PR too, one of the packages does not get max score).

Do we need to fix that before deployment?

Issue #6398, opened 2023-02-03
Redirect on non-canonical versions.


isoos:
The following URL is not a valid version:
https://pub.dev/packages/retry/versions/3.0.001

Instead of displaying the content for the canonical value, it should redirect to:
https://pub.dev/packages/retry/versions/3.0.1

PR #7457, opened 2024-02-02
Redirect matched mixed-case package names to lowercase names.

### LINKED ISSUE #7455: Redirect to case-normalized package page
We could let
pub.dev/packages/Retry redirect to pub.dev/packages/retry

isoos:
Fixes #7455 (99.9% of the cases)

### DIFF
--- a/app/lib/frontend/handlers/package.dart
+++ b/app/lib/frontend/handlers/package.dart
@@ -249,6 +249,19 @@ Future<shelf.Response> _handlePackagePage({
   if (redirectPackageUrls.containsKey(packageName)) {
     return redirectResponse(redirectPackageUrls[packageName]!);
   }
+  // May redirect to the lowercased package name, but only if [versionName] and
+  // [assetKind] is unspecified.
+  // NOTE: we have legitimate mixed-case package names that must not be redirected.
+  final lowerCasePackageName = packageName.toLowerCase();
+  if (versionName == null &&
+      packageName != lowerCasePackageName &&
+      assetKind == AssetKind.readme) {
+    if (!await packageBackend.isPackageVisible(packageName) &&
+        await packageBackend.isPackageVisible(lowerCasePackageName)) {
+      return redirectResponse(urls.pkgPageUrl(lowerCasePackageName));
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


sigurdm (commented):
LGTM

But why only for the main package-page and unspecified version?

app/lib/frontend/handlers/package.dart
```diff
@@ -249,6 +249,19 @@ Future<shelf.Response> _handlePackagePage({
...
     return redirectResponse(redirectPackageUrls[packageName]!);
   }
+  // May redirect to the lowercased package name, but only if [versionName] and
+  // [assetKind] is unspecified.
+  // NOTE: we have legitimate mixed-case package names that must not be redirected.
+  final lowerCasePackageName = packageName.toLowerCase();
+  if (versionName == null &&
+      packageName != lowerCasePackageName &&
+      assetKind == AssetKind.readme) {
+    if (!await packageBackend.isPackageVisible(packageName) &&
```
sigurdm, line 259: Is this lookup cheap?

isoos: Yes, it should be: a redis-cached boolean flag.

sigurdm: But it will mean an extra request in the common path - right?

Maybe better to put the check where we otherwise would have returned 404, so the cost is not paid in the common case.

isoos:
> But why only for the main package-page and unspecified version?

Simplicity: in this case the URL pattern is simple, we don't need to figure out where to redirect to. I've also started to look into #6398, but it turns out the generic case is rather complicated, I'd just implement the low-hanging items, and deal with the rest much later.

app/lib/frontend/handlers/package.dart
sigurdm: Ah - sorry you are right, I had not understood the full logic.

LGTM

isoos:
Let's put this on hold for now. I have an idea that could make the generic redirect (also with version canonicalization) work without too much additional complexity.

isoos:
Closing this in favor of #7460.

Issue #7177, opened 2023-11-11
Enable more automated workflows by allowing publishing from the `workflow_dispatch` trigger


spydon:
> Note: Pub.dev only allows automated publishing from GitHub Actions when the workflow is triggered by pushing a git tag to GitHub. Pub.dev rejects publishing from GitHub Actions triggered without a tag. This ensures that new versions cannot be published by events that should never trigger publishing.

I understand the reasoning here, but this also makes it impossible to do fully automated publishing since a tag that is created by another workflow isn't allowed to trigger other workflows. So with this restriction from pub.dev in place the publishing will always need manual intervention from the command-line/IDE, since a user has to push a tag manually to trigger the workflow.

What we would like is to trigger the workflow that publishes the new versions from another workflow by using the `workflow_dispatch` trigger (currently only `push` is allowed).

The simplest flow for our use-case would be this:

1. The users goes to their action tab and presses "Run release action"
2. The melos-action versions and creates changelogs for all packages, and then published a PR with the changes (it also checks if it gets any publish --dry-run warnings in this step)
3. The PR is then reviewed
4. When the PR is merged it triggers the melos publish job which will tag and publish all of newly versioned the packages.

For context this will for example be used by the [melos-action](https://github.com/bluefireteam/melos-action) that should semi-automate (PRs are still used) the whole publishing flow in monorepos, or any repositories using [melos](https://github.com/invertase/melos).

Design doc: https://flutter.dev/go/pub-automated-publishing

spydon:
One solution could be for pub to also allow `workflow_dispatch` events, they can also be connected to tags.
I thought that might already work, but sadly it didn't:
> The calling GitHub Action is not allowed to publish, because: publishing is only allowed from 'push' events, this token originates from a 'workflow_dispatch' event.

spydon:
What do you think @isoos @sigurdm?

isoos:
@spydon: I am not that familiar with the workflow internals. Is there a document that describes what the users are doing and what the system is doing in such case?

spydon:
@isoos So `workflow_dispatch` is the event that is used when a workflow is triggered manually:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

It can also be connected to a tag, so the rest of the server-side checks on your side should still be the same as they are for `push`.

isoos:
[...]

PR #7462, opened 2024-02-05
Allow GitHub `workflow_dispatch` events to publish

spydon:
This PR allows new versions of packages to be published from GitHub `workflow_dispatch` events.
The same restriction on the `refType` is in place, it still needs to be run targeting a tag.

Closes: #7177

Reason for change:
This is currently a blocker for actions like the [melos-action](https://github.com/marketplace/actions/melos-action) to be used to its full potential, where the user can create releases directly from the GitHub UI in the following manner:

1. The maintainer presses  ["Run workflow"](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow).
image
2. The action creates a branch with updated `CHANGELOG.md` and `pubspec.yaml` files and uploads it as a PR
3. A maintainer reviews and merges the release PR
4. Another workflow is triggered by the merge that then tags the packages and triggers yet another workflow through `workflow_dispatch`, since tags created from a GitHub action [aren't allowed to trigger other actions](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow)
5. The last workflow uploads the newly versioned packages to pub.dev

This is a lot less error prone than doing manual releases from the command line, especially in bigger monorepos.

---

<details>

</details>


jonasfj:
Are you sure these checks will still work?
```dart
    if (agent.payload.refType != 'tag') {
      throw AuthorizationException.githubActionIssue(
          'publishing is only allowed from "tag" refType, this token has "${agent.payload.refType}" refType');
    }
    final expectedRefStart = 'refs/tags/';
    if (!agent.payload.ref.startsWith(expectedRefStart)) {
      throw AuthorizationException.githubActionIssue(
          'publishing is only allowed from "refs/tags/*" ref, this token has "${agent.payload.ref}" ref');
    }
    final expectedTagValue = tagPattern.replaceFirst('{{version}}', newVersion);
    if (agent.payload.ref != 'refs/tags/$expectedTagValue') {
```

Will it require that the workflow_dispatch is triggered on a tag?

spydon:
@jonasfj added a test in https://github.com/dart-lang/pub-dev/pull/7462/commits/1be34bf71bfd10ed3a9ea8d8440b17d319f524fa

I'm not sure if it is possible to trigger the second `if`, possibly by manipulating the request somehow, because if the `refType` is `tag` the `ref` must surely start with `refs/tags` right? Anyhow, if anyone would manage to do that it is still covered, just like it was before.

app/lib/package/backend.dart
```diff
@@ -1268,7 +1268,8 @@ class PackageBackend {
 
     // TODO: consider allowing other events from
     //       https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows
-    if (agent.payload.eventName != 'push') {
+    if (agent.payload.eventName != 'push' &&
+        agent.payload.eventName != 'workflow_dispatch') {
       throw AuthorizationException.githubActionIssue(
           'publishing is only allowed from "push" events, this token originates from a "${agent.payload.eventName}" event');
```
clragon, line 1274: We probably wanna adjust this error message

clragon (changes requested):

app/lib/package/backend.dart
spydon: Good catch, updated!

Gustl22:
@spydon seems, there are some conflicts.

spydon:
> @spydon seems, there are some conflicts 😸 I'm excited about this feature ;D

It's just the changelog, it has to be updated every time something else is merged (gives me throwbacks from before we used Melos in all our projects :sweat_smile:).

Gustl22:
@clragon @jonasfj anything blocking to finish the PR? Best regards :)

spydon:
> @clragon @jonasfj anything blocking to finish the PR? Best regards :)

The design document needs to be reviewed and accepted first, can be seen here: https://github.com/dart-lang/pub-dev/issues/7177#issuecomment-1977676681

And hopefully after that there only needs to be a pair of check boxes added and theeen we can  get this in. 😄

isoos:
@spydon: We've come to a conclusion and we'd like to allow this event, but only as a UI option after an admin allowed it on the admin UI. Would you be interested to implement it as such?

It would need to be stored on the [`GithubPublishingConfig`](https://github.com/dart-lang/pub-dev/blob/master/pkg/_pub_shared/lib/data/package_api.dart#L69) object, and be displayed on the [UI of the package page admin](https://github.com/dart-lang/pub-dev/blob/master/app/lib/frontend/templates/views/pkg/admin_page.dart#L240).

@jonasfj: I think we could use a simple `bool? isWorkflowDispatchEnabled` flag on the object and a checkbox on the UI - or would you like to see a more open text field for the additional event names?

jonasfj:
@isoos I think we should do two booleans.

```dart
bool isFromWorkflowDispatchEventEnabled;
bool isFromPushEventEnabled;
```

Or maybe `allowPublishFrom<Event>`, but we probably need to make two checkboxes in the UI:
 * [x] Enable publishing from `push` events.
 * [ ] Enable publishing from `workflow_dispatch` events.

Where the first defaults to being enabled, and the second can be enabled.

I think the ability to disallow publishing from `push` events is nice. I think we can allow the freedom of letting people enable both, but let's give them the ability to only allow the thing they use.

jonasfj:
@isoos this is why I thought it might be necessary with some data migration, in order to just create the properties on all entities. That's something we could easily do with a backfill job, right?

isoos:
> @isoos this is why I thought it might be necessary with some data migration, in order to just create the properties on all entities. That's something we could easily do with a backfill job, right?

Yeah, makes sense.

spydon:
> Would you be interested to implement it as such?

Sounds great! 😃

isoos:
Note: I'm working on a PR that will add the checkboxes to the UI + other changes and tests.

isoos:
Closing as #7766 was merged.

PR #7479, opened 2024-02-13
Use version fallbacks when initializing Dart SDK index for search.

isoos:


jonasfj (approved):
I do think we should consider downloading these things as part of the deployment, when we install the Dart SDK.

But that might be an issue we should address later.

Issue #7482, opened 2024-02-15
Crashing isolate due to uncaught exception: ClientException


jonasfj:
@isoos if we're using `RetryClient` from `package:http` am I correct in that it won't retry if partial response is received. Like in the case of `ClientException`?

I'm guessing this error isn't the end of the world, but we probably shouldn't crash due to this. I'm also guessing that this isn't the only place where we have this issue.

```
pub.analyzer: ERROR from index-builder isolate #1

Error:
    [Exception: Crashing isolate due to uncaught exception: ClientException: Connection closed before full header was received, uri=https://storage.googleapis.com/resumable/upload/storage/v1/b/dartlang-pub--search-snapshot/o?name=snapshot%2F2024.02.13.json.gz&uploadType=resumable&alt=json, #0      _wrapper.<anonymous closure> (package:pub_dev/service/entrypoint/_isolate.dart:316:9)
    #1      _RootZone.runBinary (dart:async/zone.dart:1666:54)
    #2      StackZoneSpecification._handleUncaughtError (package:stack_trace/src/stack_zone_specification.dart:161:20)
    #3      _Zone._processUncaughtError (dart:async/zone.dart:1081:14)
    #4      _CustomZone.handleUncaughtError (dart:async/zone.dart:1285:5)
    #5      Future._propagateToListeners (dart:async/future_impl.dart:760:16)
    #6      Future._completeError (dart:async/future_impl.dart:655:5)
    #7      Future._asyncCompleteError.<anonymous closure> (dart:async/future_impl.dart:745:7)
    #8      StackZoneSpecification._run (package:stack_trace/src/stack_zone_specification.dart:207:15)
    #9      StackZoneSpecification._registerCallback.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:114:48)
    #10     _rootRun (dart:async/zone.dart:1399:13)
    #11     _CustomZone.run (dart:async/zone.dart:1301:19)
    #12     _CustomZone.runGuarded (dart:async/zone.dart:1209:7)
    #13     _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:async/zone.dart:1249:23)
    #14     _microtaskLoop (dart:async/schedule_microtask.dart:40:21)
    #15     _startMicrotaskLoop (dart:async/schedule_microtask.dart:49:5)
    #16     _runPendingImmediateCallback (dart:isolate-patch/isolate_patch.dart:118:13)
[...]

PR #7486, opened 2024-02-16
Refactored try-catch in claim refresh loop.

isoos:
- #7482
- moved the try-catch loop from inside to cover the entire microtask

### DIFF
--- a/app/lib/task/global_lock.dart
+++ b/app/lib/task/global_lock.dart
@@ -41,8 +41,8 @@ class GlobalLock {
     final claimId = c._entry.claimId;
     var refreshed = Future.value(true);
     final done = Completer<void>();
-    try {
-      scheduleMicrotask(() async {
+    scheduleMicrotask(() async {
+      try {
         while (c.valid && !done.isCompleted) {
           // Await for 50% of the time until expiration is gone, then we refresh
           var delay = c.expires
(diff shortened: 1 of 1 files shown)


app/lib/task/global_lock.dart
```diff
@@ -63,21 +63,19 @@ class GlobalLock {
...
-              _log.severe('Error refreshing claim $claimId on $_lockId', e, st);
-              return;
+            if (!await refreshed) {
+              _log.warning('failed to refresh claim $claimId on $_lockId');
+              return; // Stop refreshing, if refresh failed.
             }
           }
         }
-      });
+      } on Exception catch (e, st) {
```
jonasfj, line 72: Something like this should really only happen in `await refreshed`.

But I guess I don't really mind this.

You probably should consider also eating the exception in `await refreshed; // wait to any ongoing refreshing attempt is done` later...

app/lib/task/global_lock.dart
```diff
@@ -41,8 +41,8 @@ class GlobalLock {
     final claimId = c._entry.claimId;
     var refreshed = Future.value(true);
     final done = Completer<void>();
-    try {
-      scheduleMicrotask(() async {
+    scheduleMicrotask(() async {
```
jonasfj, line 44: This doesn't move try-catch into the microtask... it moves it out from `await refreshed` which is otherwise the only thing that could cause this.

app/lib/task/global_lock.dart
```diff
@@ -63,21 +63,19 @@ class GlobalLock {
...
+            if (!await refreshed) {
+              _log.warning('failed to refresh claim $claimId on $_lockId');
+              return; // Stop refreshing, if refresh failed.
             }
           }
         }
-      });
+      } on Exception catch (e, st) {
+        // Stop refreshing, and log the error
+        _log.severe('Error refreshing claim $claimId on $_lockId', e, st);
```
jonasfj, line 74: If we want to harden the micro-task for unhandled exceptions/errors.

I'd suggest keeping the try/catch we had before and wrap everything inside `scheduleMicrotask` in a new one that has a more correct error message.

This says that there was an error during refreshing. But that's only the case if it was thrown by `await refreshed`.

Thus, I'd suggest that a try-catch that covers the entire body of the microtask should say something less clear. Like `Internal error while maintaining claim for GlobalLock`. And it should probably catch both errors and exceptions. To be clear such a high-level try catch is not there because there are any errors/exceptions to catch -- I don't see any. But it's there because: if there is an internal error, then we can stop refreshing the claim, let it expire and continue gracefully -- we do not need to crash the whole application. Ofcourse this is a risk, because we don't know why this internal error happened, whether it's temporary or permanent and whether it could be resolved by restarting.

isoos:
Closing this for now, let's see if/how #7483 will fix the error in the log.

PR #7492, opened 2024-02-20
Minimal report page + API + email sending.

isoos:


sigurdm (approved):

pkg/web_app/lib/src/api_client/api_client.dart
```diff
@@ -26,3 +27,17 @@ PubApiClient get unauthenticatedClient =>
 PubApiClient get client {
   return PubApiClient(_baseUrl, client: http.createClientWithCsrf());
 }
+
+Future<Map<String, dynamic>> requestJson(
```
jonasfj, line 31: Probably best to document what this does in a  documentation comment.

And if we make `POST` the default verb, I'd suggest removing the `verb` option.
We could also make the `verb` a required property.

I'm just a bit confused, we'll think the method does something else in the future.

```suggestion
Future<Map<String, dynamic>> postJson(
```

app/lib/frontend/handlers/report.dart
```diff
@@ -0,0 +1,67 @@
...
+import '../../shared/utils.dart';
+import '../request_context.dart';
+import '../templates/report.dart';
+
+/// Handles GET /report
+Future<shelf.Response> reportPageHandler(shelf.Request request) async {
+  if (!requestContext.experimentalFlags.isReportPageEnabled) {
+    return notFoundHandler(request);
+  }
+  final unauthenticatedRs = await checkAuthenticatedPageRequest(request);
```
jonasfj, line 26:
```suggestion
  // TODO: Final report page cannot require authentication!
  final unauthenticatedRs = await checkAuthenticatedPageRequest(request);
```

app/lib/frontend/handlers/report.dart
```diff
@@ -0,0 +1,67 @@
...
+  InvalidInputException.checkStringLength(
+    form.description,
+    'description',
+    minimum: 20,
+    maximum: 8192,
+  );
+
+  final bodyText = <String>[
+    'New report recieved on ${now.toIso8601String()}: $id',
+    'Form content:',
```
jonasfj, line 56:
```suggestion
```

app/lib/shared/email.dart
```diff
@@ -238,3 +243,19 @@ ${_footer('transfer')}
...
+/// Creates the report page [EmailMessage] that we be sent to pub dev admins.
+EmailMessage createReportPageAdminEmail({
+  required String id,
+  required String userEmail,
+  required String bodyText,
+}) {
+  return EmailMessage(
+    _notificationsFrom,
+    [EmailAddress('support@pub.dev')],
+    'New report: $id',
```
jonasfj, line 256:
```suggestion
    'pub.dev case-id: $id',
```

app/test/frontend/handlers/report_test.dart
```diff
@@ -0,0 +1,105 @@
...
+import 'package:pub_dev/frontend/request_context.dart';
+import 'package:test/test.dart';
+
+import '../../shared/handlers_test_utils.dart';
+import '../../shared/test_services.dart';
+import '_utils.dart';
+
+void main() {
+  group('Report handlers test', () {
+    testWithProfile('requires authentication', fn: () async {
```
jonasfj, line 18: We don't require that

jonasfj (commented):
I don't mind landing this and iterating from here.

But this is still a bit from what we want.

pkg/web_app/lib/src/api_client/api_client.dart
isoos: Updated this to `sendJson` + required method parameters, PTAL.

app/test/frontend/handlers/report_test.dart
isoos: I'd leave it here for now. Once we allow the non-authenticated page, we can update the test.

isoos:
Landing this for now, iterating on it in follow-up PRs.

PR #7496, opened 2024-02-22
Temporary dartdoc workaround: remove leading href parents from extension type sidebars.

isoos:
- This is only a temporary measure to unblock the site until https://github.com/dart-lang/dartdoc/issues/3663 gets fixed.
- As the sidebars are already broken, there is little risk to do a crude string replace (before sanitization).

### DIFF
--- a/app/test/shared/versions_test.dart
+++ b/app/test/shared/versions_test.dart
@@ -146,7 +146,7 @@ and do not format to also bump the runtimeVersion.''',
     final content =
         await File('../pkg/pub_worker/lib/src/bin/pana_wrapper.dart')
             .readAsString();
-    expect(content, contains("dartdocVersion: '$dartdocVersion'"));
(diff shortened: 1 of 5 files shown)


isoos:
Note: this has been verified locally with `package:web`.

sigurdm:
Are we sure there are no other places we generate `../` that are going to break as a result of this?

isoos:
> Are we sure there are no other places we generate ../ that are going to break as a result of this?

I think the risk is very low to that:
- We are filtering on extension type sidebars via the filename.
- In addition to that, the regular sidebar does not have the `../` segment in the hrefs in the first place (hence the bug).
- I'd expect that the dartdoc fix will remove the leading `../` we are just doing it in a not-that-nice way.

sigurdm (approved):

PR #7498, opened 2024-02-22
Catch markdown rendering errors and output in <pre> block.

isoos:


app/lib/shared/markdown.dart
```diff
@@ -53,6 +53,12 @@ String markdownToHtml(
       nodes = _groupChangelogNodes(nodes).toList();
     }
     return _renderSafeHtml(nodes, disableHashIds: disableHashIds);
+  } catch (e, st) {
+    _logger.shout('Error rendering markdown.', e, st);
+    // disabling the opening and closing tags makes sure we don't have any
+    // HTML elements that would break out of the wrapper <pre>
+    final safeText = text.replaceAll('<', ' ').replaceAll('>', ' ');
```
sigurdm, line 60: Could we replace with `&lt;` and also replace `&` with &amp; 

See https://www.geeksforgeeks.org/which-characters-should-be-escaped-inside-a-pre-tag/

isoos: Done.

jonasfj (approved):
I guess we can't make test cases for this, right? Since a known bug in markdown is not something we expect there to be

PR #7519, opened 2024-02-28
On-demand download of preview/beta analysis SDKs.

isoos:
- We will only use stable SDKs as part of the worker image.
- When the worker detects a need for newer SDK, it can download it dynamically into the `<path>/<sdk>/preview` directory.
- Dart SDK is selected by querying the latest version endpoint for `beta` channel.
- Flutter SDK is selected by downloading the Flutter version archive and selecting the version manually.
- In each case, if there is no such version available, the `master` channel/branch will be used.
- Updated `setup-dart.sh` to also accept `master` as version.

### DIFF
--- a/pkg/pub_worker/lib/src/bin/pana_wrapper.dart
+++ b/pkg/pub_worker/lib/src/bin/pana_wrapper.dart
@@ -257,6 +257,7 @@ Future<String?> _installSdk({
   }
   if (!await Directory(sdkPath).exists()) {
     final configHomePath = _configHomePath(sdkKind, configKind);
+    // TODO: setup/download with retries (optionally with CRC/hash checks)
     await runConstrained(
       [
         'tool/setup-$sdkKind.sh',
--- a/pkg/pub_worker/lib/src/bin/pana_wrapper.dart
+++ b/pkg/pub_worker/lib/src/bin/pana_wrapper.dart
@@ -6,6 +6,8 @@ import 'dart:async';
 import 'dart:convert' show json, utf8;
 import 'dart:io' show Directory, File, Platform, exit, gzip;
 
+import 'package:_pub_shared/utils/dart_sdk_version.dart';
+import 'package:_pub_shared/utils/flutter_archive.dart';
 import 'package:clock/clock.dart';
 import 'package:collection/collection.dart';
 import 'package:logging/logging.dart' show Logger, Level, LogRecord;
@@ -160,8 +162,9 @@ Future<void> main(List<String> args) async {
 final _workerConfigDirectory = Directory('/home/worker/config');
 late final _workerConfigPath =
     _workerConfigDirectory.existsSync() ? _workerConfigDirectory.path : null;
+late final _isInsideDocker = _workerConfigPath != null;
 String? _configHomePath(String sdk, String kind) {
-  if (_workerConfigPath == null) {
+  if (!_isInsideDocker) {
     return null;
   }
   return p.join(_workerConfigPath!, '$sdk-$kind');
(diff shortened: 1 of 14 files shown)


pkg/pub_worker/lib/src/bin/pana_wrapper.dart
```diff
@@ -162,50 +165,103 @@ Future<(SdkConfig, SdkConfig)> _detectSdks(Pubspec pubspec) async {
...
     path: Directory(Platform.environment['FLUTTER_ROOT'] ?? '').parent,
   );
 
   // Choose stable Dart and Flutter SDKs for analysis
-  var dartSdk = dartSdks.firstWhereOrNull((sdk) => !sdk.version.isPreRelease) ??
-      (dartSdks.isNotEmpty ? dartSdks.first : null);
-  var flutterSdk =
+  final installedDartSdk =
+      dartSdks.firstWhereOrNull((sdk) => !sdk.version.isPreRelease) ??
+          (dartSdks.isNotEmpty ? dartSdks.first : null);
```
jonasfj, line 183: `package:collection` has a `firstOrNull` extension on `iterable`

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
```diff
@@ -162,50 +165,103 @@ Future<(SdkConfig, SdkConfig)> _detectSdks(Pubspec pubspec) async {
...
     path: Directory(
       Platform.environment['DART_SDK'] ??
           Directory(Platform.resolvedExecutable).parent.parent.path,
     ).parent,
   );
-  final flutterSdks = await InstalledSdk.fromDirectory(
+  final flutterSdks = await InstalledSdk.scanDirectory(
     kind: 'flutter',
     path: Directory(Platform.environment['FLUTTER_ROOT'] ?? '').parent,
   );
```
jonasfj, line 178: Why do we scan here? are we not only interested in the stable version?

And are we not interested in NOT being affected by previous pana invocations?

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
```diff
@@ -162,50 +165,103 @@ Future<(SdkConfig, SdkConfig)> _detectSdks(Pubspec pubspec) async {
...
+}) async {
+  if (!_isInsideDocker) {
+    return null;
+  }
+  if (!await Directory(sdkPath).exists()) {
+    final configHomePath = _configHomePath(sdkKind, configKind);
+    await runConstrained(
+      [
+        'tool/setup-$sdkKind.sh',
+        sdkPath,
```
jonasfj, line 254: this is probably not reliable.

Like, we should download with retries, probably, and possibly consider checking the `x-goog-hash` header, like crc32.. This could be a follow up, but worth considering.

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
jonasfj: Shouldn't this be:
 * (A) Find the stable SDK that was shipped in the current docker image
 * If (A) works, use it!
 * (B) List available prerelease SDKs (beta and dev releases)
 * (C) Find the oldest SDK from (B) that satisfies the SDK-constraints for the package
 * Check if the SDK found in (C) has already been downloaded, if so use it
 * Download the SDK found in (C), install it and use it.

jonasfj: This way, multiple pre-releases of the same SDK doesn't affect eachother.

I'm a little uncertain of (C). Choosing the minimum SDK that satisfies the SDK constraint makes some sense. That way the SDK won't change when the package is re-analyzed.

On the flip side, analyzing with the latest SDK pre-release could make sense too.

------------

oh, I'm missing logic with preferable a new stable SDK if one is available, and the current one in the image doesn't satisfy the SDK constraint.

jonasfj: So maybe the priority is:

 * (1) SDK from the docker image, if that satisfies constraints
 * (2) Latest stable SDK, if that satisfies constraints
 * (3) Latest beta SDK, if that satisfies constraints
 * (4) Latest dev SDK, if that satisfies constraints

Once an SDK has been identified, check if it's installed, if not install it.

jonasfj (approved):

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
isoos: Agreed on most parts, adding a TODO and will do a follow-up in later PRs.

PR #7524, opened 2024-02-29
Test to ensure all pub get is clean.

### LINKED ISSUE #7516: We need a test to check whether `pub get` was called in all packages.

isoos:
Fixes #7516

### DIFF
--- a/app/test/ensure_all_pub_get_is_clean_test.dart
+++ b/app/test/ensure_all_pub_get_is_clean_test.dart
@@ -1,0 +1,56 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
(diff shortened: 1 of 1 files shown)


app/test/ensure_all_pub_get_is_clean_test.dart
```diff
@@ -0,0 +1,56 @@
...
+        final workingDirectory = p.join(Directory.current.path, '..', dir);
+        await runConstrained(
+          ['dart', 'pub', 'get'],
+          workingDirectory: workingDirectory,
+          throwOnError: true,
+          retryOptions: RetryOptions(maxAttempts: 2),
+          retryIf: (_) => true,
+        );
+        final pr = await runConstrained(
+          ['git', 'diff', '--relative'],
```
sigurdm, line 37: Instead of using `git diff` you could do `dart pub get --enforce-lockfile`

sigurdm (approved):

PR #7528, opened 2024-03-05
Add 'pub_display_url' property to the SecurityAdvisory entity

szakarias:


app/lib/service/security_advisories/backend.dart
```diff
@@ -123,6 +130,21 @@ class SecurityAdvisoryBackend {
...
+    if (githubId != null) {
+      return 'https://github.com/advisories/$githubId';
+    }
+
+    final cveId = idAndAliases.firstWhereOrNull((id) => id.startsWith('CVE'));
+    if (cveId != null) {
+      return 'https://osv.dev/vulnerability/$cveId';
+    }
+
+    return idAndAliases.first;
```
sigurdm, line 145: Maybe we should just return null if we cannot find a url.

The client probably has to handle null here anyways.

szakarias: We agreed to show the plain id in this case

jonasfj, line 145:
```suggestion
    return 'https://osv.dev/vulnerability/'${idAndAliases.first}';
```

jonasfj (approved):

app/lib/service/security_advisories/backend.dart
sigurdm: Yeah - but I think the client should decide on that. 'pub_display_url' suggests that it contains a url

szakarias: Actually since we in pub-dev always fetch advisories from osv, we will be sure to find it at osv.dev.

sigurdm (approved):

PR #7531, opened 2024-03-05
Close HTTP client in finally block in flutter archive fetching.

### LINKED ISSUE #7527: client.close() in pkg/_pub_shared/lib/utils/flutter_archive.dart should be in a finally block
`fetchFlutterArchive` from `pkg/_pub_shared/lib/utils/flutter_archive.dart` doesn't use a `finally` block to close the `client`.

Also it'd probably be better to not use `httpRetryClient` and instead rely on wrapping with a retry function.

isoos:
Fixes #7527.


isoos:
Closing in favor of #7534.

PR #7542, opened 2024-03-07
Dynamic SDK selection may download latest stable versions too.

isoos:


pkg/pub_worker/lib/src/bin/pana_wrapper.dart
```diff
@@ -197,61 +197,130 @@ Future<(SdkConfig, SdkConfig)> _detectSdks(Pubspec pubspec) async {
...
+          constraint: pubspec.dartSdkConstraint,
+        ) &&
+        sdkMatchesConstraint(
+          sdkVersion: flutter,
+          constraint: pubspec.flutterSdkConstraint,
+        );
+  }
+
+  // try to use the latest SDKs in the docker image
+  final matchesInstallSdks = matchesSdks(
```
sigurdm, line 220: Nit:
```suggestion
  final matchesInstalledSdks = matchesSdks(
```

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
```diff
@@ -197,61 +197,130 @@ Future<(SdkConfig, SdkConfig)> _detectSdks(Pubspec pubspec) async {
...
+  final latestBetaDartSdk = await fetchLatestDartSdkVersion(channel: 'beta');
+  final latestBetaDartSdkVersion = latestBetaDartSdk?.semanticVersion;
+  final matchesLatestBeta = matchesSdks(
+    dart: latestBetaDartSdkVersion,
+    flutter: flutterArchive?.latestBeta?.semanticVersion,
+  );
+  if (matchesLatestBeta) {
+    final dartSdkPath = await _installSdk(
+      sdkKind: 'dart',
+      configKind: 'latest-beta',
```
sigurdm, line 281: Would it be possible to ask for the specific version we think is the latest beta, instead of just asking for the "latest beta" (same for stable above) to avoid however small a race.

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
```diff
@@ -197,61 +197,130 @@ Future<(SdkConfig, SdkConfig)> _detectSdks(Pubspec pubspec) async {
...
+
+  // try to use the latest stable downloadable SDKs
+  final latestStableDartSdk =
+      await fetchLatestDartSdkVersion(channel: 'stable');
+  final latestStableDartSdkVersion = latestStableDartSdk?.semanticVersion;
+  final matchesLatestStable = matchesSdks(
+    dart: latestStableDartSdkVersion,
+    flutter: flutterArchive?.latestStable?.semanticVersion,
+  );
+  if (matchesLatestStable) {
```
sigurdm, line 248: This screams for a loop (pseudo-code)
```
var found = false;
for (final sdkVersions in [latestStableVersions, latestBetaVersions]) {
  if (sdkMatchesConstraint(constraint, sdkVersions)) {
    _installSdk(sdkVersions);
    found = true;
    break;
  }
}
if (!found) {
  // use installed
}

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
isoos: Not sure: what is the race here?

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
isoos: This is running on each worker and each package-version separately on each run. If we were doing this in a loop, we'd need to query the versions each time (and we don't share a redis cache between the workers), or we'd need to implement this logic outside of the worker where we can cache things. The current implementation does it only on-demand.

sigurdm: Not to "run in a loop" just to factor out checking and installing an sdk version, so you can avoid repeating the same code for latest stable and latest beta.

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
sigurdm: If you detect that latest stable is good
a new stable is released
you ask to install latest table

The version you checked against will not be the version you install

isoos: Oh, the `latest-beta` is just the name for (a) the SKD directory as in `/home/worker/dart/latest-beta`, and also for its configuration directory as in `/home/worker/config/latest-beta`. The version we are installing (the `version` parameter) is the same we are checking the constraint against.

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
isoos: Ah, got it, I can give it another go (it got a bit complex on my first try to make it more generic and I ditched it).

sigurdm: If it is too complex we can also ship as-is

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
sigurdm: Ah - then all is good. I thought you were passing the channel name instead of the version

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
isoos: Updated, PTAL.

sigurdm (approved):

Issue #7430, opened 2024-01-23
Maybe consider different timeouts for analysis


isoos:
One particular package that does not get any analysis for a while:
https://pub.dev/packages/aws_client/score

isoos:
Apparently this should be implemented in `pana`: it should use a total time budget (overall timeout) and reduce the `dartdoc` timeout if the previous steps took a significant amount of time.

isoos:
A recent `googleapis` run on pub.dev needed the following wall-clock time:
- about 5-6 minutes for running `dartdoc`
- about 8-9 minutes for post-processing
- about 2 minutes for creating the .tar.gz archive

An unsuccessful run with `aws_client` took about 21 minutes of `dartdoc`, and timed out on post-processing. It is very likely that on such packages the upload will also fail because of the large size of files to be packages into the blob.

Couple of relatively cheap things to experiment with:
- running one of the post-processing tasks in an isolate
- emitting the `summary.json` before the dartdoc post-processing happens
- skip post-processing altogether if the combined pana+dartdoc run took too long (since we have very little chance to complete it anyway)

sigurdm:
I think it makes sense that we
* run post processing in isolate or process such that we can interrupt it if it is timing out.
* If post-processing fails only the pana report is uploaded. No docs.

If aws_client is the only package big enough to give us problems, I don't think we should do much more from our side; but try to push to get the quadratic behavior in dartdoc fixed.

isoos:
> If aws_client is the only package big enough to give us problems, I don't think we should do much more from our side; but try to push to get the quadratic behavior in dartdoc fixed.

Agreed. I think the main issue for us is the large number (and also total size) of files dartdoc outputs, and limiting it in any way would help. Until then, dropping the post-processing for `aws_client` would still be an improvement, as right now it won't get the pana results and tags either.

isoos:
`aws_client` got a good analysis without dartdoc post-processing (no docs are served). I think we can close this issue for now.

PR #7548, opened 2024-03-12
Cutoff time check while listing files in dartdoc post-processing.

isoos:
- #7430
- It appears that the stream of `Directory.list(recursive: true) does have large lags in it, trying to use a finer-tuned listing check to prevent going over the time budget.


isoos:
Discussed offline: we should try a simpler approach.

PR #7553, opened 2024-03-12
Refactored dartdoc post-processing.

isoos:
- #7430
- decreased dartdoc timeout as even large packages can manage to get it done within 30 minutes
- dart post-processing may be skipped entirely based on the time spent in pana analysis
- processing of files and archive building is in separate isolates, running in parallel

### DIFF
--- a/pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
+++ b/pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
@@ -5,9 +5,9 @@
 import 'dart:async';
 import 'dart:convert' show json, utf8, Utf8Codec;
 import 'dart:io';
+import 'dart:isolate';
 
 import 'package:_pub_shared/dartdoc/dartdoc_page.dart';
-import 'package:clock/clock.dart';
 import 'package:logging/logging.dart' show Logger;
 import 'package:path/path.dart' as p;
 import 'package:stream_transform/stream_transform.dart';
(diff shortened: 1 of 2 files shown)


pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
```diff
@@ -22,75 +22,76 @@ Future<void> postProcessDartdoc({
...
+    await _list()
+        .transform(tarWriter)
+        .transform(gzip.encoder)
+        .pipe(tmpTar.openWrite());
+  }).whenComplete(() {
+    _log.info('Finished .tar.gz archive');
+  });
 
-  _log.info('Finished .tar.gz archive');
+  await Future.wait([
```
sigurdm, line 87: What a shame `Isolate.run` doesn't support timeouts or some easy way to kill the isolate after a timeout...

Maybe we have to write that abstraction ourselves.

sigurdm (approved):

pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
isoos: Oh, I can add a timer that just kills the isolate from the inside... maybe in a follow-up PR.

sigurdm: ...Unless we have sync tasks blocking inside the isolate...

isoos: Maybe we don't need this at all, as worst case the parent process will kill pana_wrapper after the set timeout.

Issue #7535, opened 2024-03-06
New moderation and report processing flow


jonasfj:
The things we have to do are:
 * [ ] Moderation actions:
   * [x] Introduce `moderatedAt` and `isModerated` on
     * [x] `Package`
     * [x] `PackageVersion`
     * [x] `Publisher`
     * [x] `User`
   * [x] Enforce moderated semantics for:
     * [x] `Package` (not available in API, public bucket, search, and web page says moderated)
     * [x] `PackageVersion` (not available in API, public bucket, and web page says moderated)
     * [x] `Publisher` (web page says moderated, packages owned by publisher can't be updated)
     * [x] `User` (can't sign-in on pub.dev or publish using `dart pub`; packages owned exclusively by the user are marked discontinued)
   * [x] Administrative actions to:
     * [x] Set or clear `Package.isModerated/moderatedAt`
     * [x] Set or clear `PackageVersion.isModerated/moderatedAt`
     * [x] Set or clear `Publisher.isModerated/moderatedAt`
     * [x] Set or clear `User.isModerated/moderatedAt`
     * [x] Add optional case tracking: `--case=<caseId>` (use `--case=none`)
     * [x] Add optional email-body for content-owners
   * [ ] Periodic task to delete entities and associated resources if `now() - moderatedAt > 3 years`
         (users are still just `user.isDeleted = true` with `Likes` being deleted).

 * [ ] In-take report form
   * [ ] content about the process and what to expect
   * [x] form submitting + processing
   * [x] linking from public pages
   * [x] setup deep linking for other systems

 * [x] notifications
   * [x] confirmation emails on form submission
   * [x] notification emails on case resolution (with appeal-link)
   * [x] notification emails on restrictive action (with appeal-link)
 
 * [x] Tracking cases
   * [x] List active cases
   * [x] Create case with `source = notification | trusted-flagger | authorities | legal-referral`
   * [x] Close an active case
   * [x] Modify a case (including status)

 * [x] Tracking user account moderation / blocked / suspended
   * [x] When a user was moderated
   * [x] Why the user was moderated: `reason = illegal-content | policy-violation | unfounded-notifications | unfounded-appeals`
   This will be extremely rare, we can probably track this with properties on the `User` entity.

 * [ ] statistical reporting

-------------------

## Implementation details of `isModerated`/`moderatedAt`

To ensure that we reverse moderation decisions, we're going to add properties:
[...]

PR #7554, opened 2024-03-13
Integrity checks for isModerated/moderatedAt fields.

isoos:
#7535


app/lib/shared/integrity.dart
```diff
@@ -879,3 +906,20 @@ class _PublisherAttributes {
     _memberCount.clear();
   }
 }
+
+Stream<String> _checkModeratedFlags({
```
jonasfj, line 910:
```suggestion
/// Check that `isModerated` and `moderatedAt` are consistent.
Stream<String> _checkModeratedFlags({
```

jonasfj (approved):

PR #7556, opened 2024-03-13
Package.isModerated removes scores and handles non-visibility with new Exception

isoos:
#7535

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -366,6 +366,15 @@ class TaskBackend {
         if (state != null) {
           tx.delete(state.key);
         }
+        // also delete earlier runtime versions
+        for (final rv
+            in acceptedRuntimeVersions.where((rv) => rv != runtimeVersion)) {
+          final key = PackageState.createKey(_db, rv, packageName);
+          final s = await tx.lookupOrNull(key);
+          if (s != null) {
+            tx.delete(s.key);
+          }
+        }
         return true;
       }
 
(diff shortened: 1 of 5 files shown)


app/lib/task/backend.dart
```diff
@@ -366,6 +366,15 @@ class TaskBackend {
...
         }
+        // also delete earlier runtime versions
+        for (final rv
+            in acceptedRuntimeVersions.where((rv) => rv != runtimeVersion)) {
+          final key = PackageState.createKey(_db, rv, packageName);
+          final s = await tx.lookupOrNull(key);
+          if (s != null) {
+            tx.delete(s.key);
+          }
+        }
```
jonasfj, line 377: Is this strictly necessary. Won't older package versions catchup through the periodic background job?

Also we won't really be serving anything from such older runtimeVersions, right?
And even if we do, this won't really showup anywhere, we'll just get an analysis tasks that fails because it can't fetch the package.

jonasfj (approved):

app/lib/task/backend.dart
isoos: > Also we won't really be serving anything from such older runtimeVersions, right?

We may do as a fallback runtime version. It is probably not that much of an issue though, I just wanted to make sure we don't leave stale entries anywhere.

> Is this strictly necessary. Won't older package versions catchup through the periodic background job?

Not entirely sure that periodic job will clear this within a day. Maybe in a month?

PR #7565, opened 2024-03-18
Use https://pub.dev/ in download URLs (version listing API) and probably other places.

### LINKED ISSUE #7561: Archive links in package listings should point to pub.dev
Instead of pub.dartlang.org

isoos:
Fixes #7561.

### DIFF
--- a/app/config/dartlang-pub.yaml
+++ b/app/config/dartlang-pub.yaml
@@ -36,7 +36,7 @@ productionHosts:
 - pub.dartlang.org
 - pub.dev
 - api.pub.dev
-primaryApiUri: https://pub.dartlang.org/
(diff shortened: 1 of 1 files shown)


app/config/dartlang-pub.yaml
```diff
@@ -36,7 +36,7 @@ productionHosts:
 - pub.dartlang.org
 - pub.dev
 - api.pub.dev
-primaryApiUri: https://pub.dartlang.org/
+primaryApiUri: https://pub.dev/
```
sigurdm, line 39: We should maybe even remove this abstraction completely?...

isoos: Yeah, I shall do that in a follow-up PR.

PR #7571, opened 2024-03-20
Add sanity check that an advisory has an affected package with versions

szakarias:


app/lib/service/security_advisories/backend.dart
```diff
@@ -212,6 +212,28 @@ List<String> sanityCheckOSV(OSV osv) {
     errors.add('Invalid id, id too long (over 255 characters).');
   }
 
+  if (osv.affected == null || osv.affected!.isEmpty) {
```
jonasfj, line 215: hmm, I wonder if we should add annotations such that `affected` has a default value empty-list:

https://pub.dev/documentation/json_annotation/4.8.1/json_annotation/JsonKey/defaultValue.html

Same for `affected.versions` it would make life simpler.

------

If we try this, let's do so in a separate PR.

app/lib/service/security_advisories/backend.dart
```diff
@@ -212,6 +212,28 @@ List<String> sanityCheckOSV(OSV osv) {
...
+    errors.add('No affected packages for advisory ${osv.id}');
+  } else {
+    bool noAffectedHasVersion = true;
+    osv.affected!
+        .where((a) => a.package.ecosystem.toLowerCase() == 'pub')
+        .forEach((affected) {
+      if (affected.versions == null || affected.versions!.isEmpty) {
+        logger.warning('No versions specified for affected package '
+            '${affected.package.name} in advisory ${osv.id}.');
```
jonasfj, line 224: Is it wrong for `osv.affected` to contain multiple objects talking about the same package where one has `.versions` and one has `.ranges`?

Keep in mind that it's possible that `affected.package.name` appears multiple times.

app/lib/service/security_advisories/backend.dart
```diff
@@ -212,6 +212,28 @@ List<String> sanityCheckOSV(OSV osv) {
...
+      } else {
+        noAffectedHasVersion = false;
+      }
+    });
```
jonasfj, line 228: Just an idea, I haven't tested it. But it will find package names that are mentioned as being affected, but for which there isn't any of the entries in `affected` that specifies an explicit list of versions.
```suggestion
    final affectedPackagesWithoutVersions = osv.affected!
        .where((a) => a.package.ecosystem.toLowerCase() == 'pub')
        .groupBy((a) => a.package.name)
        .entries
        .where(((:key, :value)) => value.every((a) => (a.versions ?? []).isEmpty))
        .map(((:key, :value)) => key)
        .toSet();
    if (affectedPackagesWithoutVersions.isNotEmpty) {
      logger.warning(
        'Advisory ${osv.id} affects packages: '
        '${affectedPackagesWithoutVersions.join(', ')}, '
        'but does not specify explicit versions.',
      );
    }
```

app/test/service/security_advisory/security_advisory_test.dart
```diff
@@ -305,7 +349,12 @@ void main() {
         id: longid,
         modified: firstTime,
         published: firstTime,
-        affected: [],
+        affected: [
+          Affected(
+            package: Package(ecosystem: 'pub', name: 'oxygen'),
+            versions: ['1'],
```
jonasfj, line 355: Why not use an actual version number? just curious.

jonasfj (approved):

app/lib/service/security_advisories/backend.dart
szakarias: Noted.

app/lib/service/security_advisories/backend.dart
szakarias: This should at least not happen for the advisories we import from OSV.dev. And if there is extra information in the one with `ranges` we will take that into consideration.

app/lib/service/security_advisories/backend.dart
szakarias: I think it is preferrable to have a warning in the log for each object that does not have versions. In practice I doubt, this will happen very often though.

Also I think it is important with the error if non of the affected packages has `versions`, since then the advisory is not relevant.

PR #7573, opened 2024-03-20
Cache control on api

jonasfj:
We should probably add more cache-control on other API end-points that are frequently scanned by bots. Or just in general on everything that can reasonably be cached.

### DIFF
--- a/app/lib/frontend/handlers/headers.dart
+++ b/app/lib/frontend/handlers/headers.dart
@@ -13,21 +13,36 @@ class CacheHeaders {
   /// The maximum age while a response may be cached.
   final Duration maxAge;
 
-  /// The cache storage category used for signed-in users.
-  /// `private` by default.
-  final String? signedInStorage;
+  /// Add the `private` directive, indicating that the response must not be
+  /// stored in public cache (such as a CDN).
+  ///
+  /// If neither [private] or [public] is specified, `Cache-Control` will
+  /// set `private` if the request was authenticated.
+  final bool private;
+
+  /// Add the `public` directive, indicating that the response may be
+  /// stored in public cache (such as a CDN).
+  ///
+  /// If neither [private] or [public] is specified, `Cache-Control` will
+  /// set `private` if the request was authenticated.
+  final bool public;
 
   CacheHeaders._(
     this.maxAge, {
-    this.signedInStorage,
-  });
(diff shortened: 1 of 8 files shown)


app/lib/shared/handler_helpers.dart
```diff
@@ -136,7 +136,7 @@ shelf.Handler _requestContextWrapper(shelf.Handler handler) {
     if (!context.uiCacheEnabled && !CacheHeaders.hasCacheHeader(rs.headers)) {
```
jonasfj, line 136: This and the `requestContext.isNotAuthenticated` logic in app/lib/frontend/handlers/headers.dart is a bit weird.

Probably we should only have defaults in one location.

jonasfj: But I think we can clean that up later.

app/lib/frontend/handlers/headers.dart
```diff
@@ -13,21 +13,36 @@ class CacheHeaders {
...
+    assert(!private && !public, 'both private and public is not allowed');
+  }
 
   Map<String, String> call() {
     return <String, String>{
       HttpHeaders.cacheControlHeader: <String>[
-        requestContext.isNotAuthenticated
-            ? 'public'
-            : (signedInStorage ?? 'private'),
+        if (private || (!public && requestContext.isNotAuthenticated))
```
sigurdm, line 42: I find this hard to parse (too many negations).

What happens if neither `private` nor `public` but `requestContext.isAuthenticated` - is that possible? Will it end up public?

sigurdm: Maybe something along:
```suggestion
        if (public || requestContext.isNotAuthenticated)
          'public'
        else
          'private',
```

jonasfj: If `private` then we force it 'private'
else if `public` then we make it 'public'
else if requestContext.isNotAuthenticated then we make it 'public'
else 'private'

sigurdm: That's not what I read. I see the last `else` is `'public'`.

isoos:
I think the upload message change could have gone into a different PR, as now tests are failing on that message change.

isoos (approved):

PR #7576, opened 2024-03-21
New download urls

jonasfj:


sigurdm (approved):

app/lib/shared/urls.dart
```diff
@@ -143,8 +143,9 @@ String pkgActivityLogUrl(String package) =>
     pkgPageUrl(package, pkgPageTab: PkgPageTab.activityLog);
 
 String pkgArchiveDownloadUrl(String package, String version, {Uri? baseUri}) {
-  final path =
-      '/packages/${Uri.encodeComponent(package)}/versions/${Uri.encodeComponent(version)}.tar.gz';
+  final p = Uri.encodeComponent(package);
+  final v = Uri.encodeComponent(version);
+  final path = '/api/archives/$p-$v.tar.gz';
```
isoos, line 148: We should probably use `Uri.encodeComponent` here, just to be safe.

isoos: Nevermind, I didn't read two lines up.

isoos (approved):

app/lib/shared/urls.dart
jonasfj: We do that on the line above, right? :rofl:

app/lib/tool/test_profile/import_source.dart
```diff
@@ -65,8 +65,10 @@ class _PubDevImportSource implements ImportSource {
     final file = File(p.join(archiveCachePath, archiveName));
     // download package archive if not already in the cache
     if (!await file.exists()) {
+      // TODO: Use archive_url from version-listing, that is stable!
```
isoos, line 68: Nit: can't we use the same method to generate it?

isoos (approved):

app/lib/tool/test_profile/import_source.dart
jonasfj: Which method? It didn't work here because I changed it :rofl: and pub.dev doesn't support the new end-point yet.

isoos: Ah, got it.

PR #7587, opened 2024-03-25
An attempt at conservative default cache-control headers

jonasfj:
`public max-age=10min` for all responses where:
 * Request is GET
 * Response is HTML
 * Response is 200
 * Request has no authentication, xsrf-token or cookies
 * Response handler didn't explicitly set a different `Cache-Control` header

`public max-age=30min` for all responses where:
 * Request is GET
 * Request is handled by `PubApi` (ie. this is `/api/`)
 * Response is 200
 * Request has no authentication, xsrf-token or cookies
 * Response handler didn't explicitly set a different `Cache-Control` header

Also made a few exceptions:
 * set `private` on responses for `/consent` flow.
 * set `public max-age=10min` on screenshots
 * set `public max-age=10min` on version listing API `/api/packages/<package>`

### DIFF
--- a/app/lib/frontend/handlers/headers.dart
+++ b/app/lib/frontend/handlers/headers.dart
@@ -43,7 +43,9 @@ class CacheHeaders {
           'private'
         else if (public)
           'public'
-        else if (requestContext.isNotAuthenticated)
+        else if (requestContext.isNotAuthenticated &&
+            requestContext.uiCacheEnabled &&
+            requestContext.sessionData == null)
           'public'
         else
           'private',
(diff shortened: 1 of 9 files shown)


app/lib/frontend/handlers/headers.dart
```diff
@@ -57,8 +59,26 @@ class CacheHeaders {
...
+  /// Default cache-control for user-interface
+  static final defaultUI = CacheHeaders._(
+    Duration(minutes: 10),
+    // public / private is inferred from authentication + session header
+  );
+
+  /// Default cache-control for API responses
+  static final defaultApi = CacheHeaders._(
+    Duration(minutes: 30),
+    // public / private is inferred from authentication + session header
```
sigurdm, line 80: Isn't all of /api public?

sigurdm (approved):

app/lib/frontend/handlers/headers.dart
isoos: No, many endpoints are only for limited scope, e.g. admins (like, listing publisher members).

isoos:
> public max-age=30min

shouldn't we start with a lower default and increase it in a subsequent deployment?

app/lib/frontend/handlers/headers.dart
sigurdm: Ah - of course. Nvm

isoos (approved):

PR #7594, opened 2024-03-26
Rate limit check done at the time of creating the record (except for package publishing).

isoos:
Reverts most of the changes in #7570.


isoos:
Hm, CI still has "has sha256 hash mismatch." issue, which I though could be attributed to the in-transaction rate limit check. However, with this PR this won't be the case anymore, so the issue should have a different root cause.

jonasfj (approved):

PR #7602, opened 2024-03-28
Fixing integrity checks: hash checks and moderated archives.

isoos:
- Together with #7600 and #7599 this fixes the flaky tests.
- Archive hash check will always run locally (not just 0.01% of the cases).
- Canonical bucket check will always run locally (previously it was skipped).
- Check takes `isModerated` flag into consideration, checking the required state driven by the flag.

### DIFF
--- a/app/lib/shared/integrity.dart
+++ b/app/lib/shared/integrity.dart
@@ -623,7 +623,7 @@ class IntegrityChecker {
     }
 
     if (!canonicalInfo.hasSameSignatureAs(info)) {
-      yield 'Canonical archive for PackageVersion "${pv.qualifiedVersionKey}" differs in old bucket.';
+      yield 'Canonical archive for PackageVersion "${pv.qualifiedVersionKey}" differs from public bucket.';
     }
 
     final publicInfo = await storageService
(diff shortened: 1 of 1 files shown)


app/lib/shared/integrity.dart
```diff
@@ -602,28 +595,54 @@ class IntegrityChecker {
...
+      return;
+    }
+
+    if (!shouldBeInPublicBucket) {
+      yield 'PackageVersion "${pv.qualifiedVersionKey}" has matching public archive file but it must not.';
+      return;
+    }
+
+    if (!canonicalInfo.hasSameSignatureAs(info)) {
+      yield 'Canonical archive for PackageVersion "${pv.qualifiedVersionKey}" differs in old bucket.';
```
sigurdm, line 626: What is "old bucket"? Should this be:
```suggestion
      yield 'Canonical archive for PackageVersion "${pv.qualifiedVersionKey}" differs from public bucket.';
```

sigurdm (approved):
LGTM, but perhaps let jonas read it also.

jonasfj (approved):

Issue #7608, opened 2024-04-02
Slow tests in test/package/upload_test.dart


jonasfj:
I was playing around with: `DEBUG=* dart test test/package/upload_test.dart` and I noticed that tests are really slow.

I reviewed the logs and found:
```
 2445 ms [frontend.search_adapter] INFO: Unable to search packages.
ERROR: ClientException with SocketException: Connection refused (OS Error: Connection refused, errno = 111), address = localhost, port = 35688, uri=http://localhost:0/search?q=is%3Aflutter-favorite&tags=-is%3Adiscontinued&tags=-is%3Aunlisted&tags=-is%3Alegacy&offset=0&limit=100, null
```

Are we forgetting to setup something, or doing some sort of retries in the start of the server.

I think this single file takes 2 min, and most of it is probably us spending 2.5s every time we start a test case :see_no_evil:

isoos:
The main reason seems to be the search attempt to connect to `http://localhost:0` timing out after ~2-2.5 seconds, making most of the test cases run longer than needed. However, while #7677 was a simple workaround, it broke most landing page and listing page test, and this needs to be investigated a bit more.

isoos:
It looks like that for most tests we were failing to use/initialize the regular search and have used the fallback search (after a `SocketException` but not a `TimeoutException`). I think we should (a) fix the tests to use the regular search, and (b) refactor and maybe even remove the fallback search (as it will provide confusing or bad data for non-name search queries).

PR #7609, opened 2024-04-02
Allow sdk packages

jonasfj:
Fixing the issue where packages with SDK-dependencies can't be published.

CC @jakemac53 (I believe we'll have a workaround for `_macros` out today'ish; this will ship thursday or next week)

--------

Noticed some slow tests and ended up filing:
https://github.com/dart-lang/pub-dev/issues/7608

### DIFF
--- a/app/test/package/upload_test.dart
+++ b/app/test/package/upload_test.dart
@@ -1049,6 +1049,32 @@ void main() {
         );
       });
 
+      testWithProfile('has dependency does not exist', fn: () async {
+        final tarball = await packageArchiveBytes(
+            pubspecContent:
+                generatePubspecYaml('xyz', '1.0.0') + '  abc: ^1.0.0\n');
+        final rs = createPubApiClient(authToken: adminClientToken)
+            .uploadPackageBytes(tarball);
+        await expectApiException(
+          rs,
+          status: 400,
+          code: 'PackageRejected',
+          message: 'Dependency `abc` does not exist.',
+        );
+      });
+
+      testWithProfile('has SDK-dependency', fn: () async {
+        final tarball = await packageArchiveBytes(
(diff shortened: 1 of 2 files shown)


app/test/package/upload_test.dart
```diff
@@ -1049,6 +1049,32 @@ void main() {
...
+          code: 'PackageRejected',
+          message: 'Dependency `abc` does not exist.',
+        );
+      });
+
+      testWithProfile('has SDK-dependency', fn: () async {
+        final tarball = await packageArchiveBytes(
+            pubspecContent: generatePubspecYaml('xyz', '1.2.3') +
+                '  my_sdk_dep:\n'
+                    '    sdk: dart\n');
```
isoos, line 1070: This will allow package to have any dependency as long as they have `sdk: dart` (or `sdk:flutter`)?

isoos:
> Should we remove the special-case allow-listing of the macros package also?

We should probably have a tool or test that checks allow-listing regularly and indicates which to remove.

jonasfj:
As far as SDK packages are concerned this will allow dependency on any SDK package.

I think that's fine, you can pick one that doesn't exist, and then the package won't work for anyone. That's sad, but probably people will test their package first :see_no_evil: 

We could make a list of allowed SDK-dependencies. But I'm not sure it's necessary.

jakemac53:
LGTM, will the other fix (to add `_macros` to the reserved name list) not fix the problem for publishing then? Or do you think it will?

jakemac53:
> That fix will allow you to publish `_macros`

Did you mean `macros`? Fwiw, I still can't publish right now, but I don't know if it has been deployed yet.

isoos:
> Did you mean macros? Fwiw, I still can't publish right now, but I don't know if it has been deployed yet.

We were not able to deploy it yet. (needed to roll back)

PR #7611, opened 2024-04-03
Suspend rate limiting checks.

isoos:


isoos:
The investigation suggest that rate limiting is not to fault for the latency increase, this PR won't be necessary.

PR #7614, opened 2024-04-04
ModerationCase entity

isoos:
- https://github.com/dart-lang/pub-dev/issues/7535
- not in use yet, just the implementation of the specification

### DIFF
--- a/app/lib/admin/models.dart
+++ b/app/lib/admin/models.dart
@@ -58,6 +58,7 @@ class ModerationCase extends db.ExpandoModel<String> {
   String? appealedCaseId;
 
   /// One of:
+  /// - `pending`, if this is an appeal and we haven't decided anything yet.
   /// - `no-action`, if this is a notification (kind = notification) and
   ///   we decided to take no action.
   /// - `moderation-applied`, if this is a notification (kind = notification) and
@@ -70,7 +71,7 @@ class ModerationCase extends db.ExpandoModel<String> {
   ///   where we applied content moderation, and we decided to uphold that decision.
   /// - `moderation-reverted`, if this is an appeal (kind = appeal) of a notification
   ///   where we applied content moderation, and we decided to revert that decision.
-  @db.StringProperty()
+  @db.StringProperty(required: true)
   String? resolution;
 
   ModerationCase();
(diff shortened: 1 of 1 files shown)


app/lib/admin/models.dart
```diff
@@ -0,0 +1,87 @@
...
+  /// - `moderation-applied`, if this is a notification (kind = notification) and
+  ///   we decided to apply content moderation.
+  /// - `no-action-upheld`, if this is an appeal (kind = appeal) of a notification
+  ///   where we took no-action, and we decided to uphold that decision.
+  /// - `no-action-reverted`, if this is an appeal (kind = appeal) of a notification
+  ///   where we took no-action, and we decided to revert that decision.
+  /// - `moderation-upheld`, if this is an appeal (kind = appeal) of a notification
+  ///   where we applied content moderation, and we decided to uphold that decision.
+  /// - `moderation-reverted`, if this is an appeal (kind = appeal) of a notification
+  ///   where we applied content moderation, and we decided to revert that decision.
```
sigurdm, line 72: Perhaps document the `null` case.

sigurdm (approved):

app/lib/admin/models.dart
sigurdm: SG

app/lib/admin/models.dart
```diff
@@ -0,0 +1,87 @@
...
+  late DateTime opened;
+
+  /// The case was resolved at this time (or null if it has not been resolved yet).
+  @db.DateTimeProperty()
+  DateTime? resolved;
+
+  /// The source of the case, one of:
+  /// - `external-notification`,
+  /// - `internal-notification` (only used for reports from @google.com accounts), or,
+  /// - `automated-detection. (will not be used)
```
jonasfj, line 36:
```suggestion
  /// - `automated-detection`. (will not be used)
```

app/lib/admin/models.dart
jonasfj: IF we want a pending thing, we should rename it `status`.

For querying it's probably fine to have a `pending` value.

jonasfj (approved):

isoos:
> IF we want a pending thing, we should rename it status.

Maybe it is better to reduce the possible values to `pending`, `no-action`, `upheld`, `reverted`, without the `${kind}` classification?

PR #7652, opened 2024-04-17
Model + refactoring for canonical topics yaml file

isoos:
#7263


app/lib/service/topics/models.dart
```diff
@@ -0,0 +1,106 @@
...
+  checked: true,
+  explicitToJson: true,
+  disallowUnrecognizedKeys: true,
+)
+class CanonicalTopicFileContent {
+  final List<CanonicalTopic> topics;
+
+  CanonicalTopicFileContent({
+    required this.topics,
+  });
```
jonasfj, line 43: Why not?
```suggestion
  CanonicalTopicFileContent({
    required this.topics,
  }) {
    verifyContent();
  }
```
possibly even just copy pasting the checks into this constructor, so we don't need a `verifyContent` method.

jonasfj (approved):

app/lib/service/topics/models.dart
isoos: Done.

PR #7658, opened 2024-04-18
ModerationCase: replace subject<Package|Publisher> fields with subject[Kind|LocalName|Fqn]

isoos:
- #7535
- This is an alternative design proposal, that also addresses the use case of referencing a specific `PackageVersion` for moderation.

### DIFF
--- a/app/lib/admin/models.dart
+++ b/app/lib/admin/models.dart
@@ -44,21 +44,21 @@ class ModerationCase extends db.ExpandoModel<String> {
   late String kind;
 
   /// The kind of the entity this notification or appeal concerns. On of:
-  /// `Package`, `PackageVersion` or `Publisher`.
+  /// `package`, `package-version` or `publisher`.
   @db.StringProperty()
   String? subjectKind;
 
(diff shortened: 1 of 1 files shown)


app/lib/admin/models.dart
```diff
@@ -43,15 +43,23 @@ class ModerationCase extends db.ExpandoModel<String> {
...
+  /// - `Publisher:<publisherId>`
   @db.StringProperty()
-  String? subjectPublisher;
+  String? subjectFqn;
+
+  /// The local name of the entity (without the type qualifier) this notification or appeal concerns.
+  /// - `Package`: the package name
+  /// - `PackageVersion`: the `<package>/<version>`
+  /// - `Publisher`: the publisher ID
+  String? subjectLocalName;
```
jonasfj, line 62: Curious, do we think this is better?

It'll mean that we need composite indexes? hmm, I guess `subjectFqn` will work.

hmm, I suppose this will make it easier to search for things...

jonasfj (approved):

PR #7661, opened 2024-04-19
Include email messageId in sending log entry.

isoos:


app/lib/frontend/email_sender.dart
```diff
@@ -135,7 +135,7 @@ class _GmailSmtpRelay implements EmailSender {
 
   @override
   Future<void> sendMessage(EmailMessage message) async {
-    final debugHeader = '(${message.subject}) '
+    final debugHeader = '${message.localMessageId} (${message.subject}) '
```
jonasfj, line 138:
```suggestion
    final debugHeader = 'Message-ID:${message.localMessageId} (${message.subject}) '
```

jonasfj (approved):

PR #7663, opened 2024-04-19
Use GCP SecretManager

jonasfj:
Let's not merge until secrets have been created.


app/lib/service/secret/backend.dart
```diff
@@ -2,68 +2,126 @@
...
+    Duration maxAge = const Duration(hours: 1),
+  });
+}
+
+final class GcpSecretBackend extends SecretBackend {
+  final http.Client _client;
+  // Map from cached secret id to:
+  // * time it was fetched, and
+  // * future secret
+  final _cache = <String, (DateTime, Future<String?>)>{};
```
isoos, line 47: Aside: I don't mind this cache pattern, but this is yet another pattern we use for caching. May be worth to migrate/consolidate the different patterns (adding a TODO?).

isoos (commented):
lgtm

app/lib/service/secret/backend.dart
isoos: Not particularly this (hence the yet another part), and it may be worth to consolidate the other patterns to use this one.

jonasfj: Yeah, the `maxAge` parameter is pretty cool.

I don't know if we want to wrap this in a nice abstraction, maybe we should do that at some point and write some pure tests for the caching functionality. It's sadly pretty popular to get caching wrong by switching less-than for greater-than or vice versus :see_no_evil: :see_no_evil: :see_no_evil: 

Even with review, and with `.before` and `.after` we frequently get caching wrong :see_no_evil:

app/test/admin/api_actions_test.dart
```diff
@@ -155,28 +155,4 @@ void main() {
         .uploadPackageBytes(bytes);
     expect(rs2.success.message, contains('Successfully uploaded'));
   });
-
-  testWithProfile('setting and viewing secrets', fn: () async {
```
isoos, line 159: Not sure: what was wrong with this test?

isoos (approved):

app/test/admin/api_actions_test.dart
jonasfj: You can set secrets when we are using the secret manager, the action to interact with secrets is gone.

PR #7666, opened 2024-04-19
Rewrite and redirect canonical topic aliases.

isoos:
#7263

### DIFF
--- a/app/lib/package/model_properties.dart
+++ b/app/lib/package/model_properties.dart
@@ -14,6 +14,7 @@ import 'package:pubspec_parse/pubspec_parse.dart' as pubspek
     show Dependency, Pubspec;
 import 'package:yaml/yaml.dart';
 
+import '../../service/topics/models.dart';
 import '../shared/datastore.dart';
 import '../shared/utils.dart' show canonicalizeVersion;
 import '../shared/versions.dart' as versions;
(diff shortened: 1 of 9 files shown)


jonasfj (approved):

app/lib/package/model_properties.dart
```diff
@@ -198,7 +204,7 @@ class Pubspec {
   }
 
   late final List<Uri> funding = _inner.funding ?? const <Uri>[];
-  late final hasTopic = topics?.isNotEmpty ?? false;
+  late final hasTopic = canonicalizedTopics.isNotEmpty;
```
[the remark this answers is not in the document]
isoos: I think both names are valid (as long as it is consistent with other parts of the code). IIRC we don't really have much similar use cases, or we don't have `hasX[s]` methods for them, so we may just choose any variant. I don't mind either way, any preference?

app/lib/search/backend.dart
```diff
@@ -522,6 +523,30 @@ class SearchBackend {
   }
 }
 
+/// Returns a new search form that may overrides predicates to their canonical forms.
```
szakarias, line 526: overrides -> override

pkg/_pub_shared/lib/search/search_form.dart
```diff
@@ -123,6 +123,24 @@ class TagsPredicate {
     return p;
   }
 
+  /// Update the predicates by using their canonical version (if there is any available).
+  /// Does not change the predicates status (unless the canonical one is also present,
+  /// in this case the behavior is undefinited).
```
szakarias, line 128: undefinited -> undefined

szakarias (approved):
Nice!

PR #7670, opened 2024-04-23
Migrate pubSiteAudience -> pubServerAudience.

### LINKED ISSUE #7665: Cleanup leftover `pubSiteAudience`
I took a first step at this in https://github.com/dart-lang/pub-dev/pull/7664

But we probably need to rename a few things before we're all done.

isoos:
Fixes #7665.

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -538,7 +538,7 @@ class AccountBackend {
       return null;
     }
     return AuthenticatedUser(user,
-        audience: activeConfiguration.pubSiteAudience!);
+        audience: activeConfiguration.pubServerAudience!);
   }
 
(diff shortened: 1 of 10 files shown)


app/lib/account/backend.dart
```diff
@@ -538,7 +538,7 @@ class AccountBackend {
       return null;
     }
     return AuthenticatedUser(user,
-        audience: activeConfiguration.pubSiteAudience!);
+        audience: activeConfiguration.pubServerAudience!);
```
jonasfj, line 541: This was sort of a bug, right? are there any implications?

jonasfj (approved):

app/lib/account/backend.dart
isoos: No implication, it was the same value.

PR #7672, opened 2024-04-24
Report page with subject parameter + its form processing.

isoos:
#7535

### DIFF
--- a/app/lib/admin/models.dart
+++ b/app/lib/admin/models.dart
@@ -93,11 +93,27 @@ abstract class ModerationStatus {
   static const pending = 'pending';
 }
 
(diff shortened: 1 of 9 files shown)


app/lib/admin/models.dart
```diff
@@ -115,3 +97,71 @@ abstract class ModerationKind {
 abstract class ModerationStatus {
   static const pending = 'pending';
 }
+
+class ModerationSubject {
+  final String kind;
+  final String localName;
+  final String? package;
+  final String? version;
+  final String? publisherId;
```
jonasfj, line 106: Documentation comments would be nice later

jonasfj (approved):

PR #7676, opened 2024-04-26
Remove reference to obsolete dartdoc bucket.

isoos:


sigurdm (approved):
@jonasfj do we need to delete these buckets? Or are they already gone?

jonasfj (approved):

PR #7677, opened 2024-04-26
Do not try to use unconfigured search endpoint in local tests.

### LINKED ISSUE #7608: Slow tests in test/package/upload_test.dart

isoos:
- Fixes #7608.
- The attempt to connect to `http://localhost:0` timed out after ~2-2.5 seconds, making most of the test cases run longer than needed.

### DIFF
--- a/app/lib/search/search_client.dart
+++ b/app/lib/search/search_client.dart
@@ -61,6 +61,12 @@ class SearchClient {
     // returns null on timeout (after 5 seconds)
     Future<http.Response?> doCallHttpServiceEndpoint({String? prefix}) async {
       final httpHostPort = prefix ?? activeConfiguration.searchServicePrefix;
+      if (httpHostPort == null) {
+        // Skip calling the HTTP endpoint if no value was configured.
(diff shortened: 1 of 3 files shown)


isoos:
Hm, apparently way too many test start to fail with this workaround, I may be missing something. Closing this as it needs further investigation.

PR #7681, opened 2024-04-29
Run search service behind HTTP server for tests.

### LINKED ISSUE #7608: Slow tests in test/package/upload_test.dart

isoos:
- Fixes #7608.
- Cuts down waiting for search latencies (2+ seconds in each test case).
- Needed to change the fake sign-in redirect to `/help`, as otherwise it would have populated the caches for the landing page, breaking the landing page tests.
- Removed the test using mock search service, as it doesn't add much protection, the internals of search changed and it is no longer that useful.
- Moved topPackages and youtube service initialization to a later point in the service initialization, to prevent early cache population.

### DIFF
--- a/app/test/frontend/handlers/landing_test.dart
+++ b/app/test/frontend/handlers/landing_test.dart
@@ -2,9 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'package:http/testing.dart';
 import 'package:pub_dev/frontend/static_files.dart';
-import 'package:pub_dev/search/search_client.dart';
 import 'package:pub_dev/search/top_packages.dart';
 import 'package:test/test.dart';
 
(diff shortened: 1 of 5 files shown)


app/test/frontend/handlers/landing_test.dart
```diff
@@ -53,23 +51,6 @@ void main() {
       );
     });
 
-    testWithProfile('/ without a working search service', fn: () async {
```
sigurdm, line 56: Is this removing the fallback search entirely?

sigurdm (approved):

app/test/frontend/handlers/landing_test.dart
isoos: No, we are keeping that. This test became obsolete somewhere in the past 2-3 months, where we've updated the fallback mechanism and also did the isolate-based search migration. I think this kind of mocking is not that useful in tests, as it does hide structural changes, and now, months after the actual change, it is harder to trace it back.

PR #7686, opened 2024-04-30
Action for inspecting a user

sigurdm:


app/lib/admin/actions/actions.dart
```diff
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:pub_dev/admin/actions/user_info.dart';
```
isoos, line 5: nit: let's use relative links

app/lib/admin/actions/user_info.dart
```diff
@@ -0,0 +1,75 @@
...
+    InvalidInputException.check(
+      userIdOrEmail != null && userIdOrEmail.isNotEmpty,
+      'user must be given',
+    );
+
+    User? user;
+    if (looksLikeUserId(userIdOrEmail!)) {
+      user = await accountBackend.lookupUserById(userIdOrEmail);
+    } else {
+      final users = await accountBackend.lookupUsersByEmail(userIdOrEmail);
```
isoos, line 47: It may be useful here to return all the user entries with the given email. Could be in a follow-up later, maybe add a TODO for it?

isoos (approved):

app/lib/admin/actions/user_info.dart
sigurdm: Updated to report on all users with given email.

app/lib/admin/actions/actions.dart
sigurdm: Yeah - didn't notice that!

jonasfj (approved):

PR #7690, opened 2024-05-01
Folding report page sections.

isoos:
#7535

### DIFF
--- a/app/lib/admin/models.dart
+++ b/app/lib/admin/models.dart
@@ -202,8 +202,8 @@ class ModerationSubject {
   }
 
   late final fqn = '$kind:$localName';
-  bool get hasPackage => package != null;
-  bool get hasPublisher => publisherId != null;
+  bool get isPackage => package != null;
+  bool get isPublisher => publisherId != null;
 }
 
 class ModerationSubjectKind {
--- a/app/lib/admin/models.dart
+++ b/app/lib/admin/models.dart
@@ -202,6 +202,8 @@ class ModerationSubject {
   }
 
   late final fqn = '$kind:$localName';
+  bool get hasPackage => package != null;
+  bool get hasPublisher => publisherId != null;
 }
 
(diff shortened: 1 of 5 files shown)


app/lib/frontend/templates/report.dart
```diff
@@ -2,23 +2,34 @@
...
-  ModerationSubject? subject,
+  required ModerationSubject subject,
 }) {
+  final kindLabel = _subjectKindLabels[subject.kind] ?? 'about';
+  // TODO: also add `url`
+  final lcpsDeepLink =
+      Uri.parse('https://reportcontent.google.com/troubleshooter').replace(
+    queryParameters: {
+      'product': 'dart_pub',
+      'content_id': subject.fqn,
```
jonasfj, line 30: `content_id` should be a canonical URL for the content in question.

One of:

* `pub.dev/publishers/<publisher>`
* `pub.dev/packages/<package>`
* `pub.dev/packages/<package>/versions/<version>`

```suggestion
      'content_id': subject.fqn, // TODO: Use subject.canonicalUrl, also add canonicalUrl
```

app/lib/admin/models.dart
```diff
@@ -189,6 +189,8 @@ class ModerationSubject {
   }
 
   late final fqn = '$kind:$localName';
+  bool get hasPackage => package != null;
+  bool get hasPublisher => publisherId != null;
```
jonasfj, line 193:
```suggestion
  bool get isPublisher => publisherId != null;
```

jonasfj: Less, sure how `hasPackage` should be named :D

jonasfj (approved):

app/lib/admin/models.dart
isoos: Renamed to `isPackage` to keep it consistent.

PR #7693, opened 2024-05-02
Action log entries in ModerationCase.

isoos:
#7535

### DIFF
--- a/app/lib/admin/models.dart
+++ b/app/lib/admin/models.dart
@@ -2,9 +2,14 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:convert';
+
 import 'package:clock/clock.dart';
+import 'package:json_annotation/json_annotation.dart';
 
 import '../shared/datastore.dart' as db;
+
+part 'models.g.dart';
 
 /// Tracks the status of the moderation or appeal case.
 @db.Kind(name: 'ModerationCase', idType: db.IdType.String)
@@ -66,6 +71,10 @@ class ModerationCase extends db.ExpandoModel<String> {
   @db.StringProperty(required: true)
   String? status;
 
+  /// The JSON-encoded array of [ModerationActionLog] entries.
+  @db.StringProperty(propertyName: 'actionLog', indexed: false)
+  String? actionLogField;
+
   ModerationCase();
 
   ModerationCase.init({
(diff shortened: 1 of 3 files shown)


app/lib/admin/models.dart
```diff
@@ -79,6 +88,34 @@ class ModerationCase extends db.ExpandoModel<String> {
...
+    actionLogField = json.encode(value.toJson());
+  }
+
+  /// Adds a new entry after deduplication (skipping the current
+  /// [subject]+[isModerated] pair if an identical exists as the last state).
+  void addActionLogEntryWithDeduplication(String subject, bool isModerated) {
+    final log = getActionLog();
+    final lastEntry = log.entries.where((e) => e.subject == subject).lastOrNull;
+    if (lastEntry != null && lastEntry.isModerated == isModerated) {
+      // duplicate entry found, skip inserting
```
jonasfj, line 110: Only, if date-time is within 10 minutes? We just want retried request to not trigger new entries, right?

In that case, I think we should clearly state why in a comment.

app/lib/admin/models.dart
```diff
@@ -196,3 +233,35 @@ class ModerationSubjectKind {
   static const packageVersion = 'package-version';
   static const publisher = 'publisher';
 }
+
+@JsonSerializable(includeIfNull: false)
+class ModerationActionLog {
+  final List<ModerationActionLogEntry> entries;
```
jonasfj, line 239:
```suggestion
  final List<({DateTime timestamp, String subject, bool applyModeration, bool revertModeration, String message})> entries;
```

How about using records? If it works with json_serializable :see_no_evil:

jonasfj (approved):

app/lib/admin/models.dart
isoos: I'm not sure it is only 10 minutes. It may be somebody starts a restrictive admin action, which is partially completed, as not everything happens in the same transaction. It may be after an hour that the action is repeated.

app/lib/admin/models.dart
isoos: What's the benefit of records here? I think this is more than just a multi-type return value.

isoos: `json_serializable` does support this, however, I still think it is better to keep it in a separate class.

Also: added `message` field, but I'm undecided on the  `applyModeration` / `revertModeration` fields. How about the `enum ModerationAction {apply, revert}`?

jonasfj: I don't have a problem with an enum

app/lib/admin/models.dart
jonasfj: let's make it 10-15 minutes, if someone really does reapply it, what is the harm of having two entries?

We kind of want to have the date-time from both.

I think it's mostly worth deduplicating in cases where retries redo the operation.

-----------------

Actually, unless harmful I suggest we consider not deduplicating at all. If we need to do undo a restrictive action due to appeal, that will happen in a different `ModerationCase`.

PR #7696, opened 2024-05-02
Include canonical URL as 'url' attribute on the link to report page.

isoos:
#7535

### DIFF
--- a/app/lib/frontend/templates/views/shared/layout.dart
+++ b/app/lib/frontend/templates/views/shared/layout.dart
@@ -230,7 +230,10 @@ d.Node pageLayoutNode({
               ),
 
             d.element('main', classes: mainClasses, child: mainContent),
-            _siteFooterNode(moderationSubject: moderationSubject),
+            _siteFooterNode(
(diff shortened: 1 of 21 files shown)


app/lib/frontend/templates/views/shared/layout.dart
```diff
@@ -250,6 +253,7 @@ d.Node pageLayoutNode({
 
 /// Renders the footer content.
```
jonasfj, line 254:
```suggestion
/// Renders the footer content.
///
/// The [canonicalUrl] must be the canonical URL for the current page.
```

Is this correct?

jonasfj: Just to clarify that it's not the canonical URL for the content in question.

This is the URL that made the user click "report..." button/link.

jonasfj (approved):

app/lib/frontend/templates/views/shared/layout.dart
isoos: Yes, that is correct.

PR #7700, opened 2024-05-03
Display topic descriptions.

isoos:
- #7263
- `/topics` page with separate paragraph
- listing and package pages with the `title` attribute

PR #7706, opened 2024-05-06
Fix topics text/title and separate CSS styles.

### LINKED ISSUE #7703: Reduce the margin between Topics header and the first item

isoos:
- #7263
- Fixes a bug in #7700: the description was displayed as part of the `text` instead of the `title` attribute.
- Fixes #7703 by reducing the extra margin between the title tag and the topic block.
- Separates CSS styles from package listing styles, also simpler styles for `topic-item`.

### DIFF
--- a/app/lib/frontend/templates/views/page/topics_list.dart
+++ b/app/lib/frontend/templates/views/page/topics_list.dart
@@ -6,32 +6,26 @@ import '../../../../service/topics/models.dart';
 import '../../../../shared/urls.dart' as urls;
 import '../../../dom/dom.dart' as d;
 
+/// Renders the topics list node of /topics page.
 d.Node renderTopicsList(Map<String, int> topics) {
-  return d.div(children: [
-    d.h1(text: 'Topics '),
-    topicsListNode(topics),
-  ]);
-}
-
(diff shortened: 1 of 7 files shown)


sigurdm:
Screenshot plz!?

sigurdm (approved):

isoos:
screenshot:
image

(also on staging)

sigurdm:
Hmm - still too much distance between title and list items.

isoos:
> Hmm - still too much distance between title and list items.

It is coming from the bottom margin of the `h1` title, there are no other styles that make the gap larger. Unless we want to change the site-wide look, I'd keep it this way.

sigurdm:
On the front page there doesn't seem to be this issue:
image

Perhaps we do something specific in _home.scss
image

isoos:
> On the front page there doesn't seem to be this issue:

Good point. Should we also reduce the gap on e.g. help pages?
https://staging.pub.dev/help

sigurdm:
It somehow is less obvious on the help page, but IMO we should remove it there as well.
Current:
image

Removed:
image

PR #7708, opened 2024-05-07
Include URL in the reported entity.

isoos:
#7535

### DIFF
--- a/app/lib/frontend/templates/report.dart
+++ b/app/lib/frontend/templates/report.dart
@@ -20,6 +20,7 @@ const _subjectKindLabels = {
 String renderReportPage({
   SessionData? sessionData,
   required ModerationSubject subject,
+  required String? url,
 }) {
   final kindLabel = _subjectKindLabels[subject.kind] ?? 'about';
   // TODO: also add `url`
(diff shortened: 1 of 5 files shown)


app/lib/frontend/templates/report.dart
```diff
@@ -20,6 +20,7 @@ const _subjectKindLabels = {
 String renderReportPage({
   SessionData? sessionData,
   required ModerationSubject subject,
+  required String? url,
```
jonasfj, line 23: This should also be added to `lcpsDeepLink` as `&url=` in querystring.

jonasfj (approved):

app/lib/frontend/templates/report.dart
isoos: I think we need `subject.canonicalUrl` for that: https://github.com/dart-lang/pub-dev/blob/master/app/lib/frontend/templates/report.dart#L31
(Also: I have an almost ready PR for that, submitting shortly)

isoos: Ah, misremembered this, follow-up in: https://github.com/dart-lang/pub-dev/pull/7713

PR #7710, opened 2024-05-07
Moderate package action stores event on ModerationCase.

isoos:
#7535

### DIFF
--- a/app/lib/admin/actions/moderate_package.dart
+++ b/app/lib/admin/actions/moderate_package.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import '../../admin/backend.dart';
+import '../../admin/models.dart';
 import '../../package/backend.dart';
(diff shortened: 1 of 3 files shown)


app/lib/admin/actions/moderate_package.dart
```diff
@@ -16,11 +18,19 @@ final moderatePackage = AdminAction(
 Set the moderated flag on a package (updating the flag and the timestamp).
 ''',
   options: {
+    'case': 'The ModerationCase.caseId that this action is part of.',
     'package': 'The package name to be moderated',
     'state':
         'Set moderated state true / false. Returns current state if omitted.',
+    'message': 'Optional message to store and send out.'
```
jonasfj, line 25:
```suggestion
    'message': 'Optional message to store.'
```

Are we sending it to anyone?

jonasfj (approved):

app/lib/admin/actions/moderate_package.dart
isoos: Not yet, but I though we should if there is a message.

PR #7712, opened 2024-05-08
Reduced bottom margin on topics page.

isoos:
This is a follow-up to #7706.

I've considered a generic rule for all `h1` elements on standalone (e.g. non-user-content related) pages, but it feels like it is better to implement this on a case-by-case basis, and keep the rule close to the location it applies.


jonasfj (approved):
Please include a screenshot if this changes much

isoos:
> Please include a screenshot if this changes much

we display ~8 lines less 🙈

PR #7714, opened 2024-05-09
Report page rendering and form submission for appealing a resolution.

isoos:
- #7535
- The appeal page is simpler than the report page, we only have the form -> there are no folding parts.
- It shares a lot of logic with the report page (also the endpoint and tests), the new tests concentrate only on the additional parts and verification of the appeal page rendering and form submission.


isoos:
Note: this is not directly tied into the rest of the moderation flow yet, as we need the logic to send out the emails with the appeal URLs.

jonasfj (approved):

PR #7719, opened 2024-05-13
Require caseId when moderating a publisher.

isoos:
#7535


jonasfj (approved):
I'm not sure it wise to require a case, there could be things happening in other systems, where we're not responsible for transparency reporting.

In particularly, illegal content moderation stuff is not on us, and we won't necessarily have a case-id in such instances.

isoos:
> In particularly, illegal content moderation stuff is not on us, and we won't necessarily have a case-id in such instances.

I think it is still useful to have the same requirements for these flows, but we may skip reporting when we are closing the case.

PR #7726, opened 2024-05-14
Added action for package-version-retraction

jonasfj:


app/lib/admin/actions/package_version_retraction.dart
```diff
@@ -0,0 +1,96 @@
...
+
+      final pkg = await packageBackend.lookupPackage(packageName);
+      if (pkg == null) {
+        throw NotFoundException.resource(packageName);
+      }
+
+      final packageVersion =
+          await packageBackend.lookupPackageVersion(packageName, version);
+      if (packageVersion == null) {
+        throw InvalidInputException('No such package version');
```
isoos, line 52: Nit: I think this could be also a NotFoundException

app/lib/admin/actions/package_version_retraction.dart
```diff
@@ -0,0 +1,96 @@
...
+
+        if (isRetracted != pv.isRetracted) {
+          await packageBackend.doUpdateRetractedStatus(
+            SupportAgent(),
+            tx,
+            p,
+            pv,
+            isRetracted,
+          );
+        }
```
sigurdm, line 84: I think (but am not sure) that you want to call `await purgePackageCache(package);`

isoos (commented):
looks good so far

app/lib/admin/actions/package_version_retraction.dart
sigurdm: we do that here: https://github.com/dart-lang/pub-dev/blob/ca933a0d0349f900b1752dfa3faf6d29a8d93fa6/app/lib/admin/backend.dart#L416

jonasfj: Found that in testing :D

app/lib/admin/actions/package_version_retraction.dart
jonasfj: Done!

sigurdm (approved):

isoos (approved):

PR #7732, opened 2024-05-15
Get the final status of moderated subjects (also filtering subsets).

isoos:
#7535

### DIFF
--- a/app/lib/admin/models.dart
+++ b/app/lib/admin/models.dart
@@ -265,6 +265,25 @@ class ModerationSubject {
     }
     throw UnimplementedError('Unknown subject kind: "$fqn"');
   }();
+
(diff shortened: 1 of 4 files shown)


app/lib/admin/models.dart
```diff
@@ -286,6 +305,27 @@ class ModerationActionLog {
       _$ModerationActionLogFromJson(json);
 
   Map<String, Object?> toJson() => _$ModerationActionLogToJson(this);
+
+  /// Returns the list of subjects that have a final state as moderated.
+  /// If a package version and also a package is moderated, returns
+  /// only the package.
+  List<String> moderatedSubjects() {
```
jonasfj, line 312: Find state is difficult. Let's not try to do that.
Let's just return the log and let humans read it?

PR #7734, opened 2024-05-16
Admin action to send email to recipients.

isoos:
#7535

### DIFF
--- a/app/lib/admin/actions/send_email.dart
+++ b/app/lib/admin/actions/send_email.dart
@@ -24,9 +24,8 @@ emails to send out messages. Uses the provided `email-subject` and `email-body`
 without changing anything in it.
 ''',
   options: {
-    'to-email': 'A comma separated list of email addresses.',
-    'to-subject-admin':
-        'A comma separated list of subjects where the admins will get the emails.',
+    'to':
+        'A comma separated list of email addresses or subjects (where the admins will get the emails).',
     'from': 'The email address to impersonate (`support@pub.dev` by default).',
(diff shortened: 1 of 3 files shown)


app/lib/admin/actions/send_email.dart
```diff
@@ -0,0 +1,105 @@
...
+emails to send out messages. Uses the provided `email-subject` and `email-body`
+without changing anything in it.
+''',
+  options: {
+    'to-email': 'A comma separated list of email addresses.',
+    'to-subject-admin':
+        'A comma separated list of subjects where the admins will get the emails.',
+    'from': 'The email address to impersonate (`support@pub.dev` by default).',
+    'email-subject': 'The subject of the email message.',
+    'email-body': 'The text content of the email body.',
```
jonasfj, line 32:
```suggestion
    'subject': 'The subject of the email message.',
    'body': 'The text content of the email body.',
```

jonasfj (approved):

app/lib/admin/actions/send_email.dart
isoos: I've found the plain `subject` to be too close to confuse with the moderation subjects, and opted for the longer names. We could keep it short, but in that case I'd suggest to use only single `to` parameter, and resolve the emails and subjects within that. wdyt?

isoos: Updated the PR with the commits: one for dropping the prefix, the other to merge the `to-*` parameters. PTAL!

Issue #7725, opened 2024-05-14
Pass `SupportAgent` into all admin actions


jonasfj:
We should also ensure that SupportAgent isn't affected by rate-limits

isoos:
I've investigated the code in admin actions, and it seems that the most difficult part would be to rewrite some of the consent models and audit records related to invites and their handling. They require an active `User` marking from where the invite originates from (storing the `userId` on the `Consent` entity), and this id is used in multiple places.

PR #7737, opened 2024-05-20
Use the admin support email as agent in consent rejection and expiry audit logs.

isoos:
- Found this during the investigation of #7725. This has very little practical effect, but reduces the importance of using `User.userId` in invites.
- For these we were using the userId of the inviting `User` account as an agent (directly or fallback).
- The expiry happens without any real user interaction, using the `support@pub.dev` as placeholder seems to be better.
- For rejection we were using the signed in account's `userId` as primary agent, but in case we implement rejection without signing in, the fallback should be `support@pub.dev` (or something else) instead of the inviting `userId`.
- Instead of `support@pub.dev` we could use another user-friendly agent alias, but we may as well reuse it.

### DIFF
--- a/app/lib/audit/models.dart
+++ b/app/lib/audit/models.dart
@@ -464,21 +464,21 @@ class AuditLogRecord extends db.ExpandoModel<String> {
   ///
   /// Throws [RateLimitException] when the configured rate limit is reached.
   static Future<AuditLogRecord> publisherContactInviteExpired({
-    required String fromUserId,
+    required String? fromUserId,
     required String publisherId,
     required String contactEmail,
   }) {
     return _checkRateLimit(AuditLogRecord._init()
       ..kind = AuditLogRecordKind.publisherContactInviteExpired
-      ..agent = fromUserId
+      ..agent = KnownAgents.pubSupport
       ..summary = 'Contact invite for publisher `$publisherId` expired, '
           '`$contactEmail` did not respond.'
       ..data = {
-        'fromUserId': fromUserId,
+        if (fromUserId != null) 'fromUserId': fromUserId,
         'publisherId': publisherId,
         'contactEmail': contactEmail,
       }
-      ..users = [fromUserId]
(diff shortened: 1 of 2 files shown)


app/lib/audit/models.dart
```diff
@@ -589,21 +592,21 @@ class AuditLogRecord extends db.ExpandoModel<String> {
...
   static Future<AuditLogRecord> publisherMemberInviteExpired({
-    required String fromUserId,
+    required String? fromUserId,
     required String publisherId,
     required String memberEmail,
   }) {
     return _checkRateLimit(AuditLogRecord._init()
       ..kind = AuditLogRecordKind.publisherMemberInviteExpired
-      ..agent = fromUserId
+      ..agent = KnownAgents.pubSupport
```
jonasfj, line 601: I would prefer we make a different alias, we can just invent one.

It's kind of nice that `support@pub.dev` means there was some action by our support team. Some admin action or something.

Maybe for these things we could just write `pub.dev` as the agent name? Because it's the service itself, right?

jonasfj (approved):

app/lib/audit/models.dart
isoos: I don't mind either way. The benefit of `support@pub.dev` is that we implicitly communicate the contact address if something is amiss, while in the case of `pub.dev`, while it is short and to the point, needs further search to get to an email. Should I add `pub.dev` to the fixed list of known agents?

isoos: Note: we already use `support@pub.dev` in `packageRemovedFromPublisher`. Should we also migrate it to `pub.dev`?

isoos: On a revised note: let's merge this as-is, it improves the state of the audit log records, and we may decide to migrate some of it to `pub.dev` (with the appropriate format verification changes).

jonasfj: ship it

PR #7742, opened 2024-05-21
Use different example/generated URLs to unblock CI.

isoos:


isoos:
I think this should fix the CI.

PR #7746, opened 2024-05-22
New fields for github automated publishing events.

isoos:
- These are alternative names to the ones suggested in #7462
- They don't require `true` as default value, and don't need migration.
- They are closer in naming scheme to the already existing fields there (e.g. `requireEnvironment`).


pkg/_pub_shared/lib/data/package_api.dart
```diff
@@ -90,18 +90,28 @@ class GithubPublishingConfig {
   /// The GitHub environment where the publishing is required from.
   String? environment;
 
+  /// Whether the (default) `push` event is disabled.
+  bool? disablePushEvent;
+
+  /// Whether the `workflow_dispatch` event is enabled.
+  bool? enableWorkflowDispatchEvent;
```
jonasfj, line 97: Let's give all the boolean value default values

isoos: PTAL.

pkg/_pub_shared/lib/data/package_api.dart
```diff
@@ -67,7 +67,7 @@ class AutomatedPublishingConfig {
 
 @JsonSerializable(includeIfNull: false, explicitToJson: true)
 class GithubPublishingConfig {
-  bool? isEnabled;
+  late bool isEnabled;
```
jonasfj, line 70: Why is it late?

pkg/_pub_shared/lib/data/package_api.dart
```diff
@@ -85,17 +85,25 @@ class GithubPublishingConfig {
...
 
+  /// Whether the (default) `push` event is enabled.
+  late bool isPushEventEnabled;
+
+  /// Whether the `workflow_dispatch` event is enabled.
+  late bool isWorkflowDispatchEventEnabled;
+
   GithubPublishingConfig({
-    this.isEnabled,
+    this.isEnabled = false,
```
jonasfj, line 100: Have you considered using `@JsonKey(disallowNullValue: true, defaultValue: false)` annotation instead?

jonasfj (approved):
I'm guessing we need a separate backfill too for the new property.

pkg/_pub_shared/lib/data/package_api.dart
isoos: `disallowNullValue` would enforce us to do a migration. In its current form we may just use it as-is, and we may or may not do a migration, it is just fine to let the migration happen for users that change their settings in any way.

jonasfj: Ah, yeah, that's fine. Let's not use `disallowNullValue` :rofl: 


-------------

I do think it'd be wise to do a migration eventually, it easily becomes much harder to not make mistakes if we don't.

pkg/_pub_shared/lib/data/package_api.dart
```diff
@@ -85,17 +85,25 @@ class GithubPublishingConfig {
...
     this.repository,
     this.tagPattern,
-    this.requireEnvironment,
+    this.requireEnvironment = false,
     this.environment,
+    this.isPushEventEnabled = true,
+    this.isWorkflowDispatchEventEnabled = false,
```
jonasfj, line 106:
```suggestion
    // MUST default to true, because this was the default behavior before
    // isWorkflowDispatchEventEnabled was introduced!
    this.isPushEventEnabled = true,
    // MUST default to false, because the default behavior prior to the
    // introduction of isWorkflowDispatchEventEnabled was to only allow
    // publishing from push events.
    this.isWorkflowDispatchEventEnabled = false,
```

PR #7757, opened 2024-05-27
Fix Flutter archive test.

isoos:
- Fixes the test to also check stable release date.
- Increased both time range to 45 days.


pkg/_pub_shared/test/utils/flutter_archive_test.dart
```diff
@@ -10,15 +10,14 @@ void main() {
     final archive = await fetchFlutterArchive();
     expect(archive!.latestStable, isNotNull);
     expect(archive.latestBeta, isNotNull);
-    // we usually have a new stable release in every 3-5 weeks
+    // we usually have a new releases in every 3-5 weeks
```
sigurdm, line 13:
```suggestion
    // we usually have new releases every 3-5 weeks
```

sigurdm (approved):

PR #7758, opened 2024-05-27
Add datastructure for download counts

szakarias:


app/lib/service/download_counts/download_counts.dart
```diff
@@ -0,0 +1,151 @@
...
+  final versionCounts = <({String versionRange, List<int> counts})>[];
+  final patchCounts = <({String versionRange, List<int> counts})>[];
+  final minorCounts = <({String versionRange, List<int> counts})>[];
+  final majorCounts = <({String versionRange, List<int> counts})>[];
+  final totalCounts = <int>[];
+
+  // Last date with processed data.
+  // The date only contains year, month and date.
+  // Hours, minutes and seconds are disregarded.
+  DateTime? lastDate;
```
jonasfj, line 23: Each `xxxCounts` list contains a list of integers with a count for each day.

If we have `totalCounts = [1, 2, 3]`

Then how many downloads did we have on `lastDate`? 1 or 3?
What date did we have 2 downloads on? is `lastDate - 1` or `lastDate + 1`?

I think it's worth documenting how this works.. maybe with an illustrative example:
```
/// ```
/// totalCounts = [42, 21, 55]
///                 ^   ^   ^____ Download count on lastDate
///                 |   |________ Download count on lastDate - 1 day
///                 |____________ Download count on lastDate - 2 days
```

Or maybe it's the other way around, but an illustration like this would help.

app/lib/service/download_counts/download_counts.dart
```diff
@@ -0,0 +1,151 @@
...
+
+@JsonSerializable(includeIfNull: false)
+class CountData {
+  // We store at most two years of data, one data point per day.
+  static const countsCapacity = 731;
+  final versionCounts = <({String versionRange, List<int> counts})>[];
+  final patchCounts = <({String versionRange, List<int> counts})>[];
+  final minorCounts = <({String versionRange, List<int> counts})>[];
+  final majorCounts = <({String versionRange, List<int> counts})>[];
+  final totalCounts = <int>[];
```
jonasfj, line 18: Should this not default to `List.filled(countsCapacity, 0)` or something like that?

app/lib/service/download_counts/download_counts.dart
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  void _prepareDates(
+    DateTime date,
+    DateTime workingLastDate,
+    Map<String, int> dayCounts,
+    int countsIndex,
+    List<({List<int> counts, String versionRange})> versionCounts,
+  ) {
+    if (date.isAfter(DateTime.now())) {
```
jonasfj, line 54:
```suggestion
    if (date.isAfter(clock.now())) {
```

for better testing

jonasfj: also I'd suggest throwing an `ArgumentError`. Isn't it fair to call this a serious violation of our invariants. We probably want things to make noise.

app/lib/service/download_counts/download_counts.dart
jonasfj: Also didn't we want to add a boolean list for tracking which days we do have data from? instead of using -1.

app/lib/service/download_counts/download_counts.dart
```diff
@@ -0,0 +1,151 @@
...
+import 'package:basics/basics.dart';
+import 'package:json_annotation/json_annotation.dart';
+import 'package:pub_semver/pub_semver.dart';
+part 'download_counts.g.dart';
+
+@JsonSerializable(includeIfNull: false)
+class CountData {
+  // We store at most two years of data, one data point per day.
+  static const countsCapacity = 731;
+  final versionCounts = <({String versionRange, List<int> counts})>[];
```
jonasfj, line 14: What is `versionCounts`?

szakarias: A separate list for prereleases, but I was considering removing it. Let me do that for now.

app/lib/service/download_counts/download_counts.dart
szakarias: I removed this check as well. I can add it again later if we think it should be there.

app/lib/service/download_counts/download_counts.dart
szakarias: Added doc.

app/lib/service/download_counts/download_counts.dart
szakarias: Yes, the plan is to keep a bit vector as well. Not included in this PR.

app/lib/service/download_counts/download_counts.dart
```diff
@@ -0,0 +1,152 @@
...
+  // Hours, minutes and seconds are disregarded.
+  DateTime? lastDate;
+
+  /// A list of tuples containing a string describing a major `versionRange`,
+  /// e.g. '>=1.0.0-0 <2.0.0', and a list of integers with a count for each day.
+  /// The 'counts' list contains at most [maxAge] entries. The first entry in
+  /// represents the number of downloads on `lastDate` followed by the downloads
+  ///  on `lastDate` - 1 and so on. E.g.
+  ///
+  ///  counts = [ 42, 21, 55 ]
```
sigurdm, line 29: Beautiful diagram!

app/lib/service/download_counts/download_counts.dart
```diff
@@ -0,0 +1,152 @@
...
+    } else if (date.isAtOrBefore(workingLastDate)) {
+      versionCounts.forEach((versionCount) {
+        if (versionCount.counts.length < countsIndex + 1) {
+          // Fill in with 0 on older days with no data.
+          versionCount.counts.addAll(
+              List.filled(countsIndex - versionCount.counts.length + 1, 0));
+        }
+
+        // Reset the counts for this date.
+        versionCount.counts[countsIndex] = 0;
```
sigurdm, line 94: Not sure I'm reading this right. Should this not be done in an `else` branch? Such that you have:
```
if (date.isAfter(workingLastDate)) {
  // ... prepend zeros ...
} else if (date.isAtOrBefore(workingLastDate)) {
  // ... append zeros ...
} else {
  // reset counts for this date.
}
```

app/lib/service/download_counts/download_counts.dart
```diff
@@ -0,0 +1,152 @@
...
+    }
+  }
+
+  void _processCounts(
+    Map<String, int> dayCounts,
+    int countsIndex,
+    List<({List<int> counts, String versionRange})> versionCounts,
+    VersionRange Function(Version version) createVersionRange,
+  ) {
+    final countsLength = countsIndex + 1;
```
sigurdm, line 105: I'd inline this

sigurdm (approved):

app/lib/service/download_counts/download_counts.dart
szakarias: credits @jonasfj

app/lib/service/download_counts/download_counts.dart
szakarias: done

app/lib/service/download_counts/download_counts.dart
szakarias: It shouldn't as discussed.

PR #7762, opened 2024-05-27
Migrate to pub workspaces

sigurdm:
Activated workspace goodness to be activated.

`dart pub get`, `dart test` and `dart run build_runner build` all seem to work.


sigurdm:
I have now managed to deploy this to staging ... should we try landing it?

PR #7775, opened 2024-05-31
Resolve moderation case - admin action

isoos:
#7535


app/lib/admin/actions/resolve_moderation_case.dart
```diff
@@ -0,0 +1,69 @@
...
+import 'actions.dart';
+
+final resolveModerationCase = AdminAction(
+  name: 'resolve-moderation-case',
+  summary: 'Closes the moderation case and resolves a final status.',
+  description: '''
+Closes the moderation case and updates the status based on the actions logged on the case.
+''',
+  options: {
+    'case': 'The caseId to be closed.',
```
jonasfj, line 18: I think we should consider a `--resolution` argument as well, we can make it optional and then infer by default.

jonasfj (approved):

Issue #4416, opened 2021-01-17
Dark mode for pub.dev


vincevargadev:
It would be great if the pub.dev site supported dark mode (either with or without a switch).

Reading the site at night is annoying, especially if all other sites I need for development already support dark mode (which they do, for example GitHub, YouTube).

As it's (I suppose) a bigger design change, I assume it needs feedback/approval from the design team, but if you would be open to it, I could prepare some of the changes, so most color values can be just plugged in and the feature would get a jump start.

Shweta0617:
@vincevargadev 
I would like to contribute on this issue, please assign to me !!

isoos:
@Shweta0617: I'm happy to help you, but this is not a small undertaking. Would you share what you'd like to get to do?

vincevargadev:
@Shweta0617 I can't assign the issue to you, I just opened the issue.

Shweta0617:
@isoos I tried dark mode in mobile app, but in websites like pub.dev i never try dark mode. But, if someone guide me then probably i complete the issue !!
thanks for asking (i searched today about "implementation of dark theme in website")


@vincevargadev 
Okay, No Problem
can you tell me a dart project in which I can contribute.

Fayiz-A:
### **JUST A SUGGESTION**
[This package](https://pub.dev/packages/dynamic_theme) is really good for implementing dark theme.

vincevargadev:
@Fayiz-A That's a Flutter package, I don't think that would help with the dark theme for pub.dev.

Pub.dev is a simple HTML+CSS style webpage, so the dark theme changes would be implemented mainly with CSS (AFAIK). I think this guide could be a good start: https://css-tricks.com/a-complete-guide-to-dark-mode-on-the-web/

Fayiz-A:
@vincevargadev Sorry for my ignorance. I thought it was written in flutter because I saw dart files in app/lib folder.

gmpassos:
Same situation here. Everything that I use has Dark Mode (GitHub, Terminal, IDE, IntelliJ, Notes...). At night, when we open pub.dev, it's uncomfortable for the eyes.

vandadnp:
This is by no means a perfect solution and only works if you have a browser with Reader mode such as Safari. What I do is just to open a pub.dev package site and go to reader mode to get this nice dark theme.

https://user-images.githubusercontent.com/759524/154893802-da77945f-8f60-4da2-bc44-1a4488630e25.mov

[...]

PR #7784, opened 2024-06-05
Prepare dark-mode experimental flag and top-level style.

isoos:
- #4416
- Note: this PR does not add a button to switch themes, but relies on the `localStorage` flag that the `dartdoc` pages may set when clicking on the dark/light theme button. It is possible that we may introduce a similar button, or we may rely only on the CSS media selectors that are coming from the OS/browser preferences, maybe even both. But for the time being and to explore what kind of SCSS refactoring we may need, this gives an appropriate style isolation and also an easy trigger mechanism.

### DIFF
--- a/app/lib/frontend/handlers/experimental.dart
+++ b/app/lib/frontend/handlers/experimental.dart
@@ -7,6 +7,7 @@ import 'package:meta/meta.dart';
 import '../../shared/cookie_utils.dart';
 
 const _publicFlags = <String>{
+  'dark',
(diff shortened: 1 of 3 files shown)


jonasfj (approved):
As I understand it, switching to dart mode requires, **both**:
 * (1) Enabling the `dark` experiment, AND,
 * (2) Switching to dark-mode on a documentation page.

isoos:
> both

That's correct.

PR #7787, opened 2024-06-06
Download counts: use  to indicate days with no data

szakarias:


app/lib/service/download_counts/download_counts.dart
```diff
@@ -43,7 +43,14 @@ class CountData {
...
   final minorRangeCounts = <VersionRangeCount>[];
   final patchRangeCounts = <VersionRangeCount>[];
-  final totalCounts = List.filled(maxAge, 0, growable: true);
+
+  /// A list of integers representing the total number of daily downloads of any
+  /// version of the package. The list contains at most [maxAge] entries. The
+  /// first entry in represents the total number of downloads on `newestDate`
+  /// followed by the downloads on `newestDate` - 1 and so on.
+  ///
+  /// Days with no data are represented with `-1`.
```
jonasfj, line 52: I think this is an important remark to make in the first comment,

jonasfj (approved):

app/lib/service/download_counts/download_counts.dart
szakarias: I'll do that.

PR #7803, opened 2024-06-11
[Download counts] Update "CountData" constructor to enable json seria…

szakarias:


app/lib/service/download_counts/download_counts.dart
```diff
@@ -71,17 +71,28 @@ class CountData {
...
+  final List<int> totalCounts;
+
+  CountData({
+    required this.majorRangeCounts,
+    required this.minorRangeCounts,
+    required this.patchRangeCounts,
+    required this.totalCounts,
+  });
+
+  factory CountData.init() => CountData(
```
jonasfj, line 91: Why not give it today as the date? life might be simpler :rofl:

jonasfj (approved):

app/lib/service/download_counts/download_counts.dart
szakarias: I am not sure, as discussed. I'll consider it, but not in this change.

PR #7806, opened 2024-06-12
Add model and backend for download counts

szakarias:


app/lib/service/download_counts/backend.dart
```diff
@@ -0,0 +1,52 @@
...
+
+class DownloadCountsBackend {
+  final DatastoreDB _db;
+
+  DownloadCountsBackend(this._db);
+
+  Future<DownloadCounts?> lookupDownloadCounts(String pkg) async {
+    final key = _db.emptyKey.append(DownloadCounts, id: pkg);
+    return _db.lookupOrNull<DownloadCounts>(key);
+  }
```
jonasfj, line 26: Please add redis caching in between

app/lib/service/download_counts/backend.dart
```diff
@@ -0,0 +1,52 @@
...
+
+Future<void> updateDownloadCounts(
+  String pkg,
+  Map<String, int> dayCounts,
+  DateTime dateTime,
+) async {
+  final downloadCounts = await downloadCountsBackend.lookupDownloadCounts(pkg);
+  final countData = downloadCounts?.countData ?? CountData.empty();
+  countData.addDownloadCounts(dayCounts, dateTime);
+  await downloadCountsBackend.ingestDownloadCounts(pkg, countData);
```
jonasfj, line 51: No transactions?

app/lib/service/download_counts/models.dart
```diff
@@ -0,0 +1,39 @@
...
+
+import '../../shared/datastore.dart' as db;
+import 'download_counts.dart';
+
+@db.Kind(name: 'DownloadCounts', idType: db.IdType.String)
+class DownloadCounts extends db.ExpandoModel<String> {
+  String get package => id!;
+
+  @CountDataProperty(required: true)
+  CountData? countData;
```
jonasfj, line 15:
```suggestion
  @CountDataProperty(required: true)
  CountData countData = CountData.empty();
```

jonasfj (approved):

app/lib/service/download_counts/backend.dart
```diff
@@ -0,0 +1,52 @@
...
+  final DatastoreDB _db;
+
+  DownloadCountsBackend(this._db);
+
+  Future<DownloadCounts?> lookupDownloadCounts(String pkg) async {
+    final key = _db.emptyKey.append(DownloadCounts, id: pkg);
+    return _db.lookupOrNull<DownloadCounts>(key);
+  }
+
+  Future<DownloadCounts> ingestDownloadCounts(String pkg, CountData cd) async {
```
jonasfj, line 28: save is a better name, or overwrite, since this doesn't actually ingest. And save might imply some sort of transactionaility.

I'd suggest not having the method at all.

app/lib/service/download_counts/backend.dart
szakarias: Good idea.

app/lib/service/download_counts/models.dart
szakarias: done.

app/lib/service/download_counts/backend.dart
szakarias: done

app/lib/service/download_counts/backend.dart
szakarias: yes.

PR #7815, opened 2024-06-17
Additional paragraph on the appeal page.

isoos:


isoos:
@jonasfj: should we display the same message on the report page too, or just on the appeal page?

PR #7818, opened 2024-06-18
CSS refactor + start on dark theme color variables.

isoos:
#4416

### DIFF
--- a/pkg/web_css/lib/src/_base.scss
+++ b/pkg/web_css/lib/src/_base.scss
@@ -10,8 +10,8 @@
 }
 
 body {
-  background-color: var(--main-bg-color);
-  color: var(--main-text-color);
+  background-color: var(--pub-default-background-color);
+  color: var(--pub-default-text-color);
   font-size: 14px;
   line-height: 1.6;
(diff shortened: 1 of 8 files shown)


isoos:
Notes:
- implementing this one component/page at a time, this is for the `/help` page (pushed to staging)
- naming and structure to be discussed

jonasfj (approved):

sigurdm:
Beautiful!

We are going to need dark-mode variants of the images. The white background behind the packages is glaring.
image

sigurdm (approved):

isoos:
I did a bit more thinking about how the variable names are read when checking the styles, and added a note + refactored the current ones. My suggestion is to name them like this: `--pub-[component]-[role]-[property]`, where
- `[component]` may be: `default`, `button`, `link`, `card`, ...
- `[role]` may be the `background`, `text`, `border`, ...
- `[property]` may be `color`, `opacity` for specific values, or a `value` for multi-part properties.

I'm not yet sure if we should allow `-` separators inside any of the parts, but let's cross that bridge once we are there.

Issue #7817, opened 2024-06-17
Doc generation failed https://pub.dev/packages/googleapis/versions/13.2.0/score/log.txt


kevmoo:
URL: https://pub.dev/packages/googleapis/

See https://github.com/google/googleapis.dart/issues/630

jonasfj:
Have we added new docs?

Is the package larger. Could also be a fluke, this package is really challenging the limits :)

We might have to review them, or count how many packages are affected and make an allowlist for longer timeouts..

kevmoo:
It's got a bit bigger...new APIs. Nothing huge AFAICT

jonasfj:
Current timeout is 55min!

But it should be starting from the latest version, we should review if ordering changed.

https://github.com/dart-lang/pub-dev/blob/master/pkg/pub_worker/lib/src/analyze.dart#L27

isoos:
Comparing the log.txt files:

```
## Running analysis for "googleapis" version "13.1.0"
STARTED: 2024-06-17T20:04:45.385673Z
[...]
### Execution of pana exited 0
STOPPED: 2024-06-17T20:31:53.645140Z
```

```
## Running analysis for "googleapis" version "12.0.0"
STARTED: 2024-06-17T20:34:47.624734Z
[...]
### Execution of pana exited 0
STOPPED: 2024-06-17T21:02:27.623748Z
```

```
## Skipping analysis for "googleapis" version "13.2.0"
date-time: 2024-06-17T21:05:01.567911Z
```

It looks like the order was (and still is) off.

isoos:
The fix has been deployed, now the latest stable always gets the priority on task scheduling.

PR #7821, opened 2024-06-20
Keep order on task update when new version is published.

isoos:
- #7817
- #7830

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -432,17 +432,16 @@ class TaskBackend {
       }
 
       // Make changes!
-      state.versions!
-        // Remove versions that have been deselected
-        ..removeWhere((v, _) => deselectedVersions.contains(v))
-        // Add versions we should be tracking
-        ..addAll({
-          for (final v in untrackedVersions)
-            v: PackageVersionStateInfo(
+      // Note: iterating over the tracked versions list keeps the order.
+      state.versions = Map.fromEntries(versions.map((v) {
+        final oldVersion = state.versions![v];
+        final newVersion = oldVersion ??
+            PackageVersionStateInfo(
               scheduled: initialTimestamp,
               attempts: 0,
-            ),
-        });
+            );
+        return MapEntry(v, newVersion);
+      }));
       state.derivePendingAt();
 
       _log.info('Update state tracking for $packageName');
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -1166,6 +1166,7 @@ List<Version> _versionsToTrack(
         .sorted(Comparable.compare)
         .reversed
         .take(5)
+    // This maintains the order above, as Sets are iterated in insertion order.
   }.nonNulls.where(visibleVersions.contains).toList();
 }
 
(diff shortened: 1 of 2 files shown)


app/lib/task/backend.dart
```diff
@@ -432,17 +432,16 @@ class TaskBackend {
...
       // Make changes!
-      state.versions!
-        // Remove versions that have been deselected
-        ..removeWhere((v, _) => deselectedVersions.contains(v))
-        // Add versions we should be tracking
-        ..addAll({
-          for (final v in untrackedVersions)
-            v: PackageVersionStateInfo(
+      // Note: iterating over the tracked versions list keeps the order.
+      state.versions = Map.fromEntries(versions.map((v) {
```
jonasfj, line 436: Nothing specifies that this property should be sorted.
That's not an invariant, if it's an important invariant, it's an undocumented one.

Given that it's a map, I'd suggest we shouldn't attempt to sort it. Yes, Dart does tend to maintain the ordering of maps, but this is perhaps not something we want to rely on. It's very easy to break this, even if it is documented.

And it's not, see:
https://github.com/dart-lang/pub-dev/blob/bf3d7029ebcdb5ec9ab7bb045f19ed0d6124f792/app/lib/task/models.dart#L106-L108

jonasfj: I propose that we sort the versions here:
https://github.com/dart-lang/pub-dev/blob/bf3d7029ebcdb5ec9ab7bb045f19ed0d6124f792/app/lib/task/scheduler.dart#L289C31-L289C46

Or in here:
https://github.com/dart-lang/pub-dev/blob/bf3d7029ebcdb5ec9ab7bb045f19ed0d6124f792/app/lib/task/scheduler.dart#L289C31-L289C46
(in which case we should also document that they are sorted in order of priority).

Should probably use something like [Version.prioritize](https://pub.dev/documentation/pub_semver/latest/pub_semver/Version/prioritize.html).

Or maybe we want to make our own sorting, where we prioritize the latest stable release highest, and then everything else in order of version number.

 * `2.5.0` (this is special, otherwise this is the order of versions)
 * `3.0.0-dev.5`
 * `3.0.0-dev.4`
 * ...
 * `3.0.0-dev.1`
 * `2.4.0`
 * `2.4.0-dev.1`
 * `2.3.0`
 * ...

Or maybe, we prioritize everything in order of [Version.prioritize](https://pub.dev/documentation/pub_semver/latest/pub_semver/Version/prioritize.html) with the exception of the latest pre-release which is prioritized second highest.

 * `2.5.0`
 * `3.0.0-dev.5` (this is special, otherwise this is order of [Version.prioritize](https://pub.dev/documentation/pub_semver/latest/pub_semver/Version/prioritize.html))
 * `2.4.0`
 * ...
 * `2.3.0`
 * `3.0.0-dev.4`
 * ...
 * `3.0.0-dev.1`
 * `2.4.0-dev.1`
 * ...

--------------

Regardless, let's not try to maintain an ordering in datastore. It's much safer to sort all entries when we create the payload we send to the worker.

Yes, I realize we'll have to parse the versions, but this is extremely cheap (compared to actually downloading an analyzing the versions). Also these are sort lists.

jonasfj (changes requested):

isoos:
Implemented the second sort in #7832. Closing this PR now.

Issue #7813, opened 2024-06-14
Consider making the magnifying glass clickable


sigurdm:
<img width="1140" alt="image" src="https://github.com/dart-lang/pub-dev/assets/8613953/fd170ae4-9130-4048-bba2-d0e1c4e27ddb">

We had a request for this.

"please make the lens icon clickable instead of relying on enter key press to execute search. Some browsers don't respond to enter key press and this is frustrating when browsing your magnificent website. thank you"

isoos:
Reading upon this a bit, and my finding so far is: 

- The HTML specification encourages browsers to allow implicit submission (as in our case to submit the form on hitting enter after typing):
https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#implicit-submission

- My non-scientific sampling of browsers and forum entries suggest to me that most users have the problem the other way around: the form submits even when they are not intending to. There also may be extensions that interfere with form submission.

I see two alternatives here:

- We could add a submit button with the icon on it, styling it the same way as the magnifier glass today. The drawback here would be tab order: the page starts with the input field focused, and then it would jump upfront on tabbing.

- We could catch the enter key in the input field, and trigger a form submission via JavaScript. We should also prevent propagation, so that there is no accidental double-submission. (sending out a PR for it soon)

jonasfj:
@sigurdm do we know what browsers are affected?

sigurdm:
No - there is unfortunately no more context than what I quoted.

I first suggested the poster to file an issue, and when they didn't after a few days I posted it here.

isoos:
As we don't have any way to reproduce the issue, and e.g. Google search has similar icon that is not clickable, I'm closing this issue too.

PR #7823, opened 2024-06-20
Force search form submission on enter key.

isoos:
#7813

### DIFF
--- a/pkg/web_app/lib/src/search.dart
+++ b/pkg/web_app/lib/src/search.dart
@@ -10,10 +10,30 @@ import 'gtm_js.dart';
 import 'page_updater.dart';
 
 void setupSearch() {
+  _setFormSubmitKeyboardHandler();
   _setEventForKeyboardShortcut();
   _setEventsForSearchForm();
   _setEventForFiltersToggle();
   _setEventForSortControl();
+}
+
+void _setFormSubmitKeyboardHandler() {
+  final inputElem = document.querySelector('form.search-bar input[name="q"]');
+  if (inputElem != null && inputElem is InputElement) {
+    final form = inputElem.form;
+    if (form == null) {
+      return;
+    }
+    inputElem.onKeyDown.listen((e) {
+      // only handle `enter` key
+      if (e.keyCode != 13) {
+        return;
+      }
+      e.preventDefault();
+      e.stopPropagation();
+      form.submit();
+    });
+  }
 }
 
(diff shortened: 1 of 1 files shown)


sigurdm:
Could you deploy this on staging to play with?

isoos:
> Could you deploy this on staging to play with?

Deployed to staging, I think it should be online now.

sigurdm:
I must admit I don't understand why this is better than relying on the default behavior? (I see no differences in Chrome)

Do we know which browsers have problems with the default behavior?

Could you link to the relevant discussions you found.

isoos:
> (I see no differences in Chrome)

That is to be expected, also in Firefox, they both work as expected (by default, not sure how that changes with extensions).

> Do we know which browsers have problems with the default behavior?

This was reported through this issue: #7813. Do you recall the source?

sigurdm:
> This was reported through this issue: https://github.com/dart-lang/pub-dev/issues/7813. Do you recall the source?

Yeah the source is a support email. Though I quoted the entirety of the email.

Maybe we do nothing until we can confirm this problem is real?

sigurdm:
I thought the suggestion of making the icon clickable made some sense.

However checking out the google.com search bar, their icon does not seem clickable...

pkg/web_app/lib/src/search.dart
```diff
@@ -10,12 +10,32 @@ import 'gtm_js.dart';
...
+void _setFormSubmitKeyboardHandler() {
+  final inputElem = document.querySelector('form.search-bar input[name="q"]');
+  if (inputElem != null && inputElem is InputElement) {
+    final form = inputElem.form;
+    if (form == null) {
+      return;
+    }
+    inputElem.onKeyDown.listen((e) {
+      // only handle `enter` key
+      if (e.keyCode != 13) {
```
jonasfj, line 29: Don't submit on shift+enter

jonasfj (commented):
I don't mind this, but do we need it?

What browser are we having issues in?

isoos:
> I don't mind this, but do we need it?

I think it is important to note that by default the major browsers with their default settings have the expected behavior and we wouldn't need this, unless we want to handle special configurations.

sigurdm:
I'd say we don't land this - reimplementing default behavior seems odd - especially if we don't have clarity of which browsers have the problem, and if the problem is indeed solved by this.

PR #7824, opened 2024-06-21
lighthouse tool to track accessibility score changes

isoos:
- updated #5318 for new versions and goals
- tracking only the accessibility score in `README.md`, as that is relevant for the dark mode
- added `?force-experimental-dark=1` query parameter (only applied at local testing and fake server) to trigger dark mode theme for lighthouse
- I'm planning to run this script on each dark-mode change to check that we don't have any regression in light or dark mode

### DIFF
--- a/app/lib/frontend/handlers/experimental.dart
+++ b/app/lib/frontend/handlers/experimental.dart
@@ -12,6 +12,7 @@ const _publicFlags = <String>{
 };
 
 const _allFlags = <String>{
+  'dark-as-default',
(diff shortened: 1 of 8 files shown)


sigurdm (approved):
LGTM

Perhaps we could set up a regular github action to run this...

isoos:
> Perhaps we could set up a regular github action to run this

Let's see how useful it is during the dark mode experiments, and if it is worth it, we should do a test that runs regularly.

PR #7828, opened 2024-06-25
'updated:<duration>' search query filter for packages updated in the last <duration> days

### LINKED ISSUE #7701: [Feature request] ⭐ Filter old packages
There are a lot of abandoned packages (some with more than 2 years since last update, which will lead to dependency hell or the inability to upgrade the project in the long run because of outdated dependencies).

It would be nice if we could configure in the search results the maximum package age, in months (ordering by update date doesn't solve because the only useful sort ordering is search relevance).

Alternatively, a nice gauge showing the quality of the package would be awesome. Something like the pub points, where it reaches 0 when the last package update is >= 12 months.

Something like

★★★★★ = Package have the same current Flutter/Dart SDK (brand new)
★★★★✩ = Package is not updated in the last 3 months
★★★✩✩ = Package is not updated in the last 6 months
★★✩✩✩ = Package is not updated in the last 12 months
★✩✩✩✩ = Package is not Dart 3 compatible
✩✩✩✩✩ = Package is not Dart 2 compatible

isoos:
- Fixes #7701
- Removes unused `updatedInDays` direct parameter from `ServiceSearchQuery`, applies the same logic in the parsed query.
- Note: only `7m` or `5d` is supported, the combination of `7m5d` is not supported.
- Note: deployed to staging (though its data has not been updated in the past few months, but https://staging.pub.dev/packages?q=updated%3A13w shows it working.

### DIFF
--- a/pkg/_pub_shared/lib/search/search_form.dart
+++ b/pkg/_pub_shared/lib/search/search_form.dart
@@ -15,6 +15,7 @@ final RegExp _refDependencyRegExp =
 final RegExp _allDependencyRegExp =
     RegExp(r'dependency\*:([_a-z0-9]+)', caseSensitive: false);
 final _sortRegExp = RegExp('sort:([a-z]+)');
+final _updatedRegExp = RegExp('updated:([0-9]+[dwmy]?)');
 final _tagRegExp =
     RegExp(r'([\+|\-]?[a-z0-9]+:[a-z0-9\-_\.]+)', caseSensitive: false);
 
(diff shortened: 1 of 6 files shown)


pkg/_pub_shared/lib/search/search_form.dart
```diff
@@ -69,6 +70,33 @@ SearchOrder? parseSearchOrder(String? value) {
   return null;
 }
 
+int? parseUpdatedInDays(String value) {
```
jonasfj, line 73: Feel free to ignore this suggestion from the JS I did years ago, but we have the same regex engine... It's been adopted, just inspired by the regex we used there:
```dart
final timeExp = new RegExp(
  [
    '^((-|\\+))?',
    '((?<years>\\d+)(y|year|years|yr))?',
    '((?<months>\\d+)(months|month|mo))?',
    '((?<weeks>\\d+)(weeks|week|wk|w))?',
    '((?<days>\\d+)(days|day|d))?',
    '((?<hours>\\d+)(hours|hour|hr|h))?',
    '((?<minutes>\\d+)(minutes|minute|min|m))?',
    '((?<seconds>\\d+)(seconds|second|sec|s))?',
    '\$',
  ].join(''),
  caseSensitive: true,
);

Duration parseTime(String duration) {
  // Parse the string
  final match = timeExp.firstMatch(duration);
  if (match == null) {
    throw FormatException('String: "$duration" isn\'t a time expression');
  }
  // Negate if needed
  final neg = match.group(2) == '-' ? -1 : 1;
  // Return parsed values
  return Duration(
    days: (365 * (int.tryParse(match.namedGroup('years') ?? '0') ?? 0) * neg +
        31 * (int.tryParse(match.namedGroup('months') ?? '0') ?? 0) * neg +
        7 * (int.tryParse(match.namedGroup('weeks') ?? '0') ?? 0) * neg +
        (int.tryParse(match.namedGroup('days') ?? '0') ?? 0) * neg),
    hours: int.tryParse(match.namedGroup('hours') ?? '0') ?? 0 * neg,
    minutes: int.tryParse(match.namedGroup('minutes') ?? '0') ?? 0 * neg,
    seconds: int.tryParse(match.namedGroup('seconds') ?? '0') ?? 0 * neg,
  );
}
```

jonasfj (approved):

pkg/_pub_shared/lib/search/search_form.dart
isoos: updated the parsing with expressions from here, using `m` for months and ignoring minute/seconds level details.

isoos: PTAL

Issue #7830, opened 2024-06-25
Rewrite PackageState.versions to not depend on the order of the keys for analysis


isoos:
Options:
- do sort/prioritization before submitting the payload to a worker
- keep track of the list of versions in their priority order as a List

isoos:
We've decided that sorting will be enough for now: #7832.

PR #7831, opened 2024-06-25
Populate new field: PackageVersionStateInfo.version

isoos:
#7830


jonasfj:
I don't mind this, but are we sure we want to finish this migration?

isoos:
Closing as #7832 will be enough for now.

PR #7832, opened 2024-06-26
Sort task version list when creating worker payload.

isoos:
- https://github.com/dart-lang/pub-dev/issues/7817
- https://github.com/dart-lang/pub-dev/issues/7830

alternative to https://github.com/dart-lang/pub-dev/pull/7821 and https://github.com/dart-lang/pub-dev/pull/7831

### DIFF
--- a/app/lib/task/scheduler.dart
+++ b/app/lib/task/scheduler.dart
@@ -7,6 +7,7 @@ import 'package:clock/clock.dart';
 import 'package:logging/logging.dart' show Logger;
 import 'package:meta/meta.dart';
 import 'package:pub_dev/package/backend.dart';
+import 'package:pub_dev/package/models.dart';
 import 'package:pub_dev/shared/configuration.dart';
 import 'package:pub_dev/shared/datastore.dart';
 import 'package:pub_dev/shared/utils.dart';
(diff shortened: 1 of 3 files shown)


app/lib/task/scheduler.dart
```diff
@@ -307,14 +308,38 @@ Future<Payload?> updatePackageStateWithPendingVersions(
...
+              version: v,
+              token: s.versions![v]!.secretToken!,
+            ))
+        .toList();
+
+    // sort pending versions based on version priority
+    if (payloadVersions.length > 1) {
+      // loading the package entity to prioritize latest versions
+      final pkg = await tx.lookupOrNull<Package>(
+          db.emptyKey.append(Package, id: state.package));
```
jonasfj, line 322: How about no using this? just using the version number.

I/O is not super cheap, and we need to schedule many instances quickly.

app/lib/task/scheduler.dart
```diff
@@ -307,14 +307,50 @@ Future<Payload?> updatePackageStateWithPendingVersions(
     s.derivePendingAt();
     tx.insert(s);
 
+    final payloadVersions = pendingVersions
```
jonasfj, line 310: Might be a tiny bit easier if we sort `pendingVersions` before doing the `.map`.

Also then we won't need: `VersionTokenPair.semanticVersion`.

pkg/_pub_shared/lib/data/task_payload.dart
```diff
@@ -70,4 +71,6 @@ class VersionTokenPair {
   factory VersionTokenPair.fromJson(Map<String, dynamic> json) =>
       _$VersionTokenPairFromJson(json);
   Map<String, dynamic> toJson() => _$VersionTokenPairToJson(this);
+
+  late final semanticVersion = Version.parse(version);
```
jonasfj, line 75: nit.
```suggestion
  /// [version] parsed as [Version] object.
  late final semanticVersion = Version.parse(version);
```

Honestly, I'm tempted to do:
```suggestion
  /// [version] as [Version] object.
  Version get semanticVersion => Version.parse(version);
```

jonasfj (approved):

jonasfj (approved):

PR #7839, opened 2024-06-28
Dark mode overrides for material components: checkbox, drowdown, textfield, textarea and some table styles.

isoos:
#4416

### DIFF
--- a/app/lib/service/entrypoint/frontend.dart
+++ b/app/lib/service/entrypoint/frontend.dart
@@ -63,7 +63,7 @@ Future<void> watchForResourceChanges() async {
   void setupWatcher(
       String name, String path, FutureOr<void> Function() updateFn) {
     final w = Watcher(path, pollingDelay: Duration(seconds: 3));
-    final subs = w.events.debounce(Duration(milliseconds: 200)).listen(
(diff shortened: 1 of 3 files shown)


app/lib/service/entrypoint/frontend.dart
```diff
@@ -63,7 +63,7 @@ Future<void> watchForResourceChanges() async {
   void setupWatcher(
       String name, String path, FutureOr<void> Function() updateFn) {
     final w = Watcher(path, pollingDelay: Duration(seconds: 3));
-    final subs = w.events.debounce(Duration(milliseconds: 200)).listen(
+    final subs = w.events.debounce(Duration(milliseconds: 2500)).listen(
```
sigurdm, line 66: Is this related?

sigurdm (approved):

app/lib/service/entrypoint/frontend.dart
isoos: Yes, although only in development mode with `--watch` flag: without it on my machine the rebuild got triggered way too fast, often failing with syntax error due to unfinished line being saved by the IDE. A slightly longer time provides much better workflow.

PR #7840, opened 2024-07-01
Download counts: fix date formatting in file name

szakarias:


app/lib/service/download_counts/sync_download_counts.dart
```diff
@@ -123,6 +123,11 @@ Future<bool> processDownloadCounts(
 }
 
 const numberOfSyncDays = 5;
+String formatDateForFileName(DateTime date) {
+  final month = date.month > 9 ? date.month : '0${date.month}';
+  final day = date.day > 9 ? date.day : '0${date.day}';
```
sigurdm:
```suggestion
  final month = date.month.toString().padLeft(2, '0');
  final day = date.month.toString().padLeft(2, '0');
```

app/lib/service/download_counts/sync_download_counts.dart
```diff
@@ -143,7 +148,7 @@ Future<void> syncDownloadCounts() async {
     final syncDate = today.addCalendarDays(-i);
     final fileName = [
       'daily_download_counts',
-      '${syncDate.year}-${syncDate.month}-${syncDate.day}T00:00:00Z',
+      formatDateForFileName(syncDate),
```
sigurdm, line 146: Another option would be to do
```suggestion
      DateTime(syncDate.year, syncDate.month, syncDate.day).toIso8601String(),
```

sigurdm (approved):
I suggest some alternatives but LGTM

app/lib/service/download_counts/sync_download_counts.dart
```diff
@@ -171,3 +171,9 @@ Future<void> syncDownloadCounts() async {
     }
   }
 }
+
+String formatDateForFileName(DateTime date) {
+  final month = date.month > 9 ? date.month : '0${date.month}';
+  final day = date.day > 9 ? date.day : '0${date.day}';
+  return '${date.year}-$month-${day}T00:00:00Z';
```
jonasfj, line 178: Have you considered `date.toIso8601String().split('T').first` 

See: [DateTime.toIso8601String()](https://api.dart.dev/stable/3.4.4/dart-core/DateTime/toIso8601String.html).

Or maybe just use [String.padLeft](https://api.dart.dev/stable/3.4.4/dart-core/String/padLeft.html).

```dart
final day = '${date.day}'.padLeft(2, '0');
```

app/lib/service/download_counts/sync_download_counts.dart
szakarias: Are you sure this works? I think it ends with `000` and not `00Z`

app/lib/service/download_counts/sync_download_counts.dart
jonasfj: Or even just `DateTime.utc(2023, 2, 2).toIso8601String().replaceAll(RegExp(r'\.0+Z$'), 'Z')`

In Dart `toIso8601String` includes ms (and if us is not zero, it also includes us).
So typically, you just want to strip ms and us away.

jonasfj (approved):

app/lib/service/download_counts/sync_download_counts.dart
szakarias: done

app/lib/service/download_counts/sync_download_counts.dart
szakarias: I think `padLeft` is the simplest. Going with that.

PR #7844, opened 2024-07-02
Download counts: handle incorrect versions in data

szakarias:


app/lib/service/download_counts/sync_download_counts.dart
```diff
@@ -87,39 +92,89 @@ Future<bool> processDownloadCounts(
...
-        final package = data['package'];
-        if (package is! String) {
+
+        if (data['package'] is! String) {
           throw FormatException('"package" must be a String');
         }
-
+        package = data['package'] as String;
+        List<String> versions;
+        try {
```
sigurdm, line 105:
```suggestion
        try {
          // Validate that the package and version exists.
          // First do it via the cached data, fall back to query for invisible and moderated packages.
```

app/lib/service/download_counts/sync_download_counts.dart
```diff
@@ -72,7 +73,11 @@ Future<bool> processDownloadCounts(
     return false;
   }
 
-  bool failedLines = false;
+  bool hasFailedLines = false;
+  bool hasPartiallyFailedLines = false;
+  // Before '2024-07-02' the query generating the download count data had a bug
+  // in the reg exp causing incorrect versions to be stored.
+  final regExpQueryFixDate = DateTime.parse('2024-07-02');
```
sigurdm, line 80: I've backfilled with good counts until 2024-05-03

app/lib/service/download_counts/sync_download_counts.dart
szakarias: done

jonasfj (approved):

PR #7846, opened 2024-07-03
Do not use pana's License.path, use 'license' label instead.

isoos:
Context: for a long while now, we only accept the `LICENSE` file as the source of the package license(s). `pana` will remove the `path` field in an upcoming release, we should not use it anymore: https://github.com/dart-lang/pana/pull/1385


isoos:
> Are there not old packages that follow the old convention, and need this?

Short answer: we don't need it.  Couple of details:
- This is used only for the label.
- We no longer extract license content from anything else except `LICENSE`. 
- We store the extracted license text in the datastore, and pana analysis will not affect it much. (Which may be a good or a bad thing, depending on how I look at it).

sigurdm (approved):

Issue #7541, opened 2024-03-07
Prevent submitting packages with `hook/xx.dart` that are not allowlisted


dcharkes:
@jonasfj suggested that we should prevent people from uploading packages to pub.dev with hooks that we haven't specified.

This should be relatively easy to add to the validator.

For now, we'd only allowlist `hook/build.dart` and `hook/link.dart`.

Context:

* https://github.com/dart-lang/sdk/issues/54334

cc @mosuem @sigurdm

jonasfj:
I think we should at-least consider it. Probably we need to scan all packages to see if anyone is already using this directory.

sigurdm:
I guess it would be an overridable warning. Another pub client-side validation?

Or do we want a server-side check?

mosuem:
What about helper files, if my `build.dart` is getting too complex? Should they not be stored in `hooks/src/` or similar?

dcharkes:
> What about helper files, if my `build.dart` is getting too complex? Should they not be stored in `hooks/src/` or similar?

I'd probably put these in `src/`. I think for `bin/` scripts we also store the complex logic in `src/`. The `bin/` and `hook/` directories should only contain the entry-points. cc @jonasfj @sigurdm

sigurdm:
Is this language versioned? (will future sdk's potentially allow more files here)?

sigurdm:
I think we are converging on a server-side check. That is easier to update/relax over time.

jonasfj:
Let's make a server side map:
```dart

final _hookMinimumRequiresSdkForFile = {
  // because if you have "hook/build.dart" in the package
  // then you must have 'environment.sdk' with a minimum bound of 3.45.0
  // The actual version number is illustrative.
  'build.dart': '3.45.0',
};
```

dcharkes:
> I think we are converging on a server-side check. That is easier to update/relax over time.

Ah right, because we can just deploy a new version. With a client-side check we have old versions of Dart/pub client floating around forever. (Though I guess users could never submit a package with a newer version of Dart than the SDK they have locally.)

You're the experts, sgtm!

isoos:
Closing as this has been implemented.

PR #7847, opened 2024-07-03
Minimal check for hook/ files in new uploads.

isoos:
#7541


sigurdm (approved):

sigurdm:
cc @dcharkes

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -832,3 +833,20 @@ Iterable<ArchiveIssue> checkTopics(String pubspecContent) sync* {
     }
   }
 }
+
+Iterable<ArchiveIssue> checkHooks(Set<String> fileNames) sync* {
```
dcharkes, line 837: Should this use the SDK version as @jonasfj suggested in the issue?

jonasfj: Yeah, if we could pass along the SDK constraint and require that this greater than `3.6.0-0` ensuring that it's at-least Dart 3.6, then I think that'd help a bit.

jonasfj (approved):
Don't mind landing this as is though. But it'd be nice to add the SDK constraint check later.

pkg/pub_package_reader/lib/pub_package_reader.dart
jonasfj: @dcharkes we don't know what SDK it'll land in yet, right? Just that it definitely won't be 3.5

isoos: Updated, PTAL.

pkg/pub_package_reader/lib/pub_package_reader.dart
```diff
@@ -832,3 +836,44 @@ Iterable<ArchiveIssue> checkTopics(String pubspecContent) sync* {
...
+}
+
+Iterable<ArchiveIssue> checkHooks(
+  Version? minimumSdkConstraint,
+  Set<String> fileNames,
+) sync* {
+  final hookFiles = fileNames.where((n) {
+    final parts = p.split(n);
+    return parts.length > 1 && parts.first == 'hook';
+  }).toList();
```
jonasfj, line 866: Have we considered:
```suggestion
  final hookFiles = fileNames.where((f) => f.startsWith('hook/'));
```

We don't even need to use `toList()` :rofl:

jonasfj (approved):

PR #7849, opened 2024-07-04
Download counts: allow custom date and number of sync days when syncing

szakarias:


app/lib/service/download_counts/sync_download_counts.dart
```diff
@@ -180,10 +180,17 @@ Future<bool> processDownloadCounts(
   }
 }
 
-const numberOfSyncDays = 5;
+const defaultNumberOfSyncDays = 5;
```
isoos, line 183: nit: `@visibleForTesting`

isoos (approved):

app/lib/service/download_counts/sync_download_counts.dart
szakarias: I would like to use this value from an admin action, so I will keep it public.

PR #7851, opened 2024-07-04
Add admin action to backfill download counts

szakarias:


app/lib/admin/actions/actions.dart
```diff
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:pub_dev/admin/actions/backfill_download_counts.dart';
```
isoos, line 5: Nit: let's make this a relative path.

isoos (approved):

app/lib/admin/actions/actions.dart
szakarias: done

PR #7854, opened 2024-07-05
Add admin action for deleting DownloadCount entities

szakarias:


app/lib/admin/actions/actions.dart
```diff
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:pub_dev/admin/actions/backfill_download_counts.dart';
+import 'package:pub_dev/admin/actions/delete_download_counts.dart';
```
isoos, line 6: Nit: let's use relative paths here (and the line above).

isoos (approved):

PR #7863, opened 2024-07-11
Fix retraction-related latest stable/prerelease/preview version calculation.

### LINKED ISSUE #7852: Retracted package should show the latest available one instead of the latest retracted one if they are not compatible between
Package: https://pub.dev/packages/ord_dart

I've retracted all v0.1 versions and published a v0.0.8 version. But the page still shows that the v0.1.0+2 has been retracted.

image

isoos:
- Fixes #7852.
- I was able to create a reproduction of the bug, and it turns out it was not in the version update at retraction, it was at upload time, when all of the versions are retracted.

### DIFF
--- a/app/lib/package/models.dart
+++ b/app/lib/package/models.dart
@@ -340,6 +340,7 @@ class Package extends db.ExpandoModel<String> {
     PackageVersion pv, {
     required Version dartSdkVersion,
     required Version flutterSdkVersion,
+    bool existingLatestIsRetracted = false,
   }) {
     final newVersion = pv.semanticVersion;
     final isOnStableSdk = !pv.pubspec!.isPreviewForCurrentSdk(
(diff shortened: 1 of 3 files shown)


app/lib/package/models.dart
```diff
@@ -340,27 +340,31 @@ class Package extends db.ExpandoModel<String> {
     PackageVersion pv, {
     required Version dartSdkVersion,
     required Version flutterSdkVersion,
+    bool existingLatestIsRetracted = false,
   }) {
```
jonasfj, line 344: I think we should seriously consider being more functional here.

Instead of looking at:
 * `Package` entity,
 * `PackageVersion` entity that was just published, and,
 * SDK versions.
And then try to decide of `pv` should replace any of the current designations as latest, preview, prerelease in the `Package` entity.

----

We should instead:
 * Fetch all `PackageVersion` entities for the given package.
 * Determine latest, preview and prerelease using:
    * the SDK versions
    * All `PackageVersion` entities for the given package.
 * Update the `PackageEntity` to the newly determined values for latest, preview and prerelease.

In particular, when determining what `latestVersionKey` should be, **we should never read** `PackageEntity.latestVersionKey`.

The value of `PackageEntity.latestVersionKey` should be derived from:
 * the SDK versions,
 * All `PackageVersion` entities for the given package.

That removes a LOT of corner cases, and it becomes much simpler to keep the code correct.

Yes, it's more expensive to load all `PackageVersion` entities for a given package. But we rarely change a property that causes us to recompute `PackageEntity.latestVersionKey` (and friends). This mostly happens when:
 * a package is published,
 * an admin action flips retracted state on a version, or,
 * a new SDK version is published.

All of these are rare events.

jonasfj (approved):

app/lib/package/models.dart
isoos: Added a tracking issue, merging the fix for now.

PR #7867, opened 2024-07-12
Dark mode hack for SVG icons

isoos:
- #4416
- pushed to staging

image

image


isoos:
I've added notes to the CSS file of this style. I think eventually we should migrate off of it, but right now this had the best benfit/effort ratio.

jonasfj (approved):

jonasfj:
I think it'd be okay to file an issue for us to follow up on fixing these icons.

isoos:
Filed #7868 to track follow-ups.

Issue #7897, opened 2024-08-01
Error: Null check operator used on a null value in checkhooks


sigurdm:
```
Error: Null check operator used on a null value

at .checkHooks ( package:pub_package_reader/pub_package_reader.dart:869 )
at ._SyncStarIterator.moveNext
at .List.addAll
at .summarizePackageArchive ( package:pub_package_reader/pub_package_reader.dart:275 )
at PackageBackend.publishUploadedBlob.<fn> ( package:pub_dev/package/backend.dart:914 )
at withTempDirectory.<fn>.<fn> ( package:pub_dev/shared/utils.dart:43 )
at .PackageBackend.publishUploadedBlob ( package:pub_dev/package/backend.dart:897 )
at .PubApi.finishPackageUpload ( package:pub_dev/frontend/handlers/pubapi.dart:131 )
at .PubApi.packageUploadCallback ( package:pub_dev/frontend/handlers/pubapi.dart:125 )
at ._$PubApiRouter.<fn> ( package:pub_dev/frontend/handlers/pubapi.g.dart:140 )
at RouterEntry.invoke.<fn> ( package:shelf_router/src/router_entry.dart:107 )
at .RouterEntry.invoke ( package:shelf_router/src/router_entry.dart:104 )
at .Router.call ( package:shelf_router/src/router.dart:184 )
at createAppHandler.<fn> ( package:pub_dev/frontend/handlers.dart:71 )
at _redirectLoopDetectorWrapper.<fn> ( package:pub_dev/shared/handler_helpers.dart:93 )
at _cspHeaderWrapper.<fn> ( package:pub_dev/shared/handler_helpers.dart:147 )
at _userAuthWrapper.<fn>.<fn> ( package:pub_dev/shared/handler_helpers.dart:243 )
at AccountBackend.withBearerToken.<fn> ( package:pub_dev/account/backend.dart:274 )
at ._ServiceScope._runScopeExitHandlers ( package:gcloud/service_scope.dart:241 )
at .AccountBackend.withBearerToken ( package:pub_dev/account/backend.dart:272 )
at _userAuthWrapper.<fn> ( package:pub_dev/shared/handler_helpers.dart:242 )
at _requestContextWrapper.<fn> ( package:pub_dev/shared/handler_helpers.dart:135 )
at _httpsWrapper.<fn> ( package:pub_dev/shared/handler_helpers.dart:258 )
[...]

PR #7898, opened 2024-08-01
Avoid looking up unknown hooks

sigurdm:
https://github.com/dart-lang/pub-dev/issues/7897


isoos:
I think it would be better if we added a `continue;` in the `if (){}` block, instead of putting it in the else block.

However, what worries me here is that the following test should have caught this, and I don't see why it did not:
https://github.com/dart-lang/pub-dev/blob/master/pkg/pub_package_reader/test/package_archive_test.dart#L1009-L1014

Any idea there?

isoos:
Figured out the test issue: the `.first.` iteration will stop the processing of the `sync*` looping, and the test won't touch the rest.
Submitting https://github.com/dart-lang/pub-dev/pull/7899 to fix both the test (by using `.single` instead of `.first`) and also the loop.

sigurdm:
Closed in favour of https://github.com/dart-lang/pub-dev/pull/7899

PR #7907, opened 2024-08-06
Dark-mode images for Flutter Favorite badge

isoos:
- #4416
- The banner's gray background becomes a bit darker, otherwise unchanged.
- The images are duplicated (now we have 4 in the HTML), and CSS rules decide which one of these is displayed.
- Added generic `displayed-in-dark-theme` and `displayed-in-light-theme` rules for the image hiding, as `<picture> + <source>` tag are not able to chose the image based on top-level CSS class presence.


app/lib/frontend/templates/views/shared/detail/header.dart
```diff
@@ -44,7 +48,25 @@ d.Node detailHeaderNode({
                   ),
                 ),
                 d.img(
-                  classes: ['ff-banner', 'ff-banner-mobile'],
+                  classes: [
```
sigurdm, line 51: Should we consider using a `<picture>` instead of classes to avoid loading both images?
Like suggested in https://stackoverflow.com/a/74639288

sigurdm (approved):

app/lib/frontend/templates/views/shared/detail/header.dart
isoos: `<picture>` cannot select (`<source>`) the image based on the presence of the top-level CSS class. We could use it for desktop/mobile view switch, but then we would need to emit the hardcoded width rules into the HTML `<source media="">` selector, which would complicate it a bit (as we have that by using SCSS variables, and we don't have a direct access to it while rendering the HTML).

isoos: In practice `<picture>` is not better for such use case, it only has a different tradeoff (which we could choose, but we may as well stick with the current practice).

isoos:
Note: this has been pushed to staging and can be observed there now.

sigurdm:
LGTM

PR #7911, opened 2024-08-07
Dark theme: early initialization and OS/browser-level setting detection.

isoos:
#4416

The `window.onLoad` event happens too late in the page rendering process, some styles and the background's white color may have already rendered at that point. Since every other part of our script code is rendered synchronously in the `main` method, we should be fine doing it here.
My local testing did not produced the quick flush of the light-themed content anymore, but we may eventually need to move it to a blocking script (instead of the current deferred script execution). Let's keep it here for now for simplicity.

Also added the OS/browser-level detection through `matchMedia` check. It may not be supported with all the browsers, we need to use defensive try-catch to prevent unimplemented errors blocking further execution.

### DIFF
--- a/pkg/web_app/lib/script.dart
+++ b/pkg/web_app/lib/script.dart
@@ -26,19 +26,20 @@ void main() {
   window.onPageShow.listen((_) {
     adjustQueryTextAfterPageShow();
   });
+  _setupDarkTheme();
+}
 
+void _setupDarkTheme() {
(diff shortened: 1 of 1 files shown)


sigurdm (approved):

isoos:
> After trying this, it seems to have improved the white flashes a lot, but on my device there are still very short flashes of white between page-loads.

Which OS/browser is that?

sigurdm:
macos/chrome

PR #7919, opened 2024-08-08
Validate package name in task_bump_priority action

sigurdm:


app/lib/admin/actions/task_bump_priority.dart
```diff
@@ -23,9 +24,14 @@ This is intended for debugging, or solving one-off issues.
     'package': 'Name of package whose priority should be bumped',
   },
   invoke: (options) async {
-    final package = options['package']!;
+    final package = options['package'] ??
+        (throw InvalidInputException('Needs a package name'));
     InvalidInputException.checkPackageName(package);
-
+    // Make sure package exists.
+    final pkg = await packageBackend.lookupPackage(package);
```
isoos, line 31: The lookup could be part of the adminBumpPriority method too, but this is fine as-is.

isoos (approved):

app/lib/admin/actions/task_bump_priority.dart
sigurdm: Ack

PR #7928, opened 2024-08-12
Extra mem for search

sigurdm:
https://github.com/dart-lang/pub-dev/issues/7927

### DIFF
--- a/search.yaml
+++ b/search.yaml
@@ -8,7 +8,7 @@ service: search
 
 resources:
   cpu: 2
-  memory_gb: 8
+  memory_gb: 12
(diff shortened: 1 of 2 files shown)


isoos (changes requested):
Let's also add an entry into changelog.

search.yaml
```diff
@@ -8,7 +8,7 @@ service: search
 
 resources:
   cpu: 2
-  memory_gb: 8
+  memory_gb: 12
```
isoos, line 11: we could go with 16

isoos (approved):

Issue #7868, opened 2024-07-15
Cleanup tasks after dark mode gets stable.


isoos:
[ ] duplicate icons for dark mode (not inverted ones)

PR #7929, opened 2024-08-12
Synchronous JS initionalization of the dark theme.

isoos:
- #7868 and #4416
- The initialization is rewritten in Javascript and not compiled for Dart only because it is very limited in scope and should be straightforward to maintain.
- The initialization needs to be blocking JS executed right after the body tag becomes present. Other JS codes that we use use async or deferred loading.

### DIFF
--- a/app/lib/frontend/templates/views/shared/layout.dart
+++ b/app/lib/frontend/templates/views/shared/layout.dart
@@ -167,6 +167,10 @@ d.Node pageLayoutNode({
             if (activeConfiguration.isStaging) 'staging-banner',
           ],
           children: [
+            // The initialization of dark theme must be in a synchronous, blocking
+            // script execution, as otherwise users may see flash of unstyled content
+            // (usually white background instead of a dark theme).
+            d.script(src: staticUrls.getAssetUrl('/static/js/dark-init.js')),
             if (activeConfiguration.isStaging)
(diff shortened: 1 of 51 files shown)


isoos:
Note: I'm pushing this to staging for further evaluation.

app/lib/frontend/templates/views/shared/layout.dart
```diff
@@ -167,6 +167,10 @@ d.Node pageLayoutNode({
             if (activeConfiguration.isStaging) 'staging-banner',
           ],
           children: [
+            // The initialization of dark theme must be in a synchronous, blocking
+            // script execution, as otherwise users may see flash of unstyled content
+            // (usually white background instead of a dark theme).
+            d.script(src: staticUrls.getAssetUrl('/static/js/dark-init.js')),
```
sigurdm, line 173: Just to confirm: `/static/` is served with caching headers, meaning that this file will not be fetched over the network on most page loads?

isoos: Yes. `getAssetUrl` rewrites the URL to use hashes and as long as we are on the same hash it will be served from cache.

sigurdm (approved):

jonasfj (approved):

PR #7946, opened 2024-08-15
Shorter ModeratedCase.caseId

isoos:
Keeps the date and the 10 character hex random component, connected with a `-` character.


jonasfj (approved):
Solid

PR #7950, opened 2024-08-16
In-Reply-To id for outgoing emails

isoos:


app/lib/shared/email.dart
```diff
@@ -114,19 +120,31 @@ class EmailMessage {
...
+  /// Throws [ArgumentError] if the [localMessageId] or the
+  /// [inReplyToLocalMessageId] field doesn't look like UUID or ULID.
   ///
   /// TODO: double-check that we follow https://www.jwz.org/doc/mid.html
-  void verifyLocalMessageId() {
-    final uuid = localMessageId;
-    if (uuid == null || uuid.length < 25 || uuid.length > 36) {
-      throw ArgumentError('Invalid uuid: `$uuid`');
+  void verifyLocalMessageIds() {
+    void verifyUuid(String? uuid) {
```
jonasfj, line 133: you need to call it something else, if we're no longer using a UUID

jonasfj (approved):

PR #7952, opened 2024-08-16
Use ModerationCase.caseId as the report's email messageId and also in in-reply-to

isoos:


app/lib/admin/actions/send_email.dart
```diff
@@ -31,11 +31,13 @@ The `to` argument is comma separated list of:
...
   options: {
-    'to':
-        'A comma separated list of email addresses or subjects (where the admins will get the emails).',
+    'to': 'A comma separated list of email addresses or subjects '
+        '(the recipients of the messages).',
     'from': 'The email address to impersonate (`support@pub.dev` by default).',
     'subject': 'The subject of the email message.',
     'body': 'The text content of the email body.',
+    'in-reply-to': 'The local message id of the email that this is a reply to '
+        '(e.g. moderation case id).',
```
jonasfj, line 40:
```suggestion
    'case': '(optional) Moderation case id that this should be sent as In-Reply-To',
```

I guess it's the same thing :rofl: 

-----------

Maybe, the other is better... I don't know... I'm thinking at at-least writing `(optional)` is nice :D

jonasfj (approved):

app/lib/admin/actions/send_email.dart
jonasfj, line 40:
```suggestion
    'in-reply-to': '(optional) The local message id of the email that this is a reply to '
        '(e.g. moderation case id).',
```

PR #7956, opened 2024-08-19
Rename admin actions and file names.

isoos:


jonasfj (approved):

app/lib/admin/actions/moderation_cases_list.dart
```diff
@@ -7,8 +7,8 @@ import 'package:pub_dev/shared/datastore.dart';
 
 import 'actions.dart';
 
-final listModerationCases = AdminAction(
-  name: 'list-moderation-cases',
+final moderationCasesList = AdminAction(
+  name: 'moderation-cases-list',
```
jonasfj, line 11:
```suggestion
  name: 'moderation-case-list',
```

app/lib/admin/actions/moderation_cases_list.dart
jonasfj, line 1: let's rename it `moderation_case_list.dart`

PR #7979, opened 2024-08-26
Experimental search completion

jonasfj:
This could probably use a few tests and some more polishing of CSS (and maybe position offsets).

image

Might be worth doing some tests in follow up PRs.


pkg/web_css/lib/src/_search.scss
```diff
@@ -104,3 +104,29 @@
...
+}
+
+.search-completer-option {
+  cursor: pointer;
+  white-space: nowrap;
+  padding: 0px 3px;
+}
+
+.search-completer-option:hover {
+  background-color: #ccc;
```
isoos, line 127: Note to self: we should have tests that prevent submitting CSS values like this, and forces us to use the color palette that is for both light and dark mode.

This is fine for now, I will update these is a follow-up PR (as I want to also update the naming of the CSS variables).

app/lib/frontend/templates/views/shared/search_banner.dart
```diff
@@ -64,3 +80,120 @@ d.Node searchBannerNode({
...
+/// will automatically start. It'll always try to use the longest match.
+/// If not specified options available are assumed to be `terminal`, that is
+/// they will be followed by whitespace.
+/// If `forcedOnly` is specified, completion can only be initiated with
+/// Ctrl+Space.
+///
+/// This can generate completion data of different sizes given [topics] and
+/// [licenses]. If [minimal] is specified, then only the most important
+/// completion keys are included.
+String completionDataJson({
```
isoos, line 106: I'm wondering if this should be rather part of the `pkg/web_app` script we load, as it will be cached and not re-created on each page load? It could be loaded as an asynchronous library import.

isoos: Maybe it could be two-fold, part of it may be in the `pkg/web_app`, part of it loaded dynamically from the page (like `topics`). Maybe it is not worth to optimize for this yet.

app/lib/frontend/templates/views/shared/search_banner.dart
```diff
@@ -16,17 +20,29 @@ d.Node searchBannerNode({
...
         autofocus: autofocus,
         value: queryText,
         attributes: {
           'title': 'Search',
+          if (requestContext.experimentalFlags.isSearchCompletionEnabled)
+            'data-widget': 'completion',
+          'data-completion-base64': _initialCompletionData,
+          'data-completion-src': '/api/search-input-completion-data',
+          'data-completion-class': 'search-completer',
+          'data-completion-option-class': 'search-completer-option',
```
isoos, line 43: Not sure: what is the role of these class names - do we need them as dynamic options?

jonasfj: I think it's class for the dropdown, and the options. But I could be sold with us making the option and selected options hardcoded classes.

There is an attempt here to keep the completer generic enough that it could be applied to other things as well. Maybe that's overkill :see_no_evil:

app/lib/frontend/templates/views/shared/search_banner.dart
jonasfj: It probably could be part of it.

app/lib/frontend/templates/views/shared/search_banner.dart
isoos: I think we can start with hardcoded now and add customization later - when needed?

jonasfj: Okay, I'll simplify a bit..

sigurdm (commented):
Nice.

I'd love the styling to be more "material" eg. like: https://material.angular.io/components/autocomplete/overview#autocomplete-filter

app/lib/frontend/templates/views/shared/search_banner.dart
```diff
@@ -16,17 +20,29 @@ d.Node searchBannerNode({
...
+    },
     action: formUrl,
     children: [
       d.input(
         classes: ['input'],
+        type: 'search',
         name: 'q',
         placeholder: placeholder,
-        autocomplete: 'on',
+        autocomplete: 'off',
```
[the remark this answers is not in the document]
jonasfj: no, we don't want the browser to auto-complete, it just suggests what you've typed before.

app/lib/frontend/handlers/custom_api.dart
```diff
@@ -281,6 +283,170 @@ Future<shelf.Response> apiTopicNameCompletionDataHandler(
...
+        .readAsBytes(topicsJsonFileName);
+    final topicsMap = json.decode(utf8.decode(topicsJson));
+    final topics = (topicsMap as Map<String, Object?>).keys.toList();
+    return gzip.encode(utf8.encode(completionDataJson(
+      topics: topics,
+      licenses: _commonLicenses,
+    )));
+  });
+
+  if (!request.acceptsGzipEncoding()) {
```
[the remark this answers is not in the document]
jonasfj: I'm 50/50 on that, maybe it's the better choice.

pkg/web_app/lib/src/completion.dart
```diff
@@ -0,0 +1,697 @@
...
+import 'dart:math' as math;
+
+import 'package:collection/collection.dart';
+import 'package:http/http.dart' deferred as http show read;
+import 'package:web/web.dart';
+
+extension on HTMLCollection {
+  /// Take a snapshot of [HTMLCollection] as a Dart [List].
+  ///
+  /// Notice that it's not really safe to use [Iterable], because the underlying
```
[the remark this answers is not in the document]
jonasfj: yes, if you remove things while iterating :see_no_evil: 

Things like NodeList are defined as live-collections :see_no_evil:

pkg/web_app/lib/src/completion.dart
```diff
@@ -0,0 +1,697 @@
...
+      int start,
+      int end,
+      double score,
+    })>;
+
+final class _State {
+  /// Completion is not active, happens whens:
+  ///  * The input element doesn't have focus, or,
+  ///  * There is a non-empty selection.
+  final bool inactive;
```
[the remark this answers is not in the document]
jonasfj: In theory yes, I think so.

pkg/web_app/lib/src/completion.dart
```diff
@@ -0,0 +1,697 @@
...
+      dropdown.style.display = 'none';
+      return;
+    }
+
+    // Detect changes, first we try just a quick reference equality
+    var changed = _renderedSuggestions != state.suggestions;
+    if (changed) {
+      // Compare all entries
+      changed = _renderedSuggestions.length != state.suggestions.length;
+      for (var i = 0; !changed && i < _renderedSuggestions.length; i++) {
```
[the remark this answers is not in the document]
jonasfj: it will, check the condition :D

jonasfj:
> I'd love the styling to be more "material" eg. like: https://material.angular.io/components/autocomplete/overview#autocomplete-filter

Yes and no,

I want the dropdown to follow the caret, not the input field. Like how auto-complete works in an IDE.

pkg/web_app/lib/src/completion.dart
```diff
@@ -0,0 +1,697 @@
...
+    }
+
+    // If explicily closed using 'Escape', then we handle no more keys
+    if (state.closed) {
+      return;
+    }
+
+    // If not triggered automatically and not forced by Ctrl+Space, then
+    // we're done.
+    if (!state.triggered && !state.forced) {
```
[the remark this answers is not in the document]
jonasfj: It is possible that a class hierarchy would be nice like:
```dart
sealed class State {}
final class InactiveState extends State {}
sealed class ActiveState extends State {}
final class NoSuggestionsState extends ActiveState {}
sealed class CompletingState extends ActiveState {}
final class ActiveTriggeredState extends CompletingState {}
final class ActiveForcedState extends CompletingState {}
```

Or something like that.. But it's also a bit in-sane to be doing that :rofl: 

After all, all the state here is inside one file, and it's a private thing.
It's used when:
 * rendering HTML,
 * tracking state (responding to input, focus change, selection change, etc),
 * handling keyboard input.

That's it.

---------------

The only messy thing right now is that `suggestions` depends on `forced` -- that could be fixed.

app/lib/frontend/templates/views/shared/search_banner.dart
```diff
@@ -64,3 +76,108 @@ d.Node searchBannerNode({
...
+/// Ctrl+Space.
+///
+/// This can generate completion data of different sizes given [topics] and
+/// [licenses].
+String completionDataJson({
+  List<String> topics = const [],
+  List<String> licenses = const [],
+}) =>
+    json.encode({
+      'completions': [
```
isoos, line 98: I think these structures should be typed (similarly to what we have in `pkg/_pub_shared/lib/data/`). It can be a follow-up though, let's add a TODO here.

app/test/frontend/static_files_test.dart
```diff
@@ -208,18 +208,18 @@ void main() {
     test('script.dart.js and parts size check', () {
       final file = cache.getFile('/static/js/script.dart.js');
       expect(file, isNotNull);
-      expect((file!.bytes.length / 1024).round(), closeTo(324, 1));
+      expect((file!.bytes.length / 1024).round(), closeTo(343, 10));
```
isoos, line 211: 1kB sensitivity seems to be more useful to track size changes. With 10kB we could have a change that pushes it up with 9.5kB and then months later a small one pushes it up with 1kB, and we would be bummed.

In this case the small threshold is also a good reminder that this component could be loaded asynchronously, and we should try to skip the extra increase on the main JS file.

Let's add a TODO somewhere.

pkg/web_app/lib/src/widgets.dart
```diff
@@ -0,0 +1,43 @@
...
+      console.error(e.toString().toJS);
+      console.error(st.toString().toJS);
+    }
+  }
+}
+
+extension on NodeList {
+  /// Take a snapshot of [NodeList] as a Dart [List].
+  ///
+  /// Notice that it's not really safe to use [Iterable], because the underlying
```
isoos, line 38: That's a good catch, I'm sure we are using plain `NodeList` as iterable in many places. Any idea how to enforce this better?

pkg/web_css/lib/src/_search.scss
```diff
@@ -20,7 +20,7 @@
   }
 }
 
-.search-bar > .input {
+.search-bar>.input {
```
isoos, line 23: I think the spaces are better for readability. Although, we should rather use embedding instead...

pkg/web_app/lib/src/widgets.dart
isoos: Also: move them into a single place (from `completion.dart` too)

isoos (commented):
I think at this point this is close to be merged, and we should aim to do follow-up changes on specific parts separately from this PR.

pkg/web_css/lib/src/_search.scss
jonasfj: embedding is not the same though, right?

I'm not even sure spaces are -- probably, why am I this bad at CSS :rofl:

pkg/web_app/lib/src/widgets.dart
jonasfj: This is from `package:web` until the rest of the JS is moved to `package:web` it doesn't really make sense.

pkg/web_css/lib/src/_search.scss
isoos: Yeah, embedding is not the same, I meant:
```
.search-bar {
  > .input {
  }
}
```

jonasfj: formatter removes the space though :rofl:

jonasfj: but, yes I'm sold on embedding :D

pkg/web_app/lib/src/widgets.dart
jonasfj: introducing `web_util.dart`.

app/test/frontend/static_files_test.dart
jonasfj: fine I'll set it at 2kb..

I kind of dislike this test, because it's annoying to find it failing so late in the process.

That said, I see the point.

-----------------

> In this case the small threshold is also a good reminder that this component could be loaded asynchronously, and we should try to skip the extra increase on the main JS file.

Yeah, I could see a point in every widget being a thing is asynchronously loaded :rofl:

app/lib/frontend/templates/views/shared/search_banner.dart
jonasfj: yeah, maybe... I kind of wish we didn't have multiple packages...

it'd be much cleaner if we had:
```
lib/src/
  widget/
    completion/
      src/
        ... private bits ...
      widget.dart
      html_as_code.dart
      completions.dart
    widgets.dart  // automatically generated to import widget.dart from all widgets
  ...
```

so that things that are related are located next to each other. Rather than being split out through the code base because they are used in different locations.

But that's a thing I can dream about some other day :rofl:

isoos (approved):
let's merge this + iterate over it

PR #8000, opened 2024-08-29
Admin actions for minimal package, version and publisher info.

### LINKED ISSUE #7988: Admin action: package-info / package-version-info / publisher-info
Should outline information about the package, most notably the publication date.

isoos:
- Fixes #7988
- I think it would be worth to move the related admin action tests from the now-very-big `api_actions_test.dart` into the separate files that I've started for the info actions.


sigurdm (approved):
LGTM

Issue #7981, opened 2024-08-27
Package analysis isn't starting.


mhammerc:
Hello!

Thank you all for this amazing website and ecosystem. Greatly appreciated.

I have published my first package [googleapis_grpc](https://pub.dev/packages/googleapis_grpc/versions).

On the first few versions, the analysis crashed with an error from me, which I have fixed.
Since I have fixed it, analysis isn't starting. Is says "pending analysis" until it abandon waiting for it and says "unknown platforms".

The analysis log content is: "no log".

https://pub.dev/packages/googleapis_grpc/versions/1.0.4/score/log.txt

Therefore, I have no insight on what is wrong in my package to fix it.

I don't know where to go next. Could you help me?

Thank you very much!

isoos:
Thanks for reporting this!

First investigation showed that the backend did not create an appropriate tracking object (`PackageState`) for the analysis, and the backfill for that happens every 7 days. Now you have published a new version, it updated the said object, we'll monitor how the analysis will go. On top of that the package is a bit large, and we may not be able to handle the timeout feedback all the time.

 We have plans to make sure the tracking object is monitored/updated more frequently than one week (it is really about sanity check), and we'll try to check how the analysis of this large package may break down in other ways.

mhammerc:
Thank you for your answer !

The package is not large anymore. Since the last few versions, it is only 3mb. (unless 3mb is considered large!)

isoos:
Note: after a longer investigation we've found that sometimes we are running out of disk space and did not recover from it. In such cases the analysis was stuck in perpetual pending status. Now it seems we have found a way to at least partially recover from it, and the package has a completed analysis: https://pub.dev/packages/googleapis_grpc/versions/1.0.12/score
(Note: the latest published version did not complete analysis yet.)

Keeping the issue open until we collect information about similar cases and what we may need to implement to prevent future occurances.

mhammerc:
Congrats!

On `googleapis_grpc`, I have just discovered that the `dart doc` command produce a `doc` folder of `1.2G`. It is enormous!

moluopro:
This issue already occurred back in August last year, and it still hasn't been resolved? That's really serious, because I've encountered a similar problem too.

https://github.com/dart-lang/pub-dev/issues/8714

sigurdm:
@isoos can this issue be closed?

It seems https://pub.dev/packages/googleapis_grpc/versions has analysis now.

isoos:
[...]

PR #8005, opened 2024-09-02
Restart the task scheduling scan with the initially larger window every 6 hours.

isoos:
#7981


app/lib/task/backend.dart
```diff
@@ -326,6 +332,11 @@ class TaskBackend {
       // be relevant to the next iteration.
       seen.removeWhere((_, updated) => updated.isBefore(since));
 
+      // Do not start waiting if after that we would be over the stop threshold.
+      if (clock.fromNow(minutes: 10).isAfter(stopAfter)) {
+        break;
+      }
```
jonasfj, line 338: Yes, it'll cause it to restart the scanning, but it'll also release the lock, so someone will have to grab the lock again. It'll probably be the same process, but it's just seems like releasing the lock is unnecessary. Maybe I'm overthinking it.

```dart
var nextLongScan = clock.fromNow(hours: 6);

...
var since = clock.ago(days: 3);
while (claim.valid && !abort.isCompleted) {
  ...

  if (clock.now().isAfter(nextLongScan)) {
    // Next time we'll do a longer scan
    since = clock.ago(days: 1);
    nextLongScan = clock.fromNow(hours: 6);
  } else {
    // Next time we'll only consider changes since now - 5 minutes
    since = clock.ago(minutes: 5);
  }

  ...
```

It probably does about the same..

app/lib/task/backend.dart
```diff
@@ -289,8 +289,14 @@ class TaskBackend {
     // Map from package to updated that has been seen.
     final seen = <String, DateTime>{};
 
+    // The scan process will give up control after 6 hours and the restart
+    // will trigger the 3-days period scan at the beginning of the loop.
+    final stopAfter = clock.fromNow(hours: 6);
```
jonasfj, line 294: If we want this the behavior of `_scanForPackageUpdates` changes, so documentation comments **must** be updated accordingly.

If you don't want to write a longer documentation comment that explains how this function weirdly stops running every 6 hours. Then perhaps we should structure it differently :rofl:

jonasfj (changes requested):

app/lib/task/backend.dart
isoos: Updated.

jonasfj (approved):

PR #8011, opened 2024-09-03
Download counts: handle several data files

szakarias:


app/lib/service/download_counts/sync_download_counts.dart
```diff
@@ -56,120 +56,157 @@ Map<String, int> _extractDayCounts(Map<String, dynamic> json) {
   return dayCounts;
 }
 
-Future<bool> processDownloadCounts(
-    String downloadCountsFileName, DateTime date) async {
-  final downloadCountsBucket =
+Future<Set<String>> processDownloadCounts(DateTime date) async {
+  final downloadCountsFileName = [
```
sigurdm, line 60: Is this still used (except for reporting where we probably should use the actual problematic file name)?

app/lib/service/download_counts/sync_download_counts.dart
```diff
@@ -56,120 +56,157 @@ Map<String, int> _extractDayCounts(Map<String, dynamic> json) {
...
-        return;
-      }
-      final String package;
-      final Map<String, int> dayCounts;
-      try {
-        final data = json.decode(line);
-        if (data is! Map<String, dynamic>) {
-          throw FormatException('Download counts data is not valid json');
-        }
+  await for (final f in bucket.list(prefix: fileNamePrefix)) {
```
sigurdm, line 80: Try to only invoke bucket.list once

Maybe do `bucket.list(...).toList();`

app/lib/service/download_counts/sync_download_counts.dart
szakarias: nope, removed it.

sigurdm (approved):

app/lib/service/download_counts/sync_download_counts.dart
szakarias: done

PR #8019, opened 2024-09-05
Expose more information through log.txt when log is missing.

isoos:
#7981

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -975,7 +975,7 @@ class TaskBackend {
     if (data == null) {
       final status = await packageStatus(package);
       if (status.runtimeVersion == null) {
-        return 'no log - `PackageState` entity missing';
+        return 'no log - no tracking information';
       }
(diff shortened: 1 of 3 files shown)


app/lib/task/backend.dart
```diff
@@ -973,13 +973,21 @@ class TaskBackend {
   Future<String?> taskLog(String package, String version) async {
     final data = await gzippedTaskResult(package, version, 'log.txt');
     if (data == null) {
-      return null;
+      final status = await packageStatus(package);
+      if (status.runtimeVersion == null) {
+        return 'no log - `PackageState` entity missing';
```
sigurdm, line 978: I dislike revealing our datastore schema internals in a public facing endpoint.

isoos: how about - `no tracking information`?

sigurdm: Much better!

sigurdm (approved):

PR #8020, opened 2024-09-05
Policy update

jonasfj:


doc/help.md
```diff
@@ -6,9 +6,12 @@ The following help pages are available:
  - [Publishing packages](/help/publishing)
  - [Package scores & pub points](/help/scoring)
  - [`pub.dev` API for developers](/help/api)
+ - [Content moderation](/help/content-moderation)
```
isoos, line 9: We also need handlers similar to this:
https://github.com/dart-lang/pub-dev/blob/master/app/lib/frontend/handlers/routes.dart#L275-L277

sigurdm (approved):

doc/help.md
jonasfj: Yeah, realized that while I was pushing this... got distracted writing https://github.com/dart-lang/pub-dev/issues/8021 :rofl:, but now have I have added a route for the new help page.

PR #8037, opened 2024-09-09
Remove unnecessary getLatestVersion call.

isoos:
By using the resolved URL segment (`/latest/`) we can remove the extra call, and it also makes the rendered dartdoc page's metadata and SEO slightly better (non-latest versions shouldn't be indexed, and they should provide alternative URL).

### DIFF
--- a/app/test/task/testdata/goldens/documentation/oxygen/2.0.0/index.html
+++ b/app/test/task/testdata/goldens/documentation/oxygen/2.0.0/index.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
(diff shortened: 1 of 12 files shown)


app/test/task/testdata/goldens/documentation/oxygen/2.0.0/index.html
```diff
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <meta name="robots" content="noindex"/>
```
jonasfj, line 4: Is this intentional?

jonasfj (approved):

PR #8041, opened 2024-09-09
Update to use and analyze with 3.5 language version

parlough:
- Update to a language version of 3.5
  - Fix resulting analysis warnings from type inference/promotion improvements
- Remove lints removed in recent SDK releases

Feel free to merge if the change looks good to you. I don't have commit access to this repo. Thanks!


isoos:
Note: I the runtime SDK changes, we also need to change the `_acceptedRuntimeVersions` and with that the `runtimeVersion` (`app/lib/shared/versions.dart`). I'm curious: is there any reason to upgrade? (we don't usually track all minor releases).

isoos (approved):

jonasfj (approved):

PR #8054, opened 2024-09-12
Initial draft of a dismissible widget

jonasfj:
Hmm, not sure I might have a SASS formatter that took some liberties :see_no_evil: I can look into fixing that sometime -- or we can consider adopting something like it :/

Hmm, until it's dismissed the content is a bit jumpy.. Ideally, we should make the announcement banner cover some our margin, so that the page doesn't jump whether it's visible or not. The animation was only because it always starts hidden.


pkg/web_app/lib/src/widget/dismissible/widget.dart
```diff
@@ -0,0 +1,105 @@
...
+/// Don't save more than 50 entries
+const _maxMissedMessages = 50;
+
+/// Create a dismissible widget on [element].
+///
+/// A `data-dismissible-by` is required, this must be a CSS selected for a
+/// child element that when clicked will dismiss the message by removing
+/// [element].
+///
+/// A hash is computed from `innerText` of [element], once dismissed this hash
```
sigurdm, line 24: By relying on `innerText` we cannot fix the spelling of a dismissible message without having it pop up again. This might or might not be what we want.

An alternative is a manual id.

pkg/web_app/lib/src/widget/dismissible/widget.dart
```diff
@@ -0,0 +1,105 @@
...
+
+import 'dart:js_interop';
+
+import 'package:collection/collection.dart';
+import 'package:web/web.dart';
+
+import '../../web_util.dart';
+
+/// Forget dismissed messages that are more than 2 years old
+late final _deadline = DateTime.now().subtract(Duration(days: 365 * 2));
```
sigurdm, line 13: Why not store forever (or whatever the browser sets as limit)?

pkg/web_app/lib/src/widget/dismissible/widget.dart
```diff
@@ -0,0 +1,105 @@
...
+
+/// Don't save more than 50 entries
+const _maxMissedMessages = 50;
```
sigurdm, line 16: Why?

jonasfj: Because I don't want the whole thing to be slow.

pkg/web_app/lib/src/widget/dismissible/widget.dart
jonasfj: Because we don't want the thing to be slow...

pkg/web_app/lib/src/widget/dismissible/widget.dart
jonasfj: I think we don't care that a message is displayed twice.

And using `innerText` and `String.hashCode` is unreliable, but worst case is that we display a message more than once, which is not a problem.

pkg/web_css/lib/src/_base.scss
```diff
@@ -41,17 +44,39 @@ body,
...
     font-weight: 400;
   }
 
-  h1 { font-size: 36px; }
-  h2 { font-size: 24px; }
-  h3 { font-size: 18px; }
-  h4 { font-size: 18px; }
-  h5 { font-size: 16px; }
-  h6 { font-size: 16px; }
+  h1 {
```
isoos, line 57: Can we revert these changes? Even if we want to have a new formatting, it would be better to have a separate PR for them.

pkg/web_app/lib/src/widget/dismissible/widget.dart
```diff
@@ -0,0 +1,105 @@
...
+/// with the same `innerText` it'll be removed immediately.
+///
+/// Hashes of dismissed messages will be stored for up to 2 years.
+/// No more than 50 dismissed messages are retained in `localStorage`.
+void create(HTMLElement element, Map<String, String> options) {
+  final by = options['by'];
+  if (by == null) throw UnsupportedError('data-dismissible-by required');
+
+  late final hash = _cheapNaiveHash(element.innerText);
+  if (_dismissed.any((e) => e.hash == hash)) {
```
isoos, line 35: Could we store an id instead (from a `...-dismissible-id` attribute)? The id could have a `<yyyy><mm>-<short-slug>` structure...

pkg/web_app/lib/src/widget/dismissible/widget.dart
```diff
@@ -0,0 +1,105 @@
...
+  return text.length.toRadixString(16).padLeft(2, '0') +
+      text.hashCode.toRadixString(16).padLeft(2, '0') +
+      text.substring(0, half).hashCode.toRadixString(16).padLeft(2, '0') +
+      text.substring(half).hashCode.toRadixString(16).padLeft(2, '0');
+}
+
+/// LocalStorage key where we store the hashes of messages that have been
+/// dismissed.
+///
+/// Data is stored on the format: `<hash>@<date>;<hash>@<date>;...`
```
isoos, line 78: An idea: using a JSON array here would simplify parsing and mostly eliminate the risk of having an unexpected parsing bug for an extra overhead of JSON string escaping...

isoos, line 78: Another idea: let's store the milliseconds instead of the formatted date.
Also: let's document that the format of the `<date>`.

pkg/web_app/lib/src/widget/dismissible/widget.dart
```diff
@@ -0,0 +1,105 @@
...
+void create(HTMLElement element, Map<String, String> options) {
+  final by = options['by'];
+  if (by == null) throw UnsupportedError('data-dismissible-by required');
+
+  late final hash = _cheapNaiveHash(element.innerText);
+  if (_dismissed.any((e) => e.hash == hash)) {
+    element.remove();
+    return;
+  } else {
+    element.style.display = 'revert';
```
isoos, line 39: Nit: the current content security policies allow this, but we had a plan to lock down the page by not allowing any such style update. I'm not sure if it is possible with the completion widget's positioning, but in this case, we could have a class that makes this widget visible.

isoos (changes requested):
Mostly nits, looks good overall, but I really want to separate the CSS formatting changes.

pkg/web_app/lib/src/widget/dismissible/widget.dart
jonasfj: Yes, but then every time we make a new message we'd have to make a new ID.

Hmm, actually, I buy it, we can compute the hash on the server as an id, sold.

pkg/web_app/lib/src/widget/dismissible/widget.dart
isoos: Hm, at that point we could just do `<hash/id>:<DateTime/millisecondsAsInt>` Map, instead of a list.

pkg/web_app/lib/src/widget/dismissible/widget.dart
jonasfj: Are there CSP policies to prevent messing with style on individual elements?

I made it into a `dismissed` class instead, that seems pretty reasonable.

pkg/web_app/lib/src/widget/dismissible/widget.dart
jonasfj: I'll document it, though it's read / written just below, and is local to this file.

But yes, it's nice to have when we go back to debug things.

I opted for `YYYY-MM-DD` because anyone seeing this in localstorage will know that it's a date. So it's easier to debug :D

And I see no need for accuracy.

jonasfj: actually, now that `data-dismiss-message-id` is a thing, I can't assume that the ID won't include `@` and `;` -- this was safe before when I know it was a hash.

jonasfj: How about just base64 encode it... `window.btoa` is easy to use :rofl:

isoos (approved):

sigurdm (approved):

PR #8055, opened 2024-09-13
Remove moderated package and version assets after 3 years.

isoos:
- Reuses existing remove package logic, with the following updates:
  - reordering delete operations, in order to prevent information loss if the process fails mid-deletion,
  - also updates `replacedBy` fields if they are referencing the moderated package,
  - also deleted `AuditLogRecord` entities, otherwise the integrity check fails on them, (we could update the integrity checks to accept `ModertedPackage` existence in such cases though)
  - also adds previously deleted versions to the moderated package entity's list

- Deleting package version is simpler in comparison, it needs to remove only a few references.
- Deleting moderated publisher or user is implemented separately, left TODO for them.

### DIFF
--- a/app/lib/admin/backend.dart
+++ b/app/lib/admin/backend.dart
@@ -12,7 +12,9 @@ import 'package:clock/clock.dart';
 import 'package:collection/collection.dart';
 import 'package:convert/convert.dart';
 import 'package:gcloud/service_scope.dart' as ss;
+import 'package:gcloud/storage.dart';
 import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
 import 'package:pool/pool.dart';
 
 import '../account/backend.dart';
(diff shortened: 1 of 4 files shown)


app/lib/admin/backend.dart
```diff
@@ -300,10 +307,15 @@ class AdminBackend {
   Future<void> removePackage(String packageName) async {
     final caller =
         await requireAuthenticatedAdmin(AdminPermission.removePackage);
-
     _logger.info('${caller.displayId}) initiated the delete '
         'of package $packageName');
+    await _removePackage(packageName);
+  }
 
+  Future<void> _removePackage(
```
jonasfj, line 315: nit: documentation comment :rofl:

app/lib/admin/backend.dart
```diff
@@ -312,6 +324,48 @@ class AdminBackend {
             .toList())
         .toSet();
 
+    final pool = Pool(10);
+    final futures = <Future>[];
+    versions.forEach((final v) {
+      futures.add(pool.withResource(
+          () => packageBackend.removePackageTarball(packageName, v)));
```
jonasfj, line 331: will a 404 error be ignored?

app/lib/admin/backend.dart
```diff
@@ -312,6 +324,48 @@ class AdminBackend {
             .toList())
         .toSet();
 
+    final pool = Pool(10);
+    final futures = <Future>[];
+    versions.forEach((final v) {
```
jonasfj, line 329: nit: Consider `versions.map`

jonasfj (approved):

app/lib/admin/backend.dart
isoos: Yes, they are ignored. Also added a comment about it here.

Issue #8053, opened 2024-09-12
Create an allow-list of packages that can publish more than 1000 versions


jonasfj:
By default we limit packages to publishing at-most 1000 package versions.
This aims to:
 * Keep the size of the version listing API end-point reasonable (which would otherwise slow down `pub get`).
 * Mitigate abuse of the system.

However, if a package hits this limit we should have an allow-list that a package can be added to such that the limit can be increased for a specific package.

We would ask package authors to reduce the publication frequency, when they are allow-listed to published more than 1000 package versions.

------------

Ideally, it shouldn't be necessary to publish updates more than once per week (if even that much). Obviously, there can be extenuating circumstances, if there is a serious bug, security vulnerability, etc. or perhaps you're making a big release and publishing a few quick fixes.

But in general, we don't believe that it's a good idea to publish tiny changes extremely frequently. This causes churn for all the people who depend on your package, as they will be prompted to upgrade.

------------

We should probably add a section in our documentation with guidelines around publishing frequency.

PR #8058, opened 2024-09-16
Increase version limit for package already reaching 1000.

isoos:
#8053

### DIFF
--- a/app/lib/package/overrides.dart
+++ b/app/lib/package/overrides.dart
@@ -103,3 +103,8 @@ bool isDartDevPublisher(String? publisherId) {
   if (publisherId.endsWith('.google.com')) return true;
   return false;
 }
+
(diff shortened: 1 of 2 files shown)


app/lib/package/overrides.dart
```diff
@@ -103,3 +103,8 @@ bool isDartDevPublisher(String? publisherId) {
   if (publisherId.endsWith('.google.com')) return true;
   return false;
 }
+
+/// Overriding the default maximum number of the allowed package version counts.
+const maxVersionsPerPackageOverrides = <String, int>{
+  'masamune': 1100, // last updated: 2024-09-16
```
jonasfj, line 109:
```suggestion
  'masamune': 1500, // last updated: 2024-09-16
```

jonasfj (approved):

PR #8060, opened 2024-09-17
Deletion of moderated publishers and users (after three years)

isoos:
#7535


jonasfj (approved):

jonasfj:
Would be great if @sigurdm or @szakarias could take a look too. Just would be sad to this one wrong :)

sigurdm (commented):
Looks solid to me

PR #8063, opened 2024-09-18
Long-term rate limits for package uploads.

isoos:
- #8053
- The current rate limit implementation relies on reading all of the audit log records from the datastore and keeping them in memory. While we could read a full year's worth of logs into the memory, it seemed better to provide a separate path for package uploads, as we do have the timestamps at the time of rate limit checks.
- I've added additional checks that (a) we shouldn't specify long-term rate limits for other actions and (b) if we do and we don't provide the timestamps, it should fail.
- In order to reduce the surprises, I think we could start introducing the `weekly` limit, and then shortly after that the `monthly`, while the `quarterly` limit should be added only later this year, and the `yearly` in the next year. That way packages that would be over the specific limit have time to adopt.
- Giving a feedback of the remaining limit may come in a subsequent PR.

### DIFF
--- a/app/config/dartlang-pub.yaml
+++ b/app/config/dartlang-pub.yaml
@@ -75,6 +75,10 @@ rateLimits:
     burst: 3
     hourly: 6
     daily: 12
+    weekly: 20      #  20 versions/week
+    # monthly: 32   # ~ 8 versions/week
+    # quarterly: 50 # ~ 4 versions/week
+    # yearly: 100   # ~ 2 versions/week
   - operation: package-published
     scope: user
     burst: 200
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -1085,11 +1085,17 @@ class PackageBackend {
           'Package "${newVersion.package}" has no admin email to notify.');
     }
 
-    // check rate limits before the transaction
+    final existingVersions = await db
+        .query<PackageVersion>(ancestorKey: newVersion.packageKey!)
+        .run()
+        .toList();
+
+    // check long term (over a day) rate limits with version created timestamps
     await verifyPackageUploadRateLimit(
       agent: agent,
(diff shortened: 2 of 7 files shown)


app/config/dartlang-pub.yaml
```diff
@@ -75,6 +75,10 @@ rateLimits:
     burst: 3
     hourly: 6
     daily: 12
+    weekly: 20      #  20 versions/week
+    # monthly: 32   # ~ 8 versions/week
+    # quarterly: 50 # ~ 4 versions/week
+    # yearly: 100   # ~ 2 versions/week
```
jonasfj, line 81: I think the problem with monthly, quarterly and yearly is that these limits are so long that we need a mechanism to override them.

If you publish ~50 versions per quarter, then suddenly you can't publish for 6 months.
I don't think that's viable.

------------

I do think that 20 publications per week seems like a reasonable limit. And having to wait 5 days is not an entirely unreasonable ask (in such scenarios, hopefully there are few).

jonasfj: I think a client-side warning or error is probably better:
https://github.com/dart-lang/pub/issues/4392

We just want a soft limit, really.

isoos: > If you publish ~50 versions per quarter, then suddenly you can't publish for 6 months.

The daily, weekly and monthly limit still apply, and it is unlikely that you would trigger only the quarterly and nothing before. However, if you do hit this scenario, your events are stretched out over the quarter. As the window is rolling, your prior events will move out of it, you won't need to wait for the entire quarter to get more quota.

Maybe the events could have an exponential decay in them. E.g. after a month an event may count as only 0.5 event, and as such, the emergency quota gets freed up a bit faster?

isoos:
Closing this in favor of #8441.

PR #8064, opened 2024-09-19
New naming scheme for theme roles + migrating some background colors.

isoos:
- Deployed to staging for further verification.
- Started to separate named colors in an explicit way like `smokeWhite` and `aliceBlue`. These names are either from the HTML color names or if there is no such name, a widely used name is picked for it.
- Merged very similar colors into a single one, e.g. `f8f8f8` and `f8f9fa` is both merged into `f5f5f7`. Fewer colors makes a more consistent color palette, and the contrast/separation between them is easier to review.
- Started to refactor and use names that first focus on their role `neutral`, `inset`, `selected` and then specify their property like `bgColor` or `hover-bgColor`. This makes it much easier to pick the appropriate variable when selecting colors for a component, e.g. selection or hovering should be consistent across the components this way.
- The PR is mostly for some of the background colors, but wanted to verify the naming style before spending much time on it.

### DIFF
--- a/pkg/web_css/lib/dartdoc.scss
+++ b/pkg/web_css/lib/dartdoc.scss
@@ -31,7 +31,7 @@
 
 /* Note: light-theme was added to prevent non-visible table text in dartdoc's dark mode. */
 .light-theme .markdown-body table tr {
-  background-color: var(--pub-default-background-color);
+  background-color: var(--pub-neutral-bgColor);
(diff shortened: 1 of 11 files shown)


sigurdm:
Can we keep the colder highlight color?

isoos:
Closing as we have split this into smaller parts and reviewed it separately.

PR #8066, opened 2024-09-20
Foldable sections on package and publisher admin UI

isoos:
image

(Uploaded to staging)


isoos:
Closing as we don't want to loose easy in-page search.

PR #8068, opened 2024-09-23
Do not cache top packages if search result list empty (search not ready).

### LINKED ISSUE #8057: Empty landing page - instance restart and cache may not be populated?
I've encountered an empty landing page on `pub.dev`. Locally this happens when the local caches are not yet populated.
We should probably make an attempt at service startup to populate them before opening the HTTP port and marking it ready.

isoos:
Fixes #8057

### DIFF
--- a/app/lib/search/top_packages.dart
+++ b/app/lib/search/top_packages.dart
@@ -140,7 +140,10 @@ CachedValue<List<PackageView>> _cachedValue(
         // Do not apply rate limit here.
         rateLimitKey: null,
       );
-      return page.packageHits;
+      // Returning `null` will not cache the value, returning empty list gets cached.
+      return page.errorMessage == null && page.packageHits.isNotEmpty
+          ? page.packageHits
+          : null;
     },
   );
 }


app/lib/search/top_packages.dart
```diff
@@ -140,7 +140,10 @@ CachedValue<List<PackageView>> _cachedValue(
         // Do not apply rate limit here.
         rateLimitKey: null,
       );
-      return page.packageHits;
+      // Returning `null` will not cache the value, returning empty list gets cached.
+      return page.errorMessage == null && page.packageHits.isNotEmpty
```
sigurdm, line 144: Is it not enough to just look at the `page.errorMessage`?

isoos: In theory (and with the current state of the code) it would be enough to just look for the `errorMessage`. On the other hand I think it wouldn't hurt to be more defensive and cache only non-empty results.

sigurdm (approved):
I don't know why we would not trust an empty successful result.

But I doubt it makes much difference in practice...

jonasfj (approved):

PR #8069, opened 2024-09-23
Defending pana log writes from cross-stream data.

isoos:
- #7981
- There may be a little chance that the stdout and stderr streams get crossed, using line splitting makes it separated.
- Added additional logging and also killing the pana process when unexpected error happen.

### DIFF
--- a/pkg/pub_worker/lib/src/analyze.dart
+++ b/pkg/pub_worker/lib/src/analyze.dart
@@ -3,7 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:convert' show JsonUtf8Encoder, utf8;
+import 'dart:convert' show JsonUtf8Encoder, LineSplitter, Utf8Decoder, utf8;
 import 'dart:io'
     show Directory, File, IOException, Platform, Process, ProcessSignal, gzip;
 import 'dart:isolate' show Isolate;
@@ -152,8 +152,14 @@ Future<void> _analyzePackage(
       await pana.stdin.close();
 
       await Future.wait<void>([
-        pana.stderr.forEach(log.add),
-        pana.stdout.forEach(log.add),
+        pana.stderr
+            .transform(Utf8Decoder(allowMalformed: true))
+            .transform(LineSplitter())
+            .forEach(log.writeln),
+        pana.stdout
+            .transform(Utf8Decoder(allowMalformed: true))
+            .transform(LineSplitter())
+            .forEach(log.writeln),
         pana.exitOrTimeout(_panaTimeout, () {
           log.writeln('TIMEOUT: pana sending SIGTERM/SIGKILL');
         }),


pkg/pub_worker/lib/src/analyze.dart
```diff
@@ -152,12 +152,28 @@ Future<void> _analyzePackage(
       await pana.stdin.close();
 
       await Future.wait<void>([
-        pana.stderr.forEach(log.add),
-        pana.stdout.forEach(log.add),
+        pana.stderr
+            .transform(utf8.decoder)
+            .transform(LineSplitter())
```
jonasfj, line 157: What kind of things can this throw?

Certainly it can throw FormatException, do we want it to be UTF8? If so I certainly think we should accept malformed utf8 from the subprocess.

isoos: How about `Utf8Decoder(allowMalformed: true)`? I think unconditional stream merge is similarly error-prone.

pkg/pub_worker/lib/src/analyze.dart
```diff
@@ -152,12 +152,28 @@ Future<void> _analyzePackage(
...
+      ]).catchError((e, st) {
+        log.writeln('Error waiting for pana process: $e');
+        log.writeln(st);
+        // kill the process if not already killed
+        try {
+          pana.kill(ProcessSignal.sigkill);
+        } finally {
+          // ignore
+        }
+        return const [];
```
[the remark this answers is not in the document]
isoos: I think the killing the process part is important. The extra log exposure is not that important, but we should log that we are killing the process here with `sigkill`.

isoos:
PTAL.

pkg/pub_worker/lib/src/analyze.dart
```diff
@@ -152,12 +152,27 @@ Future<void> _analyzePackage(
...
-      ]).catchError((e) => const [/* ignore */]);
+      ]).catchError((e, st) {
+        log.writeln('TIMEOUT: sending SIGKILL to pana.');
+        // kill the process if not already killed
+        try {
+          pana.kill(ProcessSignal.sigkill);
+        } finally {
+          // ignore
+        }
+        return const [];
```
jonasfj, line 174: `Future.wait` won't return until all futures are done.

And `exitOrTimeout` does afaik the right thing:
https://github.com/dart-lang/pub-dev/blob/1f628e218caca0a3fecc9201ac465068c2d1e391/pkg/pub_worker/lib/src/analyze.dart#L302-L326

It even does the graceful shutdown thing.

If for some reason the pipe to stderr should fail (say diskspace is full and `log.writeln` fails), then we'll just hang waiting for the timeout.

Then what is the point of calling `sigkill` again.

I think we should ignore the error here as we did before, I don't see any reason to catch log it.

I wouldn't mind a `.catchError((e, st) => _log....` after `.forEach(log.writeln)` (in both places).
But we could also just live without it.

isoos: Good points. I've reverted this part.

jonasfj (approved):

PR #8070, opened 2024-09-23
Attempt to report failed analysis + also include the log.txt generated up to that point.

isoos:
#7981

### DIFF
--- a/pkg/pub_worker/lib/src/analyze.dart
+++ b/pkg/pub_worker/lib/src/analyze.dart
@@ -55,15 +55,18 @@ Future<void> analyze(Payload payload) async {
   final workerDeadline = clock.now().add(_workerTimeout);
   final client = Client().withUserAgent(pubWorkerUserAgent);
   try {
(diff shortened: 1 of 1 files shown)


pkg/pub_worker/lib/src/analyze.dart
```diff
@@ -91,10 +95,30 @@ Future<void> analyze(Payload payload) async {
...
+          client,
+          api,
+          payload.package,
+          vtp.version,
+          reason: [
+            'Processing of `${payload.package}/${vtp.version}` failed with the following error:\n$e',
+            st.toString(),
+            if (logContent != null) logContent,
+          ].join('\n\n'),
+        );
```
jonasfj, line 119: I think the result of this will be that the package won't be retried with backoff.

jonasfj: When pana fails, analysis should not be retried.

We only want to retry infrastructure failures.

isoos:
I think we may not need this after all.

PR #8077, opened 2024-09-25
Refactor CSS names and merge colors: neutral background.

isoos:
- Partial reimplementation of #8064, only changing one color role at a time, starting with the neutral (default) background.
- The `neutral` role could be thought as synonymous to `default`, but it is a bit more, e.g. the home screen's cards are also having the same background (over e.g. a grid or a grey background). Also, with the two themes, there is no default background, as the system settings may use the dark mode as default, hence the rename to `neutral`.
- The change merges dark mode's `#303030` home card background with the neutral (default) background `#212121`. This is to follow the light theme, where the card has the same background as the default page. `#212121` was later switched to `darkGunmetal` as Flutter website's new dark mode will use that.
- The new CSS role variable is used as a generic value for setting component's backgrounds, instead of having a separate variable for each component.

### DIFF
--- a/pkg/web_css/lib/dartdoc.scss
+++ b/pkg/web_css/lib/dartdoc.scss
@@ -31,7 +31,7 @@
 
 /* Note: light-theme was added to prevent non-visible table text in dartdoc's dark mode. */
 .light-theme .markdown-body table tr {
-  background-color: var(--pub-default-background-color);
+  background-color: var(--pub-neutral-bgColor);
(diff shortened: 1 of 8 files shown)


sigurdm (approved):

isoos:
Note: applied `darkGunmetal` color from the Flutter website's upcoming dark mode (it was one of the named colors closest to what they are using).

PR #8079, opened 2024-09-26
Mitigate out of free space issues while running dartdoc postprocessing.

isoos:
- #7981
- By deleting the input dartdoc data, the post-processing in both isolates will stop.
- We are deleting the partial output as we won't use it, and we may need free space to complete writing the log.txt and the blob files.

### DIFF
--- a/pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
+++ b/pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
@@ -103,6 +103,7 @@ Future<void> postProcessDartdoc({
     await tmpTar.deleteIgnoringErrors();
 
     _log.shout('Failed to run dartdoc post-processing.', e, st);
+    rethrow;
   }
 }
 
--- a/pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
+++ b/pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
@@ -98,12 +98,6 @@ Future<void> postProcessDartdoc({
     await tmpTar.rename(p.join(outputFolder, 'doc', 'package.tar.gz'));
   } catch (e, st) {
     // We are likely out of free disk space.
-    //
-    // One of the isolate has already stopped, the other may be still working.
-    // By deleting the dartdoc-generated content, the post-processing of the other
-    // isolate will stop too.
-    await Directory(docDir).deleteIgnoringErrors(recursive: true);
-
     // Cleanup the content as we won't use it.
     await tmpOutDir.deleteIgnoringErrors(recursive: true);
     await tmpTar.deleteIgnoringErrors();
--- a/pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
+++ b/pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
@@ -97,8 +97,8 @@ Future<void> postProcessDartdoc({
     await tmpOutDir.rename(p.join(outputFolder, 'doc'));
     await tmpTar.rename(p.join(outputFolder, 'doc', 'package.tar.gz'));
   } catch (e, st) {
-    // We are likely out of free disk space.
-    // Cleanup the content as we won't use it.
+    // We don't know what's wrong with the post-processing, and we treat
+    // the output as incorrect. Deleting, so it won't be included in the blob.
     await tmpOutDir.deleteIgnoringErrors(recursive: true);
     await tmpTar.deleteIgnoringErrors();
 
--- a/pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
+++ b/pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
@@ -24,15 +24,15 @@ Future<void> postProcessDartdoc({
   required String docDir,
 }) async {
   _log.info('Running dartdoc post-processing');
-  final tmpOutDir = p.join(outputFolder, '_doc');
-  await Directory(tmpOutDir).create(recursive: true);
+  final tmpOutDir = Directory(p.join(outputFolder, '_doc'));
+  await tmpOutDir.create(recursive: true);
   final processingFuture = Isolate.run(() async {
     final files = Directory(docDir)
         .list(recursive: true, followLinks: false)
         .whereType<File>();
     await for (final file in files) {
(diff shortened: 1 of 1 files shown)


pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
```diff
@@ -85,14 +85,39 @@ Future<void> postProcessDartdoc({
...
+    // This helps if there is a timeout along the way.
+    await tmpOutDir.rename(p.join(outputFolder, 'doc'));
+    await tmpTar.rename(p.join(outputFolder, 'doc', 'package.tar.gz'));
+  } catch (e, st) {
+    // We are likely out of free disk space.
+    //
+    // One of the isolate has already stopped, the other may be still working.
+    // By deleting the dartdoc-generated content, the post-processing of the other
+    // isolate will stop too.
+    await Directory(docDir).deleteIgnoringErrors(recursive: true);
```
jonasfj, line 105: I don't understand this.

Who owns the `docDir`, I think we should say that it's owned by the caller to `postProcessDartdoc` as is currently the case.

And that `postProcessDartdoc` only read from `docDir`.

It feels messy to be sometimes deleting it here and sometimes not.
If we want to delete it here, we should always do so, and we should document it for the caller.

However, I see no reason to do that.

--------------------------

Here we're trying to stop the isolates from running by deleting their input files.
I think that's a bit wild. Perhaps it'd be more reasonable to:
 * (A) Not run the two thing concurrently, so we don't have to use isolates.
 * (B) Create a `runInIsolate<T>(FutureOr<T> Function() computation, {Completer<void> cancel})`.

Obviously, (B) would require that we create a tiny helper library for running stuff in an isolate with cancellation. And that we have tests :rofl: 

I think the correct approach for what you're trying to do is (B), but that the simply solution is (A). Afterall, it's fairly likely that both these operations are going to be I/O bound (maybe not post-processing).

pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
```diff
@@ -85,14 +85,39 @@ Future<void> postProcessDartdoc({
...
-  await tmpTar.rename(p.join(outputFolder, 'doc', 'package.tar.gz'));
+    // Move from temporary output directory to final one, ensuring that
+    // documentation files won't be present unless all files have been processed.
+    // This helps if there is a timeout along the way.
+    await tmpOutDir.rename(p.join(outputFolder, 'doc'));
+    await tmpTar.rename(p.join(outputFolder, 'doc', 'package.tar.gz'));
+  } catch (e, st) {
+    // We are likely out of free disk space.
+    //
+    // One of the isolate has already stopped, the other may be still working.
```
jonasfj, line 102: That's not the semantics of [`Future.wait`](https://api.dart.dev/stable/3.5.3/dart-async/Future/wait.html).

pkg/pub_worker/lib/src/bin/pana_wrapper.dart
```diff
@@ -167,7 +167,10 @@ Future<void> main(List<String> args) async {
     docDir: rawDartdocOutputFolder.path,
   );
 
-  await rawDartdocOutputFolder.delete(recursive: true);
+  // existence check is needed as post-processing may delete this directory
```
jonasfj, line 170: When this method creates the folder, this method should probably also delete it.

pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
isoos: Good catch. Fixed it. We may wait for both, as it is likely that both will be running out of the free space the same time.

pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
isoos: Fixed the `docDir` deletion. Agreed on (B) being the better one, but for a simple handling we may just do the cleanup.

pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
```diff
@@ -85,14 +85,33 @@ Future<void> postProcessDartdoc({
...
+    await tmpOutDir.rename(p.join(outputFolder, 'doc'));
+    await tmpTar.rename(p.join(outputFolder, 'doc', 'package.tar.gz'));
+  } catch (e, st) {
+    // We are likely out of free disk space.
+    // Cleanup the content as we won't use it.
+    await tmpOutDir.deleteIgnoringErrors(recursive: true);
+    await tmpTar.deleteIgnoringErrors();
+
+    _log.shout('Failed to run dartdoc post-processing.', e, st);
+  }
```
jonasfj, line 106: So now now it'll just complete successfully without any content.

Are we sure that's the behavior we want?

The log messages here go to stackdriver. Not the package task-log.

Which is IMO also the right thing.

But successfully completing the task is probably not desirable. Or at-least if we do, we should write in the task-log that something failed.

jonasfj: To be fair, I'm not sure we fully understand what it is we want to do. We should perhaps chat a bit more about this.

pkg/pub_worker/lib/src/bin/dartdoc_wrapper.dart
```diff
@@ -85,14 +85,33 @@ Future<void> postProcessDartdoc({
...
+    // This helps if there is a timeout along the way.
+    await tmpOutDir.rename(p.join(outputFolder, 'doc'));
+    await tmpTar.rename(p.join(outputFolder, 'doc', 'package.tar.gz'));
+  } catch (e, st) {
+    // We are likely out of free disk space.
+    // Cleanup the content as we won't use it.
+    await tmpOutDir.deleteIgnoringErrors(recursive: true);
+    await tmpTar.deleteIgnoringErrors();
+
+    _log.shout('Failed to run dartdoc post-processing.', e, st);
```
jonasfj, line 105: At a minimum we should exit non-zero here, right?

jonasfj: or rethrow?

jonasfj (approved):

PR #8085, opened 2024-09-27
Name matches at the beginning of search.

### LINKED ISSUE #7938: An exact match with the package name should be shown first
Search for the `stream_channel` package on pub.dev: https://pub.dev/packages?q=stream_channel

Package named `stream_channel` is only the 4th result, it should be the first result:

image

isoos:
Fixes #7938
image

### DIFF
--- a/app/lib/frontend/templates/listing.dart
+++ b/app/lib/frontend/templates/listing.dart
@@ -5,9 +5,11 @@
 import 'dart:math';
 
 import 'package:_pub_shared/search/search_form.dart';
+import 'package:collection/collection.dart';
 
 import '../../package/search_adapter.dart';
 import '../../search/search_service.dart';
+import '../../shared/urls.dart' as urls;
 import '../dom/dom.dart' as d;
 
 import '_consts.dart';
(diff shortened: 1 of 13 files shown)


app/lib/frontend/templates/listing.dart
```diff
@@ -121,3 +123,20 @@ class PageLinks {
...
 }
+
+d.Node? _nameMatches(List<String>? matches) {
+  if (matches == null || matches.isEmpty) {
+    return null;
+  }
+  final singular = matches.length == 1;
+  final namePluralized = singular ? 'name' : 'names';
+  return d.p(children: [
+    d.b(text: 'Matching package $namePluralized: '),
```
sigurdm, line 134: Could we detect if the match is exact, and in that case write 'Exact package name match:'

sigurdm:
I'd like the message to be more specific when applicable. But I really like this!

sigurdm (approved):

jonasfj (approved):

app/lib/frontend/templates/listing.dart
isoos: Updated.

PR #8087, opened 2024-09-30
Move getting the blobId closer to its actual use.

isoos:
I have no evidence that the index building process time would affect the errors much, but moving it closer to its use seems to be a safe bet.

### DIFF
--- a/pkg/pub_worker/lib/src/analyze.dart
+++ b/pkg/pub_worker/lib/src/analyze.dart
@@ -164,13 +164,6 @@ Future<void> _analyzePackage(
       log.writeln('STOPPED: ${clock.now().toUtc().toIso8601String()}');
     }
 
-    // Upload results, if there is any
-    _log.info('api.taskUploadResult("$package", "$version")');
-    final r = await retry(
-      () => api.taskUploadResult(package, version),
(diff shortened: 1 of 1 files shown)


pkg/pub_worker/lib/src/analyze.dart
```diff
@@ -201,6 +194,13 @@ Future<void> _analyzePackage(
...
 
+    // Upload results, if there is any
+    _log.info('api.taskUploadResult("$package", "$version")');
+    final r = await retry(
+      () => api.taskUploadResult(package, version),
+      retryIf: _retryIf,
+    );
+
     // Create BlobIndex
     final index = await builder.buildIndex(r.blobId);
```
jonasfj, line 205: If you want to, there is an extremely efficient `blobIndex.update` method, if you'd rather update the blobId after the fact.

jonasfj (approved):

jonasfj:
This is really nice, it could fix some of the permission issues.

pkg/pub_worker/lib/src/analyze.dart
isoos: Maybe, in a subsequent change.

PR #8092, opened 2024-10-01
Improve banner CSS

jonasfj:
* On standalone pages the banner will sit in page and be hidden.
 * On package page (and landing page) the banner will float over content with position absolute, as these pages have sufficient margin to allow for this.

In both cases content will not jump when the banner is dismissed, or when the banner is loaded.

### DIFF
--- a/pkg/web_css/lib/src/_base.scss
+++ b/pkg/web_css/lib/src/_base.scss
@@ -407,19 +407,6 @@ pre {
   }
 
   z-index: 1000;
-
-  position: absolute;
-  left: 0px;
-  right: 0px;
-}
-
-.page-standalone .announcement-banner {
-  position: initial;
-
-  &.dismissed {
-    display: block;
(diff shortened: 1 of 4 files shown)


isoos (approved):
Can you push this to staging?

jonasfj:
Pushed to `staging-b`

pkg/web_css/lib/src/_base.scss
```diff
@@ -393,31 +393,20 @@ pre {
   text-align: center;
 
   .dismisser {
-    float: right;
+    position: absolute;
```
isoos, line 396: Not sure: does this need a `position: relative` in the parent element?

isoos (approved):

pkg/web_css/lib/src/_base.scss
jonasfj: I don't think so...

PR #8093, opened 2024-10-01
Fix automatic status on resolving an appeal.

isoos:
- Previous behavior (with bad test): (1) package gets moderated (2) appeal is filed (3) moderation is reverted (4) appeal is closed with automatic status: `moderation-upheld`. The correct status for this is `moderation-reverted`.
- The new behavior merged the actions from the original (appealed) `ModerationCase` and the current (appeal) `ModerationCase` and calculates the cumulative effect. If there are no moderated entities left, the case is considered `reverted` otherwise `upheld`.


isoos:
Closing in favor of #8095

PR #8096, opened 2024-10-02
Admin action to generate report for moderation transparency metrics.

isoos:


app/lib/admin/actions/moderation_transparency_metrics.dart
```diff
@@ -0,0 +1,328 @@
...
+          mc.status == ModerationStatus.noActionReverted) {
+        totalModerationCount++;
+        violations.increment(mc.violation ?? '');
+        sources.increment(mc.source);
+
+        final hasUserRestriction = mc
+            .getActionLog()
+            .entries
+            .any((e) => ModerationSubject.tryParse(e.subject)!.isUser);
+        if (hasUserRestriction) {
```
jonasfj, line 85:
```suggestion
        // If actions resulted in a user being blocked, then we count it as
        // "provision", even if packages were also removed.
        // Reasoning that it's natural that blocking a user would also
        // remove some of the users content.
        if (hasUserRestriction) {
```

jonasfj (approved):

PR #8101, opened 2024-10-03
Table of Contents for package admin page.

isoos:
- Deployed to staging for verification.
- The floating part is extremely limited, but provides 80% of the value for very little investment, and we can iterate on it later. When the display is narrow, the table of contents is inlined as part of the admin page.
- Floating example:
image

### DIFF
--- a/app/lib/frontend/templates/views/shared/toc.dart
+++ b/app/lib/frontend/templates/views/shared/toc.dart
@@ -1,0 +1,51 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import '../../../dom/dom.dart' as d;
+
+/// Renders Table of Contents from [nodes].
+///
+/// The hiearchy of the sections is encoded in [TocNode.children].
+d.Node renderToc(Iterable<TocNode> nodes) {
+  return d.div(
+    classes: ['pub-toc-container'],
+    child: d.div(
+      classes: ['pub-toc'],
+      children: [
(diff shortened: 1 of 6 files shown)


isoos:
Note: removed the floating part for now, let's iterate on it separately.

jonasfj:
We should probably deploy this to staging.

jonasfj:
@sigurdm and @szakarias do you have opinions?

app/lib/frontend/templates/views/shared/toc.dart
```diff
@@ -0,0 +1,51 @@
...
+/// Renders Table of Contents from [nodes].
+///
+/// The hiearchy of the sections is encoded in [TocNode.children].
+d.Node renderToc(Iterable<TocNode> nodes) {
+  return d.div(
+    classes: ['pub-toc-container'],
+    child: d.div(
+      classes: ['pub-toc'],
+      children: [
+        d.div(child: d.b(text: 'Contents')),
```
sigurdm, line 16: Open question. Should we (specifically for the admin page) call this 
```suggestion
        d.div(child: d.b(text: 'Sections')),
```

sigurdm:
There's ...something.. about the layout. I cannot quite put words to it.

But I think it is good enough for starters. We can always iterate on it.

sigurdm (approved):
This is fine.

I'm OK with keeping 'Contents' just wanted to consider it.

app/lib/frontend/templates/views/shared/toc.dart
isoos: I think it is fine to call it `Sections` on any page really. It is not that we can guarantee that a random README actually has a content, only the section headers.

PR #8104, opened 2024-10-04
Interface for writing exported API

jonasfj:
This is a proposal for an interface for writing the exported API.

The idea is to put everything related to managing the actual objects of the bucket into a single abstraction.

This interface isn't responsible for figuring out:
 * what to write, or,
 * when to write it.

Only **where** to write things.

The hope is that this makes it less likely that we accidentally writes file to the wrong location, forget to update them, or forget to GC particular files.

### DIFF
--- a/app/lib/package/api_export/exported_api.dart
+++ b/app/lib/package/api_export/exported_api.dart
@@ -1,0 +1,327 @@
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:clock/clock.dart';
+import 'package:gcloud/storage.dart';
+import 'package:logging/logging.dart';
+import 'package:pool/pool.dart';
+import 'package:pub_dev/shared/utils.dart';
+import '../../shared/versions.dart'
+    show runtimeVersion, runtimeVersionPattern, shouldGCVersion;
+
+final _log = Logger('api_export:exported_bucket');
+
+/// Interface for [Bucket] containing exported API that is served directly from
+/// Google Cloud Storage.
+///
+/// This interface is responsible for:
+///  * Naming of files in the bucket.
+///  * Deletion of all files related to a package.
+///  * Garbage collection of unknown package entries.
+///  * Maintaining two prefixes with files in the bucket.
+///    * "latest" (that all runtimeVersions write to)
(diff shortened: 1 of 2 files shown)


app/lib/shared/utils.dart
```diff
@@ -299,4 +299,25 @@ extension StringExt on String {
...
+  /// If [suffix] is not present, it will not be removed.
+  ///
+  /// **Example:**
+  /// ```
+  /// assert('dart:io'.without(prefix: 'dart:') == 'io');
+  /// assert('file.txt'.without(suffix: '.txt') == 'file');
+  /// ```
+  String without({String prefix = '', String suffix = ''}) {
+    var val = this;
+    if (startsWith(prefix)) {
```
isoos, line 315:
```suggestion
    if (prefix.isNotEmpty && startsWith(prefix)) {
```

app/lib/shared/utils.dart
```diff
@@ -299,4 +299,25 @@ extension StringExt on String {
...
+      val = val.substring(prefix.length);
+    }
+    if (endsWith(suffix)) {
```
isoos, line 318:
```suggestion
    if (suffix.isNotEmpty && endsWith(suffix)) {
```

app/lib/package/api_export/exported_api.dart
```diff
@@ -0,0 +1,327 @@
...
+    }));
+  }
+
+  /// Copy binary blob from [absoluteObjectName] to this file.
+  ///
+  /// Notice that [absoluteObjectName] must be an a GCS URI including `gs://`.
+  /// This means that it must include bucket name.
+  /// Such URIs can be created with [Bucket.absoluteObjectName].
+  Future<void> copyFrom(String absoluteObjectName) async {
+    await Future.wait(_owner._prefixes.map((prefix) async {
```
isoos, line 317: Is it possible that we have a race condition here which may update the second object earlier? (Is that bad?) -- actually this applies for most of the similar Future.wait calls in this file...

isoos (approved):

sigurdm (commented):
Can we find a better name for this concept? Perhaps `baked_api_responses`

PR #8105, opened 2024-10-04
ReservedPackage + admin action.

### LINKED ISSUE #8082: Admin action for reserving package names
Currently we need a new deployment for reserving a name.

Would be great to have a dynamic system for this. A datastore entry and an action

isoos:
- Fixes #8082

### DIFF
--- a/app/lib/admin/actions/package_reserve.dart
+++ b/app/lib/admin/actions/package_reserve.dart
@@ -12,11 +12,14 @@ final packageReserve = AdminAction(
   name: 'package-reserve',
   summary: 'Creates a ReservedPackage entity.',
   description: '''
-Reserves a package name that can be claimed by @google.com accounts or
-the allowed list of email addresses.
+Reserves a package name that can be claimed by the specified list of email addresses.
 
-The action can be re-run with the same package name. In such cases the provided
-email list will be added to the existing ReservedPackage entity.
+The action can be re-run with the same package name. In such cases the previous
+email list will be discarded, and the specified email list will be updated to the
+existing ReservedPackage entity.
+
+When no emails are specified, the package will be reserved, but no user may be
+able to claim it.
 ''',
   options: {
     'package': 'The package name to be reserved.',
@@ -43,9 +46,7 @@ email list will be added to the existing ReservedPackage entity.
       final existing = await tx.lookupOrNull<ReservedPackage>(
           dbService.emptyKey.append(ReservedPackage, id: package));
       final entry = existing ?? ReservedPackage.init(package);
-      if (emails != null) {
-        entry.emails = <String>{...entry.emails, ...emails}.toList();
-      }
+      entry.emails = <String>{...?emails}.toList();
       tx.insert(entry);
       return entry;
     });
--- a/app/lib/admin/actions/package_reserve.dart
+++ b/app/lib/admin/actions/package_reserve.dart
@@ -1,0 +1,61 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:pub_dev/package/backend.dart';
+import 'package:pub_dev/package/models.dart';
+import 'package:pub_dev/shared/datastore.dart';
+
+import 'actions.dart';
+
+final packageReserve = AdminAction(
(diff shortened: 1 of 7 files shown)


app/lib/admin/actions/package_reserve.dart
```diff
@@ -0,0 +1,61 @@
...
+  description: '''
+Reserves a package name that can be claimed by @google.com accounts or
+the allowed list of email addresses.
+
+The action can be re-run with the same package name. In such cases the provided
+email list will be added to the existing ReservedPackage entity.
+''',
+  options: {
+    'package': 'The package name to be reserved.',
+    'emails': 'The list of email addresses, separated by comma.'
```
sigurdm, line 23: Should we add an option to delete the entity? Maybe it is not worth it...

sigurdm (approved):

app/lib/admin/actions/package_reserve.dart
```diff
@@ -0,0 +1,61 @@
...
+    if (mp != null) {
+      throw InvalidInputException('ModeratedPackage `$package` exists.');
+    }
+
+    final entry = await withRetryTransaction(dbService, (tx) async {
+      final existing = await tx.lookupOrNull<ReservedPackage>(
+          dbService.emptyKey.append(ReservedPackage, id: package));
+      final entry = existing ?? ReservedPackage.init(package);
+      if (emails != null) {
+        entry.emails = <String>{...entry.emails, ...emails}.toList();
```
jonasfj, line 47:
```suggestion
        entry.emails = <String>{...emails}.toList();
```
I think it's safer to overwrite than append.

app/lib/admin/actions/package_reserve.dart
jonasfj: We could just add:
 * `package-reserve-list`
 * `package-reserve-delete`

Life is easier, if we split them into multiple actions. We can still put them all in the same file if want that.

We could even consider naming them: 
 `package-reservation-create`
 `package-reservation-list`
 `package-reservation-delete`

naming is a nit, but it's probably more intuitive to have separate options. And it's probably easier to read the code too.

isoos: Like a `--delete` flag?

app/lib/package/backend.dart
```diff
@@ -1014,10 +1022,19 @@ class PackageBackend {
...
-    final isReservedName = matchesReservedPackageName(name);
-    final isExempted = isGoogleComUser && isReservedName;
+    final reservedPackage = await lookupReservedPackage(name);
+    final reservedEmails = reservedPackage?.emails ?? const <String>[];
+
+    bool isAllowedUser = false;
+    if (agent is AuthenticatedUser) {
+      final email = agent.user.email;
+      isAllowedUser = email != null &&
+          (email.endsWith('@google.com') || reservedEmails.contains(email));
```
jonasfj, line 1032: I suggest that now we can allow list specific emails, we don't do this general thing for `@google.com` emails.

It's much better to be specific, this is generally rare.

app/lib/admin/actions/package_reserve.dart
sigurdm: Yeah - probably `--delete=true` as we didn't implement real flags for actions,

jonasfj: a separate command is MUCH easier to do.

jonasfj: it's really just something like:
```dart
final package = options['package'];
InvalidInputException.check(
  package != null && package.isNotEmpty,
  '`package` must be given',
);

await withRetryTransaction(dbService, (tx) async {
  final existing = await tx.lookupOrNull<ReservedPackage>(
      dbService.emptyKey.append(ReservedPackage, id: package));
      
  if (exiting == null) {
    return {'message': 'ReservedPackage for "$package" does not exist!'};
  }
  
  tx.delete(existing.key);
});

return {'message': 'ReservedPackage for "$package" deleted.'};
```

A `--delete=true` flag requires a lot more explanation in the documentation, it's not very intuitive.

sigurdm:
Yeah - multiple commands is probably better.

isoos:
Updated, PTAL.

isoos:
I will add list/delete in a follow-up PR.

jonasfj (approved):

sigurdm (approved):

PR #8117, opened 2024-10-07
Download counts: cache 30-days totals in backend

szakarias:


app/lib/service/entrypoint/frontend.dart
```diff
@@ -52,6 +53,7 @@ Future _main() async {
   await announcementBackend.start();
   await topPackages.start();
   await youtubeBackend.start();
+  await downloadCountsBackend.start();
```
jonasfj, line 56: @isoos is this this the right place to start it?

unrelated: Should we investigate if these could be started concurrently?

jonasfj (approved):

app/lib/service/entrypoint/frontend.dart
isoos: I think it is a good place to start it, unless we only want to start it in the `default` and not in the others.

app/lib/service/download_counts/backend.dart
```diff
@@ -19,7 +27,59 @@ DownloadCountsBackend get downloadCountsBackend =>
...
+        maxAge: Duration(days: 14),
+        interval: Duration(minutes: 30),
+        updateFn: _updateThirtyDaysTotals);
+  }
+
+  Future<Map<String, int>> _updateThirtyDaysTotals() async {
+    try {
+      final info = await storageService
+          .bucket(activeConfiguration.reportsBucketName!)
+          .info(downloadCounts30DaysTotalsFileName);
```
isoos, line 45: IIRC we have a tryInfo in the `lib/shared/storage.dart` which could at least handle the 404 part of the exceptions.

isoos (approved):

PR #8119, opened 2024-10-07
Fix completition widget: body placement + style.

isoos:


pkg/web_app/lib/src/widget/completion/widget.dart
```diff
@@ -87,7 +87,7 @@ void create(HTMLElement element, Map<String, String> options) {
       data: data,
     );
     // Add dropdown after the <input>
```
jonasfj, line 89:
```suggestion
    // Append dropdown to body
```

jonasfj (approved):

sigurdm (approved):

PR #8120, opened 2024-10-07
Fix CSS variable name + test to prevent failures.

### LINKED ISSUE #8106: site-header-nav backround color not displaying in mobile size
### Issue Description

There is a problem with the background color in the `site-header-nav` for mobile size. The background color is not being applied correctly due to a missing `--` in the CSS variable.

#### Steps to Reproduce
- Open the website on a mobile device or resize the browser window to mobile size.
- Observe the background color of the `site-header-nav`.

#### Expected Behavior
- The background color should be applied correctly using the CSS variable `--pub-site_header_popup-background-color`.

#### Actual Behavior
- The background color is not applied correctly due to the missing `--` in the CSS variable.

#### Fix
- Add the missing `--` to the CSS variable in the `background` property.

#### Screenshot
image

#### Example Code
```scss
.site-header-nav {
  @media (max-width: $device-mobile-max-width) {
    background: var(--pub-site_header_popup-background-color);
  }
}

### PR DESCRIPTION
**Pull request by isoos**
Fixes #8106

### DIFF
--- a/pkg/web_css/test/variables_test.dart
+++ b/pkg/web_css/test/variables_test.dart
@@ -75,5 +75,28 @@ void main() {
 
       expect(unused, isEmpty);
     });
+
+    test('all variables used have definition', () async {
+      final files = await Directory('lib')
+          .list(recursive: true)
+          .where((f) => f is File && f.path.endsWith('.scss'))
+          .cast<File>()
+          .toList();
+      final varRegExp = RegExp(r'var\((.*?)\)');
+      for (final file in files) {
+        final content = await file.readAsString();
+        for (final m in varRegExp.allMatches(content)) {
+          final name = m.group(1)!.trim();
+          if (!variables.contains(name)) {
+            // exempt Material Design variables
+            if (name.startsWith('--mdc-')) {
+              continue;
+            }
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/web_css/test/variables_test.dart [hunk 1]
```diff
@@ -75,5 +75,28 @@ void main() {
 
       expect(unused, isEmpty);
     });
+
+    test('all variables used have definition', () async {
```
jonasfj, line 79: Are we sure we want this?

Having an unused variable is pretty harmless is it not?

-----

If we want lints for our SASS files why not pick a dedicated tool for the purpose:
https://stylelint.io/

We can run it with extremely limited set of lints on github actions, or enable it in vscode and install stylelint in our development setups.

If we really want to do this, then I suggest we make the error more clearly document what is wrong and where.
Ideally, use something like `source_span` so that we can print nice messages.

jonasfj (approved):
No strong feelings, just a slight concern about the churn we add here.

But I can also see more churn by adding more linters/tools that everyone will have to configure/install.

isoos:
> Having an unused variable is pretty harmless is it not?

I'd consider it the same as unused code/variable lints in the Dart code. Since we filter out those, we should filter out the same here.

PR #8126, opened 2024-10-09
Experimental: Show download counts instead of popularity score

szakarias:


pkg/_pub_shared/lib/format/number_format.dart
```diff
@@ -6,15 +6,24 @@
...
 /// rounded down, in order to prevent exaggerated counts.
 String formatWithSuffix(int value) {
+  final f = computeValueWithSuffix(value);
+  return '${_toFixed(f.value)}${f.suffix}';
+}
+
+({num value, String suffix}) computeValueWithSuffix(int value) {
   if (value >= 1000000) {
-    return '${_toFixed(value, 1000000)}m';
+    return (value: _singleDigitDivision(value, 1000000), suffix: 'm');
```
jonasfj, line 15: Upper case `M` to avoid confusion with milli?

jonasfj (approved):

pkg/_pub_shared/lib/format/number_format.dart
szakarias: done

PR #8129, opened 2024-10-10
More retry on pub_worker: increased delay factor + more exceptions.

isoos:
After reviewing the error messages and stack traces, I think the bulk of the issue originates from either not retrying on some exceptions, or giving up too soon. This PR should address most of these, both for uploading the result or reporting them back to the pub.dev server.

### DIFF
--- a/pkg/pub_worker/lib/src/analyze.dart
+++ b/pkg/pub_worker/lib/src/analyze.dart
@@ -32,6 +32,9 @@ const _panaTimeout = Duration(minutes: 50);
 
 List<int> encodeJson(Object json) => JsonUtf8Encoder().convert(json);
 
+/// Retry requests with a longer delay between them.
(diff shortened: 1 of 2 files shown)


pkg/pub_worker/lib/src/analyze.dart
```diff
@@ -205,6 +205,7 @@ Future<void> _analyzePackage(
     final r = await retry(
       () => api.taskUploadResult(package, version),
       retryIf: _retryIf,
+      delayFactor: Duration(seconds: 5),
```
sigurdm, line 208: Nit: perhaps use a shared [RetryOptions](https://pub.dev/documentation/retry/latest/retry/RetryOptions-class.html)

sigurdm (approved):

PR #8144, opened 2024-10-13
Resolve json_serializable warning in `advisories_api.dart`

parlough:
Resolves the following warning on rebuilding the generated file, by relying on specifying the default value in the constructor.

```
[WARNING] json_serializable on lib/data/advisories_api.dart:
The constructor parameter for `aliases` has a default value `const <String>[]`, but the `JsonKey.defaultValue` value `[]` will be used for missing or `null` values in JSON decoding.
```


jonasfj (approved):

jonasfj:
I'd be surprised if this has any impact.

It's only if we create these objects from JSON and then try to append to `aliases`.

PR #8145, opened 2024-10-13
[web_app] Minor cleanup and consistency improvements

parlough:
To help a migration to `package:web` to be a bit more straightforward and focused.


pkg/web_app/lib/src/screenshot_carousel.dart
```diff
@@ -162,7 +165,7 @@ void _setEventForScreenshot() {
     }
 
     if (event.key == 'Escape') {
-      event.stopPropagation;
+      event.stopPropagation();
```
jonasfj, line 168: I'm guessing this is a bug fix?

jonasfj (approved):
I think this is fine,

PR #8146, opened 2024-10-13
Replace usages of `@sealed` annotation

parlough:


isoos (approved):

kevmoo:
@parlough – you rock!

jonasfj (approved):

jonasfj:
There is probably like a million places we should add `final class` :see_no_evil:

PR #8148, opened 2024-10-14
Minor cleanup to search in-memory index

parlough:
Primarily moves some declarations to be `static`, to reduce the large amount of instance members that end up in scope for those methods. This will make the route to the instance fields no longer being `late` a bit easier.

Also avoids some unnecessary non-null checks and assertions.


app/lib/search/mem_index.dart
```diff
@@ -19,15 +20,15 @@ final _logger = Logger('search.mem_index');
 final _textSearchTimeout = Duration(milliseconds: 500);
 
 class InMemoryPackageIndex {
-  final Map<String, PackageDocument> _packages = <String, PackageDocument>{};
-  final _packageNameIndex = PackageNameIndex();
+  final Map<String, PackageDocument> _packages = {};
+  final PackageNameIndex _packageNameIndex = PackageNameIndex();
```
isoos, line 24: Not sure: why do we need to specify the type here?

app/lib/search/mem_index.dart
```diff
@@ -19,15 +20,15 @@ final _logger = Logger('search.mem_index');
...
   final TokenIndex _descrIndex = TokenIndex();
   final TokenIndex _readmeIndex = TokenIndex();
   final TokenIndex _apiSymbolIndex = TokenIndex();
 
   /// Adjusted score takes the overall score and transforms
   /// it linearly into the [0.4-1.0] range.
-  final _adjustedOverallScores = <String, double>{};
+  final Map<String, double> _adjustedOverallScores = {};
```
isoos, line 31: Same here: type inference seems to be enough, why the extra specification?

app/lib/search/mem_index.dart
```diff
@@ -49,36 +50,34 @@ class InMemoryPackageIndex {
     }
     _updateOverallScores();
     _lastUpdated = clock.now().toUtc();
-    _overallOrderedHits = _rankWithComparator(_compareOverall,
+    _overallOrderedHits = _rankWithComparator(_packages, _compareOverall,
```
isoos, line 53: Not sure: is there any benefit of extracting these methods outside of the class? It is not that we will reuse them in other places.

app/lib/search/mem_index.dart
parlough: The inference is enough, but this change allows it to be consistent with the other nearby field entries.

Since these are fields and can be used outside of the direct vicinity, I generally think it makes sense to specify (and as a result, think about) the type.

app/lib/search/mem_index.dart
parlough: This was more of a transitionary change to a world where a lot of the mutable state doesn't have to escape construction. But that proved too large, so I thought this initial change would still be an improvement for now while getting a bit closer.

I do think making it `static` helps make it clear (and enforces that) it's not impacting any state of the class.

app/lib/search/mem_index.dart
isoos: Oh, now I understand the intent here. The other nearby field entries are older code that preceded the inference capabilities in the language, so if we wanted to make them consistent, I'd rather move them towards using the more modern features. wdyt?

PR #8161, opened 2024-10-18
Account for sass import deprecation

parlough:


isoos (approved):

isoos:
Note: before merging this should be pushed to staging first, let me know if this is considered ready.

parlough:
@isoos Yeah that would be appreciated. More eyes to make sure there are no regressions would be helpful :)

isoos:
Tested this locally and it seems mostly okay, except for e.g. the markdown code block with highlight js in dark mode:
image

It is not apparent to me what part is breaking... Maybe we should split the changes in the PR?

isoos (approved):
Fixed the rendering issue by moving the embedded styles into their own `_third_party.scss` file.

Issue #8157, opened 2024-10-17
Restrict appeals to one appeal per `ModerationCase`


jonasfj:
This comes along with a policy that we only allow one appeal per issue.

jonasfj:
For simplicity, I think we should simply say one appeal for each `(userId, caseId)`.

Because a case may be appealed by both the reporter and the content-owner.

PR #8168, opened 2024-10-21
One appeal per incident policy

jonasfj:
Formalizing new policy that we only allow one appeal per incident.

This is also enforced by technical measure, a user can only submit an appeal once.


sigurdm (approved):

jonasfj:
In support of https://github.com/dart-lang/pub-dev/issues/8157

PR #8170, opened 2024-10-21
Notice that we allow at-most one appeal

jonasfj:
Or maybe we should write something else.

### DIFF
--- a/app/lib/frontend/templates/report.dart
+++ b/app/lib/frontend/templates/report.dart
@@ -208,6 +208,9 @@ Iterable<d.Node> _appeal(
           '(source repositories, screenshots, logs, bug tracker entries, etc) '
           'consider uploading these and sharing a link.',
     ),
+    d.p(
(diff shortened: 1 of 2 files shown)


app/lib/frontend/templates/report.dart
```diff
@@ -208,6 +208,9 @@ Iterable<d.Node> _appeal(
           '(source repositories, screenshots, logs, bug tracker entries, etc) '
           'consider uploading these and sharing a link.',
     ),
+    d.p(
+      child: d.b(text: 'You only have one appeal per incident.'),
```
sigurdm, line 212:
```suggestion
      child: d.b(text: 'You can appeal at most once per case.'),
```

szakarias (approved):
LGTM, with Sigurd's comment.

jonasfj:
In support of https://github.com/dart-lang/pub-dev/issues/8157

sigurdm (approved):

PR #8172, opened 2024-10-21
Add tooltip to 30 days downloads label

szakarias:


app/lib/frontend/templates/views/pkg/labeled_scores.dart
```diff
@@ -30,6 +30,7 @@ d.Node labeledScoresNode({
       ),
       requestContext.experimentalFlags.showDownloadCounts
           ? d.div(
+              attributes: {'title': '30 days total download counts'},
```
sigurdm, line 33: How about:
```suggestion
              attributes: {'title': 'Number of downloads of this package during the past 30 days'},
```

sigurdm (approved):

Issue #7268, opened 2023-12-08
Extract pubspec.yaml from the package archive instead of the API endpoint before analysis SDK selection.


isoos:
Original issue reported: #7267

The root cause was a stale cache: upload happened on a different `runtimeVersion` than the analysis instance, and post-upload cache purge does not happen on other runtimes. The analysis started with reading the stale-cached API endpoint, failing to access the version. The proper long-term solution is to update the analysis and not use the cached endpoint.

isoos:
I think the issue was now fixed, and the potential further step (as implemented in #8862) is not really needed. Closing as completed.

PR #8173, opened 2024-10-21
Download archive in pub_worker before the analysis begins (and work on the downloaded directory).

isoos:
#7268


sigurdm (approved):

isoos:
Note: it seems that the `inspectDir` method will fail if the downloaded archive does not have a git repository. We'll need to update pana first to make this work, switching this to draft PR for the time being.

jonasfj (approved):

PR #8174, opened 2024-10-21
SDK selection: do not accept 3.5 Dart SDK for known macros incompatibility.

isoos:


sigurdm (commented):
Do we have a link to somewhere describing this issue?

isoos:
> Do we have a link to somewhere describing this issue?

We have packages that depend on `package:macros`, but do not update their SDK constraint for it. Similar example is `dartdoc`, and the current plan is to just wait it out: https://github.com/dart-lang/dartdoc/issues/3887

Because their SDK would allow the current stable, we are trying and failing analysis on that. I think it is better to give a temporary fix so that these packages do get analysis.

sigurdm (approved):

PR #8179, opened 2024-10-23
Use download counts in search index and sort control.

isoos:


isoos:
Note: I've pushed this to staging, but the index may not have been updated (and/or it is possible that the service is not started).

jonasfj (approved):

isoos:
Note: staging seems to have the required data already, so I'll just merge this.

PR #8180, opened 2024-10-23
More tests for ExportedApi

jonasfj:
* Minor refactoring so we easily write more tests
 * Additional tests for GC'ing
   * Files in old runtimeVersions, and,
   * stray files in old runtimeVersions.
 
I don't think we want to GC stray files in current or latest runtimeVersions.
Because:
 * There shouldn't be any.
 * If there are any, then it's because we're migrating incorrectly.
   * In `latest/` it's probably best to ignore stray files
   * In `$runtimeVersion/` stray files are probably a bug, but maybe it's best to ignore them. Eventually the runtimeVersion will be GC'ed.

I guess that stray files in `latest/` could become a problem over time.

### DIFF
--- a/app/test/package/api_export/exported_api_test.dart
+++ b/app/test/package/api_export/exported_api_test.dart
@@ -17,29 +17,22 @@ import 'package:test/test.dart';
 import '../../shared/test_services.dart';
 
 void main() {
-  testWithFakeTime('ExportedApi', (fakeTime) async {
+  final retryPkgData1 = PackageData(
+    name: 'retry',
+    latest: VersionInfo(
+      version: '1.2.3',
+      retracted: false,
+      pubspec: {},
+      archiveUrl: '-',
+      archiveSha256: '-',
+      published: clock.now(),
+    ),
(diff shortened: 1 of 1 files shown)


isoos:
> I guess that stray files in latest/ could become a problem over time.

I think we should change the logic to update the files (or their metadata) even if they are the same as existing ones, every month or so. If that update process is reliable, we may GC the files that were not updated after 6 months or so...

isoos (approved):

jonasfj:
> I think we should change the logic to update the files (or their metadata) even if they are the same as existing ones

Yeah, we could use meta-data to do this.

We could do a `x-goog-meta-updated` property that:
 * We update if it's more than 72 hours old.
 * We never set it to more than `max(Package.update, clock.ago(hours: 6))`
And GC will:
 * Delete files where `x-goog-meta-updated` is more than 14 days old.

Then we'll eventually cleanup stray files.

------------------

But we could not detect stray files, just files that aren't being updated :see_no_evil:

PR #8191, opened 2024-10-28
Added "updated" metadata to the ExportedApi

jonasfj:


app/lib/package/api_export/exported_api.dart
```diff
@@ -259,21 +259,24 @@ final class ExportedJsonFile<T> extends ExportedObject {
...
-  late final _metadata = ObjectMetadata(
-    contentType: 'application/json; charset="utf-8"',
-    contentEncoding: 'gzip',
-    cacheControl: 'public, max-age=${_maxAge.inSeconds}',
-  );
-
   /// Write [data] as gzipped JSON in UTF-8 format.
   Future<void> write(T data) async {
     final gzipped = _jsonGzip.encode(data);
+    final metadata = ObjectMetadata(
```
isoos, line 265: nit: can't we use the same method here as in `_metadata()` below?

isoos (approved):

PR #8192, opened 2024-10-28
Download counts: add computations for weekly downloads

szakarias:


app/lib/service/download_counts/computations.dart
```diff
@@ -42,3 +42,29 @@ Future<void> upload30DaysTotal(Map<String, int> counts) async {
...
+      await downloadCountsBackend.lookupDownloadCountData(package);
+  if (countData == null) {
+    return (weeklyPoints: <int>[], newestDate: null);
+  }
+
+  final totals = countData!.totalCounts;
+
+  var sum = 0;
+  for (int i = 0; i < 52 * 7; i++) {
+    if (totals[i] < 0) {
```
sigurdm, line 59: Just to be clear: this drops all days from the last partial week?

sigurdm (approved):

app/lib/service/download_counts/computations.dart
```diff
@@ -42,3 +42,29 @@ Future<void> upload30DaysTotal(Map<String, int> counts) async {
   await uploadBytesWithRetry(reportsBucket, downloadCounts30DaysTotalsFileName,
       jsonUtf8Encoder.convert(counts));
 }
+
+Future<({List<int> weeklyPoints, DateTime? newestDate})> computeWeeklyDownloads(
```
jonasfj, line 46:
```suggestion
/// Compute [weeklyDownloads] starting from [newestDate].
///
/// Each number in [weeklyDownloads] is the number of downloads for
/// a given 7 day period. Starting from [newestDate],
Future<({List<int> weeklyDownloads, DateTime? newestDate})> computeWeeklyDownloads(
```

nit: I'd consider avoid using `Points` to avoid confusing with pub points.

app/lib/service/download_counts/computations.dart
```diff
@@ -42,3 +42,29 @@ Future<void> upload30DaysTotal(Map<String, int> counts) async {
...
+  final countData =
+      await downloadCountsBackend.lookupDownloadCountData(package);
+  if (countData == null) {
+    return (weeklyPoints: <int>[], newestDate: null);
+  }
+
+  final totals = countData.totalCounts;
+
+  var sum = 0;
+  for (int i = 0; i < 52 * 7; i++) {
```
jonasfj, line 58: This is probably fine, but math would be easier to read if it was:
```suggestion
  for (int w = 0; w < 52; w++) {
    var sum = 0;
    for (int d = 0; d < 7; d++) {
      sum += totals[w * 7 + d];
    }
    weeklyPoints[w] = sum;
  }
```
Or maybe I'm missing something that it's supposed to do.

Or if one wants to be fancy :rofl: 
```suggestion
  weeklyPoints = List.generate(52, (w) => totals.skip(w * 7).take(7).sum);
```

jonasfj (approved):

app/lib/service/download_counts/computations.dart
jonasfj: will totals ever be short?

app/lib/service/download_counts/computations.dart
jonasfj: `List.generate(52, (w) => totals.skip(w * 7).take(7).where((c) => c > 0).sum);`

app/lib/service/download_counts/computations.dart
szakarias: Yes, but we can add it if you think.

app/lib/service/download_counts/computations.dart
szakarias: Updated comment.

app/lib/service/download_counts/computations.dart
szakarias: The nested loop is nice, the fancy I think is not the best. :-D

PR #8193, opened 2024-10-28
Introduce `ExportedPackage.garbageCollect(allVersionNumbers)`.

jonasfj:
* Garbage collection of version numbers, specifically archives.
 * Use `tryDelete` instead of `delete` to avoid errors from racing deletion, when GCS then returns 404.
 * Introduce `_minGarbageAge` to avoid deleting files before they are at-least a day old. This way GC won't race against creation of new files.

### DIFF
--- a/app/lib/package/api_export/exported_api.dart
+++ b/app/lib/package/api_export/exported_api.dart
@@ -20,6 +20,13 @@ import '../../shared/versions.dart'
 
 final _log = Logger('api_export:exported_bucket');
 
+/// Minimum age before an item can be consider garbage.
+///
+/// This ensures that we don't delete files we've just created.
+/// It's entirely possible that one process is writing files, while another
+/// process is running garbage collection.
+const _minGarbageAge = Duration(days: 1);
+
 /// Interface for [Bucket] containing exported API that is served directly from
 /// Google Cloud Storage.
 ///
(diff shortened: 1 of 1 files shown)


app/lib/package/api_export/exported_api.dart
```diff
@@ -94,8 +101,8 @@ final class ExportedApi {
       }
       if (!allPackageNames.contains(packageName)) {
         final info = await _bucket.info(item.name);
-        if (info.updated.isBefore(clock.ago(days: 1))) {
-          // Only delete if the item is more than one day old
+        if (info.updated.isBefore(clock.agoBy(_minGarbageAge))) {
+          // Only delete if the item if it's older than _minGarbageAge
```
isoos, line 105:
```suggestion
          // Only delete the item if it's older than _minGarbageAge
```

app/lib/package/api_export/exported_api.dart
```diff
@@ -111,8 +118,8 @@ final class ExportedApi {
       final packageName = item.name.without(suffix: '-').split('/').last;
       if (!allPackageNames.contains(packageName)) {
         final info = await _bucket.info(item.name);
-        if (info.updated.isBefore(clock.ago(days: 1))) {
-          // Only delete if the item is more than one day old
+        if (info.updated.isBefore(clock.agoBy(_minGarbageAge))) {
+          // Only delete if the item if it's older than _minGarbageAge
```
isoos, line 122:
```suggestion
          // Only delete the item if it's older than _minGarbageAge
```

isoos (approved):

PR #8194, opened 2024-10-28
Refactoring more methods into PackageStorage.

isoos:
Continuing #8183.

### DIFF
--- a/app/lib/package/package_storage.dart
+++ b/app/lib/package/package_storage.dart
@@ -2,12 +2,15 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:typed_data';
+
 import 'package:gcloud/storage.dart';
 import 'package:logging/logging.dart';
-import 'package:pub_dev/package/backend.dart';
-import 'package:pub_dev/package/models.dart';
-import 'package:pub_dev/shared/datastore.dart';
-import 'package:pub_dev/shared/storage.dart';
+import '../shared/datastore.dart';
+import '../shared/storage.dart';
+import '../shared/utils.dart';
+import 'backend.dart';
+import 'models.dart';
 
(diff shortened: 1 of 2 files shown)


app/lib/package/package_storage.dart
```diff
@@ -53,6 +56,71 @@ class PackageStorage {
...
+    String package,
+    String version,
+    Uint8List bytes,
+  ) async {
+    final objectName = tarballObjectName(package, version);
+    final info = await _canonicalBucket.tryInfo(objectName);
+    if (info == null) {
+      return ContentMatchStatus.missing;
+    }
+    // TODO: implement quick md5 match that doesn't require to download full content
```
sigurdm, line 76: This sounds important. Perhaps open an issue and attach the issue number here.

sigurdm (approved):
I think the title of the PR is wrong - there seems to be going more on than just moving code...

isoos:
> I think the title of the PR is wrong - there seems to be going more on than just moving code...

The logic of the method did not change, only a return value was introduced, because it has three result states, which was handled inlined previously. I can separate it out in a new PR if you prefer that.

sigurdm:
You can also just update the title - "moving" code to me means you can more or less find each deleted line added somewhere else.

app/lib/package/package_storage.dart
isoos: I think this is more of a shortcut on repeated upload checks than a critical part, because it would only help the rejection path, which may be rare. Instead, I think we should move all of these checks into a separate isolate and not do it in the frontend-serving one as part of the direct request path. I'll file an issue for that and add this to it.

isoos: I've updated the code to do the md5 hash check already with this refactor. PTAL.

sigurdm: LGTM.

PR #8196, opened 2024-10-29
Do not retry (or panic in logs) on upload exceptions when the version is no longer scheduled for analysis.

isoos:


app/lib/task/backend.dart
```diff
@@ -686,13 +686,14 @@ class TaskBackend {
...
-        throw NotFoundException.resource('$package/$version');
+        throw NotAcceptableException(
+            '$package/$version is no longer selected for analysis.');
       }
       final versionState = state.versions![version]!;
 
       // Check the secret token
       if (!versionState.isAuthorized(_extractBearerToken(request))) {
-        throw AuthenticationException.authenticationRequired();
+        throw NotAcceptableException('Secret token is no longer accepted.');
```
jonasfj, line 696: Let's use a 401 `AuthenticationException`, but give it a custom message?

app/lib/task/backend.dart
```diff
@@ -686,13 +686,14 @@ class TaskBackend {
       final key = PackageState.createKey(_db, runtimeVersion, package);
       final state = await tx.lookupOrNull<PackageState>(key);
       if (state == null || state.versions![version] == null) {
-        throw NotFoundException.resource('$package/$version');
+        throw NotAcceptableException(
```
jonasfj, line 689: 406 is probably not the right status code to use here.

We can use 404, because if the resources we're accessing (as indicated by URL) doesn't exist anymore.
We could also choose a generic 400 error.

I'm not sure that 409 is accurate because it's indicates that the client could do something to resolve the issue.

------------

I'm actually pretty sure the correct answer here is 404. Because someone could making a POST request to `/api/tasks/$package/$version/finished` with a `package` and `version` that simply don't exist.

We could add a custom error message to `NotFoundException`. But if `package` and `version` combination don't exist, then this is probably the correct message.

It's probably a better option to check datastore to see if the given package and version exists.
If they do, but they are not scheduled for analysis in `PackageState` we could return `400` instead. But then I'd suggest we make a custom `400` with its own `code`:
```dart
/// Thrown when results are being uploaded for a package/version where
/// analysis is not pending.
class AnalysisNotPendingException extends ResponseException {
  AnalysisNotPendingException._(String message)
      : super._(
          400,
          'AnalysisNotPending',
          message,
        );
```

Then we can parse the JSON and take out the `code` property.

------------

But I'm also inclined to suggest that what we could do is:
 * Handle 404 in `pub_worker` and just silence it.
   (this seems a bit risky, we could hiding future issues)
 * Investigate if we can avoid removing `PackageState` entries that are actively pending analysis (where an instance currently running).
   * This is probably difficult
   * We'd have to rely on timestamps to know if an instance is likely alive.
   * We'd need to review `PackageState` other methods, etc. as we are possibly changing invariants.
  (This might not be something we'll want to do, just considering the shear complexity)
 * Handle `404` in `pub_worker` and fetch version-listing from pub.dev and review the `published` date-time on the `latest` entry.
   * If there is a newer version not in the list of versions we've been told to analyze, this could be an indicator of a 404 error or 401 error not being important.
   * We could also just say that if `published` is recent, then `404` and `401` are reasonable errors to get?

app/lib/task/backend.dart
```diff
@@ -686,13 +686,14 @@ class TaskBackend {
...
-        throw NotFoundException.resource('$package/$version');
+        throw TaskAbortedException(
+            '$package/$version is no longer selected for analysis.');
       }
       final versionState = state.versions![version]!;
 
       // Check the secret token
       if (!versionState.isAuthorized(_extractBearerToken(request))) {
-        throw AuthenticationException.authenticationRequired();
+        throw TaskAbortedException('Secret token is no longer accepted.');
```
jonasfj, line 696: How do we know that this isn't because the secret is malformed?

isoos: Does it matter to the worker if it is aborted because the secret was malformed?
I think it is only the server that may be interested in malformed tokens, as it could report them as potential attack attempt... or is there any other use case for it?

jonasfj: Well,

 * On the **server side**, a _malformed token_ is an indicator of:
    * (A.1) Unknown internal bug (producing garbled payload, etc),
    * (B.1) Worker reporting back a canceled job, or,
    * (C) Being under attack (in which case we wish to reveal as little as possible, to an attacker).
 * On the **worker side**, a _malformed token_ is an indicator of:
    * (A.2) Unknown internal bug (receiving garbled payload, decoding payload incorrectly, etc), or,
    * (B.2) Worker reporting back a canceled job.

(A) is something we shout into a log, otherwise we'll never find out.
(B) is harmless, should be raised as an info or warning in logs.
(C) we should return `401`.

If (A) is a silent error, then why would we even bother looking at worker logs?

isoos: Let's spell out the cases:

If it is missing the bearer token, or it has an empty one, then it is a `401`.

If the server has no token to compare with, then it is likely that the task was aborted and we could return `400`+`TaskAborted`. But it may have been an attack attempt to inject something for a non-analyzed version. I think we can't genuinely decide here, but erring for the task being aborted seems to be okay.

After both is present, IIRC we don't have internal structure for these tokens, and do only string comparison. If it is non-matching, we could probably return `403` and shout in the logs...

Is these all what we are looking for here?

jonasfj: I'm not sure :see_no_evil: 

Perhaps we should consider introducing a `runId` or similar thing, I'm not sure.

isoos: Updated with tracking the aborted tokens. PTAL!

app/lib/task/backend.dart
```diff
@@ -1169,6 +1161,41 @@ String? _extractBearerToken(shelf.Request request) {
   return parts.last.trim();
 }
 
+PackageVersionStateInfo _extractAndVerifyVersionState(
+  String package,
+  String version,
+  PackageState? state,
+  shelf.Request request,
+) {
```
jonasfj, line 1169: It's not clear what this method does.

Indeed if it's responsibility is to do authorization, that seems like something very important to highlight.

Perhaps rename it:
```suggestion
/// Authorize a worker callback for [package] / [version].
///
/// Returns the [PackageVersionStateInfo] that the worker is authenticated for.
/// Or throw [ResponseException] if authorization is not possible.
PackageVersionStateInfo _authorizeWorkerCallback(
  shelf.Request request,
  PackageState state,
  String package,
  String version,
) {
```

Avoid making `PackageState` nullable, if it's missing, that's not an authorization issue.

app/lib/task/backend.dart
```diff
@@ -1169,6 +1161,41 @@ String? _extractBearerToken(shelf.Request request) {
...
+  final token = _extractBearerToken(request);
+  if (token == null) {
+    throw AuthenticationException.authenticationRequired();
+  }
+  if (state == null) {
+    throw NotFoundException.resource('$package/$version');
+  }
```
jonasfj, line 1176:
```suggestion
```

Not a concern for this method, if it does authorization, that should be it's primary thing.

app/lib/task/backend.dart
```diff
@@ -1169,6 +1161,41 @@ String? _extractBearerToken(shelf.Request request) {
...
+      throw TaskAbortedException('$package/$version has been aborted.');
+    }
+    // otherwise throw a generic not found error
+    throw NotFoundException.resource('$package/$version');
+  }
+
+  // Check the secret token
+  if (!versionState.isAuthorized(token)) {
+    throw AuthenticationException.authenticationRequired();
+  }
```
jonasfj, line 1192: This is vulnerable to timing attacks!

I'd suggest adding an `isAuthorized` method to `AbortedTokenInfo` object.

Also `versionState != null` doesn't imply that the token couldn't be aborted.

I'd suggest:
 * Always check `abortedTokens`, ideally always compare all of them (not just the first, because we want fixed time behavior)
 * Then check `versionState.isAuthorized`

app/lib/task/backend.dart
```diff
@@ -436,9 +437,19 @@ class TaskBackend {
...
+          deselectedVersions.add(e.key);
+
+          final token = e.value.secretToken;
+          if (token != null) {
+            state.abortedTokens ??= <AbortedTokenInfo>[];
+            state.abortedTokens!.add(AbortedTokenInfo(
+                token: token, expires: clock.fromNow(days: 1)));
+          }
+        }
+      }
```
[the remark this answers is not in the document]
isoos: Updated (using `Map.entries` for now)

app/lib/task/backend.dart
isoos: Done.

app/lib/task/backend.dart
```diff
@@ -454,6 +455,20 @@ class TaskBackend {
...
 
+      state.abortedTokens = [
+        ...state.versions!.entries
+            .where((e) => deselectedVersions.contains(e.key))
+            .map((e) => e.value)
+            .where((vs) => vs.secretToken != null)
+            .map(
+              (vs) => AbortedTokenInfo(
+                token: vs.secretToken!,
+                expires: vs.scheduled.add(maxTaskExecutionTime),
```
jonasfj, line 466: We should have stored `scheduled` instead of `expires`, not that it matters more, but it would have been 10% more canonical.

Anyways, I don't think it matters :D

jonasfj (approved):

app/lib/task/backend.dart
isoos: We could replace it, but I'm not sure it is worth it...

PR #8200, opened 2024-10-29
Stray file detection

jonasfj:
Use the `validated` meta-data to detect stray files.

### DIFF
--- a/app/lib/package/api_export/exported_api.dart
+++ b/app/lib/package/api_export/exported_api.dart
@@ -76,10 +76,18 @@ final class ExportedApi {
   /// This will remove prefixes other than `latest/` where [shouldGCVersion]
   /// returns true.
   Future<void> garbageCollect(Set<String> allPackageNames) async {
+    _log.info(
+      'Garbage collection started, with ${allPackageNames.length} package names',
+    );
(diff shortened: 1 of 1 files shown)


app/lib/package/api_export/exported_api.dart
```diff
@@ -175,6 +183,40 @@ final class ExportedApi {
...
+  ///
+  /// If there are stray files we don't really dare to delete them. They could
+  /// be introduced by a newer [runtimeVersion]. Or it could bug, but if that's
+  /// the case, what are the implications of deleting such files?
+  /// In all such cases, it's best alert and leave deletion of files as bug to
+  /// be fixed.
+  Future<void> _findStrayFiles(String prefix) async {
+    final validatedDeadline = clock.agoBy(_unvalidatedStrayFileAfter);
+    await _listBucket(prefix: prefix, delimiter: '', (item) async {
+      assert(item.isObject);
```
isoos, line 202: Not sure about this assert, maybe throwing an error is better? Otherwise we could be missing some strange thing in prod that we have no test for...

isoos (approved):
lgtm overall, but I'd change the assert to an exception

PR #8201, opened 2024-10-29
Introduce a `Stream.parallelForEach` for bounded parallelism.

jonasfj:
From https://github.com/dart-lang/async/pull/249 which we never managed to land in `dart:async`.

-------

```dart
Future<void> Stream<T>.parallelForEach(
    int maxParallel,
    FutureOr<void> Function(T item) each, {
    FutureOr<void> Function(Object e, StackTrace? st) onError = Future.error,
  })
```

Call `each` for each item in this stream with `maxParallel` invocations.

This method will invoke `each` for each item in this stream, and wait for
all futures from `each` to be resolved. `parallelForEach` will call `each`
in parallel, but never more then `maxParallel`.

If `each` throws and `onError` rethrows (default behavior), then
`parallelForEach` will wait for ongoing `each` invocations to finish,
before throw the first error.

If `onError` does not throw, then iteration will not be interrupted and
errors from `each` will be ignored.

```dart
// Count size of all files in the current folder
var folderSize = 0;
// Use parallelForEach to read at-most 5 files at the same time.
await Directory.current.list().parallelForEach(5, (item) async {
  if (item is File) {
    final bytes = await item.readAsBytes();
    folderSize += bytes.length;
  }
});
print('Folder size: $folderSize');

### DIFF
--- a/app/lib/shared/parallel_foreach.dart
+++ b/app/lib/shared/parallel_foreach.dart
@@ -1,0 +1,156 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// TODO: This library is a decent proposal for addition to `dart:async` or
+//       other similar utility package. It's extremely useful when processing
+//       a stream of objects, where I/O is required for each object.
+
+import 'dart:async';
+
+/// A [Notifier] allows micro-tasks to [wait] for other micro-tasks to
+/// [notify].
+///
+/// [Notifier] is a concurrency primitive that allows one micro-task to
+/// wait for notification from another micro-task. The [Future] return from
+/// [wait] will be completed the next time [notify] is called.
+///
+/// ```dart
+/// var weather = 'rain';
+/// final notifier = Notifier();
+///
+/// // Create a micro task to fetch the weather
+/// scheduleMicrotask(() async {
+///   // Infinitely loop that just keeps the weather up-to-date
+///   while (true) {
+///     weather = await getWeather();
+///     notifier.notify();
+///
+///     // Sleep 5s before updating the weather again
+///     await Future.delayed(Duration(seconds: 5));
+///   }
+/// });
+///
+/// // Wait for sunny weather
+/// while (weather != 'sunny') {
+///   await notifier.wait;
+/// }
+/// ```
+final class Notifier {
+  var _completer = Completer<void>();
+
+  /// Notify everybody waiting for notification.
+  ///
+  /// This will complete all futures previously returned by [wait].
+  /// Calls to [wait] after this call, will not be resolved, until the
+  /// next time [notify] is called.
+  void notify() {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: app/lib/shared/parallel_foreach.dart [hunk 1]
```diff
@@ -0,0 +1,156 @@
...
+  /// // Use parallelForEach to read at-most 5 files at the same time.
+  /// await Directory.current.list().parallelForEach(5, (item) async {
+  ///   if (item is File) {
+  ///     final bytes = await item.readAsBytes();
+  ///     folderSize += bytes.length;
+  ///   }
+  /// });
+  /// print('Folder size: $folderSize');
+  /// ```
+  Future<void> parallelForEach(
```
isoos, line 96: Shouldn't we call this `concurrentForEach` instead? Parallel would indicate separate isolates for me.

app/lib/shared/parallel_foreach.dart
```diff
@@ -0,0 +1,156 @@
...
+            }
+          } finally {
+            // When [each] is done, we decrement [running] and notify
+            running -= 1;
+            itemDone.notify();
+          }
+        }());
+
+        if (running >= maxParallel) {
+          await itemDone.wait;
```
isoos, line 136: I'm wondering: will this implicitly pause the stream listener until the pending processing is done? Can we leave a not on that, or a TODO to investigate it at one point?

isoos (approved):
lgtm, but I'd use the `concurrentForEach` name instead

app/lib/shared/parallel_foreach.dart
jonasfj: > I'm wondering: will this implicitly pause the stream listener until the pending processing is done?

Yes, that's kind of the point:
```
await stream.parallelForEach(N, each);
```
will:
 * Call `each` for each item in the stream.
 * Ensure that there are no more than `N` concurrent invocations of `each`.
   * This means pausing the stream, while items are being processed.
 * Wait for all `each` invocations to be completed.
   * Unless, `each` throws, in which case the stream will be canceled, and but pending `each` invocations will be awaited.

I would think that pausing the stream is desirable. Supposed you have a `Stream<Package> query` resulting from a datastore query.
 * You don't really want to load all `Package` entities into memory, if you did, you could do: `await query.toList();`.
 * You also don't really want to process each `Package` entity one at the time, because that's sort of slow. If you did could just do `await for (final p in query) {...}`.
 * You don't really want to process all `Package` entities concurrently because it'll do a crazy amount of I/O, won't reuse TCP connections and cause pain. If you did you could do `await Future.wait(await query.map((p) async {...}).toList());`.

You really want to process `N` number of entities concurrently and not more.

app/lib/shared/parallel_foreach.dart
jonasfj: I recall discussing this on the PR along ago. I don't really like `concurrentForEach`.

Maybe `asyncForEach` or `boundedAsyncForEach` or something completely different.

I think I'm going to keep the current name, just because it's not conflicting with other extensions on `Stream`.

PR #8202, opened 2024-10-29
Make generation of `ListSecurityAdvisoryResponse` reusable.

jonasfj:
This should allow us to use the same logic as we use in the response handler when the `ApiExporter` wants to generate responses for `/advisories`.

### DIFF
--- a/app/lib/service/security_advisories/backend.dart
+++ b/app/lib/service/security_advisories/backend.dart
@@ -5,7 +5,8 @@
 import 'dart:async';
 import 'dart:convert';
 
-import 'package:_pub_shared/data/advisories_api.dart' show OSV;
+import 'package:_pub_shared/data/advisories_api.dart'
+    show ListAdvisoriesResponse, OSV;
 import 'package:basics/basics.dart';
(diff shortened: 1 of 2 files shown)


app/lib/service/security_advisories/backend.dart
```diff
@@ -50,6 +51,16 @@ class SecurityAdvisoryBackend {
     }))!;
   }
 
+  /// Create a [ListAdvisoriesResponse] for [package] using advisories from
+  /// cache.
+  Future<ListAdvisoriesResponse> listAdvisoriesResponse(String package) async {
+    final advisories = await lookupSecurityAdvisories(package);
+    return ListAdvisoriesResponse(
+      advisories: advisories.map((e) => e.advisory).toList(),
```
isoos, line 59:
```suggestion
      advisories: advisories.map((a) => a.advisory).toList(),
```

isoos: Nit change to make it similar to the line below.

isoos (approved):

PR #8206, opened 2024-10-30
Add encoding functionality for int list to string

szakarias:


pkg/_pub_shared/lib/format/encoding.dart
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:typed_data';
+
+/// Base64 encodes [dataPoints] as 32-bit unsigned integers serialized with
+/// big endian byte order.
```
isoos, line 9:
```suggestion
/// little endian byte order.
```

pkg/_pub_shared/lib/format/encoding.dart
```diff
@@ -0,0 +1,28 @@
...
+String encodeIntsAsBigEndianBase64String(List<int> dataPoints) {
```
isoos, line 10: nit: rename here and the other `...BigEndian` -> `...LittleEndian`.

isoos (approved):
with renames

pkg/_pub_shared/lib/format/encoding.dart
szakarias: done

pkg/_pub_shared/lib/format/encoding.dart
szakarias: done

isoos (approved):

pkg/_pub_shared/lib/format/encoding.dart
```diff
@@ -0,0 +1,28 @@
...
+import 'dart:typed_data';
+
+/// Base64 encodes [dataPoints] as 32-bit unsigned integers serialized with
+/// little endian byte order.
+String encodeIntsAsLittleEndianBase64String(List<int> dataPoints) {
+  final byteData = ByteData(4 * dataPoints.length);
+  for (int i = 0; i < dataPoints.length; i++) {
+    byteData.setUint32(4 * i, dataPoints[i], Endian.little);
+  }
+  return base64Encode(byteData.buffer.asUint8List());
```
jonasfj, line 15: You can just use the Uint32List constructor. Give it a length and use `setAll`

jonasfj: @szakarias I'd suggest:
```dart
  return base64.encode(Uint8List.sublistView(Uint32List.fromList(dataPoints)));
```

Or

```dart
  return base64.encode(Uint32List.fromList(dataPoints).buffer.asUint8List());
```

But I'd advise against getting used to using `.buffer.asUint8List()`. Because when you have a `Uint32List` the underlying buffer could be longer than the contents of the actual `Uint32List`.

So technically, you'd have to read the documentation for the `Uint32List` constructor to see that it says:
> The list is backed by a [ByteBuffer](https://api.dart.dev/stable/3.5.4/dart-typed_data/ByteBuffer-class.html) containing precisely `length` times 4 bytes.

Relying on this promise it is safe to use `.buffer.asUint8List()`.

If we had gotten the `Uint32List` from somewhere else, we'd have to read [`offsetInBytes`](https://api.dart.dev/stable/3.5.4/dart-typed_data/TypedData/offsetInBytes.html) and [`lengthInBytes`](https://api.dart.dev/stable/3.5.4/dart-typed_data/TypedData/lengthInBytes.html) and pass them when doing `.buffer.asUint8List()`.

Hence, why I'd generally causing against accessing `.buffer` directly, because unless you know where the buffer was created, the location of your data within the buffer is not guaranteed. But `Uint8List.sublistView` will avoid such concerns.

PR #8207, opened 2024-10-30
Rename PackageStorage -> TarballStorage.

isoos:


jonasfj (approved):

app/lib/package/tarball_storage.dart
```diff
@@ -16,13 +17,14 @@ import 'models.dart';
...
 /// The GCS object name of a tarball object - excluding leading '/'.
+@visibleForTesting
 String tarballObjectName(String package, String version) =>
     '${_tarballObjectNamePackagePrefix(package)}$version.tar.gz';
 
 /// The GCS object prefix of a tarball object for a given [package] - excluding leading '/'.
 String _tarballObjectNamePackagePrefix(String package) => 'packages/$package-';
 
-class PackageStorage {
+class TarballStorage {
```
jonasfj, line 27: Is our thinking that this object will mainly be responsible for the `canonicalBucket`, right?

As eventually, the public bucket will go out of use, when `ApiExporter` starts updating `ExportedApi` and we've migrated the load balancer to use this.

Note. I don't mind that we let this object keep responsibility for the public bucket, until the public bucket is phased out.

jonasfj (approved):

app/lib/package/tarball_storage.dart
isoos: Yes, correct.

PR #8208, opened 2024-10-30
Add widget for weekly downloads chart with downloads data

szakarias:


app/lib/frontend/templates/views/pkg/info_box.dart
```diff
@@ -105,6 +109,23 @@ d.Node packageInfoBoxNode({
...
+        'data-sparkline-points':
+            _encodeForWeeklySparkline(wdc.weeklyDownloads, wdc.newestDate),
+      });
+  return container;
+}
+
+String _encodeForWeeklySparkline(List<int> downloads, DateTime newestDate) {
+  final date = newestDate.toUtc().millisecondsSinceEpoch ~/ 1000;
+  return encodeIntsAsLittleEndianBase64String([date, ...downloads]);
+}
```
jonasfj, line 127: You could just do:
```suggestion
String _encodeForWeeklySparkline(List<int> downloads, DateTime newestDate) {
  final date = newestDate.toUtc().millisecondsSinceEpoch ~/ 1000;
  final data = Uint32List.fromList([date, ...downloads]);
  return base64.encode(Uint8List.sublistView(data));
}
```
See: https://github.com/dart-lang/pub-dev/pull/8206/files#r1822801873

app/lib/frontend/templates/views/pkg/info_box.dart
```diff
@@ -105,6 +109,23 @@ d.Node packageInfoBoxNode({
   ]);
 }
 
+d.Node _downloadsChart(WeeklyDownloadCounts wdc) {
+  final container = d.div(
+      classes: ['weekly-downloads-sparkline'],
+      id: '-weekly-downloads-sparkline',
+      attributes: {
+        'data-widget': 'weekly-sparkline',
+        'data-sparkline-points':
```
jonasfj, line 118:
```suggestion
        'data-widget': 'weekly-sparkline',
        'data-weekly-sparkline-points':
```

app/lib/shared/redis_cache.dart
```diff
@@ -247,6 +247,16 @@ class CachePatterns {
...
       ))[package];
 
+  Entry<WeeklyDownloadCounts> weeklyDownloadCounts(String package) => _cache
+      .withPrefix('weekly-download-counts/')
+      .withTTL(Duration(hours: 6))
+      .withCodec(utf8)
+      .withCodec(json)
+      .withCodec(wrapAsCodec(
+        encode: (WeeklyDownloadCounts wdc) => wdc.toJson(),
+        decode: (d) => WeeklyDownloadCounts.fromJson(d as Map<String, dynamic>),
```
jonasfj, line 257: By default it's possible to store binary data in redis.

That's why we have `.withCodec(utf8)`.

So why not just store:
```dart
base64.encode(Uint8List.sublistView(Uint32List.fromList(dataPoints)));
```
Or:
```dart
return Uint8List.sublistView(Uint32List.fromList(dataPoints));
```

jonasfj (approved):
You'll name to name options for the widget as `data-<widget>-<option>`.

So if you have `data-widget="weekly-sparkline"` the `points` option should be: `data-weekly-sparkline-points="..."`.

app/lib/service/download_counts/download_counts.dart
```diff
@@ -215,3 +215,18 @@ class CountData {
       _$CountDataFromJson(json);
   Map<String, dynamic> toJson() => _$CountDataToJson(this);
 }
+
+@JsonSerializable(includeIfNull: false)
+class WeeklyDownloadCounts {
```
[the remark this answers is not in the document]
szakarias: Yes, this is indeed an optimization. But as you say it's not pretty. Also the other things we store in redis are human readable strings which is nice for debugging.

app/lib/frontend/templates/views/pkg/info_box.dart
szakarias: Great, I didn't know `Uint8List.sublistView()` can translate any given typed data object to bytes. I'll update this in a follow up.

Issue #8189, opened 2024-10-28
Tracking completion widget issues


isoos:
- [ ] fixing absolute positioning bug
- [ ] style change: blend with input field background
- [ ] style change: padding + rounded corners in the bottom
- [ ] style change: already typed text with different color
- [ ] display it in mini search widget (e.g. on package page)
- [ ] include the prefix in the result list
- [ ] `#` as a shortcut for `topic:` completion (also in query canonicalization)

PR #8209, opened 2024-10-30
Include matched prefix as part of the display options.

isoos:
#8189

deployed to staging

image


jonasfj (approved):

jonasfj:
For the record I still think it's more natural to not display the part that triggered completion. But I don't really mind it either way :rofl:

isoos:
> For the record I still think it's more natural to not display the part that triggered completion. But I don't really mind it either way 🤣

I'm kind of 50-50 on this, let's wait @sigurdm's vote on it.

PR #8217, opened 2024-11-01
Completion for `is:swiftpm-plugin` tag

sigurdm:
Re https://github.com/dart-lang/pana/issues/1402

### DIFF
--- a/app/lib/frontend/templates/views/shared/search_banner.dart
+++ b/app/lib/frontend/templates/views/shared/search_banner.dart
@@ -127,6 +127,7 @@ String completionDataJson({
             'plugin',
             'unlisted',
             'wasm-ready',
+            'swiftpm-plugin',
           ],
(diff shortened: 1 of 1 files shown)


app/lib/frontend/templates/views/shared/search_banner.dart
```diff
@@ -127,6 +127,7 @@ String completionDataJson({
             'plugin',
             'unlisted',
             'wasm-ready',
+            'swiftpm-plugin',
```
isoos, line 130: nit: let's keep alphabetical sort here?

isoos (approved):

app/lib/frontend/templates/views/shared/search_banner.dart
sigurdm: Done

Issue #8195, opened 2024-10-29
Move upload checks into a separate isolate.


isoos:
The current upload checks may have a large memory footprint, we should do it outside of the request-processing isolate:
- [ ] move pre-transaction archive verification into a separate isolate
- [ ] do length and md5 hash checks when verifying the newly uploaded archive against the canonical bucket

PR #8222, opened 2024-11-01
Reduced memory use on processing the new upload.

isoos:
- This is a the low-hanging fruit before having an implementation of separate isolates (#8195).
- Hash calculation is stream based, and content comparison depends on the chunk of bytes we get back from the storage bucket. In theory this could be still arbitrarily large, but in practice I think it won't be.


isoos:
Note: I think the separate isolate refactor will take a lot of work, as we need to setup a full service environment in that isolate to make it work through all of the interdependent code (especially bucket access). However, this change may deliver a small benefit before we find enough time to do the refactor with isolates.

sigurdm (approved):

PR #8224, opened 2024-11-04
Introduce synchronizeTarballs that uses Bucket.list which now yield objectinfo

jonasfj:


app/lib/package/api_export/exported_api.dart
```diff
@@ -270,6 +270,69 @@ final class ExportedPackage {
...
+  ) async {
+    await Future.wait([
+      ..._owner._prefixes.map((prefix) async {
+        final pfx = prefix + '/api/archives/$_package-';
+
+        final versionsForUpload = versions.keys.toSet();
+        await _owner._listBucket(prefix: pfx, delimiter: '', (entry) async {
+          final item = switch (entry) {
+            final BucketDirectoryEntry _ => throw AssertionError(
+                'directory entries should not be possible',
```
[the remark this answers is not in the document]
sigurdm: Perhaps make this a shout, like the "stray-file" case.

app/lib/package/api_export/exported_api.dart
```diff
@@ -450,49 +539,58 @@ final class ExportedBlob extends ExportedObject {
...
+        await _copyToPrefixFromIfNotContentEquals(
+          prefix,
+          source,
+          await _owner._bucket.tryInfo(dst),
         );
       });
     }));
   }
+
+  Future<void> _copyToPrefixFromIfNotContentEquals(
```
isoos, line 564: This method has a scary name and should have more documentation.
Maybe it would be worth it to separate it into two: checking if it needs to copy, and then copy it.

app/lib/package/api_export/exported_api.dart
```diff
@@ -450,49 +539,58 @@ final class ExportedBlob extends ExportedObject {
...
+    String prefix,
+    SourceObjectInfo source,
+    ObjectInfo? destinationInfo,
+  ) async {
+    final dst = prefix + _objectName;
+
+    // Check if the dst already exists
+    if (destinationInfo case final dstInfo?) {
```
isoos, line 572: Nit: I am not convinced that this is a better format as opposed to `if (destinationInfo != null) {`. I could accept it if `destinationInfo` was not a single variable, and instead we collapse a new variable creation and null check into one, but this is just strange.

jonasfj: Sold, this was carried over from when there was an inline `tryInfo` call.

app/lib/package/api_export/exported_api.dart
jonasfj: The name is intentionally scary, it's also an internal method because nobody outside ExportedApi (and friends) should use it.

But I'll add some documentation and throw `ArgumentError` if `destinationInfo.name != dst`.

app/lib/package/api_export/exported_api.dart
```diff
@@ -270,6 +270,67 @@ final class ExportedPackage {
         Duration(hours: 2),
       );
 
+  Future<void> synchronizeTarballs(
```
isoos, line 273: Let's document what the input parameter and with this method does, including the GC part.

app/lib/package/api_export/exported_api.dart
```diff
@@ -270,6 +270,67 @@ final class ExportedPackage {
...
+
+            // Done, we don't need to delete this item
+            return;
+          }
+
+          // Delete the item, if it's old enough.
+          if (item.updated.isBefore(clock.agoBy(_minGarbageAge))) {
+            // Only delete if the item if it's older than _minGarbageAge
+            // This avoids any races where we delete files we've just created
+            await _owner._bucket.tryDelete(item.name);
```
isoos, line 316: Nit: I think deleting while listing should not cause any issues, but I do wonder about it: should we be more defensible here by creating a list of objects to delete first? I am not sure, it may be unnecessary.

isoos (approved):
looks good to me, pending the documentation on the public method.

app/lib/package/api_export/exported_api.dart
jonasfj: There is racing regardless, so I don't think we should worry about it.

Afaik, objects are sorted alphabetically and for each page (when listing) the token used to get the next page, is basically a key from which the listing should start.

Request the first or middle page of a huge list shouldn't make any difference.

PR #8225, opened 2024-11-04
Reduce the memory allocation during search with mutable IndexedScore.

isoos:
- Follow-up to #8152.
- Fewer operation is required as document weight is part of the pre-calculation at index creation. Some scores in test did change, I think these changes are because of the changes of the generic method that is being used in search and they are fine.
- This change does not completely eliminate the Map-based immutable `Score`, but reduces its use (and its map-building operations) by extending the use of the list-based score across multiple token index and multiple words (fewer list allocations).
- Local search benchmark shows about 3-6% total query time improvements.
- Note: I'm also planning a follow-up PR that expands the concept on the filtering that happens before the text search part, possibly getting further benefits.

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -311,32 +311,28 @@ class InMemoryPackageIndex {
       // We cannot update the main `packages` variable yet, as the dartdoc API
       // symbols are added on top of the core results, and `packages` is used
       // there too.
-      final coreScores = <Score>[];
-      var wordScopedPackages = packages;
+      final coreScores = IndexedScore(_packageNameIndex._packageNames);
+      for (var i = 0; i < _documents.length; i++) {
+        if (packages.contains(_documents[i].package)) {
+          coreScores.setValue(i, 1.0);
+        }
+      }
+
       for (final word in words) {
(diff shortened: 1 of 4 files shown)


app/lib/search/mem_index.dart
```diff
@@ -501,49 +497,60 @@ class _TextResults {
...
+  /// Search using the parsed [word] and return the matching packages with scores
+  /// as a new [IndexedScore] instance.
+  ///
+  /// When [filterOnNonZeros] is present, only the indexes with an already
+  /// non-zero value are evaluated.
+  IndexedScore searchWord(
+    String word, {
+    IndexedScore? filterOnNonZeros,
+  }) {
+    final scoreore = IndexedScore(_packageNames);
```
sigurdm, line 539: What is a `scoreore`?

isoos: bad rename, fixed

isoos:
Note: after some offline discussion, I'll split this PR into multiple parts, leaving this as a draft for now.

isoos:
Closing, as all parts have been submitted for review or merged already.

PR #8227, opened 2024-11-05
Rewrite/redirect '#<topic>' search expressions to 'topic:<topic>'

isoos:
- This was mentioned while we have discussed #8189.
- We don't actively keep track of the topics, if we want to restrict the rewrite to existing topics, we should probably cache the topic names locally too.

### DIFF
--- a/app/test/search/backend_test.dart
+++ b/app/test/search/backend_test.dart
@@ -69,5 +69,12 @@ void main() {
         'topic:widget abc',
       );
     });
+
+    test('query with topic shortcut `#`', () {
+      expect(
+        canonicalizeSearchForm(_parse('#widget #testing'))?.query,
+        'topic:widget topic:testing',
+      );
+    });
   });
 }
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -19,16 +19,14 @@ import 'package:meta/meta.dart';
 // ignore: implementation_imports
 import 'package:pana/src/dartdoc/pub_dartdoc_data.dart';
 import 'package:pool/pool.dart';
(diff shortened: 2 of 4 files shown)


sigurdm (approved):
I don't really worry that the rewrite happens for non-existing topics. That is what I would expect.

My only question is, if the rewrite should go the other way (such that `#topic` is the canonical version) @szakarias any opinion?

app/test/search/backend_test.dart
```diff
@@ -69,5 +69,12 @@ void main() {
         'topic:widget abc',
       );
     });
+
+    test('query with topic shortcut `#`', () {
+      expect(
+        canonicalizeSearchForm(_parse('#widget #testing'))?.query,
+        'topic:widget topic:testing',
+      );
```
jonasfj, line 77: Can we have a test case where I search for `"#hash"` -- and would that work if I actually want to search for a keyword that starts with `#` ?

jonasfj (approved):

app/test/search/backend_test.dart
isoos: `#` is by default removed from the content at indexing and from queries (used as a separator).

The end result of this change is that `#hash` search from https://pub.dev/packages?q=%23hash gets redirected to https://pub.dev/packages?q=topic%3Ahash and yeah, we can add a test for that.

isoos:
> My only question is, if the rewrite should go the other way (such that #topic is the canonical version)

I'm 50:50 on this one. However, going to the other direction would require more change in the query parsing, and we can start with this one first, allowing the use of `#`, while still reverse the canonicalization at a later point.

PR #8233, opened 2024-11-06
Tool to import parts of dart-lang/site-shared into third_party/site-shared

isoos:


third_party/site-shared/update-site-shared.sh
```diff
@@ -0,0 +1,27 @@
...
+set -e
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+TEMP_DIR=$(mktemp -d)
+
+# Clone dartdoc repository
+git clone https://github.com/dart-lang/site-shared "$TEMP_DIR"
+
+# Delete resources/
```
sigurdm, line 11: Should we make this simpler, and just import everything?

sigurdm (approved):

third_party/site-shared/update-site-shared.sh
isoos: For the other vendored parts in `third_party/` we only keep the build artifacts. I don't see much benefit of keeping the tests or the pubspec.yaml here too, but I don't mind importing them either.

isoos: I think having a few extra lines here is better than dealing with additional files, let's keep it this way for now, until we need the full package instead.

jonasfj (approved):

PR #8238, opened 2024-11-07
Utilize BucketObjectEntry in GC logic

jonasfj:


app/lib/package/api_export/exported_api.dart
```diff
@@ -452,14 +449,14 @@ const _validatedCustomHeader = 'validated';
 ///
 /// This allows us to detect files that are present, but not being updated
 /// anymore. We classify such files as _stray files_ and write alerts to logs.
-const _updateValidatedAfter = Duration(days: 1);
+const _updateValidatedAfter = Duration(days: 3);
```
jonasfj, line 452: Idea being that we don't want to update meta-data all the time.

app/lib/package/api_export/exported_api.dart
```diff
@@ -98,40 +98,37 @@ final class ExportedApi {
   Future<void> _gcPrefix(String prefix, Set<String> allPackageNames) async {
```
jonasfj, line 98: Change here simplified a bit, just list all files, extract package name from the object name and if the package doesn't exist, we delete it if it's older than `_minGarbageAge`

app/lib/package/api_export/exported_api.dart
```diff
@@ -26,7 +26,7 @@ final _log = Logger('api_export:exported_bucket');
 /// This ensures that we don't delete files we've just created.
 /// It's entirely possible that one process is writing files, while another
 /// process is running garbage collection.
-const _minGarbageAge = Duration(days: 1);
+const _minGarbageAge = Duration(hours: 3);
```
jonasfj, line 29: This shouldn't be high, if something is published, it'll be at-least 3 hours before we can delete it.

I think that's perhaps somewhat reasonable. But we don't want this to be too high.

However, 3 hour should be more than enough for any eventual consistency issues between datastore and GCS to have been resolved.

app/lib/package/api_export/exported_api.dart
```diff
@@ -452,14 +449,14 @@ const _validatedCustomHeader = 'validated';
...
-const _updateValidatedAfter = Duration(days: 1);
+const _updateValidatedAfter = Duration(days: 3);
 
 /// Duration after which a file that haven't been updated is considered stray!
 ///
 /// We don't delete stray files, because there shouldn't be any, so a stray file
 /// is always indicative of a bug. Nevertheless, we write alerts to logs, so
 /// that these inconsistencies can be detected.
-const _unvalidatedStrayFileAfter = Duration(days: 7);
+const _unvalidatedStrayFileAfter = Duration(days: 12);
```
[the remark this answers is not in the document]
isoos: I'm wondering: will we look at the logs or have alerts that fire within this period?

isoos (approved):

app/lib/package/api_export/exported_api.dart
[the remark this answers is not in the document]
jonasfj: No, after 12 days we'll consider a file stray, and then we'll see it in logs.

PR #8243, opened 2024-11-07
Add metadata to fake_gcloud

jonasfj:


pkg/fake_gcloud/test/mem_storage_test.dart
```diff
@@ -93,7 +93,6 @@ void main() {
 
     final sb = StringBuffer();
     storage.writeTo(sb);
-    expect(sb.length, 118);
```
jonasfj, line 96: I think this test is dumb, we're checking that `MemStorage` can be serialized, we should assert what size it has.

Just that we get the same data out when we're deserialize it, which is tested further down.

szakarias (approved):

pkg/fake_gcloud/test/mem_storage_test.dart
isoos: Exact length check may be important if we are checking for a specific serialization format or making sure we are not doing excess or stupid things here. However, I do agree that this is not that important here.

isoos (approved):

PR #8257, opened 2024-11-08
Increase concurrency and handle errors better

jonasfj:


app/lib/package/api_export/api_exporter.dart
```diff
@@ -157,6 +161,12 @@ class ApiExporter {
     });
 
     await _api.garbageCollect(allPackageNames);
+
+    if (errCount > 0) {
+      throw Exception(
+        '$errCount exceptions happened while calling synchronizePackage',
```
sigurdm, line 167:
```suggestion
        '$errCount exceptions happened while calling synchronizeExportedApi',
```

sigurdm (approved):

isoos (approved):

PR #8278, opened 2024-11-12
Tests for admin action of API exporter sync.

isoos:
- tests cover object deletion, object override and metadata override of the exported bucket
- implemented `Bucket.updateMetadata` in `fake_gcloud`, hopefully close to what the real gcloud would do

### DIFF
--- a/app/test/admin/exported_api_sync_test.dart
+++ b/app/test/admin/exported_api_sync_test.dart
@@ -1,0 +1,194 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:_pub_shared/data/admin_api.dart';
+import 'package:gcloud/storage.dart';
+import 'package:pub_dev/shared/configuration.dart';
+import 'package:pub_dev/shared/storage.dart';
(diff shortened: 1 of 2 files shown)


app/test/admin/exported_api_sync_test.dart
```diff
@@ -0,0 +1,194 @@
...
+      await api.adminInvokeAction(
+        'exported-api-sync',
+        AdminInvokeActionArguments(arguments: {
+          'packages': packages?.join(' ') ?? 'ALL',
+          if (forceWrite) 'force-write': 'true',
+        }),
+      );
+    }
+
+    Future<Map<String, dynamic>> _list() async {
```
jonasfj, line 31:
```suggestion
    /// Return map of all objects in the Exported API bucket.
    ///
    /// Returns a map from object name to JSON on the form:
    /// ```js
    /// {
    ///   'updated': '<date-time>',
    ///   'metadata': {
    ///     'contentType': '<contentType>',
    ///     'custom': {...},
    ///   },
    ///   'length': <length>,
    ///   'bytes': [...],
    /// }
    /// ```
    Future<Map<String, dynamic>> listExportedApi() async {
```

This will make the tests a lot easier to read.

jonasfj (approved):

jonasfj (approved):

PR #8280, opened 2024-11-13
Use styles from site-shared/dash_design

isoos:
- Updated the variable tests to also use the dash_design files.
- I've commented out the two colors that this replaces for now. I'm going to remove them with a follow-up PR when we have more colors in the shared style sheet that we import.
- No effective color change for us.

### DIFF
--- a/pkg/web_css/lib/src/_variables.scss
+++ b/pkg/web_css/lib/src/_variables.scss
@@ -22,7 +22,6 @@
   --pub-color-snowWhite:     #fafafa; // slight deviation from official snow-white (fffafa)
   --pub-color-smokeWhite:    #f5f5f7; // slight deviation from official white-smoke (f5f5f5)
   --pub-color-bubblesBlue:   #e7f8ff; // slight deviation from official bubbles-blue (e7feff)
-  // --pub-color-gunpowderGray: #4a4a4a; // hsl(0, 0%, 29%);
 
   --pub-color-dangerRed: #ff4242;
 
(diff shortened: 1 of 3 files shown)


pkg/web_css/lib/src/_variables.scss
```diff
@@ -16,17 +16,19 @@
...
 
 :root {
+  @include dash_variables.light-theme;
+
   --pub-color-white:         #ffffff;
   --pub-color-snowWhite:     #fafafa; // slight deviation from official snow-white (fffafa)
   --pub-color-smokeWhite:    #f5f5f7; // slight deviation from official white-smoke (f5f5f5)
   --pub-color-bubblesBlue:   #e7f8ff; // slight deviation from official bubbles-blue (e7feff)
-  --pub-color-gunpowderGray: #4a4a4a; // hsl(0, 0%, 29%);
+  // --pub-color-gunpowderGray: #4a4a4a; // hsl(0, 0%, 29%);
```
sigurdm, line 25: Should this just be deleted?

sigurdm (approved):

pkg/web_css/lib/src/_variables.scss
```diff
@@ -114,15 +116,17 @@
...
 
 .dark-theme {
+  @include dash_variables.dark-theme;
+
   --pub-color-darkGunmetal:    #1f262a; // close to #1d2026
   --pub-color-shadowBlack:     #373737;
   --pub-color-anchorBlack:     #41424c;
   --pub-color-nipponUltraBlue: #23607f;
-  --pub-color-gainsboro:       #dcdcdc;
+  // --pub-color-gainsboro:       #dcdcdc;
```
sigurdm, line 125: And this..

pkg/web_css/lib/src/_variables.scss
isoos: I'd rather keep this for a short while, until we have a better consolidated view on how the naming + shared naming will work. But at the same time I can also just delete it.

sigurdm: As you prefer - just note that the usual policy is to not leave commented out code.

You can always find it in git history.

PR #8293, opened 2024-11-15
Change font of score key figures

szakarias:
Goes from this  
image
to this
image

Reducing the height difference between the numbers and letters.


sigurdm:
image
Looks like the 3 is higher than the other numbers... I guess that is a font style choice.

sigurdm (approved):

PR #8297, opened 2024-11-18
Reuse IndexedScore instances, as emptying the values is faster than discarding and re-allocating memory.

isoos:
Local benchmark showed about 5-6% difference on a continuous load.

### DIFF
--- a/app/lib/search/token_index.dart
+++ b/app/lib/search/token_index.dart
@@ -100,6 +100,7 @@ class TokenIndex<K> {
   /// scoring.
   IndexedScore<K> searchWords(List<String> words, {double weight = 1.0}) {
     IndexedScore<K>? score;
+
     weight = math.pow(weight, 1 / words.length).toDouble();
     for (final w in words) {
       final s = IndexedScore(_ids);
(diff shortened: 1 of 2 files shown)


app/lib/search/token_index.dart
```diff
@@ -142,6 +143,30 @@ extension StringTokenIndexExt on TokenIndex<String> {
...
+  }) {
+    late IndexedScore<K> score;
+    if (_pool.isNotEmpty) {
+      score = _pool.removeLast();
+      score._values.setAll(0, Iterable.generate(score.length, (_) => value));
+    } else {
+      score = IndexedScore<K>(_keys, value);
+    }
+    final r = fn(score);
+    _pool.add(score);
```
sigurdm, line 165: This logic means that we will never have more scores then the high-watermark of scores that we need at one time - right?

Thus we don't really risk leaking excessive memory... right?

sigurdm (approved):

app/lib/search/token_index.dart
isoos: That's correct. We also replace the index every 15 minutes, and if there is an extreme outlier period, its memory footprint won't be around after that.

PR #8300, opened 2024-11-18
Parse `#<topic>` expressions in SearchForm's parsed query text, and support it through the data serialization.

isoos:
- part of #8189, follow-up to #8227
- this does not reverse the topic expression canonicalization of #8227, but enables that option to be done later
- separated the new tests into two commits, they can be compared before-after the current change

### DIFF
--- a/pkg/_pub_shared/lib/search/search_form.dart
+++ b/pkg/_pub_shared/lib/search/search_form.dart
@@ -18,6 +18,7 @@ final _sortRegExp = RegExp('sort:([a-z]+)');
 final _updatedRegExp = RegExp('updated:([0-9][0-9a-z]*)');
 final _tagRegExp =
     RegExp(r'([\+|\-]?[a-z0-9]+:[a-z0-9\-_\.]+)', caseSensitive: false);
+final _topicRegExp = RegExp(r'([\+|\-])?#([a-z0-9-]{2,32})');
 
 /// The tag prefixes that we can detect in the user-provided search query.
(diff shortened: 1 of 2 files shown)


pkg/_pub_shared/lib/search/search_form.dart
```diff
@@ -18,6 +18,7 @@ final _sortRegExp = RegExp('sort:([a-z]+)');
 final _updatedRegExp = RegExp('updated:([0-9][0-9a-z]*)');
 final _tagRegExp =
     RegExp(r'([\+|\-]?[a-z0-9]+:[a-z0-9\-_\.]+)', caseSensitive: false);
+final _topicRegExp = RegExp(r'([\+|\-])?#([a-z0-9-]{2,32})');
```
sigurdm, line 21: Is '|' meant to be an option for the sign?
```
void main() {
  final _topicRegExp = RegExp(r'([\+|\-])?#([a-z0-9-]{2,32})');
  print(_topicRegExp.firstMatch(r'|#21')?[1] );
}
```
prints '|'

isoos: good catch, we have the same bug in `_tagRegExp`, fixing both.

isoos:
I'm closing this without merging, as it seems to be affecting too many internals, and they are better to be kept clean. I'd focus on doing this transformation only in the user-facing `SearchForm` in a separate PR.

Issue #8307, opened 2024-11-19
Home page cards need different background differentiator in dark mode


parlough:
In light mode, the current box shadows work, but in dark mode, they look out of place.

They could be replaced with borders in dark mode or a differently configured box shadow.

<img width="400" alt="Cards with odd box shadow in dark mode" src="https://github.com/user-attachments/assets/08056642-fd3c-4c7f-90d2-ed751c8105d6">

parlough:
The same applies for the "Sort by" dropdown menu: 
<img width="233" alt="Screenshot of sort by dropdown menu in dark mode" src="https://github.com/user-attachments/assets/f3fa6d28-dabe-4bfd-bfe1-14714d425c8d">

isoos:
I think both of these issues were fixed now.

PR #8319, opened 2024-11-19
Use border and toned-down drop-shadow for landing-page cards.

isoos:
#8307

image
image


isoos:
I've started to push this to staging for better testability.

sigurdm:
I think the general advice is to not use drop-shadows in dark mode at all.

See https://medium.com/@tundehercules/designing-effective-dark-mode-interfaces-17f38ecea2e9#:~:text=Shadows%20%2CElevation%20and%20Depth

isoos:
I was planning to keep the two themes conceptually the same, but the argument for not using shadows entirely is a good one. I'll review the other parts of the article and redo this afterwards.

PR #8322, opened 2024-11-20
Use different font size for h3 and h4 in markdown formatted content.

### LINKED ISSUE #8109: h3 and h4 headings have the same font size
* The font size for `h3` and `h4` headings are the same
 * Different heading sizes should have different font sizes (or at least have some other **distinguishing property** so that users can visually see a difference)
 * I also observed the same problem in multiple browsers
 * https://www.loom.com/share/fc9ab0f49dab4e5eb596a86b958bf599?sid=2aa47d04-c5d0-409c-97e9-eb6bc5cc0872

isoos:
Fixes #8109.

### DIFF
--- a/pkg/web_css/lib/src/_base.scss
+++ b/pkg/web_css/lib/src/_base.scss
@@ -46,12 +46,12 @@ body,
     font-weight: 400;
   }
 
-  h1 { font-size: 36px; }
-  h2 { font-size: 24px; }
-  h3 { font-size: 18px; }
-  h4 { font-size: 18px; }
-  h5 { font-size: 16px; }
-  h6 { font-size: 16px; }
+  h1 { font-size: 36px; } // note: github uses 2em, ~32px
+  h2 { font-size: 24px; } // note: github uses 1.5em, ~24px
+  h3 { font-size: 21px; } // note: github uses 1.25em, ~20px
+  h4 { font-size: 18px; } // note: github uses 1em, ~16px
+  h5 { font-size: 16px; } // note: github uses 0.875em, ~14px
+  h6 { font-size: 16px; } // note: github uses 0.85em, ~13.6px
 }
 
 img {
--- a/pkg/web_css/lib/src/_base.scss
+++ b/pkg/web_css/lib/src/_base.scss
@@ -50,7 +50,7 @@ body,
   h2 { font-size: 24px; } // note: github uses 1.5em, ~24px
   h3 { font-size: 21px; } // note: github uses 1.25em, ~20px
   h4 { font-size: 18px; } // note: github uses 1em, ~16px
-  h5 { font-size: 16px; } // note: github uses 0.875em, ~14px
+  h5 { font-size: 17px; } // note: github uses 0.875em, ~14px
   h6 { font-size: 16px; } // note: github uses 0.85em, ~13.6px
 }
 


parlough:
> /cc @parlough this looks like another item that should be easy to standardize on, what do you think?

Makes sense to me, perhaps besides `h1` which can be different depending on the content.

pkg/web_css/lib/src/_base.scss
```diff
@@ -46,12 +46,12 @@ body,
...
-  h3 { font-size: 18px; }
-  h4 { font-size: 18px; }
-  h5 { font-size: 16px; }
-  h6 { font-size: 16px; }
+  h1 { font-size: 36px; } // note: github uses 2em, ~32px
+  h2 { font-size: 24px; } // note: github uses 1.5em, ~24px
+  h3 { font-size: 21px; } // note: github uses 1.25em, ~20px
+  h4 { font-size: 18px; } // note: github uses 1em, ~16px
+  h5 { font-size: 16px; } // note: github uses 0.875em, ~14px
+  h6 { font-size: 16px; } // note: github uses 0.85em, ~13.6px
```
sigurdm, line 54: Should we differentiate h6 now we are here? (Not sure we ever use h6 thought)
```suggestion
  h6 { font-size: 16px; } // note: github uses 0.85em, ~13.6px
```

isoos: The idea to have a headline text that is smaller than the paragraph text (even though it may be bolder) is a bit strange to me, but apparently others dig up its origins: https://stackoverflow.com/questions/55696808/why-do-h5-and-h6-have-smaller-font-sizes-than-p-in-most-user-agent-default
https://ux.stackexchange.com/questions/126039/is-it-a-problem-if-h4-h5-and-h6-are-smaller-than-regular-text

With the later, I think I'd keep the 16px, as our current `<p>` is 16px.

pkg/web_css/lib/src/_base.scss
```diff
@@ -46,12 +46,12 @@ body,
...
-  h2 { font-size: 24px; }
-  h3 { font-size: 18px; }
-  h4 { font-size: 18px; }
-  h5 { font-size: 16px; }
-  h6 { font-size: 16px; }
+  h1 { font-size: 36px; } // note: github uses 2em, ~32px
+  h2 { font-size: 24px; } // note: github uses 1.5em, ~24px
+  h3 { font-size: 21px; } // note: github uses 1.25em, ~20px
+  h4 { font-size: 18px; } // note: github uses 1em, ~16px
+  h5 { font-size: 16px; } // note: github uses 0.875em, ~14px
```
sigurdm, line 53: We could then make h5 17px.

```suggestion
  h5 { font-size: 17px; } // note: github uses 0.875em, ~14px
```

isoos: Sounds good!

pkg/web_css/lib/src/_base.scss
```diff
@@ -46,12 +46,12 @@ body,
...
-  h3 { font-size: 18px; }
-  h4 { font-size: 18px; }
-  h5 { font-size: 16px; }
-  h6 { font-size: 16px; }
+  h1 { font-size: 36px; } // note: github uses 2em, ~32px
+  h2 { font-size: 24px; } // note: github uses 1.5em, ~24px
+  h3 { font-size: 21px; } // note: github uses 1.25em, ~20px
+  h4 { font-size: 18px; } // note: github uses 1em, ~16px
+  h5 { font-size: 17px; } // note: github uses 0.875em, ~14px
+  h6 { font-size: 16px; } // note: github uses 0.85em, ~13.6px
```
jonasfj, line 54: I suspect these are a bit close, but I'm not too worried.

From what I can tell wikipedia has the same size for h4 - h6

image

with lines under h1 and h2, and h3 only being slightly larger.

Just saying it's an option to not have different sizes.

jonasfj (approved):
LGTM, I don't have strong feelings here.

pkg/web_css/lib/src/_base.scss
jonasfj: I note that it might be more importnat to have a larger difference between h3 and h4, than having any difference between h4-h6

People rarely use h5 and h6.

If we wanted to make using h1-h6 usable, we should perhaps consider numbering headlines.

> # 1. h1
> ## 1.1 h2
> ### 1.1.1 h3
> #### 1.1.1.1 h4
> ##### 1.1.1.1.1 h5
> ###### 1.1.1.1.1.1 h6

But this can also be a bit degenerate in it's own right... Though we could put `1.1.1.1.1` in the margin and make it only show up when hovering a headline, or something like that.

isoos: I think a table of contents hovering/updating/folding/unfolding as you scroll may be better than numbers.

PR #8324, opened 2024-11-21
Fix code coverage collection for integration tests.

isoos:
With the `pub` tool being part of the executable, the process failed with:
```
Package "coverage" is not an immediate dependency.
Cannot run executables in transitive dependencies.
```


sigurdm (approved):
This might be a bug in `pub run`.

However that is deprecated, and `dart run` should be the way to go anyways.

PR #8328, opened 2024-11-21
Integration test to use pub get with the exported API bucket (+ minimal content rewrite)

isoos:


sigurdm (approved):

pkg/pub_integration/test/exported_bucket_test.dart
```diff
@@ -0,0 +1,138 @@
...
+          try {
+            // the proxy url with the appropriate path prefix
+            final proxyUri = storageUri.replace(pathSegments: [
+              ...storageUri.pathSegments,
+              ...rq.requestedUri.pathSegments
+            ]);
+
+            // archive files are served from the storage bucket directly
+            if (rq.requestedUri.path.endsWith('tar.gz')) {
+              return shelf.Response.seeOther(proxyUri);
```
jonasfj, line 44: this is not what GCLB does, that's what this proxy is emulating, right?

GCLB just serves directly, but yes, this is fine.

jonasfj (approved):

pkg/pub_integration/test/exported_bucket_test.dart
isoos: Good point, removing this.

PR #8331, opened 2024-11-23
Use and fix core lints across workspace

parlough:
Now that there is a shared workspace (amazing for pub-dev's setup!!!), the shared analysis config can use `package:lints`.

Start with just the core lints for easier review, and follow-up can eventually use the full recommended set.


sigurdm (approved):

pkg/fake_gcloud/pubspec.yaml
```diff
@@ -7,9 +7,11 @@ environment:
 resolution: workspace
 
 dependencies:
+  clock: ^1.1.0
+  crypto: ^3.0.0
   gcloud: ^0.8.18
   logging: '>=0.11.3 <2.0.0'
-  clock: ^1.1.0
+  _discoveryapis_commons:
```
sigurdm, line 14: Should this have a cosntraint

PR #8334, opened 2024-11-25
Use dartdoc 8.3.0 for packages that depend on macros.

isoos:
This is a temporary override to get `package:analyzer` documentation generated.


sigurdm:
Could you do something along the lines of `dart pub deps --json | jq '.packages.[]|select(.name=="_macros")'` to also catch transitive dependency on package _macros ?

isoos:
I think this + #8339 is a good baseline, pushing these to staging right now. We still need to implement the new redirect from `-library.html` to the directory-root, and also update some test assumptions, and we seem to provide a reasonable documentation page for these packages.

sigurdm (approved):

Issue #8336, opened 2024-11-25
Fix integrity check wrt. blocked packages


jonasfj:
Also we should consider migrating "blocked" to "moderatedAt" and cleanup old unused properties.

PR #8341, opened 2024-11-27
Reduce false reporting of isBlocked packages in integrity checks.

isoos:
#8336


sigurdm (approved):

sigurdm:
Is this reducing or eliminating the reporting?

isoos:
> Is this reducing or eliminating the reporting?

I'm not confident enough to claim that it will eliminate all of the reports we have seen this week, but it has a high chance that it will catch all of the issues being reported as false positives.

PR #8345, opened 2024-11-29
Implementing redirect on dartdoc-generated redirect file.

isoos:
- Follow-up to #8342, implementing the redirect instead of in-place rendering.
- Updated the document serving handler by adding a second cached value about the status of a page, whether is served, missing or redirected.
- Optimized the code for the optimistic happy path, trying to serve the cached bytes first (if any), then trying the cached status (if any), then calculating and updating both as the final fallback path.
- Added a test that should fail / should be updated when we migrate to 8.3 dartdoc.


isoos:
Note: this is available on staging and can be tried on the `analyzer` package.

sigurdm (approved):
As discussed, perhaps move the sanity check from the handler to the dartdoc post-processing (and consider logging if a violation is found).

PR #8348, opened 2024-12-02
First pass at a Table of Contents generator from markdown

Levi-Lesches:
This PR doesn't integrate with anything yet, just provides a new script in the `app` directory, `dart run :toc`.

This script generates a fake Pub page (couldn't get all the JS/CSS to load) based on [`flutter_local_notifications`](https://pub.dev/packages/flutter_local_notifications). Running the script generates a `app/toc_experiment/index.html` you can load in a localhost server. The generated page is the standard README plus a table of contents pinned to the side, with functioning links. This PR is not about the exact format of the page but rather about the parsing logic, all contained within `app/bin/toc.dart`.

To test, fill the content of `app/toc_experiment/readme.md` with whatever you like (now it has the notifications package), then: 

```bash
cd app
dart run :toc
cd toc_experiment
dart pub global activate dhttpd
dhttpd --port 8888
```
Navigate to `http://localhost:8888` in your browser, and there should be a functional table of contents!

image

---

<details>

</details>

### DIFF
--- a/app/pubspec.yaml
+++ b/app/pubspec.yaml
@@ -53,6 +53,7 @@ dependencies:
   ulid: '2.0.1'
   tar: '2.0.0'
   api_builder:
+  simple_mustache: ^2.1.0
 
 dev_dependencies:
   build_runner: '^2.0.0'
--- a/app/bin/md.dart
+++ b/app/bin/md.dart
@@ -1,0 +1,132 @@
+import 'dart:io' as io;
+
+import 'package:markdown/markdown.dart';
+import 'package:pub_dev/frontend/static_files.dart';
+import 'package:pub_dev/frontend/templates/views/shared/layout.dart';
+import 'package:simple_mustache/simple_mustache.dart';
+import 'package:pub_dev/frontend/dom/dom.dart' as dom;
+
+const _structuralHeaderTags = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'];
+final currentUri = Uri();
+const rootDir = './toc_experiment';
+
+String generateFragment(String text) => BlockSyntax.generateAnchorHash(
+  Element.text('h1', text),
+);
+
+/// A section of the Table of Contents
+class TocSection {
+  int level;
+  Node titleNode;
+  List<TocSection> children;
+  TocSection? parent;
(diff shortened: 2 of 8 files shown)


app/pubspec.yaml
```diff
@@ -53,6 +53,7 @@ dependencies:
   ulid: '2.0.1'
   tar: '2.0.0'
   api_builder:
+  simple_mustache: ^2.1.0
```
jonasfj, line 56: we can't pull in arbitrary dependencies.

jonasfj (commented):
Is TOC generation something we should consider adding to `package:markdown` directly?

isoos:
> Is TOC generation something we should consider adding to package:markdown directly?

Good point, I think besides the more users for the same feature, we could address the id generation (and customizations on conflicts for the same titles) better at the core library.

app/pubspec.yaml
Levi-Lesches: I'm not using this package for the parsing code, don't worry. 

In the previous thread you said not to focus on the UI elements of the parsing yet, so I didn't integrate this into the standard Pub package page, and instead make some quick-and-dirty HTML to show off the results. That's what this package is for and should not be used in the final site.

Levi-Lesches:
Makes sense to me -- should I open a PR there instead?

parlough:
> Makes sense to me -- should I open a PR there instead?

I think you should start with an issue there to discuss with its maintainers. I agree this tooling should be shared and accessible, but I'm not sure how much this makes sense in the core markdown package.

I think it might work best as a separate package and if the extension capabilities in `package:markdown` aren't enough, figuring out how to add them.

Levi-Lesches:
@isoos See https://github.com/dart-lang/markdown/issues/631

Levi-Lesches:
@isoos, thoughts as to the next steps here? https://github.com/dart-lang/tools/issues/1456 didn't get much discussion, but I can still open a PR on the markdown package if you feel that's the right direction.

sigurdm:
Closing this, we probably want a TOC at some point, but not exactly this way.

* We want it for dartdoc first
* we want to do it after html sanitization

PR #8356, opened 2024-12-04
Migrate isBlocked flags to isModerated.

isoos:
- #8336
- added the migration as part of the new-field pub-task, as there seems to be little reason to create a separate task for it (which will be removed and then reported as not working)
- added a test that does not use the admin action to set the flags - they can be also removed (in a separate PR)
- added the migration to be part of the post-test part of the profile-based tests, so that we can make sure we don't have any inconsistency after they run (to be extended to all periodic tasks in a follow-up PR)

### DIFF
--- a/app/lib/tool/backfill/backfill_new_fields.dart
+++ b/app/lib/tool/backfill/backfill_new_fields.dart
@@ -2,7 +2,14 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'package:clock/clock.dart';
 import 'package:logging/logging.dart';
+import 'package:pub_dev/account/models.dart';
+import 'package:pub_dev/package/api_export/api_exporter.dart';
+import 'package:pub_dev/package/backend.dart';
+import 'package:pub_dev/package/models.dart';
+import 'package:pub_dev/publisher/models.dart';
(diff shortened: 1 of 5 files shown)


app/lib/tool/backfill/backfill_new_fields.dart
```diff
@@ -12,5 +19,72 @@ final _logger = Logger('backfill_new_fields');
...
+
+    // sync exported API(s)
+    await apiExporter?.synchronizePackage(entity.name!, forceDelete: true);
+
+    // retract or re-populate public archive files
+    await packageBackend.tarballStorage.updatePublicArchiveBucket(
+      package: entity.name!,
+      ageCheckThreshold: Duration.zero,
+      deleteIfOlder: Duration.zero,
+    );
```
jonasfj, line 53: Are we sure we want to do this? Or should we just let this happen naturally?

I don't really have any strong feelings on the subject :rofl:

jonasfj (approved):

app/lib/tool/backfill/backfill_new_fields.dart
isoos: > Are we sure we want to do this? Or should we just let this happen naturally?

The tests would fail with integrity checks, otherwise we could let it happen naturally. I don't mind that it happens right away.

PR #8367, opened 2024-12-06
Hide downloads chart widget behind experimental flag

szakarias:


app/lib/frontend/templates/views/pkg/score_tab.dart
```diff
@@ -79,7 +79,8 @@ d.Node scoreTabNode({
           ],
         ),
         _reportNode(report),
-        if (card.weeklyVersionDownloads != null)
+        if (card.weeklyVersionDownloads != null &&
+            requestContext.experimentalFlags.showDownloadCounts)
```
sigurdm, line 83: Don't you want to make another experiment, so the sparkline can be enabled without the versions chart?

szakarias: No, I was planning on removing the flag from the sparkline and just using it for this, since this chart will not be out before the sparkline launches.

sigurdm (approved):

PR #8369, opened 2024-12-06
Randomize search isolate renewal delays.

isoos:


sigurdm (approved):

sigurdm:
Should we log something around takeover, such that we can correlate it in logs?

isoos:
Note: renewal logging added in a separate PR: #8372.

PR #8370, opened 2024-12-06
Keep isSoftRemoved calls in-iteration, prevent List updates.

isoos:
This is about 1-1.5% of the time when loading the search snapshot.

### DIFF
--- a/app/bin/tools/search_benchmark.dart
+++ b/app/bin/tools/search_benchmark.dart
@@ -19,9 +19,9 @@ Future<void> main(List<String> args) async {
       json.decode(utf8.decode(gzip.decode(await file.readAsBytes())))
           as Map<String, Object?>;
   final snapshot = SearchSnapshot.fromJson(content);
-  snapshot.documents!
-      .removeWhere((packageName, doc) => isSoftRemoved(packageName));
-  final index = InMemoryPackageIndex(documents: snapshot.documents!.values);
+  final index = InMemoryPackageIndex(
+      documents:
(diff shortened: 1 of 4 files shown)


sigurdm (commented):
Not sure I like this change - it seems to distribute complexity

isoos:
> it seems to distribute complexity

Not sure: why?

In two cases the change is local: it is replacing a `List.removeWhere()` into a `List.where` iteration.
In the snapshot building case we don't add the soft-removed package to the snapshot in the first place, so we don't need to remove them later.

sigurdm:
> Not sure: why?

It seems to me that it distributes the burden of attention from the creation of the snapshot to the usage of it. That feels backwards, and not worth it for a 1% improvement. 

Is there a way to construct the snapshot so it never contains the soft-removed packages in the first place?

isoos:
> Is there a way to construct the snapshot so it never contains the soft-removed packages in the first place?

The changes on line 135 in `app/lib/search/backend.dart` do exactly that.

sigurdm:
Ah - but why do you then need to account for them in app/lib/frontend/handlers/custom_api.dart line 82?

isoos:
> Ah! - but why do you then need to account for them in app/lib/frontend/handlers/custom_api.dart line 82?

That's a different use case with a different data source. We get the names from the name tracker, which has all the names from all the `Package` entities. Having a second look on it, I will move that logic inside the name-tracker, as it should be part of the `getVisiblePackageNames` method.

sigurdm (approved):

PR #8371, opened 2024-12-06
Reduce complexity of 'normalizeBeforeIndexing' method.

isoos:
- Shaves off couple of seconds locally on the index loading (~5-7%).
- The method first generated multiple spaces in place of non-[a-z0-9] characters, then coalesced those into a single space. With the updated regular expression it does the same in a single operation.

### DIFF
--- a/app/lib/search/text_utils.dart
+++ b/app/lib/search/text_utils.dart
@@ -11,7 +11,7 @@ import '../shared/markdown.dart';
 final _separatorChars = '_.?!,;:=()<>[]{}~@#\$%&|+-*\\/"\'`';
 final _escapedSeparators = _separatorChars.split('').map((s) => '\\$s').join();
 final _separators = RegExp('[$_escapedSeparators]|\\s');
-final RegExp _nonCharacterRegExp = RegExp('[^a-z0-9]');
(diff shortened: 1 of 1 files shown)


app/lib/search/text_utils.dart
```diff
@@ -58,12 +58,7 @@ String compactReadme(String? text) {
 
 String normalizeBeforeIndexing(String? text) {
   if (text == null) return '';
-  final String t = text
-      .toLowerCase()
-      .replaceAll(_nonCharacterRegExp, ' ')
-      .replaceAll(_multiWhitespaceRegExp, ' ')
-      .trim();
-  return t;
+  return text.toLowerCase().replaceAll(_nonCharactersRegExp, ' ').trim();
```
sigurdm, line 61: Don't you still need the multiWhitespace replacement?

isoos: As whitespace is also outside of `[a-z0-0]`, we already have it in the first replacement.

sigurdm: Ah - I think I get it now

sigurdm (approved):

PR #8376, opened 2024-12-09
Add a few init related lints from recommended set

parlough:
Part of migrating to `lints/recommended.yaml` from `lints/core.yaml`. Split up to enable easier review.

### DIFF
--- a/app/lib/search/token_index.dart
+++ b/app/lib/search/token_index.dart
@@ -40,8 +40,8 @@ class TokenIndex<K> {
     List<K> ids,
     List<String?> values, {
     bool skipDocumentWeight = false,
-  }) : _ids = ids {
-    assert(ids.length == values.length);
+  })  : _ids = ids,
+        assert(ids.length == values.length) {
     final length = values.length;
     for (var i = 0; i < length; i++) {
       final text = values[i];
--- a/app/lib/tool/neat_task/datastore_status_provider.dart
+++ b/app/lib/tool/neat_task/datastore_status_provider.dart
@@ -42,12 +42,13 @@ class NeatTaskStatus extends db.ExpandoModel<String> {
 
   NeatTaskStatus();
 
-  NeatTaskStatus.init(String name, {required bool isRuntimeVersioned}) {
-    this.name = name;
-    runtimeVersion =
-        _runtimeVersion(name, isRuntimeVersioned: isRuntimeVersioned);
+  NeatTaskStatus.init(String name, {required bool isRuntimeVersioned})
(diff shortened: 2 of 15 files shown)


app/lib/search/token_index.dart
```diff
@@ -40,8 +40,8 @@ class TokenIndex<K> {
     List<K> ids,
     List<String?> values, {
     bool skipDocumentWeight = false,
-  }) : _ids = ids {
-    assert(ids.length == values.length);
+  })  : _ids = ids,
+        assert(ids.length == values.length) {
```
isoos, line 44: I'm slightly biased towards having this in the constructor block. Is that against the lint?

app/lib/tool/neat_task/datastore_status_provider.dart
```diff
@@ -42,12 +42,13 @@ class NeatTaskStatus extends db.ExpandoModel<String> {
...
-    this.name = name;
-    runtimeVersion =
-        _runtimeVersion(name, isRuntimeVersioned: isRuntimeVersioned);
+  NeatTaskStatus.init(String name, {required bool isRuntimeVersioned})
+      // ignore: prefer_initializing_formals
+      : name = name,
+        runtimeVersion =
+            _runtimeVersion(name, isRuntimeVersioned: isRuntimeVersioned),
+        updated = clock.now().toUtc() {
     id = _compositeId(name, isRuntimeVersioned: isRuntimeVersioned);
```
isoos, line 51: Why is `id` left out?

pkg/_popularity/lib/popularity.dart
```diff
@@ -70,10 +70,9 @@ class VoteData {
 
   int get score => _score(this);
 
-  VoteData(this.direct, this.dev, this.total) {
-    assert(direct >= 0 && direct <= total);
-    assert(dev >= 0 && dev <= total);
-  }
+  VoteData(this.direct, this.dev, this.total)
+      : assert(direct >= 0 && direct <= total),
```
isoos, line 74: Same here. My reasoning: while asserts here are simple, we may put more complex checks in them, and initializing block may not be the best place for it.

app/lib/tool/neat_task/datastore_status_provider.dart
parlough: You can't initialize fields inherited from a parent class in the initializer list. Added a comment with more context.

isoos:
Thanks! I think this looks good, let's rebase on master and we may merge it.

isoos (approved):

isoos (approved):

PR #8377, opened 2024-12-09
Faster predicate matching with integer tag ids.

isoos:


sigurdm (approved):
LGTM

Did you measure the speedup?

isoos:
> Did you measure the speedup?

Yes, roughly half of the latency on the particular query we were looking at.

PR #8384, opened 2024-12-10
Add seperate flag for download counts chart

szakarias:


app/lib/frontend/handlers/experimental.dart
```diff
@@ -96,6 +96,7 @@ class ExperimentalFlags {
   bool get isDarkModeDefault => isEnabled('dark-as-default');
 
   bool get showDownloadCounts => isEnabled('download-counts');
+  bool get showDownloadCountsChart => isEnabled('download-counts-chart');
```
sigurdm, line 99: How about calling it `download-counts-version-chart`?

sigurdm (approved):

app/lib/frontend/handlers/experimental.dart
szakarias: I have no strong preference, so I'll change it :-)

PR #8387, opened 2024-12-10
Enable download counts

szakarias:


sigurdm (approved):
Do we need to change something, such that the example doesn't show up on pub.dev/experimental?

isoos:
> Do we need to change something, such that the experiment doesn't show up on pub.dev/experimental?

I think it is harmless if it shows up on it... 

However, the tests are failing because of an unused method in templates test.

PR #8396, opened 2024-12-11
Different tag index: document index lists for each tag.

isoos:
- This is a follow-up on #8377 to implement the idea where we keep the document index list for each tag.
- The change itself is mostly neutral on memory, and local benchmark shows about 2-3% improvements in latencies.
- However, it reduces the "smart" factor (complexity) of the index, and I think this has even further potential for improvement, if we change it to value ranges + pre-sort the document list for the default predicates. (to be explored in a separate PR).

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -28,7 +28,10 @@ class InMemoryPackageIndex {
   late final TokenIndex<String> _readmeIndex;
   late final TokenIndex<IndexedApiDocPage> _apiSymbolIndex;
   late final _scorePool = ScorePool(_packageNameIndex._packageNames);
-  final _tagDocumentIndexes = <String, List<int>>{};
+
+  /// Maps the tag strings to a list of document index values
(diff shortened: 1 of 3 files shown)


app/lib/search/mem_index.dart
```diff
@@ -28,7 +28,7 @@ class InMemoryPackageIndex {
   late final TokenIndex<String> _readmeIndex;
   late final TokenIndex<IndexedApiDocPage> _apiSymbolIndex;
   late final _scorePool = ScorePool(_packageNameIndex._packageNames);
-  final _tagIds = <String, int>{};
+  final _tagDocumentIndexes = <String, List<int>>{};
```
sigurdm, line 31: Nit
```suggestion
  final _tagDocumentIndices = <String, List<int>>{};
```

sigurdm, line 31: Could you add a dartdoc explaining what gets mapped to what here?

sigurdm (approved):

Issue #8400, opened 2024-12-12
Weekly downloads graph not visible on Safari


RepliedSage11:
The new Weekly Downloads graph is not visible on Safari.

**Safari (18.1.1)**
<img width="219" alt="Screenshot 2024-12-12 at 10 44 54" src="https://github.com/user-attachments/assets/884db43a-9504-4e31-b862-9db50b208c3e" />

**Chrome**
![telegram-cloud-photo-size-4-5969713008106128804-x](https://github.com/user-attachments/assets/003be36f-88eb-4293-8273-6a699ca12688)

sigurdm:
Seems like using 100% for width and height also works:
<img width="657" alt="image" src="https://github.com/user-attachments/assets/e5a5c37e-e2ba-4bca-99d6-e976870e514e" />

PR #8410, opened 2024-12-13
Provide width and height for sparkline

sigurdm:
Fixes https://github.com/dart-lang/pub-dev/issues/8400

Tested manually with Safari:

image


isoos (approved):

isoos:
Nit: wouldn't it be better to set these in CSS?

sigurdm:
> Nit: wouldn't it be better to set these in CSS?

YES! That is much smarter.

Which style-file would you put this into?

isoos:
Somewhere here maybe: https://github.com/dart-lang/pub-dev/blob/master/pkg/web_css/lib/src/_detail_page.scss#L238-L285

sigurdm:
It doesn't work - safari needs these to be set as attributes, not as part of the style it seems.

Issue #8337, opened 2024-11-25
Retry `ClientException` from `package:http`


jonasfj:
We should probably also retry `ClientException` from `package:http`:

https://github.com/dart-lang/pub-dev/blob/031c368ea0840f2bdcf2bcaee7886a13c19b90b4/app/lib/shared/storage.dart#L142-L154

isoos:
I think we should refactor/merge the various retry methods we have to make sure every one of them uses the same retry logic. Keeping this issue open to track that.

isoos:
Another case in storage access:

```
Error:
    ClientException: Connection closed before full header was received, uri=https://storage.googleapis.com/storage/v1/b/dartlang-pub-exported-api/o/latest%2Fapi%2Farchives%2Fstacked-2.3.14.tar.gz?projection=full&alt=json

Stack:
    #0      IOClient.send (package:http/src/io_client.dart:156:7)
    <asynchronous suspension>
    #1      AuthenticatedClient.send (package:googleapis_auth/src/auth_http_utils.dart:41:22)
    <asynchronous suspension>
    #2      RetryClient.send (package:http/retry.dart:115:20)
    <asynchronous suspension>
    #3      ApiRequester.request (package:_discoveryapis_commons/src/api_requester.dart:69:20)
    <asynchronous suspension>
    #4      ObjectsResource.get (package:googleapis/storage/v1.dart:3422:23)
    <asynchronous suspension>
    #5      _BucketImpl.info.<anonymous closure> (package:gcloud/src/storage_impl.dart:255:15)
    <asynchronous suspension>
    #6      BucketExt.tryInfo (package:pub_dev/shared/storage.dart:79:14)
    <asynchronous suspension>
    #7      ExportedApi._findStrayFiles.<anonymous closure> (package:pub_dev/package/api_export/exported_api.dart:215:11)
    <asynchronous suspension>
```

dejadior440:
app/lib/search/backend.dart

PR #8412, opened 2024-12-13
retry more operations when accessing storage + enforcer in tests

isoos:
- #8337
- The `RetryEnforcerStorage` wrapper (used only in tests) and related classes parse the current stacktrace when a storage-method is called, and throw an assertion error if `retry` is not in the stack. The detection is rather rudimentary (e.g. needs reentry detection for list + arbitrary processing), but it provides some value already. Further improvements can be done with better parsing (TBD later).
- Most of the methods are currently ignoring the lack of retry, as it seems a lot to fix everything all at once. The current updates focus on the API exporter admin test (it seems to be covered 100%), enabling the assertion error while I was investigating the test case.
- I think for some operation it could have been a simple `withStorage<R>(Future<R> Function(Storage storage)) {...}` helper method that retries internally, but since these operation are not in a transaction, it seems wasted to retry more than one at once.

### DIFF
--- a/pkg/fake_gcloud/lib/retry_enforcer_storage.dart
+++ b/pkg/fake_gcloud/lib/retry_enforcer_storage.dart
@@ -1,0 +1,247 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:gcloud/storage.dart';
+
+void _verifyRetryOnStack() {
+  final st = StackTrace.current.toString();
+  if (st.contains('package:retry/')) return;
+  if (st.contains('retryAsync')) return; // lib/shared/utils.dart
+
+  // detect direct test calls
+  final linesWithoutThisFile = st
+      .split('\n')
+      .where((l) => !l.contains('retry_enforcer_storage.dart'))
+      .toList();
+  if (linesWithoutThisFile.isNotEmpty &&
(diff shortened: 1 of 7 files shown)


pkg/fake_gcloud/lib/retry_enforcer_storage.dart
```diff
@@ -0,0 +1,247 @@
...
+      .split('\n')
+      .where((l) => !l.contains('retry_enforcer_storage.dart'))
+      .toList();
+  if (linesWithoutThisFile.isNotEmpty &&
+      linesWithoutThisFile.first.contains('_test.dart')) {
+    return;
+  }
+
+  print('Missing retry detected:\n$st\n');
+  throw AssertionError('retry is not present in stacktrace: $st');
```
sigurdm, line 25: Should we also detect that we only have a single retry call on the stack?

isoos: I think we should allow multiple re-entry in the retry stack, e.g. one for a high-level iteration of items, while the second one may be narrower operations on it.

We could ignore re-entry detection in the retry stack if we could check if the current call is sufficiently close to the latest retry invocation. But I would consider that check to come later, after the low-hanging fixes (e.g. no retry at all) has been done.

sigurdm: > I think we should allow multiple re-entry in the retry stack, e.g. one for a high-level iteration of items, while the second one may be narrower operations on it.

Yeah - we would need to have some kind of design here. But nested retries easily leads to n^2 attempts...

> But I would consider that check to come later, 

SG.

PR #8414, opened 2024-12-16
Report unmapped fields in integrity checks + delete Package.isWithheld and withheldReason

isoos:
Note: isBlocked could also be removed, but to be on the safe side I'd do that in the release after the current one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ AppEngine version, listed here to ease deployment and troubleshooting.
 ## Next Release (replace with git tag when deployed)
  * Bump runtimeVersion to `2024.12.12`.
  * Upgraded runtime Dart SDK to `3.6.0`.
- * Note: Started reporting unmapped fields and deleting `Package.isWithheld` and `Package.withheldReason`.
+ * Note: Started reporting unmapped fields 
+ * Started and deleting `Package.isWithheld` and `Package.withheldReason`.
 
 ## `20241212t111200-all`
  * Bump runtimeVersion to `2024.12.11`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ AppEngine version, listed here to ease deployment and troubleshooting.
 ## Next Release (replace with git tag when deployed)
  * Bump runtimeVersion to `2024.12.12`.
  * Upgraded runtime Dart SDK to `3.6.0`.
+ * Note: Started reporting unmapped fields and deleting `Package.isWithheld` and `Package.withheldReason`.
 
 ## `20241212t111200-all`
  * Bump runtimeVersion to `2024.12.11`.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ AppEngine version, listed here to ease deployment and troubleshooting.
 ## Next Release (replace with git tag when deployed)
  * Bump runtimeVersion to `2024.12.12`.
  * Upgraded runtime Dart SDK to `3.6.0`.
+ * Note: Started reporting unmapped fields and deleting `Package.isWithheld` and `Package.withheldReason`.
```
sigurdm, line 7:
```suggestion
 * Note: Started reporting unmapped fields 
 * Started and deleting `Package.isWithheld` and `Package.withheldReason`.
```

app/lib/shared/integrity.dart
```diff
@@ -93,7 +98,13 @@ class IntegrityChecker {
       yield* _checkAuditLogs();
       yield* _checkModerationCases();
       yield* _reportPubspecVersionIssues();
-      // TODO: report unmapped properties
+
+      if (_unmappedFields.isNotEmpty) {
+        for (final field in _unmappedFields) {
+          if (_knownUnmappedFields.contains(field)) continue;
+        yield 'Unmapped field found: $field.';
```
sigurdm, line 105: Can we report more? Like give some entity with this field, and the entitty type

app/lib/shared/integrity.dart
```diff
@@ -40,6 +40,11 @@ class IntegrityChecker {
   final DatastoreDB _db;
   final int _concurrency;
 
+  static const _knownUnmappedFields = {
```
sigurdm, line 43: Please document the scope of this (is it only about packages?)

app/lib/shared/integrity.dart
```diff
@@ -40,6 +40,11 @@ class IntegrityChecker {
...
+    'Package.isWithheld',
+    'Package.withheldReason',
+  };
+  final _unmappedFields = <String>{};
```
sigurdm, line 47: Just to make sure, do we create a new IntegrityChecker for each run?

app/lib/shared/integrity.dart
isoos: We are reporting `EntityType.fieldName` here. I'm not sure if we need to have a full entity identifier, as these fields are typically unused, but maybe we could report one id for each of the reported entries? That way we can store it in a Map and only keep the first one?

app/lib/shared/integrity.dart
isoos: Yes, this is a one-off object on each run.

app/lib/shared/integrity.dart
isoos: renamed + done.

app/lib/shared/integrity.dart
isoos: Implemented this, PTAL.

sigurdm (approved):

PR #8417, opened 2024-12-16
Stricter retry verification.

isoos:
- #8337
- As we only check the first frame of the stack trace, we don't need to handle further re-entry detection.

### DIFF
--- a/pkg/fake_gcloud/lib/retry_enforcer_storage.dart
+++ b/pkg/fake_gcloud/lib/retry_enforcer_storage.dart
@@ -5,19 +5,41 @@
 import 'dart:async';
 
 import 'package:gcloud/storage.dart';
+import 'package:stack_trace/stack_trace.dart';
+
+const _skippedFirstFrames = {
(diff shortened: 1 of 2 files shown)


pkg/fake_gcloud/lib/retry_enforcer_storage.dart
```diff
@@ -5,19 +5,41 @@
...
 import 'package:gcloud/storage.dart';
+import 'package:stack_trace/stack_trace.dart';
+
+const _skippedFirstFrames = {
+  'package:fake_gcloud/retry_enforcer_storage.dart',
+  'package:pub_dev/shared/storage.dart',
+};
 
 void _verifyRetryOnStack() {
   final st = StackTrace.current.toString();
```
sigurdm, line 16: Is this still used?

isoos: good catch, removed

sigurdm (approved):

PR #8441, opened 2025-01-02
Notify uploader(s) when remaining version count is less than 100.

### LINKED ISSUE #8053: Create an allow-list of packages that can publish more than 1000 versions

isoos:
- Fixes #8053.
- Prior discussion concluded that #8063 has way more complexity than needed, but we could send a message to the uploader(s) that they are near the current version count limit. This PR creates a message for the pub client and also includes it in the notification email that is sent to the admins of the package / publisher.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -882,8 +882,9 @@ class PackageBackend {
     );
   }
 
-  /// Finishes the upload of a package.
-  Future<PackageVersion> publishUploadedBlob(String guid) async {
+  /// Finishes the upload of a package and returns the list of messages
+  /// related to the publishing.
+  Future<List<String>> publishUploadedBlob(String guid) async {
     final restriction = await getUploadRestrictionStatus();
     if (restriction == UploadRestrictionStatus.noUploads) {
       throw PackageRejectedException.uploadRestricted();
@@ -984,7 +985,7 @@ class PackageBackend {
       sw.reset();
       final entities = await _createUploadEntities(db, agent, archive,
           sha256Hash: sha256Hash);
-      final version = await _performTarballUpload(
+      final (version, uploadMessages) = await _performTarballUpload(
         entities: entities,
         agent: agent,
         archive: archive,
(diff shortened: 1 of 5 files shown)


app/lib/package/backend.dart
```diff
@@ -1156,10 +1156,20 @@ class PackageBackend {
...
+      if (remainingVersionCount <= 0) {
         throw PackageRejectedException.maxVersionCountReached(
             newVersion.package, maxVersionCount);
       }
+      if (remainingVersionCount <= 100) {
+        // We need to decrease the remaining version count as the newly uploaded
+        // version is not yet in it.
+        final limitAfterUpload = remainingVersionCount - 1;
+        uploadMessages.add(
+            'The package "${package!.name!}" has $limitAfterUpload versions left '
```
sigurdm, line 1169: Nit:
```suggestion
        final limitAfterUpload = remainingVersionCount - 1;
        final s = limitAfterUpload == 1 ? '' : 's';
        uploadMessages.add(
            'The package "${package!.name!}" has $limitAfterUpload version$s left '
```

sigurdm (commented):
Nice!

It would be nice to have a documentation link we could point to in the message. But I don't know if that is too much work.

We could perhaps link to a github issue...

sigurdm (approved):

PR #8446, opened 2025-01-06
Remove popularity from search index.

isoos:
- Follow-up to #8435.
- The sort value is not removed for now, we could probably keep it for a while, in case somebody bookmarked it.

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -40,7 +40,6 @@ class InMemoryPackageIndex {
   late final List<IndexedPackageHit> _overallOrderedHits;
   late final List<IndexedPackageHit> _createdOrderedHits;
   late final List<IndexedPackageHit> _updatedOrderedHits;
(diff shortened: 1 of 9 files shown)


sigurdm (approved):
I guess we could forward the old popularity searches to search ranked by download counts if we want to preserve old links.

Alternatively they could go to the generic ranking.

isoos:
> I guess we could forward the old popularity searches to search ranked by download counts if we want to preserve old links.

If you do https://pub.dev/packages?sort=x it will use the default sort order. So maybe what we need here is to canonicalize the search URLs and when we get a non-canonical one, we redirect to it. We already have parts of the redirect in place, maybe we just need to extend it.

sigurdm:
> If you do https://pub.dev/packages?sort=x it will use the default sort order.

I think that is good enough. Let's just remove the popularity one then. Old links would still lead to **some** search result.

PR #8459, opened 2025-01-13
Upgrade dependencies (incl. mailer, analyzer, mime, source_gen)

isoos:


app/pubspec.yaml
```diff
@@ -49,7 +49,7 @@ dependencies:
   # pana version to be pinned
   pana: '0.22.18'
   # 3rd-party packages with pinned versions
-  mailer: '6.2.0'
+  mailer: '6.3.0'
```
sigurdm, line 52: 6.4.0 is already available (though doesn't seem to have anything interesting)

sigurdm (approved):

app/pubspec.yaml
isoos: It has a conflicting and arguably worse dependency on intl, see: https://github.com/kaisellgren/mailer/pull/261

isoos: I've added a comment on it, I think we don't need to upgrade to it.

PR #8472, opened 2025-01-16
Fix caching of Youtube API pagination + filter duplicate videoIds.

### LINKED ISSUE #8471: Package of the week section can end up with the same video twice
Happened today:
Image

isoos:
Fixes #8471.

### DIFF
--- a/app/lib/service/youtube/backend.dart
+++ b/app/lib/service/youtube/backend.dart
@@ -121,54 +121,64 @@ class _PkgOfWeekVideoFetcher {
     final youtube = YouTubeApi(apiClient);
 
     try {
+      final pageTokensVisited = <String>{};
+      String nextPageToken = '';
+
       final videos = <PkgOfWeekVideo>[];
-      String? nextPageToken;
-      for (var check = true; check && videos.length < 50;) {
-        final rs = await cache.youtubePlaylistItems().get(
+      final videoIds = <String>{};
+      while (videos.length < 50) {
+        // check visited status
+        if (pageTokensVisited.contains(nextPageToken)) {
+          break;
+        }
+        pageTokensVisited.add(nextPageToken);
+
+        // get page from cache or from Youtube API
+        final rs = await cache.youtubePlaylistItems(nextPageToken).get(
               () async => await youtube.playlistItems.list(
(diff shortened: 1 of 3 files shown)


app/lib/service/youtube/backend.dart
```diff
@@ -121,54 +121,64 @@ class _PkgOfWeekVideoFetcher {
...
+            ));
+          } catch (e, st) {
+            // this item will be skipped, the rest of the list may be displayed
+            _logger.pubNoticeShout(
+                'youtube', 'Processing Youtube PlaylistItem failed.', e, st);
+          }
+        }
+
+        // advance to next page token
+        nextPageToken = rs.nextPageToken ?? '';
```
sigurdm, line 181: Can't you just break here if `rs.nextPageToken == null` ?

app/lib/service/youtube/backend.dart
```diff
@@ -121,54 +121,64 @@ class _PkgOfWeekVideoFetcher {
     final youtube = YouTubeApi(apiClient);
 
     try {
+      final pageTokensVisited = <String>{};
+      String nextPageToken = '';
```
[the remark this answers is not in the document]
isoos: Updated, PTAL.

app/lib/service/youtube/backend.dart
```diff
@@ -121,54 +121,63 @@ class _PkgOfWeekVideoFetcher {
...
+                'youtube', 'Processing Youtube PlaylistItem failed.', e, st);
+          }
+        }
+
+        pageTokensVisited.add(nextPageToken ?? '');
+        // advance to next page token
         nextPageToken = rs.nextPageToken;
-        check = nextPageToken != null && nextPageToken.isNotEmpty;
+        if (nextPageToken == null ||
+            pageTokensVisited.contains(nextPageToken)) {
```
sigurdm, line 178: Can this really happen?

If so, please explain in a comment. I would expect the nextPageToken to not link us in a cycle.

sigurdm (approved):

app/lib/service/youtube/backend.dart
isoos: Removed. It may be a bit more defensive coding than needed.

PR #8475, opened 2025-01-17
Alternative box-shadow for dark theme.

isoos:
I've been looking into the recommendations for shadows in dark mode, and the most prominent one is just to remove them like in [this article](https://medium.com/@tundehercules/designing-effective-dark-mode-interfaces-17f38ecea2e9). However, [another article](https://codyhouse.co/nuggets/beautiful-css-shadows) suggested a different solution, and after a bit of experimentation, I think I like that we can have the same style for both modes.

Note: while the dark background is not really visible in dark mode, the additional opaque black shadow does create a slight contrast with the opaque white inset. To me it does feel like it has some kind of natural depth.

I am a bit undecided upon the "up" or "down" effect: in [this article](https://www.joshwcomeau.com/css/designing-shadows/) the "elevation" of an element is connected with the darkness and narrowness of the shadow. So if we want to keep being consistent, we want to have smaller shadow on hover (as well as darker, but that's already done). However, the current pub.dev implementation increases the shadow on hover, so I've kept it for now, but maybe we should rather go with the different direction?

image

image

### DIFF
--- a/pkg/web_css/lib/src/_home.scss
+++ b/pkg/web_css/lib/src/_home.scss
@@ -108,14 +108,14 @@
   .mini-list-item {
     background: var(--pub-neutral-bgColor);
     border-radius: 4px;
-    box-shadow: 0px 2px 7px 0px var(--pub-home_card-box_shadow-color);
+    box-shadow: var(--pub-box-shadow);
     padding: 28px 30px 30px; // title's top gap is about 2px (30-2 => 28)
     margin-bottom: 10px;
     min-height: 100px;
 
     &:hover {
       background: var(--pub-neutral-hover-bgColor);
-      box-shadow: 0px 4px 9px 0px var(--pub-home_card_hover-box_shadow-color);
+      box-shadow: var(--pub-box-hover-shadow);
 
       @media (min-width: variables.$device-desktop-min-width) {
         .mini-list-item-body {
--- a/pkg/web_css/lib/src/_variables.scss
+++ b/pkg/web_css/lib/src/_variables.scss
@@ -80,12 +80,23 @@
   --mdc-theme-secondary: #0066d9;
   --mdc-typography-font-family: var(--pub-default-text-font_family);
 
+  // Download chart colors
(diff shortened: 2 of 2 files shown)


isoos:
Note: a staging should have this online soon (alongside with a restore from backup).

/cc @parlough

sigurdm:
Could you also add a screenshot showing the alternative (dark mode without shadows) for comparison?

isoos:
One possibility: only border:

image

Another, using box-shadow to draw solid border:

image

sigurdm:
I like both.

By the way, did you read the material dark-mode guide on this?
https://m2.material.io/develop/android/theming/dark#:~:text=Shadows%20are%20less,implied%20light%20source.

isoos:
I think the remaining part from the guide may be the different surface color, but I didn't want to go there yet, only to remove the not-that-nice current implementation of the shadow.

My preference would be the contrast that the PR gives, but I also like the simple box-shadow marking the border (as in the "cron" example above). I think the border-only alternative is too flat, I'd rather use any of the other versions.

pkg/web_css/lib/src/_home.scss
```diff
@@ -108,14 +108,14 @@
   .mini-list-item {
     background: var(--pub-neutral-bgColor);
     border-radius: 4px;
-    box-shadow: 0px 2px 7px 0px var(--pub-home_card-box_shadow-color);
+    box-shadow: var(--pub-box-shadow);
```
jonasfj, line 111: Should use something other than `shadow` at all? just curious.

isoos: See the linked articles: the basic suggestion is to use border + lighter background for elevating count out of its surroundings. I still find those mostly flat, and I thought the inset + minimal shadow contrast may be interesting to check out. If that is not winning here, we shall do border + lighter background (though the inset is kind of that, with slight gradient changes).

isoos:
Closing in favor of #8489.

PR #8476, opened 2025-01-20
Fix styles.scss to include alert block styles + test.

isoos:


pkg/web_css/test/file_use_test.dart
```diff
@@ -0,0 +1,32 @@
...
+            l.substring(4).trim().split(';').first.replaceAll("'", '').trim())
+        .where((v) => v.isNotEmpty)
+        .toSet();
+    expect(references, isNotEmpty);
+    expect(references, contains('src/_tags'));
+
+    final files = await Directory('lib')
+        .list(recursive: true)
+        .where((f) => f is File && f.path.endsWith('.scss'))
+        .cast<File>()
```
sigurdm, line 24: Perhaps something like this would work?
```suggestion
        .whereType<File>().where((f) => f.path.endsWith('.scss'))
```

pkg/web_css/test/file_use_test.dart
```diff
@@ -0,0 +1,32 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:test/test.dart';
+
+void main() {
+  test('all scss files are used from style.scss', () async {
```
sigurdm, line 11: Maybe something like this:
```suggestion
  test('scss files in lib/ are exactly those referenced by style.scss ', () async {
```

sigurdm (approved):

PR #8477, opened 2025-01-20
Fix h3 issue on landing page.

isoos:
Lighthouse reported on the landing page that `Heading elements are not in a sequentially-descending order`, complaining about semantic level. At first, I've tried to remove the element content of the `a` link, but for some weird styling issue, it cropped out the bottom of the label, e.g. `g` characters became unreadable. Switching the `h3` to `span` and adding ARIA attributes seems to be the proper fix here.


sigurdm (approved):
Surprising that we need this - but can't say I have a better idea.

PR #8484, opened 2025-01-22
Darker background colors in dark mode for increased contrast.

isoos:
Following further Lighthouse reports on not enough contrast in certain context, and also comparing the brightness and (de)saturation of these colors with other dark mode palettes.

- `#1b1b1b` -> `#121317` (default background)
- `#373737` -> `#2f3034` (hover bg)
- `#41424c` -> `#242b32` (inset, code highlight bg)

Before/after example for hover: 
image
image

Before/after example for code highlight:
image
image

### DIFF
--- a/pkg/web_css/lib/src/_variables.scss
+++ b/pkg/web_css/lib/src/_variables.scss
@@ -148,16 +148,16 @@
 .dark-theme {
   @include dash_variables.dark-theme;
 
-  --pub-color-eerieBlack:      #1b1b1b;
-  --pub-color-shadowBlack:     #373737;
-  --pub-color-anchorBlack:     #41424c;
+  --pub-color-darkBlack:       #121317;
+  --pub-color-darkGunmetal:    #242b32;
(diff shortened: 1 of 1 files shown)


jonasfj:
I don't have any strong feelings here, mostly I hope we can align with dart.dev plans @parlough ?

parlough:
Thanks for working to improve the contrast here!

I'm not sure I love completely losing the saturation. I feel it makes the site seem a bit grey and not Dart related.

I think to maintain some saturation and further improve the contrast, we can go a bit darker. I imagine many developers would want a bit darker anyway.

---

Playing around a bit with the same content, I found something around this combination works pretty well. What do you think?

Page background: `#121317`
Code block background: `#242b32`

Darker, more saturated colors

Slightly lower saturation and/or lightness could work as well, particularly for the code block.

isoos:
Updated the PR and its images with the darker colors @parlough suggested + experimented with a relevant hover background too. (also staging update ongoing).
PTAL.

sigurdm (approved):

PR #8489, opened 2025-01-23
Replace drop-shadow with only borders in dark mode.

isoos:
I think this is the least controversial change to the dark-mode shadows we could do: replacing them with a thin border. We could still incorporate parts of #8475 if we wanted to...

Note: this does not add the extra border to the light mode, but I think we should also consider it (separately from this).

Before/after:
image
image

Before/after with hovering:
image
image

### DIFF
--- a/pkg/web_css/lib/src/_home.scss
+++ b/pkg/web_css/lib/src/_home.scss
@@ -106,16 +106,16 @@
   }
 
   .mini-list-item {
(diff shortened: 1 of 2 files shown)


parlough:
I think this is a lot less visually jarring and helps the hover style stand out a bit more, thanks!

sigurdm (approved):

PR #8490, opened 2025-01-24
Refactor proposal: use unnamed callback function in with*HttpPubApiClient methods.

isoos:
This is more an idea than a strong opinion, but #8485 reminded me that we started using named `fn` callbacks before the language allowed different orders for named and ordered parameters. Now, since the language allows the mix, it may be worth checking if the all-named or the mixed-order version is working better for us. Any opinions?


sigurdm (approved):
I like it!

Issue #8493, opened 2025-01-24
Opacity change on hover may not be a good idea


isoos:
Our default link stylesheet decreases the link's opacity on hovering, affecting both text and images inside the link. In both light mode and dark mode this makes the content closer to the background, also decreasing the overall contrast, which is undesirable.

isoos:
I think this has been done and can be closed.

PR #8494, opened 2025-01-24
Dark mode link hover: increase brightness instead lower opacity.

isoos:
- #8493
- Note: `filter: brightness()` did not work for me on images, but maybe other factors are at play.
- I'm not yet sure how to resolve the same opacity issue in light mode, given we are using it the current way for a long time, and also the brightness issue above, so I'm limiting this change to dark mode for now, as it seems to be an improvement there.

### DIFF
--- a/pkg/web_css/lib/src/_variables.scss
+++ b/pkg/web_css/lib/src/_variables.scss
@@ -242,6 +242,13 @@
   .displayed-in-light-theme {
     display: none !important;
   }
+
+  a {
+    &:hover {
+      opacity: 1.0;
+      filter: brightness(120%);
+    }
+  }
 }
 
 @mixin elevated-content-border {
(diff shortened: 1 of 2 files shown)


isoos:
I've started a staging push with this, so that we can take a look at it at least in dark mode, and let's discuss how we should try to solve it for images too.

/cc @parlough

pkg/web_css/lib/src/_variables.scss
```diff
@@ -242,6 +242,13 @@
   .displayed-in-light-theme {
     display: none !important;
   }
+
+  a {
+    &:hover {
+      opacity: 1.0;
```
jonasfj, line 248: do we have opacity other than 1?

isoos: Yes, the version tab icons have it on 0.7 by default. However, in this case this is to counter the main rule here:
https://github.com/dart-lang/pub-dev/blob/master/pkg/web_css/lib/src/_base.scss#L98-L100

I shall move that out to a separate light-theme/dark-theme styles so we don't need the override.

isoos: Done, PTAL.

jonasfj (approved):

PR #8504, opened 2025-01-28
Timeouts and VM exit for index building process.

isoos:
- The callback will start a timer that, if not cancelled, will call `exit` to make sure the VM is restarted.
- Added regular timeouts to storage download and also upload calls.

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -149,7 +149,7 @@ class SearchBackend {
       // update or remove the document
       await retry(() async {
         try {
-          final doc = await loadDocument(package);
+          final doc = await loadDocument(package).timeout(Duration(minutes: 2));
           snapshot.add(doc);
         } on RemovedPackageException catch (_) {
           snapshot.remove(package);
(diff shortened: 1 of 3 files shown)


app/lib/shared/utils.dart
```diff
@@ -305,3 +308,31 @@ extension ByteFolderExt on Stream<List<int>> {
...
+  final timer = Timer(timeout, () {
+    // Give the logging a short time to be stored outside of the machine.
+    Timer(Duration(seconds: 10), () async {
+      exit(-1);
+    });
+
+    stderr.writeln('Timeout triggering VM termination\n$trace');
+    _logger.pubNoticeShout(
+      'vm-termination',
+      'Timeout triggering VM termination',
```
jonasfj, line 328: Does this terminate the VM or does AppEngine just restart the process?

app/lib/search/backend.dart
```diff
@@ -158,29 +158,34 @@ class SearchBackend {
     }
 
     // initial scan of packages
-    final pool = Pool(concurrency);
-    final futures = <Future>[];
-    await for (final package in dbService.query<Package>().run()) {
-      if (package.isNotVisible) {
-        continue;
+    await withVmTerminationTimeout(timeout: Duration(hours: 12), () async {
```
jonasfj, line 161: Why not just add `.timeout(Duration(hours: 12))` when calling `doCreateAndUpdateSnapshot`?

If such a timeout does happen there, isn't that a good place to just not catch the timeout and let it bubble all the way up? Do we need to explicitly kill the VM?

isoos:
Closing in favor of #8525.

PR #8505, opened 2025-01-29
Increase font weight in detail page tabs labels to improve contrast.

### LINKED ISSUE #8306: Red tab text on account pages too low contrast in dark mode
Tab text red color is too low contrast in dark mode

---

It's also not super clear to me why these tabs are red. To indicate the contents are only accessible if logged in? Perhaps the text should be colored normally and the lock icon could be red? Too much red on the page is a bit jarring.

isoos:
Fixes #8306

I've tried to find different (red) color combinations for the dark mode, but it turns out the issue was more about the thin font over the dark gray background. Once the weight is increased, Lighthouse seems to be happier with the contrasts.

Before/after in light mode:
image
image

Before/after in dark mode:
image
image

An alternative could be to apply the font width only in the dark theme.

### DIFF
--- a/pkg/web_css/lib/src/_detail_page.scss
+++ b/pkg/web_css/lib/src/_detail_page.scss
@@ -373,6 +373,7 @@ $detail-tabs-tablet-width: calc(100% - 240px);
 
   > .tab-button,
   > .tab-link > a {
+    font-weight: 500;
     display: block;
     color: var(--pub-detail_tab-text-color);
     padding: 12px 0px;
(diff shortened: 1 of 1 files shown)


jonasfj (approved):
Curious if @parlough has any thoughts?

parlough:
I think this is fine if necessary, but instead or in addition, a color with a higher lightness could be used, such as `#ff6666`. Still ends up red but a bit less harsh.

Example of the higher-lightness red being used in the taps mneu

isoos:
Closing in favor of #8533.

PR #8509, opened 2025-01-30
Remove isolates from the analyzer service.

isoos:
- Added a TODO to rewrite the index building loop (doing it separately).
- Note: we should test this on staging first for a longer time.

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -96,7 +96,7 @@ class SearchBackend {
   ///
   /// When other process has the claim, the loop waits a minute before
   /// attempting to get the claim.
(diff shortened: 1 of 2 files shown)


app/lib/search/backend.dart
```diff
@@ -96,7 +96,7 @@ class SearchBackend {
   ///
   /// When other process has the claim, the loop waits a minute before
   /// attempting to get the claim.
-  Future<Never> updateSnapshotInForeverLoop() async {
+  Future<void> updateSnapshotInForeverLoop() async {
```
sigurdm, line 99: Is this change needed?

jonasfj (approved):

jonasfj (approved):

PR #8512, opened 2025-01-31
Detect incompatible local Chrome versions when running puppeteer.

isoos:
It looks like there is some known incompatibility with new chrome versions and the puppeteer package. Updating the code to check for version when attempting to run local chrome instance, and also removing the extra check from the CI environment (in case it upgraded the chrome version).

See https://github.com/xvrh/puppeteer-dart/issues/364

### DIFF
--- a/pkg/pub_integration/lib/src/test_browser.dart
+++ b/pkg/pub_integration/lib/src/test_browser.dart
@@ -45,13 +45,27 @@ class TestBrowser {
     final binaries = [
       '/usr/bin/google-chrome',
     ];
-    for (final binary in binaries) {
-      if (File(binary).existsSync()) return binary;
-    }
 
-    // sanity check for CI
-    if (Platform.environment['CI'] == 'true') {
-      throw StateError('Could not detect chrome binary while running in CI.');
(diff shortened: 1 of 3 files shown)


pkg/pub_integration/lib/src/test_browser.dart
```diff
@@ -45,13 +45,27 @@ class TestBrowser {
...
+            .split(' ')
+            .expand((p) => p.split('.'))
+            .map(int.tryParse)
+            .whereType<int>()
+            .firstOrNull;
+        // No version string, better not running it.
+        if (mainVersion == null) continue;
+        // Main version is after the currently supported version, will fail.
+        // https://github.com/xvrh/puppeteer-dart/issues/364
+        if (mainVersion >= 132) continue;
```
sigurdm, line 65: Won't we all soon only have versions after this?

isoos: Maybe, I don't really know. We probably will, when we upgrade the puppeteer version next time.

sigurdm (approved):

PR #8519, opened 2025-02-03
Increase contrast of pub detail tags in dark mode

parlough:
**After:**

Screenshot of pub detail tags on provider package after change<br>

**Before:**

Screenshot of pub detail tags on provider package before change


pkg/web_css/lib/src/_variables.scss
```diff
@@ -152,13 +152,14 @@
   --pub-color-darkGunmetal:    #242b32;
   --pub-color-darkGray:        #2f3034; /* derived from dark black by making it lighter */
   --pub-color-nipponUltraBlue: #23607f;
```
isoos, line 154:
```suggestion
```

isoos (approved):

PR #8522, opened 2025-02-03
Remove caching of download counts in PackageView

szakarias:


jonasfj (approved):
We will need to bump `runtimeVersion` to avoid conflicting data in cache.

PR #8524, opened 2025-02-03
Darker Flutter Favorite logos for dark mode.

isoos:
The separate image diffs are below, before/after in context:
image
image

Note: This was done via pixel color manipulation in gimp, and it is likely we should get a designer to re-do the logos with proper vectorized assets.


isoos:
/cc @parlough 

Also: this is now on staging.

sigurdm:
I like the slightly darker background of the logo - but to me the contrast of the text is now to small.

isoos:
> I like the slightly darker background of the logo - but to me the contrast of the text is now to small.

I have done couple of iterations, but I can't really keep the contrast up, it is a dark text after all. Especially not with this pixelated version of the logo... Let's keep it as-is then?

sigurdm:
Maybe the text should now be brighter than the surface?

isoos:
@sigurdm: Updated the image, slightly darker background than currently, also turning the label text color down. wdyt?

sigurdm:
I like it!

sigurdm (approved):

PR #8525, opened 2025-02-03
Strict index builder timeout + exceptions are not catched.

isoos:
- A safeguard to wait at least an hour before crashing the VM, just in case we have a crash loop.
- Note: this only has a top-level timeout, we could add lower-level timeouts, but the top-level catch-all should also be updated to rethrow either way.


app/lib/search/backend.dart
```diff
@@ -101,15 +106,28 @@ class SearchBackend {
...
-          await doCreateAndUpdateSnapshot(claim);
+          // Force timeout exception if the process does not release the lock.
+          final lockHoldTimeout = _maxLockHoldPeriod + Duration(hours: 1);
+          await doCreateAndUpdateSnapshot(claim).timeout(lockHoldTimeout);
         });
       } catch (e, st) {
-        _logger.warning('Snapshot update failed.', e, st);
+        _logger.pubNoticeShout(
+            'snapshot-building', 'Snapshot update failed.', e, st);
+        // Force waiting at least an hour before we rethrow the exception
```
sigurdm, line 120: Is this to avoid too quick cycling of restarts? Could you explain that in this comment?

sigurdm (approved):

PR #8529, opened 2025-02-06
Downloads chart: Add tooltip with exact counts

szakarias:
image

### DIFF
--- a/pkg/web_css/lib/src/_pkg.scss
+++ b/pkg/web_css/lib/src/_pkg.scss
@@ -391,32 +391,32 @@
   }
 
   .downloads-chart-square-blue {
-    background-color:var(--pub-downloads-chart-color-0-bg);
+    background-color:var(--pub-downloads-chart-color-bg-0);
     color:var(--pub-downloads-chart-color-0);
   }
 
   .downloads-chart-square-red {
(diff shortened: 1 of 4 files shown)


pkg/web_css/lib/src/_pkg.scss
```diff
@@ -303,6 +303,49 @@
     stroke: var(--pub-downloads-chart-frame-color);
   }
 
+  .downloads-chart-tooltip {
+    border-radius: 5px;
+    margin: 0px;
+    padding: 2px 8px;
+    box-shadow: rgba(0, 0, 0, 0.1) 0px 4px 12px;
+    background-color:var(--pub-neutral-bgColor);
+    opacity: 0.9;
```
sigurdm, line 312: Do we want this?
It might make the text in the tooltip slightly harder to read...

szakarias: I kind of like it with a little bit opacity, but I see your point too....

szakarias: Made it such that it is only the background that has a small opacity, keeping the contrast in the text and squares.

sigurdm (approved):

PR #8533, opened 2025-02-06
Tab redesign: removing icon, orange text in dark mode.

isoos:
- using the orange color as discussed
- removed icon, added title instead, not sure if we should have different title for `my-<page>` links vs `/admin`
- verified in lighthouse, contrast score is great with this color
  
new screens:

image

--

image

--

image

--

image


isoos:
Note: available on staging.

app/lib/frontend/templates/views/shared/detail/tabs.dart
```diff
@@ -19,6 +19,9 @@ d.Node detailTabsNode({
           children: tabs.map(
             (t) => d.li(
               classes: t.titleClasses,
+              attributes: {
+                if (t.isPrivate) 'title': 'For authorized users only.',
```
sigurdm, line 23: I think this is slightly misleading.

I would like the tooltip to reflect the tab title, not the reason for the different color.

sigurdm (approved):
I like how this looks!

app/lib/frontend/templates/views/shared/detail/tabs.dart
isoos: Do you have some examples in mind for that different text?

sigurdm: I think we should have a title text for each tab:

Something along the lines of:

Readme: "Description of the package"
Changelog: "Describes changes between versions"
Example: "An example of how to use this package"
Installing: "How to use this package in your own project"
Versions: "All published versions of this package"
Scores: "Automated analysis and download metrics"
Admin: "Admin settings for this package - only for admins of this package"
Activity log: "Historical admin and publication events - only for admins of this package"

sigurdm: Sorry that I was not more clear, but I don't think it should block the landing of this PR. I can live with the titles as they are.

isoos:
I think I'll land this, and add new descriptions as suggested above in a follow-up PR.

PR #8542, opened 2025-02-07
Use pub-monochrome-icon in the package title's copy icon.

### LINKED ISSUE #8536: [Dark mode] Package ID copy button is too dark
On the package package (`/packages/foo`, there's a copy button to the right of the package's ID:

Image

This button is difficult to see as it is too dark. Perhaps we should make it lighter?

isoos:
- Fixes #8536
- Expanded monochrome icon style to also include focus and transition.
- Using CSS variable to control `opacity`, so that components that want to have different opacity ranges doesn't need to repeat all the hover/focus rules, though in this case we still need the title as parent (may be refactored in the future).

Before:
image
image

After:
image
image

Before:
image
image

After:
image
image

### DIFF
--- a/app/lib/frontend/templates/views/pkg/title_content.dart
+++ b/app/lib/frontend/templates/views/pkg/title_content.dart
@@ -20,7 +20,11 @@ d.Node copyIcon({required String package, required String version}) {
     classes: ['pkg-page-title-copy'],
     children: [
       d.img(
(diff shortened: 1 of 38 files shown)


isoos:
Note: staging deployment is ongoing.

sigurdm (approved):

Issue #7391, opened 2024-01-15
Links should be visually distinct without relying on color


szakarias:
We should consider and discuss how this can look without making our site look messy.

sigurdm:
The standard way here is to underline all links...

sigurdm:
We could also consider underlining on hover only... seems to be good enough for google search results.

sigurdm:
https://dequeuniversity.com/rules/axe/4.8/link-in-text-block

PR #8544, opened 2025-02-07
Underline links on hover by default

parlough:
dart.dev and docs.flutter.dev updated to do this for most links as well to meet the requirement of using more than color to distinguish link interactivity.

Resolves https://github.com/dart-lang/pub-dev/issues/7391

### DIFF
--- a/pkg/web_css/lib/src/_base.scss
+++ b/pkg/web_css/lib/src/_base.scss
@@ -120,6 +120,10 @@ a {
     padding: 4px 12px;
   }
 
+  &:hover {
+    text-decoration: underline;
+  }
+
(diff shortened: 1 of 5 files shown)


isoos:
Note: this is on staging right now.

isoos (approved):

isoos:
I've found the underline a bit harsh - I guess I'm used to using the site without it. Maybe we could have a different underline color using https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-color ?

jonasfj (approved):
Let's land this after the release on Wednesday, just to derisk a bit.

PR #8562, opened 2025-02-13
Support test profile metadata to fine-tune generated content.

isoos:


isoos:
Closing as #8634 and its follow-up will be implemented instead.

PR #8563, opened 2025-02-13
Fix test processing Flutter API index.

isoos:


isoos:
Looks like Flutter API is now generated with a newer dartdoc too :)

jonasfj (approved):

Issue #8518, opened 2025-02-02
Search for a package with its verbatim name still does not display the package in first position


crcdng:
When I search for `flutter_animate`, a well-maintained package with high stats, the default ("Sort by search relevance") search result does not display the package in first position. 

Instead, it is shown somewhere down below, currently at 8th position, below a totally unrelated package named `map_picker`.

The expected result is:

* `flutter_animate` in first position
* followed by other packages that are related, e.g. with similar names.

This is a long-standing issue with pub.dev that affects numerous packages.

isoos:
@crcdng  Thanks for reporting it!

@jonasfj @sigurdm:

This seems to be a tough case, because for the given search query:
- it is not the most liked https://pub.dev/packages?q=flutter_animate&sort=like
- it is not the most downloaded https://pub.dev/packages?q=flutter_animate&sort=downloads
- it does not have maximum points (150/160) while other packages have that

I'm 95% certain that if those static analysis points were fixed, it would be on the first few spots, likely on the first.

I'm inclined to say that the points in this case are causing a bit more divide in the ranking than warranted. Maybe we should dampen the difference somehow?

sigurdm:
To me it seems fair that the package is not at the top, given that it is not the most popular by downloads nor the most liked nor has the most analysis points. This looks like it is working as intended.

We do show the exact package name match at the top:

![Image](https://github.com/user-attachments/assets/1d0bfe43-ce0e-4dd3-bbde-759caac03469)

crcdng:
> To me it seems fair that the package is not at the top, given that it is not the most popular by downloads nor the most liked nor has the most analysis points. This looks like it is working as intended.
> 
> We do show the exact package name match at the top:
> 
> ![Image](https://github.com/user-attachments/assets/1d0bfe43-ce0e-4dd3-bbde-759caac03469)

It's funny - I did not see the "Matching package names..." until mentioned.

I would still argue that a package repository should **always** display the (one) direct match in the first place regardless of metrics.

sigurdm:
> I would still argue that a package repository should always display the (one) direct match in the first place regardless of metrics.

I think this was once the actual behavior, and was decided against. There can be abandoned packages that happen to have squatted some keyword. They should not be shown at the top of the list IMO.

We show direct matches, and we show a ranked list of matches.

I believe having the match in the package name ranks the package slightly higher (@isoos please confirm) but in this case it seems to be not enough.

[...]

PR #8572, opened 2025-02-18
Use square root of points for normalized score.

isoos:
Improves ranking for almost-perfect packages in #8518.

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -313,9 +313,10 @@ class InMemoryPackageIndex {
     _adjustedOverallScores = _documents.map((doc) {
       final downloadScore = doc.downloadScore ?? 0.0;
       final likeScore = doc.likeScore ?? 0.0;
-      final combinedScore = (downloadScore + likeScore) / 2;
-      final points = doc.grantedPoints / math.max(1, doc.maxPoints);
-      final overall = combinedScore * 0.5 + points * 0.5;
+      final popularityScore = (downloadScore + likeScore) / 2;
+      final pointScore =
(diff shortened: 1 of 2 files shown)


app/lib/search/mem_index.dart
```diff
@@ -313,9 +313,10 @@ class InMemoryPackageIndex {
     _adjustedOverallScores = _documents.map((doc) {
       final downloadScore = doc.downloadScore ?? 0.0;
       final likeScore = doc.likeScore ?? 0.0;
-      final combinedScore = (downloadScore + likeScore) / 2;
-      final points = doc.grantedPoints / math.max(1, doc.maxPoints);
-      final overall = combinedScore * 0.5 + points * 0.5;
+      final popularityScore = (downloadScore + likeScore) / 2;
+      final pointScore =
+          math.sqrt(doc.grantedPoints / math.max(1, doc.maxPoints));
```
sigurdm, line 318: This looks wrong. `math.max(1, doc.maxPoints)` will always be one (unless you got exactly 0 points?

Should it not be

```
          math.max(1, math.sqrt(doc.grantedPoints / doc.maxPoints));
```

Which then simplifies to:

```suggestion
          math.sqrt(doc.grantedPoints / doc.maxPoints);
```

assuming `grantedPoints <= maxPoints`.

isoos: `math.max(1, doc.maxPoints)` prevents division by zero (in case we have data for whatever reason).

we could have an additional check that `grantedPoints/maxPoints` doesn't go over 1, adding that too.

sigurdm: Ah - I was reading `min` - sorry for my confusion :)

sigurdm (approved):

PR #8573, opened 2025-02-18
Additional style for exact package name highlights.

isoos:
https://github.com/dart-lang/pub-dev/issues/8518

image


sigurdm (approved):
Not sure this improves visibility enough.

But it is clearly an improvement, and I don't have much better ideas - let's try this for now.

PR #8575, opened 2025-02-19
Take screenshots from pub_integration tests (store locally).

isoos:
- Automatically takes separate screenshots for typical mobile, tablet and desktop widths and for dark and light themes (6 combinations total).
- Starting with 11 locations (66 screenshots), some full-page, some per-component screenshots.
- Comparison tool and github CI integrations in subsequent PRs.

### DIFF
--- a/pkg/pub_integration/lib/src/screenshot_utils.dart
+++ b/pkg/pub_integration/lib/src/screenshot_utils.dart
@@ -1,0 +1,102 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:puppeteer/puppeteer.dart';
+
+// Default screen with 16:10 ratio.
+final desktopDeviceViewport = DeviceViewport(width: 1280, height: 800);
+
+final _screenshotDir = Platform.environment['PUB_SCREENSHOT_DIR'];
+final _isScreenshotDirSet =
+    _screenshotDir != null && _screenshotDir!.isNotEmpty;
+
+// Set this variable to enable screenshot files to be updated with new takes.
+// The default is to throw an exception to prevent accidental overrides from
+// separate tests.
+final _allowScreeshotUpdates =
+    Platform.environment['PUB_SCREENSHOT_UPDATE'] == '1';
+
+// Note: The default values are the last, so we don't need reset
+//       the original values after taking the screenshots.
+final _themes = ['dark', 'light'];
+final _viewports = {
+  'mobile': DeviceViewport(width: 400, height: 800),
+  'tablet': DeviceViewport(width: 768, height: 1024),
+  'desktop': desktopDeviceViewport,
+};
+
+extension ScreenshotPageExt on Page {
+  Future<void> writeScreenshotToFile(String path) async {
+    await File(path).writeAsBytes(await screenshot());
+  }
+
+  /// Takes screenshots **if** `PUB_SCREENSHOT_DIR` environment variable is set.
+  ///
+  /// Iterates over viewports and themes, and generates screenshot files with the
+  /// following pattern:
+  /// - `PUB_SCREENSHOT_DIR/$prefix-desktop-dark.png`
+  /// - `PUB_SCREENSHOT_DIR/$prefix-desktop-light.png`
+  /// - `PUB_SCREENSHOT_DIR/$prefix-mobile-dark.png`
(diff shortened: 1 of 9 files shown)


pkg/pub_integration/lib/src/screenshot_utils.dart
```diff
@@ -0,0 +1,102 @@
...
+// separate tests.
+final _allowScreeshotUpdates =
+    Platform.environment['PUB_SCREENSHOT_UPDATE'] == '1';
+
+// Note: The default values are the last, so we don't need reset
+//       the original values after taking the screenshots.
+final _themes = ['dark', 'light'];
+final _viewports = {
+  'mobile': DeviceViewport(width: 400, height: 800),
+  'tablet': DeviceViewport(width: 768, height: 1024),
```
jonasfj, line 28: Might I propose we don't do `tablet` (just because of the number of screenshots)? Also don't those normally have much higher resolution (same with phones).

I'm assuming this is more about aspect ratio?

pkg/pub_integration/lib/src/screenshot_utils.dart
```diff
@@ -0,0 +1,102 @@
...
+
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:puppeteer/puppeteer.dart';
+
+// Default screen with 16:10 ratio.
+final desktopDeviceViewport = DeviceViewport(width: 1280, height: 800);
+
+final _screenshotDir = Platform.environment['PUB_SCREENSHOT_DIR'];
```
jonasfj, line 13: How about we just always store the screenshots in `.dart_tool/screenshots/`.

 * The `.dart_tool/` folder is already there.
 * The `.dart_tool/` folder is always in `.gitignore`.

jonasfj (approved):

pkg/pub_integration/lib/src/screenshot_utils.dart
jonasfj: Borderline crazy idea:

```
.dart_tool/
  screenshots/
    <revision>/
      image1.png
    current/
      image1.png
    comparison.html
```

Whenever, we create screenshots we:
 * Always write to `.dart_tool/screenshots/current/`
 * If `git status --porcelain` (meaning we have no changes), then:
   * We **also** write to `.dart_tool/screenshots/<revision>/`
 * We always create `.dart_tool/screenshots/comparison.html`
   * A simple dump HTML file showing screenshots side by side
   * Contents of `.dart_tool/screenshots/current/` on the right side
   * Contents of `<revision>` on the left side, where you pick `<revision>` from a dropdown
     * We create a dropdown at the top of the HTML file.
     * We run `git log` to find relevant revisions
        * We put relevant revisions in dropdown, if  `.dart_tool/screenshots/<revision>/` exists locally.
        * We order the dropdown by `git log`, and default to the most recent revision we have.

We could also use `gh api` to download artifacts from github:
https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#get-an-artifact

**Note**
 * An artifact on github is a zip folder.
 * It's not possible to download an artifact without being authenticated!
 * Artifacts can't be embedded in comments as images :(

If wanted to have before/after images on github, we'll probably have to do something borderline crazy like uploading the images to a separate branch or a GCS bucket.
Which probably involve more infrastructure than we care to stand up and maintain.

pkg/pub_integration/lib/src/screenshot_utils.dart
isoos: Why don't do tablet? I think it is a good way to check how the site behaves in narrower screen, and tbh. I don't think we will manually check tablet screens, but may catch if something degrades.

isoos: > I'm assuming this is more about aspect ratio?

Mostly the width.

pkg/pub_integration/lib/src/screenshot_utils.dart
isoos: Not a crazy idea, but let's not rush to it yet, and reach it maybe step-by-step. There is an immediate benefit with this + a comparison tool, and then we can add further automatization with github artifacts and/or per-revision storage...

pkg/pub_integration/lib/src/screenshot_utils.dart
jonasfj: I'm mostly thinking that we already have a LOT of screenshots, so I don't mind it, I'm only concerned that it's hard to get an overview.

isoos: Let's keep them for now, in the hope they are useful. If they are not, create too much noise, we may remove them or ignore them in most comparison reports.

sigurdm (approved):

PR #8579, opened 2025-02-21
Compare screenshots, generate report of differences.

isoos:
The tool and the report is very simple (calling imagemagick's `compare`), but it already revealed issues around the screenshot generation. Submitting it as-is, and we should discuss the issues outside of this PR.

### DIFF
--- a/pkg/pub_integration/tool/compare_screenshots.dart
+++ b/pkg/pub_integration/tool/compare_screenshots.dart
@@ -10,7 +10,7 @@ import 'package:path/path.dart' as p;
 /// Compares the screenshots from the previous and current test runs.
 /// Uses imagemagick for image processing.
 ///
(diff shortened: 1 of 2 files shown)


pkg/pub_integration/tool/compare_screenshots.dart
```diff
@@ -0,0 +1,125 @@
...
+
+import 'dart:io';
+
+import 'package:markdown/markdown.dart';
+import 'package:path/path.dart' as p;
+
+/// Compares the screenshots from the previous and current test runs.
+/// Uses imagemagick for image processing.
+///
+/// dart `<script.dart>` `<before-dir>` `<after-dir>` `<report-dir>`.
```
sigurdm, line 13:
```suggestion
/// `dart <script.dart> <before-dir> <after-dir> <report-dir>.
```

jonasfj (approved):

PR #8598, opened 2025-02-27
List uploaders in action package-info

sigurdm:


app/lib/admin/actions/package_info.dart
```diff
@@ -27,11 +29,18 @@ Loads and displays the package information.
       throw NotFoundException.resource(package);
     }
 
+    final uploaderIds = p.uploaders;
+    List<User>? uploaders;
+    if (uploaderIds != null) {
+      uploaders =
+          (await accountBackend.lookupUsersById(uploaderIds)).nonNulls.toList();
```
isoos, line 36: Note: `getEmailsOfUserIds` will use cache and may be faster.

isoos (approved):

app/lib/admin/actions/package_info.dart
sigurdm: Nice, switched

isoos (approved):

PR #8616, opened 2025-03-04
Migrate publisher-member-invite to action

sigurdm:
The 'publisher-members-list' is already covered by an existing action.


app/lib/admin/backend.dart
```diff
@@ -59,8 +58,6 @@ final Map<String, Tool> availableTools = {
   'notify-service': executeNotifyService,
   'package-publisher': executeSetPackagePublisher,
   'recent-uploaders': executeRecentUploaders,
-  'publisher-member': executePublisherMember,
```
isoos, line 62: Maybe `publisher-member-list` and `publisher-member-invite`?

isoos: Ah, sorry, wrong location.

app/lib/admin/actions/actions.dart
```diff
@@ -116,6 +117,7 @@ final class AdminAction {
     publisherCreate,
     publisherDelete,
     publisherInfo,
+    publisherInviteMember,
```
isoos, line 120: maybe we should call this `publisherMemberInvite` (with the name of the action also `publisher-member-invite`) so that it aligns with `publisherMembersList`?

PR #8662, opened 2025-03-21
Do not display HTML links when current search has no results and adding another filter wouldn't add more.

isoos:
- Similarly to #8661, reduces the link surface, so indiscriminate crawlers have fewer things to request from us.
- The complexity comes from `is:unlisted`, as it is not restricting, rather expanding the result set, so we need to pass the appropriate information down as well.
- The web app's dynamic update is also adjusted to work with the `<span>` as well as the `<a>` element.


isoos:
Note: pushed to staging for further testing.

isoos:
Offline discussion prompted to put a bit more effort into the index before removing these links, closing for now (#8670, #8671).

PR #8680, opened 2025-03-28
Built files are not built, only asserted when the app starts in a container.

### LINKED ISSUE #8673: Production server startup should fail if built resources are not present
Right now we check and build them, but that should happen in the Docker setup instead.

isoos:
Fixes #8673.


Dockerfile.app
```diff
@@ -10,6 +10,7 @@ RUN apt-get update && \
 # Let the pub server know that this is not a "typical" pub client but rather a bot.
 ENV PUB_ENVIRONMENT="bot.pub_dartlang_org.docker"
 ENV PUB_CACHE="/project/.pub-cache"
+ENV PUB_DEV_IN_CONTAINER="1"
```
sigurdm, line 13: Could you add a comment explaining what this does?

jonasfj (approved):

isoos:
Closing in favor of #8729

Issue #8671, opened 2025-03-25
Package search index improvements (tracking issue)


isoos:
- [x] Detect high search loads (high latencies) and dynamically turn off API docs or readme search.
- [x] Investigate if predicate-only search can use a separate bitmap index.
- [x] Make sure SDK index is not on a blocking path (part of #8670), e.g. it doesn't take longer to serve that part of the results.
- [x] Reduce List-building when the result may not need to be used (using Iterator instead)
- [x] Investigate if top-k sort algorithms could be used instead of sorting everything and selecting the top k items.

PR #8684, opened 2025-04-01
Use bit arrays for predicate matching in search.

isoos:
- #8671
- There are two further potential improvements, but I would implement them in a subsequent PRs: (a) reusing the bitarray instance, and (b) not creating/using a `PackageScore` instance when the `BitArray` would suffice.
- Measured around 5-10% improvement for predicate-only search, and around 1-4% for more complex text matching + predicates.

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -10,6 +10,7 @@ import 'package:collection/collection.dart';
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 import 'package:pub_dev/service/topics/models.dart';
+import 'package:pub_dev/third_party/bit_array/bit_array.dart';
 
 import '../shared/utils.dart' show boundedList;
 import 'models.dart';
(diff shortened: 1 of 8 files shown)


app/lib/search/mem_index.dart
```diff
@@ -208,22 +199,29 @@ class InMemoryPackageIndex {
...
     final updatedDuration = query.parsedQuery.updatedDuration;
     if (updatedDuration != null && updatedDuration > Duration.zero) {
       final now = clock.now();
-      packageScores.removeWhere(
-          (i, _) => now.difference(_documents[i].updated) > updatedDuration);
+      packages.clearWhere(
+          (i) => now.difference(_documents[i].updated) > updatedDuration);
+    }
+
+    // TODO: find a better way to handle predicate-only filtering and scoring
```
sigurdm, line 214: Should we open an issue for this?

app/pubspec.yaml
```diff
@@ -48,6 +48,7 @@ dependencies:
   # pana version to be pinned
   pana: '0.22.20'
   # 3rd-party packages with pinned versions
+  bit_array: 2.3.0
```
sigurdm, line 51: @jonasfj are we ok with this dependency?

isoos: Note: the 80% of `package:bit_array` is probably not needed for `pub-dev`, and I have no issue with vendoring it.

Also: the package is using 32-bit ints for cross-compatibility in browsers, and it is possible we should benchmark using 64-bit ints for the server-only use case.

isoos:
I've modified and benchmarked the 64-bit version of `BitArray` and couldn't point out much different performance to the current 32-bit implementation. With that in mind, I've vendored the library as-is, with minor updates (e.g. removing the `part of` and adding imports). I'll also update `package:bit_array` to reflect these changes, so next time the update will be straightforward file copy.

Should we also vendor some of the tests too?

Issue #8688, opened 2025-04-01
"exact match" package name results are easy to ignore


eseidel:
Very similar to https://github.com/dart-lang/pub-dev/issues/8518.

Searching for "redis" the "redis" package is the 10th highest result.  Maybe that's intentional?  Maybe that's because it hasn't been updated in 17 months?

But also the first two matches don't have anything to do with "redis" whereas there are other results that do.  You don't get to a package that really has anything to do with "redis" until the 8th match. 🤣   Maybe this is just that "search is hard, especially over small corpuses" and we should just close this 🤷 

There is an attempt to link to it from the "exact match", but it's very easy to glance over the "exact match" line.

<img width="1166" alt="Image" src="https://github.com/user-attachments/assets/5f0e40f1-d665-428c-9fd9-d80c94d1bcd4" />

isoos:
Hm, couple of quick observations:

`neat_cache` (3rd result) has `redis` in the description and internally it does have a redis client - though it is not exposed to the public API, not sure if that is intended /cc @jonasfj 

`shorebird_redis_client` will rank higher if they get more use.

The query seems to be reduced to `redi` with a clueless de-pluralization, and matched with `credit` in the name or description of the other results ranking high. The de-pluralized version should probably rank lower if other versions are also present (or we should recognize words and brands).

isoos:
I think the difference between the two results can be highlighted when you enable exact phrase match:
https://pub.dev/packages?q=%22redis%22

It kind of gives away that we are matching much more than we should with the non-exact phrase version.

isoos:
Also: I don't think that users should use the exact phrase match for such results, the search index should do a better job. (And again, thanks for reporting it!)

PR #8692, opened 2025-04-02
Slightly lower score for depluralized matches.

isoos:
#8688

- fixes the oversight that the substring check should be done on the lowercased value.
- does not really solve the de-pluralization problem space, but slightly reduces the miss-attribution ranks


isoos:
Closing as we've decided to remove the depluralization code instead. #8696

Issue #8670, opened 2025-03-25
SDK search index improvements (tracking issue)


isoos:
- [x] Index content should be downloaded at the time we create the docker image, reduces the isolate startup latencies.
- [x] We should consider using a separate isolate for the SDK index, can be a fixed one after the content doesn't need a reload.
- [ ] We should track latencies separately, optionally skipping the index if otherwise latencies are high.
- [x] Consider merging the two SDK index and search them only a single time instead of sequentially.

isoos:
I think this can be closed now.

PR #8699, opened 2025-04-04
Download SDK index.jsons at docker build.

isoos:
- #8670
- The cache location is reused from the test, but instead of hand-provided filename, I've opted to use the URL as the template.
- The TTL is there only for the test to re-download the files periodically, so that we catch regressions. (TBD: should we download the beta channel instead in tests?)
- The docker build will download the files, or will fail during the build. However, if #8680 lands, we could also assert the presence of the files and not fall back to downloading them.

### DIFF
--- a/Dockerfile.app
+++ b/Dockerfile.app
@@ -33,6 +33,7 @@ WORKDIR /project/app
 RUN dart /project/tool/pub_get_offline.dart /project/app
 
 RUN /project/tool/setup-webp.sh /usr/local/bin
+RUN /project/tool/download-sdk-index-jsons.sh
 
 # Clear out any arguments the base images might have set
 CMD []
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -17,6 +17,7 @@ import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 // ignore: implementation_imports
 import 'package:pana/src/dartdoc/pub_dartdoc_data.dart';
+import 'package:path/path.dart' as p;
 import 'package:pool/pool.dart';
 import 'package:pub_dev/shared/monitoring.dart';
 import 'package:retry/retry.dart';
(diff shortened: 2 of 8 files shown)


Dockerfile.app
```diff
@@ -33,6 +33,7 @@ WORKDIR /project/app
 RUN dart /project/tool/pub_get_offline.dart /project/app
 
 RUN /project/tool/setup-webp.sh /usr/local/bin
+RUN /project/tool/download-sdk-index-jsons.sh
```
jonasfj, line 36: I think this file is missing :rofl:

app/lib/search/backend.dart
```diff
@@ -427,17 +428,40 @@ class SearchBackend {
...
+  /// non-characters with a single dash `-`, like:
+  /// - `https-api.dart.dev-stable-latest-index.json`
+  Future<String> loadOrFetchSdkIndexJsonAsString(
+    Uri uri, {
+    @visibleForTesting Duration? ttl,
   }) async {
-    final uri = baseUri.resolve(relativePath);
+    final fileName = uri.toString().replaceAll(RegExp(r'[^a-z0-9\.]+'), '-');
+    final file = File(p.join('.dart_tool', 'pub-search-data', fileName));
+    if (await file.exists()) {
```
jonasfj, line 444: It's always safer to read and then handle the `IOException` if the file doesn't exist.

jonasfj (approved):

Dockerfile.app
isoos: Oh. Nice catch. Fixed, PTAL.

app/lib/search/backend.dart
isoos: Right. As the same thing can be said of the `_http.get` that this method is also using, I'll address that try-catch in one level up, in a follow-up PR.

isoos: Note: the SDK index builder methods (both for Dart and Flutter) already have a try-catch block, handling exceptions.

PR #8702, opened 2025-04-07
Export /feed.atom to the exported bucket.

isoos:


isoos:
Note: for some reason the fake-time based exporter test fails, and I couldn't figure out the reason yet. I think the rest of the change is worth for a review though.

isoos:
Note: I've found and fixed the test issue: we were using two different instance of `ApiExporter` in two different scope. Fixing that by resetting the existing one at the beginning of the tests.

app/lib/package/backend.dart
```diff
@@ -1271,8 +1271,10 @@ class PackageBackend {
         if (activeConfiguration.isPublishedEmailNotificationEnabled)
           emailBackend.trySendOutgoingEmail(outgoingEmail),
         taskBackend.trackPackage(newVersion.package, updateDependents: true),
-        if (apiExporter != null)
+        if (apiExporter != null) ...[
           apiExporter!.synchronizePackage(newVersion.package),
+          apiExporter!.synchornizeAllPackagesAtomFeed(),
```
sigurdm, line 1276:
```suggestion
          apiExporter!.synchronizeAllPackagesAtomFeed(),
```

sigurdm (approved):

PR #8705, opened 2025-04-14
RFC: Options to SQL-migration.

isoos:
- `sql_datastore.dart` shows how it would look like if we were to keep the `DatastoreDB` as-is, but switch out the underlying implementation.
- `task/backend.dart` shows how it would look like to refactor all current Datastore access to narrowly defined access methods, and later on the SQL migration could focus only on these methods, without their associated application logic.


isoos:
Closing this for now, per our discussion we shall go ahead with the separate methods, but with a slightly different kind of class structure + extensions on DatastoreDb.

Issue #5267, opened 2021-11-15
Suggestion: Please add an RSS/Atom feed for each package's changelog


jonmountjoy:
For example,  https://pub.dev/packages/flutter_local_notifications/changelog

It would be fab if there was a feed, which updated with the latest change.  That way I can pay attention to all changes to all packages I wish.

I realise "latest change" is difficult as it's probably a single file for now - so just the latest, if it's changed, would do.

As a bonus, provide a single feed as part of my profile that aggregates all packages I subscribed to :-)

isoos:
a related design suggestion from @jtmuller5: https://twitter.com/Mullr33/status/1477286273144524804

Issue #1378, opened 2018-06-17
Get notified by mail on package updates


cbenhagen:
It would be nice if a user could star a package to get notified by mail on updates.

mit-mit:
We don't have a notion of users on the site, so that would be a fairly big change.

Idea: do this based on RSS feeds

sigurdm:
We now have users, and likes.
We could add an option to get mails if any of these are updated.

We could also create one rss feed per package - that would make it easy to subscribe to each package one is interested in.

We now have dependabot support that notifies to new versions relevant for a project.

PR #8709, opened 2025-04-15
Link to the RSS/atom feed from the versions page.

isoos:
- #1378 and #5267 

image
image


sigurdm:
how about
image

isoos:
> how about

I'd probably use a full sentence, regardless if it is at the top or the bottom of the page. I'm fine if it goes to the bottom though...

sigurdm:
I'm ok with a full sentence. Can we have the feed logo?

isoos:
Updated the PR with a new logo and layout. Note: I've also considered the inline icon, but as the following screenshot shows, it is not that prominent, and it would need additional positioning and fine-tuning of the margins:

image

isoos:
Closing in favor of #8731.

PR #8711, opened 2025-04-18
Use new HTTP retry logic when fetching SDK's index.json + refactor

isoos:
- Alternative implementation direction for retrying HTTP APIs (see suboptimal attempt in #8558).
- Reduces the `SearchBackend.loadOrFetchSdkIndexJsonAsString` dependence on scope-specific state, moving it out of the backend class, which will make it easier to move the SDK index to a separate isolate (#8670).


isoos:
I think we should land this this week. @sigurdm @jonasfj

pkg/_pub_shared/lib/utils/http.dart
```diff
@@ -35,3 +37,45 @@ http.Client httpRetryClient({
...
+/// The HTTP GET and the [responseFn] callback is retried on the transient
+/// network errors.
+Future<K> httpGetWithRetry<K>(
+  Uri uri, {
+  required FutureOr<K> Function(http.Response response) responseFn,
+  int maxAttempts = 3,
+}) async {
+  return await retry(
+    () async {
+      final client = http.Client();
```
sigurdm, line 53: Perhaps we pass in a client instead of creating a new client for every url to fetch?

isoos: Planning to do that in subsequent PRs (when migrating more functions to use this).

The practical difference in this specific case was negligible: we would only call the method when the index.json is not fetched at container creation time (or in local tests).

sigurdm (approved):

PR #8719, opened 2025-04-29
Upgrade Dart analysis SDK

isoos:


sigurdm (approved):

isoos:
Note: reverted the dependency upgrade, as it involves a lot of deprecation in package:analyzer and I'd rather do that separately.

PR #8725, opened 2025-05-01
Create a working copy of the uploaded blob before processing it.

isoos:


app/lib/package/backend.dart
```diff
@@ -891,28 +893,48 @@ class PackageBackend {
...
+      final workGuid = createUuid();
+      final workObjectName = '${tmpObjectName(workGuid)}-$uploadGuid';
+      try {
+        await _storage.copyObjectWithRetry(
+          _incomingBucket.absoluteObjectName(uploadObjectName),
+          _incomingBucket.absoluteObjectName(workObjectName),
+        );
+      } catch (e) {
+        throw InvalidInputException(
+            'Failed to copy uploaded file (uuid:$uploadGuid).');
```
sigurdm, line 925: Perhaps add the error 
```suggestion
            'Failed to copy uploaded file (uuid:$uploadGuid): $e.');
```

isoos: Hm, it may be worth logging, but not exposing to the end user.

sigurdm: Ah yes, sorry, the user should definitely **not** get this error 😅.

sigurdm (approved):

PR #8726, opened 2025-05-01
Refactor: use await for loop instead of Stream.pipe.

isoos:


app/lib/package/backend.dart
```diff
@@ -1825,21 +1825,20 @@ Future _saveTarballToFS(Stream<List<int>> data, String filename) async {
...
+      if (receivedBytes <= UploadSignerService.maxUploadSize) {
+        sink.add(chunk);
+      } else {
+        await sink.close();
+        throw PackageRejectedException.archiveTooLarge(
+            UploadSignerService.maxUploadSize);
+      }
+    }
+    await sink.flush();
+    await sink.close();
```
sigurdm, line 1841: Should the close be in a finally block?

sigurdm (approved):

PR #8732, opened 2025-05-02
Use new HTTP GET retry logic in SearchClient.

isoos:
This extends the retry utility method with timeout, headers and caller-provided client. However, on a retry, this client is not recreated (whether and how to express this is to be done in a subsequent PR).

### DIFF
--- a/pkg/_pub_shared/lib/utils/http.dart
+++ b/pkg/_pub_shared/lib/utils/http.dart
@@ -47,23 +47,41 @@ Future<K> httpGetWithRetry<K>(
   Uri uri, {
   required FutureOr<K> Function(http.Response response) responseFn,
   int maxAttempts = 3,
+
+  /// The HTTP client to use.
+  ///
+  /// Note: when the client is specified, the inner loop will not create a new client object on retries.
+  http.Client? client,
+  Map<String, String>? headers,
+
+  /// Per-request time amount that will be applied on the overall HTTP request.
+  Duration? timeout,
+
+  /// Additional retry conditions (on top of the default ones).
+  bool Function(Exception e)? retryIf,
 }) async {
   return await retry(
     () async {
-      final client = http.Client();
+      final closeClient = client == null;
+      final effectiveClient = client ?? http.Client();
       try {
-        final rs = await client.get(uri);
+        var f = effectiveClient.get(uri, headers: headers);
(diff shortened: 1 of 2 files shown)


pkg/_pub_shared/lib/utils/http.dart
```diff
@@ -47,23 +47,41 @@ Future<K> httpGetWithRetry<K>(
...
   int maxAttempts = 3,
+
+  /// The HTTP client to use.
+  ///
+  /// Note: when the client is specified, the inner loop will not create a new client object on retries.
+  http.Client? client,
+  Map<String, String>? headers,
+
+  /// Per-request time amount that will be applied on the overall HTTP request.
+  Duration? timeout,
```
sigurdm, line 58: I think you  (in principle at least) need to put dartdoc documentation of these in the main function dartdoc.

Perhaps rename this to clarify that it is per request.
```suggestion
  Duration? perRequestTimeout,
```

pkg/_pub_shared/lib/utils/http.dart
```diff
@@ -47,23 +47,41 @@ Future<K> httpGetWithRetry<K>(
   Uri uri, {
   required FutureOr<K> Function(http.Response response) responseFn,
   int maxAttempts = 3,
+
+  /// The HTTP client to use.
+  ///
+  /// Note: when the client is specified, the inner loop will not create a new client object on retries.
```
sigurdm, line 53: I would formulate this the other way around
```suggestion
  /// Note: when the client is not specified, the inner loop will create a new [Client] object on each retry attempt.
```

pkg/_pub_shared/lib/utils/http.dart
```diff
@@ -47,23 +47,41 @@ Future<K> httpGetWithRetry<K>(
...
+        if (timeout != null) {
+          f = f.timeout(timeout);
+        }
+        final rs = await f;
         if (rs.statusCode == 200) {
           return responseFn(rs);
         }
-        throw http.ClientException(
-            'Unexpected status code for $uri: ${rs.statusCode}.');
+        throw UnexpectedStatusException(rs.statusCode, uri);
```
sigurdm, line 76: Perhaps document that the `retry` function can check for this.

sigurdm (approved):

Issue #8652, opened 2025-03-19
Investigating the work needed for Datastore to SQL migration


isoos:

isoos:
By running the `app/test` tests, I've collected the entity types and attached the types that they were together in the write or delete part of a transaction.

```
AuditLogRecord
AuditLogRecord:Consent
AuditLogRecord:OutgoingEmail,Package
AuditLogRecord:OutgoingEmail,Package,PackageVersion,PackageVersionAsset,PackageVersionInfo
AuditLogRecord:OutgoingEmail,Package,PackageVersion,PackageVersionAsset,PackageVersionInfo,ReservedPackage
AuditLogRecord:Package
AuditLogRecord:Package,PackageVersion
AuditLogRecord:Publisher
AuditLogRecord:Publisher,PublisherMember
Consent
Consent:AuditLogRecord
Consent:OutgoingEmail
DownloadCounts
GlobalLockState
Like:Package,User
Like:User
ModeratedPackage
ModeratedPackage:Package
ModerationCase
ModerationCase:Package
ModerationCase:Package,PackageVersion
ModerationCase:Publisher
ModerationCase:User
NeatTaskStatus
OAuthUserID:User
OutgoingEmail
OutgoingEmail:AuditLogRecord,Package
OutgoingEmail:AuditLogRecord,Package,PackageVersion,PackageVersionAsset,PackageVersionInfo
OutgoingEmail:AuditLogRecord,Package,PackageVersion,PackageVersionAsset,PackageVersionInfo,ReservedPackage
OutgoingEmail:Consent
Package
Package:AuditLogRecord
Package:AuditLogRecord,OutgoingEmail
Package:AuditLogRecord,OutgoingEmail,PackageVersion,PackageVersionAsset,PackageVersionInfo
Package:AuditLogRecord,OutgoingEmail,PackageVersion,PackageVersionAsset,PackageVersionInfo,ReservedPackage
Package:AuditLogRecord,PackageVersion
Package:Like,User
Package:ModeratedPackage
Package:ModerationCase
Package:ModerationCase,PackageVersion
Package:PackageVersion
Package:PackageVersion,PackageVersionAsset,PackageVersionInfo
Package:SecurityAdvisory
PackageState
PackageVersion:AuditLogRecord,OutgoingEmail,Package,PackageVersionAsset,PackageVersionInfo
[...]

PR #8733, opened 2025-05-05
Extract most UserSession DatastoreDB access methods.

isoos:
- #8652
- Note: `createOrUpdateClientSession` could be also moved to the data access parts, but it felt it has more direct and transactional logic and we should rather keep it there. However, during the SQL migration we will migrate it alongside the data access methods.

### DIFF
--- a/app/lib/account/backend.dart
+++ b/app/lib/account/backend.dart
@@ -428,7 +428,8 @@ class AccountBackend {
     // try to update old session first
     if (oldSession != null) {
       final rs = await withRetryTransaction(_db, (tx) async {
-        final session = await tx.lookupOrNull<UserSession>(oldSession.key);
+        final session =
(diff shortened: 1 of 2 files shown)


app/lib/account/backend.dart
```diff
@@ -682,3 +659,63 @@ Future<void> purgeAccountCache({
...
+      await expire(session.sessionId);
+    }
+  }
+
+  /// Removes the session data from the Datastore and from cache.
+  Future<void> expire(String sessionId) async {
+    final key = _db.emptyKey.append(UserSession, id: sessionId);
+    try {
+      await _db.commit(deletes: [key]);
+    } catch (_) {
```
sigurdm, line 698: Can we get away with catching only Exceptions?

isoos: Updated.

sigurdm (approved):

PR #8736, opened 2025-05-06
Do not export RSS feed for soft-removed packages.

isoos:


isoos:
Closing, as #8739 is preferred.

PR #8738, opened 2025-05-07
HTTP Client that can be renewed / reset after a retry.

isoos:


isoos:
Note: I'm unsure if this is needed, but it seems like a good idea to periodically renew connections, and the retry event may be just as good trigger for it as any.

isoos:
Discussed yesterday: we don't need to renew the client, it should be done inside `package:http`.

PR #8740, opened 2025-05-07
Add function for computing package trend

szakarias:


app/lib/service/download_counts/package_trends.dart
```diff
@@ -0,0 +1,96 @@
...
+/// normalized by the average daily downloads, allowing for comparison across
+/// packages of different popularity. For example, a slope of +10 downloads/day
+/// is more significant for a package with 100 average daily downloads (10%
+/// relative growth) than for a package with 10000 average daily downloads (0.1%
+/// relative growth).
+double computeRelativeGrowthRate(List<int> totalDownloads) {
+  if (totalDownloads.length < analysisWindowDays) {
+    // insufficient data points.
+    return -double.maxFinite;
+  }
```
jonasfj, line 35: Why?

jonasfj (approved):

app/lib/service/download_counts/package_trends.dart
szakarias: removed.

PR #8741, opened 2025-05-07
Handle 0 downloads in growth function

szakarias:


app/lib/service/download_counts/package_trends.dart
```diff
@@ -35,6 +35,10 @@ double computeRelativeGrowthRate(List<int> totalDownloads) {
       recentDownloads.reduce((prev, element) => prev + element) /
           recentDownloads.length;
 
+  if (averageRecentDownloads == 0) {
```
jonasfj, line 38: < 1 might also be worth considering...

Packages with average downloads less than one shouldn't have their growthrate increased..

jonasfj (approved):

app/lib/service/download_counts/package_trends.dart
szakarias: I'll keep it in mind but I'll keep it as it is for now. This is the mathematically correct calculation, and hence keeps the abstraction of the function clear.  On the caller site we will handle packages with few downloads.

PR #8743, opened 2025-05-08
Select only a single name match in search.

isoos:
- While it keeps the list of results (may be updated in a follow-up PR), at most a single name match is returned.
- Removes per-word name matches, only full-query name match is considered.
- In the rare case of potential name collapse conflicts, it selects the most downloaded one.

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -230,20 +230,26 @@ class InMemoryPackageIndex {
 
     String? bestNameMatch;
     if (parsedQueryText != null) {
-      final matches = _packageNameIndex.lookupMatchingNames(parsedQueryText);
-      if (matches != null && matches.isNotEmpty) {
-        bestNameMatch = matches.length == 1
-            ? matches.single
-            :
-            // Note: to keep it simple, we select the most downloaded one from competing matches.
-            matches.reduce((a, b) {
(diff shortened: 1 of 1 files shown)


app/lib/search/mem_index.dart
```diff
@@ -225,11 +225,27 @@ class InMemoryPackageIndex {
...
       packageScores,
       parsedQueryText,
-      includeNameMatches: (query.offset ?? 0) == 0,
       textMatchExtent: query.textMatchExtent ?? TextMatchExtent.api,
     );
 
-    final nameMatches = textResults?.nameMatches;
+    String? bestNameMatch;
+    if (parsedQueryText != null) {
+      final matches = _packageNameIndex.lookupMatchingNames(parsedQueryText);
```
sigurdm, line 233: Not sure if this is already how it works, but can we make it such that if there are multiple matches, and one is an exact match, then we use it regardless of download-count?

isoos: Fair point, fixed.
Note: I think the multiple matches are mostly theoretical, e.g. `a_bc` and `ab_c` would be matched for the query `abc`, however, `pub.dev` wouldn't allow such package names in the first place (unless they were exempted), and then we would fall back to the download counts.

sigurdm (approved):

PR #8744, opened 2025-05-09
Add function and task for computing and uploading package trends

szakarias:


sigurdm (approved):

app/lib/service/download_counts/computations.dart
```diff
@@ -6,15 +6,45 @@ import 'dart:math';
...
+  await uploadTrendScores(trendScores);
+}
+
+Future<Map<String, double>> computeTrend() async {
+  final res = <String, double>{};
+
+  await for (final pkg in packageBackend.allPackages()) {
+    final name = pkg.name!;
+    final downloads =
+        (await downloadCountsBackend.lookupDownloadCountData(name))
```
jonasfj, line 31: You could consider a retry around this part, so that `computeTrend` doesn't fail just because a single lookup here failed.

Indeed, you could also consider ignore errors here.. or maybe ignore up-to 50 errors, so if there is less than 50 errors you just ignore them (log them), but don't fail the computation.

Just an idea, not sure it's smart.

jonasfj (approved):

app/lib/service/download_counts/computations.dart
jonasfj: or mabye `lookupDownloadCountData` already does retries

PR #8745, opened 2025-05-09
Promote name matches to top position + add name match badge after the package name.

isoos:
- Previous name match block is removed.
- The best matched name is promoted to the first spot, in certain conditions (no other active filters or sorting, but score is kept in API).
- The "name match" badge is added after the name:

image


isoos:
Note: staging deployment is pending.

isoos:
PTAL. The latest version is on staging.

szakarias (approved):
LGTM

Issue #8752, opened 2025-05-13
Exported API for `/api/packages/<package>/*`


jonasfj:
Let's do exported API for all `GET` requests for:
```
/api/packages/<package>/likes
/api/packages/<package>/metrics
/api/packages/<package>/options
/api/packages/<package>/publisher
/api/packages/<package>/score
/api/packages/<package>/versions/<version>/options
/api/packages/<package>/versions/<version>/score
```

PR #8758, opened 2025-05-15
Export more package API endpoints.

isoos:
- exporting `/likes`, `/options`, `/publisher` and `/score` endpoints 
- extracted `getPackageOptions` method + fixed the missing `replacedBy` field
- extracted `getVersionScore` method
- updated triggers on cache invalidation and API export
- #8752


isoos:
As discussed, I've revisited this PR, and I think we can merge it, and will do a follow-up with the proposal for better cache invalidation.

PR #8783, opened 2025-05-22
Use DateTime.utc() for sanity

jonasfj:
I'm not entirely sure why this fails when using local time, but not using local time is probably a good idea.

I suspect that if we wanted to make the logic more robust, we should convert `DateTime` objects to milliseconds, and do a floating point division with `24 * 60 * 60 * 1000` (ms per day), and then round to nearest count of days.

I suspect the problem with `Duration.inDays` is that it doesn't round up, it'll always round down.

That may be a better fix.


isoos (approved):

szakarias (approved):
LGTM

PR #8784, opened 2025-05-22
Initial migration of task backend to typed_sql

jonasfj:


app/lib/task/backend.dart
```diff
@@ -349,81 +344,84 @@ class TaskBackend {
...
+    if (package == null) {
+      return; // assume package was deleted!
+    }
 
-      // Update the timestamp for when the last version was published.
-      // This is used if we need to update dependents.
-      lastVersionCreated = packageVersions.map((pv) => pv.created!).max;
+    if (package.isNotVisible) {
+      await db.tasks.delete(runtimeVersion, packageName).execute();
+      return; // TODO: Purge caches
```
isoos, line 381: Why not call `_purgeCache` (and remove the TODO) or, if needed, specify which cache need to be purged.

app/lib/task/backend.dart
```diff
@@ -534,52 +549,42 @@ class TaskBackend {
...
+                ))
+            .execute();
+      } catch (e, st) {
+        _log.warning(
+          'failed to propagate lastDependencyChanged to "$package"',
+          e,
+          st,
+        );
+      }
+    });
```
isoos, line 587: I like this very much, but a nit: here and maybe at other places of the `parallelForEach` I would have preferred to indent it under/below the `parallelForEach` method. Would an extra comma here achieve that?

isoos (changes requested):
Otherwise I like really the new sql-way :)
I think some files like `database/model.dart` is missing. Also needs merge fixes.

app/lib/task/backend.dart
jonasfj: No, an extra comma would just result in:
```dart
    await db.tasks
        .join(db.taskDependencies)
        .usingTask()
        .where((task, dependency) =>
            dependency.package.equalsValue(package) &
            dependency.runtimeVersion.equalsValue(runtimeVersion) &
            task.lastDependencyChanged.isBeforeValue(publishedAt))
        .select((task, dependency) => (task.package, task.state))
        .stream()
        .parallelForEach(
      20,
      (row) async {
        final (package, state) = row;
        try {
          await db.tasks
              .byKey(runtimeVersion, package)
              .update((_, set) => set(
                    lastDependencyChanged: publishedAt.asExpr,
                    pendingAt: derivePendingAt(
                      state: state,
                      lastDependencyChanged: publishedAt,
                    ).asExpr,
                  ))
              .execute();
        } catch (e, st) {
          _log.warning(
            'failed to propagate lastDependencyChanged to "$package"',
            e,
            st,
          );
        }
      },
    );
```

It it perhaps tempting to do intermediate `q` variable:
```dart
    final q = db.tasks
        .join(db.taskDependencies)
        .usingTask()
        .where((task, dependency) =>
            dependency.package.equalsValue(package) &
            dependency.runtimeVersion.equalsValue(runtimeVersion) &
            task.lastDependencyChanged.isBeforeValue(publishedAt))
        .select((task, dependency) => (task.package, task.state));

    await q.stream().parallelForEach(20, (row) async {
      final (package, state) = row;
      try {
        await db.tasks
            .byKey(runtimeVersion, package)
            .update((_, set) => set(
                  lastDependencyChanged: publishedAt.asExpr,
                  pendingAt: derivePendingAt(
                    state: state,
                    lastDependencyChanged: publishedAt,
                  ).asExpr,
                ))
            .execute();
      } catch (e, st) {
        _log.warning(
          'failed to propagate lastDependencyChanged to "$package"',
          e,
          st,
        );
      }
    });
```

But when there only is one, and the parallelForEach block isnt' to long, I think it's fine to do this.

app/lib/task/backend.dart
jonasfj: I think this was a change I missed during migration.

app/lib/task/backend.dart
```diff
@@ -1216,20 +1243,12 @@ List<Version> _versionsToTrack(
   }.nonNulls.where(visibleVersions.contains).toList();
 }
 
-List<String> _updatedDependencies(
-  List<String>? dependencies,
-  List<String>? discoveredDependencies, {
+List<String> _newDependencies(
```
isoos, line 1246: Let's document what this method does.

Aside: This may be in another step, but with SQL, we should probably check if we can get away with simpler behavior, e.g. not reading and processing everything, or, maybe updating the dependent package's timestamp through a query.

app/lib/database/model.task.dart
```diff
@@ -0,0 +1,65 @@
...
+
+@PrimaryKey(['runtimeVersion', 'package', 'dependency'])
+@ForeignKey(
+  ['runtimeVersion', 'package'],
+  table: 'tasks',
+  fields: ['runtimeVersion', 'package'],
+  name: 'task',
+  as: 'dependencies',
+)
+abstract final class TaskDependency extends Row {
```
isoos, line 33: Doesn't this need `@PrimaryKey[]`?

Also, I'm wondering if the order of the keys should start with the dependency, maybe it could have performance benefits if/when later we update the dependent package's flag/timestamp with a query.

app/lib/database/model.task.dart
```diff
@@ -0,0 +1,65 @@
...
+  String get package;
+
+  /// Name of a package that is either a direct or transitive dependency of
+  /// [package].
+  String get dependency;
+}
+
+@immutable
+@JsonSerializable()
+final class TaskState implements CustomDataType<String> {
```
isoos, line 44: Nit: How far would it be to implement `CustomDataType<Json>` and using `JSONB` under the hood? I think it is so wasteful to keep this data as string instead of `JSONB`...

app/lib/task/backend.dart
```diff
@@ -488,9 +499,12 @@ class TaskBackend {
...
   }
 
-  /// Garbage collect [PackageState] and results from old runtimeVersions.
+  /// Garbage collect [Task] and results from old runtimeVersions.
   Future<void> garbageCollect() async {
-    await _db.tasks.deleteBeforeGcRuntime();
+    await db.tasks
+        .where((task) => task.runtimeVersion < gcBeforeRuntimeVersion.asExpr)
+        .delete()
+        .execute();
```
isoos, line 507: Nit: wouldn't the `delete()` call before the `where()` make more sense? It kind of make sense for the `select()` projection to be just before e.g. `stream()` lie a few lines below, but for `delete()`, I could imagine it above it. This feels like the `delete()` and `execute()` calls should be a single method call, but maybe that fades more with further familiarity of the library use.

isoos (commented):
Still looking good.

app/lib/shared/env_config.dart
```diff
@@ -37,6 +37,11 @@ class _EnvConfig {
   /// Youtube API key to use (skips Datastore secret).
   late final youtubeApiKey = Platform.environment['YOUTUBE_API_KEY'];
 
+  /// True, if the process is running in a CI environment.
+  late final isContinuesIntegration = !['false', '0', ''].contains(
```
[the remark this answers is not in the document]
isoos: though, I think we should rather have the explicit `true` (or maybe `1`) check instead of negating the negative.

isoos:
Closing in favor of #9126 - very similar as of now, without the merge conflicts :)

Issue #8730, opened 2025-05-02
Package deletion admin action should just remove packages from public bucket


sigurdm:
A background process should remove packages from the canonical bucket after they have been deleted for more than 60 days.

sigurdm:
(after the 60 days the current behavior of creating a tombstone should still apply)

PR #8788, opened 2025-05-23
Updated admin removal for package versions: using the flag on the action

isoos:
- part of #8730
- only changes the deletion of a version and not the package (yet, to be done in a follow-up PR)
- follows the same admin action pattern as the moderate actions
- reuses most of the update code with moderating and deleting a package version
- updated tests with only a few relevant bits, we won't need to duplicate everything, as the main code will be already tested with the moderation action
- does the actual deletion after 2 months

### DIFF
--- a/app/lib/admin/actions/package_version_delete.dart
+++ b/app/lib/admin/actions/package_version_delete.dart
@@ -4,47 +4,38 @@
 
 import '../../account/backend.dart';
 import '../../shared/configuration.dart';
-import '../backend.dart';
 import 'actions.dart';
+import 'moderate_package_versions.dart';
 
 final packageVersionDelete = AdminAction(
-    name: 'package-version-delete',
-    options: {
-      'package': 'name of package to delete',
-      'version': 'version of package',
-    },
-    summary: 'Deletes package <package> version <version>.',
-    description: '''
-Deletes package <package> version <version>.
-
-Deletes all associated resources:
-
-* PackageVersions
-* PackageVersionAsset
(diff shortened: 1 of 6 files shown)


isoos:
Note: also updated to merge/reuse the deletion of a package version.

app/lib/admin/actions/package_version_delete.dart
```diff
@@ -4,47 +4,38 @@
...
-* PackageVersionAsset
-* archives (might be retrievable from backup)
-
-The package version will be "tombstoned" and same version cannot be published
-later.
+  name: 'package-version-delete',
+  summary:
+      'Set the admin-deleted flag on a package version (making it not visible).',
+  description: '''
+Set the admin-deleted flag on a package version (updating the flag and the timestamp).
```
sigurdm:
```suggestion
Set the admin-deleted flag on a package version (updating the flag and the timestamp). After 2 months it will be fully deleted.
```

app/lib/admin/backend.dart
```diff
@@ -802,6 +804,7 @@ class AdminBackend {
     }
 
     // delete package versions
+    // TODO: merge this with the removePackageVersion method.
```
sigurdm: Consider creating an issue for this TODO

isoos: Ah, that's actually a TODO that was implemented in the follow-up commit, will remove the line.

sigurdm (approved):

PR #8797, opened 2025-05-29
Implement bitarray pool in package search index.

isoos:
- Provides a consistent 1-2% latency improvement.
- Note: I've also investigated how to reduce the use of `IndexedScore`, but seems to be a bit more involved, postponing it for now (with a follow-up PR that may prepare it).

### DIFF
--- a/app/lib/search/token_index.dart
+++ b/app/lib/search/token_index.dart
@@ -204,9 +204,9 @@ class BitArrayPool<K> {
   R withBitArrayAllSet<R>({
     required R Function(BitArray array) fn,
   }) {
-    final score = _acquireAllSet();
-    final r = fn(score);
-    _release(score);
+    final array = _acquireAllSet();
+    final r = fn(array);
+    _release(array);
     return r;
   }
 }
--- a/app/lib/search/token_index.dart
+++ b/app/lib/search/token_index.dart
@@ -5,6 +5,7 @@
 import 'dart:math' as math;
 
 import 'package:meta/meta.dart';
+import 'package:pub_dev/third_party/bit_array/bit_array.dart';
 
 import 'text_utils.dart';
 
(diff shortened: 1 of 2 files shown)


app/lib/search/token_index.dart
```diff
@@ -183,6 +184,33 @@ class ScorePool<K> {
...
+  }
+
+  void _release(BitArray array) {
+    _pool.add(array);
+  }
+
+  R withBitArrayAllSet<R>({
+    required R Function(BitArray array) fn,
+  }) {
+    final score = _acquireAllSet();
```
sigurdm:
```suggestion
    final array = _acquireAllSet();
```

app/lib/search/token_index.dart
```diff
@@ -183,6 +184,33 @@ class ScorePool<K> {
...
+/// A reusable pool for [BitArray] instances to spare some memory allocation.
+class BitArrayPool<K> {
+  final int _length;
+  final _pool = <BitArray>[];
+
+  BitArrayPool(this._length);
+
+  BitArray _acquireAllSet() {
+    final array = _pool.isNotEmpty ? _pool.removeLast() : BitArray(_length);
+    array.setRange(0, _length);
```
sigurdm, line 196: I guess you could do:
```suggestion
    final array = _pool.isNotEmpty ? _pool.removeLast()..setRange(0, _length) : BitArray(_length);
```

isoos: Both need the `setRange` (which is probably a bad name, but a counter-operation of `clearRange` and sets (or clears) all the bits within the given range.

app/lib/search/token_index.dart
```diff
@@ -183,6 +184,33 @@ class ScorePool<K> {
   }
 }
 
+/// A reusable pool for [BitArray] instances to spare some memory allocation.
+class BitArrayPool<K> {
+  final int _length;
+  final _pool = <BitArray>[];
+
+  BitArrayPool(this._length);
```
sigurdm, line 192: Doesn't matter too much, but I guess it could be generic, sharing implentation with ScorePool (if you pass a constructor).

```suggestion
class AllocationPool<T> {
  final _pool = <T>[];
  final T Function() constructor;

  AllocationPool(this.constructor);
```

sigurdm (approved):

app/lib/search/token_index.dart
sigurdm: Ah, sorry, thought the constructor would create a zeroed instance.

app/lib/search/token_index.dart
isoos: I've checked but with the ScorePool's variable initialization parameter, it requires a bit of refactoring first. I'll do that in a follow-up PR.

PR #8798, opened 2025-06-02
Shared code for ScorePool and BitArrayPool.

isoos:
Follow-up to #8797


sigurdm (commented):
If you cannot make a single generic class that covers both kinds of Pool, then I actually prefer just having the code duplicated.

sigurdm (approved):

PR #8808, opened 2025-06-04
Updated admin removal for packages: using the flag on the action

isoos:
- Fixes https://github.com/dart-lang/pub-dev/issues/8730 following the pattern of #8788
- follows the same admin action pattern as the moderate actions
- reuses most of the update code with moderating and deleting a package
- updated tests with only a few relevant bits, we won't need to duplicate everything, as the main code will be already tested with the moderation action
- does the actual deletion after 2 months

### DIFF
--- a/app/lib/admin/actions/package_delete.dart
+++ b/app/lib/admin/actions/package_delete.dart
@@ -17,7 +17,7 @@ A package in this state will appear deleted from the public. But its archive fil
 
 After 2 months it will be fully purged.
 
-To undo a deletion <insert action here>
+To undo a deletion run the same command with `state: false`.
 ''',
   options: {
     'package': 'The package name to be deleted',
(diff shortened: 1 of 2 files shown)


app/lib/admin/actions/package_delete.dart
```diff
@@ -4,54 +4,34 @@
...
-        'message': '''
-Package and all associated resources deleted.
-
-A tombstone has been created
-
-'NOTICE: Redis caches referencing the package will expire given time.'
+  name: 'package-delete',
+  summary: 'Set the admin-deleted flag on a package (making it not visible).',
+  description: '''
+Set the admin-deleted flag on a package (updating the flag and the timestamp). After 2 months it will be fully deleted.
```
sigurdm:
```suggestion
Set the admin-deleted flag on a package (updating the flag and the timestamp).

A package in this state will appear deleted from the public. But its archive file will still exist in the canonical bucket, and the metadata will still be present.

After 2 months it will be fully purged.

To undo a deletion <insert action here>
```

sigurdm (approved):

PR #8809, opened 2025-06-06
Lazy initialization of score and list objects in search.

isoos:
- Part of #8671, with multiple small changes, each verified that improves or does not degrades the performance. Overall improvement seems to be 3-8% depending on the query.
- `IndexedScore` object is initalized on-demand with the `withItemGetter` wrapper method. This skips the allocation and more importantly the reset of the score array, and also its population when there is no text search component.
- Early return of query processing when `_TextResults.hasNoMatch` is set (skips additional post-processing).
- Pre-computed lists can be filtered through either `BitArray` or `IndexedScore` index iteration, making it possible to skip the `IndexedScore` initialization.
- The total hit count calculation is based on the bit array or the indexed score object, there is no need to build a full `*Hit` list of the result just for counting them.
- Removed nullability in multiple places.
- No functional change otherwise.

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -516,7 +516,7 @@ class InMemoryPackageIndex {
     IndexedScore<String> score, {
     // if the item count is fewer than this threshold, an empty list will be returned
     required int requiredLengthThreshold,
-    // note: when no best name match is applied, this parameter will be `-1`
+    // When no best name match is applied, this parameter will be `-1`
     required int bestNameIndex,
   }) {
     final list = <IndexedPackageHit>[];
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -12,7 +12,6 @@ import 'package:meta/meta.dart';
 import 'package:pub_dev/service/topics/models.dart';
 import 'package:pub_dev/third_party/bit_array/bit_array.dart';
 
-import '../shared/utils.dart' show boundedList;
 import 'models.dart';
 import 'search_service.dart';
 import 'text_utils.dart';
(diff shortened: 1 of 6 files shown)


isoos:
Note: I've got mixed results with top-k sorting (vs full result list sorting), not sure why, this needs some further exploration.

Also: now that the overall improvement is verified, I'm happy to split up this PR to incremental parts, if needed.

app/lib/search/token_index.dart
```diff
@@ -185,6 +185,24 @@ abstract class _AllocationPool<T> {
...
+    List<T>? items;
+    T itemFn() {
+      items ??= <T>[];
+      final item = _acquire();
+      items!.add(item);
+      return item;
+    }
+
+    final r = fn(itemFn);
+    if (items != null) {
```
sigurdm: nit:
```suggestion
    final r = fn(itemFn);

    if (items != null) {
```

app/lib/search/token_index.dart
```diff
@@ -185,6 +185,24 @@ abstract class _AllocationPool<T> {
     _release(item);
     return r;
   }
+
+  R withItemGetter<R>(R Function(T Function() itemFn) fn) {
```
sigurdm, line 194: Perhaps add a dartdoc explaining what this does, and why to use it...

app/lib/search/mem_index.dart
```diff
@@ -500,19 +515,18 @@ class InMemoryPackageIndex {
   List<IndexedPackageHit> _rankWithValues(
     IndexedScore<String> score, {
     // if the item count is fewer than this threshold, an empty list will be returned
-    int? requiredLengthThreshold,
-    String? bestNameMatch,
+    required int requiredLengthThreshold,
+    // note: when no best name match is applied, this parameter will be `-1`
```
sigurdm: nit:
```suggestion
    // When no best name match is applied, this parameter will be `-1`
```

sigurdm (approved):
LGTM

PR #8823, opened 2025-06-17
Search index isolate without text indexing for predicate-only queries.

isoos:
- #8671
- The reduced index will be used only when the search query has no text search component. In effect, such predicate-only queries will should be served faster, lowering the overall latencies.


isoos:
Note: to my surprise, the now-added benchmark tool did not show any clear benefit over the current behavior, in fact, it shows that the second index made it worse in most of the benchmarks. I'm not yet sure why it could be worse though... I'll extract the still-useful parts from this PR, but I don't think the two-isolate solution implemented here shows clear benefits.

isoos:
Update: I've identified an implementation bug (fixed in a commit) and also a measurement bug in the benchmark (will fix it soon), will re-run the benchmark again.

sigurdm (approved):
Still not sure I see how this is better than just having 2 isolates with a full index (to utilize more cores of each search instance).
But let's give it a try

PR #8826, opened 2025-06-19
Use top-k sorted list builder with heap for search.

isoos:
- Gets about 1-3% performance boost compared to the current status.
- Implemented using a simple heap implementation, tested up to page 100 (top-k ~ 1000), benefits are slightly degrading but still present.

### DIFF
--- a/app/lib/search/heap.dart
+++ b/app/lib/search/heap.dart
@@ -8,15 +8,35 @@
 /// than or equal to its children's values.
 ///
 /// The provided comparator decides which kind of heap is being built.
-class _Heap<T> {
+class Heap<T> {
   final Comparator<T> _compare;
   final _items = <T>[];
+  bool _isValid = true;
 
-  _Heap(this._compare);
+  Heap(this._compare);
 
   int get length => _items.length;
 
-  void _pushDown(int index) {
+  /// Collects [item] and adds it to the end of the internal list, marks the [Heap]
+  /// as non-valid.
+  ///
+  /// A separate operation may trigger the restoration of the heap proprerty.
+  void collect(T item) {
+    _items.add(item);
+    _isValid = false;
+  }
+
+  /// Collects [items] and adds them the end of the internal list, marks the [Heap]
+  /// as non-valid.
+  ///
+  /// A separate operation may trigger the restoration of the heap proprerty.
+  void collectAll(Iterable<T> items) {
+    _items.addAll(items);
+    _isValid = false;
+  }
+
+  /// Ensures that the tree structre below the [index] is a valid heap.
+  void _heapify(int index) {
     final maxLength = _items.length;
     final item = _items[index];
     while (index < maxLength) {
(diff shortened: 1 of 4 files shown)


app/lib/search/heap.dart
```diff
@@ -0,0 +1,111 @@
...
+class TopKSortedListBuilder<T> {
+  final int _k;
+  final Comparator<T> _compare;
+  final Heap<T> _heap;
+
+  TopKSortedListBuilder(this._k, this._compare)
+      : _heap = Heap<T>((a, b) => -_compare(a, b));
+
+  void addAll(Iterable<T> items) {
+    for (final item in items) {
```
sigurdm: This is worst case no better than sorting I believe.

Consider doing a linear time construction of the heap: https://www.youtube.com/watch?v=MiyLo8adrWw

sigurdm (approved):

app/lib/search/heap.dart
isoos: Done, PTAL. Accidentally merged it, so it is not a separate PR. Measured a bit benefit around k=1000, but otherwise it was roughly the same.

app/lib/search/heap.dart
```diff
@@ -0,0 +1,97 @@
...
+      }
+    }
+    return true;
+  }
+}
+
+/// Builds a sorted list of the top-k items using the provided comparator.
+///
+/// The algorithm uses min-heap to select the top-k items, and then builds
+/// a max-heap and uses heap sort to return the items in descending order.
```
sigurdm: Is this description correct?

I only see a single heap in the code. Also, why would you use a min-heap for selecting top-k items?

isoos: Ah, no, this is now obsolete, will update.

app/lib/search/heap.dart
sigurdm: Not sure I follow. What I tried to suggest was that to choose the top K elements do something like:

```
buildHeap(items); // Linear time, establishes heap property on `items`.
final result = [];
for (var i = 0; i < K; i++) {
  result.add(removeMax(items)); // Take the top element, O(log N)
}
```

sigurdm: (Though it might be higher overhead than what you are doing now - probably have to measure)...

isoos: This is happening in the getTopK method: https://github.com/dart-lang/pub-dev/pull/8826/files#diff-a1d3a2c1e0fee6c6f113d7e173db423f2e974218eb2f666d455a7ab327d0650cR87

- The loop limits the returned values until we have or reach the target k.
- The `yield` is returning the top item.
- Then we remove the top item from the heap, replace it with the last item, and restore the heap property.

Maybe the `k` parameter should be part of the `getTopK` method, there is no use to provide it at construction time.

app/lib/search/heap.dart
```diff
@@ -8,15 +8,35 @@
...
+  /// Collects [items] and adds them the end of the internal list, marks the [Heap]
+  /// as non-valid.
+  ///
+  /// A separate operation may trigger the restoration of the heap proprerty.
+  void collectAll(Iterable<T> items) {
+    _items.addAll(items);
+    _isValid = false;
+  }
+
+  /// Ensures that the tree structre below the [index] is a valid heap.
```
sigurdm:
```suggestion
  /// Ensures that the tree structure below the [index] is a valid heap.
```

sigurdm (approved):

PR #8834, opened 2025-06-25
Reduce memory usage during index construction.

isoos:
- The removal of `_documentsByName` (little use) and `_topics` (unused) is mostly symbolic, there isn't much memory overhead there.
- The change in API doc page index construction (using separate strings instead of concatenating and the splitting the same string) reduces index building time, but more importantly reduces memory usage by several hundred megabytes.


sigurdm (approved):

isoos:
Note: I've confirmed the memory effect on staging too (it has a newer search snapshot).

PR #8836, opened 2025-06-27
fix: flash of white while loading page in dark mode on Firefox

adil192:
- This PR adds a `<meta name="color-scheme" content="light dark"/>` so the browser's default dark mode can be applied before the css is loaded (https://web.dev/articles/color-scheme)
- When the css is loaded, that meta tag is overridden with `color-scheme: dark;` or `light` as needed.

---

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.
  (I'm unsure if the pubspec package version needs increasing)

<details>

</details>

### DIFF
--- a/app/lib/dartdoc/dartdoc_page.dart
+++ b/app/lib/dartdoc/dartdoc_page.dart
@@ -287,6 +287,7 @@ extension DartDocPageRender on DartDocPage {
     final dataBaseHref = p.relative('', from: p.dirname(options.path));
     return d.element('body', classes: [
       'light-theme',
+      'loading-theme',
     ], attributes: {
       'data-base-href':
           baseHref ?? (dataBaseHref == '.' ? '' : '$dataBaseHref/'),
--- a/app/lib/frontend/templates/views/shared/layout.dart
+++ b/app/lib/frontend/templates/views/shared/layout.dart
@@ -184,6 +184,8 @@ d.Node pageLayoutNode({
             requestContext.experimentalFlags.isDarkModeDefault
                 ? 'dark-theme'
                 : 'light-theme',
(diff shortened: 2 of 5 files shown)


adil192:
When the browser starts loading the page, it loads the html first before the css. So having the color-scheme defined in the html means the browser can apply a dark background sooner, instead of having to wait for the css to load.

https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme:
> Along with the above CSS, also include the [`<meta name="color-scheme">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/color-scheme) HTML [`<meta>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta) tag in the [`<head>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/head), before any CSS style information, to inform user agents about the preferred color scheme, helping prevent unwanted screen flashes during the page load.

isoos:
Upon closer inspection, I see a difference between pub.dev and `dartdoc` in this regard: pub.dev blocks the rendering until the variable is read and the main style is set, while dartdoc seems to not block it, and the pub.dev embedding of the dartdoc pages doesn't fix that either. I prefer to fix that in dartdoc (or its pub.dev embedding) the same way we use it on the rest of pub.dev, to preserve the consistency - by blocking the rendering for now.

As a follow-up change, we should also support media selectors, and after that, we should consider if/how the feature should work with non-blocking rendering. I wouldn't mind having the metadata here on the page (after the above changes), but I don't think it will help for those who leave their OS theme settings on default but switch their theme on pub.dev.

sigurdm:
@isoos is this still relevant?

PR #8837, opened 2025-06-27
Render-blocking dark-theme initialization for dartdoc pages.

isoos:
- A simpler alternative to #8836 that focuses only on the rendering glitch.
- Uses the same initialization script and the same (blocking) way on dartdoc pages as the rest of pub.dev.
- dartdoc's own initalization does not support the OS-provided theme-preference yet https://github.com/dart-lang/dartdoc/blob/main/web/theme.dart - removing that part from the initialization script. We need to implement that in dartdoc first (https://github.com/dart-lang/dartdoc/issues/4069).

### DIFF
--- a/static/js/dark-init.js
+++ b/static/js/dark-init.js
@@ -14,13 +14,6 @@
     // ignore errors e.g. when media query matching is not supported
   }
 
-  // NOTE: dartdoc currently does not support externally supported color scheme,
-  //       and we disable that part for the affected URLs.
-  // TODO: remove this part when dartdoc supports the above detection too
-  if (window.location.pathname.startsWith('/documentation/')) {
-    mediaPrefersDarkScheme = false;
-  }
-
   // Detects whether the control widget was set to use a specific theme
   let colorTheme = window.localStorage.getItem('colorTheme');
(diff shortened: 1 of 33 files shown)


static/js/dark-init.js
```diff
@@ -14,6 +14,13 @@
     // ignore errors e.g. when media query matching is not supported
   }
 
+  // NOTE: dartdoc currently does not support externally supported color scheme,
+  //       and we disable that part for the affected URLs.
+  // TODO: remove this part when dartdoc supports the above detection too
```
sigurdm: Can we create an issue for dartdoc for this TODO?

sigurdm (approved):

static/js/dark-init.js
isoos: https://github.com/dart-lang/dartdoc/issues/4069

isoos:
I have checked again with both Firefox and Chrome, with various different settings (including the `<script blocking="render" ...` attribute). The current init script improves the flashing on both browsers, and e.g. the `blocking` attribute doesn't add much on top of it. However, the site is not completely flash-free, in certain cases I've noticed a few more full or partial flashing, opening tracking issues for them.

PR #8839, opened 2025-06-30
WIP/RFC: change event processing.

isoos:
- This is an exploration on how an internal event processing abstraction could look like.
- The emitting of the changes could be automatic, not yet sure what limitations we would have with Datastore or SQL, but having keys is a possibility.
- The processing then matches the change and may emit more post-processing event(s). I've written examples for single- and multi-event processing.
- There are many ways in the rules or in the semantics that is subject to change, let's discuss them!


isoos:
Closing as we've discussed a more direct approach.

PR #8843, opened 2025-07-02
Restrict Flutter SDK libraries in the search index.

isoos:
Searching for `intl` or `meta` is confusing, as it will expose the SDK-included package link and the regular package as results. Restricting the SDK packages to the list defined explicitly (the first section on `api.flutter.io`) and also packages that start with `flutter_` in the index json, right now the following:

```
flutter_driver
flutter_driver_extension
flutter_gpu
flutter_localizations
flutter_test
flutter_web_plugins
```

In the long term we should probably use some kind of additional field in the `index.json` that `dartdoc` generates so that we can have this decision automatically, without any external list.

### DIFF
--- a/app/lib/search/sdk_mem_index.dart
+++ b/app/lib/search/sdk_mem_index.dart
@@ -108,7 +108,7 @@ class SdkMemIndex implements SdkIndex {
       'flutter',
       _flutterUri,
       flutterIndex,
-      libraryFn: (library) =>
+      includeLibraryFn: (library) =>
           library.startsWith('flutter_') ||
           _flutterLibraryAllowlist.contains(library),
     );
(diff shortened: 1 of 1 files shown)


app/lib/search/sdk_mem_index.dart
```diff
@@ -99,8 +122,9 @@ class SdkMemIndex implements SdkIndex {
   void _addDartdocIndex(
     String sdk,
     Uri baseUri,
-    DartdocIndex index,
-  ) {
+    DartdocIndex index, {
+    bool Function(String library)? libraryFn,
```
sigurdm: Can this have a more descriptive name, or a dartdoc description?

sigurdm (approved):

PR #8846, opened 2025-07-03
Trigger post-update events for Package updates.

isoos:
- Single and explicit method call for all potential post-update events.
- Cache is always purged, task update and API export can be skipped.
- API export partial skip is indicated with a new flag (`skipVersionsExport`), but it is not implemented yet.
- Updated all but one uses of `purgePackageCache` (TBD later if/how that should be changed).
- The trigger method and also the `AsyncQueue` update returns an optional `Future` that can be awaited, but doesn't cause linter issues when not awaited.
- If we like this pattern, I would also implement it for the other entities too.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -381,7 +381,11 @@ class PackageBackend {
       return true;
     });
     if (updated) {
-      await purgePackageCache(package);
+      triggerPackagePostUpdates(
+        package,
+        skipTask: true,
+        skipExport: true,
+      );
     }
     return updated;
   }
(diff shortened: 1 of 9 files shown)


app/lib/package/backend.dart
```diff
@@ -2096,3 +2098,49 @@ class _VersionTransactionDataAcccess {
     return await _tx.query<PackageVersion>(pkgKey).run().toList();
   }
 }
+
+/// Triggers post-update event processing after a [Package] object is part of
+/// a transaction.
+///
+/// Returns a record with an optionally awaitable [Future] in case the caller needs
+/// wait the updates before yielding its response.
```
sigurdm:
```suggestion
/// Returns a record with an optionally awaitable [Future] in case the caller needs to
/// wait for the updates before yielding its response.
```

app/lib/package/backend.dart
```diff
@@ -2096,3 +2098,49 @@ class _VersionTransactionDataAcccess {
...
+({Future future}) triggerPackagePostUpdates(
+  String package, {
+  /// Skip trigger a new analysis on the package.
+  bool skipTask = false,
```
sigurdm: task sounds too generic.

Perhaps:
```suggestion
  bool skipReanalysis = false,
```

Issue #5532, opened 2022-02-23
Feed for new versions.


bsutton:
I'm looking to see if there is a feed containing updates for new versions of packages uploaded to pub.dev.

Possibly this might be a REST end point that I can ask for new versions since some date/time.
I might the have to poll the end point every few minutes to get list of new packages/versions since the last poll.

Some time ago I seem to recall reading about something along these lines.

Does such an service exist in pub.dev?

isoos:
There is no such feed except for the https://pub.dev/feed.atom

What's the use case you are looking to solve?

bsutton:
We are building a private package repo saas for dart.
We want the ability to send out notifications to users when a new version
of a public package is released.

sigurdm:
Perhaps our progress on [dependabot-integration](https://github.com/dependabot/dependabot-core/pull/4510), and especially the [dependency_services](https://github.com/dart-lang/pub/commit/941191f7f83ad60259348860197cfcdd83bb8e6f) we made could be of interest here.
It will only query for new versions of packages that are already dependencies.

But maybe that is just enough.

bsutton:
So I'm not familiar with how dependabot works outside of its integration
with github.

I assume it works as a service that scans dependencies in a git repo and
sends alerts if a new dependency is available.
Is this intended for organisations to implement internally or is it
intended to allow github to support dart?
How does it talk to pub.dev?
I'm not certain how this might provide us with an alternative, given my
(vague) understanding of what it does.
I assume that it must somehow be pulling version updates out of pub.dev?
This is a service we want to offer to our private package repo customers.

We have played with the atom feed and it looks like we can poll that every
five minutes or so and then pull the new versions' meta data.
Haven't been able to find any doco on the atom feed but it looks to have
about the last 6 hours of updates.
Is based on a count of the last 'n' entries or a time span?

We are keen not to have to do major re-sinks after any outages (at our
end). Provided we resume services within the window of the atom feed this
should be doable.



S. Brett Sutton
Noojee Contact Solutions
03 8320 8100

sigurdm:
> I'm not certain how this might provide us with an alternative, given my
(vague) understanding of what it does.

The dependency_services we implemented is akin to a machine-parsable `pub outdated` with more detailed output (it detects possible updates for each dependency separately).
described [here](https://github.com/dependabot/dependabot-core/blob/96354d5649e45bacb479a3eafa2cab2c0a5bf1cf/pub/README.md#dependency-service-interface).

[...]

Issue #5183, opened 2021-10-14
RSS feed should contain the relevant snippet extracted from CHANGELOG


sigurdm:

sigurdm:
We perhaps need to formalize the changelog format first... though pub.dev is already parsing and splitting the changelog based on some heuristic, that might be good enough.

I don't know if the rss feed has enough usage to make this worth investing in.

Issue #5127, opened 2021-09-21
We're markdown rendering of changelog - some headers with dates, others not


kevmoo:
URL: https://pub.dev/packages/vector_math/changelog

See

![image](https://user-images.githubusercontent.com/17034/134260191-e046b3f7-e166-48f8-ab60-dc2c58e26044.png)

isoos:
The content comes from the markdown source of the changelog:
https://raw.githubusercontent.com/google/vector_math.dart/master/CHANGELOG.md

isoos:
However, the parsing should recognize the `[version] - [something else]` pattern, it seems that it doesn't do it properly at the moment.

creativecreatorormaybenot:
@isoos @kevmoo I want to note that this pattern is _not_ encouraged anymore.

It was (mistakenly) kept in the Flutter plugin templates for a long time, which I only fixed in February of this year (https://github.com/flutter/flutter/pull/76019).

I think fixing it is a good idea as some older packages decide to keep it, however, I think it is way more important to encourage consistency and communicate the recommended style properly.

sigurdm:
Seems we have improved here at some point:
https://pub.dev/packages/vector_math/versions/2.1.1/changelog
![image](https://github.com/user-attachments/assets/bb81de19-0ec7-4081-bcbd-3ebdaa89692d)

PR #8856, opened 2025-07-17
New changelog parser + typed representation of the changelog structure.

### LINKED ISSUE #7395: Show versions in changelog consistently
Currently we allow prefixing the versions with a `v`, we could make a decision on how we show it on pub.dev , so it looks the same on all package pages - either with or without the `v`.

isoos:
- This only a small step to get improved markdown processing (see #3331).
- The new parser is designed to parse a wide range of changelog formats, and in itself it is **not a recommendation** for any particular format, a recommendation for that is outside of the scope of this PR.
- While the date parsing is improved a bit, it can be better in the future (see #5127).
- The parsed output is integrated only in the markdown processing for now, but the typed structure enables further features to be implemented, e.g. #5183, #5267, #5532.
- Small consistency-related changes in the output on the changelog tab:
  - `[]` around the version are removed
  - the `v` before the version is removed (fixes #7395)
  - the date format is `yyyy-MM-dd`, it may be `yyyy/MM/dd` in the source
  - the date is always separated by ` - ` from the version (optional in the source)

### DIFF
--- a/app/lib/shared/changelog.dart
+++ b/app/lib/shared/changelog.dart
@@ -20,7 +20,6 @@ library;
 import 'package:collection/collection.dart';
 import 'package:html/dom.dart' as html;
 import 'package:html/parser.dart' as html_parser;
-import 'package:markdown/markdown.dart' as m;
 import 'package:pub_semver/pub_semver.dart';
 
 /// Represents the entire changelog, containing a list of releases.
@@ -116,14 +115,6 @@ class ChangelogParser {
   })  : _strictLevels = strictLevels,
         _partOfLevelThreshold = partOfLevelThreshold;
 
-  Changelog parseMarkdown(String input) {
-    final nodes =
-        m.Document(extensionSet: m.ExtensionSet.gitHubWeb).parse(input);
-    final rawHtml = m.renderToHtml(nodes);
-    final root = html_parser.parseFragment(rawHtml);
-    return parseHtmlNodes(root.nodes);
-  }
-
   /// Parses markdown nodes into a [Changelog] structure.
   Changelog parseHtmlNodes(List<html.Node> input) {
     String? title;
@@ -161,7 +152,7 @@ class ChangelogParser {
             node.localName != firstReleaseLocalName) {
           continue;
         }
-        final headerText = _extractText(node).trim();
+        final headerText = node.text.trim();
 
         // Check if this looks like a version header first
         final parsed = _tryParseAsHeader(node, headerText);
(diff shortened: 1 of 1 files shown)


app/lib/shared/changelog.dart
```diff
@@ -0,0 +1,345 @@
...
+
+/// The [ChangelogParser] accommodates various formatting styles. It can
+/// effectively parse changelogs with inconsistent header levels or those
+/// that include additional information beyond just the version number in
+/// the release header.
+///
+/// The parser is designed to support the widely adopted "Keep a Changelog"
+/// format (see https://keepachangelog.com/en/1.1.0/ for details).
+/// Additionally, it has been tested with a diverse set of changelog files
+/// available a part of the packages on https://pub.dev/.
```
sigurdm: Broken sentence
```suggestion
/// available as part of the packages on https://pub.dev/.
```

app/lib/shared/changelog.dart
```diff
@@ -0,0 +1,345 @@
...
+    required this.version,
+    this.anchor,
+    this.label,
+    this.date,
+    this.note,
+    required this.content,
+  });
+}
+
+/// Describes an arbitrary content (e.g. a changelog description or inside an entry).
```
sigurdm:
```suggestion
/// Describes an arbitrary piece of content (e.g. the description of a single version).
```

app/lib/shared/changelog.dart
```diff
@@ -0,0 +1,345 @@
...
+      description: description,
+      releases: releases,
+    );
+  }
+
+  String _extractText(html.Node node) {
+    if (node is html.Text) {
+      return node.text;
+    } else if (node is html.Element) {
+      return node.nodes.map(_extractText).join();
```
sigurdm: I worry that by doing recursive calls here we are subject to crash by stack-overflow...

app/lib/shared/changelog.dart
```diff
@@ -0,0 +1,345 @@
...
+  final bool _strictLevels;
+  final int _partOfLevelThreshold;
+
+  ChangelogParser({
+    bool strictLevels = false,
+    int partOfLevelThreshold = 2,
+  })  : _strictLevels = strictLevels,
+        _partOfLevelThreshold = partOfLevelThreshold;
+
+  Changelog parseMarkdown(String input) {
```
sigurdm: Does the input get sanitized in this pipeline, or does that happen later?

isoos: Sanitization is the last step, anything we do here, will get sanitized at the end.

Note: since the `parseMarkdown` is no longer used (helper method in the test, and post-markdown-rendering step in the lib/ code), removing it.

app/lib/shared/changelog.dart
isoos: Yeah, it is probably a not really necessary item here, switching to always do `node.text.trim()` (it is using a tree visitor inside, but it doesn't need such joins, it appends the text to a buffer).

sigurdm (approved):

PR #8862, opened 2025-07-24
Run 'pub get' for worker's SDK selection.

### LINKED ISSUE #7268: Extract pubspec.yaml from the package archive instead of the API endpoint before analysis SDK selection.

isoos:
- Fixes #7268.


isoos:
Note: While I was implementing this, I got some second thoughts whether this was really needed, or we should only use the SDK constraints. My main worry is that e.g. a beta SDK is needed, but for some reason the `pub get` fails, and it falls back to using the stable SDK, and it gets worse results with that.

PR #8863, opened 2025-07-25
Track only visible packages as dependency in the search index.

isoos:
- https://github.com/dart-lang/pana/issues/1167 (side comment)
- I don't think this has caused any issue yet, but it is better for consistency.


sigurdm (commented):
Should this have a test?

Issue #7879, opened 2024-07-17
Consider a "Maintainer Wanted" tag


jonasfj:
If set on a package and not removed in 6 months, then maybe the package should change to "discontinued".

PR #8870, opened 2025-07-30
Maintainer wanted model, badge, action and admin UI.

isoos:
- #7879
- Note: this may also need a search checkbox and search autocomplete (TBD), and we should also consider how the expiry should work as in the original suggestion.


isoos:
Closing this as it has drifted off significantly.

PR #8879, opened 2025-08-06
New 404 page rendering: link to package page or search.

### LINKED ISSUE #8871: `pub.dev/<package_name>` not found page should provide relevant actions
Right now navigating to `pub.dev/<package_name>` leads to a 404 not found page. However, it's quite likely someone navigating to this path with a valid package name wants to visit that package's page.

Rather than just listing possible next steps, I think we could provide two more prominent buttons:

1. A button to go directly to searching with the same query.
2. If a package with a matching name does exist, additionally add a button that goes directly to that package's page.

An alternative or addition to **1** could be pre-filling the search bar.

isoos:
- Fixes #8871.

Screenshots:
image
image
image
image

### DIFF
--- a/app/lib/frontend/handlers/misc.dart
+++ b/app/lib/frontend/handlers/misc.dart
@@ -272,18 +272,58 @@ Future<shelf.Response> experimentalHandler(shelf.Request request) async {
   });
 }
 
-String _notFoundMessage(Uri requestedUri) {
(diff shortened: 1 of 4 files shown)


sigurdm:
I don't think we should use a button - this is a link more than it is an action

We can always propose the pre-filled search, even when there is a package match

Maybe something like: 

image

PR #8885, opened 2025-08-08
Cleanup ModerationCase entities after 3 years.

### LINKED ISSUE #8883: Delete ModerationCase entities after 3 years

isoos:
Closes #8883.

### DIFF
--- a/app/lib/admin/backend.dart
+++ b/app/lib/admin/backend.dart
@@ -779,6 +779,34 @@ class AdminBackend {
     return refCase;
   }
 
+  /// Scans datastore and deletes [ModerationCase] entities where the last
+  /// action was more than 3 years ago.
+  Future<void> deleteModerationCases({
+    @visibleForTesting DateTime? before,
+  }) async {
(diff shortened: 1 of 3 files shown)


app/lib/admin/backend.dart
```diff
@@ -779,6 +779,30 @@ class AdminBackend {
     return refCase;
   }
 
+  /// Scans datastore and deletes [ModerationCase] entities where the last
+  /// action was more than 3 years ago.
```
sigurdm:
```suggestion
  /// Scans datastore and deletes [ModerationCase] entities opened more than 3 years ago.
```

app/test/admin/moderation_case_resolve_test.dart
```diff
@@ -81,14 +91,27 @@ void main() {
...
+
+      // close case
       expect(
         await _close(
           mc.caseId,
           reason: 'The package violated our policy.',
         ),
         'moderation-applied',
       );
+
```
sigurdm, line 109: An ideal test would run the actual cleanup task with a clock set 3 years into the future.

Not a blocker

Issue #8891, opened 2025-08-12
Create better integration test(s) for search isolate + search handler + search client.


isoos:
Right now we have tests for each component independently, but not in a single end2end test.

Issue #8887, opened 2025-08-08
Display the /my-liked-packages page the same way as other package listing.


isoos:
Background: Because we don't have the private package-likes information in the search index, we couldn't easily apply a search filter that does search in the scope of the packages that the current user liked. A recent discussion outlined a plan to implement it with the following components:

- [ ] Maintain a (redis) cache of liked packages for a user. It is not clear yet if and how we need to limit the number of liked packages, or at least the packages that go into the cached list. As a start this could be a List of the names, and if it grows too large, we may try to build a bitmap instead (if/when a consistent package ordering is implemented).
- [ ] When the user visits the `/my-liked-packages` page, we can use this cache to pass it to the search service, and the search service can apply a filter on those packages.
- [ ] Right now the internal search backend is queried with HTTP GET queries. Due to the increased parameter length, we should transition to POST requests, where the body has no size limit. However, we should probably keep some query information in the URL, to help tracking issues if latency jumps. (We should have a few overlapping release with both endpoints, just in case a rollout or a rollback may need them for a few minutes).
- [ ] We should keep the feature of unliking a package, either only when the  user is on the `/my-liked-packages` or also when they are on the generic search page (`/packages?q=...`). We could add the same like icon to the listings, or add the current "Unlike" button at the bottom of the package listings.

PR #8892, opened 2025-08-12
Search handler and client with HTTP POST-based request serialization.

isoos:
- #8887
- moved `TextMatchExtent` to `pkg/_pub_shared`, and added the new data class there, as the build command's inclusion rules otherwise couldn't identify all the required types.
- Updated handler creation and processing, but kept both search variants on `/search`, since we have so little change based on the HTTP method
- Update the search client to call the POST only when the new experimental flag is enabled.
- Restructured the retry-GET handler and created `withRetryHttpClient` to reuse some of its logic
- Note: we need to have a better search isolate/handler integration test, but that is independent of this change, tracking in #8891

### DIFF
--- a/app/lib/search/handlers.dart
+++ b/app/lib/search/handlers.dart
@@ -52,6 +52,7 @@ Future<shelf.Response> _debugHandler(shelf.Request request) async {
 }
 
 /// Handles GET /search requests.
+/// Handles POST /search requests.
 Future<shelf.Response> _searchHandler(shelf.Request request) async {
   final info = await searchIndex.indexInfo();
   if (!info.isReady) {
--- a/app/lib/search/handlers.dart
+++ b/app/lib/search/handlers.dart
@@ -3,9 +3,12 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:convert';
 
+import 'package:_pub_shared/search/search_request_data.dart';
 import 'package:logging/logging.dart';
 import 'package:shelf/shelf.dart' as shelf;
+import 'package:shelf_router/shelf_router.dart';
 
 import '../shared/env_config.dart';
 import '../shared/handlers.dart';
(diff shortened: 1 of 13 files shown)


app/lib/search/handlers.dart
```diff
@@ -48,21 +45,27 @@ Future<shelf.Response> _readinessCheckHandler(shelf.Request request) async {
...
 
-/// Handler /debug requests
+/// Handler GET /debug requests
 Future<shelf.Response> _debugHandler(shelf.Request request) async {
   final info = await searchIndex.indexInfo();
   return debugResponse(info.toJson());
 }
 
-/// Handles /search requests.
+/// Handles GET /search requests.
```
sigurdm, line 54: This seems to contradict the code below that branches on the method....

isoos: Updated comment, it should handle both.

pkg/_pub_shared/lib/search/search_form.dart
```diff
@@ -53,7 +53,6 @@ enum SearchOrder {
   /// WARNING: The value shouldn't be used anymore.
   ///
   /// TODO: remove in a future release.
-  @Deprecated('Popularity is no longer used.')
   popularity,
```
sigurdm: Was this intentional?

sigurdm (approved):

pkg/_pub_shared/lib/search/search_form.dart
isoos: Intentional, but more of a convenience: if we keep it, the generated code uses it and analyzer complains about using a deprecated field. I did not want to remove it yet, but maybe we should?

isoos: Removed the field. I think we did have enough time since we have changed from popularity to downloads. Old bookmarks may break, but that is going to happen no matter when we remove it.

PR #8897, opened 2025-08-15
Upgrade runtime and analysis SDKs, dependencies, CI.

isoos:


isoos:
> `test/publish_rejection_test.dart` has a failure, think you need to update a constraint somewhere in that test.

Yeah, filed https://github.com/dart-lang/pub-dev/issues/8898 to make it more automatic.

PR #8907, opened 2025-08-20
Allow searching and filtering on packages that are provided in the query

isoos:
#8887

### DIFF
--- a/app/lib/search/mem_index.dart
+++ b/app/lib/search/mem_index.dart
@@ -207,6 +207,7 @@ class InMemoryPackageIndex {
     BitArray packages,
     IndexedScore<String> Function() scoreFn,
   ) {
+    _resetBitArray(packages, query.packages);
     final predicateFilterCount = _filterOnPredicates(query, packages);
     if (predicateFilterCount <= query.offset) {
       return PackageSearchResult.empty();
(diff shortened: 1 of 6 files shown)


app/lib/search/mem_index.dart
```diff
@@ -333,6 +334,21 @@ class InMemoryPackageIndex {
     );
   }
 
+  /// The [BitArrayPool] does not resets the reused pool items, because initialization
```
sigurdm: typo
```suggestion
  /// The [BitArrayPool] does not reset the reused pool items, because initialization
```

app/test/search/mem_index_test.dart
```diff
@@ -543,6 +544,27 @@ server.dart adds a small, prescriptive server (PicoServer) that can be configure
           ServiceSearchQuery.parse(query: '=', order: SearchOrder.text));
       expect(rs.isEmpty, isTrue);
     });
+
+    test('only packages list filter', () {
+      final rs = index
+          .search(ServiceSearchQuery(SearchRequestData(packages: ['http'])));
```
sigurdm: Perhaps include two packages in this test, just for good measure.

sigurdm (approved):

PR #8913, opened 2025-08-22
Send changelog excerpt in package uploaded email.

### LINKED ISSUE #2028: Feature request: include change log in "Package uploaded" emails
Would be nice if the "Package uploaded: " email that is sent to package uploaders will include the changelog, that's the main thing I care about when I see these emails.

Thanks!

isoos:
- Fixes #2028.
- Excerpt is generated with the following processing: markdown -> html -> changelog parser -> html content of the release -> markdown-ified version of the release ->  taking only 10 lines, with limited character counts.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -1175,7 +1175,7 @@ class PackageBackend {
       versionKey: newVersion.qualifiedVersionKey,
       changelogContent: entities.changelogAsset?.textContent,
     );
-    if (changelogExcerpt != null) {
+    if (changelogExcerpt != null && changelogExcerpt.isNotEmpty) {
       uploadMessages
           .add('Excerpt of the changelog:\n```\n$changelogExcerpt\n```');
     }
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -9,6 +9,7 @@ import 'package:_pub_shared/data/account_api.dart' as account_api;
 import 'package:_pub_shared/data/package_api.dart' as api;
 import 'package:_pub_shared/utils/sdk_version_cache.dart';
 import 'package:clock/clock.dart';
+import 'package:collection/collection.dart';
 import 'package:convert/convert.dart';
 import 'package:crypto/crypto.dart';
 import 'package:gcloud/service_scope.dart' as ss;
@@ -21,6 +22,8 @@ import 'package:pub_dev/package/tarball_storage.dart';
 import 'package:pub_dev/scorecard/backend.dart';
 import 'package:pub_dev/service/async_queue/async_queue.dart';
 import 'package:pub_dev/service/rate_limit/rate_limit.dart';
+import 'package:pub_dev/shared/changelog.dart';
+import 'package:pub_dev/shared/monitoring.dart';
 import 'package:pub_dev/shared/versions.dart';
 import 'package:pub_dev/task/backend.dart';
 import 'package:pub_package_reader/pub_package_reader.dart';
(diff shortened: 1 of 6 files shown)


isoos:
Note: maybe we should also remove `<` and `>` characters from the excerpt, so that we don't accidentally create a HTML tag in the email?

app/lib/shared/changelog.dart
```diff
@@ -101,6 +103,176 @@ class Content {
...
+        visitChildren(node);
+        _listDepth--;
+        if (_listDepth == 0) _writeln();
+        break;
+      case 'li':
+        final parent = node.parent?.localName;
+        final indent = '  ' * (_listDepth - 1);
+        _write(indent);
+        if (parent == 'ol') {
+          _write('1. ');
```
sigurdm: Should we attempt counting here?

app/lib/package/backend.dart
```diff
@@ -1315,6 +1327,47 @@ class PackageBackend {
...
+      final lines = text.split('\n');
+      final excerpt = lines
+          // filter empty or decorative lines to maximalize usefulness
+          .where((line) =>
+              line.isNotEmpty &&
+              !line.startsWith('```') && // also removes the need to escape it
+              !line.startsWith('---'))
+          .take(10)
+          // prevent accidental HTML-tag creation
+          .map((line) => line
```
sigurdm: Should we do `&lt;`-style escaping? Would that work in an email?

app/lib/package/backend.dart
```diff
@@ -1315,6 +1327,47 @@ class PackageBackend {
...
+              .replaceAll('<', '[')
+              .replaceAll('>', ']')
+              .replaceAll('&', ' '))
+          .map((line) =>
+              line.length < 76 ? line : '${line.substring(0, 70)}[...]')
+          .join('\n');
+      return excerpt;
+    } catch (e, st) {
+      _logger.pubNoticeShout('changelog-parse-error',
+          'Unable to parse changelog for $versionKey', e, st);
```
sigurdm, line 1369: Will this log the package name also?

app/test/package/upload_test.dart
```diff
@@ -1210,6 +1214,15 @@ void main() {
         expect(email.subject, 'Package uploaded: oxygen 3.0.0');
         expect(email.bodyText,
             contains('https://pub.dev/packages/oxygen/versions/3.0.0\n'));
+        expect(
+            email.bodyText,
+            contains('\n'
+                'Excerpt of the changelog:\n'
+                '```\n'
+                'Some bug fixes:\n'
```
sigurdm, line 1222: Should we consider also including the header of the relevant section?

app/lib/package/backend.dart
isoos: Yes, it is `<package>/<version>`.

app/test/package/upload_test.dart
isoos: Not sure. For most packages, it will be only the version string. Maybe if the changelog itself is not too long?

app/lib/package/backend.dart
isoos: I think we could have emails with both text and HTML versions, but that would complicate things way too much. I'd rather focus on the text content here.

sigurdm (approved):

app/lib/package/backend.dart
```diff
@@ -1315,6 +1327,47 @@ class PackageBackend {
...
+          .take(10)
+          // prevent accidental HTML-tag creation
+          .map((line) => line
+              .replaceAll('<', '[')
+              .replaceAll('>', ']')
+              .replaceAll('&', ' '))
+          .map((line) =>
+              line.length < 76 ? line : '${line.substring(0, 70)}[...]')
+          .join('\n');
+      return excerpt;
```
sigurdm, line 1366: Should we return null if the excerpt is empty/only whitespace?

isoos: We are filtering out most of the empty spaces and lines, maybe empty list items will be still kept. I'll add a bit more conditions to prevent / post-filter those.

sigurdm: I was thinking if the "whole message" is whitespace only - it would look weird in the email.
```
Here's an excerpt from the CHANGELOG:

And now for something completely different... 
```

kevmoo:
"markdown -> html -> changelog parser -> html content of the release -> markdown-ified version of the release -> taking only 10 lines, with limited character counts." – WOW!

PR #8925, opened 2025-08-26
Expose admin-deleted timestamp in admin info actions

isoos:


app/lib/admin/actions/package_info.dart
```diff
@@ -46,6 +46,8 @@ Loads and displays the package information.
         'isModerated': p.isModerated,
         if (p.moderatedAt != null)
           'moderatedAt': p.moderatedAt?.toIso8601String(),
+        if (p.adminDeletedAt != null)
```
sigurdm, line 50: Do we also have isAdminDeleted?

I guess it is kind of redundant, but better be aligned with isModerated/moderatedAt

PR #8928, opened 2025-08-28
New tag to indicate search on the current user's liked packages.

isoos:
- #8887
- behind (updated) experimental flag
- the presence of the tag is not propagated to the backend (since there won't be any package with it), it needs to be removed before sending the query to the backend
- when the tag is present, we must not cache the results (or should add a userid to the cache key, TBD)
- updated like backend to work with `userId` instead of `User` object


sigurdm (approved):
Nice!

Issue #5545, opened 2022-03-02
Make it easy to find unendorsed implementation of federated plugins


stuartmorgan-g:
Currently there's no good way, in general, to find unendorsed implementations of a federated plugin. As an example, https://pub.dev/packages/just_audio currently has unendorsed Windows and Linux implementations, which are in a sort of pre-release/evaluation stage. The only way for someone to know that right now is because the developer of just_audio actually added information to the README about it.

Some things we could do:
1. Add a sidebar entry when a plugin is federated (which we could heuristically define as "has at least one `default_package` entry in its plugin section"), listing all unendorsed packages that implement it (which is detectable using the `implements` entry in the implementation packages).
    - A less convenient but more spam-resistant option would be just to have a link to a search for implementations, if any unendorsed implementations exist.
2. An explicit search option to find implementations of a package.
3. Some kind of special handling in the supported platform section, to address the issue that searching by platform will currently exclude things without endorsed implementations for that platform. For instance, searching for audio packages that support Windows won't turn up `just_audio`. It might be interesting to augment the metadata with information about unendorsed implementations such that it would turn up in searches but be indicated in some way. E.g., `PLATFORM: [ANDROID] [IOS] [MACOS][WEB] (Unendorsed: [LINUX] [WINDOWS])`, where the word "unendorsed" is a link to a search for unendorsed implementation packages.

(/cc @mit-mit @jonasfj filed per recent discussion)

jonasfj:
I like the idea that an implementation-package gets listed in the sidebar of the package it implements.

But to mitigate spam we should probably consider:
 * Forcing implementation-packages to be _unlisted_ (ie. won't show up in search).
 * Allow original package author to untag implementation-packages.

stuartmorgan-g:
> Is this still relevant?

Yes.

> Currently pub.dev doesn't explicitly treat federated plugins differently in any way.

Correct, leading to this issue's problem statement:

>> Currently there's no good way, in general, to find unendorsed implementations of a federated plugin.

stuartmorgan-g:
Endorsing a package is [defined](https://docs.flutter.dev/packages-and-plugins/developing-packages#endorsed-implementations) as adding a `default_package` entry in the plugin block pointing to that package.

(In practice if the app-facing package has a dependency on the package it has basically the same effect even if the `default_package` entry is missing, but pub.dev doesn't have to deal with that; it can just rely on the explicit endorsement.)

isoos:
[...]

PR #8933, opened 2025-08-29
Tagging 'implements-federated-plugin:<name>' for packages that implement a plugin.

isoos:
#5545

### DIFF
--- a/pkg/_pub_shared/lib/search/tags.dart
+++ b/pkg/_pub_shared/lib/search/tags.dart
@@ -16,7 +16,8 @@ const allowedTagPrefixes = [
   'sdk:',
   'show:',
   'has:',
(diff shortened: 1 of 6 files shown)


pkg/_pub_shared/lib/search/tags.dart
```diff
@@ -16,7 +16,8 @@ const allowedTagPrefixes = [
   'sdk:',
   'show:',
   'has:',
-  'topic:'
+  'topic:',
+  'plugin-for:',
```
sigurdm: Perhaps we call this something like `implementsFederatedPlugin:`?

sigurdm (approved):

isoos:
Also verified on staging.

PR #8939, opened 2025-09-02
Parse change headers with "Version <v>" pattern too.

isoos:


app/lib/shared/changelog.dart
```diff
@@ -401,6 +401,11 @@ class ChangelogParser {
       }
     }
 
+    // remove leading 'Version' label
+    if (inputLowerCase.startsWith('version ')) {
+      input = input.trim().substring(7).trim();
```
sigurdm: Do you really need both `.trim`s?

sigurdm:
```suggestion
      input = input.trim().substring('version '.length).trim();
```

isoos: good point, as `input` is already trimmed when the method is being called, and we don't need trim for `inputLowerCase` creation either.

PR #8947, opened 2025-09-08
Update SDK constraint and language version to 3.9 + formatting

isoos:
- updated SDK constraint in `pubspec.yaml` files and ran `pub get` + `dart format`
- no other changes has been done


isoos:
Note: this is a step that is very likely needed for the analyzer upgrade, since `json_serializer` needs updated language version for new output format.

sigurdm:
I think you need to re-codegen

PR #8951, opened 2025-09-10
Move screenshot-related utilities to a separate package.

isoos:
- Part of creating a shared/separate tool for creating screenshots in both pub-dev and dartdoc (https://github.com/dart-lang/dartdoc/pull/4074)
- Dropped the `PUB_` prefix from the environment variable, otherwise kept files unchanged.
- Updated/expanded `README.md`.


pkg/pub_integration/lib/src/pub_puppeteer_helpers.dart
```diff
@@ -4,8 +4,8 @@
 import 'dart:async';
 
 import 'package:puppeteer/puppeteer.dart';
+import 'package:screenshot_tools/screenshot_utils.dart';
```
sigurdm: nit: Should we change the name of the main library to align with the package name?


```suggestion
import 'package:screenshot_tools/screenshot_tools.dart';
```

sigurdm: And should we have the name of the new package indicate somehow what it is screenshotting?

Something like:

```suggestion
import 'package:browser_screenshots/browser_screenshots.dart';
```

isoos: I'll do the library name changes in a follow-up, after more changes has been done, e.g. the dartdoc-related tooling is also implemented.

isoos: How about `puppeteer_screenshots`?

PR #8954, opened 2025-09-12
Refactor search isolate control + minimal test.

isoos:
- Part of #8891. Started using it in fake_server, seems to work, but tests are very flaky with the current code, so submitting only the stable parts and will continue working on the full end-to-end testing in subsequent PRs.
- Updated the local snapshot loading from an arbitrary URL + created a simple web server for it. While the previous file could have been used too, this reduces the complexity of on-demand updates for search.
- In the search entrypoint, the new `runSearchInstanceController` will use a Stream for triggering index reloads. This allows better control than the previous scheduling, and also provides an easy feedback through the completer.

### DIFF
--- a/app/lib/search/updater.dart
+++ b/app/lib/search/updater.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:_pub_shared/utils/http.dart';
 import 'package:gcloud/service_scope.dart' as ss;
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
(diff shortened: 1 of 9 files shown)


app/test/service/entrypoint/search_test.dart
```diff
@@ -0,0 +1,59 @@
...
+            renewWait: Duration.zero,
+            snapshot: 'http://localhost:${snapshotServer.server.port}/',
+          );
+          await handlerStartedCompleter.future;
+
+          // force renew
+          final c = Completer();
+          renewController.add(c);
+          await c.future;
+
```
sigurdm, line 41: Don't we need to expect something?

app/lib/search/updater.dart
```diff
@@ -28,14 +29,20 @@ void registerIndexUpdater(IndexUpdater updater) =>
...
-) async {
-  final file = File(path);
-  final content =
-      json.decode(utf8.decode(gzip.decode(await file.readAsBytes())))
-          as Map<String, Object?>;
+/// Loads a search snapshot from an URL or from a local file and builds
+/// an in-memory package index from it.
+Future<InMemoryPackageIndex> loadInMemoryPackageIndexFromUrl(String url) async {
+  late String textContent;
+  if (url.startsWith('http://') || url.startsWith('https://')) {
```
sigurdm, line 36: Where do you call this with an https url? Or is it preparing for something?

sigurdm (approved):

app/test/service/entrypoint/search_test.dart
isoos: At this point the fact that it runs without issues and the future completes is already a good test. I will add search queries and actually changing index at a follow-up pr, when the fake-service is also fixed.

app/lib/search/updater.dart
isoos: More like completeness. Though, we could make the bucket content available through such URL, which would be `https`... and that would remove the special case here. While the intention was not to prepare for something, it can be :)

PR #8963, opened 2025-09-18
Refactor Github Actions + remove mono_repo

isoos:
- Update and completion of #8787.
- New CI workflow is no longer generated, simpler to update.
- Removes `mono_repo` dependency (`tool/run-pub-in-all.sh` to keep the most used feature).
- Updated integration test to ignore some messages that are apparently race conditions in puppeteer. Needs further investigation, but asserts are running and work as expected.


tool/run-pub-in-all.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
sigurdm: Do we need this?

`pub upgrade` in the root should upgrade the entire workspace.

pkg/pub_integration/tool/trigger_puppeteer_download.dart
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
```
sigurdm, line 1: Is this intended to be part of this PR?

pkg/pub_integration/lib/src/fake_test_context_provider.dart
```diff
@@ -17,7 +17,7 @@ import 'test_browser.dart';
 import 'test_scenario.dart';
 
 /// The timeout factor that should be used in integration tests.
-final testTimeoutFactor = 6;
+final testTimeoutFactor = 8;
```
sigurdm, line 20: Belonging to a different PR?

pkg/pub_integration/lib/src/test_browser.dart
```diff
@@ -12,6 +12,9 @@ import 'package:puppeteer/puppeteer.dart';
 import 'package:puppeteer_screenshots/puppeteer_utils.dart';
 import 'package:retry/retry.dart';
 
+/// Filters server or client messages, ignoring and not reporting them as errors.
```
sigurdm, line 15: Belonging to a different PR?

pkg/pub_integration/test/search_update_test.dart
```diff
@@ -18,7 +18,15 @@ void main() {
     final httpClient = http.Client();
 
     setUpAll(() async {
-      fakeTestScenario = await TestContextProvider.start();
+      fakeTestScenario = await TestContextProvider.start(
```
sigurdm, line 21: Belonging to a different PR?

.github/workflows/all-test.yml
```diff
@@ -0,0 +1,120 @@
...
+    - name: Install webp
+      run: sudo apt-get update -yq && sudo apt-get install webp
+    - run: dart test -P presubmit --shard-index ${{matrix.shard}} --total-shards 8
+      working-directory: app/
+
+  test_packages:
+    strategy:
+      fail-fast: false
+      matrix:
+        package:
```
sigurdm, line 72: Not sure if this can be computed dynamically, but we have `dart pub workspace list --json` to list all the packages in the workspace.

isoos:
The seemingly unrelated changes are part of the PR, because the `pkg/pub_integration` tests become extremely flaky when migrating away from the previous model (which had them run in 4 different batches). To stabilize the tests:
- We need to trigger the puppeteer download (if not present), otherwise the download's internal cleanup is being hit by race condition and causes failures.
- There is a race condition coming from puppeteer logs that is absent when running with `-j1` and may be present when running test concurrently. Ignoring the related messages does make the test stable, and I've confirmed that the expectations are still run and met, these messages can be ignored.
- The increase of the timeout may not be needed, but wouldn't hurt.

Issue #8992, opened 2025-10-03
Allow disabling publication of a package from cli when automated publishing is enabled


sigurdm:
That would make it easier to avoid publication mistakes.

isoos:
Couple of notes on the backend side:

- I think the related flag should not be part of `PkgOptions`, since we don't want to expose it publicly.
- The related flag could be part of the existing `AutomatedPublishing` data structure, but we've discussed that it could be toggled independently. We may even rename the `AutomatedPublishing` field to `PublishingSettings`, or just leave the name as-is in this case.
- Or, we could introduce a `ManualPublishing` class with a now-single bool flag, or it could be a top-level `Package.isManualPublishingDisabled` field. In either case we'll need to have new HTTP admin endpoint(s) that update (and get) the field value.

On the package admin page it would make sense to create a new top-level entry after the "Automated publishing" section like the following:

<img width="208" height="232" alt="Image" src="https://github.com/user-attachments/assets/72f39cba-d7f3-46cd-89d8-a5fc8fda496f" />

PR #8995, opened 2025-10-09
Option to disable manual publishing

isoos:
- #8992
- admin page option is behind experimental flag
- created a `ManualPublishingConfig` type for it, that way it is easy to handle it separately from the automatic publishing options.
- The enclosing class name, entity field name and endpoint may be renamed in follow-up PRs.

### DIFF
--- a/app/lib/frontend/templates/views/pkg/admin_page.dart
+++ b/app/lib/frontend/templates/views/pkg/admin_page.dart
@@ -473,14 +473,12 @@ d.Node _manualPublishing(Package package) {
   return d.fragment([
     d.a(name: 'manual-publishing'),
     d.h3(text: 'Manual publishing'),
-    d.markdown(
-      '''
+    d.markdown('''
 Manual publishing, using personal credentials for the `pub` client (`pub login`) .
 
 Disable to prevent accidental publication from the command line.
 
-It is recommended to disable when automated publishing is enabled.''',
-    ),
+It is recommended to disable when automated publishing is enabled.'''),
     d.div(
       classes: ['-pub-form-checkbox-row'],
       child: material.checkbox(
--- a/app/lib/shared/exceptions.dart
+++ b/app/lib/shared/exceptions.dart
@@ -16,6 +16,7 @@ library;
 import 'dart:io';
 
 import 'package:api_builder/api_builder.dart' show ApiResponseException;
+import 'package:pub_dev/shared/urls.dart';
 import 'package:pub_dev/shared/utils.dart';
 
 /// Base class for all exceptions that are intercepted by HTTP handler wrappers.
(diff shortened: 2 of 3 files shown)


app/lib/shared/exceptions.dart
```diff
@@ -572,6 +572,12 @@ class AuthorizationException extends ResponseException {
     'The calling service account is not allowed to publish, because: $reason.\nSee https://dart.dev/go/publishing-with-service-account',
   );
 
+  /// Signaling that the manual publishing was disabled and cannot be authorized.
+  factory AuthorizationException.manualPublishingDisabled() =>
+      AuthorizationException._(
+        'The manual publishing with the `pub` tool is disabled on the package admin page.',
```
sigurdm:
```suggestion
        'Manual publishing with the `pub` tool has been disabled. This usually means this package should be published via automated publishing (see https://dart.dev/tools/pub/automated-publishing). To re-enable manual publishing, go to the package admin page.',
```

app/lib/shared/exceptions.dart
```diff
@@ -572,6 +572,12 @@ class AuthorizationException extends ResponseException {
     'The calling service account is not allowed to publish, because: $reason.\nSee https://dart.dev/go/publishing-with-service-account',
   );
 
+  /// Signaling that the manual publishing was disabled and cannot be authorized.
+  factory AuthorizationException.manualPublishingDisabled() =>
+      AuthorizationException._(
+        'Manual publishing with the `pub` tool has been disabled. This usually means this package should be published via automated publishing (see https://dart.dev/tools/pub/automated-publishing). To re-enable manual publishing, go to the package admin page.',
```
sigurdm: Actually, you would always use the pub tool to publish - no?
```suggestion
        'Manual publishing has been disabled. This usually means this package should be published via automated publishing (see https://dart.dev/tools/pub/automated-publishing). To re-enable manual publishing, go to the package admin page.',
```

app/lib/frontend/templates/views/pkg/admin_page.dart
```diff
@@ -453,6 +468,27 @@ d.Node _automatedPublishing(Package package) {
...
 
+d.Node _manualPublishing(Package package) {
+  final manual = package.automatedPublishing?.manualConfig;
+  return d.fragment([
+    d.a(name: 'manual-publishing'),
+    d.h3(text: 'Manual publishing'),
+    d.markdown(
+      'The manual publishing of new versions using the `pub` tool is enabled by default in all packages. '
+      'Disabling it may protect the package from accidental publishing events when the package is otherwise using '
+      'automated publishing, or in other cases, is discontinued.',
```
sigurdm: How about:

```suggestion
      '''
Manual publishing, using personal credentials for the `pub` client (`pub login`) .

Disable to prevent accidental publication from the command line.

It is recommended to disable when automated publishing is enabled.''',
```

app/lib/shared/exceptions.dart
```diff
@@ -572,6 +572,12 @@ class AuthorizationException extends ResponseException {
     'The calling service account is not allowed to publish, because: $reason.\nSee https://dart.dev/go/publishing-with-service-account',
   );
 
+  /// Signaling that the manual publishing was disabled and cannot be authorized.
+  factory AuthorizationException.manualPublishingDisabled() =>
+      AuthorizationException._(
+        'Manual publishing has been disabled. This usually means this package should be published via automated publishing (see https://dart.dev/tools/pub/automated-publishing). To re-enable manual publishing, go to the package admin page.',
```
sigurdm: It would be great if we could give a link to the package admin page here. Is it possible to pass the package name in the constructor?

PR #8996, opened 2025-10-09
Image proxy

sigurdm:


pkg/image_proxy/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Use latest stable channel SDK.
+FROM dart:3.9.0 AS build
```
isoos, line 2: I think we should also add this file to the test here:
https://github.com/dart-lang/pub-dev/blob/master/app/test/shared/versions_test.dart#L82

pkg/image_proxy/lib/image_proxy_service.dart
```diff
@@ -0,0 +1,306 @@
...
+            final request = await client.getUrl(parsedImageUrl);
+            Timer? timeoutTimer;
+            void scheduleRequestTimeout() {
+              timeoutTimer?.cancel();
+              timeoutTimer = Timer(timeoutDelay, () {
+                request.abort(RequestTimeoutException('No response'));
+              });
+            }
+
+            request.headers.add('user-agent', 'pub-proxy');
```
isoos: Let's add a URL to the user-agent string, pointing to the repository or some other document describing what this proxy does.

pkg/image_proxy/Dockerfile
sigurdm: I think we are going to deploy this separately, and hopefully much more rarely. I think we should allow the sdk versions to drift apart.

isoos: I am a tiny bit worried that we are testing it on an ever-rolling CI SDK, and potentially updating the code if/when needed to pass CI, which may not be backward-compatible.

sigurdm: That's a good point... Ok - I'm convinced to keep the version updated.

sigurdm: Done

PR #8999, opened 2025-10-10
Migrate admin action to set Package.publisherId

isoos:


app/lib/admin/actions/package_publisher_set.dart
```diff
@@ -0,0 +1,61 @@
...
+        skipReanalysis: true,
+        skipVersionsExport: true,
+      );
+      if (currentPublisherId != null) {
+        await purgePublisherCache(publisherId: currentPublisherId);
+      }
+    }
+
+    final pkg = await packageBackend.lookupPackage(packageName);
+    return {'publisherId': pkg!.publisherId};
```
sigurdm: Can we return the previous publisher/uploader also?

isoos: Done.

sigurdm (approved):

PR #9024, opened 2025-10-21
Do not fail Postgres init on staging or prod.

isoos:
This should prevent the log messages about not having the secret entry and also allow us to experiment with the connection string without breaking the rest of the service.


app/lib/service/services.dart
```diff
@@ -251,7 +251,16 @@ Future<R> _withPubServices<R>(FutureOr<R> Function() fn) async {
           await storageService.verifyBucketExistenceAndAccess(bucketName);
         }
 
-        await PrimaryDatabase.tryRegisterInScope();
+        try {
+          await PrimaryDatabase.tryRegisterInScope();
+        } catch (e, st) {
```
sigurdm: Please catch the specific error.

sigurdm (approved):

PR #9026, opened 2025-10-21
Image proxy Add security headers to all responses

sigurdm:


pkg/image_proxy/lib/image_proxy_service.dart
```diff
@@ -104,13 +114,17 @@ final maxImageSize = 1024 * 1024 * 10; // At most 10 MB.
 Future<shelf.Response> handler(shelf.Request request) async {
   try {
     if (request.method != 'GET') {
-      return shelf.Response.notFound('Unsupported method');
+      return shelf.Response.notFound(
+        'Unsupported method',
+        headers: {...securityHeaders},
```
isoos: I think we can write `headers: securityHeaders` here (and later on), since it won't update the headers map. (maybe make it `const` to be sure)

sigurdm: Done

isoos: missing push on the commit?

PR #9046, opened 2025-10-31
Decrease aborted task log level + tag log message.

### LINKED ISSUE #9037: Lower severity of "TaskAborted","message":"<package>/<version> has been aborted."
They cannot be totally prevented, so having a warning is mostly just noise.

We could consider making a log-metric that track them over time.

isoos:
Fixes #9037


sigurdm (approved):
LGTM, but perhaps let's have @jonasfj also agree

jonasfj (approved):

Issue #9036, opened 2025-10-27
Some failed package analyses


sigurdm:
Stacktrace:

```
CRITICAL 2025-10-20T15:17:28.314871397Z [resource.labels.instanceId: 7854299047641478585] pub_worker.process_payload: Failed to process trxact_io / 1.0.1 Error: RequestException(status = 404) {"error":{"code":"NotFound","message":"Could not find `trxact_io/1.0.1`."},"code":"NotFound","message":"Could not find `trxact_io/1.0.1`."} Stack: #0 Client.requestJson.<anonymous closure> (package:api_builder/_client_utils.dart:67:5) <asynchronous suspension> #1 PubApiClient.taskUploadResult (package:_pub_shared/src/pubapi.client.dart:515:7) <asynchronous suspension> #2 RetryOptions.retry (package:retry/retry.dart:131:16) <asynchronous suspension> #3 _analyzePackage (package:pub_worker/src/analyze.dart:254:15) <asynchronous suspension> #4 analyze (package:pub_worker/src/analyze.dart:103:11) <asynchronous suspension> #5 main.<anonymous closure> (file:///home/worker/pub-dev/pkg/pub_worker/bin/pub_worker.dart:36:7) <asynchronous suspension> #6 main (file:///home/worker/pub-dev/pkg/pub_worker/bin/pub_worker.dart:35:5) <asynchronous suspension>
  {
    "insertId": "1ucvksmf4hwcei",
    "jsonPayload": {
[...]

PR #9047, opened 2025-10-31
Extend pub_worker-related NotFound exception with additional context.

isoos:
- #9036
- This should expose if the `runtimeVersion` - for any reason - is not matching the caller worker's runtime.

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -610,7 +610,9 @@ class TaskBackend {
 
     final state = await _db.tasks.lookupOrNull(package);
     if (state == null) {
-      throw NotFoundException.resource('$package/$version');
+      throw NotFoundException.resource(
+        'PackageState($runtimeVersion/$package)',
+      );
     }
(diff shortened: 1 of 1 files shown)


app/lib/task/backend.dart
```diff
@@ -610,7 +610,9 @@ class TaskBackend {
 
     final state = await _db.tasks.lookupOrNull(package);
     if (state == null) {
-      throw NotFoundException.resource('$package/$version');
+      throw NotFoundException.resource(
+        'PackageState($runtimeVersion/$package)',
```
sigurdm, line 614: Don't you want to also include the version here?

isoos: The lookup above depends only on the `runtimeVersion` and the `package` name, the version check comes only later. I think removing the version from the message makes it better focused on the operation.

sigurdm: Fair

sigurdm (approved):

PR #9055, opened 2025-11-04
Move a transaction block into the task db extension methods.

isoos:


app/lib/task/backend.dart
```diff
@@ -1397,9 +1397,7 @@ final class _TaskDataAccess {
     return await withRetryTransaction(_db, (tx) async {
       // Reload [state] within a transaction to avoid overwriting changes
       // made by others trying to update state for another package.
-      final s = await tx.lookupOrNull<PackageState>(
-        PackageState.createKey(_db.emptyKey, runtimeVersion, package),
-      );
+      final s = await tx.tasks.lookupOrNull(package);
```
sigurdm, line 1400: How come the runtimeVersion is no longer part of the key?

isoos: tx.tasks.lookupOrNull is a wrapper method that has it: https://github.com/dart-lang/pub-dev/blob/master/app/lib/task/backend.dart#L1313-L1323

sigurdm (approved):

PR #9057, opened 2025-11-05
Trigger cache invalidation and API export after an advisory affects a package.

### LINKED ISSUE #9056: Do we need to purge cache / re-export the package after a security advisory is digested?
In `ingestSecurityAdvisory` we may update `Package.latestAdvisory`, but we don't immediately purge the cache or export the package entities. Should we do it?

/cc @szakarias

isoos:
Fixes #9056


isoos:
Note: I don't really know if this is important or needed at all, it just felt that we should be probably consistent with the other updates (e.g. also updating the `updated` field).

PR #9065, opened 2025-11-07
Expose license detection edit operations (incomplete, behind experimental).

isoos:
- #5359
- Colors and the interactive feature is incomplete (clicking on the removed icon exposes the removed text segment - or a snippet of it).
- Note: the input from pana seems to be not consistent enough for this feature to be useful at this point, but I'd rather have it behind experimental, so that we can better find detection runs where the offsets or operations are off.

image
image

### DIFF
--- a/app/lib/frontend/templates/package.dart
+++ b/app/lib/frontend/templates/package.dart
@@ -404,6 +404,7 @@ Tab _licenseTab(PackagePageData data) {
           d.span(
             classes: ['license-op-insert'],
             text: text.substring(offset, op.start),
+            attributes: {'title': 'inserted content'},
           ),
         );
         offset = op.start;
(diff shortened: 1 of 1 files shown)


sigurdm:
This is quite awesome!

I found a case where the diff splits in an unlucky place:

image

Could we force it to split in "logical" places only (start of sections or something...)

This view probably has to be enabled by clicking a button or something...

There needs to be more context information given (tool-tips and a color legend, a link to the original licence file we compare with)...

Unclear what happens here in the bottom:
image

Would a side-by-side diff view be easier to comprehend?

app/lib/frontend/handlers/experimental.dart
```diff
@@ -14,6 +14,7 @@ const _publicFlags = <PublicFlag>{
 
 final _allFlags = <String>{
   'dark-as-default',
+  'license',
```
sigurdm: expose-licence-diff

sigurdm (approved):

PR #9072, opened 2025-11-13
Upgrade analysis SDKs.

isoos:


jonasfj (approved):

sigurdm (approved):

pkg/pub_worker/lib/src/sdks.dart
```diff
@@ -68,7 +69,23 @@ final class InstalledSdk {
...
+        p.join(path, 'bin', 'cache', 'flutter.version.json'),
+      );
+      if (await fallbackFile.exists()) {
+        final map = json.decode(await fallbackFile.readAsString()) as Map;
+        final version =
+            (map['flutterVersion'] as String?) ??
+            (map['frameworkVersion'] as String?);
+        sdkVersion = version!;
+      }
+    }
```
sigurdm: This reads a bit like we expect the `version` file to come back at some point (which it isn't). How about something like:

```suggestion
    final String sdkVersion
    if (kind == 'flutter') {
      final versionFile = File(
        p.join(path, 'bin', 'cache', 'flutter.version.json'),
      );
      sdkVersion = (json.decode(await versionFile.readAsString()) as Map)['flutterVersion'] as String;
    } else {
      sdkVersion = await File(p.join(path, 'version')).readAsString().trim();
    }
```

isoos: Merged + updated.

PR #9076, opened 2025-11-14
Updated task tests: working with non-synchronous backends.

isoos:
- Apparently these work with both almost-sync backends (current HEAD) and also with #9069.
- In most cases the change only tracks/awaits the creation of the cloud worker instance, but for some tests I needed to modify the assertion logic too.
- Note: I think we should rewrite the scheduling logic and its async delegation (which is implemented with scheduled microtasks), exposing more control and also some event over it, in order to make these tests more readable.

### DIFF
--- a/app/test/task/task_test.dart
+++ b/app/test/task/task_test.dart
@@ -832,12 +832,6 @@ void main() {
   );
 }
 
-extension<T> on Stream<T> {
-  Future<bool> get isNotEmpty async {
-    return !await this.isEmpty;
-  }
-}
-
 Future<void> upload(
   http.Client client,
   UploadInfo destination,


app/test/task/task_test.dart
```diff
@@ -58,7 +58,9 @@ void main() {
       await taskBackend.backfillTrackingState();
       await clockControl.elapse(minutes: 1);
 
+      final created = cloud.waitForCreated();
       await taskBackend.start();
+      await created;
       await clockControl.elapse(minutes: 5);
```
jonasfj, line 64: What if you did:
```
await Future.wait([
  created,
  await clockControl.elapse(minutes: 5),
]);
```

PR #9079, opened 2025-11-17
Schedule pending tasks on a single thread.

isoos:
Only indent changes for the inner function.

### DIFF
--- a/app/lib/task/scheduler.dart
+++ b/app/lib/task/scheduler.dart
@@ -243,6 +243,7 @@ Future<void> schedule(
       });
     }
 
+    // Creating an instance can be slow, we want to schedule them concurrently.
     await Future.wait(
       (await db.tasks.selectSomePending(selectLimit).toList()).map(
         scheduleInstance,


app/lib/task/scheduler.dart
```diff
@@ -147,104 +147,100 @@ Future<void> schedule(
       _maxInstancesPerIteration,
       max(0, activeConfiguration.maxTaskInstances - instances),
     );
-    await Future.wait(
-      await (db.tasks.selectSomePending(selectLimit)).map<Future<void>>((
-        selected,
-      ) async {
-        pendingPackagesReviewed += 1;
+    await for (final selected in db.tasks.selectSomePending(selectLimit)) {
+      pendingPackagesReviewed += 1;
```
jonasfj: let's `await for`

 and instead use `db.tasks.selectSomePending(selectLimit).toList()` first, so that we can still do all instances concurrently.

jonasfj (approved):

PR #9080, opened 2025-11-18
Experiment: track and wrap Timers and run* blocks in the clock control zone.

isoos:


app/lib/task/clock_control.dart
```diff
@@ -18,9 +18,66 @@ Future<T> withClockControl<T>(
...
-      () async => await fn(clockCtrl),
+    () => runZoned(
+      () => withClock(
+        Clock(clockCtrl._controlledTime),
+        () async => await fn(clockCtrl),
+      ),
+      zoneSpecification: ZoneSpecification(
+        run: <R>(self, parent, zone, R Function() fn) {
+          if (R is Future) {
+            return parent.run(zone, () => clock.instant(() async => fn())) as R;
```
jonasfj: I don't think I have wisdom to know what this does.

app/lib/task/clock_control.dart
```diff
@@ -18,9 +18,66 @@ Future<T> withClockControl<T>(
...
+              current.contains('ClockController._triggerPendingTimers') ||
+              current.contains('ClockController._cancelPendingTimer')) {
+            return parent.createTimer(zone, duration, f);
+          }
+          final clockCtrl = zone[_clockCtrlKey];
+          if (clockCtrl is ClockController) {
+            return clockCtrl._createTimer(duration, f);
+          } else {
+            return parent.createTimer(zone, duration, f);
+          }
```
jonasfj: I think this is remarkably brave, and what does this do?

isoos: The first few lines detects if the timer is created inside the `ClockControl`'s internal timer-related methods or outside of it, and if inside, we return the default timers.

I think this part could be much smarter, e.g. `ClockControl` should create these timers outside of the current zone, maybe in the root zone.

The next part is a sanity check: if we are inside the clock-controlled zone (which we always should be as we fork it inside of it), and creates a timer that is both an internal object we use to track and elapse, and an implementation of `Timer` which can be returned (so that the caller may cancel it, also removing it from the tracking).

app/lib/task/clock_control.dart
isoos: With the recent change (using Zone.root's timers) I was able to remove these.

isoos:
Closing this, as we will go with the steppable task loops approach.

PR #9082, opened 2025-11-18
Refactor task loop: scanning for package updated timestamps and updating tracking state.

isoos:


app/lib/task/loops/scan_packages_updated.dart
```diff
@@ -0,0 +1,93 @@
...
+import 'package:clock/clock.dart';
+
+/// The internal state for deciding which package needs to be updated.
+class ScanPackagesUpdatedState {
+  /// The last time the algorithm checked on a package.
+  final Map<String, DateTime> seen;
+
+  final DateTime since;
+
+  final DateTime nextLongScan;
```
jonasfj: We'd need a better name than `nextLongScan`, what is a "long scan"?

isoos: renamed to `nextLongOverlap` + added some description

PR #9085, opened 2025-11-20
Increase the accepted uncompressed total size of an uploaded archive to 256 MiB

isoos:


sigurdm (approved):

sigurdm:
We also need to update the documentation at https://dart.dev/tools/pub/publishing#:~:text=Keep%20package%20size%20to%20less%20than%20100%20MB%20after%20gzip%20compression

PR #9097, opened 2025-11-28
Refactor + separately test the instance deletion logic in task scheduling

isoos:
- Refactored the instance deletion cycle similarly to the package scan logic in #9082.
- Added tests for the cycle's typical use cases.

### DIFF
--- a/app/lib/task/loops/delete_instances.dart
+++ b/app/lib/task/loops/delete_instances.dart
@@ -69,13 +69,15 @@ Future<DeleteInstancesState> scanAndDeleteInstances(
 
     deletionInProgress[instance.instanceName] = clock.now();
 
-    futures.add(Future.microtask(() async {
-      try {
-        await cloudCompute.delete(instance.zone, instance.instanceName);
-      } catch (e, st) {
-        _log.severe('Failed to delete $instance', e, st);
-      }
-    }));
+    futures.add(
+      Future.microtask(() async {
+        try {
+          await cloudCompute.delete(instance.zone, instance.instanceName);
+        } catch (e, st) {
+          _log.severe('Failed to delete $instance', e, st);
+        }
+      }),
+    );
   }
 
   await Future.wait(futures);


app/lib/task/loops/delete_instances.dart
```diff
@@ -0,0 +1,133 @@
...
+import 'package:basics/basics.dart';
+import 'package:clock/clock.dart';
+import 'package:logging/logging.dart';
+import 'package:pub_dev/task/clock_control.dart';
+import 'package:pub_dev/task/cloudcompute/cloudcompute.dart';
+
+final _log = Logger('pub.task.scan_instances');
+
+/// The internal state for scanning and deleting instances.
+class DeleteInstancesState {
```
jonasfj: nit: final

app/lib/task/loops/delete_instances.dart
```diff
@@ -0,0 +1,133 @@
...
+  final int instances;
+
+  DeleteInstancesNextState({required this.state, required this.instances});
+}
+
+/// Tracks the latest instance-specific deletion's start timestamp
+/// and progress [completer].
+class DeletionInProgress {
+  final DateTime started;
+  final Completer completer;
```
jonasfj: How is anyone to know who is supposed to complete this, when, and who is supposed to await an why?

app/lib/task/loops/delete_instances.dart
```diff
@@ -0,0 +1,133 @@
...
+  // Wait at-least 5 minutes from start of deletion until we remove the
+  // progress tracking, to give the Cloud API some time to reconcile the state.
+  bool isRecent() => started.isAfter(clock.ago(minutes: 5));
+}
+
+/// Calculates the next state of delete instances loop by processing
+/// the input [stream].
+Future<DeleteInstancesNextState> scanAndDeleteInstances(
+  DeleteInstancesState state,
+  Stream<CloudInstance> stream,
```
jonasfj: What is this?
```suggestion
  Stream<CloudInstance> instances,
```

app/lib/task/loops/delete_instances.dart
```diff
@@ -0,0 +1,133 @@
...
+  Future<void> Function(String zone, String instanceName) deleteInstanceFn,
```
jonasfj: Or give it an instance of `CloudCompute`, but a tearoff is also fine.

app/lib/task/loops/delete_instances.dart
```diff
@@ -0,0 +1,133 @@
...
+  bool Function() isAbortedFn, {
```
jonasfj, line 30: IMO, a `Completer<void>` is not a bad choice here.

I don't like using completer, because it's kind of unclear from methods what we're completing, but it does the right job here.

Ideally, we'd have a type like:
```dart
final class Context {
  Future<void> get aborted;
  bool get isAborted;
}
```

But such a thing doesn't exist in `dart:async`, so let's just use `Completer`. Building it ourselves could be interesting one day, but ideally it should be something we all agree on -- let's not try to do it now.

jonasfj: Notably, without a completer, you won't be able to race with `isAborted.future`.

app/lib/task/loops/delete_instances.dart
```diff
@@ -0,0 +1,133 @@
...
+      completer: completer,
+    );
+
+    scheduleMicrotask(() async {
+      try {
+        await deleteInstanceFn(instance.zone, instance.instanceName);
+      } catch (e, st) {
+        _log.severe('Failed to delete $instance', e, st);
+      } finally {
+        completer.complete();
```
jonasfj: From `CloudCompute`:
```dart
  /// Delete the instance with [instanceName] from [zone].
  ///
  /// If this operation fails, it may throw an [Exception] or discard the error.
  /// Systems should assume deletion is best-effort and periodically list
  /// instances to delete instances that are no longer needed.
  Future<void> delete(String zone, String instanceName);
```

Notably, deleting an instance is a long-running operation in GCE, meaning that we start an operation that we then can periodically pool to see how it's doing, whether the operation is pending, failed or finished.

We don't really bother doing this.

See: `GoogleCloudCompute`:
```dart
  @override
  Future<void> delete(String zone, String instanceName) async {
    try {
      await _retryWithRequestId(
        (rId) =>
            _api.instances.delete(_project, zone, instanceName, requestId: rId),
      );
      // Note. that instances.delete() technically returns a custom long-running
      // operation, we have no reasonable action to take if deletion fails.
      // Presumably, the instance would show up in listings again and eventually
      // be deleted once more (with a new operation, with a new requestId).
      // TODO: Await the delete operation...
    } on DetailedApiRequestError catch (e) {
      if (e.status == 404) {
        // If we get a 404, then we shall assume that instance has been deleted.
        // Worst case the instance will eventually show up in listings again and
        // then be deleted once more.
        return;
      }
      rethrow;
    }
  }
```


So I don't think we should track deletion at this level.

-------------

In theory we could, but if we wanted to do that we should add such supported to `GoogleCloudCompute` first.

Instead, I suggest we send a best-effort delete call, log exceptions (but otherwise ignore them), and we avoid calling delete for a particular `instanceName` for 5 minutes or so. Hopefully, after 5 minutes the deletion operation will have completed, and the instance will have been deleted, and instance will have disappeared from the instance listing.

But if not, we'll just try to delete it again, most likely the API will return `404` (which we ignore in `GoogleCloudComplete`), or it'll start a long-running operation, which will possible:
 * Delete the instance, 
 * Fail with a 404 message,
 * Fail with some sort of conflict message, who knows 🤣 

I'm not sure I care, so long as the instance is eventually deleted.

------------------

Yes, we could and in theory probably should care how the long-running delete operation is resolved, failed / finished.

But if it fails, I don't know what we can do other than call `delete()` again, and hope that eventually disappears from the _instance list_.

So perhaps it's better to just keep it simple.

app/lib/task/loops/delete_instances.dart
```diff
@@ -0,0 +1,133 @@
...
+  // progress after we started to delete the instance.
+  final Map<String, DeletionInProgress> deletions;
+
+  DeleteInstancesState({required this.deletions});
+
+  factory DeleteInstancesState.init() => DeleteInstancesState(deletions: {});
+
+  // An arbitrary wait to attempt an async cleanup for our tests, since otherwise we
+  // keep the progress entry for 5 minutes, and ignore the completer status.
+  Future<void> waitSomeSeconds(int seconds) async {
```
jonasfj: if we stick to `final Map<String, DateTime> deletionsInProgress;` then we probably won't need this.

app/lib/task/loops/delete_instances.dart
```diff
@@ -0,0 +1,103 @@
...
+      } finally {
+        completer.complete();
+      }
+    });
+    futures.add(completer.future);
+  }
+
+  return DeleteInstancesNextState(
+    state: DeleteInstancesState(deletions: deletionInProgress),
+    deletionsDone: futures.isEmpty ? Future.value() : Future.wait(futures),
```
jonasfj: I don't like this...

I think that in testing we could either do:
```dart
await Future.delayed(microseconds: 10);
```
Instead of waiting for `deletionsDone`.

----------

Or we could change the tests to wait for `deleteInstanceFn` to have been called N times.

-----------

Or we could `await deleteInstanceFn` in here, with some short timeout like 30s and if it has happened then we log a warning. Nor sure this necessary.

jonasfj: Or we could make the agreement that `deleteInstanceFn` invocations will have been started before we return from `scanAndDeleteInstances`.

But we won't have awaited the futures from `deleteInstanceFn` invocations.

That way, if `deleteInstanceFn` does anything sync, the sync thing will have run.

jonasfj: Maybe, `scanAndDeleteInstances` could return `({DeleteInstancesState nextState, Future<void> done})`.

I don't like having a type for this... I don't know why.

It feels like we're letting internals escape.

-----------

I think we should make the promise that `deleteInstanceFn` will not be invoked after the `Future` from `scanAndDeleteInstances` has been resolved.

We're not promising that `Future`s from `deleteInstanceFn` have been resolved, just that we won't make any more calls to `deleteInstanceFn`.

This requires us to not use `scheduleMicrotask(...`, but instead do:

```dart
Future? f;
try {
  // We've promised to never all deleteInstanceFn after the return
  // from this function! But have not promised to await Futures
  // from deleteInstanceFn in this function!
  f = deleteInstanceFn(...)
} catch (e, st) {
  _log.severe('Failed to delete $instance', e, st);
}
if (f != null) {
  scheduleMicroTask(() async {
    try {
      await f;
    } catch (e, st) {
      _log.severe('Failed to delete $instance', e, st);
    }
  });
}
```

or something like that.

We should declare the invariant in the documentation comment for `scanAndDeleteInstances`.

isoos: Updated as discussed offline.

app/lib/task/loops/delete_instances.dart
```diff
@@ -0,0 +1,82 @@
...
+Future<DeleteInstancesState> scanAndDeleteInstances(
+  DeleteInstancesState state,
+  CloudCompute cloudCompute,
+  bool Function() isAbortedFn, {
+  required int maxTaskRunHours,
+}) async {
+  final keepTreshold = clock.ago(minutes: 5);
+  final deletionInProgress = {
+    ...state.deletions.whereValue((v) => v.isAfter(keepTreshold)),
+  };
```
jonasfj, line 38:
```suggestion
  // Purge deletionsInProgress after 5 minutes, if instance has not disappeared
  // after 5 minutes we'll try to delete it again.
  final keepTreshold = clock.ago(minutes: 5);
  final deletionInProgress = {
    ...state.deletions.whereValue((v) => v.isAfter(keepTreshold)),
  };
```

app/lib/task/loops/delete_instances.dart
```diff
@@ -0,0 +1,82 @@
...
+    deletionInProgress[instance.instanceName] = clock.now();
+
+    final future = Future.microtask(() async {
+      try {
+        await cloudCompute.delete(instance.zone, instance.instanceName);
+      } catch (e, st) {
+        _log.severe('Failed to delete $instance', e, st);
+      }
+    });
+    futures.add(future);
```
jonasfj: nit :D
```suggestion
    futures.add(Future.microtask(() async {
      try {
        await cloudCompute.delete(instance.zone, instance.instanceName);
      } catch (e, st) {
        _log.severe('Failed to delete $instance', e, st);
      }
    }));
```

app/lib/task/backend.dart
```diff
@@ -128,6 +131,13 @@ class TaskBackend {
         await _scanForPackageUpdates(claim, abort: aborted);
       },
     );
+    final deleteLoop = _createLoop(
+      name: 'delete-instances',
+      aborted: aborted,
+      fn: (claim, aborted) async {
+        await _scanForInstanceDeletion(claim, abort: aborted);
+      },
```
jonasfj, line 139: I got confused I thought `_createLoop` would call `fn` in a loop.

Not sure this makes me less confused, but perhaps:

```suggestion
      fn: (claim, aborted) async => await _scanForInstanceDeletion(claim, abort: aborted),
```

Couldn't we make `_createLoop<T>` where `fn` returns `(T, Duration)`, `T` being state and duration being time to wait before next iteration.

And then in `_createLoop<T>` we rename `fn` to `iteration`.

------

Anyways, this is all orthogonal -- ignore me 🤣

app/lib/task/backend.dart
```diff
@@ -321,6 +331,40 @@ class TaskBackend {
...
+      }
+      // Wait until aborted or 10 minutes before scanning again!
+      await abort.future.timeoutWithClock(
+        Duration(minutes: 10),
+        onTimeout: () => null,
+      );
+    }
+  }
+
+  Future<void> _runOneInstanceDeletion({
```
jonasfj, line 357: is this is a function we need?

Or should we rename `scanAndDeleteInstances` to `runOneInstanceDeletionCycle` or `runOneInstanceDeletionIteration`

app/lib/task/backend.dart
isoos: Yeah, I want to do that change in a follow-up PR.

app/lib/task/backend.dart
isoos: I think the loop/cycle/iteration naming is confusing at this point, and my plan is to refactor everything in a way that there is:
- loop (as is right now + its delay logic)
- single operation (maybe we can call it cycle)

But I didn't want to put that in this PR, as it is mostly an unrelated refactor.

jonasfj:
I suggest we ship this!

PR #9099, opened 2025-12-01
Task backend events exposed for testing.

isoos:
- Event-related codes and updated tests from #9097 (without the inner loop refactor).
- Note: the test verifies a different timeline. However, async delays from #9069 still fail this, even with zone-based timer tracking from #9080.


isoos:
Closing as not planned.

PR #9102, opened 2025-12-02
Refactor scan-packages loop kernel to enable direct call.

isoos:
- Follow-up to #9082
- Moves the `ScanPackagesUpdatedState` as a field on `TaskBackend`.

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -17,7 +17,6 @@ import 'package:gcloud/storage.dart' show Bucket;
 import 'package:googleapis/storage/v1.dart' show DetailedApiRequestError;
 import 'package:indexed_blob/indexed_blob.dart' show BlobIndex, FileRange;
 import 'package:logging/logging.dart' show Logger;
-import 'package:meta/meta.dart';
 import 'package:pana/models.dart' show Summary;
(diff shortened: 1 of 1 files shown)


app/lib/task/backend.dart
```diff
@@ -341,6 +328,34 @@ class TaskBackend {
...
+      if (isAbortedFn()) {
+        return;
+      }
+      // Check the package
+      await trackPackage(p, updateDependents: true);
+    }
+  }
+
+  @visibleForTesting
+  Future<void> runOneIterationOfBackgroundLoops() async {
```
sigurdm: Is this ever called?

isoos: Not yet, but tests will use this. I first wanted to also unroll the scheduling loops in this PR, but its a lot things that will be moved, but kept it for now.

sigurdm: Why not just expose the actual _runOneScanPackagesUpdate function to tests? It seems more honest to test that than some wrapper that is otherwise not used in production.

isoos: Removing for now. Wanted to keep it as a placeholder for the name, but can re-add later.

PR #9105, opened 2025-12-03
Refactor: track compute zones bans and round-robin cursor in a separate component.

isoos:


app/lib/task/cloudcompute/zone_tracker.dart
```diff
@@ -0,0 +1,112 @@
...
+import 'package:pub_dev/task/cloudcompute/cloudcompute.dart';
+
+final _log = Logger('pub.task.zone_tracker');
+
+/// Tracks compute zones and custom ban periods.
+class ComputeZoneTracker {
+  final List<String> _zones;
+  final _bannedUntil = <String, DateTime>{};
+
+  int _cursor = 0;
```
jonasfj: What is this, is it possible `_nextZoneIndex` ?

app/lib/task/cloudcompute/zone_tracker.dart
```diff
@@ -0,0 +1,112 @@
...
+
+import 'package:clock/clock.dart';
+import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
+import 'package:pub_dev/task/cloudcompute/cloudcompute.dart';
+
+final _log = Logger('pub.task.zone_tracker');
+
+/// Tracks compute zones and custom ban periods.
+class ComputeZoneTracker {
```
jonasfj: `final`?

app/lib/task/cloudcompute/zone_tracker.dart
```diff
@@ -0,0 +1,112 @@
...
+        return zone;
+      }
+    }
+    return null;
+  }
+
+  @visibleForTesting
+  List<String?> tryPickZones(int n) {
+    return List.generate(n, (_) => tryPickZone());
+  }
```
jonasfj: If this is for testing just move it to the test code, this doesn't need to be here. It doesn't access anything internal.

app/lib/task/cloudcompute/zone_tracker.dart
```diff
@@ -0,0 +1,112 @@
...
+
+  /// Executes [fn] in compute [zone] and handles zone-related exceptions
+  /// with the appropriate bans.
+  Future<void> withZoneAndInstance(
```
jonasfj, line 60: let's not move this logic. Do we need to?

The interactions here are subtle, this is not reusable, and shouldn't attempt to be.

Whether to ban or not ban a zone is a scheduling decision.

I'm not keen on `ComputeZoneTracker` being mutable, nor am I keen on it containing much logic. Isn't it mostly just a container for state between iterations.

Keeping state immutable, means `runSchedulerIteration` (I assume we are working towards) will have to return a new state object, that it can mutate with timers after returning it.

I think spreading logic across files will make it harder to track.


----------

wouldn't it be better for `runSchedulerIteration` to return a `ZoneTrackerState` object? that is effectively immutable.

PR #9115, opened 2025-12-05
Refactoring cloud compute instance creation with external state of banned zones.

isoos:


app/lib/task/backend.dart
```diff
@@ -324,45 +308,33 @@ class TaskBackend {
...
-      if (isAbortedFn()) {
-        return;
-      }
-      // Wait until aborted or 10 minutes before scanning again!
-      await abort.future.timeoutWithClock(
-        Duration(minutes: 10),
-        onTimeout: () => null,
-      );
-    }
+    return Duration(minutes: 10);
```
```suggestion
    return Duration(minutes: 10); // TODO: consider if we scan more frequently.
```

app/lib/task/backend.dart
```diff
@@ -324,45 +308,33 @@ class TaskBackend {
...
-  }) async {
-    _deleteInstancesState = await scanAndDeleteInstances(
+  Future<Duration> _runOneInstanceDeletion(bool Function() isAbortedFn) async {
+    _deleteInstancesState = await runOneDeleteInstancesCycle(
       _deleteInstancesState,
       taskWorkerCloudCompute,
       isAbortedFn,
       maxTaskRunHours: activeConfiguration.maxTaskRunHours,
     );
+    return Duration(minutes: 10);
```
```suggestion
    return Duration(minutes: 10); // TODO: consider if this should be dynamic
```

app/lib/task/scheduler.dart
```diff
@@ -11,41 +11,33 @@ import 'package:pub_dev/shared/configuration.dart';
...
-      await Future.microtask(() {});
-    } else {
-      await abort.future.timeoutWithClock(delay, onTimeout: () => null);
-    }
-  }
-
   // Map from zone to DateTime when zone is allowed again
   final zoneBannedUntil = <String, DateTime>{
     for (final zone in compute.zones) zone: DateTime(0),
+    ...state.zoneBannedUntil,
```
jonasfj, line 40: It's remove the old bans that are too old, so `zoneBannedUntil` can't grow indefinitely, should compute zones over time be large.

app/lib/task/scheduler.dart
```diff
@@ -63,168 +55,169 @@ Future<void> schedule(
...
+    } on ZoneExhaustedException catch (e, st) {
+      // A zone being exhausted is normal operations, we just use another
+      // zone for 15 minutes.
+      _log.info(
+        'zone resources exhausted, banning ${e.zone} for 30 minutes',
+        e,
+        st,
+      );
+      // Ban usage of zone for 30 minutes
+      banZone(e.zone, minutes: 30);
```
jonasfj, line 148: 15 or 30 minutes, did we change this recently.

Code and comments are out of sync.

app/lib/task/scheduler.dart
isoos: Note: the current code relies on the zoneBannedUntil to contain all available zones. I think it is better to rely on the zones list directly, will refactor.

isoos: Actually, will do this in a follow-up

PR #9120, opened 2025-12-08
sparse zoneBannedUntil in instance creation

isoos:
Note: this will remove the timed-out bans, and uses the compute's zone list to check currently available zones, but removes the extra check for zone names.

### DIFF
--- a/app/lib/task/scheduler.dart
+++ b/app/lib/task/scheduler.dart
@@ -70,7 +70,10 @@ Future<(CreateInstancesState, Duration)> runOneCreateInstancesCycle(
   // Determine which zones are not banned
   final allowedZones =
       compute.zones
-          .where((zone) => (zoneBannedUntil[zone] ?? DateTime(0)).isBefore(now))
+          .where(
+            (zone) =>
+                (zoneBannedUntil[zone] ?? DateTime(0)).isBefore(clock.now()),
+          )
           .toList()
         ..shuffle(rng);
   var nextZoneIndex = 0;


app/lib/task/scheduler.dart
```diff
@@ -36,18 +37,14 @@ Future<(CreateInstancesState, Duration)> runOneCreateInstancesCycle(
 }) async {
   // Map from zone to DateTime when zone is allowed again
   final zoneBannedUntil = <String, DateTime>{
-    for (final zone in compute.zones) zone: DateTime(0),
-    ...state.zoneBannedUntil,
+    ...state.zoneBannedUntil.whereValue((v) => v.isAfter(clock.now())),
```
jonasfj, line 40:
```suggestion
// Map from zone to DateTime when zone is allowed again
final now = clock.now();
final zoneBannedUntil = <String, DateTime>{
  for (final zone in compute.zones) zone: DateTime(0)
    ...state.zoneBannedUntil.whereValue((v) => v.isAfter(now)),
```

And then not do the other changes.

isoos: Curious: why not chose the zones always from the currently available ones?

The current way seems to lag one cycle behind if/when a zone gets in a known unavailable status.

app/lib/task/scheduler.dart
```diff
@@ -72,10 +69,7 @@ Future<(CreateInstancesState, Duration)> runOneCreateInstancesCycle(
 
   // Determine which zones are not banned
   final allowedZones =
-      zoneBannedUntil.entries
-          .where((e) => e.value.isBefore(clock.now()))
-          .map((e) => e.key)
-          .toList()
+      compute.zones.where((zone) => !zoneBannedUntil.containsKey(zone)).toList()
```
jonasfj:
```suggestion
      compute.zones.where((zone) => (zoneBannedUntil[zone] ?? DateTime(0)).isBefore(now)).toList()
```

PR #9124, opened 2025-12-09
Refactor task tests: increment time but do not trigger timers.

isoos:
- Passes as-is, delayed datastore tests and the current working state of task sql migration (revealed a bug!).
- Unrelated elapse-based tests will be migrated separately and elapse methods will be removed at that point.

### DIFF
--- a/app/lib/task/clock_control.dart
+++ b/app/lib/task/clock_control.dart
@@ -469,6 +469,31 @@ final class ClockController {
       await Future.delayed(Duration(microseconds: 0));
     }
   }
+
+  void incrOffset({int hours = 0, int minutes = 0, int seconds = 0}) {
+    _offset += Duration(hours: hours, minutes: minutes, seconds: seconds);
+  }
+
+  Future<void> incrUntil(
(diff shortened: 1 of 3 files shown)


app/lib/task/clock_control.dart
```diff
@@ -469,6 +469,31 @@ final class ClockController {
       await Future.delayed(Duration(microseconds: 0));
     }
   }
+
+  void incrOffset({int hours = 0, int minutes = 0, int seconds = 0}) {
+    _offset += Duration(hours: hours, minutes: minutes, seconds: seconds);
+  }
+
+  Future<void> incrUntil(
```
jonasfj, line 477: I think we should rename these: `elapse` and `elapseUntil`, and then we should remove everything from clock_control.dart but I'm okay doing that in a followup.

jonasfj (approved):

app/lib/task/clock_control.dart
isoos: Let's do it in a follow-up, since it also involves unrelated tests.

PR #9126, opened 2025-12-10
Migrating task model and backend to use typed_sql.

isoos:


app/lib/database/database.dart
```diff
@@ -179,3 +181,23 @@ Future<void> _dropCustomDatabase(String url, String dbName) async {
...
+        } on TransactionAbortedException catch (e) {
+          if (e.reason is ResponseException) {
+            // TODO: we should keep and use the original stacktrace in typed_sql's exception
+            throw e.reason;
+          }
+          rethrow;
+        }
+      },
+      maxAttempts: 3,
+      retryIf: (e) => e is DatabaseException,
```
sigurdm: I guess we should not retry if e is a https://pub.dev/documentation/typed_sql/latest/typed_sql/OperationException-class.html ?

isoos: I think, since this also covers deadlocks or timeouts, we should retry on it too.

isoos:
I think this is ready for review.

app/lib/task/backend.dart
```diff
@@ -1233,142 +1333,93 @@ List<Version> _versionsToTrack(package_api.PackageData data) {
...
-extension TaskTransactionWrapperExt on TransactionWrapper {
-  _TaskTransactionDataAcccess get tasks => _TaskTransactionDataAcccess(this);
+/// Low-level, narrowly typed data access methods for [Task] entity.
+extension TaskDatabaseExt on Database<PrimarySchema> {
+  _TaskDataAccess get tasksAccess => _TaskDataAccess(this);
 }
 
 final class _TaskDataAccess {
-  final DatastoreDB _db;
+  late final Database<PrimarySchema> _schema;
```
```suggestion
  late final Database<PrimarySchema> _db;
```
or am I the only one that `_schema` confuses.

app/lib/task/backend.dart
```diff
@@ -1233,142 +1333,93 @@ List<Version> _versionsToTrack(package_api.PackageData data) {
...
   Stream<({String package})> listAllForCurrentRuntime() async* {
-    final query = _db.query<PackageState>()
-      ..filter('runtimeVersion =', runtimeVersion);
-    await for (final ps in query.run()) {
-      yield (package: ps.package);
+    final query = _schema.tasks
+        .where((task) => task.runtime_version.equalsValue(runtimeVersion))
+        .select((task) => (task.package,));
+    await for (final row in query.stream()) {
+      yield (package: row);
```
```suggestion
    return _schema.tasks
        .where((task) => task.runtime_version.equalsValue(runtimeVersion))
        .select((task) => (task.package,))
        .stream()
        .map((p) => (package: p));
```

app/lib/task/backend.dart
```diff
@@ -1233,142 +1333,93 @@ List<Version> _versionsToTrack(package_api.PackageData data) {
...
-      ..filter('finished >=', since)
-      ..order('-finished');
-    await for (final s in query.run()) {
-      yield (package: s.package, finished: s.finished);
+    final query = _schema.tasks
+        .where((task) => task.finished >= since.asExpr)
+        .orderBy((task) => [(task.finished, Order.descending)])
+        .select((task) => (task.package, task.finished));
+    await for (final row in query.stream()) {
+      yield (package: row.$1, finished: row.$2);
```
```suggestion
    await for (final (package, finished) in query.stream()) {
      yield (package: package, finished: finished);
```

jonasfj: There was at some point some experimental support in `package:typed_sql` for returning named-records.

It would scan you're entire code base looking for all named-records and generate type-parameterized extension methods that would allow one to use them as return values in database queries.

There is really nothing wrong with doing:

```dart
    final query = _schema.tasks
        .where((task) => task.finished >= since.asExpr)
        .orderBy((task) => [(task.finished, Order.descending)])
        .select((task) => (package: task.package, finished: task.finished))
        .stream(); // <-- this extension method won't exist
```

It could totally work, we would just need to generate a tiny extension method for:
```dart
extension<A, B> QueryPackageFinished on Query<(package: Expr<A>, finished: Expr<B>)> {
  ...
}
```

Or something along those lines, that would have been really nice to have here.

It's a bit exponential, but one could generate such extension method for all named-records of using a subset of field names from a defined table :D

Or more likely, we should support it somehow else like having people explicitly declare what named-records they want extension methods for.
```dart
@namedRecordQuery
typedef<A, B> PackageFinishedRow = (package: A, finished: B);
```

Then have `@namedRecordQuery` be a special annotation that `package:typed_sql` code generator picks up on. And generates pre-requisite extension methods.

Okay, sorry that got off topic.

app/lib/task/backend.dart
```diff
@@ -1233,142 +1333,93 @@ List<Version> _versionsToTrack(package_api.PackageData data) {
...
-      ..filter('dependencies =', package)
-      ..filter('lastDependencyChanged <', publishedAt);
-    await for (final ps in query.run()) {
-      yield (package: ps.package);
+    final query = _schema.task_dependencies
+        .join(_schema.tasks)
+        .usingTask()
+        .where(
+          (dep, task) =>
+              dep.runtime_version.equalsValue(runtimeVersion) &
```
jonasfj, line 1413:
```dart
              dep.runtime_version.equalsValue(runtimeVersion) &
```

If you ever get tired of writing this, you can just do:
```dart
extension CurrentTasksExt on Database<PrimarySchema> {
  /// [tasks], but filtered the current [runtimeVersion].
  Query<Task> currentTask =>
    tasks.where((t) => t.runtime_version.equalsValue(runtimeVersion));

  /// [task_dependencies], but filtered the current [runtimeVersion].
  Query<Task> currentTaskDependencies =>
    task_dependencies.where((d) => d.runtime_version.equalsValue(runtimeVersion));
}
```

Chaining multiple `.where()` calls is practically free, should result in the same SQL at-least.

jonasfj: Okay, maybe applying the `where` before the `join` could be different, not sure if it's better or worse 🤣

app/lib/task/backend.dart
```diff
@@ -1378,87 +1429,79 @@ final class _TaskDataAccess {
     String package,
     DateTime publishedAt,
   ) async {
-    return await withRetryTransaction(_db, (tx) async {
+    return await _schema.transact(() async {
       // Reload [state] within a transaction to avoid overwriting changes
       // made by others trying to update state for another package.
-      final s = await tx.tasks.lookupOrNull(package);
+
+      final s = await _schema.tasks.byKey(runtimeVersion, package).fetch();
```
```suggestion
      final s = await _schema
          .tasks
          .byKey(runtimeVersion, package)
          .where((t) => t.last_dependency_changed.isBeforeValue(publishedAt))
          .fetch();
```
or don't, this is really a nit.

app/lib/task/backend.dart
```diff
@@ -1378,87 +1429,79 @@ final class _TaskDataAccess {
...
-      if (s.lastDependencyChanged!.isBefore(publishedAt)) {
-        tx.insert(
-          s
-            ..lastDependencyChanged = publishedAt
-            ..pendingAt = derivePendingAt(
-              versions: s.versions!,
-              lastDependencyChanged: publishedAt,
-            ),
-        );
+      if (s.last_dependency_changed.isBefore(publishedAt)) {
```
jonasfj: You can move `last_dependency_changed.isBefore(publishedAt)` into the `byKey` query.

app/lib/task/backend.dart
```diff
@@ -1378,87 +1429,79 @@ final class _TaskDataAccess {
...
-      final state = await tx.tasks.lookupOrNull(packageName);
-      if (state != null) {
-        state.pendingAt = initialTimestamp;
-        tx.insert(state);
-      }
-    });
+    await _schema.tasks
+        .where(
+          (task) =>
+              task.runtime_version.endsWithValue(runtimeVersion) &
```
```
              task.runtime_version.endsWithValue(runtimeVersion) &
                                   // ^ ?
```

Is `endsWithValue` intentional? Also did we implement that correctly across all databases 🙈

app/lib/task/backend.dart
```diff
@@ -662,58 +706,97 @@ class TaskBackend {
...
+          isValid = true;
+        } on ResponseException {
+          _log.shout(
+            'pub_worker responses with summary.allDependencies containing "$dependency"'
+            ' in package "$package" version "$version"',
+          );
         }
+
+        if (!isValid) continue;
+        await schema.task_dependencies
```
jonasfj: bulk insert is a missing feature here!

It's kind of bad to do a lot of inserts in a loop, rather than doing them concurrently (using multiple connections). Granted then we'd have to do it outside the transaction.

But doing them concurrently is also a bit sad (because it'd be outside the transaction), so we really should some day look into bulk insert.

jonasfj: Maybe a sanity limit here is reasonable, I think we had that before.

1000 inserts in a loop is not a good idea inside a transaction.

10000 inserts in a loop inside a transaction could be a bit bad.

Also there is no limit to `existingDependencyList`, it could be 70k.

We could sort summary?.allDependencies  and only take the first K (500) entries or so?

jonasfj (commented):
I don't I think have any major concerns.

The dependency tracking insert loop is potentially a bit bad.

app/lib/task/backend.dart
isoos: Oh, bummer, great catch!

app/lib/task/backend.dart
isoos: Note: implemented limits, but we may also want to support "on conflict" clauses here.

jonasfj (approved):

PR #9137, opened 2025-12-19
Use Image proxy (with experiment)

sigurdm:
Adds configuration to locate the image proxy and the secret.

Adds an experiment to enable the image proxy

If the experiment is enabled:
  - Transform all images in rendered markdown to use proxied urls. (Except for those from our own properties.)
 
Still not enabled for dartdoc. It has its own sanitization pipeline I believe.

Also a few fixes to the image_proxy deployment script.

### DIFF
--- a/app/lib/service/image_proxy/backend.dart
+++ b/app/lib/service/image_proxy/backend.dart
@@ -32,7 +32,8 @@ class ImageProxyBackend {
 
   static Future<ImageProxyBackend> create() async {
     final instance = ImageProxyBackend._();
-    await instance._dailySecret.update();
+
+    unawaited(instance._dailySecret.update());
     return instance;
   }
 


app/lib/service/image_proxy/backend.dart
```diff
@@ -0,0 +1,102 @@
...
+ImageProxyBackend get imageProxyBackend =>
+    ss.lookup(#_imageProxyBackend) as ImageProxyBackend;
+
+/// Represents the backend for the Youtube handling and related utilities.
+class ImageProxyBackend {
+  ImageProxyBackend._();
+
+  static Future<ImageProxyBackend> create() async {
+    final instance = ImageProxyBackend._();
+    await instance._dailySecret.update();
```
jonasfj: Can we make this unawaited (and catch any exception, so that we can log it), I don't want something to throw an Exception during startup.

If CloudKMS is down, we want `ImageProxyBackend.imageProxyUrl` to return `null`.

We do not want the server to refuse starting. We'll just accept that we don't have images, which is a much better way to degrade.

sigurdm: I added unawaited.

I believe CachedValue has logic for catching and logging errors of the updateFn

app/lib/shared/configuration.dart
```diff
@@ -218,6 +218,10 @@ final class Configuration {
   /// The rate limits for auditable operations.
   final List<RateLimit>? rateLimits;
 
+  /// The Cloud KMS HMAC key version used to sign image proxy URLs.
+  final String? imageProxyHmacKeyVersion;
+  final String? imageProxyServiceBaseUrl;
```
isoos, line 223:
```suggestion
  /// The identifier of the Cloud KMS HMAC key version used to sign image proxy URLs.
  final String? imageProxyHmacKeyVersion;
  /// The public base URL of the image proxy service.
  final String? imageProxyServiceBaseUrl;
```

PR #9150, opened 2026-01-06
Sanity check the report's moderation subject.

isoos:


app/lib/admin/models.dart
```diff
@@ -391,6 +396,9 @@ class ModerationSubject {
         }
         final package = pvParts.first;
         final version = pvParts[1];
+        if (package.length > 64 || version.length > 64) {
```
sigurdm, line 399: Can we use named constants here instead of magic numbers?

isoos: Will do in a follow-up.

PR #9162, opened 2026-01-13
Script skeleton for sandbox runner logic.

isoos:


pkg/pub_worker/bin/sandbox_runner.dart
```diff
@@ -0,0 +1,112 @@
...
+      Platform.environment['SANDBOX_NETWORK_ENABLED'] == 'true';
+
+  final passThroughKeys = {
+    'CI',
+    'NO_COLOR',
+    'PATH',
+    'XDG_CONFIG_HOME',
+    'FLUTTER_ROOT',
+    'PUB_ENVIRONMENT',
+    'PUB_HOSTED_URL',
```
jonasfj: Should we not pass eveyrthing through? Not that I mind this is perhaps fine. A sandbox should limit what goes through.

pkg/pub_worker/bin/sandbox_runner.dart
```diff
@@ -0,0 +1,112 @@
...
+
+    print('Write mounts:');
+    for (final m in writeMounts) {
+      print('- $m');
+    }
+
+    print('Needs network: $needsNetwork');
+  }
+
+  final p = await Process.start(
```
jonasfj, line 79: maybe catch IOException

isoos: Can we do anything about it?
Should we do anything about it?

jonasfj: Printing a prettier error I guess.

isoos: Let's push that to a later point if/when we are unhappy if there is such an error inside the logs. We may be swallowing important information, which we could just expose instead.

PR #9164, opened 2026-01-14
Setup gvisor script + running in worker docker.

isoos:
- Adapting https://github.com/dart-lang/pub-dev/blob/master/pkg/runsc/tool/fetch-gvisor.sh but also applying similar steps as in `setup-webp.sh`
- Running it in the worker's docker, downloading it to its separate directory, so that we may not mount it inside the sandbox environment alongside with e.g. `webp`.

### DIFF
--- a/tool/setup-gvisor.sh
+++ b/tool/setup-gvisor.sh
@@ -18,8 +18,8 @@ ARCH=$(uname -m)
 # See more: https://gvisor.dev/docs/user_guide/install/#specific-release
 URL=https://storage.googleapis.com/gvisor/releases/release/20260112/${ARCH}
 
-curl -o runsc "${URL}"
-curl -o runsc.sha512 "${URL}.sha512"
+curl -o runsc "${URL}/runsc"
+curl -o runsc.sha512 "${URL}/runsc.sha512"
 sha512sum -c runsc.sha512
 rm -f runsc.sha512
 


tool/setup-gvisor.sh
```diff
@@ -0,0 +1,22 @@
...
+  echo "Target directory argument is missing."
+  exit 1
+fi
+
+TARGET_DIRECTORY=${1}
+mkdir -p "${TARGET_DIRECTORY}"
+cd "${TARGET_DIRECTORY}"
+
+ARCH=$(uname -m)
+URL=https://storage.googleapis.com/gvisor/releases/release/latest/${ARCH}
```
sigurdm: Should we consider using a fixed version and do manual upgrades when needed?
Then we could even bake in the hash code.

Not that I don't trust gvisor to put out good stuff. But locked dependencies will make this stable.

tool/setup-gvisor.sh
```diff
@@ -0,0 +1,22 @@
...
+
+wget ${URL}/runsc ${URL}/runsc.sha512
```
sigurdm: Not that it really matters, but we already have curl installed,  using that would make us need one tool less.

tool/setup-gvisor.sh
isoos: Also added a tracking issue to remind us to update it periodically, and the best way would be to create a test for it (not urgent right now though) https://github.com/dart-lang/pub-dev/issues/9165

jonasfj: maybe, but honestly see:
https://gvisor.dev/docs/user_guide/install/#latest-release

I can't find the list of stable releases, or maybe they do one every day or so.

isoos: It seems to be weekly: https://github.com/google/gvisor/tags
I think pinning is good, and occasionally upgrading should be fine.

PR #9168, opened 2026-01-15
Utility to setup rootfs for sandbox

jonasfj:


.dockerignore
```diff
@@ -1,2 +1,3 @@
 app/.pub
 .git
+.dart_tool/
```
jonasfj, line 3: is there ever any reason to copy these into docker images?

I'm thinking they might container precompiled build_runner binaries we might not need.

isoos (approved):

PR #9170, opened 2026-01-16
AOT compile pub_worker

jonasfj:


Dockerfile.worker
```diff
@@ -52,6 +52,10 @@ RUN dart pub get
 RUN dart compile exe -o bin/sandbox_runner bin/sandbox_runner.dart
 RUN DEBUG_SANDBOX_RUNNER=true bin/sandbox_runner ls
 
+# AOT compile pub_worker
+RUN tool/build.sh
```
isoos: I think this command is not that complex and would be better in a one-liner.

isoos (changes requested):

Dockerfile.worker
```diff
@@ -52,6 +52,11 @@ RUN dart pub get
 RUN dart compile exe -o bin/sandbox_runner bin/sandbox_runner.dart
 RUN DEBUG_SANDBOX_RUNNER=true bin/sandbox_runner ls
 
+# AOT compile pub_worker
+RUN mkdir -p build
+RUN dart compile exe -o "build/pub_worker" "bin/pub_worker.dart"
```
isoos, line 57: Should we just put it in `/home/worker/bin`? That is my intent in https://github.com/dart-lang/pub-dev/pull/9172 for dartdoc...

PR #9176, opened 2026-01-20
Use the precompiled dartdoc with pana in the pub_worker.

isoos:


pkg/pub_worker/bin/pub_worker_subprocess.dart
```diff
@@ -98,6 +98,15 @@ Future<void> main(List<String> args) async {
...
   final detected = await _detectSdks(pkgDir.path);
 
+  final dartdocDirEnvValue = Platform.environment['DARTDOC_DIR'];
+  final dartdocCommand = dartdocDirEnvValue == null
+      ? null
+      : <String>[
+          p.join(dartdocDirEnvValue, 'build', 'dartdoc'),
+          '--resources-dir',
+          p.join(dartdocDirEnvValue, 'lib', 'resources'),
+        ];
```
jonasfj: This feels a bit hacky, we should just use `dartdoc` from `PATH`.

Then update our image to put `dartdoc` into `/home/worker/bin` or something like that.

And then make an env var `DARTDOC_DATA` or `DARTDOC_DATA_DIR` pointing at where `lib/resources` are located.

-----

But I propose that we do such cleanup work later.

isoos: Updated to use `DARTDOC_BINARY` and `DARTDOC_RESOURCES_DIR`, at least reducing the hacky part. Agreed on later cleanup.

PR #9180, opened 2026-01-21
Run some pub_worker subprocess in gvisor sandbox.

isoos:


pkg/pub_worker/bin/sandbox_runner.dart
```diff
@@ -61,30 +81,30 @@ Future<void> main(List<String> args) async {
...
-    args.first,
-    args.skip(1).toList(),
-    mode: ProcessStartMode.inheritStdio,
-    environment: environment,
-    includeParentEnvironment: false,
-    runInShell: false,
-    workingDirectory: currentWorkingDir,
+  final p = await runsc(
+    runscExecutable: '/home/worker/gvisor/runsc',
+    env: {'TERM': 'xterm', ...environment},
```
jonasfj, line 86: might be wise to also set `HOME: /home/sandbox/` or something like that...

isoos: Let's keep it as pass-through for now, restricting it may come later.

PR #9183, opened 2026-01-23
Additional network config + debug log for runsc.

isoos:


pkg/pub_worker/bin/sandbox_runner.dart
```diff
@@ -103,6 +103,7 @@ Future<void> main(List<String> args) async {
       ),
     ],
     processStartMode: ProcessStartMode.inheritStdio,
+    debugLogDir: Platform.environment['SANDBOX_DEBUG_LOG_DIR'],
```
jonasfj, line 106: does pana need to provide this?

Can't we just decide that we write this to where we want? Is it something we want to show up in the pana log?

isoos: This is more for us (until we figure out the network stack).

PR #9197, opened 2026-01-30
Re-enable pub_worker sandbox with docker run --privileged.

isoos:


isoos:
Verified on staging: both Dart- and Flutter-packages were analyzed. Checked various results, successes and failures (e.g. version resolution or analysis errors), seems to work as expected.

PR #9201, opened 2026-02-04
AOT-compiled pub_worker_subprocess

isoos:


pkg/pub_worker/lib/src/analyze.dart
```diff
@@ -170,14 +165,28 @@ Future<void> _analyzePackage(
 
     log.writeln('## Running analysis for "$package" version "$version"');
     log.writeln('STARTED: ${clock.now().toUtc().toIso8601String()}');
-    log.writeln(''); // empty-line before the next headline
 
+    var subprocessCommand = 'dart';
+    var subprocessParams = <String>[
+      p.absolute('./bin/pub_worker_subprocess.dart'),
+    ];
```
jonasfj, line 172: This is for testing to work, right?

See suggestion for `pubWorkerSubprocessCommand` above, then we don't have to do things like this.

pkg/pub_worker/lib/src/analyze.dart
```diff
@@ -170,14 +165,28 @@ Future<void> _analyzePackage(
...
+    final subprocessBinary =
+        Platform.environment['PUB_WORKER_SUBPROCESS_BINARY'];
+    if (subprocessBinary != null) {
+      subprocessCommand = subprocessBinary;
+      subprocessParams = <String>[];
+    }
+    log.writeln(
+      'PUB_WORKER_SUBPROCESS_BINARY: $subprocessCommand ${subprocessParams.join(' ')}',
```
jonasfj: this doesn't seem like something we should write to log... this will go into logs visible to end-users.

This is more of an internal thing.

pkg/pub_worker/lib/src/analyze.dart
```diff
@@ -47,10 +47,7 @@ bool _retryIf(Exception e) =>
     e is ClientException ||
     e is TimeoutException;
 
-Future<void> analyze(
-  Payload payload, {
-  String pubWorkerSubprocessPath = './bin/pub_worker_subprocess.dart',
```
jonasfj, line 52: Why don't we keep `pubWorkerSubprocessPath` and rename it `pubWorkerSubprocessCommand` or something like that.

We can default it to `'pub_worker_subprocess'`, and we simply require that whatever this command is, it is something that is either an absolute path or something that can be resolved from PATH.

In docker image it's easy to compile to a `pub_worker_subprocess` binary and put it on PATH.

In `test/analyze_test.dart` we can pass an explicit value for `pubWorkerSubprocessCommand` and we can just make it bash script in `tool/` that wraps `dart bin/pub_worker_subprocess.dart`.

Dockerfile.worker
```diff
@@ -56,14 +56,23 @@ ENV DARTDOC_RESOURCES_DIR="${DARTDOC_DIR}/lib/resources"
...
-RUN mkdir -p build
+ENV PUB_WORKER_BUILD_DIR="/home/worker/pub-dev/pkg/pub_worker/build"
+RUN mkdir -p "${PUB_WORKER_BUILD_DIR}"
 RUN dart compile exe -o "build/pub_worker" "bin/pub_worker.dart"
-ENV PATH="/home/worker/pub-dev/pkg/pub_worker/build:${PATH}"
+ENV PATH="${PUB_WORKER_BUILD_DIR}:${PATH}"
+
+# AOT compile the pub_worker_subprocess
+RUN dart compile exe -o "build/pub_worker_subprocess" "bin/pub_worker_subprocess.dart"
+ENV PUB_WORKER_SUBPROCESS_BINARY="${PUB_WORKER_BUILD_DIR}/pub_worker_subprocess"
```
jonasfj, line 71: This puts it in `/home/worker/bin`, right? Which is already in `PATH`.

So why not simply refer to it as `pub_worker_subprocess` available from `PATH`.
We probably don't even need an env var for `PUB_WORKER_SUBPROCESS_BINARY`.

Since default is to assume that it's available from `PATH` and that's it's called `pub_worker_subprocess`.

isoos: > This puts it in /home/worker/bin, right?

No, `PUB_WORKER_BUILD_DIR` is `/home/worker/pub-dev/pkg/pub_worker/build` (see above).

> So why not simply refer to it as pub_worker_subprocess available from PATH.

`/home/worker/bin` will be mounted into the sandbox, while `/home/worker/pub-dev/pkg/pub_worker/build` will not be mounted. One less item to worry about.

pkg/pub_worker/lib/src/analyze.dart
isoos: Removed.

Dockerfile.worker
jonasfj: > /home/worker/bin will be mounted into the sandbox, while /home/worker/pub-dev/pkg/pub_worker/build will not be mounted. One less item to worry about.

Good point

PR #9209, opened 2026-02-09
New style for downloads sparkline.

### LINKED ISSUE #8301: fix: downloads sparkline hover is hidden under mouse
When you hover the sparkline the number tracks under your cursor (rather than the line), so you can't move the mouse off from over the number to see it.

image

Screenshots don't show the mouse cursor, so it's hard to show you the bug.

FYI @szakarias

isoos:
- Fixes #8301
- Built on top of @bartvanandel's suggestion, but the offline discussion voted for the top-right position.


bartvanandel:
Regarding "top-right position": while `translateX(-50%)` can be used to center an element, `translateX(-10%)` may not be exactly what you intend, because it depends on the width of the rendered element. For instance, 10 downloads would only result in a tiny translation, whereas 100,000 downloads would result in a lot more. If you want to offset the position with a fixed amount (which I think is what you're after, e.g. to align the number itself, not the popup border), better use `rem` units.

PR #9218, opened 2026-02-13
feat: Improve the output of release emails

kevmoo:
Format the changelog excerpt like code

### DIFF
--- a/app/lib/service/email/email_templates.dart
+++ b/app/lib/service/email/email_templates.dart
@@ -42,8 +42,8 @@ With appreciation, the Dart package site admin''';
 String _changeLogDiv(String excerpt) =>
     '''
 Excerpt of the changelog:<br/>
-<div style="background-color: #f6f8fa; border-radius: 6px; padding: 16px; overflow: auto; border: 1px solid #d0d7de; margin: 20px 0;">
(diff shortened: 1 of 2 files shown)


app/lib/service/email/email_templates.dart
```diff
@@ -39,6 +39,13 @@ Thanks for your contributions to the Dart community!
 With appreciation, the Dart package site admin''';
 }
 
+String _changeLogDiv(String excerpt) =>
+    '''
+Excerpt of the changelog:<br/>
+<div style="background-color: #f6f8fa; border-radius: 6px; padding: 16px; overflow: auto; border: 1px solid #d0d7de; margin: 20px 0;">
```
sigurdm: It hurts me to see sizes in px. I wonder if we can use font-relative units, such that the default font size of the email client has effect. And if that is worth it.

```
Excerpt of the changelog:<br/><div style="background-color: #f6f8fa; border-radius: .5rem; padding: 2rem; overflow: auto; border: 1px solid #d0d7de; margin: 3rem 0;">
  <pre style="margin: 0; font-family: 'Courier New', Courier, monospace; font-size: 1rem; line-height: 1.5; color: #24292f;">
```

Image

isoos: Updated.

PR #9220, opened 2026-02-16
Avoid inline styles in chart scripts

sigurdm:


pkg/web_app/lib/src/widget/downloads_chart/widget.dart
```diff
@@ -473,8 +473,8 @@ void drawChart(
   chart.append(cursor);
 
   void hideCursor(_) {
-    cursor.setAttribute('style', 'opacity:0');
-    toolTip.setAttribute('style', 'opacity:0;position:absolute;');
```
isoos, line 477: position: absolute is missing.

sigurdm: I wonder if that matters when the opacity is 0.

isoos: Maybe not, let's try it.

jonasfj: it might matter a bit if there is a CSS transition on the opacity :D

PR #9265, opened 2026-03-06
Increased scaling threshold and cool-down for search instances.

isoos:


search.yaml
```diff
@@ -20,10 +20,17 @@ automatic_scaling:
   min_num_instances: 10
   max_num_instances: 30
```
sigurdm: I think 30 is slightly too high for our current quota.
20?

PR #9274, opened 2026-03-11
Script to create always forward-rolling migrations SQL with atlas.

isoos:
Note: the migrations scripts are not executed yet.

### DIFF
--- a/app/lib/database/schema.task.dart
+++ b/app/lib/database/schema.task.dart
@@ -2,11 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-<<<<<<< migrations
-// TODO: remove this after typed_sql supports automatic snake_case convention
-=======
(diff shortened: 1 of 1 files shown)


app/tool/create-migration.sh
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env bash
```
sigurdm: Shouldn't this be dart?

isoos: It could, though in this case, I think the code wouldn't be that much more readabke, would it?

.gitignore
```diff
@@ -9,7 +9,6 @@ build/
 *.js_
 *.js.deps
 # *.js.map
-*.sum
```
sigurdm, line 12: Not needed anymore

app/tool/create-migration.sh
isoos: PTAL. Script is now in dart.

PR #9278, opened 2026-03-12
Attempt to keep image proxy key up to date

sigurdm:


app/lib/service/image_proxy/backend.dart
```diff
@@ -33,7 +33,16 @@ class ImageProxyBackend {
   static Future<ImageProxyBackend> create() async {
     final instance = ImageProxyBackend._();
 
-    unawaited(instance._dailySecret.update());
+    scheduleMicrotask(() async {
+      while (true) {
```
isoos: I think we may need to stop this loop, e.g. in tests and local testing that needs to exit the scope. E.g. task has this loop, we could reuse/copy some parts of it without the global locking:

https://github.com/dart-lang/pub-dev/blob/master/app/lib/task/backend.dart#L153-L199

app/lib/service/image_proxy/backend.dart
```diff
@@ -30,13 +30,28 @@ ImageProxyBackend get imageProxyBackend =>
...
 
-    unawaited(instance._dailySecret.update());
+    scheduleMicrotask(() async {
+      while (!instance._stopped) {
+        try {
+          await instance._dailySecret.update();
+        } catch (e, st) {
+          logger.severe('Failed to update daily secret', e, st);
+        }
+        await Future.delayed(Duration(minutes: 15));
```
isoos: Nit: maybe this need to be lower while not in appengine mode (`envConfig.isRunningLocally ? seconds: 10 : minutes:15`)

isoos: Hm, maybe the Completer + awaiting its future with a timeout is a more robust way to do it.

PR #9281, opened 2026-03-12
Add swiftpm-ready 

### LINKED ISSUE #8407: Add "Swift Package Manager Support" Tag for Flutter Packages
## Description

With Flutter's new support for Swift Package Manager (SPM), it would be helpful to identify packages on pub.dev that already support SPM. Adding a specific tag like "Swift Package Manager Support" to these packages would make it easier for developers to find and utilize them.

## Proposal

1. Add functionality to pub.dev to automatically identify if a package supports Swift Package Manager.
2. Once identified, pub.dev should automatically add a "Swift Package Manager Support" tag to the package's page.

This would be a small but valuable addition to improve discoverability and adoption of SPM-compatible packages within the Flutter ecosystem.

johnpryan:
This adds a swiftpm-ready tag that can be selected when searching for Pub packages.

Resolves #8407

### DIFF
--- a/pkg/_pub_shared/lib/search/search_form.dart
+++ b/pkg/_pub_shared/lib/search/search_form.dart
@@ -443,6 +443,7 @@ class SearchForm {
       parsedQuery.tagsPredicate.hasTag(PackageTags.showUnlisted) ||
       parsedQuery.tagsPredicate.hasTag(PackageVersionTags.hasScreenshot) ||
       parsedQuery.tagsPredicate.hasTag(PackageVersionTags.isPlugin) ||
+      parsedQuery.tagsPredicate.hasTag(PackageVersionTags.isSwiftpmReady) ||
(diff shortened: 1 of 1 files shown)


app/lib/package/model_properties.dart
```diff
@@ -218,6 +218,23 @@ class Pubspec {
...
+  late final isSwiftpmReady = () {
+    if (_flutterPluginMap == null) {
+      return false;
+    }
+    final platforms = _flutterPluginMap['platforms'];
+    if (platforms is! Map) {
+      return false;
+    }
+    for (final platform in platforms.values) {
+      if (platform is Map && platform['swift_package'] == true) {
```
johnpryan, line 232: We need to verify that this is the correct way to check.

johnpryan:
Since we already have a `is:swiftpm-plugin` tag, I think there's a simpler approach we can take to just add a checkbox that includes this in the search bar.

PR #9290, opened 2026-03-16
Add create-session-cookie admin action

sigurdm:
new admin action 'create-session-cookie' that takes an OAuth access token and returns browser session cookies. 

Useful for integration testing where we need the cookies to drive browser testing.

### DIFF
--- a/app/lib/admin/actions/create_session_cookie.dart
+++ b/app/lib/admin/actions/create_session_cookie.dart
@@ -1,4 +1,4 @@
-// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 


app/lib/admin/actions/create_session_cookie.dart
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
```
isoos:
```suggestion
// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
```

app/lib/admin/actions/create_session_cookie.dart
```diff
@@ -0,0 +1,42 @@
...
+import 'package:pub_dev/account/backend.dart';
+import 'package:pub_dev/account/session_cookie.dart' as session_cookie;
+import 'package:pub_dev/admin/actions/actions.dart';
+
+final createSessionCookie = AdminAction(
+  name: 'create-session-cookie',
+  summary: 'Create a session cookie for a given oauth access token.',
+  description: '''
+Validates the given oauth access token and creates a session cookie for the user.
+Returns the cookie headers that can be used to authenticate the user in the browser.
```
jonasfj, line 15: Maybe write a bit more... like say that this is only used for test users and only for integration testing.

jonasfj: say we created this for integration testing.

sigurdm: Now only allowing creating session cookies for test accounts.

PR #9291, opened 2026-03-16
Minimal tracker table + SQL migration script

isoos:
Note: this is an attempt to implement it with `typed_sql` so that we may move this to the package itself.

Limitations and design questions encountered so far:

- `typed_sql` has no specific control to put a table into a different schema (or to override a table's name). Without that FQN prefix, we can't use it in the same transaction as the DDL, unless we put it in the same schema. Probably a limit we can accept, but let's consider alternatives.

- There may be a sequence of events, where race-conditions may arise in competing migrations executions. To protect for all different kind of issues, we should probably lock the entire `schema_migrations` table for the transaction, but that may be DB-dependent (e.g. `LOCK` in postgres, I don't see it supported by sqlite, but maybe we can make it conditional).

### DIFF
--- a/app/lib/database/database.dart
+++ b/app/lib/database/database.dart
@@ -121,6 +121,7 @@ class PrimaryDatabase {
     );
 
     // create migration_schema table (if not exists)
+    // TODO(https://github.com/google/dart-neats/issues/348): use the output as-is after typed_sql supports it
     final createSql = createSchemaMigrationSchemaTables(
       _dialect,
     ).replaceFirst('CREATE TABLE "', 'CREATE TABLE IF NOT EXISTS "');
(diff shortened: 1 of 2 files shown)


app/lib/database/database.dart
```diff
@@ -113,6 +114,29 @@ class PrimaryDatabase {
     }
   }
 
+  Future<void> migrateSchema() async {
+    final migrationDb = Database<SchemaMigrationSchema>(
+      _adapter,
+      SqlDialect.postgres(),
+    );
+
+    // create migration_schema table (if not exists)
```
sigurdm, line 123: Perhaps add a todo with an associated issue to remove the need for this hack

PR #9312, opened 2026-03-24
Remove Image proxy experiment

sigurdm:


isoos (commented):
Hm, how about to not remove the code yet, instead, only update the experimental check to always true?

(We can keep this PR open and do another one for that to keep the cleanup parts readily available later.)

jonasfj (approved):
Agreed, generally, it's best not to remove the dead code yet.

isoos:
Closing in favor of #9517

PR #9338, opened 2026-04-10
Build the app as jit-snapshot in a from-scratch docker image to reduce dependencies.

sigurdm:
* Create a helper in go for verifying signatures, to remove dependency on an openssl binary.
* Remove unused dependencies from the docker build script: git, wepb,  unzip. They are all only used in the worker.
* Use kernel snapshots for the search isolates


app/bin/server.dart
```diff
@@ -7,11 +7,20 @@ import 'package:args/command_runner.dart';
...
 import 'package:pub_dev/service/entrypoint/search.dart';
+import 'package:pub_dev/shared/env_config.dart';
 
 void main(List<String> args) async {
   final runner = CommandRunner('pub_dev', 'pub.dev services')
     ..addCommand(AnalyzerCommand())
     ..addCommand(DefaultCommand())
     ..addCommand(SearchCommand());
+
+  if (args.isEmpty) {
```
isoos:
```suggestion
  if (args.isEmpty && envConfig.isRunningInAppengine) {
```

app/lib/service/entrypoint/search.dart
```diff
@@ -79,9 +79,11 @@ Future<void> runSearchInstanceController({
   final sdkIsolate = await startQueryIsolate(
     logger: _logger,
     kind: 'sdk',
-    spawnUri: Uri.parse(
-      'package:pub_dev/service/entrypoint/sdk_isolate_index.dart',
-    ),
+    spawnUri: envConfig.isRunningInAppengine
```
isoos: Nit: maybe use a separate `bool get hasPrecompiledBinaries => isRunningInAppengine` flag?

isoos:
> Create a helper in go for verifying signatures, to remove dependency on an openssl binary.

It is not clear if this will be compiled and used locally all the time (in which case let's update the `doc/development.md` file - which is outdated, but at least add the relevant new tool requirements section).

> Use kernel snapshots for the search isolates

It may be worth to have this separately in a different PR.

sigurdm:
> It is not clear if this will be compiled and used locally all the time (in which case let's update the doc/development.md file - which is outdated, but at least add the relevant new tool requirements section).

It is used all the time. Added a section that you need the go sdk.

jonasfj (approved):
I have no problems, if istvan is happy :D

doc/development.md
```diff
@@ -12,6 +12,8 @@ instructions how to work with the local, in-memory server (`fake_server`).
 - Development and runtime support is best on Linux, compatible system may work,
   Windows isn't supported yet (e.g. bash scripts are not compatible).
 
+- The **Go SDK** is equired to build the `signature_verifier` helper used for RSA signature verification (replaces the dependency on `openssl`).
```
isoos:
```suggestion
- The **Go SDK** is required to build the `signature_verifier` helper used for RSA signature verification (replaces the dependency on `openssl`).
```

isoos (approved):

PR #9339, opened 2026-04-13
Use a Zone variable to indicate that a database retry block has been already started.

isoos:
- Follow-up to #9336.
- I was updating the task migration (#9126), and while the callbacks introduced some friction, I think the complexity was manageable, with the exception of the loop calls, where it was not clear where we were (e.g. 3x repeat vs 3x3 repeat). Adding the zone tracking should mitigate that.
- This also handles the case where `withRetry()` is called, and then internally these is a `schema.transact()` call which could throw an exception that needs to be unboxed.

### DIFF
--- a/app/lib/database/database.dart
+++ b/app/lib/database/database.dart
@@ -237,7 +237,13 @@ final _retryKey = #dbRetryKey;
 extension PrimaryDatabaseExt on PrimaryDatabase {
   /// Runs [fn] in a retry block (without wrapping it in a transaction).
   ///
-  /// The call is retried only if [DatabaseConnectionException] is throw.
+  /// The retry block is tracked with a [Zone], and if multiple retries are embeded
+  /// (with or without transaction), only the outermost block will be retried.
+  ///
+  /// The call is retried if the generic [DatabaseException] is throw, which may be a
+  /// connection issue, deadlock, timeout, constraint or any query-related problem.
+  ///
+  /// However, if inside the transaction an [Error] is thrown, or if the wrapped exception
   Future<K> withRetry<K>(
     Future<K> Function(Database<PrimarySchema> db) fn,
   ) async {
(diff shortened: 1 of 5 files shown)


app/lib/database/database.dart
```diff
@@ -232,18 +232,16 @@ Future<void> _dropCustomDatabase(String url, String dbName) async {
   await conn.close(force: true);
 }
 
+final _retryKey = #dbRetryKey;
+
 extension PrimaryDatabaseExt on PrimaryDatabase {
   /// Runs [fn] in a retry block (without wrapping it in a transaction).
   ///
   /// The call is retried only if [DatabaseConnectionException] is throw.
```
sigurdm: This probably needs an update.

sigurdm (commented):
Could you add tests for this mechanism?

isoos:
Updated. PTAL!

app/lib/database/database.dart
```diff
@@ -232,22 +232,29 @@ Future<void> _dropCustomDatabase(String url, String dbName) async {
...
   /// Runs [fn] in a retry block (without wrapping it in a transaction).
   ///
-  /// The call is retried only if [DatabaseConnectionException] is throw.
+  /// The retry block is tracked with a [Zone], and if multiple retries are embeded
+  /// (with or without transaction), only the outermost block will be retried.
+  ///
+  /// The call is retried if the generic [DatabaseException] is throw, which may be a
+  /// connection issue, deadlock, timeout, constraint or any query-related problem.
+  ///
+  /// However, if inside the transaction an [Error] is thrown, or if the wrapped exception
```
sigurdm, line 246: Broken sentence

app/test/database/retry_zone_test.dart
```diff
@@ -0,0 +1,189 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
```
sigurdm:
```suggestion
// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
```

sigurdm (approved):

PR #9340, opened 2026-04-13
Start reading .tar.gz snapshot file (with .json.gz fallback)

isoos:
- Follow-up to #9327
- Starts reading the tar.gz file, logging if it is missing. (If such logs do not show up, it is safe to remove the .json.gz fallback).

### DIFF
--- a/app/lib/search/backend.dart
+++ b/app/lib/search/backend.dart
@@ -457,7 +457,7 @@ class SearchBackend {
       } on DetailedApiRequestError catch (e) {
         if (e.status == 404) {
           // ignore and log
-          _logger.warning('Snapshot archive file not found.');
+          _logger.warning('Snapshot tar.gz archive file not found.');
         } else {
           rethrow;
         }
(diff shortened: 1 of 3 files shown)


app/lib/search/backend.dart
```diff
@@ -450,7 +451,19 @@ class SearchBackend {
...
   Future<List<PackageDocument>?> fetchSnapshotDocuments() async {
     try {
-      final map = await _snapshotStorage.getContentAsJsonMap();
+      Map<String, dynamic>? map;
+      try {
+        map = await _snapshotStorage.getContentAsJsonMapFromTarGz();
+      } on DetailedApiRequestError catch (e) {
+        if (e.status == 404) {
+          // ignore and log
+          _logger.warning('Snapshot archive file not found.');
```
sigurdm:
```suggestion
          _logger.warning('Snapshot tar.gz archive file not found.');
```

sigurdm (approved):

PR #9341, opened 2026-04-15
Replace package:mdc_web with a small JS binding.

isoos:
Also migrated some Element creation to the `package:web`-compatible version.


pkg/web_app/pubspec.yaml
```diff
@@ -11,7 +11,6 @@ dependencies:
   collection: ^1.18.0
   http: ^1.0.0
   markdown: ^7.2.0
-  mdc_web: ^0.6.0
```
sigurdm, line 14: Yay!

sigurdm (approved):

PR #9345, opened 2026-04-17
New database init block.

isoos:
- the `try-catch` is moved to the database method (since it already starts with `try...`
- `DatabaseException` from typed_sql is also handled
- schema migration is run in a retry block
- production now tries to connect to the database, but does not migrate the schema yet

### DIFF
--- a/app/lib/database/database.dart
+++ b/app/lib/database/database.dart
@@ -86,7 +86,10 @@ class PrimaryDatabase {
   /// When NOT running in the AppEngine environment (e.g. testing or local fake),
   /// the initilization will create a new database, which will be dropped when the
   /// [close] method is called.
-  static Future<PrimaryDatabase> createAndInit({String? url}) async {
+  static Future<PrimaryDatabase> createAndInit({
+    String? url,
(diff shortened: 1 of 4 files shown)


app/lib/database/database.dart
```diff
@@ -47,21 +49,33 @@ class PrimaryDatabase {
...
+      if (envConfig.isRunningInAppengine) {
+        // ignore setup issues for now
+        _logger.warning('Could not connect to Postgresql database.', e, st);
+      } else {
+        rethrow;
+      }
+    } on DatabaseException catch (e, st) {
+      if (envConfig.isRunningInAppengine) {
+        // ignore setup issues for now
+        _logger.warning('Could not initialize typed_sql.', e, st);
```
sigurdm, line 74: Is this not going to be a fatal error?

sigurdm (approved):

app/lib/database/database.dart
isoos: I thought we shall start with warning, since we are not yet requiring it (while testing if prod works).

PR #9361, opened 2026-04-28
Hash-based, two-level index for IndexedBlob.

isoos:


pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -111,160 +140,236 @@ final class IndexedBlobBuilder {
...
   /// returned by [BlobIndex.lookup]. It is intended to store an identifier for
-  /// blob containing the file, or a URL to the blob.
-  Future<BlobIndex> buildIndex(String blobId) async {
+  /// the blob containing the file, or a URL to the blob.
+  ///
+  /// ## Blob layout
+  ///
+  /// The written blob contains, in order:
+  /// 1. **File entries** – for each file added via [addFile]:
+  ///    `utf8(path) + 0x00 + <raw content bytes>`
```
sigurdm: Should we add a file size prefix?
```suggestion
  ///    `utf8(path) + 0x00 + <length> + <raw content bytes>`
```

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -111,160 +140,236 @@ final class IndexedBlobBuilder {
...
+    }
 
-    // Note. it's a bit of a hack, but keys are expected to be order as follows:
-    // - version
-    // - blobId
-    // - index
+    if (_entries.isEmpty) {
+      await _blob.close();
+      return BlobIndex.empty(blobId: blobId);
+    }
```
sigurdm, line 218: Can we leave this out? Don't we get an equivalent behavior below?

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -111,160 +140,236 @@ final class IndexedBlobBuilder {
...
+    _entries.sort((a, b) => _compareHash(a.pathHash, b.pathHash));
+
+    final maxCommonHashPrefix = _entries.indexed
+        .skip(1)
+        .map((a) => _commonPrefix(_entries[a.$1 - 1].pathHash, a.$2.pathHash))
+        .fold(0, (a, b) => max(a, b));
+    final hashPrefixBytes = max(4, maxCommonHashPrefix + 1);
+
+    final worstCaseIndexSize = (hashPrefixBytes + 8 + 2 + 8) * _entries.length;
```
sigurdm: Are the entries not fixed size? Why is this worst case?

pkg/indexed_blob/lib/indexed_blob.dart
isoos: It could be used to read through the blob sequentially, but I don't think we have any use case for that (maybe for debugging).

pkg/indexed_blob/lib/indexed_blob.dart
isoos: E.g. at the end of this method we are using `slices.first`, and if we don't check the length here, we need to check there (and that code would need a bit restructure for that). I don't think it is worth to remove this early return.

pkg/indexed_blob/lib/indexed_blob.dart
isoos: At this point it is not decided how many subindexes we will have, and those subindexes are not written to the blob yet. It is possible that the offset bytes will take 4 bytes or 8, and it is easier to assume the worst (8). With that, we can calculate the maximum offset, and from there we can calculate the bytes needed for it.

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -282,171 +387,319 @@ final class BlobIndex {
...
+///
+/// Parameters:
+/// - [length]: the number of elements to search over.
+/// - [compare]: called with an index; should return:
+///     - negative: element at index is **less than** the target
+///     - zero:     element at index **equals** the target
+///     - positive: element at index is **greater than** the target
+/// - [exactMatch]: when `true`, only a zero-comparison result is accepted and
+///   `-1` is returned if no exact match exists.  When `false`, the method
+///   returns the largest index whose element is **≤ the target** (i.e. the
```
sigurdm, line 581: Not sure it really matters but you return mid immediately if cmp == 0. In a sequence with duplicate values, this will return any index that matches the target, not necessarily the largest (rightmost) index.

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -72,11 +75,17 @@ final class IndexedBlobBuilder {
     }
   }
 
+  /// Estimated maximum length of the index file.
+  int get indexUpperLength => 128 + _entries.length * (32 + 8 + 2 + 8);
```
sigurdm: Are we using this?

pkg/indexed_blob/lib/indexed_blob.dart
sigurdm: Not sure it matters in practice, but if we allow arbitrary strings for path, I belive they could contain \0 bytes themselves. Maybe a length prefix for the path is better than a 0 separator?

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -111,160 +140,236 @@ final class IndexedBlobBuilder {
...
+  ///
+  /// See [HashIndexHeader] for the exact byte layout of each record type.
+  ///
+  /// ## Subindex logic
+  ///
+  /// When the total index size would exceed [indexSizeThresholdKiB] KiB,
+  /// entries are divided into equally-sized slices:
+  ///
+  /// - The **first slice** is stored in the returned index buffer (the main
+  ///   index), covering the lowest-hash portion of the key space.
```
sigurdm, line 192: In my mind it would be a lot simpler if we **either** store a single index **or** sub-indexes.

It confuses me that we handle an arbitrary portion of the paths differently from the rest in the subindex case.

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -111,160 +140,236 @@ final class IndexedBlobBuilder {
...
+    final hashPrefixBytes = max(4, maxCommonHashPrefix + 1);
+
+    final worstCaseIndexSize = (hashPrefixBytes + 8 + 2 + 8) * _entries.length;
+    final worstCaseOffset =
+        _offset +
+        worstCaseIndexSize +
+        /* arbitrary padding for headers and other overheads */
+        128 * 1024;
+
+    // always use path length of 2 bytes
```
sigurdm:
```suggestion
    // We always use 2 bytes for the length-prefix for the path.
```

pkg/indexed_blob/lib/indexed_blob.dart
isoos: It may be cleaner, but in practice, I think the transition to no subindex to subindexes is like this:

Option A, current impl:
- main index with about 256 KiB content, subindex with about 256 KiB content
- lookups have 50% chance that they are in the main index, no second lookup is needed

Option B, storing either entries or subindexes:
- very small main index, two ~256 KiB subindex
- lookups are always need to read the second level

I think that at the time when subindexes reach 10 or more, it is less important to store the entries in the main index, but for this case it may be worth it.

We may tune this without changing the format though.

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -282,171 +387,319 @@ final class BlobIndex {
...
+    var dataOffset =
+        _header.getEntryOffset(selectedEntry) + _header.hashPrefixBytes;
+    final entryOffset = data.getUint(_header.offsetBytes, dataOffset);
+    dataOffset += _header.offsetBytes;
+    final pathLength = data.getUint(_header.pathLengthBytes, dataOffset);
+    dataOffset += _header.pathLengthBytes;
+    final contentLength = data.getUint(_header.contentLengthBytes, dataOffset);
+
+    // sanity check on path length
+    if (pathLength != pathBytes.length) return null;
```
sigurdm: Should we also sanity-check the path itself?

pkg/indexed_blob/lib/indexed_blob.dart
isoos: We need to know the length in the index, no need to repeat it here. We could remove the trailing zero though.

pkg/indexed_blob/lib/indexed_blob.dart
sigurdm: And should we throw if there is a mis-match? I guess it indicates something has broken?

pkg/indexed_blob/lib/indexed_blob.dart
isoos: Obsolete, removed.

sigurdm (commented):
Looks pretty solid.

Should we change the filename of the index away from json in the same PR, such that we are sure not to match up indices from different versions?

sigurdm (approved):

pkg/indexed_blob/lib/indexed_blob.dart
isoos: We do that on reading the bytes. This is another sanity check, a bit cheaper (without lookup from the blob).

pkg/indexed_blob/lib/indexed_blob.dart
sigurdm: But in that case we actually don't need the path at all, right? If we never want to enumerate the blob file on its own...

pkg/indexed_blob/lib/indexed_blob.dart
isoos: Oh, reverting, since we do use it in the pub_worker, limiting the index building (as another sanity check).

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -90,9 +99,29 @@ final class IndexedBlobBuilder {
...
-      final start = _offset;
+
+      final startOffset = _offset;
+
+      // sanity check for the encoded paths
+      final pathBytes = utf8.encode(path);
+      if (pathBytes.length > 4096) {
+        throw ArgumentError(
+          'Path too long (must not exceed 4096 UTF-8 bytes).',
+        );
```
jonasfj, line 116: nit: Check this before `if (!_paths.add(path)) {` so the error doesn't break state

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -455,34 +708,266 @@ final class FileRange {
...
+/// ## Entry record layout (`entryCount` records after the header)
+///
+/// Each entry describes one file stored in the blob:
+///
+/// ```
+/// Size              Field
+/// ----              -----
+/// hashPrefixBytes   leading bytes of SHA-256(blobIdBytes ++ pathBytes)
+/// offsetBytes       start offset of the entry in the blob (uint)
+/// pathLengthBytes   byte-length of the file path (uint)
```
jonasfj: Why do we need or even want this in the index?

Why not just have the blob be:
```
                        // <-- offset points to here.
[uint32: path-length]
[variable: path-length]
[variable: content]     // <-- offset + length points to here
```

We should never need to load paths separately, and if we do really want to do that, you can do a read from `offset` to `min(offset + length, offset + 4096 + 4)` because you know paths cannot be more than 4kb -- as we disallow this.

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -455,34 +708,266 @@ final class FileRange {
...
+/// ```
+/// Offset    Size      Field
+/// ------    ----      -----
+///      0       1      Magic byte 'I' (0x49)
+///      1       1      Magic byte 'B' (0x42)
+///      2       2      version (uint16); currently always 1
+///      4       1      hashPrefixBytes: number of hash bytes stored per entry (≥ 4)
+///      5       1      offsetBytes: byte-width used for blob offsets (2, 4, or 8)
+///      6       1      pathLengthBytes: byte-width used for path lengths (always 2)
+///      7       1      contentLengthBytes: byte-width used for content lengths
```
jonasfj: do these really need to be variable, life is MUCH simpler if they are always 4

And 4 GB blobs is an acceptable limit.

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -111,160 +140,236 @@ final class IndexedBlobBuilder {
...
+}
 
-    w.endObject();
+Uint8List _hash(List<int> saltBytes, List<int> pathBytes) {
+  Digest? digest;
+  final input = sha256.startChunkedConversion(
+    ChunkedConversionSink.withCallback((list) => digest = list.single),
+  );
+  input.add(saltBytes);
+  input.add(pathBytes);
```
jonasfj: Isn't an hmac safer?

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -282,171 +387,319 @@ final class BlobIndex {
...
+  ///
+  /// Entries are yielded in hash order (not lexicographic path order).
+  ///
+  /// **Caution:** issues one [readBlob] call per entry plus one per subindex
+  /// block. For remote blobs, callers may want to buffer the stream rather
+  /// than issuing large numbers of individual range requests.
+  ///
+  /// Throws [FormatException] if the path bytes read from the blob do not
+  /// match the hash stored in the index for that entry.
+  Stream<FileRange> listFiles(BlobSliceReader readBlob) async* {
```
jonasfj: Do we need this? I'd be okay not having it?

or if we do, then have it by streaming down the entire blob. At which point, we should just give it a `Stream<List<int>>`.

I imagine we only ever need this for testing or debugging.

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -282,171 +387,319 @@ final class BlobIndex {
...
+  /// Returns `true` if [path] is stored in the indexed blob **and** the path
+  /// bytes read from the blob match the expected value.
+  ///
+  /// Unlike [lookup], which only checks the hash index, this method also reads
+  /// the path bytes from the blob via [readBlob] to confirm there is no hash
+  /// collision or index/blob mismatch.
+  ///
+  /// [readBlob] may be called up to twice: once to resolve a subindex block
+  /// (when the main index has subindexes), and once to read the path bytes.
+  Future<bool> hasFile(String path, BlobSliceReader readBlob) async {
```
jonasfj: Do we need this?

pkg/indexed_blob/lib/indexed_blob.dart
```diff
@@ -282,171 +387,319 @@ final class BlobIndex {
...
+    final range = await lookup(path, readBlob);
+    if (range == null) return false;
+    final blobPathBytes = await readBlob(
+      range.entryOffset,
+      range.entryOffset + range.pathLength + 1,
+    );
```
jonasfj: You don't need `range.pathLength`, you can just do:
```dart
    final blobPathBytes = await readBlob(
      range.entryOffset,
      math.min(range.entryOffset + range.length, range.entryOffset + 4096 + 1),
    );
```

Reading less than 4kb doesn't make sense anyways.

app/lib/task/backend.dart
```diff
@@ -857,7 +867,7 @@ class TaskBackend {
 
     FileRange range;
     try {
-      final r = index.lookup(path);
+      final r = await index.lookup(path, _blobSliceReader(index.blobId));
```
jonasfj: Why don't we just have `index.readFile(path, _blobSliceReader(index.blobId))` and make that:
 * return `Uint8List`,
 * return `null` if not there,
 * throw `FormatException` if index is invalid.

I would also love it, if the `blobId` was relative to the index file location, I don't even think that's too hard to do.

app/lib/task/backend.dart
```diff
@@ -661,7 +662,10 @@ class TaskBackend {
       version,
       await _gzippedTaskResult(index, 'summary.json'),
     );
-    final hasDocIndexHtml = index.lookup('doc/index.html') != null;
+    final hasDocIndexHtml = await index.hasFile(
```
jonasfj: If we don't want to build `hasFile` because this is possibly the only use case, why don't we just read `doc/index.html`. Not as fast, but ideally we could also put this into `summary.json` or something (one day).

jonasfj: Granted if we can avoid carrying around the path length and do the min(length, 4096+1)` hack instead, the `hasFile` is cheaper than loading the entire file.

It just should just only be used in places where reading the entire file is the only alternative.

pkg/indexed_blob/lib/indexed_blob.dart
isoos: The path length has been removed from the index.

pkg/indexed_blob/lib/indexed_blob.dart
isoos: Path length is now removed from the index.

app/lib/task/backend.dart
```diff
@@ -620,7 +621,7 @@ class TaskBackend {
       ),
       uploadSigner.buildUpload(
         _bucket.bucketName,
-        '$runtimeVersion/$package/$version/index.json',
+        '$runtimeVersion/$package/$version/index.bin',
```
```suggestion
        '$runtimeVersion/$package/$version/blob.index',
```

is perhaps cuter :D

isoos:
Note: full removal of `FileRange` (and also the `BlobIndex.lookup()`) is not beneficial, because of its use in `_gzippedTaskResult`, but otherwise the reduction of its API surface was possible. If we tradeoff some efficiency in that method (e.g. calling the cache even though we wouldn't cache the data because it is large), we could remove these from the public API.

PTAL.

jonasfj (approved):

sigurdm (approved):

Issue #9248, opened 2026-03-02
Get rid of old public archive bucket


sigurdm:
dartlang-pub-public-packages
dartlang-pub-dev-public-packages

We are now serving the package archives out of the exported api bucket. But we still seem to maintain the old public bucket.
We need to:

* [x] Update the deployment script to rely on the exported api bucket instead.
* [ ] Make the old bucket private for a week or two, to make sure noone is relying on it
* [ ] Remove the updates and any other references to the old bucket
* [ ] Remove the old bucket

AlexV525:
Out of curious: we've seem bucket changes by user reporting they cannot download from us (the CFUG host). Could someone share more details and progress that can help us to keep up-to-date seemlessly? FYI @chenglu

sigurdm:
Not sure i understand the issue here.
Who has this problem? Where are they downloading from? and Why.

If you are downloading directly from dartlang-pub-public-packages please don't do that.
Please always use the archiveUrl that you get from the [package-listing API](https://github.com/dart-lang/pub/blob/master/doc/repository-spec-v2.md#list-all-versions-of-a-package), that is the only stable api we provide for downloading packages.

chenglu:
In the China mirror, the Pub API is a critical component of the CLI toolchain. We have optimized our network infrastructure specifically to ensure the fastest possible response times.

Previously, we utilized direct redirects to the GCS bucket for binary downloads to maintain speed. The recent change now forces all downloads through the API server, which created massive instantaneous pressure and led to service unavailability for our users. Fortunately, Alex reported this promptly, and we’ve implemented a fix downstream.

While I understand the need for these changes, I strongly suggest incorporating these into our coordination process to avoid breaking the ecosystem, such as including a connectivity and download test for the China mirror in the CI suite to ensure it remains functional.

Thanks.

sigurdm:
I understand now, for the Chinese mirror, please use the `dartlang-pub-exported-api` bucket, folder `latest/api/packages`

chenglu:
Got it. We will update the redirect logic for the China mirror to point to the new bucket later.

This should effectively resolve the performance issues we encountered. Thanks for the continuous support and for pointing us to the right path!

sigurdm:
No problem. Thanks for supporting the Flutter community in China! And I'm sorry for the interruptions this might have caused

chenglu:
No worries at all, and thank you for the kind words, @sigurdm!

[...]

PR #9374, opened 2026-05-05
Stop populating public package bucket (replaced by exported-api)

sigurdm:
Towards https://github.com/dart-lang/pub-dev/issues/9248

Remaining TarBallStorage functionality was folded into PackageBackend.


app/test/admin/moderate_package_version_test.dart
```diff
@@ -580,16 +533,10 @@ void main() {
 
     testWithProfile(
       'cleanup deletes datastore entities and canonical archive file',
-      expectedLogMessages: [
-        'SHOUT Deleting object from public bucket: "packages/oxygen-1.0.0.tar.gz".',
```
isoos, line 584: Wondering: why is this login message absent in the new version? I think we do want to log object deletions the same way.

sigurdm: Fair - deletion only happens in the exported-api backend now. Made that shout

sigurdm:
@isoos PTAL

isoos (approved):

PR #9390, opened 2026-05-15
Test updates: await on async queue, check for no dart:html use.

isoos:
I think the await is beneficial to fix a very small fraction of flaky tests.

### DIFF
--- a/pkg/web_app/test/no_dart_html_test.dart
+++ b/pkg/web_app/test/no_dart_html_test.dart
@@ -1,0 +1,19 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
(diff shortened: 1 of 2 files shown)


pkg/web_app/test/no_dart_html_test.dart
```diff
@@ -0,0 +1,19 @@
...
+
+import 'package:test/test.dart';
+
+void main() {
+  test('no dart:html use in pkg/web_app', () async {
+    final dir = Directory('lib');
+    await for (final f in dir.list(recursive: true)) {
+      if (f is File && f.path.endsWith('.dart')) {
+        final content = await f.readAsString();
+        expect(content, isNot(contains('dart:html')), reason: f.path);
```
kevmoo, line 15: We have a LINT for this!! 😄

isoos: Sure, this is dumb, but we also had `// ignore` for it :see_no_evil:

sigurdm (approved):

PR #9395, opened 2026-05-18
Report advisoriesUpdated: null if there are no advisories

### LINKED ISSUE #4818: pub get became very slow
Here's the `flutter pub get -v` output.

[pub_get.log](https://github.com/user-attachments/files/27511238/pub_get.log)

### Observed timings
- Total: `~408.79s`
- Version solving: `~2.01s`
- Root project phase before first `Got dependencies!`: `~380.01s`
- Metadata HTTP requests count: `364`
- Sum of HTTP request durations: `~370.37s`
- Median request latency: `~0.91s`
- p95: `~1.40s`
- p99: `~3.16s`

### Expected behavior
`dart pub get` should complete significantly faster for already-cached packages / lockfile-stable graphs, and should not require so much wall-clock time in metadata lookups.

### Environment
- Dart (from Flutter SDK): `Pub 3.11.5` (from verbose log)
- OS: macOS (arm64)
- Command path: `flutter pub get --example` -> `dart pub get -v`

sigurdm:
This should fix: https://github.com/dart-lang/pub/issues/4818

We should also have the client fetch advisories in parallel.

### DIFF
--- a/app/lib/package/models.dart
+++ b/app/lib/package/models.dart
@@ -169,6 +169,10 @@ class Package extends db.ExpandoModel<String> {
   @db.DateTimeProperty()
   DateTime? latestAdvisory;
 
+  /// Whether the package is affected by any active security advisories.
+  @db.BoolProperty(required: true)
+  bool hasAdvisories = false;
+
   Package();
 
   /// Creates a new [Package] and populates all of its fields from [version].
--- a/app/lib/service/security_advisories/backend.dart
+++ b/app/lib/service/security_advisories/backend.dart
@@ -69,10 +69,6 @@ class SecurityAdvisoryBackend {
       skipCache: skipCache,
     );
 
-    if (advisories.isEmpty) {
-      return ListAdvisoriesResponse(advisories: [], advisoriesUpdated: null);
-    }
-
     final p = await packageBackend.lookupPackage(package);
     final latestAdvisory =
         p?.latestAdvisory ?? DateTime.fromMillisecondsSinceEpoch(0);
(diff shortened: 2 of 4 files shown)


app/lib/service/security_advisories/backend.dart
```diff
@@ -69,6 +69,10 @@ class SecurityAdvisoryBackend {
       skipCache: skipCache,
     );
 
+    if (advisories.isEmpty) {
+      return ListAdvisoriesResponse(advisories: [], advisoriesUpdated: null);
```
jonasfj: We can't do this, it breaks the promises we've made in the API.

Why can we not use `DateTime.fromMillisecondsSinceEpoch(0);´?

jonasfj (changes requested):

app/lib/service/security_advisories/backend.dart
sigurdm: Sorry, I was fixing the wrong API.

Updated

app/lib/package/models.dart
```diff
@@ -169,6 +169,10 @@ class Package extends db.ExpandoModel<String> {
   @db.DateTimeProperty()
   DateTime? latestAdvisory;
 
+  /// Whether the package is affected by any active security advisories.
+  @db.BoolProperty(required: true)
+  bool hasAdvisories = false;
```
isoos, line 174: Let's start with required: false, I think the required annotation will throw regardless if the field has an initializer. Not sure if we can start with this `bool hasAdvisories` vs `bool?` - worth considering.

sigurdm:
Sorry, we probably need a revert here instead

PR #9398, opened 2026-05-18
Track and join pending database transaction when using transactWithRetry

isoos:
Note: `typed_sql` (or specifically the postgresql driver) doesn't support re-entrant calling of `db.transact()`. With this change we don't force-start a new transaction, just use the outer one (same logic as with joining the outermost retry block).

### DIFF
--- a/app/lib/database/database.dart
+++ b/app/lib/database/database.dart
@@ -358,6 +358,7 @@ Future<void> _dropCustomDatabase(String url, String dbName) async {
 }
 
 final _retryKey = #dbRetryKey;
+final _txKey = #dbTxKey;
 
 extension PrimaryDatabaseExt on PrimaryDatabase {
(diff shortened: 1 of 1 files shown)


jonasfj:
> Note: typed_sql (or specifically the postgresql driver) doesn't support re-entrant calling of db.transact(). With this change we don't force-start a new transaction, just use the outer one (same logic as with joining the outermost retry block).

I think it's only concurrent nested subtransactions that disallowed.

jonasfj:
See also https://github.com/google/dart-neats/pull/372

That said, one can argue that we probably don't need nested transactions, and instead just want single transactions.

Where in our code is this even relevant?

isoos:
Note: this may not be needed after all. Though, we could use this to detect re-entrant transaction calls, which we probably don't want, and maybe block them?

isoos:
Dropping this after offline discussion, we will do this in typed_sql.

PR #9399, opened 2026-05-18
Reapply "Load/save Postgres database state into local state file. (#9389)

isoos:
Fixed the broken code that was caused by relying on the task sql migration PR.


sigurdm (approved):

isoos:
Note: the test fails because it expects that the task backed will store the analysis results in the SQL database. Since we have rolled back that part, merging this PR is blocked until it is re-added again.

PR #9402, opened 2026-05-19
Simpler CachedValue logic + more tests.

isoos:
Cached value is kept indefinitely, retry logic is based on the interval, and does not trigger excessive updates.

### DIFF
--- a/app/lib/shared/cached_value.dart
+++ b/app/lib/shared/cached_value.dart
@@ -29,18 +29,17 @@ typedef UpdateFn<T> = Future<T?> Function();
 ///
 /// - The default `timeout` value is half of the `interval`.
 ///
-/// - `maxAge` is used to affect `isAvailable` and also to raise the
-///   level of logging if the updates happen to fail longer than the
-///   specified age.
+/// Once a value is successfully fetched, it is held indefinitely (even if
+/// subsequent updates fail). `isAvailable` returns `true` as long as a
+/// non-null value has been set.
 ///
 /// `CacheValue` will eventually retry the `update` function when the next scheduled update happens.
 class CachedValue<T> {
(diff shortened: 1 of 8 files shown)


app/lib/shared/cached_value.dart
```diff
@@ -107,12 +121,16 @@ class CachedValue<T> {
...
+      _lastAttempted = clock.now();
       _value = await _updateFn().timeout(_timeout);
       if (_value != null) {
         _lastUpdated = clock.now();
       }
     } catch (e, st) {
-      if (age <= _maxAge) {
+      final failureDuration = _lastUpdated != null
+          ? clock.now().difference(_lastUpdated!)
+          : clock.now().difference(_createdAt);
```
sigurdm: Do we need both here? Could _lastUpdated not start out as the creation time?

sigurdm:
I'm not sure I fully follow the logic here. What improved behavior do we expect. Is it possible to test for it?

isoos:
> I'm not sure I fully follow the logic here. What improved behavior do we expect. Is it possible to test for it?

The two core issues with the old code:
- `_lastUpdated` was initialized to `clock.now()`, which made the freshly created cache appear as recently updated
- `_lastAttempted` was not tracked, so a failed attempt would be rescheduled right away, causing potentially a pressure loop we don't need

This affected both scheduling and how availability was treated. I'm adding a few more tests to highlight the correct behavior.

isoos:
PTAL. I've update the code to keep the value indefinitely + using simpler rules for schedule and logging.

sigurdm (approved):
Love it!

sigurdm:
Perhaps update the pr description

PR #9403, opened 2026-05-20
Split TokenIndex's weighted doc list + use quantized values for weights.

isoos:
- Local benchmark shows no change in processing latencies, but reduces memory requirements by about 6-7%.
- Useful step for migrating to disk- or sqlite-stored `TokenIndex`.

### DIFF
--- a/app/test/search/dartdoc_index_parsing_test.dart
+++ b/app/test/search/dartdoc_index_parsing_test.dart
@@ -89,9 +89,9 @@ void main() {
                     'https://api.flutter.dev/flutter/widgets/StatelessWidget-class.html',
               },
               {
-                'path': 'widgets/StatelessWidget/StatelessWidget.html',
+                'path': 'widgets/StatelessWidget/createElement.html',
                 'url':
-                    'https://api.flutter.dev/flutter/widgets/StatelessWidget/StatelessWidget.html',
+                    'https://api.flutter.dev/flutter/widgets/StatelessWidget/createElement.html',
               },
               {
                 'path': 'widgets/StatelessWidget/build.html',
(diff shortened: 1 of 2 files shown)


app/test/search/dartdoc_index_parsing_test.dart
```diff
@@ -89,9 +89,9 @@ void main() {
                     'https://api.flutter.dev/flutter/widgets/StatelessWidget-class.html',
               },
               {
-                'path': 'widgets/StatelessWidget/StatelessWidget.html',
+                'path': 'widgets/StatelessWidget/createElement.html',
```
sigurdm, line 92: Why did this change?

isoos: That's likely the side-effect of quantization: previously some numbers were rounded up, which now gets to rounded down, and the elements switch places in the final ranking.

isoos: Before:
- `0.9841605470367712 widgets/StatelessWidget-class.html`
- `0.9823977944917073 widgets/StatelessWidget/StatelessWidget.html`
- `0.9823977944917073 widgets/StatelessWidget/build.html`
- `0.9796291837976918 widgets/StatelessWidget/createElement.html`
- `0.5717456113001694 widgets/StatelessElement-class.html`

After:
- `0.984375 widgets/StatelessWidget-class.html`
- `0.98046875 widgets/StatelessWidget/createElement.html`
- `0.98046875 widgets/StatelessWidget/build.html`
- `0.98046875 widgets/StatelessWidget/StatelessWidget.html`
- `0.5718701972134389 widgets/StatelessElement/StatelessElement.html`

So the top 2-4 items are the same score (both before and after), but for some reason this changed the order.

isoos: The heap building could also include the long string key, but I think it would just slow down the calculation for very little consistency gain. Let's leave it as-is.

sigurdm (approved):

PR #9404, opened 2026-05-20
Upgrade the runtime SDK + CI.

isoos:


app/test/admin/exported_api_sync_test.dart
```diff
@@ -119,10 +119,13 @@ void main() {
     testWithProfile(
       'deleted files + full sync',
       fn: () async {
+        print('a');
```
sigurdm: Remember to remove

sigurdm (approved):

PR #9408, opened 2026-05-22
signature_verifier in Dart+web_crypto

sigurdm:
Removing dependency on golang.


isoos (approved):

jonasfj (approved):
We probably should consider moving this in process at some point.

pkg/signature_verifier/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
...
+description: A tool/library to verify RSA SHA-256 signatures.
+version: 0.1.0
+
+environment:
+  sdk: ^3.11.0
+
+resolution: workspace
+
+dependencies:
+  webcrypto: ^0.6.0
```
jonasfj:
```suggestion
  webcrypto: ^0.6.1
```

sigurdm:
Yeah - let's move it in process later if everything looks good. I just like to have the process boundary for the ffi calls a little while.

PR #9410, opened 2026-05-25
Migrating task model and backend to use typed_sql.

isoos:
- This reverts commit 3796fabc5a2e65f918a85228cecab6f32f74d3d2.

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -7,6 +7,7 @@ import 'package:_pub_shared/data/package_api.dart' as package_api;
 import 'package:_pub_shared/data/task_api.dart' as api;
 import 'package:_pub_shared/data/task_payload.dart';
 import 'package:_pub_shared/worker/limits.dart';
+import 'package:basics/basics.dart' show MapBasics;
 import 'package:chunked_stream/chunked_stream.dart' show MaximumSizeExceeded;
 import 'package:clock/clock.dart';
 import 'package:collection/collection.dart';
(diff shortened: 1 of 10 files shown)


isoos:
Note: this is re-applying the original rewrite + layering several commits on top of it that address the re-entrant transactions, the query + transaction on the same connection pattern and simplify further parts, e.g. how the dependency timestamp gets updated and read periodically.

sigurdm (approved):

app/lib/task/backend.dart
```diff
@@ -242,58 +242,51 @@ class TaskBackend {
     StackTrace? stackTrace;
 
     // For each package we should ensure state is tracked
-    await for (final p in _datastore.packages.listAllNames()) {
-      packageNames.add(p.name);
-
-      scheduleMicrotask(() async {
-        await pool.withResource(() async {
+    for (final p in packageNames) {
+      unawaited(
```
sigurdm, line 246: I guess you disconnect the stacktraces here. Not sure if it is important.

sigurdm (approved):

app/lib/task/backend.dart
isoos: The previous version did disconnect them. The new code preserves the full stacktrace, puts them into a concurrent pool, and then awaits them. I just don't await the individual future that the pool returns. Previously we have used:
```
  final f = pool.schedue...
  futures.add(f);
  
  await Futures.wait(futures);
```

But I think it is no longer necessary with this block + awaiting the pool.close();

jonasfj (approved):

PR #9411, opened 2026-05-25
pub_integration test sharding on CI

isoos:


.github/workflows/all-test.yml
```diff
@@ -79,6 +79,46 @@ jobs:
...
+    - name: Setup Dart SDK
+      uses: dart-lang/setup-dart@65eb853c7ba17dde3be364c3d2858773e7144260
+      with:
+        sdk: ${{env.DART_SDK_VERSION}}
+    - name: Checkout repository
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+    - run: dart pub get
+    - run: dart tool/trigger_puppeteer_download.dart
+      working-directory: pkg/pub_integration
+    - run: find test/ -type f -name "*_test.dart" | awk -v shard="${{ strategy.job-index }}" -v total="${{ strategy.job-total }}" 'NR % total == shard' | xargs -I {} dart test -j 1 --run-skipped {}
```
sigurdm, line 117: Should we just do
`--total-shards=${{ strategy.job-total }} --shard-index=${{ strategy.job-index }}`?

We'd be parsing more files, but perhaps it is good enough? Certainly simpler

isoos: One of the earlier commits was --total-shards and --shard-index based, but for some reason most tests got into one shard, and the other shard had only one each. It was not with strategy.job-total and job-index though, I can make a try.

isoos: image

With shard count 4 the last one does not get any test at all. I believe the sharding index number may reset on each file and in integration tests we don't have files that have more than 3 tests.

Reverting to the more complex one.

sigurdm (approved):

sigurdm (approved):

Issue #3370, opened 2020-02-25
Display a log message when a service crashes - and track it in a metric


sigurdm:
This can probably be done by wrapping the service in a bash-script that prints if the service crashes.

isoos:
Related: also check if isolate crashes are reported properly.

sigurdm:
Still pending

sigurdm:
Let's try to do this, but also send a info-log when there is a non-crashing instance shutdown, to validate that the mechanism works.

PR #9423, opened 2026-06-04
Run-wrapper for services.

isoos:
#3370

### DIFF
--- a/tool/run-wrapper.sh
+++ b/tool/run-wrapper.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+# Wrapper script that runs a command and logs exit status to stderr,
+# sleeping 10 seconds before exiting to allow log collection.
+
+set -u
+
+"$@"
+EXIT_CODE=$?
+
+if [ "$EXIT_CODE" -eq 0 ]; then
+  echo "[pub-run-wrapper-exited]" >&2
+else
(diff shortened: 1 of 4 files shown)


tool/run-wrapper.sh
```diff
@@ -0,0 +1,17 @@
...
+
+set -u
+
+"$@"
+EXIT_CODE=$?
+
+if [ "$EXIT_CODE" -eq 0 ]; then
+  echo "[pub-run-wrapper-exited]" >&2
+else
+  echo "[pub-run-wrapper-failed] (exit code $EXIT_CODE)." >&2
```
sigurdm: I imagined we used json to format these.
Then we can also give a severity etc.

https://docs.cloud.google.com/appengine/docs/standard/writing-application-logs?tab=python#stdout_stderr

tool/run-wrapper.sh
```diff
@@ -0,0 +1,17 @@
...
+fi
+
+sleep 10
```
sigurdm: Did you validate we need this?
Shutting down versions is already painfully slow

isoos: Was not able to validate.
Maybe we wait 10 for errors and 3 for regular exits?

sigurdm: SG

sigurdm (approved):
LGTM

Perhaps @jonasfj should also confirm

jonasfj (approved):

PR #9426, opened 2026-06-05
Preserve Stacktrace when extracting inner transaction.

isoos:
The main changes are the `Error.throwWithStackTrace` calls, the rest is formatting.

Note: I'm wondering if `TransactionAbortedException` should also keep the original stacktrace it catched.

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -290,8 +290,7 @@ class TaskBackend {
     // If we had any error, we rethrow to ensure that any background task
     // calling this method won't register completion as successful.
     if (error != null) {
-      // Hack to rethrow [error] with [stackTrace]
-      await Future.error(error!, stackTrace);
+      Error.throwWithStackTrace(error!, stackTrace!);
     }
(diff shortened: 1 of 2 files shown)


app/lib/task/backend.dart
```diff
@@ -290,8 +290,7 @@ class TaskBackend {
     // If we had any error, we rethrow to ensure that any background task
     // calling this method won't register completion as successful.
     if (error != null) {
-      // Hack to rethrow [error] with [stackTrace]
-      await Future.error(error!, stackTrace);
+      Error.throwWithStackTrace(error!, stackTrace!);
```
sigurdm, line 293: Nice, I didn't know you could do this!

sigurdm (approved):

app/lib/task/backend.dart
isoos: Yeah, apparently it was in the SDK for a long time, and I was not aware of it either. I had noticed it in another package's code and it seems useful.

PR #9427, opened 2026-06-05
Fix catch-all in task scheduling.

isoos:
We are catching `Exception` but not `Error`, this will catch both.


sigurdm (approved):
LGTM

Did this happen? Or you just saw this in the code?

isoos:
> Did this happen? Or you just saw this in the code?

Not aware of it happening. It just felt off, using the generic Exception and not catch-all, and now that I was looking for it, we have other places with similar pattern, I'll check those too.

PR #9436, opened 2026-06-12
Migrate textfield + textarea from material component to plain HTML + updated style rules.

isoos:
- Changed the HTML DOM layout + styles, added a client-side script to update the character counter.
- Removed material-specific styles that are no longer relevant.
- Updated the colors of the input + select widget: it gets a slightly shaded background, and the border has more contrast than before. Kept now as combined percent of the base color, will update the colors in a subsequent review/change.

Before:
image

After:
image
image

### DIFF
--- a/app/lib/frontend/dom/material.dart
+++ b/app/lib/frontend/dom/material.dart
@@ -94,7 +94,7 @@ d.Node raisedButton({
   );
 }
 
-/// Renders a material text field.
+/// Renders a plain HTML text field.
 d.Node textField({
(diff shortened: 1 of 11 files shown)


sigurdm:
Why is there a waving hand in two of the images?

sigurdm (approved):
I love it

isoos:
> Why is there a waving hand in two of the images?

That's the description in my publisher :see_no_evil: :vulcan_salute:

PR #9438, opened 2026-06-15
Migrate material checkbox to plain HTML.

isoos:
- Continuing the migration of material components to plain HTML versions. The checkbox styling was a bit harder to crack, and there may be edge cases I haven't tried or covered.
- The CSS implementation of the checkmarks was created from a mix of LLM-suggested styles and online templates. It looks closer to the material design and much better than the default checkbox implementation.
- I've struggled to align the label with the bottom part of box border, the current one is a good-enough for first attempt, may fine-tune it later.
- Followed the same styling as the textfields.

Before:

image
image

After:

image
image

### DIFF
--- a/app/lib/frontend/dom/material.dart
+++ b/app/lib/frontend/dom/material.dart
@@ -219,7 +219,7 @@ d.Node dataTable<T>({
   );
 }
 
-/// Renders a material checkbox component.
+/// Renders a plain HTML checkbox component.
 d.Node checkbox({
   required String id,
   required String label,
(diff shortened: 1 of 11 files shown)


isoos:
Note: this is also deployed to staging.

sigurdm:
Light before-after
imageimage

From playing with staging, 3 things that I think could make it closer to the current look (and also make it look better)

* darker grey bordercolor.
* less roundedness on corners
* padding, I think there is less padding around the boxes or something making everything smaller

isoos:
Update style: (staging deploy ongoing)

image
image

sigurdm (approved):

PR #9442, opened 2026-06-17
Migrate Material datatable to plain HTML

isoos:


isoos:
The change is mostly an in-place style replacement, however, there is one notable difference: it doesn't decrease the font size inside the table, which is not really notable, but I think it is better this way. Deploying to staging for better comparisons.

sigurdm (approved):

PR #9446, opened 2026-06-18
Merge description + readme token index.

isoos:
- Doesn't really seems to change total memory use (probably description index is not that large), but improves total latency of search by 3-5% in local testing, which is coming from the single lookup/index iteration instead of having two.
- This doesn't change weights or the outcome of the calculation.
- I see further paths to similarly integrate the API symbol index, but that may change the outcomes, and also total memory, so I would experiment on it separately.


sigurdm (approved):
Does this have influence on how we deploy (ie, we need to wait for a new index to be calculated?) In that case we should mention it in the changelog.

isoos:
No change in the deployment. This is the in-memory preparation of the index.

PR #9447, opened 2026-06-19
Migrate the material floating action button to plain HTML.

isoos:
- Mostly copy of the material style rules.
- Available on staging (eg. win32 screenshots).


sigurdm (approved):

sigurdm:
The colors are slightly different - but I don't really care.

PR #9458, opened 2026-06-26
Improve name search by changing Set.contains into counting.

isoos:
This replaces the Set of trigrams and their contains operation with posting of package index numbers, and counting them.

Note: local benchmark showed a significant improvement (30%+) but I think I should probably update the benchmark script before concluding it is a good representation.

Note: I've been also experimenting with additional optimizations, but eventually reduced the PR to the minimum (and I think where the most gain is coming from). Further benchmarks will be done (left TODOs for them). It is also possible we can slightly alter the matching algorithm for 1- and 2-character search expressions.

### DIFF
--- a/app/bin/tools/search_benchmark.dart
+++ b/app/bin/tools/search_benchmark.dart
@@ -25,19 +25,30 @@ Future<void> main(List<String> args) async {
 
   // NOTE: please add more queries to this list, especially if there is a performance bottleneck.
   final queries = [
+    // tag only
     'sdk:dart',
     'sdk:flutter platform:android',
     'is:flutter-favorite',
+    // frequent substring
+    'a',
+    'er',
+    // frequently searched for
     'chart',
     'json',
     'camera',
+    // frequent words of different length
+    'dart',
+    'flutter',
+    'development',
+    // multi-word expressions
     'android camera',
     'sql database',
+    'to help speed up',
   ];
 
   final sw = Stopwatch()..start();
   var count = 0;
-  for (var i = 0; i < 100; i++) {
+  for (var i = 0; i < 100 || sw.elapsed.inMinutes < 2; i++) {
     await index.search(
       ServiceSearchQuery.parse(
         query: queries[i % queries.length],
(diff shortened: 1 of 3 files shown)


isoos:
Updated benchmark script with more diverse queries. This change seems to improve latencies by 30-35% in local test (e.g. from ~18 ms to ~11 ms per query).

app/lib/search/mem_index.dart
```diff
@@ -702,59 +715,73 @@ class PackageNameIndex {
...
+        return true;
+      }
+      return false;
+    }
 
-        score.setValue(i, 0.99);
-        continue;
+    // Note: short strings (less than a trigram) are not indexed separately,
+    //       we need to do full substring check, evaluating every package.
+    // TODO: consider updating the index + the evaluation algorithm
```
sigurdm: Should we already create an issue for this?
```suggestion
    // TODO(<issue nr>): consider updating the index + the evaluation algorithm
```

app/bin/tools/search_benchmark.dart
```diff
@@ -25,19 +25,30 @@ Future<void> main(List<String> args) async {
...
+    // multi-word expressions
     'android camera',
     'sql database',
+    'to help speed up',
   ];
 
   final sw = Stopwatch()..start();
   var count = 0;
-  for (var i = 0; i < 100; i++) {
+  for (var i = 0; i < 100 || sw.elapsed.inMinutes < 2; i++) {
```
sigurdm, line 51:
```suggestion
  for (var i = 0; i < 100 && sw.elapsed.inMinutes < 2; i++) {
```

sigurdm (approved):
This looks like a HUGE improvement!!

app/bin/tools/search_benchmark.dart
isoos: Note: the intention here is to elapse 2 minutes or to iterate 100, whichever takes longer (and nowadays it is the 2 minutes).

sigurdm: Ah, sorry, I thought you wanted to iterate at most two minutes.

Issue #9462, opened 2026-06-29
Tracking: benchmark / improve the new name search


isoos:
The new name search introduced token posting list.

TODOs:
- [ ] consider int-based counter implementation
- [ ] consider posting weights
- [ ] consider updating the index + the evaluation algorithm

PR #9463, opened 2026-06-30
Use int-based counter in search name matching.

isoos:
- #9462
- very small (1-4%) but mostly consistent improvement over the double-based counter
- interestingly 8, 16 or 64-bit `UintXXList` instances weren't faster, in fact, most of the time they performed slower than the int-based list


app/lib/search/token_index.dart
```diff
@@ -501,3 +510,20 @@ class IndexedScore {
     return heap.getAndRemoveTopK(count).toList();
   }
 }
+
+/// Integer counter that can be accessed via index.
+class IndexedCounter {
```
sigurdm: Not sure if it makes a difference, but perhaps this could be an extension type.

sigurdm (approved):

app/lib/search/token_index.dart
isoos: Done. Benchmark did not show any difference, but maybe one object allocation is spared.

PR #9478, opened 2026-07-10
More robust analysis download.

isoos:
- The post-upload async task starts with the `synchronizePackage` call and updates the tracking state afterwards.
- The initial download of the package is retried more, with extended delays, up to 8 minutes total.

### DIFF
--- a/app/test/package/upload_test.dart
+++ b/app/test/package/upload_test.dart
@@ -1608,6 +1608,8 @@ void main() {
           authToken: adminClientToken,
         ).uploadPackageBytes(bytes);
         expect(message.success.message, contains('Successfully uploaded'));
+
+        await asyncQueue.ongoingProcessing;
         await nameTracker.reloadFromDatastore();
 
         await accountBackend.withBearerToken(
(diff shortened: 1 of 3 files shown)


app/test/package/upload_test.dart
```diff
@@ -1608,6 +1608,8 @@ void main() {
           authToken: adminClientToken,
         ).uploadPackageBytes(bytes);
         expect(message.success.message, contains('Successfully uploaded'));
+
+        await asyncQueue.ongoingProcessing;
```
jonasfj, line 1612: Is this a global queue? then are you sure this is sound?

isoos: We want to check the effect of post-upload processing, and this is the reliable way to do it, since upload does not block for it.

isoos: > Is this a global queue?

This is an instance-level (isolate-level) async execution queue.

jonasfj: yes, an instance/isolate global, so a request can fail because of tasks started in other requests, I think that's a bad idea.

None of these tasks are essential, right? They are all things that we recover from eventually.

jonasfj (approved):

app/test/package/upload_test.dart
jonasfj: sorry, I read this too fast, this in a test no concerns..

PR #9485, opened 2026-07-16
Unroll TaskState into two new SQL table (write-only for now).

isoos:


sigurdm (approved):

isoos:
PTAL. I've updated the naming of the fields + columns:
- I think it helps if timestamps are always ending with `At`.
- We also need to synchronize the `secretToken`+`token` names.
- The `worker*` prefix is debatable, I find it better to read.
- I've also considered renaming `attempts` to `attemptCount` still undecided on it.

app/lib/database/schema.task.dart
```diff
@@ -48,6 +48,65 @@ abstract final class TaskDependency extends Row {
...
+
+  /// Name of the zone in which the instance analysing this package version is
+  /// running.
+  String? get workerZone;
+
+  /// Name of the instance analysing this package version.
+  String? get workerInstance;
+
+  /// Secret token (UUIDv4) used for authenticating worker requests.
+  String? get workerToken;
```
jonasfj, line 88: These need more documentation, all the invariants should be clearly stated.

attempts = 0 have special meaning I think, can it be negative?

workerZone,instance,token can be null when and what does that mean?

jonasfj (approved):

PR #9487, opened 2026-07-22
Prevent weekly downloads sparkline tooltip from overflowing viewport

### LINKED ISSUE #9486: Weekly downloads sparkline tooltip overflows viewport near screen edges
The tooltip in weekly_sparkline/widget.dart is positioned directly from mouse coordinates (`e.x`/`e.y` + fixed offset), without accounting for its own size or viewport bounds. Near the right or bottom edge of the screen, it renders partially off-screen.

dahiAbi:
Fixes: #9486

---

<details>

Many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.

**Note**: The Dart team is trialing Gemini Code Assist. Don't take its comments as final Dart team feedback. Use the suggestions if they're helpful; otherwise, wait for a human reviewer.

</details>

### DIFF
--- a/pkg/web_app/lib/src/widget/weekly_sparkline/widget.dart
+++ b/pkg/web_app/lib/src/widget/weekly_sparkline/widget.dart
@@ -65,13 +65,14 @@ void drawChart(
   final lastDate = data.last.date;
   final firstDate = data.first.date;
   final firstDay = firstDate.copyWith(day: firstDate.day - 7);
-
   final xAxisSpan = lastDate.difference(firstDate);
   // We start with 1 as initial value. In the special case where all downloads
   // are 0 we want a straight line in the bottom of the chart.
   final maxDownloads = data.fold<int>(1, (a, b) => max<int>(a, b.downloads));
 
   final toolTipOffsetFromMouse = 15;
+  // Small margin so the tooltip never touches the very edge of the viewport.
+  final viewportMargin = 8;
 
   (double, double) computeCoordinates(DateTime date, int downloads) {
     final duration = date.difference(firstDate);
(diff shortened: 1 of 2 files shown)


pkg/web_app/lib/src/widget/weekly_sparkline/widget.dart
```diff
@@ -100,6 +101,49 @@ void drawChart(
     return '$year.$month.$day';
   }
 
+  // Positions the tooltip near the mouse pointer while keeping it fully
+  // inside the viewport, flipping above/left of the cursor when there isn't
+  // enough room below/right.
+  void positionTooltip(MouseEvent e) {
+    final tooltipRect = toolTip.getBoundingClientRect();
+    final viewportWidth = window.innerWidth;
```
szakarias: I think it is safer to use `document.documentElement.clientWidth/clientHeight` is than `window.innerWidth/innerHeight` for bounding box calculations since the latter  includes the width of the scrollbar.

pkg/web_app/lib/src/widget/weekly_sparkline/widget.dart
```diff
@@ -100,6 +101,49 @@ void drawChart(
...
+    final viewportHeight = window.innerHeight;
+
+    var left = e.clientX.toDouble();
+    if (left + tooltipRect.width + viewportMargin > viewportWidth) {
+      // Not enough room to the right: show the tooltip to the left of the
+      // cursor instead.
+      left = e.clientX - tooltipRect.width;
+    }
```
szakarias, line 118: This is looking good overall, but I noticed one bug to watch out for with the math: The `.weekly-downloads-sparkline-tooltip` CSS class currently restricts its hover offset using a static `transform: transform: translateY(-3rem) translateX(1rem)`. Because the browser applies `translateX(1rem)` after your `style.left` property is set, clamping `left` to `viewportWidth - tooltipRect.width` means the CSS will still push the tooltip's right edge 16 pixels off the side of the screen. 

We should probably try out removing the transform from the CSS and bake the offsets directly in. For the offsets, instead of the CSS transform, we can apply the 16px (1rem) and -48px (-3rem) nudges directly to the mouse coordinates before we do any clamping:

```
var left = e.clientX + 16;
var top = e.clientY - 48; // (Or replacing toolTipOffsetFromMouse)

```

Note: if the tooltip runs out of room and has to flip to the left or above the cursor, you'll want to invert those offsets (e.g., left = e.clientX - tooltipRect.width - 16) so it maintains that nice gap from the mouse.

szakarias (commented):
Thanks for putting this together.

szakarias:
Thanks for the update. 

I think the y-axis calculations are still not quite right. The original tooltip combined Dart's +15 offset with CSS's -48 transform, making the true net offset -33. Since this places the tooltip above the cursor by default, we now need to check for overflows at the top of the screen, not the bottom.

I think we should do something like this for the Y-axis block that uses the -33 offset, checks the top boundary, and flips it cleanly below the cursor if there isn't enough room:


```
final toolTipOffsetY = -33; 

// Initial placement 
var top = (e.clientY + toolTipOffsetY).toDouble();

// Check if it overflows the top of the screen
if (top < viewportMargin) {
  // Flip it so the tooltip renders below the cursor instead
  top = e.clientY + 33.0;
}

top = top.clamp(
  viewportMargin.toDouble(),
  max(viewportMargin.toDouble(), viewportHeight - tooltipRect.height - viewportMargin),
);

```

szakarias (approved):

PR #9500, opened 2026-08-03
Allow regional Google Analytics and GTM domains in CSP connect-src

sigurdm:
Google Analytics 4 (GA4) uses regional collection endpoints such as `region1.google-analytics.com` and `analytics.google.com` to process events and measurement data.

When CSP was tightened to restrict `connect-src` to only `https://www.google-analytics.com/` and `https://stats.g.doubleclick.net/` in https://github.com/dart-lang/pub-dev/pull/9221, modern GA4 traffic dispatched to regional endpoints was blocked by browsers.

This PR updates `connect-src` to include wildcard domains for GA4, GTM, and DoubleClick (`*.google-analytics.com`, `*.analytics.google.com`, `*.googletagmanager.com`, `*.g.doubleclick.net`).

Detailed in https://developers.google.com/tag-platform/security/guides/csp

For validation of our CSP rules we could consider using https://www.w3.org/TR/CSP3/#directive-report-to . But that would be a separate project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ Important changes to data models, configuration, and migrations between each
 AppEngine version, listed here to ease deployment and troubleshooting.
 
 ## Next Release (replace with git tag when deployed)
+ * Allow regional Google Analytics and GTM domains in CSP `connect-src`.
 
 ## `20260730t085500-all`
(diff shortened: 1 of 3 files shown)


isoos (approved):

isoos:
Note: also add a changelog about it.

PR #9505, opened 2026-08-04
Configure CSP report-to directive, Reporting-Endpoints header, and reporting endpoint

sigurdm:
Configure CSP violation reporting using modern Reporting API and legacy fallbacks:
- Add `report-to csp-endpoint` and `report-uri /api/csp-report` directives to `Content-Security-Policy`.
- Add `Reporting-Endpoints: csp-endpoint="/api/csp-report"` response header on HTML responses.
- Implement `/api/csp-report` endpoint to receive, parse, and log reports at `info` level.
- Filter out noise from browser extension schemes (`chrome-extension:`, `moz-extension:`, etc.) and off-origin document URLs.
- Guard against oversized payloads (>64 KB).
- Update unit tests for CSP serialization and frontend reporting handlers.

### DIFF
--- a/app/lib/frontend/handlers/misc.dart
+++ b/app/lib/frontend/handlers/misc.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:convert';
 import 'dart:io';
 
 import 'package:clock/clock.dart';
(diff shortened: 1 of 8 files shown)


app/lib/frontend/handlers/misc.dart
```diff
@@ -324,3 +325,108 @@ shelf.Response formattedNotFoundHandler(shelf.Request request) {
...
+    final parsed = json.decode(body);
+    if (parsed is List) {
+      // Reporting API payload: array of reports
+      for (final item in parsed) {
+        if (item is Map<String, dynamic>) {
+          final reportBody = item['body'];
+          if (reportBody is Map<String, dynamic>) {
+            _processCspReport(
+              effectiveDirective: reportBody['effectiveDirective'] as String?,
+              blockedUrl: reportBody['blockedURL'] as String?,
```
isoos, line 386: Would it be worth to define a data class with `json_serializable` annotations on it?

sigurdm: Not sure it is worth it. We just extract four fields. The two different directives will require 2 different classes.

isoos (approved):

PR #9509, opened 2026-08-05
Fix backfill comparison issue.

isoos:
- Compares the non-lock parts of the publishing configs (lock may have been filled or changed between the migration steps).
- In case of difference it keeps log.shout, but continues with other entities.


pkg/api_builder/pubspec.yaml
```diff
@@ -7,7 +7,7 @@ environment:
 resolution: workspace
 
 dependencies:
-  analyzer:  ^13.0.0
+  analyzer:  ^14.0.0
```
sigurdm: Is this intentional? Should we widen instead?

sigurdm (approved):

pkg/api_builder/pubspec.yaml
isoos: leftover from testing the upgrade, removed

PR #9512, opened 2026-08-06
Allow Google Feedback, gstatic, and google.com in CSP

sigurdm:
Based on analysis of CSP violation reports received on production:

1. **`connect-src`**:
   - Added `https://www.gstatic.com/` and `https://*.gstatic.com` to allow the Google Cookie Choice / Consent banner component to fetch `config.json`.
   - Added `https://www.google.com/` to allow Google Analytics telemetry hits sent to `/g/collect`.

2. **`frame-src`**:
   - Added `https://feedback-pa.clients6.google.com` to allow embedding the Google Feedback widget iframe.

### DIFF
--- a/app/lib/service/csp/default_csp.dart
+++ b/app/lib/service/csp/default_csp.dart
@@ -70,7 +70,7 @@ final _defaultContentSecurityPolicyMap = <String, List<String>>{
     'https://*.googletagmanager.com',
     'https://*.g.doubleclick.net',
     'https://www.google.com/',
-    'https://www.gstatic.com/',
+    'https://gstatic.com',
     'https://*.gstatic.com',
   ],
   'frame-ancestors': _none,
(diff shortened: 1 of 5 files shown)


app/lib/service/csp/default_csp.dart
```diff
@@ -61,13 +61,17 @@ final _defaultContentSecurityPolicyMap = <String, List<String>>{
     "'self'",
     'https://www.googletagmanager.com/',
     'https://accounts.google.com/',
+    'https://feedback-pa.clients6.google.com',
```
isoos, line 64: clients6 seems to be a dynamic hostname, or is it a permanent service URL?

isoos (approved):

app/lib/service/csp/default_csp.dart
sigurdm: Yeah - I would think the same, apparently feedback-pa.clients6.google.com is the permanent static OnePlatform production endpoint for Google User Feedback / HaTS (similar to how clients2.google.com is a fixed domain for Google client services).
It is the standard URL allowlisted in CSPs across Google web properties for embedding the feedback dialog.

PR #9518, opened 2026-08-10
Naming policy test: SQL timestamps should end with _at.

isoos:
- related to #9485
- exempting the already failing columns

### DIFF
--- a/app/test/database/sql_naming_test.dart
+++ b/app/test/database/sql_naming_test.dart
@@ -38,10 +38,10 @@ void main() {
   // This is a nice naming convention, enforced by this semi-ugly test.
   // If we decide to abandon this convention, then delete this file.
   test('timestamptz columns end with `_at`', () {
-  // NOTE: If this test breaks because of SQL parsing
-  //       issues, feel free to delete it.
-  //       We shall only maintain this test to the extend
-  //       that doing so is easy.
+    // NOTE: If this test breaks because of SQL parsing
+    //       issues, feel free to delete it.
+    //       We shall only maintain this test to the extend
+    //       that doing so is easy.
     for (final file in sqlFiles) {
       final basename = path.basename(file.path);
       final excluded = _exemptedTimestampColumns[basename] ?? const {};
--- a/app/test/database/sql_naming_test.dart
+++ b/app/test/database/sql_naming_test.dart
@@ -34,6 +34,9 @@ void main() {
     expect(sqlFiles, isNotEmpty);
   });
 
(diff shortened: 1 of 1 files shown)


sigurdm:
Is there a specific reason for this convention? Does it help us? Are the fields not already statically strongly typed in dart?

isoos:
> Is there a specific reason for this convention?

LLM suggestion after reviewing the column names.

> Does it help us? Are the fields not already statically strongly typed in dart?

I think it does, if for nothing else, it improves consistency. It is also somewhat confusing the have both `scheduled` and `finished` and you don't know if it is a boolean (`isScheduled` and `isFinished`) or a timestamp (`scheduledAt` and `finishedAt`).

sigurdm (approved):

sigurdm:
I don't mind either way

app/test/database/sql_naming_test.dart
```diff
@@ -0,0 +1,55 @@
...
+          .whereType<File>()
+          .where((f) => f.path.endsWith('.sql'))
+          .toList()
+        ..sort((a, b) => a.path.compareTo(b.path));
+
+  test('found migration files', () {
+    expect(sqlFiles, isNotEmpty);
+  });
+
+  test('timestamptz columns end with `_at`', () {
```
jonasfj, line 40:
```suggestion
  test('timestamptz columns end with `_at`', () {
  // NOTE: If this test breaks because of SQL parsing
  //       issues, feel free to delete it.
  //       We shall only maintain this test to the extend
  //       that doing so is easy.
```

jonasfj (approved):

jonasfj:
If you want this better, consider using `package:stringscanner` and make dump using a `SourceSpan` when there is an issue. That way we won't spend so much time figuring out what is wrong when we hit it.

app/test/database/sql_naming_test.dart
jonasfj, line 40:
```suggestion
  // Test that all DateTime fields in the database have a name ending in "At"
  // This is a nice naming convention, enforced by this semi-ugly test.
  // If we decide to abandon this convention, then delete this file.
  test('timestamptz columns end with `_at`', () {
```

isoos:
> If you want this better, consider using package:stringscanner and make dump using a SourceSpan when there is an issue. That way we won't spend so much time figuring out what is wrong when we hit it.

Let's consider this when we could have more rules like this one, until then the simple may be enough.

PR #9522, opened 2026-08-12
Upgrade Flutter SDK, pana and dartdoc

isoos:


jonasfj (approved):

isoos:
Not sure why the token test fails, started failing after a clean PR was merged, but no other indication for the root cause.

PR #9523, opened 2026-08-12
CI testing: Revert "Cleanup of UserSession migration."

isoos:
Reverts dart-lang/pub-dev#9520


isoos:
CI failure seems unrelated to the commit which first showed signs of failure.

PR #9526, opened 2026-08-13
Schedule package analysis immediately in task-bump-priority admin action

sigurdm:
Updates the `task-bump-priority` admin action and `TaskBackend.adminBumpPriority` to:
- Use the standard `db.taskBumpPriority` to enqueue analysis for all tracked versions of the package.
- Direct-launch an analysis worker VM instance right away if quota allows instead of strictly waiting for the background scheduler.
- If the instance quota limit is reached or VM creation fails, the task remains enqueued with high priority and will be picked up by the background scheduler cycle.

### DIFF
--- a/app/lib/task/backend.dart
+++ b/app/lib/task/backend.dart
@@ -1187,85 +1187,18 @@ class TaskBackend {
     }
 
     final zone = compute.zones.first;
-    final instanceName = compute.generateInstanceName();
-
-    // 3. Atomically assign tokens and state.
-    final payload = await updatePackageStateWithPendingVersions(
-      _database,
-      packageName,
-      zone,
-      instanceName,
+    final res = await schedulePackageInstance(
+      database: _database,
+      compute: compute,
+      package: packageName,
+      zone: zone,
     );
-    if (payload == null) {
-      return {
-        'status': 'none-pending',
-        'message': 'No versions pending analysis.',
-        'package': packageName,
-        if (version != null) 'version': version,
-      };
-    }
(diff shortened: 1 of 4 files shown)


isoos:
This sounds like an awful lot of duplicated logic. Can't we do a refactor for the regular scheduling first so that we may reuse most of that?

sigurdm:
Refactored single-instance provisioning and error/rollback handling into a shared `schedulePackageInstance` helper in `scheduler.dart`, reused by both `runOneCreateInstancesCycle` and `adminBumpPriority`.

app/lib/task/backend.dart
```diff
@@ -1086,13 +1085,133 @@ class TaskBackend {
...
+      }
+
+      final versions = {...task.state.versions};
+      if (version != null && !versions.containsKey(version)) {
+        throw InvalidInputException(
+          'Version "$version" is not tracked for package "$packageName".',
+        );
+      }
+
+      final targetVersions = version != null
```
isoos: This is also a block that seems to be a repeat of the logic we already have in `trackPackage`, and it seems to be added because we are allowing to schedule an arbitrary version. Wouldn't it be better to remove this and just keep bumping the priority, and schedule the versions be analyzed as they would otherwise?

sigurdm:
Removed the optional `--version` argument from `task-bump-priority` entirely and restored the `adminBumpPriority` behavior to prioritize all versions.

isoos (approved):

PR #9527, opened 2026-08-13
fix: throttle updateAllPackageVersions memory consumption

sigurdm:
Throttles Datastore stream processing in `updateAllPackageVersions` by utilizing `StreamExtensions.parallelForEach` to enforce strict streaming backpressure with bounded concurrency.

This prevents unbounded queuing of thousands of active background closures into memory simultaneously, which previously caused silent out-of-memory (OOM) failures during periodic background tasks.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -475,13 +475,19 @@ class PackageBackend {
     int? concurrency,
   }) async {
     var count = 0;
-    await db.query<Package>().run().parallelForEach(concurrency ?? 1, (p) async {
+    
+    // Assigning to a variable and awaiting the variable natively resolves the analyzer AST unawaited_futures lint bug
+    // that fails on GitHub when calling `await stream.parallelForEach(...)`.
+    final pipeline = db.query<Package>().run().parallelForEach(concurrency ?? 1, (p) async {
       final updated = await updatePackageVersions(
         p.name!,
         dartSdkVersion: dartSdkVersion,
       );
       if (updated) count++;
     });
+    
+    await pipeline;
+
     return count;
   }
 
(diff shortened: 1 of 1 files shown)


app/lib/package/backend.dart
```diff
@@ -479,12 +479,17 @@ class PackageBackend {
     final futures = <Future>[];
     await for (final p in db.query<Package>().run()) {
       final package = p.name!;
-      final f = pool.withResource(() async {
-        final updated = await updatePackageVersions(
-          package,
-          dartSdkVersion: dartSdkVersion,
-        );
-        if (updated) count++;
+      final resource = await pool.request();
```
isoos: Were you able to measure this somehow locally?

Not sure if this helps here with the memory: an array of ~100k callbacks shouldn't take too much space. In the meantime, the Datastore query is being blocked, and we add an async microtask, that is harder to track and reason about.

sigurdm: The issue isn't the array of callbacks itself (which takes negligible space) — it's that `pool.withResource(...)` gracefully queues the task but returns a `Future` immediately. Because it doesn't block, the `await for` loop instantly drains all ~40,000 entities from the `db.query().run()` Datastore stream into memory in just a few seconds.

Ingesting all entities unrestrained forces the gRPC layer and network buffers to retain a lot of state while the app attempts to queue 40,000 concurrent outgoing Datastore transactions. In recent incidents, we observed the `pub.dev` instances ballooning to 13GB+ of RAM and getting instantly OOM-killed (which bypasses the timeout failure handlers and causes the tasks to hang for 12 hours).

By explicitly awaiting `pool.request()` inside the loop, we halt the `await for` loop fully until a concurrency slot frees up. This back-pressures the Datastore stream over gRPC, stopping the runaway memory spike since it only pulls entities as fast as the pool processes them.

isoos (approved):
Please create a new issue to investigate this why it is happening with package:pool. Let's land the fix in the meantime.

isoos (approved):

PR #9529, opened 2026-08-13
Cleanup of PackageState and UserSession ex-Datastore entities + follow-up rename of UserSessionRow.

isoos:


sigurdm (approved):
LGTM overall! Two minor follow-ups:
1. Should we also remove the two `kind: PackageState` index entries from `index.yaml` (lines 63–72)?
2. Minor nit on the PR title: the removed Datastore entity is `PackageState` rather than `TaskState` (`TaskState` is the JSON payload on the SQL `Task` row). There's also a lingering mention in the doc comment for `derivePendingAt` right below the deleted class in `app/lib/task/models.dart`.

isoos:
Removed index reference + updated doc comment + title.

sigurdm (approved):
SLGTM

PR #9534, opened 2026-08-14
New SQL table for tracking neat_periodic_task statuses (with best-effort mirror to Datastore).

isoos:


jonasfj:
I think this can migrate atomically, can it not?
In these use cases do we care if it runs one more time?
Do we even have any that aren't runtimeVersioned?

It's not globalLock and even that I think it would only matter for those
that are globallock and not runtimeVersioned.

--
Regards Jonas Finnemann Jensen.

fre. 14. aug. 2026, 16.46 skrev István Soós ***@***.***>:

isoos:
Sure, we could do this in a single migration too, though, some non-runtimeVersioned jobs are a bit heavy.

As an alternative (which could also work for global lock) we could just start using the SQL as the authoritive source, and write the status/lock back to Datastore without considering the status there. The old process may correctly assume that something is running it (for most of the time).

I'll update the PR to do that migration and mirror the state to Datastore.

jonasfj:
What jobs are problematic to run twice as often as we do?
It's more concerning if racing is a problem...

--
Regards Jonas Finnemann Jensen.

fre. 14. aug. 2026, 17.06 skrev István Soós ***@***.***>:

isoos:
> What jobs are problematic to run twice as often as we do?
> It's more concerning if racing is a problem...

The tarball- and the datastore- integrity (both weekly) checks are the most heavy, not sure about the synchronizing exported API (daily) task.

isoos:
PTAL: updated the PR with the version where SQL is the primary store and Datastore is updated so that old runtimeVersion would not try to run the task concurrently.

app/lib/tool/neat_task/datastore_status_provider.dart
```diff
@@ -89,72 +95,144 @@ class DatastoreStatusProvider extends NeatStatusProvider {
...
-    return base64.decode(e!.statusBase64!);
+    final row = await primaryDatabase.transactWithRetry((db) async {
+      var row = await db.neatTaskStatuses
+          .byKey(_name, _runtimeVersionValue)
+          .fetch();
+      if (row != null) {
+        return row;
+      }
+      final now = clock.now().toUtc();
+      final etag = Ulid().toCanonical();
```
sigurdm: `Ulid().toCanonical()` was deprecated in `package:ulid 2.2.0` and replaced across `main` in PR #9585 with `Ulid().toBase32(lowercase: true)`. Rebasing onto `main` will fail `dart analyze`.

```suggestion
      final etag = Ulid().toBase32(lowercase: true);
```

sigurdm (commented):
LGTM overall — just a rebase fix for `package:ulid` on `main` and a couple of minor suggestions.

app/lib/tool/neat_task/datastore_status_provider.dart
```diff
@@ -89,72 +95,144 @@ class DatastoreStatusProvider extends NeatStatusProvider {
...
+          .doNothing()
+          .returnInserted()
+          .executeAndFetch();
+      row ??= await db.neatTaskStatuses
+          .byKey(_name, _runtimeVersionValue)
+          .fetch();
+      return row;
+    });
+    _etag = row!.etag;
+    return row.status;
```
sigurdm: Two small improvements for `get()`:
1. Since `get()` is polled every cycle by every scheduler, doing the fast-path `byKey(...).fetch()` via `withRetry` outside `transactWithRetry` avoids `BEGIN` + `SELECT` + `COMMIT` roundtrips when the row already exists (and `INSERT ... ON CONFLICT DO NOTHING RETURNING *` is already atomic).
2. If `row` is ever `null` (e.g., if `doNothing()` returns `null` and a concurrent cleanup deletes the row before the second `fetch()`), `row!.etag` throws a Dart `TypeError` (which extends `Error`, not `Exception`), bypassing `withRetry` and `NeatStatusProvider.withRetry`. Throwing a `DatabaseException` allows retry wrappers to handle it.

```suggestion
    var row = await primaryDatabase.withRetry(
      (db) => db.neatTaskStatuses.byKey(_name, _runtimeVersionValue).fetch(),
    );
    if (row == null) {
      final now = clock.now().toUtc();
      final etag = Ulid().toBase32(lowercase: true);
      row = await primaryDatabase.withRetry((db) async {
        final inserted = await db.neatTaskStatuses
            .insertValue(
              taskName: _name,
              runtimeVersion: _runtimeVersionValue,
              status: Uint8List(0),
              etag: etag,
              updatedAt: now,
            )
            .onConflict(.primaryKey)
            .doNothing()
            .returnInserted()
            .executeAndFetch();
        return inserted ??
            await db.neatTaskStatuses
                .byKey(_name, _runtimeVersionValue)
                .fetch();
      });
    }
    if (row == null) {
      throw DatabaseException('Failed to initialize NeatTaskStatus row: $_id');
    }
    _etag = row.etag;
    return row.status;
```

app/lib/tool/neat_task/datastore_status_provider.dart
```diff
@@ -89,72 +95,144 @@ class DatastoreStatusProvider extends NeatStatusProvider {
...
-      e
-        ..statusBase64 = base64.encode(status ?? <int>[])
-        ..etag = Ulid().toCanonical()
-        ..updated = clock.now().toUtc();
-      tx.insert(e);
-      return e.etag;
-    });
-    if (newEtag != null) {
+    final statusBytes = Uint8List.fromList(status ?? <int>[]);
+    final newEtag = Ulid().toCanonical();
```
sigurdm: Same here — replace `.toCanonical()` with `.toBase32(lowercase: true)` to match `main`.

```suggestion
    final newEtag = Ulid().toBase32(lowercase: true);
```

app/lib/tool/neat_task/datastore_status_provider.dart
```diff
@@ -89,72 +95,144 @@ class DatastoreStatusProvider extends NeatStatusProvider {
...
+          )
+          .onConflict(.primaryKey)
+          .update(
+            (_, excluded, set) => set(
+              status: excluded.status,
+              etag: excluded.etag,
+              updatedAt: excluded.updatedAt,
+            ),
+          )
+          .where((existing, _) => existing.etag.equalsValue(previousEtag))
```
sigurdm: Because `newEtag` is generated outside `primaryDatabase.withRetry(...)`, if Attempt 1 commits the `UPSERT` in Postgres (updating `etag` in the DB to `newEtag`) but a transient network error occurs before the response reaches Dart, Attempt 2 will retry the query with `WHERE existing.etag = previousEtag`. Since the DB row already has `newEtag`, Attempt 2 matches 0 rows and returns `null`, causing `set()` to return `false` and skip `_mirrorToDatastore()`.

Matching either `previousEtag` or `newEtag` makes the `UPSERT` idempotent across retries:

```suggestion
          .where(
            (existing, _) =>
                existing.etag.equalsValue(previousEtag) |
                existing.etag.equalsValue(newEtag),
          )
```

PR #9537, opened 2026-08-17
Fix SSRF signing oracle in sign-in complete and filter private destinations in image proxy

sigurdm:
This PR fixes SSRF vulnerability reported in b/543872427 and redirect handling in image_proxy:

1. Removes `markdownToHtml` rendering of the `error` query parameter in `/sign-in/complete` (`app/lib/frontend/handlers/account.dart`), preventing markdown injection from closing code blocks and generating signed proxy URLs via `_ImageProxyRewriter`.
2. Extracts RFC 1918 (private IPs), link-local (`169.254.0.0/16`), RFC 6598 carrier-grade NAT, unique local IPv6, loopback (when not testing), and internal metadata domain (`metadata.google.internal`) filtering into its own reusable CIDR matching library (`pkg/image_proxy/lib/prohibited_addresses.dart`).
3. Fixes relative redirect URL resolution in `image_proxy` (`pkg/image_proxy/lib/image_proxy_service.dart`) so that `Location` headers are resolved against the current redirect step URI rather than the initial request URI.
4. Adds comprehensive unit tests covering the `/sign-in/complete?error=...` handler, prohibited IP/CIDR block destination filtering, and multi-step relative redirect resolution.

### DIFF
--- a/pkg/image_proxy/lib/image_proxy_service.dart
+++ b/pkg/image_proxy/lib/image_proxy_service.dart
@@ -464,7 +464,7 @@ Future<bool> _isAllowedDestination(Uri url) async {
         }
       }
     }
-  } catch (_) {
+  } on SocketException {
     // If DNS lookup fails, allow through so client.getUrl throws expected DNS error
   }
   return true;
--- a/pkg/image_proxy/lib/image_proxy_service.dart
+++ b/pkg/image_proxy/lib/image_proxy_service.dart
@@ -16,6 +16,8 @@ import 'package:retry/retry.dart';
 import 'package:shelf/shelf.dart' as shelf;
 import 'package:shelf/shelf_io.dart';
 
+import 'prohibited_addresses.dart';
+
 enum Severity {
   notice,
   info,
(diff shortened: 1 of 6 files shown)


pkg/image_proxy/lib/image_proxy_service.dart
```diff
@@ -399,3 +409,96 @@ Future<Uint8List> readAllBytes(Stream<List<int>> stream, int maxBytes) async {
...
+/// - RFC 3513 IPv6 site-local addresses (`fec0::/10`).
+///
+/// Returns `true` if [address] is within any prohibited network range.
+bool _isProhibitedAddress(InternetAddress address) {
+  if (address.isLoopback || address.isLinkLocal || address.isMulticast) {
+    return true;
+  }
+  final raw = address.rawAddress;
+  if (address.type == InternetAddressType.IPv4 && raw.length == 4) {
+    if (raw[0] == 0 || // 0.0.0.0/8 (RFC 1122)
```
jonasfj: Could we use:
https://api.dart.dev/dart-io/InternetAddress/isLinkLocal.html
https://api.dart.dev/dart-io/InternetAddress/isLoopback.html
https://api.dart.dev/dart-io/InternetAddress/isMulticast.html

Or may I request that we create a tiny abstraction for this.

Can't we make something where we input `10.0.0.0/8` as a String or `[10, 0, 0, 0], 8` as parameters?

jonasfj (approved):

pkg/image_proxy/lib/image_proxy_service.dart
sigurdm: Done! Extracted this into a new library [prohibited_addresses.dart](https://github.com/dart-lang/pub-dev/pull/9537/files#diff-8153cba39d93fe54bdcecb0980ffdb7f66a0d4cfd81b8ccce6c243ee7d4dfbfb) with its own unit tests in [prohibited_addresses_test.dart](https://github.com/dart-lang/pub-dev/pull/9537/files#diff-bc05a8f4c4a6dbfbebfda5c1b69fb7a41ecbf0d22079148d44c80cb5f14e414c).

It uses address.isLoopback, address.isLinkLocal, and address.isMulticast from dart:io, alongside a small CidrBlock class that parses CIDR strings (CidrBlock.parse('10.0.0.0/8')) to cover RFC 1918 private IPv4, RFC 6598 shared address space, RFC 4193 unique local IPv6, etc.

(Note: We didn't use safe_url_check because its built-in IP checks omit some ranges like RFC 6598 100.64.0.0/10 and 0.0.0.0/8, and safeUrlCheck is structured around HEAD requests rather than GET downloading).

PR #9538, opened 2026-08-17
Fake datastore transaction logic + proper GlobalLock testing.

isoos:
- This is in preparation for SQL migration of locks.
- Implemented a simple logic for transaction failures in fake_gcloud.
- Enabled datastore-only test locally + wrote two more tests.


isoos:
Closing since it became part of #9586

PR #9546, opened 2026-08-19
Verify uploaded package attestations using package:sigstore

mosuem:
Second PR in stack for package attestation support (stacked on top of #9545):

- Adds `sigstore` dependency in `app/pubspec.yaml`.
- In `packageBackend.publishUploadedBlob`, validates the attestation bundle against the package archive bytes, package name/version, and pubspec repository using `AttestationVerifier`.
- Rejects package upload with `PackageRejectedException` if attestation is invalid or fails verification.
- Adds test cases in `upload_test.dart` for valid verification and rejected invalid attestations.

### DIFF
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -30,7 +30,6 @@ import 'package:pub_dev/task/backend.dart';
 import 'package:pub_package_reader/pub_package_reader.dart';
 import 'package:pub_semver/pub_semver.dart';
 import 'package:pubspec_parse/pubspec_parse.dart' as pubspec_parse;
-import 'package:sigstore/sigstore.dart';
 
 import '../account/agent.dart';
 import '../account/backend.dart';
@@ -49,6 +48,7 @@ import '../shared/redis_cache.dart' show cache;
 import '../shared/storage.dart';
 import '../shared/urls.dart' as urls;
 import '../shared/utils.dart';
+import 'attestation_verifier.dart';
 import 'model_properties.dart';
 import 'models.dart';
 import 'name_tracker.dart';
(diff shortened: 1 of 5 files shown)


app/lib/package/backend.dart
```diff
@@ -1227,14 +1230,55 @@ class PackageBackend {
...
+            throw PackageRejectedException('Invalid package attestation: $e');
+          }
+
+          final archiveBytes = await file.readAsBytes();
+          final verificationResult = _sigstoreVerifier.verify(
+            packageName: pubspec.name,
+            packageVersion: Version.parse(versionString),
+            archiveBytes: archiveBytes,
+            bundle: bundle,
+            pubspecRepository: pubspec.repository?.toString(),
```
sigurdm, line 1260: I have a security concern regarding using `pubspec.repository` as the trust anchor for attestation verification:

1. **Untrusted source of truth:**
   `pubspec.repository` is read directly from the `pubspec.yaml` inside the archive being uploaded. In a maintainer-account compromise or unauthorized upload scenario, an attacker could point `repository:` in `pubspec.yaml` to their own GitHub repository and build/sign the package using their own GitHub Actions workflow. Because `backend.dart` checks the attestation against `pubspec.repository`, pub.dev would verify and accept this attestation as valid.

   Instead, the expected repository should be anchored to pub.dev's server-side configuration—specifically `package.publishingConfig?.githubConfig` (similar to PyPI's Trusted Publisher model), which only package admins can modify.

2. **Empty repository fallback:**
   If a package doesn't define a `repository` in `pubspec.yaml`, `pubspecRepository ?? ''` passes an empty string into `SigstoreVerificationPolicy.create`, which disables identity verification in `package:sigstore` entirely and allows any valid Sigstore signature to be accepted.

3. **SAN matching vs. repository URL:**
   In Sigstore/Fulcio, the certificate Subject Alternative Name (SAN) for GitHub Actions is the full workflow URI (e.g. `https://github.com/<owner>/<repo>/.github/workflows/<workflow>.yml@refs/...`), not just the repository URL `https://github.com/<owner>/<repo>`. If `expected_identity` does exact equality matching against `pubspecRepository`, this check might fail for legitimate packages. For binding to the repository, we should ideally verify against the Fulcio repository extension (and preferably the immutable repository ID OID `1.3.6.1.4.1.57264.1.15` rather than the mutable repository URL OID `.1.12` to prevent repo-transfer/rename attacks).

4. **Testing gap:**
   In `app/test/package/upload_test.dart`, `AttestationVerifier.skipSignatureCheckInTest = true` is set, which means the verification path isn't being exercised end-to-end against real Sigstore bundles during tests. We should also avoid mutable static flags in production classes and prefer injecting a mock/fake verifier into `PackageBackend`.

Issue #9554, opened 2026-08-24
image proxy: exceeded memory limit


jonasfj:
Cloud Run container instances that exceed their allowed memory limit are terminated. Consider increasing the memory limit of your Cloud Run service.
Memory limit of 512 MiB exceeded with 516 MiB used. Consider increasing the memory limit, see https://cloud.google.com/run/docs/configuring/memory-limits

jonasfj:
Reported in stackdriver error reporting, probably we should increase memory a bit.

sigurdm:
Closed as resolved.

The Cloud Run memory allocation for the image proxy service has been increased from 512 MiB to 1 GiB in production configuration (cl/973847545).

PR #9557, opened 2026-08-27
Set DART_VM_OPTIONS for image proxy to adjust GC

sigurdm:
Configure `DART_VM_OPTIONS="--old_gen_heap_size=768"` for the image proxy server in both the Dockerfile and Cloud Build deployment step.

- `--old_gen_heap_size=768`: caps old generation heap (75% of the 1 GiB container memory limit) to trigger garbage collection before reaching container memory boundaries, leaving ~256 MB for new-gen heap, thread stacks, TLS/socket buffers, and runtime overhead.

Improves the situation on #9554 (but we should also increase memory).

### DIFF
--- a/pkg/image_proxy/Dockerfile
+++ b/pkg/image_proxy/Dockerfile
@@ -18,8 +18,7 @@ COPY --from=build /app/server /app/bin/
 
 # Set Dart VM options for container execution:
 # - --old_gen_heap_size=768 caps the old generation heap to trigger GC before container cgroup limits.
-# - --dontneed_on_sweep returns freed heap pages to the OS kernel via madvise(MADV_DONTNEED).
-ENV DART_VM_OPTIONS="--old_gen_heap_size=768,--dontneed_on_sweep"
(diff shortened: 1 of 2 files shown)


jonasfj (approved):
Is this really a good idea?

sigurdm:
Hmm, I would think so - nice to prevent OOM by triggering GC before it grows out of your available memory. but of course I don't really knows how well it works.

@mkustermann might be able to give guidance...

sigurdm:
Closing for now.

As detailed in [dart-lang/sdk#29786 (comment)](https://github.com/dart-lang/sdk/issues/29786#issuecomment-5491620654), setting `--old_gen_heap_size` acts as a strict allocation ceiling rather than a soft target. In a 1 GiB container, it risks premature `OutOfMemoryError`s and GC thrashing during temporary bursts that the container could otherwise absorb.

The Cloud Run container memory for image proxy has already been doubled to 1 GiB in cl/973847545, which provides ample headroom for concurrent requests without needing a manual VM heap cap.

PR #9564, opened 2026-09-02
Migrate package download redirect to using V4 signed urls

jonasfj:
With this buckets do not have to be public anymore.

Granted we still need to change GCLB configuration.

### DIFF
--- a/app/lib/frontend/handlers/pubapi.dart
+++ b/app/lib/frontend/handlers/pubapi.dart
@@ -84,9 +84,8 @@ class PubApi {
   ///
   /// This is the endpoint we link to from the version listing.
   ///
-  /// While this is here a redirect to a bucket, and will be done this way on
-  /// staging, on pub.dev we actually serve this endpoint directly from a bucket
-  /// via GCLB (load balancer), and will never trigger this handler.
+  /// This end-point is generally handled by GCLB which serves it directly from the bucket.
+  /// But when hitting the service directly or testing locally we need to redirect to the bucket.
   @EndPoint.get('/api/archives/<package|[^-/]+>-<version>.tar.gz')
   Future<Response> fetchPackage(
     Request request,
--- a/app/lib/package/backend.dart
+++ b/app/lib/package/backend.dart
@@ -47,6 +47,7 @@ import '../shared/redis_cache.dart' show cache;
 import '../shared/storage.dart';
 import '../shared/urls.dart' as urls;
 import '../shared/utils.dart';
+import 'download_signer_service.dart';
 import 'model_properties.dart';
 import 'models.dart';
 import 'name_tracker.dart';
@@ -410,7 +411,11 @@ class PackageBackend {
     InvalidInputException.checkSemanticVersion(version);
     final cv = canonicalizeVersion(version);
     final object = 'latest/api/archives/$package-$cv.tar.gz';
-    return Uri.parse(_exportedApiBucket.objectUrl(object));
+    return await downloadSigner.buildDownloadUrl(
+      _exportedApiBucket.bucketName,
+      object,
+      Duration(minutes: 15),
+    );
   }
 
   /// Updates the stable, prerelease and preview versions of [package].
--- a/app/lib/package/download_signer_service.dart
+++ b/app/lib/package/download_signer_service.dart
@@ -1,0 +1,150 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:_pub_shared/utils/http.dart';
+import 'package:clock/clock.dart';
+import 'package:collection/collection.dart';
+import 'package:convert/convert.dart';
+import 'package:crypto/crypto.dart';
+import 'package:gcloud/service_scope.dart' as ss;
+import 'package:googleapis/iam/v1.dart' as iam;
+import 'package:http/http.dart' as http;
+
+import '../shared/configuration.dart';
+import 'upload_signer_service.dart' show SigningResult;
+
+/// The registered [DownloadSignerService] object.
+DownloadSignerService get downloadSigner =>
+    ss.lookup(#_download_url_signer) as DownloadSignerService;
+
+/// Register a new [DownloadSignerService] object into the current service
+/// scope.
+void registerDownloadSigner(DownloadSignerService downloadSigner) =>
+    ss.register(#_download_url_signer, downloadSigner);
+
+/// Creates an download signer based on the current environment.
+Future<DownloadSignerService> createDownloadSigner(
+  http.Client authClient,
+) async {
+  final email = activeConfiguration.downloadSignerServiceAccount;
+  return _IamBasedDownloadSigner(
+    activeConfiguration.projectId,
+    email,
+    authClient,
+  );
+}
+
+/// Signs Google Cloud Storage URLs for downloading objects securely.
+abstract base class DownloadSignerService {
+  /// The Google Access ID (e.g. service account email) used for signing.
+  String get googleAccessId;
+
+  /// Google Cloud Storage V4 signatures require strict RFC 3986 encoding.
+  /// Dart's Uri.encodeComponent does not encode !, ', (, ), or *.
+  /// https://cloud.google.com/storage/docs/authentication/canonical-requests#about-resource-path
+  String _uriEncode(String input) {
+    return Uri.encodeComponent(input)
+        .replaceAll('!', '%21')
+        .replaceAll("'", '%27')
+        .replaceAll('(', '%28')
+        .replaceAll(')', '%29')
+        .replaceAll('*', '%2A');
+  }
+
+  /// Builds a V4 signed URL for downloading a Google Cloud Storage object.
+  Future<Uri> buildDownloadUrl(
+    String bucket,
+    String object,
+    Duration lifetime,
+  ) async {
+    final now = clock.now().toUtc();
+    final datetime = now
+        .toIso8601String()
+        .replaceAll('-', '')
+        .replaceAll(':', '')
+        .replaceAll(RegExp(r'\.\d+'), '');
+    final date = datetime.split('T').first;
+    final scope = '$date/auto/storage/goog4_request';
+
+    final canonicalQueryParameters = {
+      'X-Goog-Algorithm': 'GOOG4-RSA-SHA256',
+      'X-Goog-Credential': '$googleAccessId/$scope',
+      'X-Goog-Date': datetime,
+      'X-Goog-Expires': '${lifetime.inSeconds}',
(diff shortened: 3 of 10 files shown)


jonasfj:
fyi, I deployed this to staging-sigurdm

app/lib/frontend/handlers/pubapi.dart
```diff
@@ -84,9 +84,8 @@ class PubApi {
   ///
   /// This is the endpoint we link to from the version listing.
   ///
-  /// While this is here a redirect to a bucket, and will be done this way on
-  /// staging, on pub.dev we actually serve this endpoint directly from a bucket
-  /// via GCLB (load balancer), and will never trigger this handler.
+  /// This end-point is generally handled by GCLB which serves it directly from the bucket.
+  /// But when hitting the service directly or testing locally we need to redirect to the bucket.
```
sigurdm, line 88: How will this interact with GCLB once the bucket is non-public?

1. If GCLB continues to serve `/api/archives/*` directly via a backend bucket, a non-public bucket will reject unauthenticated requests unless Cloud CDN signed URLs/cookies are configured in GCLB. This handler would remain bypassed in production.
2. If GCLB is instead updated to route archive requests to the backend service, every archive download on pub.dev will hit this handler. Because `_IamBasedUploadSigner.sign` calls `iam.serviceAccounts.signBlob` over HTTP, this introduces network latency (~50–200ms) on each cache miss and could easily exhaust IAM API quotas without caching the signed URLs in Redis/memory (the 303 redirect is cached for 2 minutes, but the signed URL is valid for 15 minutes).

app/lib/package/backend.dart
```diff
@@ -410,7 +410,11 @@ class PackageBackend {
     InvalidInputException.checkSemanticVersion(version);
     final cv = canonicalizeVersion(version);
     final object = 'latest/api/archives/$package-$cv.tar.gz';
-    return Uri.parse(_exportedApiBucket.objectUrl(object));
+    return await uploadSigner.buildDownloadUrl(
+      activeConfiguration.exportedApiBucketName!,
```
sigurdm: Two things here:

1. **Encapsulation:** `PackageBackend` already has `this._exportedApiBucket` injected. Consider using `_exportedApiBucket.bucketName` instead of `activeConfiguration.exportedApiBucketName!`.

2. **IAM permissions:** Testing this on `staging-sigurdm` (`/api/archives/cancellable-2.3.0.tar.gz`) currently fails with HTTP 403 `AccessDenied`:
   ```
   package-uploader-signer@dartlang-pub-dev.iam.gserviceaccount.com does not have storage.objects.get access to the Google Cloud Storage object.
   ```
   GCS verifies that the signer identity in `X-Goog-Credential` actually has read permissions on the object. `package-uploader-signer` needs `roles/storage.objectViewer` on `dartlang-pub-dev-exported-api` (and prod `dartlang-pub-exported-api`). We should also update the doc comment on `Configuration.uploadSignerServiceAccount` to mention this requirement.

app/lib/package/upload_signer_service.dart
```diff
@@ -97,6 +104,69 @@ abstract class UploadSignerService {
...
+      'X-Goog-Expires': '${lifetime.inSeconds}',
+      'X-Goog-SignedHeaders': 'host',
+    };
+
+    final canonicalQueryString = canonicalQueryParameters.entries
+        .sortedBy((e) => e.key)
+        .map((e) => '${e.key}=${Uri.encodeComponent(e.value)}')
+        .join('&');
+
+    final canonicalUri = '/$bucket/$object';
```
sigurdm: What happens when package versions have build metadata containing `+` (e.g. `foo-1.0.0+1.tar.gz`)?

In the GCS V4 signing spec:
> *When defining the resource path, you must percent encode the following reserved characters: `?!#$&'()*+,:;@[]"`*

Here `canonicalUri` keeps a literal `+` and passes it both to the canonical request and `Uri(path: canonicalUri)`. If a client, proxy, or CDN encodes `+` as `%2B` when following the redirect, GCS calculates the canonical request with `%2B` and rejects the request with `403 SignatureDoesNotMatch` (tested and confirmed against GCS).

Reserved characters like `+` should be percent-encoded in both the canonical URI and the returned `Uri`.

app/lib/package/upload_signer_service.dart
```diff
@@ -97,6 +104,69 @@ abstract class UploadSignerService {
...
+
+    final canonicalRequest = [
+      'GET',
+      canonicalUri,
+      canonicalQueryString,
```
sigurdm: The embedded `\n` in `'host:storage.googleapis.com\n'` works because it creates the required blank line after canonical headers when joined with `\n`, but it looks like a typo.

Consider separating it explicitly so the structure is obvious:
```dart
final canonicalHeaders = 'host:storage.googleapis.com\n';
final canonicalRequest = [
  'GET',
  canonicalUri,
  canonicalQueryString,
  canonicalHeaders,
  'host',
  'UNSIGNED-PAYLOAD',
].join('\n');
```

app/lib/fake/backend/fake_upload_signer_service.dart
```diff
@@ -30,6 +33,15 @@ class FakeUploadSignerService implements UploadSignerService {
     );
   }
 
+  @override
+  Future<Uri> buildDownloadUrl(
+    String bucket,
```
sigurdm: Because `FakeUploadSignerService` completely bypasses `UploadSignerService.buildDownloadUrl`, none of our unit/integration tests actually execute the signing algorithm.

Could we add a unit test for `UploadSignerService.buildDownloadUrl` (using a mock `sign` implementation) to verify the date format, query string sorting, canonical request hashing, and URL encoding against known test vectors?

app/lib/frontend/handlers/pubapi.dart
jonasfj: > If GCLB continues to serve /api/archives/* directly via a backend bucket, a non-public bucket will reject unauthenticated requests unless Cloud CDN signed URLs/cookies are configured in GCLB.

What nonsenss is this? we just workaround this by granted read access to the GCLB service account.

jonasfj: > If GCLB is instead updated to route archive requests to the backend service, every archive download on pub.dev will hit this handler. 

Yes this would not scale.

app/lib/package/backend.dart
jonasfj: :see_no_evil: I thought I had tested this? maybe bucket permission changes hadn't gone through.

But yes, you're write package-uploader-signer doesn't have sufficient permissions, and giving them to it would be a bit of a hack.

app/lib/frontend/handlers/pubapi.dart
jonasfj: I do wonder if there is a risk any of this gets cached in redis..

jonasfj: I think we have to accept it won't scale, or we return an error code. Or we redirect to staging.pub.dev or pub.dev, but that also risk creating a redirect cycle.

So I thought this was the best option, not that I like any of the options.

app/lib/package/backend.dart
```diff
@@ -410,7 +410,11 @@ class PackageBackend {
...
+      object,
+      Duration(minutes: 15),
```
sigurdm, line 417: Is there a reason to keep the lifetime at 15 minutes?

GCS V4 signed URLs support a lifetime of up to **7 days** (`Duration(days: 7)` / 604,800 seconds). Because pub package archives are public open-source code rather than confidential user data, a longer lifetime (e.g. up to 7 days) would allow us to cache the signed URLs in Redis or memory for days at a time.

That would eliminate almost all outbound `signBlob` IAM RPCs and the associated ~100ms latency on download redirects. Note that even with a multi-day signature, if a package version is deleted or moderated, GCS returns `404 NoSuchKey` immediately.

app/lib/package/download_signer_service.dart
```diff
@@ -0,0 +1,150 @@
...
+    );
+
+    final stringToSign = [
+      'GOOG4-RSA-SHA256',
+      datetime,
+      scope,
+      canonicalRequestHash,
+    ].join('\n');
+
+    final result = await sign(utf8.encode(stringToSign));
```
sigurdm, line 110: An alternative to calling `iam.signBlob` over HTTP for download URLs is to sign locally using a GCS HMAC key (`GOOG4-HMAC-SHA256`).

GCS natively supports V4 signed URLs using HMAC keys associated with service accounts ([docs](https://cloud.google.com/storage/docs/authentication/signatures#signing-process)).

Instead of awaiting the `iam.signBlob` RPC (which adds 50–150ms of latency per redirect), we could:
1. Create an HMAC key for the service account (`gcloud storage hmac create <service-account>`).
2. Store the secret in Secret Manager.
3. Derive the signing key and calculate the signature locally in Dart using `package:crypto`:

```dart
final kDate = Hmac(sha256, utf8.encode('GOOG4$secret')).convert(utf8.encode(date)).bytes;
final kRegion = Hmac(sha256, kDate).convert(utf8.encode('auto')).bytes;
final kService = Hmac(sha256, kRegion).convert(utf8.encode('storage')).bytes;
final signingKey = Hmac(sha256, kService).convert(utf8.encode('goog4_request')).bytes;

final signature = hex.encode(Hmac(sha256, signingKey).convert(utf8.encode(stringToSign)).bytes);
```

The Canonical Request and String-To-Sign construction remain identical; only `X-Goog-Algorithm` becomes `GOOG4-HMAC-SHA256` and `X-Goog-Credential` uses the HMAC `accessId`.

This would reduce the signing latency to < 0.1ms with zero network RPCs or IAM quotas. Worth considering if archive redirects are going to be served by the backend.

app/lib/package/backend.dart
jonasfj: We don't cache these URLs, I don't think w eod, and yes we could but why?

Afaik nothing will hit this, the only thing that realistically could want to is workers, but even they will get a different `archive_url` as I recall.

So this is only like a fallback. And yes, I don't think this will handle us disabling the GCS bucket in GCLB well, maybe, who knows, I wouldn't want to try :D

---------

Importantly: we don't cache these today, I am not convinced it's worth the hassle to wire up caching.

app/lib/package/download_signer_service.dart
jonasfj: this is the AWS compatibility right, aws signature algorithm v4?

Not GCP v4 signatures.

This means long lived keys, we could, but it's not important for anything is it? Why go through the hassle of having long lived keys, if we even can in the projects we have.

sigurdm (approved):
After understanding this better - I think we could perhaps move to a model where the server doesn't accept exported-api requests (or perhaps it just redirects to the urls where we have the GCLB in front of)
I don't think we gain much by having these workarounds for whenever something is misconfigured. We are in a quite bad state when stuff is misconfigured anyways.

jonasfj:
Closing in favor of https://github.com/dart-lang/pub-dev/pull/9572/changes

PR #9569, opened 2026-09-03
Do snapshot preparation in isolate event-loop starvation

sigurdm:
The search snapshot JSON has grown significantly (currently ~156 MB compressed / >600 MB uncompressed JSON across hundreds of thousands of packages).

During periodic snapshot creation (`searchBackend.updateSnapshotInForeverLoop`), querying Datastore, building the ~600 MB document map, and serializing it previously ran on the main isolate of `analyzer` instances via `scheduleMicrotask`. This caused severe CPU and garbage collection pressure and blocked the single-threaded event loop, leading to timed-out health checks (`/liveness_check` and `/readiness_check`), IGM auto-healer SIGKILLs (exit code 137), and abandoned task locks.

This PR moves search snapshot creation and upload into a dedicated worker isolate using `startWorkerIsolate` on `analyzer` instances:
- **Event loop & GC isolation**: Snapshot building and serialization run completely independent of the main isolate serving health checks and task processing.
- **Zero inter-isolate overhead**: The worker isolate queries Datastore directly and uploads the archive directly to Cloud Storage (`_snapshotStorage.uploadDataAsJsonMap`), requiring no large inter-isolate message passing.
- **Lifecycle management**: Registers `snapshotWorker.close` with `registerScopeExitCallback` for clean termination on shutdown.

### DIFF
--- a/app/lib/service/entrypoint/analyzer.dart
+++ b/app/lib/service/entrypoint/analyzer.dart
@@ -18,6 +18,7 @@ import '../../task/backend.dart';
 import '../../tool/neat_task/pub_dev_tasks.dart';
 
 import '../download_counts/backend.dart';
+import '_isolate.dart';
 
 final Logger logger = Logger('pub.analyzer');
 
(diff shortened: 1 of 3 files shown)


jonasfj:
Any urgency here? I think we should consider migration to `.jsonl` instead, if this is viable.

But also I think this is supposed to be going on in a background isolate, I could be wrong there.

sigurdm:
I suspect this is why many of our scheduled tasks don't finish in time, because the analyzer instances gets recycled...
But well I guess it can wait.

isoos:
We do use isolates for the snapshot loads (on the search instances), but I don't recall if we use them for snapshot building (on the analyzer instances). It would be worth to run that entirely in a separate isolate.

isoos:
I think the long term solution here is to continue the migration of the container format from JSON to SQLite, but a quick fix is okay here.

sigurdm:
Ok, updated to use a worker isolate instead. Maybe we should do both, as this still synthesizes the entire thing in memory.

jonasfj (approved):

PR #9572, opened 2026-09-08
feat: Redirect production downloads directly to primary host

jonasfj:


sigurdm (approved):
This is a much cleaner and simpler approach than #9564. Relying on GCLB to serve directly from the private bucket via backend bucket permissions avoids the complexity, latency, and quota overhead of `signBlob`.

Two suggestions:

1. **Self-redirect loop guard:** If a request reaches App Engine on `pub.dev` or `staging.pub.dev` (e.g. via GCLB misconfiguration, proxying, or host spoofing), redirecting to `primarySiteUri` redirects to itself and creates an infinite loop (`ERR_TOO_MANY_REDIRECTS`). In `PubApi.fetchPackage`, consider short-circuiting with a 400:
   ```dart
   final redirectUrl = await packageBackend.downloadUrl(package, version);
   if (request.requestedUri.host == redirectUrl.host &&
       request.requestedUri.path == redirectUrl.path) {
     return Response.badRequest(
       body: 'Direct archive requests on this host are not supported.',
     );
   }
   ```

2. **Reuse `urls.pkgArchiveDownloadUrl`:** In `PackageBackend.downloadUrl`, consider using:
   ```dart
   return Uri.parse(urls.pkgArchiveDownloadUrl(
     package,
     cv,
     baseUri: activeConfiguration.primarySiteUri,
   ));
   ```
   This ensures versions containing build metadata (e.g. `+`) are percent-encoded consistently with the version listing API.

PR #9587, opened 2026-09-17
Support SANDBOX_OUTPUT_JSON in sandbox_runner

sigurdm:
Support `SANDBOX_OUTPUT_JSON` (a JSON-encoded list of directory paths) in `sandbox_runner` alongside legacy colon-separated `SANDBOX_OUTPUT`.

- `parseSandboxOutput` prefers `SANDBOX_OUTPUT_JSON` when non-empty and falls back to `SANDBOX_OUTPUT`.
- `pub_worker_subprocess.dart` passes both `SANDBOX_OUTPUT` and `SANDBOX_OUTPUT_JSON` during the transition.

Corresponding `pana` PR: https://github.com/dart-lang/pana/pull/1636


pkg/pub_worker/bin/pub_worker_subprocess.dart
```diff
@@ -102,7 +102,7 @@ Future<void> main(List<String> args) async {
     environment: {
       'PUB_CACHE': unpackPubCacheDir.path,
       'PUB_HOSTED_URL': pubHostedUrl,
-      'SANDBOX_OUTPUT': tempDir,
+      'SANDBOX_OUTPUT_JSON': json.encode([tempDir]),
```
isoos, line 106: I think we should use both for the time being, passing the same values. We don't gain much by being too strict on the either-or condition, since we control both parts, but we lose the easy ability to migrate back if needed (one more knob to turn). Once we have everything migrated and stable, we could clean it up though.

sigurdm: Done — `pub_worker_subprocess.dart` now passes both `SANDBOX_OUTPUT` and `SANDBOX_OUTPUT_JSON`, and `parseSandboxOutput` prefers `SANDBOX_OUTPUT_JSON` when both are set.

isoos (approved):

